PR #51, opened 2018-03-26
Radcliffe 2: fix fatal error when query categories and tags

lancewillett:
Sometimes WordPress returns a `WP_Error` object and not a string when querying for terms like tags and categories. The PHP fatal error in the server logs looks like this:

```
Catchable fatal error for wp-content/themes/pub/radcliffe-2/inc/template-tags.php:78 - Object of class WP_Error could not be converted to string #theme
```

This change applies a safety check to both `get_the_tag_list()` and `get_the_category_list()` for Radcliffe 2 — checking for the error before attempting to display the result as a string.

### DIFF
--- a/radcliffe-2/inc/template-tags.php
+++ b/radcliffe-2/inc/template-tags.php
@@ -66,14 +66,14 @@ function radcliffe_2_entry_footer() {
 	if ( 'post' === get_post_type() ) {
 		/* translators: used between list items, there is a space after the comma */
 		$categories_list = get_the_category_list( esc_html__( ', ', 'radcliffe-2' ) );
-		if ( $categories_list ) {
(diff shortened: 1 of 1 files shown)


thomasguillot:
LGTM 👍

thomasguillot:
@lancewillett do you think we should add this to _s? https://github.com/Automattic/_s/blob/master/inc/template-tags.php#L63

lancewillett:
@thomasguillot I'll leave it up to you; I think this is something that happens more often on WP.com sites (huge multisite install) and less likely in other environments.

Another suggestion would be to think about changing this in core WP — template functions used in themes should probably not require error handling. They should return an empty string instead.

PR #54, opened 2018-03-28
Shoreditch: Fix main stylesheet when using child themes

miroslavrussev:
When trying to extend Shoreditch using a child theme the parent style.css isn't loaded correctly. 
Instead it loads the style.css of the child theme twice.
The reason for that is when using a child theme `get_stylesheet_uri()` returns the URI of the **current** theme stylesheet:
https://codex.wordpress.org/Function_Reference/get_stylesheet_uri
In this case it returns the stylesheet of the child theme.

Change the way that the parent stylesheet is enqueued so that child themes can load it correctly instead of loading their own stylesheet twice.

#### Related issue(s):

I couldn't find any but the issue is kind of obvious.

### DIFF
--- a/shoreditch/functions.php
+++ b/shoreditch/functions.php
@@ -179,7 +179,7 @@ function shoreditch_scripts() {
 
 	wp_enqueue_style( 'shoreditch-fonts', shoreditch_fonts_url(), array(), null );
 
-	wp_enqueue_style( 'shoreditch-style', get_stylesheet_uri() );
+	wp_enqueue_style( 'shoreditch-style', get_template_directory_uri() . '/style.css' );
 
 	wp_enqueue_script( 'shoreditch-back-top', get_template_directory_uri() . '/js/back-top.js', array(), '20120206', true );
 


miroslavrussev:
Is there a particular reason for **not** fixing this?

laurelfulford:
Hi @miroslavrussev!

When following the instructions [here](https://developer.wordpress.org/themes/advanced-topics/child-themes/) to create a child theme from the Shoreditch theme, I can't recreate this issue with the child theme's stylesheet loading twice.

From that page, adding the following to the child theme's functions.php to load the styles:

```
<?php
add_action( 'wp_enqueue_scripts', 'my_theme_enqueue_styles' );
function my_theme_enqueue_styles() {
    wp_enqueue_style( 'parent-style', get_template_directory_uri() . '/style.css' );
 
}
?>
```

... loads the parent theme's and child theme's style.css one time each, like:

image

Because of this, I'm going to close this PR. But if you have any further questions, please let us know!

Issue #87, opened 2018-04-09
Multiple Themes: WooCommerce Products per row, Rows per page settings missing in the Customizer


danieldudzic:
> I'm not able to set **Product per row** or **Rows per page** with Dara theme in **Customizer > WooCommerce > Product Catalog**:

> ![Screenshot](http://cld.wthms.co/OBwUPr+)
> Screenshot: http://cld.wthms.co/OBwUPr

> With default themes like Storefront or 2017 these settings show just fine:

> ![Screenshot](http://cld.wthms.co/VzoQph+)
> Screenshot: http://cld.wthms.co/VzoQph

> Dara still uses pre-[WC 3.3](https://woocommerce.wordpress.com/2017/12/09/wc-3-3-will-look-great-on-all-the-themes/) legacy filters for columns and rows and needs to be updated to support the new Customizer options.

This has now been addressed in Dara in https://github.com/Automattic/themes/pull/85

However, the issue occurs in multiple themes:

* Radcliffe 2
* Independent Publisher 2
* Dyad 2
* Libre 2
* Karuna
* Pique
* Lodestar

PR #88, opened 2018-04-09
Radcliffe 2: Adds support for the 3.3+ new Customizer options

danieldudzic:
* Fixes WooCommerce products column layout
* Adds support for the 3.3+ new Customizer options (See #87)
* Fixes flash sale badge styling

### DIFF
--- a/radcliffe-2/inc/woocommerce.php
+++ b/radcliffe-2/inc/woocommerce.php
@@ -15,7 +15,18 @@
  * @return void
  */
 function radcliffe_2_woocommerce_setup() {
-	add_theme_support( 'woocommerce' );
+	add_theme_support( 'woocommerce', apply_filters( 'radcliffe_2_woocommerce_args', array(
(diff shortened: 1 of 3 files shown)


danieldudzic:
@allancole for your reference:

Original styling:
previous

New styling:
new

Explanation of the issue:

> There is a problem when it comes to having the Sale badge on the top right (or left) corner of the product thumbnail when you have a variable number of columns available.

> That is when the product content is centered.

> There isn’t really a great way to position the badge accurately because of the position: absolute
so if the theme has centered contents, the “flying” badge needs re-styling.

Gif showcasing the issue with the original styling:

screencast

allancole:
@danieldudzic & @jeffikus Yeah I think this design change is appropriate. I actually think it fits theme a bit better than what I did originally ;-) . My only concern would be that it might feel like a disruption for current users running a store with Radcliffe 2. It’s a small disruption but probably something worth noting for HE‘s.

PR #111, opened 2018-04-27
Pique: Updated SHOW MORE nav item to open on click

### LINKED ISSUE #109: Pique: "show more" nav not expanding on desktop
## Steps to replicate

View [theme demo](https://piquedemo.wordpress.com) and shrink the browser window down to  under 768px in width, where the "Show More" (...) menu is visible _with_ the background. Click and/or hover over the "show more" menu to ... show more menu items.

## Result

Nothing happens. Menu does not show on hover as expected. Note that it _does_ show up when the "Shoe More" menu is visible and the viewport is _above_ 768px (where the menu background is not visible).

## Expected

Expected the hidden navigation items to show on hover in desktop, even on smaller viewports.

Note that, on an actual mobile device, this works as expected, as it appears to be listening to the click event there.

Tested in Chrome, Firefox, Safari, Edge (latest - Windows 10), IE11 (Windows 10).

danielwrobert:
This listens for either a click event or a touch event, depending on the user's device, to toggle the 'Show More' navigation. This will no longer open on hover, to remain consistent with the behavior on all viewport widths.

Fixes #109.

### DIFF
--- a/pique/assets/js/navigation.js
+++ b/pique/assets/js/navigation.js
@@ -58,11 +58,9 @@
 			}
 		};
 
-		// if ( window.matchMedia( "( max-width: 767px )" ).matches ) {
-			for ( i = 0; i < parentLink.length; ++i ) {
-				parentLink[i].addEventListener( ( 'ontouchstart' in window ? 'touchstart' : 'click' ), toggleNav, false );
-			}
-		// }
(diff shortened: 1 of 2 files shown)


laurelfulford:
This works great @danielwrobert -- thanks for looking into it!

Issue #108, opened 2018-04-25
Multiple Themes: Use of `$(window).load()` clashes with ads


laurelfulford:
We've had a few reports of sites using Rowling and Rebalance not loading when ads were enabled, and Goran not loading completely (header isn't visible). 

The issue looks to be related to these themes using `$(window).load()` to show the header, or full content, in these themes. It's not working well with something in the ads code. 

Swapping to `$(document).ready()` seems to fix the above issue, but may have unintended consequences. So it will need to be tested with each theme carefully to make sure it's a good replacement, and if no, look for other alternatives.

More details and related tickets here: pNEWy-bRB-p2

(Edited to add) Affected themes (there may be more):
* Rowling ✅ 
* Rebalance ✅ 
* Goran ✅ 
* Baskerville 2 ✅ 
* Textbook ✅ 
* Illustratr ✅ 
* Altofocus ✅

KokkieH:
I commented on the P2, but in case you missed it, I've also observed the issue with Baskerville-2, Pinboard, Textbook, Illustratr, and Altofocus thus far.

laurelfulford:
Thanks @KokkieH! I was just coming here to add those after I realized I'd missed them from the thread 😅

allykmyta:
Another reported issue: #1125244-zen 
Theme: Rowling

laurelfulford:
Patch for Rowling here: D12613-code. I've removed the part where the content is hidden in the first place, because it doesn't seem needed in this theme. I'd appreciate your thoughts on this one, @allancole, if you think this change is over-reaching!

laurelfulford:
Patch for Baskerville 2 here: D12659-code.

laurelfulford:
Patch for Goran here: D12660-code

laurelfulford:
For Illustratr's fix (#114), it addresses the missing content issue, but there is other JavaScript (like the bit to make images wider than the content) that doesn't seem to work the same with `$(document).ready()`. I've left them as is for now since the pressing problem was the projects not displaying; we can circle back as need be.

laurelfulford:
I'm going to close this because all of the listed themes have been patched. Will follow up in the related P2 threads; I've also added a comment to #1125244-zen.

gpmw:
Reporting another instance of this here with the Oxygen theme.
#1135929-zen

rclations:
Is it possible to patch Cubic as well?

User report: C03TY6J1A/p1539719524000100-slack-livechat

[...]

PR #114, opened 2018-05-01
Illustratr: Replace window.load event with document.ready

laurelfulford:
Replace `$(window).load()` event with `$(document).ready()` for functions that show hidden elements on the page, to help work around ads issue. 

#### Related issue(s):
See #108

### DIFF
--- a/illustratr/js/portfolio.js
+++ b/illustratr/js/portfolio.js
@@ -28,11 +28,9 @@
 				} );
 			}
 		} );
-
 	}
 
 	$( window ).load( function() {
-
 		/*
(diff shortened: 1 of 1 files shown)


laurelfulford:
This one still needs work; I missed a couple `$(window).load()`s, in at least one case, the `$(document).ready()` replacement isn't working.

laurelfulford:
On second thought, I'm going to go ahead with this fix -- it addresses the hidden content. We can circle back on other things that don't work exactly right -- like the images being outset -- as needed.

PR #139, opened 2018-05-25
Add Gutenberg blocks XML file to design-toolkit directory.

allancole:
Creating a new `/tools` directory to house common development resources starting with a Gutenberg export file. The export data contains a single post that displays all core Gutenberg blocks for testing.


allancole:
This should probably be moved to the Design toolkit directory: https://github.com/Automattic/themes/tree/master/design-toolkit

allancole:
This should probably be moved to the Design toolkit directory: https://github.com/Automattic/themes/tree/master/design-toolkit

laurelfulford:
> This should probably be moved to the Design toolkit directory

That sounds like a good idea! :) Otherwise, this looks good!

Issue #276, opened 2018-10-05
Independent Publisher 2: Current page drop-down menu items are invisible on mobile


sjaved87:
When visiting a page on mobile, if it is in the menu, its subpages will not be visible. 

## Steps to replicate

1. Visit the site demo https://independentpublisher2demo.wordpress.com on a mobile device.

2. Open the Menu; note the items listed (Home, About me, Photography on the top level, with Adventures with an SLR camera as a sub menu of Photography)

3. Click on Photography.

4. Reopen the menu.

## Result

See, the child/submenu items of the current page are hidden because of the white color.

http://cld.wthms.co/bBWVA6

The culprit:
```
 .main-navigation>div>ul>li.current-menu-item>ul>li a, .main-navigation>div>ul>li.current_page_item>ul>li a {
    color: #fff;
}
```

## Expected

It should display the sub-menu items. Like this -> http://cld.wthms.co/VVp82a

* User report: #1465160-zen
* Follow-up ticket: #1465160-zen
* User's site: https://sanatanadhara.wordpress.com/ Issue can't e viewed there. Because as a workaround I gave them following CSS code. 

* However, you can see the issue on my test site: https://sjtesterblog.wordpress.com/about/
```
 .main-navigation>div>ul>li.current-menu-item>ul>li a, .main-navigation>div>ul>li.current_page_item>ul>li a {
    color: #00aadc;
}
```

shreeshylahs:
May i know was the above mentioned css manually added through Dashboard?

sjaved87:
It looks like this one:

```
 .main-navigation>div>ul>li.current-menu-item>ul>li a, .main-navigation>div>ul>li.current_page_item>ul>li a {
    color: #fff;
}
```

Is coming from theme style.css.

I gave the following to add through dashboard to fix/workaround:

```
 .main-navigation>div>ul>li.current-menu-item>ul>li a, .main-navigation>div>ul>li.current_page_item>ul>li a {
    color: #00aadc;
}
```

sjaved87:
I see, thank you and I am glad we were able to figure out. 👍

shreeshylahs:
I think there might be many other places which would need a fix.

laurelfulford:
Committed to WP.com in 53224-wpcom-themes.

PR #277, opened 2018-10-09
Added media query for the drop-down font fix (Git Issue #276)

shreeshylahs:
Added media query for mobile screen to fix the font color of the drop down in mobile view

#### Related issue(s):
#276


laurelfulford:
Thanks for this PR, @shreeshylahs!

I made a slightly different version of it in #534, just to remove the extra media query -- I moved the styles around a bit to make sure they're applied at the same breakpoints. We appreciate your help here!

Issue #330, opened 2018-11-09
Independent Publisher: Hebrew Text overlaps in Author Bio


damiannep:
## Steps to replicate
Create a page in Hebrew with author bio enabled. Then open the page and scroll down to the Author Bio enabled on single posts. 

## Result

The text doesn't render properly e.g. https://cld.wthms.co/8eR0eL

## Expected

The text renders correctly in other themes such as Dara: https://cld.wthms.co/CYDLV7

* User report: #7924341-hc
* Follow-up ticket: #1555074-zd
* User's site: https://adam-hofshi.blog/2018/11/09/%D7%90%D7%93%D7%9D-%D7%97%D7%95%D7%A4%D7%A9%D7%99-%D7%A4%D7%95%D7%A1%D7%98-%D7%A8%D7%90%D7%A9%D7%95%D7%9F-%D7%91%D7%9C%D7%95%D7%92-%D7%90%D7%93%D7%9D-%D7%97%D7%95%D7%A4%D7%A9%D7%99/

nagpai:
* The site-posted-on  div seems to remain right floated .. may need to change to left since this is a RTL language

* The Avatar may need to be given a top margin of 20px

Added the following custom CSS

```
@media screen and (min-width: 48.9375em) {
.site-posted-on {
   left: auto;
   right: 0;
}

.entry-author .author-avatar {
   margin-top: 20px;
  }
}
```
Trying to submit my first PR for this :)

nagpai:
I have some new information after my first futile attempt at a PR :-P 

**tl;dr -** The issue seems to be with the broken link of the Photon CDN served RTL CSS file 

After realizing that the CSS rules in the theme's source code rtl.css are correct, my eyes fell on the inspector output for errors

<img width="1013" alt="screenshot 2018-11-10 at 8 36 34 pm" src="https://user-images.githubusercontent.com/4162931/48302871-79388580-e528-11e8-8a2e-0aa1172648aa.png">

It seems that the theme CSS is being hosted by Jetpack Photon CDN which doesnt seem to have the rtl.css on it. 

The URL 
https://s2.wp.com/wp-content/themes/pub/independent-publisher-2/rtl.css?m=1541537380

seems to be broken and thats why the RTL styles are not loading. 

This doesnt seem to be a theme issue, rather its a Jetpack Photon issue. I will check where is the best place to report this

nagpai:
[...]

PR #332, opened 2018-11-10
Independent Publisher : Changes to author box CSS in rtl.css

nagpai:
Correcting the Avataar and site-posted-on div margins and alignment for RTL languages.

Edited the rtl.css to move .site-posted-on info to the left and give a margin to the Avataar. This is a suggested fix for the overlap within the author box noted in the issue for RTL languages.

#### Related issue(s):
#330

### DIFF
--- a/independent-publisher-2/rtl.css
+++ b/independent-publisher-2/rtl.css
@@ -601,8 +601,8 @@ ol.comment-list .trackback {
 	}
 
 	.site-posted-on {
-		left: auto;
-		right: 0;
+		left: 0;
+		right: auto;
 		margin-right: 0;
 		text-align: left;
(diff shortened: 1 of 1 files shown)


nagpai:
### Before Changes : 

screenshot 2018-11-10 at 12 42 47 pm

### After Changes (tested in Customizer >> CSS) :

screenshot 2018-11-10 at 12 39 29 pm

nagpai:
I need to do some more tests on a local instance. This could be an issue beyond the theme, since the theme runs fine when installed on a local test site outside of WordPress.com environment. ( sidebar appears on right if enabled), whereas on WP.com it appears on left (as though some of the RTL rules are not getting applied)

Issue #357, opened 2018-11-16
Radcliffe 2&Gutenberg: Wide width tables is left aligned


Robertght:
## Steps to replicate

[Please include screenshots, browser information, and anything else that can help us recreate the problem.]

1. Insert a table on the page/post via Gutenberg
2. Select the wide width option from the alignment bar
3. It gets aligned to the left due to the following CSS rule:
```
.wp-block-table.alignwide {
margin-left: calc(25% - 23vw);
margin-right: calc(25% - 23vw);
}
```
## Result
![image](https://user-images.githubusercontent.com/6737345/48648386-d3898880-e9f7-11e8-9c33-2814ec76d8a9.png)

## Expected
The table has to be centered.

Temp fix:

```
.wp-block-table.alignwide {
  margin: 0;
}
```

* User report: Slack-/C4WC5LYM6/p1542366377832200
* User's site: https://lamaisondescedres.net/packages/

laurelfulford:
This has been fixed with the changes in #359.

I removed the CSS fixing the issue on the user's site, since it paired with the fix caused more issues :) 

@Robertght Can you please follow up with the user? I'm not sure if there's an associated ticket, the Slack reference above only includes the time stamp, not the channel, so I can't check the discussion there. Thanks!

Robertght:
Thanks Laurel! Forgot to add it here too. Follow-up: #1578596-zen

laurelfulford:
Great! Thank you, @Robertght! 😄

PR #359, opened 2018-11-16
Radcliffe 2: Clean up Gutenberg support

laurelfulford:
* Reorders checks for which editor colour palette to load; with the original set up, the default palette was never loaded. 
* Reorders stylesheets to match other themes.
* Removes unneeded editor styles.
* Makes Cover block only wide when the wide and full alignments are applied.
* Fix issues with table full and wide widths. 
* Add updated versions of Gutenberg CSS selectors.


laurelfulford:
This should fix the issue described in #357, but I'll double-check once the changes are on WP.com and I can test the user's site.

Issue #369, opened 2018-11-20
Libretto: Site Logo Link Overlaps Menu Items


mdrockwell:
## Steps to replicate

- Change a site's theme to Libretto
- Add a site logo in the Customizer
- Add a handful of links to the menu in the header

## Result

When viewing the site in at a laptop-sized viewport width (I tested at 1440 wide) the link on the site logo overlaps with the menu, making some menu items un-clickable.

Tested in Safari, Chrome, and Firefox on macOS Mojave.

## Expected

The site logo link is constrained to only be clickable on the image itself
or
The menu switches to the mobile menu when the two items overlap.

* User report: 1576392-zen
* User's site: https://demonnier.com/ (visible on homepage)

## Screenshot/GIF

![screen recording 2018-11-20 at 9 52 15 am mov](https://user-images.githubusercontent.com/42454301/48781519-2db17480-ecaa-11e8-80af-927e4f1b65ad.gif)

<img width="1552" alt="screen shot 2018-11-20 at 9 54 43 am" src="https://user-images.githubusercontent.com/42454301/48781617-5a658c00-ecaa-11e8-934f-51194a264486.png">

laurelfulford:
This has been committed to WP.com in r53281-wpcom-themes.

@mdrockwell It looks like the original ZD ticket is closed, so I can't add a comment there - can you please follow up with the user if needed? Thanks!

mdrockwell:
Thanks for the heads up @laurelfulford, I created a follow-up ticket and let them know there.

1787595-zen

PR #370, opened 2018-11-20
Libretto: Prevent Title Block Link Overlap with Menu

torres126:

This change should mean that only the image and the site title will redirect, allowing for the users to visit the intended pages by clicking their names on the menu. I've tried to explain it in this image here: https://prnt.sc/lkvaj0

#### Related issue(s):

#369

### DIFF
--- a/libretto/style.css
+++ b/libretto/style.css
@@ -823,6 +823,7 @@ body:not(.libretto-has-header-image) .title-block {
   left: 20%;
   position: absolute;
   text-shadow: 0 1px 0 #000;
+  width: 60%;
   z-index: 1;
 }
 
--- a/libretto/style.css
+++ b/libretto/style.css
@@ -772,6 +772,7 @@ body {
 /* Style page title blocks */
 .title-block {
   text-align: center;
+  pointer-events: none;
 }
 
 .title-block h1 {
@@ -871,12 +872,8 @@ body:not(.libretto-has-header-image) .title-block {
   padding: 0;
 }
 
-:not(.libretto-has-header-image) .title-block {
-  pointer-events: none;
-}
-
-img.site-logo.attachment-libretto-logo,.site-title {
-pointer-events: all;
+.site-logo.attachment-libretto-logo,.site-title {
+  pointer-events: all;
 }
 
 /*--------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


laurelfulford:
Thanks for the patch @torres126!

I think a better approach might be to get the link out of the way, rather than changing it's behaviour -- I'm a little worried that could be confusing down the road for other people working on the theme.

I _think_ switching `.title-block .site-logo-link` to `display: inline-block` would be one option, though I need to test a bit more to make sure it won't cause other issues. Let me know what you think!

torres126:
Hmm, maybe it's just me, but though that seems to stop the menu items linking to the home page, it seems to leave them unclickable completely? Agreed though, there's probably a better solution to this.

torres126:
No luck in managing to align things yet I'm afraid. Added a commit (still a WIP) showing a possible solution, but it's pretty much useless without figuring an alignments. I'm wondering if there's even a better way to do this without touching the widths, but that does seem to be the problem here. My worry is it's the best solution too.

Out of curiosity though, what about touching the margins [here](https://github.com/Automattic/themes/blob/master/libretto/style.css#L2645)? If the icon was lowered beneath the menu, it'd solve both this issue but also more importantly the issue of menus overlapping with the site icon on tablets, and therefore being invisible. But then again, it's a pretty notable change.

torres126:
I did a bit more testing and I think that this PR should work, since it centre aligns the title block without resorting to having it take up the full width, which isn't ideal as it's a link that overlaps the menu. It does mean line breaks will be after less characters now though, it seems.

laurelfulford:
Thanks @torres126! Oof, this is a weird one. 

>  It does mean line breaks will be after less characters now though, it seems.

Thanks for catching that. That could be problematic, since this theme is already on live sites. 

What do you think about leaving the width as is, and instead give the menu a slightly higher z-index -- like: 

```
#site-navigation .menu {
  position: relative;
  z-index: 2;
}
```

This does mean a really long menu will run in front of the logo instead of behind it, but both aren't great, and I'm not sure there's a way around it that doesn't involve completely redoing the header (and likely changing how it looks).

torres126:
Oops, I forgot about this one! But yep, there's probably no ideal way to address this, though that may just be the best option. Confirmed it works and solves the issue.

laurelfulford:
Looks good! Thanks for all your work on this one, @torres126 - it was a doozy!

Issue #361, opened 2018-11-18
Libre 2: Scrolling Issue with Pages


torres126:
## Steps to replicate

1. Activate the theme "Libre 2" on a WordPress.com blog
2. View a page on the website, provided it is not the posts page
3. See issue

## Result

You'll see a scrolling bar at the bottom of the browser. When you scroll down, this bar will disappear, but this only seems to occur on pages which are not the posts page: https://prnt.sc/ljvibc

## Expected

The website shouldn't have a scrolling bar appear, and it should be as if you scrolled down. 

* Using Chrome 49.0.2623.112
* Reported in the forums: https://en.forums.wordpress.com/topic/libre-2-needs-an-urgent-fix-or-update/
* User's sites (also recreated on my test site): https://dreambacoli.wordpress.com/, https://crossdressingtales.wordpress.com/

KokkieH:
Thanks for the report.

@laurelfulford, any chance this could be related to the recent changes for Gutenberg support on this theme?

laurelfulford:
@KokkieH Yikes, yes, this change is related to those updates. Thanks for the report, @torres126!

torres126:
Thanks! :)

KokkieH:
Thanks Laurel.

Congrats on your first PR here, @torres126 :)

PR #407, opened 2018-12-04
Libre 2: Fix Scrolling on Tablets

torres126:

It seems that there's still a scrolling issue with the Libre 2 theme that wasn't fixed in #362, and that can be seen by visiting a site using the Libre 2 theme on a tablet, or setting your device's dimensions to that of a tablet (tested using 1440px, and can also see on my tablet). I'm not too sure if this is worth creating a new issue about, because it seems very similar to #361, and you can also see this bug on the sites linked in that issue. 

It seems that there should be a simple fix though, and unless I'm missing something (please let me know if I am!), this shouldn't change anything that it isn't supposed to, including on a desktop. 

**Before:**

gsdfgsfdsdfg

**After:**

dfgsgdfsdfsg

I'm able to recreate this on multiple sites, and the same fix seems to work locally for me. If I'm missing something though, please let me know! 

(cc @laurelfulford)

### DIFF
--- a/libre-2/style.css
+++ b/libre-2/style.css
@@ -621,7 +621,7 @@ textarea {
 		content: "";
 		position: absolute;
 		top: 0;
-		width: 100%;
+		width: auto;
 	}
 
 	.singular .site-header:before {


laurelfulford:
Thanks for reporting this issue, @torres126! 

I'm running into some issues recreating this that are probably just me :) Do you need to be on a specific page/post to see the issue? It looks like it's happening before you scroll down the page -- is that correct?

Also could you please share the browser/OS that you're recreating this in the screenshots above?

Thanks in advance!

torres126:
Sure thing @laurelfulford. The screenshots above are on a Mac OS X 10.6.8 using Chrome 49.0.2623.112, and setting the pixels to 1440 wide using Developer Tools/Inspect Element. I've done that because I see this issue on my Samsung Galaxy Tab S2, running Android 7, on Chrome 70.0.3538.80, where this issue is more likely to be seen unintentionally by readers. 

Any page except the posts page seems to cause this issue, and it happens both before and after you scroll down the page. Also, just to confirm, I tested (and saw) this issue on both the sites linked in #361. :)

laurelfulford:
Ah, thanks for the extra details, @torres126! I'm still stuck on recreating this on my own test site, but I can see see it on the sites shared in #361. I'm wondering if it has something to do with whether ads are present? 

The `:before` and `:after` elements were kind of a hacky way to make Libre 2's header look "right" when there are wide blocks in the content. Originally, the header looked like it spanned the entire width of the browser window when it was "sticky" as you scrolled down, but it actually didn't -- this is most obvious when scrolling over a wide image or gallery block:

image

This is what it looks like when fixed:

image

I snuck these styles in with the Gutenberg support added to the theme. Unfortunately, switching the width to `auto` makes it so the white band doesn't run across the screen, so the header looks like it does in the first screenshot again. 

Messing around in the inspector, it looks like switching the `overflow-x` to the `body` might fix this, but I'm going to test it a bit more before sticking with it :) It's possible my hacky header work-around might need to be reworked all together!

Thanks again for spotting this, @torres126!

PR #411, opened 2018-12-06
Libre 2: Fix Centre Alignment of Images & Captions

torres126:

#408 doesn't seem to work like intended, and it seems I can't modify that PR as captions aren't shown when trying to do so. Therefore, after trying to look for an alternative, I've created a new PR to try and solve the issue of images not aligning when using the Image Block on the Libre 2 theme. 

I'm not too sure if this is the most effective way at solving this problem, but I'm hoping it does work. 

**Before:**

gfsdgfsdfgsfgsd

**After:**

gfsfgddfgsgdf

(cc @laurelfulford)

### DIFF
--- a/libre-2/css/blocks.css
+++ b/libre-2/css/blocks.css
@@ -91,7 +91,7 @@ figure[class^="wp-block-"]:not(.aligncenter):not(.alignleft):not(.alignright):no
 	position: relative;
 }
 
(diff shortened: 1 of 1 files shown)


laurelfulford:
Closing to merge in #467.

torres126:
Hm, that's odd, not sure why that happened. Thanks for fixing it up!

Issue #413, opened 2018-12-07
Sketch: Post titles in new editor show as uppercase


davidakennedy:
## Steps to replicate

See: https://twitter.com/jarango/status/1070873507754438656

1. Activate Sketch on a site.
2. View a post in the editor.

## Result

Post titles show as uppercase, but those do not match the front end of the site.

## Expected

Post titles should be as the user typed them.

* User report: https://twitter.com/jarango/status/1070873507754438656
* User's site: https://jarango.com/ | https://jarango.com/2018/12/06/new-keynote-designing-distinctions/

davidakennedy:
Closing this now, since I see it's the result of customer's custom CSS.

davidakennedy:
Reopened the issue after some discussion in Slack.

We're going to remove text transform in the editor because it breaks the user experience.

PR #417, opened 2018-12-07
Lodestar, Illustratr & Galactic 2: Remove Uppercase in Post Titles

torres126:

As mentioned in #413, Illustratr, Intergalactic 2 and Lodestar all has themes which transform the post title in the Editor with intent to make the editor and front-end match, but this can cause confusion. So, this PR intends to stop that from happening so the user can decide to use capital letters.

### DIFF
--- a/illustratr/editor-blocks.css
+++ b/illustratr/editor-blocks.css
@@ -61,7 +61,6 @@
 	color: #24282d;
 	font-family: "Source Sans Pro", sans-serif;
 	font-weight: 900;
-	text-transform: uppercase;
 	font-size: 1.2em;
 }
 
--- a/intergalactic-2/editor-blocks.css
+++ b/intergalactic-2/editor-blocks.css
@@ -68,8 +68,7 @@
 	font-family: Lato, Helvetica, sans-serif;
 	font-weight: bold;
 	font-size: 30px;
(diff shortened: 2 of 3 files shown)


laurelfulford:
Thanks for the patch, @torres126! This looks good to me.

vinnie333:
Sorry for the noobie question but I edited the CSS file in intergalactic-2/editor-blocks.css inside Wordpress, hoping that Post Titles wouldn't be all caps anymore but it doesn't seem to have done anything?
I removed the two lines & added the one as it showed but nothing seems to have changed...
Any insight on what i'm missing?   I love this theme other than the darn ALL CAPS in Page titles & Post Titles...

torres126:
Hi @vinnie333, `editor-blocks.css` (which was modified in this PR) is only for the Gutenberg editor, so removing that bit there won't affect the actual content. It's generally recommended to use the Customiser instead of directly editing the theme to make CSS changes though, there's some reasons and steps listed here: https://codex.wordpress.org/CSS#Custom_CSS_in_WordPress

Try adding this. :)

```
.entry-title {
text-transform: inherit;
}
```

vinnie333:
You're amazing, thank you so much! 
No more ALL CAPS!

Issue #415, opened 2018-12-07
Multiple Themes: All caps in post title in editor can be confusing


laurelfulford:
Similar to #413, we have a few themes that display post titles in all caps; in the new block editor, this style is replicated so the front-end and editor match.

Unfortunately, this also means that these styles mess with the user input -- you can't tell what case you're actually typing in, which could result in weirdness in what the title actually is. This would be noticeable in something like Google, or when the theme's switched. 

Sketch was fixed in #413 - these themes also transform the post title:

* Illustratr
* Intergalactic 2
* Lodestar

Just of note, and perhaps future discussion, some themes in this repo also transform some/all headers in the content: 

* Apostrophe 2 (H4, H6)
* Dyad 2 (H5, H6)
* Intergalactic 2 (H1, H2, H4, H5, H6)
* Ixion (H4, H5, H6)
* Karuna (H1, H2, H3, H4, H5, H6)
* Libretto (H2, H4, H6)
* Lodestar (H1, H2, H3, H4, H5, H6)
* Pique (H2, H3)
* Radcliffe 2 (H5)
* Sketch (H1, H2, H3, H4, H5, H6)
* Textbook (H5, H6)
* Toujours (H4, H5, H6)

... and it looks like most in this repo transform buttons, and often table headers too.

laurelfulford:
Merged #417 which fixes this issue.

laurelfulford:
Illustratr committed to WP.com in r52732-wpcom-themes.

Intergalactic 2 committed to WP.com in r52733-wpcom-themes.

Lodestar committed to WP.com in r52734-wpcom-themes.

laurelfulford:
Thanks @torres126! I totally missed that, because I was searching for `text-transform` in the stylesheets :) 

I'll leave it for now for the sake of keeping the right font, but will keep an eye out for any reports of confusion due to it!

PR #418, opened 2018-12-07
Several themes: Remove Uppercase from Editor

torres126:

As mentioned in #413, there can be a lot of confusion caused by the capitalising of things in the editor, but it's done to reflect how the theme actually looks front-end. This PR intends to remove most of those, as the confusion could outweigh the benefits from matching editors, and could cause further issues, as discussed in #413, although ultimately, it may just be worth leaving it as it is. 

That mainly means remove capitals from:

- Headings
- Table Headings 
- Citations (Pullquotes and Quotes) 
- Buttons (both the Button and the File Block) 

The individual ones for each theme are mentioned on each commit, and I'm fairly certain I've not missed anything and I've not removed anything unless it is necessary to do so. 

I've done this in a different PR to #417 in case it's decided that it's worth leaving other parts of the Editor as be. 

(cc @laurelfulford)

### DIFF
--- a/affinity/editor-blocks.css
+++ b/affinity/editor-blocks.css
@@ -358,8 +358,7 @@
 	border-radius: 0;
 	display: inline-block;
 	font-size: 13.2px;
-	font-family: Raleway, "Helvetica Neue", sans-serif;
-	text-transform: uppercase;
+	font-family: Raleway, "Helvetica Neue", sans-serif;	
 	letter-spacing: 1px;
 	font-weight: bold;
 	line-height: 1;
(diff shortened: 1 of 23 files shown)


laurelfulford:
Thanks for this PR, @torres126!

I think I may have been misleading in #415 -- I was just making note of the themes that use all caps in different areas of their design, since it's something we've been discussing as potentially confusing when paired with the new editor styles. There isn't a plan at the moment to remove them.

torres126:
Thanks @laurelfulford, no worries, I just did this (separately to #417) in case it's something which comes up in the future. :)

(Spare time on a Friday evening has wonders!)

Issue #459, opened 2018-12-31
All themes: Gutenberg caption color unreadable on dark backgrounds


goblinartificer:
## Steps to replicate
1. Set up a theme on a Free or Personal site which has a dark color scheme available. I tested with *Gateway*'s dark gray or *Escutcheon*'s default. 
2. Set up two posts, one in Gutenberg and one in the old editor, with images with captions.

## Result
The Gutenberg caption is unreadable and dark gray; the other caption is an expected light color in keeping with the rest of the theme text.

## Expected
Both captions should be identical. In the initial user report, the user found the issue persisted even after converting back to the classic editor, and created a new post to avoid these.

## Additional info
One example here: [Gutenberg](https://limitedepsilon.wordpress.com/2018/12/28/caption-test-1/) | [classic](https://limitedepsilon.wordpress.com/2018/12/28/caption-test-2/)

`.wp-block-image figcaption` has its own color set, where the classic editor caption inherits the body color. 

* User report: #9081773-hc
* Follow-up ticket: #1672452-zd

laurelfulford:
Just circled through to check which themes have this issue - they are: 

* Affinity
* Canard
* Gazette
* Independent Publisher 2
* Intergalactic 2
* Photos
* Rebalance

#463 covers a number of these - testing that next!

laurelfulford:
Fixes for these themes have been committed to WP.com in:

* Affinity - 53231-wpcom-themes
* Canard - 53232-wpcom-themes
* Gazette - 53233-wpcom-themes
* IP2 - 53235-wpcom-themes
* Intergalactic 2 - 53236-wpcom-themes
* Photos - 53237-wpcom-themes
* Rebalance - 53238-wpcom-themes

... so that's all the themes that have Gutenberg styles built in fixed. Themes without (like     Escutcheon) will continue to use Gutenberg's default caption colour of grey. For now, this is a `wontfix`. 

@goblinartificer It looks like the ZD ticket is closed - can you please follow up with the user who reported this issue?

PR #463, opened 2018-12-31
Several themes: Inherit Caption Colours with Gutenberg Blocks

torres126:

Inherits the caption colour for five themes: Canard, Gazette, Independent Publisher 2, Intergalactic 2 and Rebalance 

Other themes may have a set caption colour, but even though you can change the colour of the background of the site, you can't change the background colour behind the caption. Examples of these themes include Libretto, Textbook and Toujours. The only theme where there's still an issue which this PR won't help with is Photos, the caption issue is present but setting the colour to inherit only makes it worse for some reason. 

The effect of this change is to make the colour of the caption far more readable on darker coloured backgrounds, since they currently fail the accessibility [guidelines](https://dotcombrand.wordpress.com/colors/) of a colour ratio. 

### Example 

**Before:**

before_dgdgfdgfgdf

**After:**

after_gdgfdfdg

**Not needed - changing the background only changes what's currently black, not the colour behind the caption itself**

not_needed_button_2

#### Related issue(s):
#459

### DIFF
--- a/canard/blocks.css
+++ b/canard/blocks.css
@@ -21,7 +21,7 @@ Description: Used to style Gutenberg Blocks.
 /* Captions */
 
 [class^="wp-block-"] figcaption {
-	color: #777;
+	color: inherit;
 	font-family: Lato, sans-serif;
 	font-size: 16px;
 	line-height: 1.25;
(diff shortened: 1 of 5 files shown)


laurelfulford:
Thanks @torres126!

PR #464, opened 2019-01-01
Penscratch 2: Fix No-Sidebar Layout

### LINKED ISSUE #371: Penscratch 2: No-Sidebar layout is off center at 55em
## Steps to replicate
- Enable theme
- Remove sidebar widgets
- Adjust screen width to about 55em

## Result
screen shot 2018-11-20 at 2 21 15 pm

## Expected
screen shot 2018-11-20 at 2 23 48 pm

* User report: https://en.forums.wordpress.com/topic/how-to-hide-the-side-bar-when-full-screen/
* Follow-up ticket: (same is OK)
* User's site: https://totoko.fr/

Noting the user already has custom CSS to fix it.

torres126:

Adds `margin-right (and left): auto` to the site when there isn't a sidebar. Though `margin: 14px auto` works well when the issue occurs, it results in the content being higher on larger screens because it overrides the margin [here](https://github.com/Automattic/themes/blob/master/penscratch-2/style.css#L1911). This approach doesn't cause that effect. 

**Before:**

ezgif-2-582f41f6c020

**After:**

ezgif-2-fb4e2832fb09

#### Related issue(s):
Fixes #371

### DIFF
--- a/penscratch-2/style.css
+++ b/penscratch-2/style.css
@@ -1918,6 +1918,8 @@ div.sharedaddy div.sd-block {
 }
 @media screen and (min-width: 55em) {
 	.no-sidebar .site {
+		margin-right: auto; 
+		margin-left: auto;								
 		max-width: 872px;
 		padding: 54px 108px;
 	}


laurelfulford:
This look good - thanks @torres126! :)

Issue #434, opened 2018-12-15
All Themes: Button block styles not working in themes with Gutenberg styles & support


ahmadbaig1:
Note: this issue originally specified the Shoreditch theme, but this was overlooked in any theme that we added block styles and supports to. 

## Steps to replicate

- Activate the Shoreditch theme on your site.

- Edit a Post in the Gutenberg editor. 

- Add a button block, try to select the different button styles using the option in the right sidebar- "outlined" and "squared", and they do not reflect either on the button or on the button preview like they do in other themes.

Storeditch theme-
<img width="1200" alt="screen shot 2018-12-15 at 3 28 05 pm" src="https://user-images.githubusercontent.com/40609190/50041963-f785cb80-0082-11e9-8209-5b696cce801b.png">

TwentySeventeen theme-
<img width="1200" alt="screen shot 2018-12-15 at 3 29 13 pm" src="https://user-images.githubusercontent.com/40609190/50041968-066c7e00-0083-11e9-81f4-b4346f46e8c5.png">

## Result

The button style is not changing to either to the "Outlined" or the "Squared" button type, neither on the actual button on the block nor on the preview buttons which are clicked to select the styles, on the right sidebar.

## Expected

I expected the buttons to change styles, but they do not change and stay like the default style.

* User report: [8777635-hc]
* Follow-up ticket: [1642719-zd]
* User's site: [[https://sebastianschick.com](https://sebastianschick.com)]
* User Post: [ [post/page]( https://sebastianschick.com/wp-admin/post.php?lang=en&action=edit&post_type=page&post=2342)]

laurelfulford:
Unfortunately, I know that we didn't consider the button styles in the other themes we added block editor styles to, either. 

I've updated the ticket to be more general.

jamiepalatnik:
Reported for Gazette theme in 8117471-hc.

gwwar:
@laurelfulford is any integration work here needed or do we need theme updates? Asking since, the screenshots appear to be Atomic sites.

laurelfulford:
@gwwar It's all theme updates :)

In all the themes, we followed the theme's original design for buttons... but should have deferred to Gutenberg's default rounded styles, since most of them aren't. On top of that, in a lot of cases, the theme's CSS is too specific and it's overriding the styles Gutenberg's trying to add on top, so all the options look the same.

gwwar:
Gotcha, I'll assume you have this covered. Let us know if you need a hand otherwise 👍

[...]

PR #481, opened 2019-01-09
Shoreditch: Update button block styles

laurelfulford:
Update button block styles to respect rounded, squared and outline settings. 

#### Related issue(s):
See #434.


laurelfulford:
Oh! I also cleaned out some hover styles from the editor from both the button and file block -- they are misleading there.

PR #497, opened 2019-01-16
Apostrophe 2: Ensure "Follow" Button is Readable

### LINKED ISSUE #496: Apostrophe 2: Button text on Follow Blog widget not readable
## Steps to replicate

- Add Jetpack Subscription widget to a Jetpack connected site using the WordPress.com Apostrophe 2 theme or add the Follow Blog widget to a WordPress.com site using the same theme.
- The 'Subscribe Now' Text in the button is not readable unless hovering the mouse over the button, using the default styling.

screenshot 2019-01-16 at 08 18 08

I was able to replicate this both on a WordPress.com site and a Jetpack connected site as well.

## Expected

Text on the button should be readable.

* User report: #1717150-zd
* User's Jetpack site: http://www.awakenplus.com/
* User's WordPress.com site: https://tantraplusyoga.wordpress.com/

torres126:

Inherit the colour for `input` of widgets, so that the Follow button is readable when using the default colour palletes, but isn't changed when using any other pallete. 

**Current:**

ezgif-5-8aef022aa21a

**Proposed:**

ezgif-5-8d73e0e470a4

#### Related issue(s):

Fixes #496

### DIFF
--- a/apostrophe-2/style.css
+++ b/apostrophe-2/style.css
@@ -1255,8 +1255,9 @@ a:active {
 	max-width: 100%;
 }
 
-/* Form elements should span the full width. */
+/* Form elements should span the full width and be readable. */
 .widget input {
+	color: inherit;	
 	width: 100%;
 }
 


laurelfulford:
Thanks for putting this together, @torres126!

It looks like it causes some issues -- the input in the Follow widget that shows up when you're not logged in also inherits the styles, so the font is white on white when it's not focused. Other inputs in widgets (like the search) also inherit the style, making them light grey.

I'm going to spin something together for this with the colour annotations code that's also needed for the custom palettes, since that's hard to test on a self-hosted site. 

Thanks again for your help here!

PR #508, opened 2019-01-24
Libretto: Fix Contact Form Displays

### LINKED ISSUE #504: Libretto: Dropdown form fields do not display correctly on desktop
Issue occurs both in forms created using the Calypso classic editor, and Gutenberg form block.

Reported by a user and duplicated on my own test sites.

## Steps to replicate

Classic Editor
1. Add a form to a post or page via Add > Contact Form
2. Add a "dropdown" field to the form

Gutenberg
1. Add a form block
2. Add a "select" field

## Result

The dropdown box does not display in the right location: http://cld.wthms.co/O3eEm8

## Expected

I expected the dropdown box to display in the correct location, like it does on mobile: http://cld.wthms.co/gWnSQu

* User report: #1710300-zen
* User's site: https://concorsocittamurata.com/2018/10/29/festival-citta-murate-2019-prenotazione-biglietti-eventi/
(We will be providing them with CSS as a temporary workaround).

torres126:
Fixes #504

Display the currently broken elements as a block on wider viewing, but an inline block on mobile, so that there's no visual differences except for the fix. 

**Gutenberg:**

hdgghfdghfhg

**Classic:**

Sorry, not the most clear screenshots. Hopefully they show the layout okay though. 

screenshot_20190124-190627

And on smaller screens. 

screenshot_20190124-190734

### DIFF
--- a/libretto/style.css
+++ b/libretto/style.css
@@ -2032,6 +2032,7 @@ label[for="comment"] {
 }
 .site-content .contact-form label.checkbox,
 .site-content .contact-form > div {
+  display: block;
   margin-bottom: .875em;
 }
 .site-content .contact-form textarea,
@@ -2039,6 +2040,7 @@ label[for="comment"] {
 .site-content .contact-form input[type='email'],
 .site-content .contact-form input[type='url'],
 .site-content .contact-form select {
+  display: block;
   margin-bottom: .4375em;
   max-width: 100%;
 }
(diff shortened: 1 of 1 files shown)


laurelfulford:
Thanks for taking a swing at this, @torres126!

I think it might be easier/more consistent to follow the styles used on the other inputs and piggy back off of the label and field alignments already in the theme, rather than creating something new for the select dropdown. I'm sorry it took me a while to get back to your question in the issue itself, but I think all the labels/fields should be aligned consistently, rather than having them stacked for some form fields, and side by side for others. 

I messed around with this a bit to see if it was actually possible, and think I have a workable fix -- I'm going to push that up now. 

Thanks again for working on this! Theses kinds of issues are tricky, where, not only does an issue need to be fixed, but a visual direction for the fix needs to be sorted out.

PR #517, opened 2019-01-25
Karuna: Add Gutenberg Button Block Styles

torres126:

- Let only the rounded and outline buttons have a border radius 
- In the editor, override the colours because that's not accurate or consistent of the frontend 

I'm *really* hesitant about using an override, but I feel like it's the only solution; the option can't be removed completely because they're needed for hover styles, but right now this offers a temporary solution to the most noticeable problem with the Outline Button colours. 

**Current:**

Editor:

fsdadafsasdfadfs

Frontend:

gfdsfdsgsfgdsfgd

**Proposed:**

Editor:

afdsfdasafdsafds

**Frontend:**

hdgfgfhdgfhhfgd

#### Related issue(s):

#434 

(Btw, I checked the branches and couldn't find one which addresses this, if this is being worked on and this PR isn't necessary, please feel free to close!)

### DIFF
--- a/karuna/blocks.css
+++ b/karuna/blocks.css
@@ -318,8 +318,7 @@ p.has-drop-cap:not(:focus)::first-letter {
 /* Buttons */
 
 .wp-block-button .wp-block-button__link {
-	border: 2px solid currentColor;
-	border-radius: 10px;
+	border: 2px solid currentColor;	
 	box-shadow: none;
 	font-weight: bold;
 	font-size: 16px;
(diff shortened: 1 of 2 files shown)


laurelfulford:
Thanks for putting this together, @torres126! 

I have been working on a fix for this issue, just have been moving slowly since I self assigned it :) I'll circle back to this once I get to Karuna, but it looks like it needs to be reworked to remove the `!important` from the styles.

Issue #503, opened 2019-01-21
Illustratr: Overlapping Videos on Portfolio Pages with columns block


thebud15:
## Steps to replicate

1. Add a new Portfolio project
2. Add a column block
3. Add a YouTube video link to each one (in this specific case, these links were:
https://youtu.be/BNoGUhZcmaY
https://youtu.be/ItMVtMpTScA
4. Publish the page and see the videos overlapping with the footer area of the site, as the below screenshot shows: 

![screen shot on 2019-01-16 at 20_24_21](https://user-images.githubusercontent.com/29046778/51449992-ab622280-1ce3-11e9-8393-f8eb45312dd1.png)

## Result

The videos displayed in a block side by side, but overlapped with the lines in the footer of the site. 

## Expected

I expected the two videos to sit side by side with no overlap. 

* User report: #7617547-hc
* Follow-up ticket: #1724761-zd
* User's site: https://naiiboa.wordpress.com/portfolio/sweet-creations/

laurelfulford:
I think this might have to do with Illustratr's built-in media resizing on the portfolio pages: 

https://github.com/Automattic/themes/blob/master/illustratr/js/portfolio-single.js#L8

alaczek:
Scratch that, when I created portfolio project on my test site and added two videos in columns, I could see the issue you're describing, but not the one I was seeing on customer site.

alaczek:
Allrighty, I have a patch ready for testing. It did fix the issue on my site, but didn't on the user site, when I sandboxed it. Maybe I'm sandboxing it wrong? Would appreciate second look cc @laurelfulford

laurelfulford:
This has been fixed on WP.com with r53385-wpcom-themes.

It looks like the user has removed the columns as a temporary fix. 

It looks like the original ZD ticket is now closed -- @thebud15 can you please follow up with the user who originally reported this? Thanks!

thebud15:
Done in #1805449-zen. Thanks @laurelfulford @alaczek

PR #538, opened 2019-02-06
Update responsive-videos.js

alaczek:
Prevent responsive-videos script from targeting images in columns block.

Before proceeding with the resizing, I added a check for a video inside a columns block.

Related issue(s): #503

### DIFF
--- a/illustratr/js/responsive-videos.js
+++ b/illustratr/js/responsive-videos.js
@@ -13,11 +13,11 @@
 	function responsive_videos() {
 
 		$( '.video-wrapper embed, .video-wrapper iframe, .video-wrapper object' ).each( function() {
-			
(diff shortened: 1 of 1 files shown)


jeffikus:
@alaczek looks good - just wondering if there ever would be a situation where the returned jquery element would be empty and if that's a problem? Otherwise it looks good to merge!

laurelfulford:
This isn't actually working for me :| When I add two YouTube videos to the columns, they're still too large -- I double-checked that things weren't caching by adding `console.log()`.

jeffikus:
@laurelfulford can you post a screencast of your experience so I can take a look later?

laurelfulford:
I've updated the PR with a fix that @alaczek found, since it looks like it does the trick!

PR #548, opened 2019-02-07
Canard: Fix button styles

laurelfulford:
This update corrects Canard's button block styles, so you can actually use the default rounded, and assign the outline and square options.

It also includes a re-working of Affinity's button styles, that should work better with custom colours - I was too clever the first time out.

See #434.

### DIFF
--- a/affinity/blocks.css
+++ b/affinity/blocks.css
@@ -187,15 +187,31 @@ p.has-drop-cap:not(:focus)::first-letter {
 	text-transform: uppercase;
 	letter-spacing: 1px;
 	font-weight: bold;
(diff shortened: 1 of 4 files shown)


kspilarski:
I also have a user with the same issue and can recreate it on my test site. The Poll block's submit button automatically got a red border color as soon as Canard theme is activated on the site. The workaround is to use CSS code to change the border color. 

In the editor, the poll block submit button border color is red and shows `"borderColor":"#dddddd"`, but in the block inspector, it shows as `border-color" #d11415;`: 

Screen Shot 2021-11-06 at 4 28 36 PM

Issue #225, opened 2018-08-12
Intergalactic 2: Background colors make site title disappear


GeoJunkie:
## Steps to replicate

- Create a site with the Intergalactic 2 theme
- Set a static homepage without a featured image
- Set a slideout menu
- Change the background color under Colors & Backgrounds 

## Result

For some of the colors, the text disappears:

![screen capture on 2018-08-12 at 15-17-30](https://user-images.githubusercontent.com/7596032/44005537-eae5f438-9e42-11e8-9ebf-2a9b64842962.gif)

## Expected

The font color should be one that is visible on the background that's chosen.

I've given the user a workaround to use Custom CSS to set the font color manually, but the theme should choose a reasonable color (as it does for the tagline)

* User report: #1342863-zen
* User's site: https://islandmovement.life/

laurelfulford:
I can recreate this on my test site.

nagpai:
User reported their header disappearing in the default palette here 
6874567-hc 

I asked them to change the palette from customizer, however the default palette needs a fix.

SeeRjED:
Having the same issue with background colour set to white. It hides the menu button for me and I'm pretty sure this wasn't the case a few weeks ago.

laurelfulford:
@nagpai that looks like a different issue that was introduced yesterday, reported here: #318. The issue has been fixed; I've checked their site and it looks like the header is now working as expected.

laurelfulford:
Committed to WP.com in r53280-wpcom-themes.

@GeoJunkie It looks like the original ZD ticket is closed - can you please follow up with the user if needed? If they opt to get rid of the Custom CSS fix, they will need to resave their Colour Settings to make sure the original contrast isn't cached on their site. Just let me know if you have any questions about this!

PR #572, opened 2019-02-07
Intergalactic 2: Fixes color annotations so that site title doesn’t disappear

allancole:

• Adds a `:not(.singular)` selector to page views so that the site-title can respect the colors set in the customizer

#### Related issue(s):

Fixes: #225 

## Before

intergalactic-2-colors-before

## After

intergalactic-2-colors-after


laurelfulford:
Looking good, Allan!

Issue #282, opened 2018-10-12
AltoFocus: menu overlap on IE 11


eduardozulian:
## Steps to replicate

1. Open Internet Explorer 11
2. Visit https://altofocusdemo.wordpress.com/

## Result

Using the AltoFocus theme on IE11, the menu stands over the logo:

<img width="1144" alt="captura de tela 2018-10-12 as 16 56 18" src="https://user-images.githubusercontent.com/1136160/46891305-ccdb8480-ce3f-11e8-9798-4eda83e5c4fb.png">

The same happens in the theme demo page:

<img width="1378" alt="captura de tela 2018-10-12 as 16 57 21" src="https://user-images.githubusercontent.com/1136160/46891363-fa283280-ce3f-11e8-889d-7b69a372af69.png">

## Expected

See the menu on the right, just like we do in Chrome, Safari, and Firefox.

* User report: 2167914-hc
* Follow-up ticket: 1488537-zen
* User's site: https://faithonfire.blog/

Robertght:
Howdy! I just got this in queue. Can I get someone to look at this? :) Thanks!

laurelfulford:
Confirm that I can recreate this issue on the theme demo. 

Marking as `needs-patch`, but I'm not 100% sure this is fixable -- it needs a bit of investigation. Altofocus uses some newer CSS methods that may not be possible to get working perfectly in IE11, since it was released 5+ years ago. If this comes up again in the queue, it might be good to set some expectations around that.

tanjoymor:
Another reporting of this issue on this site: https://alixkelinda.com/

7607693-hc

I have warned the user that it may not be fixable.

laurelfulford:
This update has been committed to WP.com in r53326-wpcom-themes.

It looks like the original ZD ticket has been closed -- @eduardozulian can you please follow up with the original reporter? Thanks!

eduardozulian:
@laurelfulford, sure! The follow-up was sent on 1802712-zen.

@tanjoymor, I thought about following-up with your user but it seems like they are using the site with page-based content.

PR #581, opened 2019-02-08
Altofocus: Fix menu overlap issue

allancole:

- Changing flex-direction from `row-reverse` to `column` fixes this in IE11 and doesn’t break things for other browsers. 
- May need to do more testing but this should do the trick.

#### Related issue(s):

Fixes: #282 

## Before

image

## After

image


laurelfulford:
Looks good to me, @allancole!

Issue #133, opened 2018-05-15
Dyad 2: Fade over short posts on blog page


thesacredpath:
The user in the following WP.org forum thread reports that the fade on the posts on the blog page text area is being applied to posts that are shorter than the excerpt length. The user has detailed it very well in the forum thread, so I won't repeat it here. 

Demo site screenshot: https://cldup.com/iT07Hgite8.png

It appears like an easy fix, but then again, this is "code" so who knows. :)

* User report: [https://wordpress.org/support/topic/class-too-short-always-applied/](url)
* Demo site: [https://dyad2demo.wordpress.com/](url)

PR #584, opened 2019-02-08
Dyad 2: Fix fade over short posts on blog page

allancole:

> The script function adjustPosts() in global.js, meant to apply the class “too-short” to the wrapper div, will ALWAYS apply this class, no matter the length of the content because the wrapper auto-adjusts itself to the exact height of that content. Which in turn will cause the “if smaller than or equal to” be true on all occasions.

• The above fix actually came from a user, but it seems to do the trick.

#### Related issue(s):

Fixes: #133 

## Before

Notice the `.too-short` class at the end of the class list at the bottom of this screenshot.

image

## After

That class properly goes away after the fix is applied since the fade is no longer needed on short paragraphs. The fading is also gone here but its tough to see on the screenshot.

image

### DIFF
--- a/dyad-2/js/global.js
+++ b/dyad-2/js/global.js
@@ -95,7 +95,7 @@
 				$contentHeight = $content.innerHeight(),
 				$wholeContentHeight = $headerHeight + $contentHeight;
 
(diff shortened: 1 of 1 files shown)


danieldudzic:
Tested, works great!

PR #588, opened 2019-02-09
Updated JS to changed the featured image height

### LINKED ISSUE #284: Radcliffe 2: Homepage post titles overlap on mobile when a "large" heading font is selected
## Steps to replicate

1. Set a simple site to use Radcliffe 2
2. Navigate to Customizer > Homepage Settings > set Homepage to display latest posts
3. Ensure multiple posts with featured images and relatively long post titles are published
4. Navigate to Customizer > Fonts > Headings and choose a larger font, like Playfair Display, and the "Large" font size (or larger)
5. View site on mobile (in my case, I viewed in Safari on iOS, but I experience the same when making the window very narrow on desktop as well)

## Result

Post titles overlap like this: http://cld.wthms.co/aJmKfz

## Expected

I would expect the post titles to display something more like this: https://cld.wthms.co/PCQwNK

* User report: #1492816-zen
* User's site: mumsgoneclimbingcom.wordpress.com
(They may change their Heading size after receiving my response to the ticket, but I was able to replicate this issue on my own test site as well).

crunnells:
The JS looks at the size of each `.entry-header` and compares it to the default `.entry-thumbnail` height ( Window.width * 1.33 ), and sets the image to the larger of the two. Resolves an issue when the Customizer header fonts are set to Large, AND when there are really long post titles on mobile views.

#### Related issue(s):
#284

### DIFF
--- a/radcliffe-2/assets/js/radcliffe-2.js
+++ b/radcliffe-2/assets/js/radcliffe-2.js
@@ -70,6 +70,11 @@
 		fullScreenImages();
 	});
 
+	// One more time after everything (eg: Custom Fonts) has loaded for better height accuracy
+	$( window ).load( function() {
+		fullScreenImages();
+	});
+
 	// On window resize.
 	$( window ).on( 'resize', function() {
 		fullScreenImages();
--- a/radcliffe-2/assets/js/radcliffe-2.js
+++ b/radcliffe-2/assets/js/radcliffe-2.js
@@ -21,10 +21,26 @@
 	function fullScreenImages() {
 		var $entryBackground = $( '.entry-thumbnail' ),
 			$heroArea        = $( '.hero-area' ),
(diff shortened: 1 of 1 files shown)


laurelfulford:
This is looking good overall, @crunnells!

The only issue I'm seeing is that sometimes, on the first load of the page or a hard refresh, the post titles still overlap. I think this is because the custom fonts are a little slower to load, so the theme sets the heights before the fonts increase in size. 

It's a bit hacky, but what do you think about re-firing the function after a timeout? Unless there's something else to watch for to make sure the custom font is loaded?

crunnells:
@laurelfulford I _had_ added the function call attached to `$( window ).load()` so that it runs again when _all_ assets have loaded, but forgot to put it into this PR. Updating now, and it should be good to go.

Issue #644, opened 2019-02-28
Multiple themes: Hard coded column width for Media & Text Block


iamtakashi:
In Professional Business, Sophisticated Business, and Friendly Business, the column that has an image is set to 60% in CSS for > 768px so the resizing column option doesn't work as expected.

The image in the vertical content takes indeed 60%, but that's specified in the block setting, and there is no need to specify that in themes.

The following styles are the cause.

```
@media only screen and (min-width: 768px) {
	.entry .entry-content .wp-block-media-text {
		grid-template-columns: 60% auto;
	}
	
	.entry .entry-content .wp-block-media-text.has-media-on-the-right {
		grid-template-columns: auto 60%;
	}
}
```

```
.editor-styles-wrapper .wp-block-media-text {
	grid-template-columns: 60% auto;
}

.editor-styles-wrapper .wp-block-media-text.has-media-on-the-right {
	grid-template-columns: auto 60%;
}
```

cathymcbride:
retired theme. Closing... (looks like there was a PR to fix it tho)

PR #655, opened 2019-03-04
Multiple themes: Fixes hard coded column width for Media & Text Block

allancole:

- Removes the hard coded 60% widths in the Media & Text block so that users can override this setting with the Block editor.
- Also removes some unnecessary menu styles for `[focus-within]` in Friendly Business. This appears to be a separate issue with the Sass compiler. 

#### Related issue(s):

See #644 

#### Before

sophistcated-m t

#### After

sophistcated-m t-2

#### Test

To test pull in this PR and make sure that the Media & Text block displays with 50% / 50% proportions and make sure that the proportions are flexible (like the _after_ screenshot above).

### DIFF
--- a/friendly-business/sass/blocks/_blocks.scss
+++ b/friendly-business/sass/blocks/_blocks.scss
@@ -794,8 +794,6 @@
 		}
 
 		@include media(tablet) {
-			// Modify the default 50-50 proportion between image and text.
-			grid-template-columns: 60% auto;
 
 			.wp-block-media-text__content {
 				:last-child {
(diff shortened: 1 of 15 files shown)


crunnells:
Looks like there's an issue with the drag icon in Professional, and I just noticed it on Modern as well. It seems to work fine if you add `z-index: 1` to `.wp-block-media-text.is-selected.is-stacked-on-mobile .editor-media-container__resizer .components-resizable-box__handle` but that seems like overkill. I haven't found the reason why this problem exists on some style variants, but not all.

allancole:
@crunnells I went ahead and added a fix for this to the theme. Typically, we would want a fix like this to come from Gutenberg but only if the problem persisted across all themes. Since this issue is actually caused by the subtle design tweaks in the themes themselves, fixing it via the theme is the best way to do it.

PR #659, opened 2019-03-09
Business Theme Variations: Remove horizontal padding from full-width column blocks

### LINKED ISSUE #654: All Theme Variations: Horizontal paddings add too much space when inside full-width column blocks
## Steps to replicate

- Create a 3 Column Block, set it to `wide` aligned, publish and view it on the front end.

## Result

image

## Expected

The individual columns should be wider and stretch the full width of the site minus any horizontal margins. Instead, the horizontal padding is too generous and makes the columns appear too small.

## *EDIT

Turns out this issue effects all 6 theme variations, so I updated the title to reflect that. This issue also needs to be fixed in Twenty Nineteen—I’ll file a ticket on Trac if one doesn’t already exist

crunnells:
@allancole reported in issue #654 that full-width columns had horizontal padding on the block container as well as the individual columns, which resulted in the columns being much smaller than intended (especially on smaller screens). This PR aims to fix that by removing the padding.

#### Related issue(s):
Fixes #654

### DIFF
--- a/calm-business/sass/blocks/_blocks.scss
+++ b/calm-business/sass/blocks/_blocks.scss
@@ -757,6 +757,7 @@
 		}
 
 		@include media(tablet) {
+
 			.wp-block-column > * {
 
 				&:first-child {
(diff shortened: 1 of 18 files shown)


allancole:
After #675 gets merged, we’ll need to re-sync with `master` and re-run the compilers which should remove those unnecessary `[focus-within]` selectors.

allancole:
Just completed some further testing on this issue. Everything seems to check out now, so I’ll go ahead and merge this fix.

PR #660, opened 2019-03-11
Professional Business: Fix Centre Alignment

torres126:

Remove the media query which was overriding this margin from being set, which shouldn't be overriden because an aligncenter class is present:

https://github.com/Automattic/themes/blob/040fda40e8d55dd8dcbd6219d66eae3c034100f8/professional-business/style.css#L3502-L3506

This is currently resulting in an issue with alignments with galleries. 

**Before:**

hgfghfghffgh

**After:**

gfdgfdgfdg

#### Related issue(s):

@automattic-ian mentioned this bug in Automattic/wp-calypso#28756

### DIFF
--- a/professional-business/style.css
+++ b/professional-business/style.css
@@ -3519,14 +3519,6 @@ body.page .main-navigation {
   }
 }
 
-@media only screen and (min-width: 768px) {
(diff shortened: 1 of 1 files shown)


torres126:
Closing

PR #738, opened 2019-04-21
Apostrophe 2: Offer an Option to Display a Mobile Social Menu

### LINKED ISSUE #628: Apostrophe 2: display the social networks on small devices
## Steps to replicate

1. Activate the theme
2. Create a social menu
3. Visit your site on mobile

## Result

No sign of the social icons on devices smaller than 767px.

## Expected

See the actual icons. Current we are offering workarounds like this following:

```
@media (max-width: 767px) {
  .main-navigation {
    position: relative;
  }
  .jetpack-social-navigation {
    display: block;
    position: absolute;
    top: 0;
    right: 0;
    width: 80%;
    height: 50px;
  }
  
  .jetpack-social-navigation ul li a {
      margin: 0 2rem 0 0;
    padding: .75rem 0 1.5rem;
    line-height: 45px;
  }
}
```

Result:
see screen shot
_Link:_ https://cld.wthms.co/7LcUTm

I believe we are not prioritizing changes like this at the moment, but I wanted to report it anyway. Thank you!

* User report: 10560383-hc
* Follow-up ticket: no follow-up so far.

Aurorum:

It's my understanding that changes cannot be made which will affect all sites in risk of it being unwelcome by some users, however the menu on mobile devices with Apostrophe 2 uses a fraction of the space, and it seems very reasonable for users to want to display their social menu in the remaining space.

As such, this offers an option in the Customizer to do so. This will only appear after a social menu has been created.

Screenshot 2019-04-21 at 21 12 36

If checked, then the social menu will appear on mobile devices. If unchecked, it won't (which is what currently happens now in all cases).

Screenshot 2019-04-21 at 21 17 21

#### Related issue(s):

Resolves #628

### DIFF
--- a/apostrophe-2/inc/customizer.php
+++ b/apostrophe-2/inc/customizer.php
@@ -28,8 +28,8 @@ function apostrophe_2_customize_register( $wp_customize ) {
 
 		$wp_customize->add_control( 'apostrophe_2_mobile_social', array(
   			'type' => 'checkbox',
-  			'section' => 'nav_menu[2]',
-  			'label' => __( 'Display social menu on mobile devices' ),
+  			'section' => 'jetpack_content_options',
+  			'label' => __( 'Display social menu on mobile devices', 'apostrophe-2' ),
 		) );
 	}	
 }
--- a/apostrophe-2/inc/customizer.php
+++ b/apostrophe-2/inc/customizer.php
@@ -14,6 +14,18 @@ function apostrophe_2_customize_register( $wp_customize ) {
 	$wp_customize->get_setting( 'blogname' )->transport         = 'postMessage';
 	$wp_customize->get_setting( 'blogdescription' )->transport  = 'postMessage';
 	$wp_customize->get_setting( 'header_textcolor' )->transport = 'postMessage';
+	
+	if ( function_exists( 'jetpack_social_menu' ) ) {
+		$wp_customize->add_setting( 'apostrophe_2_mobile_social', array(
+  			'capability' => 'edit_theme_options',
+		) );
+
+		$wp_customize->add_control( 'apostrophe_2_mobile_social', array(
+  			'type' => 'checkbox',
+  			'section' => 'nav_menu[2]',
+  			'label' => __( 'Display social menu on mobile devices' ),
+		) );
+	}	
 }
 add_action( 'customize_register', 'apostrophe_2_customize_register' );
 
--- a/apostrophe-2/inc/customizer.php
+++ b/apostrophe-2/inc/customizer.php
@@ -15,9 +15,15 @@ function apostrophe_2_customize_register( $wp_customize ) {
 	$wp_customize->get_setting( 'blogdescription' )->transport  = 'postMessage';
 	$wp_customize->get_setting( 'header_textcolor' )->transport = 'postMessage';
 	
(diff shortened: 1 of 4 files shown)


apostrophe-2/inc/customizer.php
```diff
@@ -14,6 +14,24 @@ function apostrophe_2_customize_register( $wp_customize ) {
...
+		$wp_customize->add_setting( 'apostrophe_2_mobile_social', array(
+  			'capability' => 'edit_theme_options',
+			'transport' => 'refresh',
+			'sanitize_callback'  => 'apostrophe_2_sanitize_checkbox',
+		) );
+
+		$wp_customize->add_control( 'apostrophe_2_mobile_social', array(
+  			'type' => 'checkbox',
+  			'section' => 'nav_menu[2]',
+  			'label' => __( 'Display social menu on mobile devices' ),
```
sixhours: This translation function needs a text domain:
```__( 'Display social menu on mobile devices', 'apostrophe-2' ),```

**Review by sixhours — CHANGES_REQUESTED** [review]
One minor change before this ships. Again, I know it's an oldie, so let me know if you'd prefer I spin up a new PR. Thanks @Aurorum

File: apostrophe-2/inc/jetpack.php [hunk 2]
```diff
@@ -61,13 +61,23 @@ function apostrophe_2_has_post_thumbnail( $post = null ) {
...
- * Return early if Social Menu is not available.
+ * Return early if Social Menu is not available and check if it should be displayed on mobile.
  */
 function apostrophe_2_social_menu() {
 	if ( ! function_exists( 'jetpack_social_menu' ) ) {
 		return;
-	} else {
+	} 
+	
+	if ( function_exists( 'jetpack_social_menu' ) && ( get_theme_mod( 'apostrophe_2_mobile_social' ) == 0 )) {
```
sixhours: I think this logic can be simplified; we already know the `jetpack_social_menu` function exists since we checked for it early on, so those can be removed:

```
        if ( get_theme_mod( 'apostrophe_2_mobile_social' ) == 0 ) {
		jetpack_social_menu();
	} else { ?>
		<div class="social-mobile-menu">
			<?php jetpack_social_menu(); ?>
		</div>
	<?php }
```

That said, I wonder if adding a special class to the body tag like `show-mobile-social-menu` when the option is active would prevent the need for extra markup. What do you think?

sixhours:
Testing this locally it works fine, but testing on WP.com isn't showing the new option for some reason. 🤔 I need to do more testing, but I wonder if moving this option somewhere more stable makes sense, since the menu assignments can change. Maybe Content Options, which are also included with Jetpack. /cc @Aurorum

apostrophe-2/inc/jetpack.php
Aurorum: Hi @sixhours, I updated this PR to remove that markup and favour the body tag approach like you suggested, along with moving the menu to Content Options. Let me know what you think. :)

sixhours (approved):
Works great! Tested on my local dev environment and WP.com and had no issues. Thanks again @Aurorum! I'll work on getting this merged today. :)

sixhours:
Deployed in r58825-wpcom-themes

Issue #745, opened 2019-04-29
All Theme Variations: Set top/bottom margins to 32px


allancole:
For some theme variations, the vertical spacing is inconsistent which makes the spacing between themes variations irregular. Some themes have a 44px margin while others have 22px, but they should all be 32px. This isn’t a breaking issue, but affects the templates used in the on-boarding process particularly noticeable for spacers, separators, headings and blocks with `alignfull`.

PR #770, opened 2019-05-02
Add 32px vertical margin between blocks in all 6 theme variations. 

allancole:

This PR adds a consistent 32px margin unit and applies it to vertical margins between sibling Blocks. This should make the spacing between all theme variations consistent for the on-boarding flow

#### Related issue(s):
See: #745


iamtakashi:
I only had a look a few themes, but stuff like blocks with `alignfull` or headings now has 64px margins (possibly more, I didn't go through all the changes.) It looks like some of calc and multiplies left untouched. Could you go through the changes and make sure the changes?

modern-business/sass/blocks/_blocks.scss
```diff
@@ -200,7 +200,7 @@
 		}
 
 		@include media(tablet) {
-			padding: 60px 0;
+			padding: $size__vertical-spacing-unit 0;
```
iamtakashi: This vertical padding inside a media and text block in this theme was 60px before, and it now makes sense to change a multiple of 32px with this change, but 32px feels a bit too small. How about 64px? You can find many 60px used around this block, and it's probably better to change all of them.

allancole: Got it. I’ll sort them all out so they’re all using the variable with the multiplier.

Issue #757, opened 2019-04-30
All Theme Variations: Style Business Hours block better and reduce the overall height.


iamtakashi:
Business Hours block currently doesn't look great with what theme defines for description details/term element. It deserves a better style :)

**Currently:**
<img width="286" alt="Screen Shot 2019-04-30 at 22 49 31" src="https://user-images.githubusercontent.com/908665/56995654-4055a100-6b9a-11e9-918b-34b02e9848b8.png">

<h2>Suggested Fix</h2>
Display the day and the time inline to reduce the overall height of the block.

iamtakashi:
This feels like something that needs to be addressed in the block. There doesn't seem any consideration about the front-end and the editor style whatsoever.

iamtakashi:
We'd like to fix this issue in the block, but whatever the fix we push to Jetpack won't land WP.com by June 1st. We would need to add a temporary fix for this in each child themes.

PR #790, opened 2019-05-09
Temp. fix for Jetpack Business Hours Block style for all 6 style variations

iamtakashi:
Ideally, this simple style fix should be added to the block itself, but we can't wait for the fix to land in WP.com for Onboarding.

We should remove this when the style is added in the block and becomes available in WP.com.

#### With this PR the block should look like this:
Screen Shot 2019-05-09 at 22 21 52

#### Related issue(s):
Fixes: #757

### DIFF
--- a/calm-business/style-editor.css
+++ b/calm-business/style-editor.css
@@ -903,3 +903,19 @@ ul.wp-block-archives li ul,
 [data-type="core/media-text"] a {
   color: inherit;
 }
+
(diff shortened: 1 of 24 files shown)


allancole:
Just adding a small clarification that these styles will eventually need to live in Jetpack. After we merge this temporary fix we may want to submit this issue and its solution to Jetpack itself. Then as Jetpack is updated on wpcom we can remove this temporary fix. 

But for now let’s merge!

Issue #760, opened 2019-04-30
All Theme Variations: Style enhancement for Mailchimp block


iamtakashi:
Currently, the input field is very short and looks strange.

Also, the concent text shouldn't have the same visual weight as a paragraph. Should use one or two smaller size in the font size variables.

<img width="770" alt="Screen Shot 2019-04-30 at 23 17 11" src="https://user-images.githubusercontent.com/908665/56996968-21590e00-6b9e-11e9-835c-60bd1a20f736.png">

<h2>Suggeste Fix</h2>
 Make the input width 100%. Use one or two smaller size in the font size variables.

scruffian:
These fixes should probably be in the block

Robertght:
Based on this https://github.com/Automattic/themes/pull/791, it appears it's related to the retired themes. 

Closing

PR #791, opened 2019-05-09
Temp style enhancements for Mailchimp block for all 6 style variations

iamtakashi:
Ideally, this should be fixed in the block itself and the Twenty Nineteen compat file in Jetpack, but we can't wait for those for Onboarding.

* 100% width input field
* Smaller consent text

#### WIth this PR the block should look
Screen Shot 2019-05-10 at 00 56 19

#### Related issue:
See #760


allancole:
Yup, this looks good. Merging :-)

PR #808, opened 2019-05-16
All Theme Variations: fixes Cover Block text align and font-size settings

allancole:
This change revises the font scale for the Cover Block to work more intuitively with the new paragraph and heading options in the latest version of the Cover Block. 

To test, for each theme variation:

1. On a post or page, create a new block and make sure that the small, normal, large, and huge font-sizes for the cover text are the same in the editor and the frontend when using `<p>` for the text. 
2. Test to make sure  `<h1>-<h6>` font sizes look correct on the front end when using headings for the text.  
3. Lastly, make sure that the Cover Block text is centered by default for both paragraphs and headings. (Caveat: It appears that the block editor sets the `wp-cover-block h2` to alignleft, and that should be likely be fixed in core).

So far only Elegant Business has this fix applied. Once we test that this fix works for Elegant, we’ll move on to the others with the same fix.

### DIFF
--- a/calm-business/package-lock.json
+++ b/calm-business/package-lock.json
@@ -1161,8 +1161,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -1183,14 +1182,12 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
-          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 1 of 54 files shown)


iamtakashi:
Changes are mostly great. Thanks for working on this big one. The only thing makes me stop this is that, with this change, Modern looks quite different.

The left is master, the right is this branch.

This is h1 in the content. In this fix the weight is 300.
Screen Shot 2019-05-18 at 03 06 18

The small font size is even smaller in this fix, and the weight of h3 is now 300.
Screen Shot 2019-05-18 at 03 06 37

Screen Shot 2019-05-18 at 03 06 51

Screen Shot 2019-05-18 at 03 07 04

And before merge this, can you make sure the font sizes match between the editor and the front end?

It's almost there!

iamtakashi:
For Modern, can we use 300 for headings for the followings so that avoid radical change to activated sites with this fix?

* Entry titles
* Page titles
* Cover block

And the rest of headings remain 700 as they were.

allancole:
@iamtakashi This should all be fixed now :-). Please review the changes once more. Thanks!

allancole:
All above issue have been fixed. Gonna go ahead and merge this.

PR #817, opened 2019-05-21
All style variations: Add style for posts list block

iamtakashi:
Posts List block is coming soon, and this makes sure the block styled accordingly.

**Testing instruction:**
* Download or clone the [plugin](https://github.com/Automattic/posts-list-block).
* Activate the plugin and add the block to a page.
* Check the posts list on the page to see something odd or out of place.

**Note:**
The style is not meant to be match exactly with the themes' archive style as the markup is different. It's only meant to be something close to the style.

### DIFF
--- a/calm-business/sass/blocks/_blocks.scss
+++ b/calm-business/sass/blocks/_blocks.scss
@@ -1100,13 +1100,16 @@
 		}
 	}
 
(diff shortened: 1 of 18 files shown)


iamtakashi:
Ready for review and merge :)

PR #828, opened 2019-05-22
Multiple Themes: Fix column block responsive pattern

### LINKED ISSUE #813: Multiple Themes: Column block issue in small screens
Although we made some changes in #752 and #753, there are still some issues with smaller screens.

Some themes stop wrapping columns at 600px. This is problematic as the column width gets too small.

Screen Shot 2019-05-18 at 01 08 38

This style should be removed from the following themes.
```
@media only screen and (min-width: 600px)
.entry .entry-content .wp-block-columns {
    flex-wrap: nowrap;
}
```
- Calm
- Friendly
- Modern
- Sophisticated

Once you removed the above, another problem becomes apparent. Column block stops wrapping columns at 782px. But some themes add side margins to the columns at 768px. This makes the column breaks completely. The side margin should be added from 782px in the following themes.

Screen Shot 2019-05-18 at 01 09 58

- Calm
- Elegant
- Friendly
- Modern

Also, Professional adds side margins to all the column. That should be removed from the first and the last child.

Screen Shot 2019-05-18 at 01 04 01

iamtakashi:
Fixes #813

**Testing:**
* Add a column block to a page. Make it to have more than 2 columns
* Add an image to each column (easier to spot the margin between columns)
* Test the responsiveness by changing the viewport from small to desktop size.

### DIFF
--- a/calm-business/sass/blocks/_blocks.scss
+++ b/calm-business/sass/blocks/_blocks.scss
@@ -781,11 +781,7 @@
 			padding-right: $size__spacing-unit;
 		}
 
-		@include media(mobile) {
-			flex-wrap: nowrap;
-		}
-
-		@include media(tablet) {
+		@include media(tabletalt) {
 
 			.wp-block-column {
 				margin-left: #{0.5 * $size__spacing-unit};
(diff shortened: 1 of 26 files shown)


iamtakashi:
Ready for test and merge :)

PR #831, opened 2019-05-24
806 entry meta structural fix

jeffikus:
@iamtakashi I've taken out the margin from the archive page as well, retaining the bottom margin for Sophisticated Business as well. 

Please give 1 final test and let me know when you are happy so I can get this merged and deployed.


iamtakashi:
Looks good to me 👍

PR #869, opened 2019-05-29
Modern Business: Add font annotations

danieldudzic:


pbking:
Considering that this is using typekit, which we are moving away from, and that Global styles is forthcoming and should ultimately be leveraged in lieu of this solution this will be closed.

Issue #818, opened 2019-05-21
AltoFocus: Cookie Consent Widget shows only on Single Posts and Archives


kriskorn:
## Steps to replicate

Add the Cookie Consent widget with AltoFocus theme active.

## Result

No cookies banner is shown on the site when you browse it.

## Expected

The cookie banner to show on the site, as the widget has been added.

The theme says this:

> AltoFocus comes with one widget area below single posts and archives.

Would it be possible to enhance the theme to show Footer widgets on all pages, so that a cookie consent could be added with this theme?

* User report: 12927316-hc

mikeicode:
Issue also reported 2831289-zen

katinthehatsite:
Also reported in 3099688-zen

PR #872, opened 2019-05-29
AltoFocus: Allow widgets to be shown on the homepage

kriskorn:

- Allows showing widgets on homepage so that all widgets could work correctly.
In this instance, the Cookie Consent & Banner widget would only show on archive and single post pages, but users would like to show it right on the homepage when a visitor lands on the site. 

#### Related issue(s):

- Fix for #818


scruffian:
This looks fine. It won't work on pages or the front page, but maybe that's ok.

Issue #846, opened 2019-05-24
Independent Publisher 2: Add basic style for Posts List block


iamtakashi:
Posts List block is coming soon, and let's make sure it's not out of the place in this popular theme.

PR #876, opened 2019-05-31
Independent Publisher 2: adds posts list block styles

jeffikus:
- Adds styles for Posts List Block
- Please test with using the block on a Page as well as Single Post.

#### Related issue(s):
#846


iamtakashi:
The theme has smaller `entry-title` and `entry-content` for smaller screens. Can you add that to the ones in the Posts List block as well?

jeffikus:
@iamtakashi added the styles, please review. Thanks!

iamtakashi:
LGTM 👍

PR #877, opened 2019-05-31
Radcliffe 2: adds styles for Posts List Block

### LINKED ISSUE #830: Radcliff2: Add basic style for Posts List block
Posts List block is coming soon, and let's make sure it's not out of the place in this popular theme.

jeffikus:
- Adds Styles for Posts List Block
- Test on single Page and single post

@allancole I've made the title and meta text-transform to uppercase as per some of the other styles in the theme. I've also lined up the button states to match. Please let me know if you don't agree with the design choices I've made!

#### Related issue(s):
#830

### DIFF
--- a/radcliffe-2/assets/css/blocks.css
+++ b/radcliffe-2/assets/css/blocks.css
@@ -805,3 +805,103 @@ ul.wp-block-gallery li {
 	background-color: #2b6e9d;
 }
 
+/*--------------------------------------------------------------
+8.0 Blocks - Posts List Block
+--------------------------------------------------------------*/
+
+li.a8c-posts-list__item {
+	margin-top: 0;
+}
+
+li.a8c-posts-list__item article {
+	margin-top: .875em;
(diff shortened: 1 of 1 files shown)


iamtakashi:
Since this theme only shows the title and metas in index/archive view, we need some design decision. I think it's best to replicate sort of its single view.

The difference in typography is glaring at the moment. And the theme has style packs that changes the appearance, and the block style should also "largely" match with each style.

Screen Shot 2019-06-03 at 11 53 48
_The screenshot was taken with the default style._

jeffikus:
@iamtakashi ok, in the screenshot on the right - that's the single post page right? I think if we want to get everything to line up, @allancole is the best person to do a full audit.

iamtakashi:
OK, no worries. I'll take this one then.

iamtakashi:
I've worked on a PR that includes all style pack versions in #902

Issue #885, opened 2019-06-03
Calm Business: Visual Inconsistency in Page Title between the editor and the site.


iamtakashi:
At least, alignment and font weight should match.

**Site:**
<img width="1281" alt="Screen Shot 2019-06-03 at 14 39 49" src="https://user-images.githubusercontent.com/908665/58806165-7d83d780-860d-11e9-99f6-47d40e911fdc.png">

**Editor:**
<img width="1281" alt="Screen Shot 2019-06-03 at 14 39 52" src="https://user-images.githubusercontent.com/908665/58806180-85dc1280-860d-11e9-9e79-a7b0c9177355.png">

PR #917, opened 2019-06-04
Calm Business: 885 Fixes editor styles for page title and content mismatches.

jeffikus:
- Fixes editor styles for page title and the content editor frame padding.

#### Related issue(s):
#885

### DIFF
--- a/calm-business/style-editor.css
+++ b/calm-business/style-editor.css
@@ -54,18 +54,22 @@ body .wp-block[data-align="full"] {
 
 @media only screen and (min-width: 768px) {
   .wp-block {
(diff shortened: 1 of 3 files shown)


iamtakashi:
LGTM 👍

jeffikus:
@iamtakashi I added 1 more commit here - I didn't notice it earlier because the font size mismatch here is so tiny, it's 1.25px difference. But it's properly fixed now like the others. Will merge unless you have any objections?

iamtakashi:
Thanks for the further fix! The font size now matches with the front-end of the site. Let's merge it 👍

PR #954, opened 2019-06-07
Sophisticated Business: Fix excess text side padding for the full wid…

### LINKED ISSUE #916: Sophisticated Business: Excess side paddings in the text column when it's stacked

Screen Shot 2019-06-04 at 18 22 30

danieldudzic:
…th stacked Media & Text. Closes #916

### DIFF
--- a/modern-business/sass/blocks/_blocks.scss
+++ b/modern-business/sass/blocks/_blocks.scss
@@ -213,7 +213,7 @@
 			}
 
 			.wp-block-media-text__content {
-				padding: 0 #{2 * $size__vertical-spacing-unit} 0 0;
+				padding: 0;
 			}
 
 			&.alignfull {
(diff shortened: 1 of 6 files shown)


iamtakashi:
It looks good in small screens, but we need some slight adjustments.

Sophisticated has a vertical line on the left of the text with a viewport larger than 1168px.

Screen Shot 2019-06-17 at 14 29 05

So there are three states that effect white space inside the column.
- Small screen (Looks good with this change)
- Tablet https://d.pr/free/i/iQsPTJ
- Desktop https://d.pr/free/i/OoQP0S

danieldudzic:
@iamtakashi Could you please provide some more details on how exactly the table and desktop views should be adjusted?

> Tablet https://d.pr/free/i/iQsPTJ
> Desktop https://d.pr/free/i/OoQP0S

My understanding is, that the space above the text dynamically increases based on the image height.

adjusts_height

iamtakashi:
Sorry for not being extra clear. I was talking about the whitespaces on both sides.

iamtakashi:
I've added further fix, and I also reverted seems-to-be accidental commit for Modern business in this branch.

PR #958, opened 2019-06-07
Modern Business: Fix excess text side padding for the full width stac…

### LINKED ISSUE #915: Modern Business: Excess side paddings in the text column when it's stacked

Screen Shot 2019-06-04 at 18 20 29

danieldudzic:
…ked Media & Text. Closes #915

### DIFF
--- a/modern-business/sass/blocks/_blocks.scss
+++ b/modern-business/sass/blocks/_blocks.scss
@@ -191,6 +191,14 @@
 			}
 		}
 
+		&.alignfull .wp-block-media-text__content {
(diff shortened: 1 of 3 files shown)


iamtakashi:
It looks good in a small screen, but with this change, it looks strange on a tablet and larger. For those screens, the following rule that the theme has should be respected.

```
@media only screen and (min-width: 768px)
.entry .entry-content .wp-block-media-text .wp-block-media-text__content {
    padding: 0 64px 0 0;
}
```

iamtakashi:
LGTM 👍

PR #959, opened 2019-06-07
Elegant Business: Fix excess text side padding for the full width sta…

### LINKED ISSUE #914: Elegant Business: Excess side paddings in the text column when it's stacked

Screen Shot 2019-06-04 at 18 15 30

danieldudzic:
…cked Media & Text. Closes #914


iamtakashi:
The padding change needs to be limited to when the columns are stacked in small screens.

iamtakashi:
LGTM 👍

PR #997, opened 2019-06-20
Fix fstream security notice

jeffikus:

- Updates package file to fix security issue with fstream.


iamtakashi:
After trying `npm install` I can see ```found 0 vulnerabilities``` 👍

PR #1050, opened 2019-07-05
Independent Publisher 2: Changed display mode of summary tag

SirPL:
By default the `<summary>` tag displays black triangle.
Due to inconsistent behavior between browsers, by design Firefox does not show the triangle when summary's `display` mode is set to `block`. To show the triangle summary should be displayed as a `list-item`. The behavior is discribed in [Firefox documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary#Default_style). This pull request brings consistency back.

#### Related issue(s):
None

### DIFF
--- a/independent-publisher-2/style.css
+++ b/independent-publisher-2/style.css
@@ -80,9 +80,12 @@ hgroup,
 main,
 menu,
 nav,
-section,
-summary {
+section {
 	display: block;
+}
+
+summary {
+	display: list-item;
 }
 
 audio,


sixhours (approved):
Yep, this works much better. A belated thanks for your contribution!

Screen Shot 2021-05-10 at 4 47 39 PM

sixhours:
Synced in r58798-wpcom-themes

PR #1057, opened 2019-07-11
Update modern business to support the full site editing plugin header

apeatling:

This change will update the modern business theme to pull in the header styles into the editor through the style-editor.scss file. This means that when using the full site editing plugin, you will see almost a one to one representation of the header design in the editor vs the front end.

To test this:

* Install the modern business theme locally
* Create a new page
* Confirm that the contents and design of the header match the front end closely (there will be some difference in spacing between the blocks)

#### Related issue(s):

### DIFF
--- a/modern-business/sass/site/header/_site-header.scss
+++ b/modern-business/sass/site/header/_site-header.scss
@@ -2,6 +2,7 @@
 
 .site-header {
 	padding: 1em;
+	text-align: center;
 
 	&.featured-image {
 		display: flex;
@@ -67,7 +68,9 @@
 
 // Site title
 
-.site-title {
+.site-title, 
+.wp-block-a8c-site-title,
+.wp-block-a8c-site-title:focus {
 	color: $color__text-main;
 	margin: 0;
 	order: 2;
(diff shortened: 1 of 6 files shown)


iamtakashi:
Thanks for working on this. Here are the screenshots.
**Front of the site**
Screen Shot 2019-07-12 at 16 52 32

**Page Editor**
Screen Shot 2019-07-12 at 16 46 34

**Template Part Editor**
Screen Shot 2019-07-12 at 16 46 26

* Is the bullet listed menu item expected at this stage?

Other than that and some spacing, the style matches with the front of the site.

apeatling:
Thanks for looking Takashi!

The menu is a known bug with the navigation block, we'll be looking at it here: https://github.com/Automattic/wp-calypso/issues/34429

Once this is fixed, it should also just fix in the theme since the markup will be what the theme is expecting.

In terms of spacing, Gutenberg needs space between blocks for inserters. It's possible to override this but it gets a bit janky. I think we can look at ways of reducing this, but this should be good enough for an initial version.

PR #1058, opened 2019-07-12
Update the Friendly Business theme to support FSE header

apeatling:

This change will update the friendly business theme to pull in the header styles into the editor through the style-editor.scss file. This means that when using the full site editing plugin, you will see almost a one to one representation of the header design in the editor vs the front end:

Screen Shot 2019-07-11 at 9 42 33 PM

To test this:

* Install the friendly business theme locally
* Create a new page
* Paste the following into the editor: https://codepen.io/apeatling/pen/VJNZVJ and update/publish the header template.
* Confirm that you see a two column header that matches the screenshot above, and this looks identical on the front end of the site.

#### Related issue(s):

### DIFF
--- a/friendly-business/style-editor.scss
+++ b/friendly-business/style-editor.scss
@@ -6,6 +6,9 @@ Friendly Business Editor Styles
 
 @import "sass/variables-site/variables-site";
 @import "sass/mixins/mixins-master";
+@import "sass/site/header/site-header";
+@import "sass/navigation/menu-main-navigation";
+@import "sass/typography/headings";
 
 /** === Editor Frame === */
 
@@ -868,6 +871,16 @@ ul.wp-block-archives,
 	.wp-block-latest-comments__comment-date {
 		font-size: $font__size-xs;
 	}
+}
+
+/** === Site Header Block == **/
+
+.wp-block-columns.site-branding {
+
+	.editor-inner-blocks {
+		width: 100%;
+	}
+
 }
 
 /** === Classic Editor === */
--- a/friendly-business/sass/site/header/_site-header.scss
+++ b/friendly-business/sass/site/header/_site-header.scss
@@ -29,7 +29,7 @@
 
 .site-branding {
 
-	align-items: center;
+	align-items: flex-start;
 	color: $color__text-dark;
 	display: flex;
 	flex-wrap: wrap;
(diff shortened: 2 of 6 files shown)


iamtakashi:
LGTM 👍

iamtakashi:
Sorry, I missed this. Like #1059, the content width is now 100% in the page editor.

Screen Shot 2019-07-12 at 18 41 12

apeatling:
Thanks for testing @iamtakashi. The full width is this issue https://github.com/Automattic/wp-calypso/issues/34558 but it's not related to the header changes, that will be fixed separately.

gwwar:
Saving folks a click from the codepen content:
```
<!-- wp:columns {"align":"full","className":"site-header site-branding"} -->
<div class="wp-block-columns alignfull site-header site-branding"><!-- wp:column -->
<div class="wp-block-column"><!-- wp:a8c/site-title /-->

<!-- wp:a8c/navigation-menu /--></div>
<!-- /wp:column -->

<!-- wp:column -->
<div class="wp-block-column"><!-- wp:a8c/site-description /--></div>
<!-- /wp:column --></div>
<!-- /wp:columns -->
```

friendly-business/style-editor.scss
```diff
@@ -6,6 +6,9 @@ Friendly Business Editor Styles
 
 @import "sass/variables-site/variables-site";
 @import "sass/mixins/mixins-master";
+@import "sass/site/header/site-header";
+@import "sass/navigation/menu-main-navigation";
```
gwwar: We can leave this for a follow up but I don't think the navigation styling here is being picked up. Note the list style, vs Andy's summary screenshot

Screen Shot 2019-07-17 at 10 46 46 AM


We probably also would like to pull in `sass/navigation/navigation` instead for the other menu types

noahtallen: I can't believe I missed that! I had a hunch based on some previous conversations, but it turns out I did not have a primary menu created in the customizer on my site. Once I made a primary menu there, the style looks correct in the template editor:

Screen Shot 2019-07-17 at 11 28 07 AM

gwwar: Good catch @noahtallen! Adding a main menu did the trick

Screen Shot 2019-07-17 at 11 31 35 AM

noahtallen: I think we're tracking it here: https://github.com/Automattic/wp-calypso/issues/34429 I remember it came up in the modern business PR

gwwar: Perfect ✨

apeatling:
I'll follow up with the colour difference in a separate PR, thanks for the reviews!

PR #1059, opened 2019-07-12
Update the Professional Business theme to support FSE header

apeatling:

This change will update the professional business theme to pull in the header styles into the editor through the style-editor.scss file. This means that when using the full site editing plugin, you will see almost a one to one representation of the header design in the editor vs the front end:

Screen Shot 2019-07-11 at 10 38 51 PM

To test this:

* Install the professional business theme locally
* Create a new page
* Paste the following into the editor: https://codepen.io/apeatling/pen/VJNZVJ and update/publish the header template.
* Confirm that you see a two column header that matches the screenshot above, and this looks identical on the front end of the site.

#### Related issue(s):

### DIFF
--- a/friendly-business/sass/site/header/_site-header.scss
+++ b/friendly-business/sass/site/header/_site-header.scss
@@ -29,7 +29,7 @@
 
 .site-branding {
 
-	align-items: center;
+	align-items: flex-start;
 	color: $color__text-dark;
 	display: flex;
 	flex-wrap: wrap;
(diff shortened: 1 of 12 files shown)


iamtakashi:
I've noticed something that I might missed in the PR for Friendly Business. 

In the page editor, the content now seems to be 100% width. It'd be easier to see the difference in the screenshots below.

**Master**
Screen Shot 2019-07-12 at 18 34 35

**With this branch**
Screen Shot 2019-07-12 at 18 34 29

apeatling:
Followed up on the reasons for full width here https://github.com/Automattic/themes/pull/1058#issuecomment-510982803 👍

apeatling:
Thanks for the reviews!

PR #1061, opened 2019-07-12
Update the Elegant Business theme to support FSE header

apeatling:

This change will update the elegant business theme to pull in the header styles into the editor through the style-editor.scss file. This means that when using the full site editing plugin, you will see almost a one to one representation of the header design in the editor vs the front end:

Screen Shot 2019-07-12 at 9 51 26 AM

To test this:

* Install the elegant business theme locally
* Create a new page
* Paste the following into the editor: https://codepen.io/apeatling/pen/rEbvMN and update/publish the header template.
* Confirm that you see a header that matches the screenshot above, and this looks identical on the front end of the site.

#### Related issue(s):

### DIFF
--- a/elegant-business/sass/site/header/_site-header.scss
+++ b/elegant-business/sass/site/header/_site-header.scss
@@ -2,6 +2,7 @@
 
 .site-header {
 	padding: 1.5em 1em 1em;
+	text-align: center;
 
 	&.featured-image {
 		display: flex;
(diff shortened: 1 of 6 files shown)


noahtallen:
The header looks good to me! I'm also seeing that 100% width thing in the page editor like some folks mentioned on the other PRs:

**Editor:**
Screen Shot 2019-07-12 at 11 01 44 AM

**Site:**
Screen Shot 2019-07-12 at 11 01 53 AM

noahtallen:
Is this being caused by the FSE plugin per https://github.com/Automattic/wp-calypso/issues/34558?

apeatling:
Yes, that's that issue. It's separate from the header improvements, but needs to be addressed.

PR #1062, opened 2019-07-13
Modern Business: add FSE support

vindl:

This adds required functionality to support rendering FSE header and
footer on all site pages (static and dynamic). It's also includes the
data population functionality, which has now been removed from the plugin
and will be delegated to each theme for maximum flexibility.

#### Testing instructions:

This PR requires its FSE counterpart PR to work properly. Please follow the testing instructions provided there: https://github.com/Automattic/wp-calypso/pull/34640

### DIFF
--- a/modern-business/inc/fse-template-data.php
+++ b/modern-business/inc/fse-template-data.php
@@ -1,0 +1,324 @@
+<?php
+/**
+ * Template data inserter file.
+ *
+ * @package full-site-editing
+ */
+
+/**
+ * Class A8C_WP_Template_Data_Inserter
+ */
+class A8C_WP_Template_Data_Inserter {
+	/**
+	 * This function will be called on plugin activation hook.
+	 */
+	public function insert_default_template_data() {
+		$current_theme_name = get_option( 'stylesheet' );
+		/**
+		 * This site option will be used to indicate that template data has already been
+		 * inserted for this theme, in order to prevent this functionality from running
+		 * more than once.
+		 */
+		$fse_template_data_option = $current_theme_name . '-fse-template-data';
+
+		if ( get_option( $fse_template_data_option ) ) {
+			/*
+			 * Bail here to prevent inserting the FSE data twice for any given theme.
+			 * Multiple themes will still be able to insert different template parts.
+			 */
+			return;
+		}
+
+		$this->register_template_post_types();
+
+		$header_id = wp_insert_post(
+			[
+				'post_title'     => 'Header',
+				'post_content'   => $this->get_header_content(),
+				'post_status'    => 'publish',
+				'post_type'      => 'wp_template_part',
+				'comment_status' => 'closed',
+				'ping_status'    => 'closed',
+			]
+		);
+
+		if ( ! term_exists( "$current_theme_name-header", 'wp_template_part_type' ) ) {
+			wp_insert_term( "$current_theme_name-header", 'wp_template_part_type' );
+		}
+
+		wp_set_object_terms( $header_id, "$current_theme_name-header", 'wp_template_part_type' );
+
+		$footer_id = wp_insert_post(
+			[
+				'post_title'     => 'Footer',
+				'post_content'   => $this->get_footer_content(),
+				'post_status'    => 'publish',
+				'post_type'      => 'wp_template_part',
+				'comment_status' => 'closed',
+				'ping_status'    => 'closed',
+			]
+		);
+
+		if ( ! term_exists( "$current_theme_name-footer", 'wp_template_part_type' ) ) {
+			wp_insert_term( "$current_theme_name-footer", 'wp_template_part_type' );
+		}
+
+		wp_set_object_terms( $footer_id, "$current_theme_name-footer", 'wp_template_part_type' );
+
+		$page_template_id = wp_insert_post(
+			[
+				'post_title'     => 'Page Template',
+				'post_content'   => $this->get_template_content( $header_id, $footer_id ),
+				'post_status'    => 'publish',
+				'post_type'      => 'wp_template',
+				'comment_status' => 'closed',
+				'ping_status'    => 'closed',
+			]
+		);
+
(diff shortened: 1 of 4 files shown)


modern-business/inc/fse-template-data.php
```diff
@@ -0,0 +1,324 @@
...
+/**
+ * Template data inserter file.
+ *
+ * @package full-site-editing
+ */
+
+/**
+ * Class A8C_WP_Template_Data_Inserter
+ */
+class A8C_WP_Template_Data_Inserter {
```
apeatling: What if instead of including all of the logic in the theme, you instead keep the logic in the plugin and expose helper functions that the theme calls to register template parts, and provide the default block template content (or block template name to request from the API)?

Even though we are using one base theme that all child themes inherit from, that may not always be the case, and you're opening the possibility of future copy-pasta. Abstracting away the hard work, but providing helper functions matches the general philosophy of WordPress and theming.

modern-business/inc/fse-template-data.php
```diff
@@ -0,0 +1,324 @@
...
+
+		add_option( $fse_template_data_option, true );
+	}
+
+	/**
+	 * Returns default header template part content.
+	 *
+	 * @return string
+	 */
+	public function get_header_content() {
```
apeatling: We talked about this in a call, but I think this data should be coming from an API.

modern-business/inc/fse-template-data.php
jeraldjuice: There's some discussion on this in the other PR, btw. :) 

https://github.com/Automattic/wp-calypso/pull/34640#issuecomment-511231921

modern-business/inc/fse-template-data.php
vindl: Agreed, that that would be a lot better than hardcoding it here. I don't see any blockers to doing that atm. 👍

modern-business/inc/fse-template-data.php
vindl: Ideally I'd like to do that and prevent the need for copy-pasta in the future. On the other hand I'm also trying to avoid posing the requirement that FSE plugin needs to be installed/active prior to theme activation - which might be problematic if the plugin code it relies on is missing.

Ideally I'd like FSE theme to be able to operate entirely independent of FSE plugin - there should be a fallback non-FSE header, and data population should work. That way, if the plugin is activated at a later time, FSE functionality will kick in immediately.

modern-business/inc/fse-template-data.php
```diff
@@ -0,0 +1,324 @@
...
+		// TODO: replace with header blocks once they are ready.
+		return '<!-- wp:group {"className":"site-header site-branding"} -->' .
```
mmtr: I have not found any specific doc indicating how many Gutenberg versions do we support in FSE, but given that we still use the older `@wordpress/editor` package rather than `@wordpress/block-editor` ([#](https://github.com/Automattic/wp-calypso/pull/34096)), it seems that we want to support at least the 2 most recent WP versions (currently 5.1 and 5.2) like Jetpack does. 

Since the group block is not present yet in the current WP version ([#](https://github.com/WordPress/gutenberg/issues/15490)), should we avoid using it?

vindl: I think this part is required to make Modern Business header layout look as expected in the editor (cc @apeatling).

Since we are now targeting only Dotcom/Atomic, in the short term we can probably rely on the fact that we'll be running the latest version in both environments

vindl:
Thanks for testing @mmtr!

> If I manually remove a template part, it is not created again automatically after switching back to Modern Business.

I kind of relied here on the fact that template parts won't be exposed and hence deletable. I could improve it to check for entries based on taxonomy (instead of site option) and reinsert them when some part is missing.

> Since the data is created after switching to the theme, any existing site currently using the Modern Business theme won't have FSE functionality. Might not be an issue if we want to target only new sites.

That's a good point - this depends on how we want to handle existing Modern Business sites (cc @apeatling @gwwar). In case we need to populate data for all such sites, we could make data population execute on `after_setup_theme`. This would effectively trigger it on each load, so we'll have to be very careful not to create duplicate entries.

> For some reason, in my WP.com test site, the template/template parts has been created 5 times when switching to the theme.

Good catch! This is weird given that it's gated/marked by site option. Wondering if it's related to option caching or maybe the `pub/` prefix.

jeraldjuice:
Stumbled across something today - this triggers an error when doing the following:
* Both the FSE plugin and Modern Theme are active on a site.
* Deactivate the FSE plugin.
* Try to activate the FSE plugin again. Failure because the Template Inserter class is already defined.

Screen Shot 2019-07-17 at 7 07 02 PM

mmtr:
> Stumbled across something today - this triggers an error when doing the following:
> 
> * Both the FSE plugin and Modern Theme are active on a site.
> * Deactivate the FSE plugin.
> * Try to activate the FSE plugin again. Failure because the Template Inserter class is already defined.
> 
> Screen Shot 2019-07-17 at 7 07 02 PM

Seems this is caused by having the theme branch diverging from the calypso branch: https://github.com/Automattic/wp-calypso/pull/34640#issuecomment-512196709

vindl:
> Stumbled across something today - this triggers an error when doing the following:
> Both the FSE plugin and Modern Theme are active on a site.
> Deactivate the FSE plugin.
> Try to activate the FSE plugin again. Failure because the Template Inserter class is already defined.

Since the corresponding Calypso PR has been merged, this should no longer be the case. I'm assuming this was occurring while testing this PR with Calypso master that tried to declare that class to with old FSE plugin approach.

vindl:
Lots of great feedback in the comments that I think should be incorporated. ✨
In order to prevent this PR from getting too bug and to keep things moving, I'm going to merge this as is for now. This should also help us parallelize some of the follow up improvements.

Thank you all for reviews and suggestions!

PR #1072, opened 2019-07-16
Full Site Editing: Fix post content nested block alignment on modern business

noahtallen:
- Make the post content block 100% width. As a result, the post content block takes up the same area that the post editor container takes up without the post content block. This is important because the theme styles do alignment of blocks based on the style of outer 
- Allow nested blocks inside the post content block to maintain the default alignment styles for top level blocks. Previously, they were targeted as 2nd level blocks, and were thus being set to a much wider width.

I did this by targeting `.wp-block[data-type="a8c/post-content"]` because we need to set the width of the outer block container. Additionally, I needed to override the default block alignment styles for 2nd level blocks inside the , so I used `.wp-block[data-type="a8c/post-content"] .wp-block` to target those. I tried using `.post-content-block .wp-block` but it didn't appear to override the styles, so I had to stick with the data type selector.

#### Testing:
1. Activate the modern business theme from this PR.
2. Also activate FSE with this companion FSE plugin PR (alternatively, just turn off full width alignment on the page content block): https://github.com/Automattic/wp-calypso/pull/34691
3. Create a new page from the pages screen.
4. Select the "portfolio" template
5. The full width blocks (gallery & video) should extend to the edge of the screen.
6. The normal paragraphs should be narrower in the middle.
7. Ultimately, the style of the post content should look identical to how it does in the non-fse post editor, so please play around with blocks and alignments that might not be covered by this PR yet!

_Update: Testing for template part blocks_
1. Activate this theme update.
1. In the page editor, select the header template part block and turn off full width alignment. _I'll be opening a PR soon which disables alignment on template part blocks, which has the same effect._
3. Make sure that the style of the Modern Business header is still looking correct.
4. Find a block like the gallery insert it into the header, setting it to full width. Save the update and go back to the page editor. That new block should extend screen to screen. 

#### Follow up tasks:
- [x] Make sure the header/footer styles are updated to work with alignment disabled.
* Make the same changes to the other 2019 child themes (will be followup PRs so long as this approach is viable.)

#### Related issue(s):
https://github.com/Automattic/wp-calypso/issues/34558

### DIFF
--- a/modern-business/style-editor.css
+++ b/modern-business/style-editor.css
@@ -811,14 +811,14 @@ body {
   font-weight: 300;
 }
 
-body .wp-block[data-type="a8c/template"],
 body .wp-block[data-align="full"] {
   width: 100%;
 }
 
 @media only screen and (min-width: 600px) {
-  body .wp-block[data-type="a8c/template"],
+  
   body .wp-block[data-align="full"],
+  body .wp-block[data-type="a8c/template"],
   body .wp-block[data-type="a8c/post-content"] .wp-block[data-align="full"] {
     width: calc( 100% + 90px);
     max-width: calc( 100% + 90px);
@@ -835,8 +835,9 @@ body .wp-block[data-align="full"] {
   body .wp-block[data-type="a8c/post-content"] .wp-block[data-align="wide"] {
     width: 100%;
   }
-  body .wp-block[data-type="a8c/template"],
+  
   body .wp-block[data-align="full"],
+  body .wp-block[data-type="a8c/template"],
   body .wp-block[data-type="a8c/post-content"] .wp-block[data-align="full"] {
     position: relative;
     left: calc( -12.5% - 14px);
--- a/modern-business/style-editor.css
+++ b/modern-business/style-editor.css
@@ -797,6 +797,15 @@ blockquote {
   }
 }
 
+/** === Helper Functions === */
+/**
+ Given a string, $alignment, returns the nested FSE block selectors
+ required for Full Site Editing to look correct. If $alignment is given,
+ we use it to target wp-blocks with a matching data-align attribute.
+ If no alignment is given, we target the base wp-block class. In both
+ of these scenarios, we target top level blocks, and any top level blocks
+ inside one of the FSE template part blocks.
+ */
 /** === Editor Frame === */
 body {
   font-weight: 300;
@@ -807,7 +816,9 @@ body .wp-block[data-align="full"] {
 }
 
 @media only screen and (min-width: 600px) {
+  
   body .wp-block[data-align="full"],
+  body .wp-block[data-type="a8c/template"] .wp-block[data-align="full"],
   body .wp-block[data-type="a8c/post-content"] .wp-block[data-align="full"] {
     width: calc( 100% + 90px);
     max-width: calc( 100% + 90px);
@@ -819,50 +830,65 @@ body .wp-block[data-align="full"] {
     max-width: 80%;
     margin: 0 10%;
   }
+  
   body .wp-block[data-align="wide"],
+  body .wp-block[data-type="a8c/template"] .wp-block[data-align="wide"],
   body .wp-block[data-type="a8c/post-content"] .wp-block[data-align="wide"] {
     width: 100%;
   }
+  
   body .wp-block[data-align="full"],
+  body .wp-block[data-type="a8c/template"] .wp-block[data-align="full"],
   body .wp-block[data-type="a8c/post-content"] .wp-block[data-align="full"] {
     position: relative;
     left: calc( -12.5% - 14px);
     width: calc( 125% + 116px);
     max-width: calc( 125% + 115px);
   }
+  
   body .wp-block[data-align="right"],
+  body .wp-block[data-type="a8c/template"] .wp-block[data-align="right"],
   body .wp-block[data-type="a8c/post-content"] .wp-block[data-align="right"] {
     max-width: 125%;
   }
 }
 
 /** === Content Width === */
+
 .wp-block,
+.wp-block[data-type="a8c/template"] .wp-block,
 .wp-block[data-type="a8c/post-content"] .wp-block {
   width: calc(100vw - (2 * 1rem));
   max-width: 100%;
 }
 
 @media only screen and (min-width: 768px) {
+  
   .wp-block,
+  .wp-block[data-type="a8c/template"] .wp-block,
   .wp-block[data-type="a8c/post-content"] .wp-block {
     width: calc(8 * (100vw / 12));
   }
 }
 
 @media only screen and (min-width: 1168px) {
+  
   .wp-block,
+  .wp-block[data-type="a8c/template"] .wp-block,
   .wp-block[data-type="a8c/post-content"] .wp-block {
     width: calc(6 * (100vw / 12 ));
   }
 }
 
+
 .wp-block .wp-block,
+.wp-block[data-type="a8c/template"] .wp-block .wp-block,
 .wp-block[data-type="a8c/post-content"] .wp-block .wp-block {
   width: 100%;
 }
 
-/** === FSE Post Content Block Override === */
+/** === FSE Template Part Blocks === */
+.wp-block[data-type="a8c/template"],
 .wp-block[data-type="a8c/post-content"] {
   width: 100%;
 }
--- a/modern-business/style-editor.css
+++ b/modern-business/style-editor.css
@@ -811,14 +811,14 @@ body {
   font-weight: 300;
 }
 
+body .wp-block[data-type="a8c/template"],
 body .wp-block[data-align="full"] {
   width: 100%;
 }
 
 @media only screen and (min-width: 600px) {
-  
+  body .wp-block[data-type="a8c/template"],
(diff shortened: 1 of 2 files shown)


modern-business/style-editor.scss
```diff
@@ -21,7 +21,7 @@ body {
 
 	@include media(mobile) {
 
-		.wp-block[data-align="full"] {
+		.wp-block[data-align="full"], .wp-block[data-type="a8c/post-content"] .wp-block[data-align="full"] {
```
mmtr: According to the [WordPress CSS Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/css/#structure), each selector should be in its own line.
```suggestion
		.wp-block[data-align="full"], 
		.wp-block[data-type="a8c/post-content"] .wp-block[data-align="full"] {
```

modern-business/style-editor.scss
```diff
@@ -34,26 +34,26 @@ body {
 			margin: 0 10%;
 		}
 
-		.wp-block[data-align="wide"] {
+		.wp-block[data-align="wide"], .wp-block[data-type="a8c/post-content"] .wp-block[data-align="wide"] {
```
mmtr:
```suggestion
		.wp-block[data-align="wide"], 
		.wp-block[data-type="a8c/post-content"] .wp-block[data-align="wide"] {
```

modern-business/style-editor.scss
```diff
@@ -34,26 +34,26 @@ body {
...
 			width: 100%;
 		}
 
-		.wp-block[data-align="full"] {
+		.wp-block[data-align="full"], .wp-block[data-type="a8c/post-content"] .wp-block[data-align="full"] {
```
mmtr:
```suggestion
		.wp-block[data-align="full"], 
		.wp-block[data-type="a8c/post-content"] .wp-block[data-align="full"] {
```

modern-business/style-editor.scss
```diff
@@ -34,26 +34,26 @@ body {
...
 			position: relative;
 			left: calc( -12.5% - 14px );
 			width: calc( 125% + 116px );
 			max-width: calc( 125% + 115px ); // Subtract 1px here to avoid the rounding errors that happen due to the usage of percentages.
 		}
 
-		.wp-block[data-align="right"] {
+		.wp-block[data-align="right"], .wp-block[data-type="a8c/post-content"] .wp-block[data-align="right"] {
```
mmtr:
```suggestion
		.wp-block[data-align="right"], 
		.wp-block[data-type="a8c/post-content"] .wp-block[data-align="right"] {
```

modern-business/style-editor.scss
```diff
@@ -34,26 +34,26 @@ body {
...
 			max-width: 125%;
 		}
 	}
 }
 
 /** === Content Width === */
 
-.wp-block {
+.wp-block, .wp-block[data-type="a8c/post-content"] .wp-block {
```
mmtr:
```suggestion
.wp-block, 
.wp-block[data-type="a8c/post-content"] .wp-block {
```

modern-business/style-editor.scss
noahtallen: Thanks for the tip! I'll try to get some sort of plugin going in VS Code so I see any of those errors while I'm working. I'll also get those fixed in the meantime!

noahtallen:
Update: I added support for the template part blocks in the same manner. I also added a helper function to generate the selectors to reduce verbosity in the scss file. I could really use some eyes on that code as it's my first shot at writing a sass function!

This will work in tandem with a yet to be created PR which disables alignment options for all template part blocks.

I edited the description to add testing instructions for this.

modern-business/style-editor.css
```diff
@@ -807,7 +816,10 @@ body .wp-block[data-align="full"] {
 }
 
 @media only screen and (min-width: 600px) {
-  body .wp-block[data-align="full"] {
+  
+  body .wp-block[data-align="full"],
+  body .wp-block[data-type="a8c/template"] .wp-block[data-align="full"],
```
apeatling: Instead of using these data attribute selectors, is it possible for us to set a class on these blocks like we do for any other block?

Using the data attribute selectors will be fragile because this could change at any point in a future Gutenberg update.

noahtallen: Great question! I did attempt this, but I wasn't able to get it working. The problem is that the theme css targets the width of every `.wp-block`, so we needed to somehow override that for the post content block. The custom class name (`.post-content-block`) we have set for this block is much further down the DOM tree, so adjusting it doesn't really help us override the size of the actual block:
Screen Shot 2019-07-17 at 2 46 32 PM

I'm not sure how we would otherwise target that top level block div. I'm very open to suggestions; as a beginner, there may be some Gutenberg way to do this I'm not aware of, or even a more clever css selector!

gwwar: Blocks generate classes with the following pattern: `wp-block-${namespace}-${block}-${name}`. These are usually generated for us, but in the case of dynamic blocks we need to generate that ourselves since we fully control the render callback.

So for a more stable selector we can do something like:
`body.block-editor-page .wp-block-a8c-post-content`

After updating the dynamic block render callbacks to include this.

noahtallen: Hm, I'm not sure if this targets what I'd like to target.

For example, with a paragraph block, I would want to target the outer layer which is selected. But it looks like the autogenerated classname is much further down the dom tree:
Screen Shot 2019-07-17 at 4 02 16 PM

Maybe I'm missing something?

mmtr: `.wp-block[data-type]` seems ok to me. Since we need to modify the outer layer of the block, I don't think we can use any other selector.

gwwar: Summarizing a slack chat with @noahtallen and @vindl and @gwwar  I think we'll +1 @mmtr's suggestion and stick with this.

If we're afraid of behavior changing, we can add a follow up E2E test to see if this still exists before upgrading Gutenberg, and we can add a note to PCYsg-mfI-p2 for followup.

modern-business/style-editor.scss
```diff
@@ -10,6 +10,28 @@ Modern Business Editor Styles
...
+
+/**
+ Given a string, $alignment, returns the nested FSE block selectors
+ required for Full Site Editing to look correct. If $alignment is given,
+ we use it to target wp-blocks with a matching data-align attribute.
+ If no alignment is given, we target the base wp-block class. In both
+ of these scenarios, we target top level blocks, and any top level blocks
+ inside one of the FSE template part blocks.
+ */
+@function get-block-selector($alignment) {
```
mmtr: Note that we usually include these functions in a separate file called `_functions.scss`.

Also, you can use [annotations](http://sassdoc.com/annotations/) to change the structure of the docstring and specify the params/returns:

```css
/**
 * Get the nested FSE block selectors
 * required for Full Site Editing to look correct. If $alignment is given,
 * we use it to target wp-blocks with a matching data-align attribute.
 * If no alignment is given, we target the base wp-block class. In both
 * of these scenarios, we target top level blocks, and any top level blocks
 * inside one of the FSE template part blocks.
 * @param {String} $alignment - Alignment
 * @return {String}
 */
```

modern-business/style-editor.scss
```diff
@@ -21,7 +43,7 @@ body {
 
 	@include media(mobile) {
 
-		.wp-block[data-align="full"] {
+		#{get-block-selector("full")} {
```
mmtr: This syntax looks a bit awkward to me. Wouldn't a mixin be a better fit for this? Or is it just me not being used to use functions like this?

noahtallen:
I updated this with the same thing that I added on #1086 - template part blocks now get the same styles as full width blocks without the toggle. It's easier to see the change on the other PR :)

modern-business/style-editor.scss
noahtallen: Good call! I'll make sure to add those if we stick with this function

modern-business/style-editor.scss
noahtallen: Yeah that's a great question. The goal here was just to reduce the amount of verbosity on this line - otherwise it would have several lines of selectors. And since those selectors are repeated so often, I figured it might be worthwhile to see what a function could do in scss. I wasn't sure if a mixin would do the same thing. Do those basically allow you to use parameters to create reusable styles? In this case I wanted to do that with a selector instead of a style.

mmtr: > Do those basically allow you to use parameters to create reusable styles?

You can also use them to create reusable selectors and inject the given styles with `@content`. This is actually how the mixins used for the responsive breakpoints work:

```
@function get-block-selector( $alignment ) {
	// ...
}

@mixin blocks($alignment) {
	#{get-block-selector( $alignment )} {
		@content;
	}
}

body {
	@include media(mobile) {
		@include blocks("full") {
			width: calc( 100% + 90px );
			max-width: calc( 100% + 90px );
		}
	}
}
```

I found that code more legible, but as I said maybe it's just me not being used to use functions on selectors :) So feel free to disregard this idea!

noahtallen: Neat! I feel like it adds a bit of extra boilerplate, but I can definitely see how it's easier to read inline

PR #1086, opened 2019-07-18
Fix block alignment in FSE context for friendly business

noahtallen:
This extends the changes proposed in https://github.com/Automattic/themes/pull/1072 to the friendly business theme. It also fixes header/footer alignment issues mentioned in the other PR which would affect themes with wide headers like this one.

#### Testing:
1. Activate the friendly business theme from this PR.
2. Also activate FSE with this companion FSE plugin PR: https://github.com/Automattic/wp-calypso/pull/34691
3. Create a new page from the pages screen.
4. Select the "small business" template
5. The full width blocks (side by side text/picture) should extend to the edge of the screen.
6. The normal paragraphs should be narrower in the middle, and most of the stuff should be centered in the middle.
7. Ultimately, the style of the post content in the page editor with FSE should look identical to how it does in the non-fse post editor, so please play around with blocks and alignments that might not be covered by this PR yet!
8. Make sure the header matches the header on the front of the site both in the header editor and in the page editor.

### DIFF
--- a/friendly-business/style-editor.css
+++ b/friendly-business/style-editor.css
@@ -849,17 +849,29 @@ blockquote {
   }
 }
 
+/** === Helper Functions === */
+/**
+ Given a string, $alignment, returns the nested FSE block selectors
+ required for Full Site Editing to look correct. If $alignment is given,
+ we use it to target wp-blocks with a matching data-align attribute.
+ If no alignment is given, we target the base wp-block class. In both
+ of these scenarios, we target top level blocks, and any top level blocks
+ inside one of the FSE template part blocks.
+ */
 /** === Editor Frame === */
 body {
   background: #fffdf6;
 }
 
+body .wp-block[data-type="a8c/template"],
 body .wp-block[data-align="full"] {
   width: 100%;
 }
 
 @media only screen and (min-width: 600px) {
-  body .wp-block[data-align="full"] {
+  body .wp-block[data-type="a8c/template"],
+  body .wp-block[data-align="full"],
+  body .wp-block[data-type="a8c/post-content"] .wp-block[data-align="full"] {
     width: calc( 100% + 90px);
     max-width: calc( 100% + 90px);
   }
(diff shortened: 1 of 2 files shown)


noahtallen:
Screen Shot 2019-07-17 at 6 22 48 PM
it looks great, but my CSS hack to disable the alignment doesn't work in the top toolbar mode.

apeatling:
Tested and this is working with the Small Business template. One update we need to make to this block is to change it to "Wide Width" to work correctly in the editor and front end.

noahtallen:
Gotcha! Do we want to make that change for both in the editor and on the front end? In my browser it looks full width on the front end too:

**Front End:**
Screen Shot 2019-07-19 at 9 06 59 AM

**Editor:**
Screen Shot 2019-07-19 at 9 07 07 AM

gwwar:
As a general note before/after screenshots work better than text descriptions

gwwar:
With Marko's approach landed, to test this properly now, I think we need to provide similar content defaults within the theme. Since we'd like to move out that default content from the theme to some API or similar, shall we land changes here so far and follow up on additional fixes after?

noahtallen:
That makes sense to me. The overall styles for this won't change as we add the header/footer population, and this at least makes the post content for this theme look correct in an FSE context in the meantime :)

noahtallen:
@apeatling Could you create an issue for me if you'd still like follow up on the wide align?

gwwar:
> @apeatling Could you create an issue for me if you'd still like follow up on the wide align?

@noahtallen feel free to add one! You're welcome to DM me if you had any questions on that.

gwwar:
Ah @apeatling I'm not sure if I read https://github.com/Automattic/themes/pull/1086#issuecomment-512891299 correctly, but are you seeing visual issues when we toggle "wide align", or that "wide align" should be the default for the page template?

apeatling:
I'll need to recheck, but setting wide width for that specific block on this template seemed to make it match in the editor and front end.

noahtallen:
Ok, gotcha! I wasn't seeing that on my end. Do you know if it happened at a particular width? Maybe it's a responsive size issue

Issue #1087, opened 2019-07-18
_Dsgnsystm: Pullquote text alignment inconsistencies


danieldudzic:
**Left and Right aligned Pullquotes in the editor:**

![Editor](https://user-images.githubusercontent.com/905781/61454238-a31c3480-a960-11e9-95ec-985ae00939b1.jpg)

----------

**Left and Right aligned Pullquotes on the front-end:**
![Front-end](https://user-images.githubusercontent.com/905781/61454308-d52d9680-a960-11e9-9dce-50c0375d1a86.jpg)

-----------
In the editor the text is aligned to the center, and on the front-end the Left aligned text is aligned to the left, and the Right aligned text is aligned to the center.

allancole:
Ahhh, good catch! I’ll sort that out now…

allancole:
Fixed in #1092 
Closing…

PR #1092, opened 2019-07-18
_dsgnsystm: Fix text-alignment and padding in pullquote block

allancole:

- Fixes text alignment in pullquote block
- Fixes left padding on blockquotes in pullquote block

**Before**

image

**After**

image

#### Related issue(s):

#1087

### DIFF
--- a/_dsgnsystm/sass/blocks/pullquote/_editor.scss
+++ b/_dsgnsystm/sass/blocks/pullquote/_editor.scss
@@ -10,6 +10,10 @@
 	border-bottom-width: #{map-deep-get($config-pullquote, "border", "width")};
 	color: #{map-deep-get($config-global, "color", "foreground", "default")};
 
+	blockquote {
+		padding-left: 0;
+	}
+
 	p {
 		font-family: #{map-deep-get($config-pullquote, "font", "family")};
 		font-size: (strip-unit(map-deep-get($config-heading, "font", "size", "h4")) + 0em);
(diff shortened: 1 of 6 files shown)


iamtakashi:
In the editor, because of the lack of max-width that are specified for the front of the site, it's a lot wider. (as you can see the right aligned pull quote is pushed down) But this looks like a core bug.

Screen Shot 2019-07-18 at 19 12 10

allancole:
@iamtakashi Yeah, I’m thinking we should remove the `max-width: 50%;` setting on all left/right aligned blocks. Even when a block is floated we don’t want to set an explicit width without also giving users a way to change that width in the editor. Using the image block as an example, a customer can set the image width to 25%, 50%, 75%, etc. If a customer chooses 25% in the editor, that’s what they get. They should also get 25% on the frontend, but with this rule in place they’d be stuck with 50% on the frontend with no way to change it.

In you’re screenshot above, the two blocks are actually floating naturally and since the right-aligned block has one extra character, it’s slightly too wide to float in that instance, but with less text it should float properly.

iamtakashi:
> I’m thinking we should remove the max-width: 50%; setting on all left/right aligned blocks. Even when a block is floated we don’t want to set an explicit width without also giving users a way to change that width in the editor.

👍

allancole:
That last commit fixes the max-width issue mentioned here: https://github.com/Automattic/themes/pull/1093

Gonna go ahead and merge this.

Issue #1084, opened 2019-07-17
_Dsgnsystm: Collapsed Button Width


danieldudzic:
![Aligned_buttons](https://user-images.githubusercontent.com/905781/61418603-4f2b3480-a8fb-11e9-9dd7-cfb69480f270.jpg)

-------
They are displaying correctly in the editor, though:
![Editor_buttons](https://user-images.githubusercontent.com/905781/61418661-7da90f80-a8fb-11e9-8608-735bdda6faf3.jpg)

iamtakashi:
It looks like this is the cause.

```
header *, main *, footer * {
    max-width: inherit;
}
```

danieldudzic:
This actually seems to be caused by:

`.alignleft {
    max-width: 50%;
}`

and

`.alignright {
    max-width: 50%;
}`

iamtakashi:
Oh. right. That's what the button is inherited.

I can understand the intention of those max-widths for the `.alignleft` and `.alignright`, but it might cause unexpected result like this when everything inside is inheriting the max-width. 

I suppose this is something new in _dsgnsystm? I don't see this rule in Twenty Nineteen. @allancole any opinion?

allancole:
> I can understand the intention of those max-widths for the .alignleft and .alignright, but it might cause unexpected result like this when everything inside is inheriting the max-width.

Yeah, that approach might be a bit too aggressive if it’s effecting the buttons negatively. I’ll work out a fix that excludes buttons (and maybe a few other elements).

allancole:
This is now fixed in #1092 . 
Closing this issue…

PR #1093, opened 2019-07-18
_dsgnsystm: Fix max-width for left/right aligned button blocks

allancole:

- Reset the max-width on buttons with align left/right utility classes.

**Before**

image

**After**

image

#### Related issue(s):

- Fixes: #1084

### DIFF
--- a/_dsgnsystm/sass/blocks/button/_style.scss
+++ b/_dsgnsystm/sass/blocks/button/_style.scss
@@ -28,6 +28,11 @@ input[type="submit"],
  */
 .wp-block-button {
 
+	&.alignleft,
+	&.alignright {
+		max-width: inherit;
+	}
+
 	// Outline Style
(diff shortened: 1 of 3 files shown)


iamtakashi:
I'll share the concern. The max width on the utility class feels too aggressive. How about we remove the max-width and not fixing the buttons?

allancole:
Good point and I agree. 

I’ll remove the max-width from the utility class, and instead place it only on the blocks that need this behavior: Cover, Image, Video, Embed, Quote, & Pullquote.

iamtakashi:
Discussed and decided to remove `max-width: 50%;` [here](https://github.com/Automattic/themes/pull/1092#issuecomment-512981118). After that, we shouldn't need this fix.

allancole:
With #1092 merged we no longer need this patch. Closing…

PR #1107, opened 2019-07-20
Varia: Reduce gap between header and content when homepage title is hidden.

iamtakashi:
Since this is for WP.com only, it's a bit awkward to test :)

Before:
Screen Shot 2019-07-19 at 19 08 48

After:
Screen Shot 2019-07-19 at 19 23 35

### DIFF
--- a/varia/inc/style-wpcom.css
+++ b/varia/inc/style-wpcom.css
@@ -9,7 +9,7 @@
 	display: none;
 }
 
-.home.page.hide-homepage-title .hentry:not(.has-post-thumbnail) .entry-content > .alignfull:first-child {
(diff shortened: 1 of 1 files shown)


varia/inc/style-wpcom.css
```diff
@@ -9,7 +9,7 @@
 	display: none;
 }
 
-.home.page.hide-homepage-title .hentry:not(.has-post-thumbnail) .entry-content > .alignfull:first-child {
+.home.page.hide-homepage-title .site-main > article > .entry-content {
```
allancole: Yup, this looks good. Merging…

PR #1109, opened 2019-07-20
Radcliffe 2: Update scrrenshot

### LINKED ISSUE #1108: Radcliffe2: Update screenshot
Since we launched Redhill, we'll need to retire Radcliffe 2, but since the theme has been referenced in many places, it might take a bit of time. Update the screenshot so that it's not so much the same.

iamtakashi:
Fixes #1108 

screenshot


iamtakashi:
@davidakennedy Do you think this is ok to update Radcliffe2 screenshot so that it's less similar to Redhill :) ?

iamtakashi:
I'm going to ahead with this.

Issue #1113, opened 2019-07-20
Redhill: Link Missing Style


carinapilar:
## Steps to replicate

1. Create a new page on your site
2. Add the "Media & Text" block on it
3. Add a hyperlinked text on the Text area of the block
4. Publish the page
5. Activate the Redhill theme

## Result

The link on this block isn't being highlighted, as follows:

![Links](https://user-images.githubusercontent.com/16104522/61584030-7f4f1f00-ab17-11e9-95bd-cef2f07a8d4c.png)

## Expected

For the link to highlight as it does on a Paragraph block.

* User report - ticket: #2194551-zd

thomasguillot:
Hi @carinapilar, this should now be fixed 😊

Also

* 745a0b8
* fa9df39
* be4d855

iamtakashi:
@thomasguillot Could you make PRs when you make a change to what's in production so that we can review and discuss the approach? I don't think I'd agree with the approach you took. I'd keep the style (inherited link text colour), but I'd make it's only applied to when background colour is added to either the background of the block or the paragraph.

catkit:
I understand this hasn't been published yet to production, so once it is let us know so we can tell user, please. Thanks.

iamtakashi:
@carinapilar @catkit The fix to add underline to text links has been deployed.

PR #1122, opened 2019-07-22
Varia: Have links inherit the color only when the block (Media & Text or Paragrpah) are using a background color

thomasguillot:
Plus Rebuild Redhill

#### Related issue(s):
See #1113

### DIFF
--- a/redhill/style-editor.css
+++ b/redhill/style-editor.css
@@ -444,7 +444,15 @@ object {
 	padding-left: 0;
 }
 
-.wp-block-media-text .block-editor-inner-blocks a {
+.wp-block-media-text[style*="background-color"]:not(.has-background-background-color) .block-editor-inner-blocks a {
+	color: currentColor;
+}
+
+p.has-background {
+	padding: 16px 16px;
+}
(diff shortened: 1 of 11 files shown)


iamtakashi:
This works well in Varia, but for Redhill, it'd be better if we restore the underline for links in body text because without it there is no way to tell this is a link.

Screen Shot 2019-07-22 at 21 48 38

Apart from the obvious links such as the post title or the site title, we should always keep underline on links for a better a11y rather than relying on a colour difference.

@thomasguillot going forward, make a PR per theme. Since there needs to be additional fix for Redhill (restoring the underline), can you remove the fix for Redhill from this PR? And make a separate PR for Redhill. Thanks in advance!

allancole:
I’m gonna go ahead and merge this as it is. We can correct course for next time. I tried breaking up the fixes into separate PRs but it isn’t as simple as I thought it would be. 

I’ll sort out the Redhill issue as a separate PR.

iamtakashi:
Thanks for sorting this out, @allancole.

PR #1147, opened 2019-07-24
Varia: Fix responsive logic for better mobile view

allancole:
Fixes responsive logic to prevent content/text from reaching the absolute edge of any container or viewport edge. Effects Group Block, Cover Block, Pullquote Block, and Gallery Blocks.

- Revises responsive logic Sass rules to allow for smarter padding on small screens. 
- Focuses the responsive logic on `.alignwide` and `.alignfull` since normal width have the correct width already. 
- Sets 50% max-width on `.alignright` and `.alignleft` Gallery Blocks.

#### Before:

varia-align-broke

#### After:

varia-align-fix

#### To test:

- Try the Group Block, Cover Block, Pullquote Block, and Gallery Blocks and make sure the text contained within them never touches the edge of the screen or the widest edge of its parent container.

#### Related issue(s):

Found while testing locally.

### DIFF
--- a/varia/sass/abstracts/_mixins.scss
+++ b/varia/sass/abstracts/_mixins.scss
@@ -77,9 +77,9 @@
  * Align container widths
  * - Sets a fixed-width on content within alignwide and alignfull blocks
  */
-@mixin align-container-width( $width: 100%, $multiplier: "2", $columns: "12" ) {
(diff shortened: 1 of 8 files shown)


iamtakashi:
Nice improvement! 👍

PR #1148, opened 2019-07-24
Varia: Improve spacer blocks on Mobile

allancole:

This PR limits the spacer-block to a 16px height (to match mobile vertical margins). This might be a bit aggressive and may not expected, but it does greatly improve the scrolling experience on mobile for the templates I tried this with. When we need to, we can use empty new lines in text areas in our templates to `fake` a taller spacer area on mobile, if needed.

This PR also forces zero vertical margins on mobile screens, which is also aggressive (as it uses `!important`) but still necessary to maintain consistency. 

#### Before: Varied spacer heights makes for a lot of awkward empty spaces.

varia-spacers-before

#### After: Consistent 16px for _all_ spacers makes the layout feel more natural and considered.

varia-spacers-after

#### Related issue(s):

Discovered while testing locally.

### DIFF
--- a/varia/sass/abstracts/_mixins.scss
+++ b/varia/sass/abstracts/_mixins.scss
@@ -5,7 +5,7 @@
 @mixin media( $res ) {
 
 	@if mobile-only == $res {
-		@media only screen and (max-width: calc(map-deep-get($config-global, "breakpoint", "sm") - 1)) {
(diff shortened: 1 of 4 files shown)


iamtakashi:
I think it's worth to try this approach. Until the spacer block becomes more robust to allow different height for mobile devices.

allancole:
Agreed 👍. 
Merging this…

PR #1152, opened 2019-07-26
Varia: Introduce a Lighter Foreground (#FAFAFA)

thomasguillot:
This allows users to create blocks with a very subtle background colour for example, which we don't support at the moment without having to rely on a custom colour.


varia/sass/abstracts/_config-global.scss
```diff
@@ -63,6 +63,7 @@ $config-global: (
 			"default": #444444,
 			"light": #767676, // must be accesible!
 			"dark": #111111, // must be accesible!
+			"lighter": #FAFAFA,
```
allancole: Yup, I support this change 👍. The only thing I’d want to think more about is wether or not it should be in the `foreground` part of the array. Assuming this is more likely to be used as a background color, it might be better if we extend the `background` array to something like this: 

```
"background": (
    "default": white,
    "light": #FAFAFA,
    "dark": #CCCCCC,
),
```

thomasguillot: That's a great idea! Even more flexibility.

thomasguillot: Reworked this in 881115c

allancole:
This looks good @thomasguillot 👍. 

It also exposes a new consideration we’ll need to document at some point later. Due to how the system works, if we merge this to `master` and re-compile `redhill` for example, it’ll throw an error. The array in `redhill/sass/_config-child-theme-deep.scss` is missing the new _sub-array_ for `"background-color": (...)`, which causes the compiler to fail. To fix this, we just need to adjust the arrays in `redhill` (and all other child themes) to include the new _sub-array_. 

In cases like this where we’re specifically **changing/updating** the variables array, I think we’ll have to also change the arrays in all the child themes in the same PR. This will prevent unexpected compiling failures for others working on child themes. I don’t anticipate us having to update the variables array very often, but when we do we’ll just need to make sure to keep this caveat in mind. 

cc: @iamtakashi @jeffikus

iamtakashi:
>  if we merge this to master and re-compile redhill for example, it’ll throw an error. 

I was wondering about this.

> In cases like this where we’re specifically changing/updating the variables array, I think we’ll have to also change the arrays in all the child themes in the same PR.

Sounds like a good idea for this sort of case. 👍

thomasguillot:
@iamtakashi @allancole I have updated Brompton, Leven and Redhill. Let me know if it works!

PR #1154, opened 2019-07-26
Varia: Simply hides the wpstats image :-)

allancole:

- Hide the wpstats image

#### Related issue(s):

- Noticed this on Redhill and Leven demos at the very bottom of the page.

### DIFF
--- a/varia/sass/components/_imports.scss
+++ b/varia/sass/components/_imports.scss
@@ -15,3 +15,5 @@
 @import "comments/comments";
 
 @import "widgets/widgets";
+
+@import "utilities/utilities";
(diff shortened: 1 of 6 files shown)


allancole:
I added this fix to `media.scss`, but I just noticed it ends yp in the editor where it’s not needed. There might be a better place for this CSS. Gonna think on it a bit more before merging.

allancole:
Added a `/sass/components/utilities` directory for this and other _utility_ styles that exist outside of the editor context. 

These styles do not get compiled to the editor, so I’ll go ahead and merge this.

PR #1167, opened 2019-07-31
Modern Business: add remote template fetching for FSE

vindl:

Previously our template parts data was hardcoded in the theme.
This has been updated so it is now fetched from the WordPress.com API.

Fixes https://github.com/Automattic/wp-calypso/issues/34924

#### Testing instructions

1. Sandbox the public API.
2. Apply D31186-code to your sandbox.
3. Map this theme to your local Gutenberg Docker dev environment, in a similar fashion you do for FSE plugin: `{themes_repo_path}/themes/modern-business:/var/www/html/wp-content/themes/modern-business`
4. On your local test site, delete all of the Templates that were previously created.
5. Delete `modern-business-fse-template-data` option if it was set on your site. Otherwise this would prevent data population here https://github.com/Automattic/themes/pull/1167/files#diff-8573df061382abec8228a31b5eb02880R84.
6. Activate Modern Business theme to trigger the activation hook.
7. Make sure that Templates are populated with correct content.
8. Verify that FSE editing and rendering still works as before.

### DIFF
--- a/modern-business/inc/fse-template-data.php
+++ b/modern-business/inc/fse-template-data.php
@@ -6,20 +6,78 @@
  */
 
 /**
- * Class A8C_WP_Template_Data_Inserter
+ * Class A8C_WP_Templates_Data_Inserter
  */
-class A8C_WP_Template_Data_Inserter {
+class A8C_WP_Templates_Data_Inserter {
+	/*
+	 * Modern Business header content.
+	 *
+	 * @var string $header_content
+	 */
+	private $header_content;
+
+	/*
+	 * Modern Business footer content.
+	 *
+	 * @var string $footer_content
+	 */
+	private $footer_content;
+
+	/*
+	 * Current theme slug.
+	 *
+	 * @var string $theme_slug
+	 */
+	private $theme_slug;
+
+	/**
+	 * A8C_WP_Template_Parts_Data_Inserter constructor.
+	 */
+	public function __construct() {
+		$this->theme_slug = get_option( 'stylesheet' );
+		$this->header_content = '';
+		$this->footer_content = '';
+	}
+
+	/**
+	 * Retrieves template parts content from WP.com API.
+	 */
+	public function fetch_template_parts() {
+		$request_url = 'https://public-api.wordpress.com/wpcom/v2/full-site-editing/templates';
+
+		$request_args = [
+			'body' => [ 'theme_slug' => $this->theme_slug ],
+		];
+
+		$response = wp_remote_get( $request_url, $request_args );
+
+		if ( is_wp_error( $response ) ) {
+			return;
+		}
+
+		$api_response = json_decode( wp_remote_retrieve_body( $response ), true );
+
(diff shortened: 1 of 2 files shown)


glendaviesnz:
Nice! I tested locally and it mostly seems to work as advertised. 

It initially didn't add the template data, but was soon as I tried doing some debugging it worked - at that point all the template parts worked as before.

Because of the initial failure I added and removed the template data and `modern-business-fse-template-data` flag multiple times, and 8/10 times it worked as expected. The times that it didn't, just deactivating and activating the theme a couple more times made it work. I wasn't able to successfully work out why. I suspect it was just a flakey connection between my local FSE install and the sandboxed APIs or something causing the header or footer to return empty. 

Probably for a follow up pull request, but I guess we need to notify the user somehow if the call to the api failed and the required template data has not been loaded.

It is probably worth at least one other person testing this locally to see if they get the same intermittent issues with template data not loading.

modern-business/inc/fse-template-data.php
```diff
@@ -29,41 +87,49 @@ public function insert_default_template_data() {
 			return;
 		}
 
+		// Set header and footer content based on data fetched from the WP.com API.
+		$this->fetch_template_parts();
+
+		// Avoid creating template parts if data hasn't been fetched properly.
+		if ( empty( $this->header_content ) || empty( $this->footer_content ) ) {
```
Copons: Are we disregarding the possibility of a site without some template parts (e.g. no footer)?
What about an `&&` here?

vindl: I was thinking about the same thing while writing this code, and I don't have a strong preference yet. I went for `||` at the time because all of the themes from our current scope provide both, and if that's not the case something has gone wrong in the process and we better bail.

In general case though `&&` might make more sense because of added flexibility. On the other hand, having an FSE editor that doesn't show the footer goes against our goal of matching the content in the editor and the front end as closely as possible? 🤔 I'm thinking we could make it a minimum requirement for themes to provide a header and footer.

gwwar: I think the current logic is 👍 We can revisit if we ever end up with a theme that doesn't have a footer.

gwwar:
Great work @vindl ! This tested well for me. Very lovely that we're pulling content from out of themes. As for next steps:

- Let's have a quick conversation sorting out how theme folks would like to build content in ( D31013-code ). 
- What folks pick should should inform the API structure in D31012-code.
- Any needed changes from the above being added to this PR

apeatling:
Great! Worked first time as expected for me after removing the option, deactivating the FSE plugin, switching to a different theme, reactivating the plugin, then switching to modern business.

vindl:
Updated for demo site approach, this one is ready for another round of reviews and testing.

Copons:
⚠️ This does not work because I've accidentally wiped the FSE template site option and can't seem to be able to recreate it with the correct value. ⚠️

PR #1168, opened 2019-07-31
Varia: Add spacing override utility classes

### LINKED ISSUE #1160: Varia: Add vertical spacing utility classes
## Enhancement

After playing around with many 3rd party Gutenberg plugins, I noticed a lot of them come with special “utility classes” that are helpful for making per-block design decisions. Currently our system enforces a pretty strict 32px vertical spacing unit between _all_ blocks, but there are some cases where we may want to reduce that spacing to communicate a relationship between  blocks. If we add some utility classes as outlined below, we can make these kinds of decisions in our templates in a non-destructive way. 

- `.zero-top-margin`: Removes top margin entirely.
- `.half-top-margin`: Reduces the top margin to half of the defined vertical spacing unit variable.
- `.default-top-margin`: Add a top margin based on the vertical spacing unit variable.
- `.zero-bottom-margin`: Removes top margin entirely.
- `.half-bottom-margin`: Reduces the top margin to half of the defined vertical spacing unit variable.
- `.default-bottom-margin`: Add a bottom margin based on the vertical spacing unit variable.

**Note:** When fixing this, we’ll likely need to remove the overriding `half` spacing that happens for text elements within the M&T and Group Blocks.

allancole:

Adds vertical spacing overrides with the following classes:

- `.margin-top-none`
- `.margin-top-half`
- `.margin-top-default`
- `.margin-bottom-none`
- `.margin-bottom-half`
- `.margin-bottom-default`

These classes can be used to add extra spacing considerations for the editor and the frontend using the “Additional CSS Classes” field in the editor as shown here: 

image

Later, this functionality will likely be replaced by a UI based tool, similar to what CoBlocks offers which also relies on the “Additional CSS Classes” field. 

This PR also addresses a few vulnerabilities in the package files for some child-themes.

#### To Do

- [x] Make sure spacing overrides work on Frontend
- [x] Make sure spacing overrides work in the editor
- [ ] Compile spacing overrides into all child-themes

#### Related issue(s):

Fixes #1160

### DIFF
--- a/brompton/package-lock.json
+++ b/brompton/package-lock.json
@@ -558,7 +558,7 @@
       "requires": {
         "bluebird": "3.5.3",
         "chokidar": "2.1.1",
-        "lodash": "4.17.13",
+        "lodash": "4.17.11",
         "yargs": "12.0.5"
       },
       "dependencies": {
@@ -1221,7 +1221,7 @@
           }
         },
         "lodash": {
-          "version": "4.17.13",
+          "version": "4.17.11",
           "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.13.tgz",
           "integrity": "sha512-cQKh8igo5QUhZ7lg38DYWAxMvjSAKG0A8wGSVimP07SIUEK2UO+arSRKbRZWtelMtN5V0Hkwh5ryOto/SshYIg==",
           "dev": true
@@ -2640,7 +2640,7 @@
       "dev": true,
       "requires": {
         "glob": "~7.1.1",
-        "lodash": "~4.17.13",
+        "lodash": "~4.17.10",
         "minimatch": "~3.0.2"
       }
     },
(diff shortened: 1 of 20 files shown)


allancole:
@takashi I’m starting to think that we should add all margins and all paddings to this. That way, in addition to what we have above we could also have the following: 

`.margin-left-none`
`.margin-left-half`
`.margin-left-default`
`.margin-right-none`
`.margin-right-half`
`.margin-right-default`

`.padding-top-none`
`.padding-top-half`
`.padding-top-default`
`.padding-right-none`
`.padding-right-half`
`.padding-right-default`
`.padding-bottom-none`
`.padding-bottom-half`
`.padding-bottom-default`
`.padding-left-none`
`.padding-left-half`
`.padding-left-default`

What do you think?

iamtakashi:
Sure, why not. It might be useful one day :)

allancole:
Coool. I’ll sort this out shortly 👍

allancole:
This PR only adds some extra classes and can’t actually break anything. Gonna go ahead and merge.

PR #1172, opened 2019-08-02
Modern Business: Full Site Editing - add header and footer selectors for block front end styles

glendaviesnz:

Add  `.fse-template-content` class to block styles in order to apply the correct front end block styles to the FSE header and footer content.

#### Testing
- Check out https://github.com/Automattic/wp-calypso/pull/35154 in your FSE test environment
- Build theme and copy or link to local FSE install
- Add a Paragraph block to the Header and Footer template parts and set the background colour to one of the primary colours
- View the page in the front end and check that the blocks have a background colour

**Before**

before

**After**

after

#### Related issue(s): 
https://github.com/Automattic/wp-calypso/issues/35058
https://github.com/Automattic/wp-calypso/pull/35154

### DIFF
--- a/modern-business/sass/blocks/_blocks.scss
+++ b/modern-business/sass/blocks/_blocks.scss
@@ -1,7 +1,8 @@
 /* !Block styles */
 
 .entry .entry-content > *,
-.entry .entry-summary > * {
+.entry .entry-summary > * 
+.site-header > *, .site-footer > * {
 	margin: $size__vertical-spacing-unit 0;
 	max-width: 100%;
 
@@ -106,7 +107,8 @@
  * - helps with plugin compatibility
  */
 .entry .entry-content,
-.entry .entry-summary {
+.entry .entry-summary,
+.site-header, .site-footer {
 
 	.entry-content,
 	.entry-summary,
@@ -123,7 +125,7 @@
 	}
 }
 
-.entry .entry-content {
+.entry .entry-content, .site-header, .site-footer {
 
 	//! Headers
 	& > h1,
--- a/modern-business/sass/blocks/_blocks.scss
+++ b/modern-business/sass/blocks/_blocks.scss
@@ -2,7 +2,7 @@
 
 .entry .entry-content > *,
 .entry .entry-summary > * 
-.site-header > *, .site-footer > * {
+.fse-template-content > * {
 	margin: $size__vertical-spacing-unit 0;
 	max-width: 100%;
 
@@ -108,7 +108,7 @@
  */
 .entry .entry-content,
 .entry .entry-summary,
-.site-header, .site-footer {
+.fse-template-content {
 
 	.entry-content,
 	.entry-summary,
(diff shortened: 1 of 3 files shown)


Copons:
So, to recap, we have two solutions to this issue:

1. Modify the theme CSS by adding all `.entry .entry-content` style to `.site-header` and `.site-footer`.
2. Modify the FSE template parts by adding `.entry .entry-content` to their wrapper when rendered on the front-end.

Both approaches have tradeoffs, but I think 1 would potentially mess up a site big time if FSE gets disabled, and the template parts are replaced by the original header and footer.

Currently, 2 would mean hardcoding theme classes in FSE, but that is changing: in https://github.com/Automattic/wp-calypso/pull/34997 we use theme classes in FSE, and my plan is to add them to the template API that @vindl is writing, so that we'll eventually drop the hardcoding.
My idea was to add to the API, say, `header_classes` and `footer_classes`, but we can also add `content_classes`.

> (currently I see different primary colours in the editor compared to front end - investigating to see if this is a separate issue)

It is a separate issue unrelated to FSE. 🙂 
I'll take care of it ASAP.

glendaviesnz:
> Both approaches have tradeoffs, but I think 1 would potentially mess up a site big time if FSE gets disabled, and the template parts are replaced by the original header and footer.
> 
> Currently, 2 would mean hardcoding theme classes in FSE

@Copons  - what about just adding a wrapper to template parts with an .fse-template-content class, eg. in /wp-calypso/apps/full-site-editing/full-site-editing-plugin/full-site-editing/utils/class-wp-template.php and adding this class to the theme _blocks.scss. This removes the risk of applying invalid styles to the header and footer if FSE is removed from a site, and still ensures that blocks styles are correctly applied to each of the template parts? But happy to just close this PR and wait until you have the theme classes added to the API.

Copons:
> > Both approaches have tradeoffs, but I think 1 would potentially mess up a site big time if FSE gets disabled, and the template parts are replaced by the original header and footer.
> > Currently, 2 would mean hardcoding theme classes in FSE
> 
> @Copons - what about just adding a wrapper to template parts with an .fse-template-content class in /wp-calypso/apps/full-site-editing/full-site-editing-plugin/full-site-editing/utils/class-wp-template.php and adding this class to the theme _blocks.scss. This removes the risk of applying invalid styles to the header and footer if FSE is removed from a site, and still ensures that blocks styles are correctly applied to each of the template parts? But happy to just close this PR and wait until you have the theme classes added to the API.

@glendaviesnz I think it's a great idea!
I'm not entirely sure of the ETA of the API, nor if my approach will be eventually used; with your suggestion we can both ship it quickly now, and in case iterate easily and painlessly on it. 👍

glendaviesnz:
> I think it's a great idea!

Ok  - gone with that approach then, so this PR is now linked with https://github.com/Automattic/wp-calypso/pull/35154

vindl:
> I'm not entirely sure of the ETA of the API, nor if my approach will be eventually used; with your suggestion we can both ship it quickly now, and in case iterate easily and painlessly on it. 👍

Just an update that the API changes have been deployed. We can proceed with this one and refactor if necessary.

glendaviesnz:
On further investigation I think we can handle this better from just the theme end without requiring the extra wrapper from the FSE plugin end, so going to close this one and put up a new one with that approach.

PR #1173, opened 2019-08-02
 wp_body_open() function added after body tag

sharazghouri:
Starting from WordPress 5.2 a new function is added – wp_body_open() – that is used to trigger a wp_body_open action. The intention of this action is to allow people to add things – like a <script> tag – directly inside the body of a page.

Ref: 
https://developer.wordpress.org/reference/functions/wp_body_open/
https://make.wordpress.org/themes/2019/03/29/addition-of-new-wp_body_open-hook/

### DIFF
--- a/affinity/header.php
+++ b/affinity/header.php
@@ -21,6 +21,7 @@
 </head>
 
 <body <?php body_class(); ?>>
+<?php wp_body_open(); ?>
 <div id="page" class="site">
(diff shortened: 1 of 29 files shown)


michaeldcain:
Fixes #1722

michaeldcain (changes requested):
Thanks for the patch - sorry it's taken us a while to respond.

Can you remove 1e7257b from the PR since it's covered in #1185, and add the `wp_body_open()` function to the Varia child themes that use their own header.php (listed in #1722)?

pbking:
Originating branch has been deleted.  Keeping closed.

PR #1174, opened 2019-08-02
Modern Business: rename template parts to templates

vindl:

Renames `wp_template_part` to `wp_template` and in general all occurrences of the word "part".

Fixes https://github.com/Automattic/wp-calypso/issues/35079

#### Testing instructions

1. Follow the testing instructions available in https://github.com/Automattic/wp-calypso/pull/35080.

### DIFF
--- a/modern-business/footer.php
+++ b/modern-business/footer.php
@@ -21,9 +21,9 @@
 
 <footer id="colophon" class="site-footer">
 	<?php
-		if ( class_exists( 'A8C_WP_Template' ) ) {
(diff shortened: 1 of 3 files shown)


Copons:
It might be because of language differences, but in my opinion those really are "template parts", and not "templates".
Like, when I think of a template, I think of the whole thing (e.g. header + content + sidebar + footer).

Also, but it's probably less important, we have been using "template parts" for a while now, and this change might be confusing, especially when we need to talk about it outside our teams.

PR #1177, opened 2019-08-02
Modern Business: Add editor palette support to the functions.php

Copons:

Editor color palettes require two parts to work:
- One in PHP where we define the colors names and codes, which is used by the editor.
- One in CSS where we assign styles to the classes automagically generated by the editor.

Modern Business only has the CSS part, while inheriting the PHP one from Twenty Nineteen.
This means that, for example, the Primary color is blue in the editor (as defined by [Twenty Nineteen `functions.php`](https://github.com/WordPress/twentynineteen/blob/e74d348ab9cfc6ad46bdf33f691611930c39e944/functions.php#L141-L171)), and pink on the front-end (as defined by [Modern Business CSS](https://github.com/Automattic/themes/blob/0afe9f26f9b91d3710fa5a9d32ac5b1d9a8210ce/modern-business/sass/blocks/_blocks.scss#L1045-L1098)).

* Add the `editor-color-palette` theme support to Modern Business, overriding the Twenty Nineteen one, to display the correct custom colors in the editor.

| Before | After | Front end (unchanged) |
| - | - | - |
| Screenshot 2019-08-02 at 13 20 07 | Screenshot 2019-08-02 at 13 21 16 | Screenshot 2019-08-02 at 13 20 47 |

### DIFF
--- a/modern-business/functions.php
+++ b/modern-business/functions.php
@@ -16,21 +16,53 @@ if ( ! function_exists( 'modern_business_setup' ) ) :
 	 * as indicating support for post thumbnails.
 	 */
 	function modern_business_setup() {
-	    /**
-	     * Add support for core custom logo.
-	     *
-	     * @link https://codex.wordpress.org/Theme_Logo
-	     */
-	    add_theme_support(
-	        'custom-logo',
-	        array(
-	            'height'      => 128,
-	            'width'       => 128,
-	            'flex-width'  => true,
-	            'flex-height' => false,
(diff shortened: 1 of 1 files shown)


iamtakashi:
Oh wow, thanks for catching this @Automattic/opportunity 

Is this something new introduced recently? I kind of think we didn't have to do this before, otherwise we would've realised about it.

Anyway, the PR works as expected. 👍

PR #1184, opened 2019-08-02
Varia: Update Cover Block for Gutenberg 6.2

allancole:

The Cover Block has no restrictions on which Blocks can be contained within the Block. It now supports pretty much all Blocks. 

#### Related issue(s):

Discovered when Gutenberg was updated recently.

#### To Do
- [x] Add uniform spacing between any elements contained within cover blocks.
- [x] Test Responsive behaviors with nested complex blocks like Columns, Media&Text, and Cover within Cover.
- [x] Test `.alignwide/full` on both the Cover block itself and any blocks contained within it that use `.alignwide/full`. Make sure it works as expected. 

#### Before:
image

#### After:
image

### DIFF
--- a/varia/sass/blocks/cover/_style.scss
+++ b/varia/sass/blocks/cover/_style.scss
@@ -79,7 +79,7 @@
 	}
 
 	&.alignwide {
-		// Make sure nested blocks respects .alignwide utility classes.
+		// Make sure nested blocks respect .alignwide utility classes.
 		& > .wp-block-cover__inner-container > .alignwide {
 			@extend %responsive-width-wide;
 		}
@@ -91,7 +91,7 @@
 	}
 
 	&.alignfull {
-		// Make sure nested blocks respects .alignwide utility classes.
+		// Make sure nested blocks respect .alignwide utility classes.
 		& > .wp-block-cover__inner-container > .alignwide {
 			@extend %responsive-width-wide;
 		}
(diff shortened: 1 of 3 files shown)


allancole:
Hey @iamtakashi, Can you give this a review when you get a free moment?

iamtakashi:
@allancole Thanks a lot for working on this. I'm excited about this change.

I have only noticed a few minor things.

* When a wide element is in a normal width Cover, the wide element overflows. I know this is an edge case.

Screen Shot 2019-08-07 at 14 13 26

* When a wide column is in wide width Cover, there is no side paddings.

Screen Shot 2019-08-07 at 14 14 49

I don't mean to block this PR. I think we can merge this once the minor issues are resolved, but I've started to wonder if we should take a little more utility approach rather than adding this sort of fix on a block by block basis. It might reduce the complexity of CSS and potential bugs in future.

allancole:
Cool. I’ll fix those two edge case conditions. 

> I've started to wonder if we should take a little more utility approach rather than adding this sort of fix on a block by block basis. It might reduce the complexity of CSS and potential bugs in future. 

Ahhh, you’re right! This logic should be “block-agnostic” so to speak, and applied to the utility class directly so that other blocks that may have similar nesting structures can work the same way. I think I’m going to make another PR that tries that approach. Good catch!

allancole:
@iamtakashi can you give this PR a try? #1218 

It should fix the two issues above and makes this change _block agnostic_ so that all blocks with nesting capabilities can have the same responsive behavior.

allancole:
Looks like we’re going to go with the solution in #1246 which introduces an improved responsive logic that produces (slightly) less CSS load. Once it’s merged we can close this.

allancole:
Closing this in favor of #1246

Issue #699, opened 2019-03-26
Independent Publisher 2: 0 minutes time stamp on all posts


rahul020691:
## Steps to replicate

1) Activate theme and write 2-3 posts

2) You will see all  show 0 minutes time stamp on the byline

## Result

Like this: https://cld.wthms.co/ODhJAa

* User report: [ #11343684-hc]
* Follow-up ticket: [ #1906000-zd]
User site: https://naprienko.blog/

nielslange:
Issue had been fixed in https://github.com/Automattic/themes/pull/1185

PR #1185, opened 2019-08-05
Independent Publisher 2:  if post word less than 250 words  post show 0 minutes time to read solved.

sharazghouri:
If post word less than 250 words. Theme show 0 Minutes read time 0 is not plural.
Before:
image

_nx() function issue it assumes 0-0.9 to the plural and only 1 to single. I fixed it if the time less than 1 I explicitly set to 1.

After:
image

#### Related issue(s):
https://github.com/Automattic/themes/issues/699


michaeldcain:
This was deployed to WordPress.com in r55048-wpcom-themes

PR #1194, opened 2019-08-05
Varia: Add bottom margin to column cells for mobile

### LINKED ISSUE #1178: Varia: Column Block has no margin on small screens
The column block can look really rough on mobile/tablet with no margin-bottom to differentiate the columns.

Screenshot 2019-08-02 at 13 33 25

iamtakashi:
Fixes #1178 

**Before**
Screen Shot 2019-08-05 at 17 35 31

**After**
Screen Shot 2019-08-05 at 17 34 59

### DIFF
--- a/varia/sass/blocks/columns/_style.scss
+++ b/varia/sass/blocks/columns/_style.scss
@@ -1,12 +1,7 @@
 .wp-block-columns {
 
 	.wp-block-column {
-
-		/* Resetting margins to match _block-container.scss */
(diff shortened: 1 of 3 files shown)


iamtakashi:
@thomasguillot Thanks for the review.

> On a side note, there's an inconsistent margin-bottom if the columns are in a Group but I suppose it's a different issue.

Can you file an issue about that so that we don't forget? Thanks in advance.

PR #1200, opened 2019-08-06
Sophisticated Business: Add missing editor color palette support

### LINKED ISSUE #1190: Sophisticated Business: Missing editor palette in functions.php
See more info in #1179 & #1177

danieldudzic:
Closes #1190

### DIFF
--- a/sophisticated-business/functions.php
+++ b/sophisticated-business/functions.php
@@ -50,9 +50,9 @@ function sophisticated_business_setup() {
 				'color' => '#b59439', // $color__border-link-hover
 			),
 			array(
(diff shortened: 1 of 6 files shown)


iamtakashi:
Instead of Dark Blue, how about Dark Gray, #1c1c1c?

iamtakashi:
This theme seems need to fix the front end as well. This is what I see and it doesn't make sense :) I'll add a fix to it.

Screen Shot 2019-08-06 at 18 50 19

iamtakashi:
**After:**

Screen Shot 2019-08-06 at 19 08 12

This should be ready to merge.

PR #1201, opened 2019-08-06
Friendly Business: Add missing editor color palette support

### LINKED ISSUE #1188: Friendly Business: Missing editor palette in functions.php
See more info in #1179 & #1177

danieldudzic:
Closes #1188

### DIFF
--- a/friendly-business/functions.php
+++ b/friendly-business/functions.php
@@ -55,7 +55,7 @@ function friendly_business_setup() {
 			array(
 				'name'  => __( 'Light Gray', 'friendly-business' ),
 				'slug'  => 'light-gray',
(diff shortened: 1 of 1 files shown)


iamtakashi:
It'd be better to not to change the slugs and the names of colours that WP.com version already has since user's customization would probably be gone if we change it. So, let's keep Dark Gray and Light Gray. Dark Gray could be `#3c2323`, and Light Gray could be `#fdf9ec`.

iamtakashi:
> Light Gray could be `#fdf9ec`

Sorry, this was a wrong suggestion. It was mismatching with the front end. I added a further fix. This is ready to merge.

PR #1211, opened 2019-08-06
Modern Business Full Site Editing  - add FSE classes to header and footer

glendaviesnz:

Add  `.fse-enabled` selectors to block css for header and footer if FSE is enabled in order to apply the correct front end block styles to the FSE header and footer content.

#### Testing
- Build modern-business theme and copy or link to local FSE install
- Checkout https://github.com/Automattic/wp-calypso/pull/35209 in your FSE environment (this is needed in order to apply the `.fse-enabled` class to the body element
- Add a Paragraph block to the Header and Footer template parts and set the background colour to one of the primary colours
- View the page in the front end and check that the blocks have a background colour - this should be due the existence of a either `.fse-template-content`  on the `site-header` and `.site-footer`  divs if testing first commit, or `fse-enabled` class on body if testing second commit.
- Disable the FSE plugin and check that the `.fse-template-content` class is no longer present on the `.site-header` and `.site-footer` divs if testing first commit, or `fse-enabled` not on body if testing second commit

**Before**

before

**After**

after

Fixes https://github.com/Automattic/wp-calypso/issues/35058

### DIFF
--- a/modern-business/functions.php
+++ b/modern-business/functions.php
@@ -127,16 +127,3 @@ function a8c_fse_insert_theme_template_data() {
 }
 
 add_action( 'after_switch_theme', 'a8c_fse_insert_theme_template_data' );
-
-/**
- * Add Full Site Editing class to body so block specific styles can be targetted
- * to theme parts if FSE enabled.
- */
-add_filter('body_class', 'fse_body_class');
-function fse_body_class($classes)
-{
-	if (class_exists('A8C\FSE\WP_Template')) {
-		$classes[] = 'fse-enabled';
-	}
-	return $classes;
-}
--- a/modern-business/functions.php
+++ b/modern-business/functions.php
@@ -127,3 +127,16 @@ function a8c_fse_insert_theme_template_data() {
 }
 
 add_action( 'after_switch_theme', 'a8c_fse_insert_theme_template_data' );
+
+/**
+ * Add Full Site Editing class to body so block specific styles can be targetted
+ * to theme parts if FSE enabled.
+ */
+add_filter('body_class', 'fse_body_class');
+function fse_body_class($classes)
+{
+	if (class_exists('A8C\FSE\WP_Template')) {
+		$classes[] = 'fse-enabled';
+	}
+	return $classes;
+}
--- a/modern-business/sass/blocks/_blocks.scss
+++ b/modern-business/sass/blocks/_blocks.scss
@@ -1,7 +1,7 @@
 /* !Block styles */
 
 .entry .entry-content > *,
-.entry .entry-summary > * 
+.entry .entry-summary > *, 
 .fse-enabled .site-header > *,
 .fse-enabled .site-footer > * {
 	margin: $size__vertical-spacing-unit 0;
--- a/modern-business/sass/blocks/_blocks.scss
+++ b/modern-business/sass/blocks/_blocks.scss
@@ -2,7 +2,8 @@
 
 .entry .entry-content > *,
 .entry .entry-summary > * 
-.fse-template-content > * {
+.fse-enabled .site-header > *,
+.fse-enabled .site-footer > * {
 	margin: $size__vertical-spacing-unit 0;
 	max-width: 100%;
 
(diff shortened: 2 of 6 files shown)


glendaviesnz:
In contrast to https://github.com/Automattic/themes/pull/1172 & https://github.com/Automattic/wp-calypso/pull/35154 this approach decouples the application of the front end block styles from the FSE plugin and allows the theme to dictate which template parts that block styles need to be applied to.

jeraldjuice:
Works great locally!

I do wonder - does it make more sense to add a class to the body of the page if the plugin is active? Considering that this is a compatibility issue rather than a direct styling issue. That approach may be easier to generalize and not duplicate into separate themes down the road.

glendaviesnz:
> I do wonder - does it make more sense to add a class to the body of the page

Yeh, this is approach has merits, so have added a commit for this, and will see which people prefer out of the two. So:

1. One approach is the original of adding fse classes to each theme part if FSE is enabled- https://github.com/Automattic/themes/pull/1211/commits/ef405376c00867899acc5ddfd40e50461bedd72a

2. Other approach suggested by @jeraldjuice is to just add a class to body if fse is enabled and use that to target block styles - https://github.com/Automattic/themes/pull/1211/commits/9922110d61c01719df6c92990eb6f5b17a8e9309

Which do people prefer?

vindl:
> Which do people prefer?

Both are fine, but I'd probably pick 2 for the following reasons: we don't have to repeat it for future template parts, and we can move this functionality from theme to the plugin. In general, any FSE logic that we include into themes will have to be duplicated, so it makes sense to keep it at a minimum in my opinion.

modern-business/functions.php
```diff
@@ -127,3 +127,16 @@ function a8c_fse_insert_theme_template_data() {
 }
 
 add_action( 'after_switch_theme', 'a8c_fse_insert_theme_template_data' );
+
+/**
+ * Add Full Site Editing class to body so block specific styles can be targetted
+ * to theme parts if FSE enabled.
+ */
+add_filter('body_class', 'fse_body_class');
```
vindl: I'd suggest moving this to a plugin (per https://github.com/Automattic/themes/pull/1211#issuecomment-519074802). Also this section is missing spaces after/before parentheses.

Copons:
> 2\. Other approach suggested by @jeraldjuice is to just add a class to body if fse is enabled and use that to target block styles - [9922110](https://github.com/Automattic/themes/commit/9922110d61c01719df6c92990eb6f5b17a8e9309)

I wonder if this has a risk of applying the post content styles to the entire site.

Now, on one hand, we might end up with, say, an image _outside the template parts_ that's supposed to be a carefully styled design element, incorrectly behaving like a post content's image.
On the other we're doing FSE here, so eh, it's kinda supposed to go that way. 😄

modern-business/functions.php
glendaviesnz: Yeh that makes sense, will do that - and will get my IDE php formatting set up properly too :-)

glendaviesnz:
> I wonder if this has a risk of applying the post content styles to the entire site.

I am using the body class to still target just the header and footer divs at this stage for this reason, eg. `.fse-enabled .site-header`

modern-business/style-rtl.css
```diff
@@ -3415,59 +3415,77 @@ footer .wp-block-a8c-navigation-menu .sub-menu {
...
+  .entry .entry-summary > *
+.fse-enabled .site-header > *,
+  .fse-enabled .site-footer > * {
     margin: 32px auto;
   }
 }
 
 .entry .entry-content > * > *:first-child,
-.entry .entry-summary > * > *:first-child {
+.entry .entry-summary > *
```
noahtallen: It looks like there are a _ton_ of changes in these files - just wanted to make sure they're expected :)

glendaviesnz: Nice catch, thanks - missing comma!

noahtallen: Just good luck haha! I didn't even realize that

modern-business/sass/blocks/_blocks.scss
```diff
@@ -1,7 +1,9 @@
 /* !Block styles */
 
 .entry .entry-content > *,
-.entry .entry-summary > * {
+.entry .entry-summary > *
```
noahtallen: is the comma also an issue here?

glendaviesnz: that is where the comma was missing, added now - you must have been looking at the previous commit

noahtallen: very strange, not sure why github hadn't caught up to that 😆 apologies!

PR #1216, opened 2019-08-07
Modern Business: remove FSE template data inserter duplication

vindl:

To avoid duplicating this file across all FSE themes I'm moving it to the FSE plugin instead and providing hooking to theme activation there. This is now viable since we are no longer hardcoding the templates in themes, but fetching them via API.

#### Testing instructions

- Follow the instructions available in https://github.com/Automattic/wp-calypso/pull/35190.


Copons:
Tested along with Automattic/wp-calypso#35190 and confirmed it works fine. 👍

Issue #1192, opened 2019-08-05
Varia: Should wide width element be wider than normal width in mobile? 


iamtakashi:
I've noticed this while I'm looking at Leven. On mobile, a wide width element is very close to the edge of the screen. I can understand the logic, and technically this is correct, but should it be? My opinion is that this technicality creates unnecessary visual noise because there is not much difference between wide and normal in small screen, and a wide element should be the same width until the viewport that a wide element can have enough difference.

<img width="349" alt="Screen Shot 2019-08-05 at 14 25 47" src="https://user-images.githubusercontent.com/908665/62469143-92a4f000-b78f-11e9-8c6b-08213bb06a0e.png">

cc @allancole

allancole:
Yeah, I’ve noticed this a bit with Coutoire and agree its its too small of a space to really matter. 

Just for clarity, are you saying you think `.alignwide` should become `.alignfull` on mobile? Or should it go to normal width on mobile?

iamtakashi:
> Just for clarity, are you saying you think .alignwide should become .alignfull on mobile? Or should it go to normal width on mobile?

Ah sorry it wasn't clear. My gut feeling is it should be the same with normal width. I think there are more cases that wide width being used with text content like the example above. If it was full width the text will bump to the edge of the screen.

allancole:
Great, yeah I think Normal width would look best too.

allancole:
’ll work out a PR shortly.

iamtakashi:
@allancole This is a different issue in my mind though.

As we discussed in #1217, we should increase the threshold for the wide breakpoint so that wide width become different from normal on desktop.

PR #1217, opened 2019-08-07
Varia: Fix Mobile `.alignwide` widths to match normal width

allancole:

- Makes `.alignwide` widths match the normal content width on mobile screens. 
- Also recompiles all Varia child themes to apply this fix. 
- Recompiling child themes also applies improvements from #1194 and #1204.  

#### Related issue(s):

#1192 

#### To do:

- [x] Improve nested calculations (ie: `calc( calc() )`): https://github.com/Automattic/themes/compare/enhancement/mobile-alignwide-width?expand=1#diff-b6d805de0184172207b2b53df97b66d7R3216
- [x] Test all child themes to make sure compiling other PRs hasn’t broken anything.
- [x] Fix issue where white space is removed from `calc()` which makes them not work in Chrome and FF.


iamtakashi:
Thanks for working on this. This is definitely better for mobile. 👍 

The tiny width difference between wide and normal can still see in every brea point above for example this is what it appears with 782px viewport:

Screen Shot 2019-08-08 at 15 33 41

I started to wonder if we should make wide width form 1024 or even 1280 and above. Similar to the responsive pattern in editor. https://cld.wthms.co/MOlSJr 

Let me know your thoughts on this.

allancole:
> The tiny width difference between wide and normal can still see in every brea point above for example this is what it appears with 782px viewport: 

> I started to wonder if we should make wide width from 1024 or even 1280 and above. Similar to the responsive pattern in editor. https://cld.wthms.co/MOlSJr

Yeah, it should match the editor behavior as closely as possible when it makes sense. I think increasing the threshold for the `.alignwide` breakpoint should be fine for the front-end. I’ll take another look there and move it up one size (from `large/tablet` to `x-large/desktop`)

iamtakashi:
Sorry for not being able to review this sooner. 

I had the impression of we're going to make wide width element behaves the same as normal until desktop or at the point where the normal width fully expanded. I'm not saying this is how it should be, just wondering what most makes sense. This is what I see in the editor.

Sep-06-2019 13-55-11

Again, I don't know if this is the best, and the normal width behaves differently between the editor and the front-end anyway.

allancole:
Ahhh, yeah I see what you mean. 

> I had the impression of we're going to make wide width element behaves the same as normal until desktop or at the point where the normal width fully expanded

Yeah, I missed that detail :-) The last commit above makes it work that way, and while it doesn’t exactly mirror the editor, I do think it’s the better solution. 

simple-wide-width-test

Check it out and let me know what you think.

iamtakashi:
That looks great. In this way, we can avoid the situation where the difference between wide and normal is just a few pixels :)

PR #1219, opened 2019-08-07
Modern Business: Set editor background color for FSE

noahtallen:
Handle background color of editor for FSE. I thought one would just add `.editor-styles-wrapper`, but it turns out that Gutenberg automagically prefixes every selector inside the editor stylesheet! So to target the background of the editor, you actually use `body`. See more info on that [here](https://github.com/WordPress/gutenberg/blob/0ad0414e7a469254ef1ad105fdcf14248a986686/docs/designers-developers/developers/themes/theme-support.md#editor-styles). 

## Testing:
0. Use this FSE PR on your local setup: https://github.com/Automattic/wp-calypso/pull/35235
1. Pull this branch and make sure you have modern business mapped to your local setup.
2. Open the page editor and the template editor.
3. The background color of each should be `#fff` as set by the theme.
4. Change the background property to `blue`. ([here](https://github.com/Automattic/themes/pull/1219/files#diff-5a116bd4bbf0de8a39434744a50c9dc9R40))
5. In `themes/modern-business` run `npm i && npm run build` to update the CSS.
6. Verify that the editor background color is now blue. This is just to make sure the selector is getting applied, since the default color in FSE is also set to white.

Resolves #35117. A task has been added to the overall theme issues to bring this support to them as well, so we don't need that issue. I'll link to this PR so that people can see what to do for it.

### DIFF
--- a/modern-business/style-editor.css
+++ b/modern-business/style-editor.css
@@ -924,6 +924,7 @@ blockquote {
 /** === Editor Frame === */
 body {
   font-weight: 300;
+  background: #fff;
 }
 
 body .wp-block[data-align="full"] {
--- a/modern-business/style-rtl.css
+++ b/modern-business/style-rtl.css
@@ -4649,17 +4649,29 @@ footer .wp-block-a8c-navigation-menu .sub-menu {
 }
 
 .entry .entry-content .wp-block-cover-image h2.has-text-align-left,
-.entry .entry-content .wp-block-cover h2.has-text-align-left {
+.entry .entry-content .wp-block-cover h2.has-text-align-left,
+.fse-enabled .site-header .wp-block-cover-image h2.has-text-align-left,
+.fse-enabled .site-header .wp-block-cover h2.has-text-align-left,
+.fse-enabled .site-footer .wp-block-cover-image h2.has-text-align-left,
+.fse-enabled .site-footer .wp-block-cover h2.has-text-align-left {
   text-align: right;
 }
 
 .entry .entry-content .wp-block-cover-image h2.has-text-align-center,
-.entry .entry-content .wp-block-cover h2.has-text-align-center {
+.entry .entry-content .wp-block-cover h2.has-text-align-center,
+.fse-enabled .site-header .wp-block-cover-image h2.has-text-align-center,
+.fse-enabled .site-header .wp-block-cover h2.has-text-align-center,
+.fse-enabled .site-footer .wp-block-cover-image h2.has-text-align-center,
+.fse-enabled .site-footer .wp-block-cover h2.has-text-align-center {
   text-align: center;
 }
 
 .entry .entry-content .wp-block-cover-image h2.has-text-align-right,
-.entry .entry-content .wp-block-cover h2.has-text-align-right {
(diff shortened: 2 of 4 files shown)


modern-business/style-editor.css
```diff
@@ -981,9 +981,15 @@ body .wp-block[data-align="full"] {
   width: 100%;
 }
 
+/** === Editor background color === */
+.editor-styles-wrapper {
+  background: #fff;
```
gwwar: This is pretty odd, inspecting in FF/Chrome this property isn't even set at all. (Not inherited from body either).

Screen Shot 2019-08-07 at 12 02 35 PM

@noahtallen have you tried a different selector, or doubling up there with an additional class?

One thing I noted was a style tag at the bottom page for related `editor-styles-wrapper` rules.

gwwar: Oh my bad, wrong prop name:

Screen Shot 2019-08-07 at 12 06 10 PM

gwwar: Screen Shot 2019-08-07 at 12 12 21 PM

Here's why the rule isn't applying. I forget which is which, but we're likely pulling in the sass as a mixin or extends.

`.editor-styles-wrapper .editor-styles-wrapper` won't apply.

noahtallen: Oh that's really odd. I kinda thought that would show up in the output CSS. Good catch!

gwwar: Ah as Noah noted, the extra class is coming from default theme support behavior https://github.com/WordPress/gutenberg/blob/0ad0414e7a469254ef1ad105fdcf14248a986686/docs/designers-developers/developers/themes/theme-support.md

glendaviesnz:
I might be missing something, but when I test it the background being set in the theme is being overridden by the:

https://github.com/Automattic/wp-calypso/blob/master/apps/full-site-editing/full-site-editing-plugin/full-site-editing/editor/style.scss#L21

ie. the blue is background is not applied unless I comment out the above line - so do we need to set the background it the theme if it is being set by the plugin?

noahtallen:
I’ll have to take a look again tomorrow then. I thought it was working! We need to be able to override that from the theme for other non-white backgrounds. I wanted to keep the color in the plugin as well just in case the theme hasn’t added it yet - if no color is set, then the background is all gray and shadowy from the outer frame.

noahtallen:
Updated instructions with this FSE PR to make the selector work properly: https://github.com/Automattic/wp-calypso/pull/35235

modern-business/style-rtl.css
```diff
@@ -4649,17 +4649,29 @@ footer .wp-block-a8c-navigation-menu .sub-menu {
 }
 
 .entry .entry-content .wp-block-cover-image h2.has-text-align-left,
-.entry .entry-content .wp-block-cover h2.has-text-align-left {
+.entry .entry-content .wp-block-cover h2.has-text-align-left,
```
noahtallen: These changes were autogenerated when I ran `npm run build`, probably just some other stuff that didn't build properly in a different PR.

noahtallen:
Fantastic!

PR #1223, opened 2019-08-08
Modern Business: Fix alignment issues of elements inside group block

### LINKED ISSUE #1137: Modern Business: Alignment issue on element inside group block
See https://gutenbergtestdata.wordpress.com/2019/07/22/group-block/

It's rough, we should fix them :)

iamtakashi:
This is a partial fix. At the moment it only fixes the issue on the front end.

I've tried the editor side, but it seems to be more tricky.

Fixes #1137

### DIFF
--- a/modern-business/sass/blocks/_blocks.scss
+++ b/modern-business/sass/blocks/_blocks.scss
@@ -1,7 +1,7 @@
 /* !Block styles */
 
 .entry .entry-content > *,
-.entry .entry-summary > *, 
+.entry .entry-summary > *,
 .fse-enabled .site-header > *,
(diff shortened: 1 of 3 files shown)


iamtakashi:
@allancole We briefly had a chat about this, and the front-end os in a good shape, but I struggled the editor CSS. Would you mind take a stab at this?

We want to refresh the business themes with Varia, but the themes being used in Signup, SPT, and FSE, we need to support them for a bit more :/

allancole:
Yup, I’ll take a stab at this @iamtakashi 👍

iamtakashi:
The theme has been retired. Closing this.

PR #1240, opened 2019-08-10
Varia: Make normal font size the same as default size

### LINKED ISSUE #1239: Varia: Normal font size is bigger than default size.
Currently, the normal size is actually bigger than the default. This is confusing, and I think it should be the same size as default size.

Screen Shot 2019-08-10 at 16 38 44

iamtakashi:
Fixes #1239


allancole:
Merging this one…

Issue #1241, opened 2019-08-10
Varia: Behaviour of alignfull element


iamtakashi:
Currently, an element with `alignfull` always tries to expand to edge to edge of the screen, this made sense until recently, but now I'm not sure anymore :)

When I try to make a little more complex layout, the behaviour of alignfull would make more sense if it expands fully _within the available area_ just like `width 100%`. For example, if you nest 5 full width group blocks that have 16px paddings, it could be like this.

<img width="1039" alt="Screen Shot 2019-08-10 at 20 26 46" src="https://user-images.githubusercontent.com/908665/62826251-258ac380-bbb0-11e9-852f-f38f228e90e7.png">

The most full width outer should still expands to edge to edge of the screen, but the nested full width elements shouldn't expand beyond the wrapper element in my opinion. This will probably simplify the CSS as well.

What's your opinion @allancole?

iamtakashi:
I've heard that a padding setting for Group being worked on now. I think it would be better yo keep the behaviour simpler would be better for the setting as well.

iamtakashi:
We're tackling this issue in #1176.

PR #1218, opened 2019-08-07
Varia: A more exhaustive approach to the solution proposed in #1184

allancole:
#### Changes proposed in this Pull Request:

- Builds on the solution presented in #1184 to be _block-agnostic_. This way all blocks that allow deep nesting can have the same responsive behavior.
- Adds nested utility selectors to deal with _any_ nested `.alignwide` and `.alignfull` blocks. 
- Moves `.wp-audio-shortcode` selector to the audio block sass file for clarity.

PR #1246, opened 2019-08-12
Varia: Refactor responsive logic to better support nested blocks with varying widths

allancole:
An even _more_ exhaustive approach to the solution proposed in #1218.

- Makes .entry-content 100% wide to let the box-model determine .alignfull widths without negative margins
- .alignwide and _.aligndefault_ widths now have a fixed width, and a responsive 100% max-width
- Improves nested block behaviors and reduces CSS load
- Updates utility classes across the theme to work with this change
- Removes .responsive-max-width utility classes from .entry-content and similar wrappers that no longer need it
- Renames responsive `@extends` and `@mixins` to match Gutenberg pre/suffixes

#### Related issue(s):

If this PR is merged, both #1184 and #1218 can be rejected in favor of this PR.

Fixes: #1241
Fixes: #1192
Fixes: #1254
Fixes: #1252

#### Screenshots

responsive-refactor-1

responsive-refactor-2


iamtakashi:
@allancole This is awesome improvement. This is definitely on the right direction. There are a few small thing I've noticed.

* Looks like core css stops the inner container of a cover block 70px short from 100%. We would want to change that with the variable.

Screen Shot 2019-08-13 at 11 40 42

* Let's not remove the padding with the negative margin on the last element.

Screen Shot 2019-08-13 at 11 52 10

* The same as above, let's keep the padding so that it's less opinionated, and handle group block like `div`

Screen Shot 2019-08-13 at 11 58 51

allancole:
Added fixes based on your comments above. Let me know if you notices anything else @iamtakashi . 

Before we merge I think we should resync with Master and update all the child themes on this PR. I’ll do that once you’ve approved these last few fixes 👍

iamtakashi:
Awesome, this works nicely, and I like the default 32px padding for Group as well.

> Before we merge I think we should resync with Master and update all the child themes on this PR. I’ll do that once you’ve approved these last few fixes 

Good idea. Hopefully, there won't be too much work for it.

allancole:
Also I think we should merge this _after_ we launch the two themes for this week. That way we can apply the fix to all child themes that are live at once.

iamtakashi:
> Also I think we should merge this after we launch the two themes for this week. That way we can apply the fix to all child themes that are live at once.

Good idea 👍

PR #1249, opened 2019-08-13
Barnsbury: A Refresh of the Friendly Business theme

allancole:

- Use Varia to recreate the Friendly Business theme in preparation for better FSE support.

#### Related issue(s):

https://github.com/Automattic/template-first-themes-private/issues/121


allancole:
@iamtakashi, I think this is ready to go :-). I’ve only spent about 5 hours on it and the homepages are nearly identical outside of some minor spacing and responsive behaviors.

**Barnsbury vs. Friendly Business:**

friendly-business-vs-barnsbury

iamtakashi:
Amazing quick turn around!

1:1 replication isn't important, and this is a good chance to improve the visual. So feel free to tweak the design to make it look nicer.

Let's improve the header pattern. I think Morden or Hever pattern is solid.

iamtakashi:
I'll have a test run and will note here what I've noticed.

iamtakashi:
- The vertical paddings for the header seems too much.
- Reduce the font size of the menu.
- Change the header layout like Morden or Hever.

Screen Shot 2019-08-14 at 18 09 04

iamtakashi:
Let's change the body text and line height to 20px/1.6 so that it sits well on the baseline.

iamtakashi:
* Post titles would look nicer if the colours are reversed.

Screen Shot 2019-08-14 at 18 15 35

* I'm not sure about the red secondary colour. Some earthy colour would be nice.

Screen Shot 2019-08-14 at 18 18 11

* The button seems a little too big?

Screen Shot 2019-08-14 at 18 19 51

* Widget title seems unnecessarily too big. The same as the leave a reply heading or smaller would be nicer.

Screen Shot 2019-08-14 at 18 20 57

allancole:
@iamtakashi, I made all the changes you mentioned above plus a few other improvements, I noticed a long the way. Should be good to go :-)

iamtakashi:
@allancole I missed this in initial review, but let's avoid the following in `_extra-child-theme.scss` that alters the style of some blocks.

```
.wp-block-separator:not(.is-style-wide):not(.is-style-dots) {
	max-width: 10rem;
}

.wp-block-cover,
.wp-block-cover-image {
	background-color: map-deep-get($config-global, "color", "foreground", "dark");
}

.entry .entry-content .wp-block-media-text {
	background-color: map-deep-get($config-global, "color", "background", "light");
} 
```

allancole:
Agreed @iamtakashi ^ 

Almost all block styles should be done with the variables system only. The `_extra-child-theme.scss` Should almost exclusively focus on non-block areas (header, footer, comments, _some_ global styles). I’ll take a note of this for the documentation.

PR #1256, opened 2019-08-14
Modern Business: FSE - fix footer alignment in front end

glendaviesnz:

* Add css to left align the footer separator and nav menu to match the editor styling

### Testing

* Build modern business and apply to FSE test environment
* Check that the the horizontal alignment of any block added to the footer matches across the page view, template edit view and the front end.
* Check that side padding is adding correctly on smaller screens to stop blocks being pushed hard against the edge

**notes:** 

* This issue occurred in both local and WP.com Simple sites. I have tested in Sandboxed simple site  - feel free to double check the fix in both also.
* Vertical alignment is still not perfect between these views, but close!

**Before**

before-footer-space

**After**

footer-space-after

#### Related issue(s):
https://github.com/Automattic/wp-calypso/issues/35389

### DIFF
--- a/modern-business/sass/blocks/_blocks.scss
+++ b/modern-business/sass/blocks/_blocks.scss
@@ -1265,7 +1265,7 @@
 	}
 }
 
-.fse-enabled .site-footer, .post-type-wp_template .site-footer {
+.fse-enabled .site-footer {
 	
 	padding: $size__spacing-unit $size__vertical-spacing-unit;
 	
--- a/modern-business/sass/blocks/_blocks.scss
+++ b/modern-business/sass/blocks/_blocks.scss
@@ -1267,3 +1267,12 @@
 		}
 	}
 }
+
+.fse-enabled .site-footer {
+	.wp-block-a8c-navigation-menu, .wp-block-separator {
+		margin: 15px 0;
+	}
+	.site-info {
+		text-align: center;
+	}
+}
--- a/modern-business/sass/blocks/_blocks.scss
+++ b/modern-business/sass/blocks/_blocks.scss
@@ -1282,9 +1282,15 @@
 		margin: $size__vertical-spacing-unit 0;
 		text-align: left;
 	}
-	.wp-block-separator, .wp-block-button {
+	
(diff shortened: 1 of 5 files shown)


modern-business/sass/blocks/_blocks.scss
```diff
@@ -1267,3 +1267,12 @@
 		}
 	}
 }
+
+.fse-enabled .site-footer {
```
noahtallen: Great call to add this class, it has been very helpful!

glendaviesnz:
> Unfortunately this isn't working for me with anything other than the nav block

Haha, you mean we can put anything in the footer :-) rats, back to the drawing board, without blinkers on this time!

vindl:
Also looking at the "after" screenshot - the `A WordPress.com Website` should be left aligned too.

glendaviesnz:
> Unfortunately this isn't working for me with anything other than the nav block

Should be better now - vertical alignment between views is not perfect,  but close. I am not 100% happy about the amount of arbitrary adjustment needed between different block types and views - doesn't scale well for bringing FSE to other things, but I suggest we go with this approach for now, but circle back and have a good think about how to handle in a more systematic way.

gwwar:
Spacing looks a bit tight on the credits vs bottom menu. Did the original theme have similar spacing?

Screen Shot 2019-08-15 at 10 39 15 AM

noahtallen:
Looking better! I noticed two things that are weird though.
1. I can't change the header alignment (always stays left aligned)
2. That button is "center aligned" but not in the center

gwwar:
On my note, it's not blocking since the original theme does this 😆 

Screen Shot 2019-08-15 at 10 43 22 AM

Let's maybe get this in and we can follow up with polish before v1?

noahtallen:
Yeah, if I disable FSE the footer looks like this lol:
Screen Shot 2019-08-15 at 10 50 27 AM
Quite a bit we aren't doing yet

gwwar:
Thanks @glendaviesnz, going to land this one for you!

PR #1257, opened 2019-08-15
FSE: Use new classnames instead of data-type selectors

noahtallen:
Affects friendly business and modern business. Also renamed one or two things to "template" from "template part".

Note that the only changes I made were in *.scss files. The .css and package lock files are autogenerated.

### Testing
1. Make sure you have both modern business and friendly business mapped to your local FSE setup
2. Make sure you have FSE active on your this site.
3. Pull this branch.
4. Activate the modern business theme.
5. Open the page editor and make sure the styles look correct. If the new selector is failing, things would be very messed up, like paragraphs in post content extending the full width of the screen.
6. Inspect the post content block and verify the style is getting applied correctly:
Screen Shot 2019-08-14 at 5 41 25 PM

7. Inspect a block nested inside the post content block and verify the style is getting applied correctly:
Screen Shot 2019-08-14 at 5 40 35 PM

8. Activate the friendly business theme. Try steps 5 through 7 with it active. (Obviously most FSE functionality doesn't work with it yet, but the post content should still show up in the right format. On a local install, FSE will also stay active if the theme is unsupported, which helps test this.)

Resolves https://github.com/Automattic/wp-calypso/issues/35246 along with https://github.com/Automattic/wp-calypso/pull/35400

### DIFF
--- a/friendly-business/package-lock.json
+++ b/friendly-business/package-lock.json
@@ -524,12 +524,12 @@
       "requires": {
         "bluebird": "3.5.3",
         "chokidar": "2.1.1",
-        "lodash": "4.17.13",
+        "lodash": "4.17.11",
         "yargs": "12.0.5"
       },
       "dependencies": {
         "lodash": {
-          "version": "4.17.13",
+          "version": "4.17.11",
           "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.13.tgz",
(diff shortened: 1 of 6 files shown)


gwwar:
@noahtallen maybe try rebasing? Overall changes look good on wpcom, but I'm noticing the publish view has the footer menu aligned center instead of aligned left.

noahtallen:
> Should be good to merge then if you can verify we're not regressing that.

I was able to verify that the footer is left aligned on the front end, page editor, and template editor once I rebased :)

PR #1258, opened 2019-08-15
Mayland: Add new theme to repo

iamtakashi:
A refresh of Calm Business.


iamtakashi:
@danieldudzic Can you take over this?

First, rebase with `master`, so it will have the latest Varia updates. And let's change the header design. This vertically stacked layout takes too much space :) Can you make the header layout full-width and like this?

Screen Shot 2019-09-01 at 12 33 55

PR #1260, opened 2019-08-15
Modern Business: add CSS custom properties

nosolosw:
This adds CSS custom properties for `font-family` using the existing font as a _variable fallback_ for when the custom property is not defined.

Note that:

- I've only modified SCSS files. I'm assuming the CSS ones are generated.
- I've left a few cases untouched because they didn't use `$font__heading` or `$font__body`, but other variables such as `$font__pre` or `$font__code`. We could add support for them anyway as they'll fall back to the previous value.
- ff3d576 adds a _browser fallback_ for those that don't support custom properties.

The theme should still work as expected with this code.

### DIFF
--- a/modern-business/sass/blocks/_blocks.scss
+++ b/modern-business/sass/blocks/_blocks.scss
@@ -315,6 +315,7 @@
 			@include button-transition;
 			border: none;
 			font-size: $font__size-sm;
+			font-family: $font__heading; // for browsers without CSS custom props support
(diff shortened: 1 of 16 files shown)


jasmussen:
Took a look just now, fonts appear to work well in the editor as far as I can see. Not really sure what more to look for, I'm guessing there are some technical details Allan and Takashi might want to look at.

nosolosw:
We're going to try a different approach in which the theme doesn't need to do any of this.

PR #1262, opened 2019-08-15
Modern Business: Refactor the Full Site Editing styles

Copons:
#### Changes proposed in this Pull Request

Requires https://github.com/Automattic/wp-calypso/pull/35387

Refactor the full Editing Styles in the Modern Business theme.

Before/after screenshots can be found in https://github.com/Automattic/wp-calypso/pull/35387

#### Testing instructions

* Build FSE with https://github.com/Automattic/wp-calypso/pull/35387.
* On an FSE site, edit a page.
* Try poking the header and footer template parts, and make sure that:
  * The Template block overlay is properly aligned to the editor frame.
  * The overlay retains the same background color and blur filter as before.
  * Selecting (clicking) a Template block does not close the sidebar (if it was open), but simply hides its "Block" tab.
  * Everything looks fine!


gwwar:
Needs a rebase. This is a pure janitorial refactor, right?

Copons:
> Needs a rebase. This is a pure janitorial refactor, right? No changes to visual styling?

That's the idea, though we should still smoke test FSE a bit, in case I've missed something, or I've been too aggressive in the refactor.

modern-business/style-editor.scss
```diff
@@ -27,9 +28,7 @@ Modern Business Editor Styles
 		$main-block-selector: ".wp-block[data-align=\"#{$alignment}\"]";
 	}
 
-	@return "
-	  #{$main-block-selector},
-	  .wp-block[data-type=\"a8c/post-content\"] #{$main-block-selector}";
+	@return '#{$main-block-selector}, .wp-block[data-type=\"a8c/post-content\"] #{$main-block-selector}';
```
noahtallen: Will need updated once https://github.com/Automattic/themes/pull/1257 is in

gwwar:
I'll test as is, but we'll need a quick ✅ after a second rebase if we're conflicting the files this easily 😆

modern-business/style-editor.scss
noahtallen: all we need to do is change `.wp-block[data-type=\"a8c/post-content\"]` to `.wp-block.post-content__block` and rebuild the CSS and the conflicts should be fine 👍

Copons: Yeah, this change was just to clear the linter, it's not related to the PR.
It's also the only conflict so it's a super easy fix! 🙂

PR #1264, opened 2019-08-15
Dalston: Add new theme to repo

allancole:
A refresh of Professional Business.

#### Related issue(s):

https://github.com/Automattic/template-first-themes-private/issues/123


allancole:
I think this is just about done @iamtakashi :-). Started working on it only 2-3 hours ago. I also went ahead and revised the header to work more like the one in Hever. Let me know what you think or if anything is missing. 

**Dalston vs. Professional Business:**

professional-business-vs-dalston

iamtakashi:
@allancole Thanks for working on this.

* Let's avoid altering the default style of Media & Text block.

```
.entry .entry-content .wp-block-media-text {

	background: #{map-deep-get($config-global, "color", "secondary", "default")};
	color: #{map-deep-get($config-global, "color", "background", "default")};
	position: relative;

	&:before {
		content: '';
		position: absolute;
		top: 10px;
		right: 10px;
		bottom: 10px;
		left: 10px;
		border: 1px solid #{map-deep-get($config-global, "color", "background", "default")};
		z-index: 1;
	}
}
```

* The masthead doesn't seem to have any bottom padding, but it's safer to have some always.

* The vertical spacing between header and the title seems a bit excessive. Let's reduce it so that the content isn't pushed down so much.

localhost_8888_wordpress__page_id=132

* The sticky tag seems too big in Chrome or Safari. The font Crimson, when it's bold(700), is rendered in webkit. Let's use semi bold(600).

* The colour of the post title should be the same as the body text, and change it to primary colour on hover.

Screen Shot 2019-08-19 at 11 05 44

* Reduce the line height for headings. Varia default works fine.

Screen Shot 2019-08-19 at 11 23 16

* The body font size is 22px. Then let's change the line-height to `1.64` so it's 36px equivalent. (32+ (8/2))

PR #1272, opened 2019-08-16
Rivington: Add new theme to repo.

allancole:
A refresh of Sophisticated Business. 

#### Related issue(s): 

https://github.com/Automattic/template-first-themes-private/issues/124


iamtakashi:
@allancole Thanks for working on this.

* This is a dark theme and we should opt-in the dark style. See more info: https://developer.wordpress.org/block-editor/developers/themes/theme-support/#dark-backgrounds

Screen Shot 2019-08-19 at 12 19 11

* Font sizes don't match between in the editor and the front of the site.

* The body font size is 18px for desktop, then let's increase the line height to `1.778` so that it sits well in the baseline grid.

* The white body text seems too stark with the dark background, can we change the all white in this themes #F1ECE4?

* Leave text decoration from links in body text. This pattern has a better accessibility rather than relying on colour.

* Because the foreground is light, the cover block appears strange. This could be fixed in Varia as well — if the background colour for the block is black, and the text is in white, it will work for most of cases.

Screen Shot 2019-08-19 at 12 33 09

* Can you make the separator much darker?

* Some contrast issue in colour settings.

Screen Shot 2019-08-19 at 12 41 34

* The sticky tag seems to big tall to me.

Screen Shot 2019-08-19 at 12 42 34

* I know the original has centred entry-title, but with this header, normal left align works fine and better. Let's remove the centre alignment.

Screen Shot 2019-08-19 at 12 50 00

iamtakashi:
@danieldudzic I've rebased this and fixed what I found in the previous review. We may want to change some style to fit with the vertical and the homepage template, it's pretty close :)

iamtakashi:
localhost_8888_wordpress__page_id=1254 (1)

Issue #1286, opened 2019-08-19
Varia: default and wide separators are identical


iamtakashi:
This is a side effect of #1246.

Now the default and wide separators are the same on the front of the site.

<img width="821" alt="Screen Shot 2019-08-19 at 17 17 39" src="https://user-images.githubusercontent.com/908665/63281742-48e6fa00-c2a5-11e9-9947-3e8870eca48a.png">

allancole:
Fixed in #1288

PR #1288, opened 2019-08-19
Varia: Fix separator styles to work with new responsive logic

allancole:

- Restrict default width of separator block
- Add separator `width` value to config
- Revise separator selectors for CSS overrides 
- Allow normal `<hr>` markup to work as expected

#### Related issue(s):

#1286


allancole:
Tested this change against all the child themes and the issue is now fixed properly. Gonna go ahead and merge.

Issue #1274, opened 2019-08-19
Varia: Provide better default background and text colours in cover block.


iamtakashi:
This surfaced during reviewing Revington that is a dark theme. https://github.com/Automattic/themes/pull/1272#issuecomment-522536303

I think it'd be better default if the background colour is black and the text colour is white. That should work almost universally, and child themes can override if it wants a unique style.

Issue #444, opened 2018-12-19
Lodestar: Project type archives not loading the rest of the projects with a grid layout


Robertght:
## Steps to replicate

1. Add a few projects and assign it to one project type.
2. Reach to the Project type archive in question and click on the "Older Posts" button.
3. The rest of the loaded projects will be loaded but won't respect the grid layout of the ones above.

## Result
The rest of the projects didn't keep the same grid layout
https://monosnap.com/file/7FoikFaolljLtIugadVJ8QRu5de4j2

## Expected
To have the same layout on load.

* User report: 8895005-hc
* Follow-up ticket: 1653607-zen
* User's site: https://projectmanifest.org/project-type/orchestra/

Same thing happened on my end. I had to increase the number of projects shown per page as workaround.

laurelfulford:
This has been committed to WP.com in r53327-wpcom-themes.

It looks like the related ZD ticket has been closed; @Robertght can you please follow up with the original reporter? Thanks!

vanemmerikbv:
Looks like it's still not working, @laurelfulford...

I applied the changed code to all 3 files, but 'filters' (or paging) still doesn't sort the desired effect.

Pretty time-consuming issue, any help will be much appreciated!

example here: [www.vanemmerikbv.nl/projecten](url) (site in Dutch...)

PR #1300, opened 2019-08-21
Varia: Update Cover Block default background-color and text color 

allancole:

- Set default Cover Block colors to `background: #000` and `color: #FFF`. 
- Remove `calc()` from Cover Block height so we get `height: 480px` instead of `height: calc(15 * 32px)`.
- Update Cover Block config file to include background and foreground color overrides for easier child-theming
- Updates the `$config-cover` array in starter child-theme to include the new colors variables. 
- Updates the `$config-cover` array for all launched Varia child themes to prevent errors when recompiling after this merges.

#### Related issue(s):

Fixes: #1274

### DIFF
--- a/brompton/sass/_config-child-theme-deep.scss
+++ b/brompton/sass/_config-child-theme-deep.scss
@@ -188,7 +188,7 @@ $config-button: (
  * Cover
  */
 $config-cover: (
-	"height": calc( 18 * #{map-deep-get($config-global, "spacing", "vertical")} ),
+	"height": #{18 * map-deep-get($config-global, "spacing", "vertical")},
 	"color": (
 		"foreground": #{map-deep-get($config-global, "color", "background", "default")},
 		"background": #{map-deep-get($config-global, "color", "foreground", "default")},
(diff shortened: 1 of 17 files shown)


iamtakashi:
@allancole Thanks for working on this. It's helpful to allow child themes to specify the colours separately. 👍 

About the default, now we have an inconsistency between the front of the site and the editor. The background is still #444, and the font colour is #f8f9f9 (Core specifies it in further down `.wp-block-cover .block-editor-block-list__block`).

allancole:
This last commit ^ should do the trick @iamtakashi 👍

iamtakashi:
The background is fixed, but the white text colour is still overridden to #f8f9f9. We need to specify the colour with a different class `.wp-block-cover .block-editor-block-list__block` don't we?

Screen Shot 2019-08-23 at 18 02 34

allancole:
Ahhh, that makes sense @iamtakashi. Fixed in this commit ^.

image

allancole:
Gonna go ahead and merge this one…

Issue #1275, opened 2019-08-19
Varia: Links in contact Info block should always be visible


iamtakashi:
Same as links on cover or media & text, links should always be visible even when the parent has background colour.

<img width="803" alt="Screen Shot 2019-08-19 at 13 45 51" src="https://user-images.githubusercontent.com/908665/63266438-d3b8fc00-c287-11e9-8c8e-96a5e3fe1d84.png">

iamtakashi:
Fixed in #1301

PR #1301, opened 2019-08-21
Varia: Make sure links are visible when parent elements have a background color.

allancole:

- Make sure links are visible when parent elements have a background color.
- Improve Sass for utility styles to optimize CSS output.

#### Related issue(s): 

#1275


iamtakashi:
Thanks for working on this, @allancole. The fix works well on the front of the site, but in the editor, links in a group that has the primary background colour are still being a chameleon :)

allancole:
Should be fixed in the above commit @iamtakashi ^.

**Before:** 

image

**After:** 

image

iamtakashi:
Thanks for fixing this. LGTM! 👍

Issue #1296, opened 2019-08-21
Varia: Latest Posts block style


iamtakashi:
Latest posts block needs some visual attention.

In Editor:
* Font size for title.
* Colour of post date when the parent has a background colour or a background image.
* Can we remove the underline from post titles? The most majority of time we don't have underlines on titles even they are links.

<img width="1153" alt="Screen Shot 2019-08-21 at 14 06 52" src="https://user-images.githubusercontent.com/908665/63435164-5a074680-c41e-11e9-84e3-01dd4538e0d5.png">

On the front of the site
* Vertical spacing.
* Colour of post date when the parent has a background colour or a background image.
* The underline on the post title.

<img width="1250" alt="Screen Shot 2019-08-21 at 14 14 21" src="https://user-images.githubusercontent.com/908665/63435200-655a7200-c41e-11e9-8a35-dceb328b9938.png">

PR #1303, opened 2019-08-21
Varia: Add support for improved latest-posts block for Gutenberg 6.3

allancole:

- Adds more styles to better support changes to the Latest Posts blog. 
- Supports both grid and list views in both the editor and the frontend. 
- Fix text colors when latest posts block is nested in a cover block or any other block with the `.has-background` utility class. 
- _Does not_ remove underline from links, for better out-of-the-box a11y support. Child themes can easily opt-out of this with one line: `a { text-decoration: underline; }. 

#### Related issue(s): 

Fixes: #1296

### DIFF
--- a/varia/sass/blocks/latest-posts/_editor.scss
+++ b/varia/sass/blocks/latest-posts/_editor.scss
@@ -1,3 +1,42 @@
 .wp-block-latest-posts {
 	padding-left: 0;
+
+	& > li > a {
+		font-family: #{map-deep-get($config-heading, "font", "family")};
+		font-size: #{map-deep-get($config-heading, "font", "size", "h4")};
+		font-weight: #{map-deep-get($config-heading, "font", "weight")};
+		line-height: #{map-deep-get($config-heading, "font", "line-height", "h4")};
+	}
+
+	&:not(.is-grid) > li {
+		/* Vertical margins logic */
+		margin-top: #{map-deep-get($config-global, "spacing", "vertical")};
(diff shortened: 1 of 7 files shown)


allancole:
##### Frontend

image

##### Editor

image

allancole:
@iamtakashi, after looking at this: https://make.wordpress.org/themes/2019/08/03/planning-for-keyboard-navigation/ I think we _can_ actually remove the underline. Just need to make sure the browsers default `focus` states for text links is still in tact. I think I’ll make a separate issue for it and address all the a11y issues in one punch.

iamtakashi:
Thanks for working on this so quickly, @allancole! 

About the underline on the title, I wasn't actually sure if we should remove it when I suggested. What you said makes sense. I'm more than fine to keep it in Varia.

> after looking at this: https://make.wordpress.org/themes/2019/08/03/planning-for-keyboard-navigation/ I think we can actually remove the underline.

Which part in the post made you said we could remove it?

allancole:
> **Text links**
> Do not remove the browsers native outline without replacing it with an accessible alternative.
> Make sure that you can see which link is focused. Focus should either incorporate a visual change that is not based on color (background change, underline, shape) or a color change where the difference between the two colors meets the WCAG 2.0 level AA contrast ratio (4.5:1).

There’s no mention of links needing to have underlines here ^. But it does require a significant visual change when focused, so I’m thinking we can remove the underline, and then show it on `:hover, :focus, :active` states instead. 

Gonna go ahead and merge this and then create a new issue for all the a11y stuff.

PR #1309, opened 2019-08-22
Full Site Editing: Add the Header and Footer template parts to the Varia front end

Copons:

As it was done for Modern Business, add the Full Site Editing template parts to the header and footer of the Varia theme.
They should be also inherited by the Varia children themes, such as Maywood, unless they overwrite the `header.php` and `footer.php` files.

#### Testing instructions

* Apply D31833-code and sandbox the API.
* Apply https://github.com/Automattic/wp-calypso/pull/35698 on an FSE site.
* On that site, change theme from Modern Business to Varia.
* Edit a page and make sure the template parts are still there in the editor.
* Check edit.php?post_type=wp_template and make sure that there are new Header and Footer template parts (although, virtually identical to the Modern Business ones).
* View a page in the front end. It should have the FSE Header and Footer instead of the original Varia template parts.

### DIFF
--- a/varia/footer.php
+++ b/varia/footer.php
@@ -16,25 +16,33 @@
 	</div><!-- #content -->
 
 	<footer id="colophon" class="site-footer responsive-max-width">
-		<?php get_template_part( 'template-parts/footer/footer', 'widgets' ); ?>
 		<?php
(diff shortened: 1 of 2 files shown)


apeatling:
Worked for me with Varia.

PR #1320, opened 2019-08-23
Varia: Add logo resizer script to customizer

### LINKED ISSUE #1287: Varia: (and possibly child themes) Logo is loading as thumbnail, scales poorly
## Steps to replicate

1. Load up Varia or one of the child themes like Redhill
2. Add a site logo over ~300px
3. Inspect the logo, it is loading as the smallest possible size, usually around 150px

## Result
If you add CSS to enlarge the image, it will scale poorly and become distorted. Hard to implement larger logos as they are being loaded as very small images.

image

1. Image shows as blurry if scaled up with CSS
2. Width is set to 250px manually
3. But the intrinsic width is 99px

## Expected
Ideally, I could load an image at a median size, ~500px, so I can scale it accordingly.

* User report: p1566230277047000-slack-livechat

allancole:

- Add logo resizer script from Radcliffe 2 to customizer
- Increase logo size restriction for retina screens.

This change should automatically get adopted by all child-themes after merging. No need to recompile anything. Example:

varia-logo-resizer

#### Related issue(s):

Fixes: #1287 

#### To do:

Test logo resizer with all child themes and make sure the header layout can support the resizing. I’m sure some themes may not be able to use this feature in which case, we’ll need to sort out a way to opt-out of the logo resizer script. 

- [x] Alves
- [x] Balasana
- [x] Barnsbury
- [x] Brompton
- [x] Coutoire
- [x] Dalston
- [x] Exford
- [x] Hever
- [x] Leven
- [x] Mayland
- [x] Maywood
- [x] Morden
- [x] Redhill
- [x] Rivington
- [x] Rockfield
- [x] Shawburn
- [x] Stow
- [x] Stratford


allancole:
When we eventually move to FSE, this PR will be obsolete which makes this pretty low priority. I’m gonna leave this PR open just in case something changes or if future, pre-FSE child themes want to borrow the code to add a flexible Logo script.

scruffian:
This looks good. My only concern is if we have Varia sites with a fixed navigation in the header. If we do then the fact that logo changes height will potentially mess these up. Can you think of any examples?

allancole:
So the one quirky thing about this solution is that there’s an odd discrepancy between what’s calculated by the slider versus what actually gets saved as a pixel value. Because of it, if you have a larger sized logo and you try to resize it with this tool, it will give you inconsistent results where the logo size changes after you publish and refresh the option. Here’s a screen capture of the issue happening: https://cloudup.com/cjIZmcuIVCr (notice how the logo gets smaller after I publish and refresh the customizer). 

> My only concern is if we have Varia sites with a fixed navigation in the header. If we do then the fact that logo changes height will potentially mess these up. Can you think of any examples?

Coutoire is the first one that comes to mind that has a fixed header, but there may be 1 or 2 others. It doesn’t handle it well:

PR #1351, opened 2019-08-29
FSE: add fse theme support flag

glendaviesnz:

* Add add_theme_support( 'full-site-editing') to each of the FSE themes so this can be used instead of a whitelist in the FSE plugin

Testing instructions at https://github.com/Automattic/wp-calypso/pull/35856

### DIFF
--- a/varia/functions.php
+++ b/varia/functions.php
@@ -194,7 +194,6 @@ if ( ! function_exists( 'varia_setup' ) ) :
 
 		// Add support for responsive embedded content.
 		add_theme_support( 'responsive-embeds' );
-		
 	}
 endif;
 add_action( 'after_setup_theme', 'varia_setup' );
--- a/varia/functions.php
+++ b/varia/functions.php
@@ -194,6 +194,10 @@ if ( ! function_exists( 'varia_setup' ) ) :
 
 		// Add support for responsive embedded content.
 		add_theme_support( 'responsive-embeds' );
+
+		// Enable Full Site Editing
+		add_theme_support( 'full-site-editing');
+		
 	}
 endif;
 add_action( 'after_setup_theme', 'varia_setup' );
--- a/varia/functions.php
+++ b/varia/functions.php
@@ -194,9 +194,6 @@ if ( ! function_exists( 'varia_setup' ) ) :
 
 		// Add support for responsive embedded content.
 		add_theme_support( 'responsive-embeds' );
-
-		// Enable Full Site Editing
-		add_theme_support( 'full-site-editing');
 		
 	}
 endif;
--- a/maywood/functions.php
+++ b/maywood/functions.php
@@ -99,6 +99,9 @@ if ( ! function_exists( 'maywood_setup' ) ) :
 				),
 			)
 		);
+
+		// Enable Full Site Editing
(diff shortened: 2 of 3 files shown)


varia/functions.php
```diff
@@ -194,6 +194,10 @@ function varia_setup() {
 
 		// Add support for responsive embedded content.
 		add_theme_support( 'responsive-embeds' );
+
+		// Enable Full Site Editing
+		add_theme_support( 'full-site-editing');
```
gwwar: @iamtakashi, @kwight, @mattwiebe I think this will enable for all child themes too? Should we enable here, then `remove_theme_support` on other child themes that we aren't ready yet for?

Copons: I think that's the case, and I'm in favour of removing this - unless we want to have explicit FSE support to Varia itself too, but then we'd need to remove it manually in all yet-unsupported children.

mattwiebe: The `functions.php` file is loaded from both the child theme (first) and the parent theme (second). Sometimes, the parent theme will define certain function like this so that the child theme can define it instead.

```php
if ( function_exists( 'varia_setup' ) ) : 
  function varia_setup() {/*etc*/}
endif;
```

That is not happening in this case. The child theme support declarations can be dropped.

Similarly, any child theme that does not declare its own `header.php` or `footer.php` will need no extra work to support FSE for those, as the parent theme's file will be loaded in those cases.

Copons: @mattwiebe I might be misunderstanding your comment, but I think the issue here is the other way around though.

If we wanted to _immediately_ add FSE support to Varia and all its children, then yes, we can just `add_theme_support( 'full-site-editing')` in Varia and let the children inherit it.

But we right now only want to add FSE to Maywood.
So I'd argue that we want to only add `add_theme_support( 'full-site-editing')` to Maywood, and remove it from Varia to prevent unintended FSE support in all the other Varia children.

gwwar: Let's see, Varia does show up as a theme folks can install, but it also looks like this:

Screen Shot 2019-08-30 at 9 21 50 AM

I'd agree with @Copons to line until we're comfortable whitelisting more themes. If there's a technical reason we need it, we can re-evaluate.

mattwiebe: Ah, I failed to understand! 

I thought that all child themes of Varia were supposed to have FSE support. If it's only Maywood, then it should be removed from Varia for now, until such a time as "all Varia child themes are FSE themes," at which point any child theme support declarations can be dropped.

Alternately, we could just add something like this to any Varia child theme we want to not support:

```php
function remove_fse_support() {
  remove_theme_support( 'full-site-editing' );
}
add_action( 'after_theme_setup', 'remove_fse_support', 11 ); // runs after `varia_setup`
```

It'll all depend which makes the most sense. I would imagine removing the declaration from Varia is the best way to go right now.

gwwar: > I thought that all child themes of Varia were supposed to have FSE support. 

Eventually. Not for the 10%

gwwar:
Going to land this with only Maywood and Modern Business. We still need to sync the themes before we land https://github.com/Automattic/wp-calypso/pull/35856

PR #1369, opened 2019-08-30
Varia: Overwrite Jetpack's default styles for the Cookies & Consents Banner

thomasguillot:

The cookie banner has some default styles which are OK but how about we make it look like the theme?

__Before:__

banner-no-style

__After:__

banner-bottom-light

banner-top-dark


varia/sass/vendors/_jetpack.scss
```diff
@@ -34,3 +34,75 @@
 		margin-bottom: 0;
 	}
 }
+
+/**
+ * Cookies & Consents Banner
+ */
+body {
```
thomasguillot: Note: We need to use `body` to make sure we actually overwrite Jetpack's styles.

PR #1371, opened 2019-08-30
Full Site Editing / Varia, Maywood: Adjust the Content block vertical margins

Copons:

Adjust the Content block vertical margin in the editor to look as close as possible to the front end.

Slice

#### Notes:

- Varia contains the defaults for all the Varia-based themes.

- The bottom margin is not exactly the same as the front end.
This is because in the front end there is the "Edit" link in that space. I think it's not worthy to go pixel perfect in this case, and so I've just picked the multiple of 32 (for the sake of it) that looked as close as possible in Maywood.

- I've added a z-index to the Content block.
This is needed because the page title frame (or, if the title was hidden, any other block's toolbar etc.) could appear _behind_ the header Template block.

- As you might notice from the screenshot, the header spacing in Maywood is not as close as it should have been.
I've decided instead to adjust the Content top margin just enough so that the page title's permalink box is aligned with the header's overlay border.
This is not just to satisfy my pedantry, but also because while the permalink box has a background color which would cover the header content and overlay, this is not the case for the actual title textarea.
Adding a background color to the title should be avoided, because then we would need to maintain it for every theme that might have a non-white background.

#### Related issue(s):

This PR is tracked by https://github.com/Automattic/wp-calypso/issues/35870

#### Testing instructions:

- Apply this PR on an FSE site with both Varia and Maywood. If needed build Varia first and then Maywood, just in case.
- Try this with both Varia and Maywood.
- Compare the front end to the page editor, and make sure the Content block vertical margins look good (or at least, look similar to the screenshot!).

### DIFF
--- a/varia/sass/full-site-editing/_editor.scss
+++ b/varia/sass/full-site-editing/_editor.scss
@@ -45,6 +45,12 @@
 	}
 }
 
+.post-content__block {
+	margin-bottom: 160px;
+	margin-top: 36px;
+	z-index: 20; // Minimum to appear above the Template block overlay
+}
+
 .site-footer {
 	@include media(tablet) {
 		display: block;
--- a/varia/sass/full-site-editing/_editor.scss
+++ b/varia/sass/full-site-editing/_editor.scss
@@ -48,7 +48,11 @@
 .post-content__block {
 	margin-bottom: 160px;
 	margin-top: 36px;
-	z-index: 20; // Minimum to appear above the Template block overlay
+
+	// Minimum z-index to appear above the Template block overlay.
+	// @see https://github.com/WordPress/gutenberg/blob/f198997e2c8e377423beb230ce5283914076d396/packages/block-editor/src/components/block-list/style.scss#L495-L496
+	// @see https://github.com/WordPress/gutenberg/blob/f198997e2c8e377423beb230ce5283914076d396/assets/stylesheets/_z-index.scss#L8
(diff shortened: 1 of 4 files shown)


varia/sass/full-site-editing/_editor.scss
```diff
@@ -41,6 +41,12 @@
 	}
 }
 
+.post-content__block {
+	margin-bottom: 160px;
+	margin-top: 36px;
+	z-index: 20; // Minimum to appear above the Template block overlay
```
gwwar: 🙈 Should we add a sass variable then over the magic number?

gwwar:
Simple site testing is blocked by https://github.com/Automattic/wp-calypso/issues/35913

varia/sass/full-site-editing/_editor.scss
Copons: I don't think it will be of such big help, though...
That magic number is to surpass the [Gutenberg's z-index value for wide- and full-aligned blocks](https://github.com/WordPress/gutenberg/blob/f198997e2c8e377423beb230ce5283914076d396/packages/block-editor/src/components/block-list/style.scss#L496) ([here the value](https://github.com/WordPress/gutenberg/blob/f198997e2c8e377423beb230ce5283914076d396/assets/stylesheets/_z-index.scss#L8)), like the Template.

I'll go ahead and clarify it better in the comment though! 👍

PR #1378, opened 2019-09-02
Mayland: Recompile to Varia 1.1.1

danieldudzic:
#1258


danieldudzic:
#1258

iamtakashi:
Hey thanks for working on it, but something isn't right.

* Full-width blocks aren't full width
* Got compile error with `npm run watch` (after `npm install` of course)

iamtakashi:
I've updated my PR, so it's easy now to keep working in #1258.

PR #1379, opened 2019-09-02
Full Site Editing / Varia, Maywood: Let the Editor Align the Footer Blocks

Copons:

Change the default footer layout from `flex` to `block` in order to let the editor properly position and align all blocks inside the Footer template.

As a byproduct, the site info (e.g. "MAYWOOD, A WordPress.com Website.") is not inlined with the footer navigation menu as shown in the [Maywood demo](https://maywooddemo.wordpress.com/).
This is unavoidable, unless we want to add the site info as FSE block, and let the user move it as they see fit.

| | Editor | Front End |
| - | - | - |
| Before | Screenshot 2019-09-02 at 15 09 09 | Screenshot 2019-09-02 at 15 09 33 |
| After | Slice | localhost_short-page_ |

#### Testing instructions

* On an FSE site apply this PR and build both Varia and Maywood.
* Activate Maywood if needed.
* Try adding some explicitly aligned blocks (e.g. Image), and change its alignment a few times.
* Check that each change looks as expected in both editor and front end (e.g. a full block should "touch" the window borders; a wide block should be wider than "normal" blocks, if there is enough horizontal space; etc.).

#### Related issue(s):

Tracked in https://github.com/Automattic/wp-calypso/issues/35870

### DIFF
--- a/maywood/sass/_full-site-editing.scss
+++ b/maywood/sass/_full-site-editing.scss
@@ -13,4 +13,20 @@
 			}
 		}
 	}
+
+	.site-footer {
+		.wp-block-a8c-navigation-menu {
+			@include media(tablet) {
+				margin-bottom: #{ 0.666 * map-deep-get($config-global, "spacing", "vertical") };
+				margin-top: #{ 0.666 * map-deep-get($config-global, "spacing", "vertical") };
+			}
+		}
+	}
+
+	#colophon {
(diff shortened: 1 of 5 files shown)


Copons:
Self-approving this as it only affects FSE-enabled sites on Varia and Maywood.

PR #1385, opened 2019-09-03
Varia: image attachment page enhancement. 

### LINKED ISSUE #1376: Varia: Attachment page enhancement
I'd like to see some enhancement in Varia's attachment page.

* Full-size image (Currently the width is restricted to the normal width)
* Image navigation `previous_image_link()`, `next_image_link()`

iamtakashi:
Fixes #1376 

This PR adds image navigations on `image.php` and changes `entry-attachment` full-width.

**Before:**
localhost_8888_wordpress__attachment_id=1570 (1)

**After:**
localhost_8888_wordpress__attachment_id=1570


iamtakashi:
@allancole Thanks for the review. BTW, this PR only changes `image.php` and other attachments such as videos are displayed in `single.php` where the “Published in” text bug needs to be fixed. I can fix it with this PR if that's ok.

Since videos will be limited to the content width, I think it's better to be contained with the content width.

In my opinion, the next/previous links don't seem necessary for videos and audios.

allancole:
> I can fix it with this PR if that's ok.
Yup, I think its a close enough issue to just fix it in this PR. 

> Since videos will be limited to the content width, I think it's better to be contained with the content width.
Agreed. 

> In my opinion, the next/previous links don't seem necessary for videos and audios.
Agreed. The links imply a “Video Gallery” which isn’t a thing so it’s probably best to just remove them.

iamtakashi:
Fixed the published in text in single.php.

Screen Shot 2019-09-26 at 00 14 59

PR #1400, opened 2019-09-05
Varia: Update Cover Block to work with flexible height setting in Gutenberg 6.4

allancole:
Gutenberg 6.4 introduces a new customizable `min-height` option for the Cover Block that conflicts with the Cover Block’s `min-height` variable in Varia. To fix this, we need to revise the way the min-height works to handle 2 conditions: 

1) The heights of “Legacy” Cover Blocks that already exist should remain unchanged.
2) New Cover Blocks should respect the new `min-height` option.

- Adapts the `min-height` CSS in Varia for the Cover Block to only apply to “legacy” Cover blocks. 
- Prevent the theme’s from conflicting with the new cover height option in the editor and the frontend.
- Sets the default `min-height` CSS to `430px` to match the default in Gutenberg.
- Does not fix this semi-related issue: https://github.com/WordPress/gutenberg/issues/17339

Going forward, Varia child-theme’s _should not_ change the `min-height` variable. Instead we should set the height in our templates directly using the new custom Cover height feature in the editor.

#### Related issue(s):

- Discovered by @danieldudzic & @iamtakashi

### DIFF
--- a/varia/sass/blocks/cover/_editor.scss
+++ b/varia/sass/blocks/cover/_editor.scss
@@ -3,7 +3,10 @@
 
 	background-color: #{map-deep-get($config-cover, "color", "background")};
 	color: #{map-deep-get($config-cover, "color", "foreground")};
-	min-height: #{map-deep-get($config-cover, "height")};
+
+	.block-library-cover__resize-container:not([style*="min-height"]) & {
+		min-height: #{map-deep-get($config-cover, "height")};
+	}
 
 	.wp-block-cover__inner-container,
 	.wp-block-cover-image-text,
--- a/varia/sass/blocks/cover/_editor.scss
+++ b/varia/sass/blocks/cover/_editor.scss
@@ -4,10 +4,6 @@
 	background-color: #{map-deep-get($config-cover, "color", "background")};
 	color: #{map-deep-get($config-cover, "color", "foreground")};
 
-	.block-library-cover__resize-container:not([style*="min-height"]) & {
-		min-height: #{map-deep-get($config-cover, "height")};
-	}
-
 	.wp-block-cover__inner-container,
 	.wp-block-cover-image-text,
 	.wp-block-cover-text,
--- a/varia/sass/blocks/cover/_style.scss
+++ b/varia/sass/blocks/cover/_style.scss
@@ -2,8 +2,11 @@
 .wp-block-cover-image {
 
 	background-color: #{map-deep-get($config-cover, "color", "background")};
-	min-height: #{map-deep-get($config-cover, "height")};
 	margin: inherit;
+
+	&:not([style*="min-height"]) {
+		min-height: #{map-deep-get($config-cover, "height")};
+	}
 
 	.wp-block-cover__inner-container,
 	.wp-block-cover-image-text,
--- a/varia/sass/blocks/cover/_style.scss
+++ b/varia/sass/blocks/cover/_style.scss
@@ -4,10 +4,6 @@
 	background-color: #{map-deep-get($config-cover, "color", "background")};
 	margin: inherit;
 
-	&:not([style*="min-height"]) {
-		min-height: #{map-deep-get($config-cover, "height")};
-	}
-
 	.wp-block-cover__inner-container,
 	.wp-block-cover-image-text,
 	.wp-block-cover-text {
--- a/varia/style-rtl.css
+++ b/varia/style-rtl.css
@@ -1179,12 +1179,16 @@ input.has-focus[type="submit"],
 .wp-block-cover,
 .wp-block-cover-image {
 	background-color: black;
-	min-height: 480px;
 	margin: inherit;
 	/* Treating H2 separately to account for legacy /core styles */
 	/**
 	 * Block Options
 	 */
+}
+
+.wp-block-cover:not([style*="min-height"]),
+.wp-block-cover-image:not([style*="min-height"]) {
+	min-height: 480px;
 }
 
 .wp-block-cover .wp-block-cover__inner-container,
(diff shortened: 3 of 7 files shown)


varia/sass/blocks/cover/_style.scss
```diff
@@ -2,9 +2,12 @@
 .wp-block-cover-image {
 
 	background-color: #{map-deep-get($config-cover, "color", "background")};
-	min-height: #{map-deep-get($config-cover, "height")};
 	margin: inherit;
 
+	&:not([style*="min-height"]) {
```
iamtakashi: Is this necessary? If there is an inline style setting a minimum height the theme CSS will be overridden anyway?

varia/sass/blocks/cover/_editor.scss
```diff
@@ -3,7 +3,10 @@
 
 	background-color: #{map-deep-get($config-cover, "color", "background")};
 	color: #{map-deep-get($config-cover, "color", "foreground")};
-	min-height: #{map-deep-get($config-cover, "height")};
+
+	.block-library-cover__resize-container:not([style*="min-height"]) & {
```
iamtakashi: What is this intending? Is there any case where `.block-library-cover__resize-container` does not have the inline style?

iamtakashi:
Thanks for working on this @allancole!

* Newly created cover block works as expected without media or background colour sticking outside of the block. 👍 
* Already published cover block on a site stays as it was in 6.3 👍

But I'm struggling to understand the benefit of the two added selectors 🤔I'll leave inline comments there.

> Going forward, Varia child-theme’s should not change the min-height variable. Instead we should set the height in our templates directly using the new custom Cover height feature in the editor.

I agree. Should we remove the config entirely from Varia? We'll need to move the minimum height css to `_extra-child-theme.scss` for the theme drastically changed the height though.

varia/sass/blocks/cover/_editor.scss
allancole: Ahh, after looking again, it seems this selector is probably not necessary. It was intended to only catch blocks that were made before 6.4, but you’re right that it doesn’t actually do that. I do think its necessary to prevent disruption for Cover Blocks that were made before 6.4. For those, we want the theme-specific height variable to still work.

varia/sass/blocks/cover/_style.scss
allancole: Yeah, I think this one is necessary so that Cover Blocks from before 6.4 fallback to the default cover height variable in child themes. 

Older Cover Blocks are not automatically converted to use the new inline `min-height` (I dont think) so we need something to make sure that the height remains unchanged without it. Once a customer converts the cover block to 6.4 or updates the height setting in the editor, than this selector gets ignored entirely.

varia/sass/blocks/cover/_editor.scss
iamtakashi: Yes, I guessed that, but I was confused because it looks like all cover blocks regardless when it was made seem to have the `min-height: 430px` in the editor after updating 6.4.

varia/sass/blocks/cover/_style.scss
iamtakashi: You're right about the min-height won't automatically be added to cover blocks on the front-end. It will only be added when you change the height from the default in the editor. If that's what we both seeing, I believe we don't need it :) Because where we currently have the min-height will be the default on the front-end for both old and new cover blocks. And when you change the height the inline style takes over.

Just to be clear, what I'm suggesting doesn't make anything better :) My point is the following two CSS will be exactly the same, and the changes made in this file doesn't change anything.

```
.wp-block-cover,
.wp-block-cover-image {
	min-height: #{map-deep-get($config-cover, "height")}; 
}
```

```
.wp-block-cover:not([style*="min-height"]),
.wp-block-cover-image:not([style*="min-height"]) {
	min-height: #{map-deep-get($config-cover, "height")}; 
}
```

allancole:
Hey @iamtakashi,

> https://github.com/Automattic/themes/pull/1400#discussion_r321921748
>> all cover blocks regardless when it was made seem to have the min-height: 430px in the editor after updating 6.4.

Ahh, I didn’t catch that but I did notice that Cover Blocks with a background-color only (no bg image) do not support a custom height. See here: 

**Editor** | **Frontend**
------------ | -------------
No option to change height when using solid color as a background. `min-height: 430px` is added _inline_ by Gutenberg. | The new selector in question is accepted when using a solid color as a background since Gutenberg only adds inline styles for a background image. `min-height: 430px` is added by Varia and since theres no inline styles here, it overrides the `430px` that comes from the Gutenberg blocks stylesheet. 
image | image

So it seems that no matter what, the default Cover Block should always default to `min-height: 430px` in all cases after 6.4 is released. We should not set a Varia-specific height in CSS _at all_ after 6.4 so that the new custom height option can take priority. I’ll go ahead and remove those two pre-6.4 fallback selectors since they no effect anyway.

Additionally, I imagine Cover Blocks with only a background-color will eventually support a custom height. So removing any theme-specific Cover Block height styles now will prevent us from having to fix this again later. 

> Should we remove the config entirely from Varia? We'll need to move the minimum height css to _extra-child-theme.scss for the theme drastically changed the height though.

Yes, we should remove the min-height from the cover block config in and add it to child themes when it makes sense.

varia/style-rtl.css
```diff
@@ -1179,14 +1179,18 @@ input.has-focus[type="submit"],
...
 	background-color: black;
-	min-height: 480px;
 	margin: inherit;
 	/* Treating H2 separately to account for legacy /core styles */
 	/**
 	 * Block Options
 	 */
 }
 
+.wp-block-cover:not([style*="min-height"]),
```
iamtakashi: Leftover?

allancole:
Closing this since the Gutenberg 6.5 update fixes the issue.

PR #1406, opened 2019-09-06
Full Site Editing / Varia, Maywood: Fix incorrect content positioning when the homepage title is hidden

Copons:

On WPCOM, Varia adds an option, which defaults to true, to hide the page title of the homepage.
I didn't know about it, and my tests on a local install of course didn't have this option.

* Add the `.hide-homepage-title` class to the block editor body as well.
* Enqueue a new stylesheet on the block editor that hides the homepage title and reposition the Post Content block.
* Restore the header's bottom padding on the front end's homepage when the page title is hidden.

| | Before | After |
| - | - | - |
| Editor | Screenshot 2019-09-06 at 16 05 42 | Screenshot 2019-09-06 at 15 45 02 |
| Front End | Screenshot 2019-09-06 at 16 05 24 | Screenshot 2019-09-06 at 15 44 43 |

#### Notes:

I haven't found a better solution for this than to add a new stylesheet to the block editor on WPCOM.
This is because I believe there is no way in PHP to add CSS classes to the `editor-styles-wrapper` component, which is automagically considered the `style-editor.css` body.
So, the approach here is to add `.hide-homepage-title` to the actual editor page body, and manually enqueue the stylesheet that repositions title and content in FSE.
This manually enqueued stylesheet is not automagically parsed by WP, and so it is able to target the page body instead of just `.editor-styles-wrapper`.

Let me know if you think there's a better way that:
- Doesn't entail modifying to the plugin: theme options shouldn't be its concern!
- Adding unnecessary JS to the theme (e.g. to use the `editor.BlockListBlock` filter).

#### Testing instructions:

* Sandbox an FSE site using Maywood.
* Copy the files modified by this PR into the sandbox `wp-content/themes/pub/`.
* Make sure the "Hide homepage title" option (in the customizer, accessible at `/customize/{ siteFragment }` or `/wp-admin/customize.php`, homepage settings) is enabled.
* Open the front end, and make sure the page title is hidden and there is enough space between the header and the content.
* Edit the front end page, and make sure the title is hidden, and there is a reasonable space between header and content block.
* Repeat the test on a different page, and make sure the title is visible, and the spaces look fine.
* Disable the "Hide homepage title" option.
* Repeat the test: this time the front page should look exactly the same as any other page.

#### Related issue(s):

Fixes https://github.com/Automattic/wp-calypso/issues/36021

### DIFF
--- a/varia/inc/wpcom.php
+++ b/varia/inc/wpcom.php
@@ -105,3 +105,21 @@ function varia_wpcom_body_classes( $classes ) {
 	return $classes;
 }
 add_filter( 'body_class', 'varia_wpcom_body_classes' );
+
+/**
+ * Adds custom classes to the admin body classes.
+ *
+ * @param string $classes Classes for the body element.
+ * @return string
+ */
+function varia_wpcom_admin_body_classes( $classes ) {
+	$is_block_editor_screen = ( function_exists( 'get_current_screen' ) && get_current_screen() && get_current_screen()->is_block_editor() );
+	$hide = get_theme_mod( 'hide_front_page_title', false );
+
+	if ( $is_block_editor_screen && true === $hide ) {
+		$classes .= ' hide-homepage-title';
+	}
+
+	return $classes;
+}
+add_filter( 'admin_body_class', 'varia_wpcom_admin_body_classes' );
--- a/varia/inc/wpcom.php
+++ b/varia/inc/wpcom.php
@@ -123,3 +123,11 @@ function varia_wpcom_admin_body_classes( $classes ) {
 	return $classes;
 }
 add_filter( 'admin_body_class', 'varia_wpcom_admin_body_classes' );
+
+/**
+ * Enqueue our WP.com styles for the block editor.
+ */
+function varia_wpcom_editor_scripts() {
+	wp_enqueue_style( 'varia-wpcom-editor-style', get_template_directory_uri() . '/inc/style-editor-wpcom.css', array(), '20190906' );
+}
+add_action( 'enqueue_block_editor_assets', 'varia_wpcom_editor_scripts' );
--- a/varia/inc/wpcom.php
+++ b/varia/inc/wpcom.php
@@ -113,10 +113,12 @@ add_filter( 'body_class', 'varia_wpcom_body_classes' );
  * @return string
  */
 function varia_wpcom_admin_body_classes( $classes ) {
+	global $post;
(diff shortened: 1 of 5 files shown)


varia/inc/wpcom.php
```diff
@@ -105,3 +105,31 @@ function varia_wpcom_body_classes( $classes ) {
...
 }
 add_filter( 'body_class', 'varia_wpcom_body_classes' );
+
+/**
+ * Adds custom classes to the admin body classes.
+ *
+ * @param string $classes Classes for the body element.
+ * @return string
+ */
+function varia_wpcom_admin_body_classes( $classes ) {
```
gwwar: Does this also apply to Atomic sites?

gwwar:
> Varia adds an option, which defaults to true

@Copons could you point out the code line where this happens? 

I'm flexible on options, but I'd prefer a solution that's straightforward for Simple/Atomic to implement and maintain. If we need core contributions to make this cleaner that's an option too, though we might need a more immediate fix while we wait for the plugin release.

> Doesn't entail modifying to the plugin: theme options shouldn't be its concern!

I could see the plugin needing to know if a page title is hidden or not. We'd have to provide alternative means of editing/updating titles, creating new pages and modifying permalinks if not visible.

Copons:
@gwwar I guess "option" is kind of a loaded word in WP. Probably "setting" is better? 🤔 

Either way, Varia uses the [Theme Customization API](https://codex.wordpress.org/Theme_Customization_API#Part_1:_Defining_Settings.2C_Controls.2C_Etc.) to add a new _setting_ to the Customizer, that lets the user choose if showing or hiding the homepage title:

https://github.com/Automattic/themes/blob/3fa0230721d0d8d389845856f0ea9c1f7abb5c43/varia/inc/wpcom.php#L31-L51

Then, if this `hide_front_page_title` setting is on, it [adds a CSS class to the front end body](https://github.com/Automattic/themes/blob/3fa0230721d0d8d389845856f0ea9c1f7abb5c43/varia/inc/wpcom.php#L91-L97) (and with this PR, to the block editor body as well).

As you may have noticed, all this stuff happens in the `inc/wpcom.php` file.
As pointed out by the [file docs](https://github.com/Automattic/themes/blob/3fa0230721d0d8d389845856f0ea9c1f7abb5c43/varia/inc/wpcom.php#L5), it's automagically imported by `wp-content/mu-plugins/wpcom-theme-compat.php` on WPCOM (first function of that file).

I haven't tested on Atomic, to be honest, but the `wpcom.php` file does load on one of my Atomic test sites (I can see the "hide title" setting in the Customizer), and I don't have any reasons to think that this PR wouldn't automatically work as well.

PR #1418, opened 2019-09-24
Varia, Maywood: Fix FSE blocks style when inside a Cover block

Copons:

Themes apply some styles overrides to blocks inside Cover, for example changing the color and adding a text shadow.

* Make sure the FSE blocks (Site Title, Site Description, and Navigation Menu) look consistently when inserted inside and outside a Cover block.
* When inside a Cover block, change the color of Site Title and Site Description blocks to white, and add a text shadow. Remove the link underline from Site Title and Navigation Menu.

| | Before | After |
| - | - | - |
| Editor | Screenshot 2019-09-24 at 11 15 10 | Screenshot 2019-09-24 at 11 44 32 |
| Front End | Screenshot 2019-09-24 at 11 15 21 | Screenshot 2019-09-24 at 11 44 40 |

Note: I gave to the Site Title (which is rendered as a `h1`) a slightly smaller shadow compared to how it is in the front end. This is because in the editor it would be cut off by the `textarea` overflow.
I haven't found a good way to work around it, and so I just reduced the shadow blur radius so that the shadow is still visible, but the cut-off is not as annoying (you can see how it would look with the larger blur radius value in the screenshot I've used for Editor/After).

#### Testing instructions

* Update Gutenberg to 6.5 (or to master) on a FSE test site.
* Apply this PR on Varia and Maywood.
* Edit the header template: add a Cover block and move the FSE blocks inside it.
* Compare how the header looks in the header editor, in the page editor, and in the front end.
* Make sure there are no visual regressions to the FSE blocks when rendered _outside_ the Cover block.

#### Related issue(s):

Fixes https://github.com/Automattic/wp-calypso/issues/35429

### DIFF
--- a/maywood/sass/_extra-child-theme.scss
+++ b/maywood/sass/_extra-child-theme.scss
@@ -219,12 +219,12 @@ a {
 .wp-block-cover,
 .wp-block-cover-image {
 	p {
-		text-shadow: 0 0 6px #000;
+		text-shadow: 0 0 6px map-deep-get($config-global, "color", "black");
 	}
 
 	h1,h2,h3,h4,h5,h6 {
 		font-weight: 300;
-		text-shadow: 0 0 10px #000;
+		text-shadow: 0 0 10px map-deep-get($config-global, "color", "black");
 	}
 
 	@include media(desktop) {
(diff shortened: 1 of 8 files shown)


Copons:
@noahtallen Yeah, I've noticed the same color clash, but there's nothing really we can do there, I guess?
Say we changed the navigation menu hover to like white, it would end up having the same problem if the user chose a white cover overlay.

We could "conditionally" adjust the text color according to the cover's `.has-background-*` class.
But we can do it _if_ we recognize an actual need, and with a design input.

PR #1422, opened 2019-09-25
Add Global Styles support

nosolosw:
This PR:

* migrates Varia and Maywood to use CSS Custom Properties.
* declares support for Global Styles in the Maywood theme.

Related:

* D32062-code
* https://github.com/Automattic/global-styles (the plugin)

Test:

- With and without GlobalStyles.
- Screen and print styles.

### DIFF
--- a/maywood/sass/style-child-theme.scss
+++ b/maywood/sass/style-child-theme.scss
@@ -10,7 +10,7 @@ License: GNU General Public License v2 or later
 License URI: LICENSE
 Template: varia
 Text Domain: maywood
-Tags: one-column, flexible-header, accessibility-ready, custom-colors, custom-menu, custom-logo, editor-style, featured-images, footer-widgets, rtl-language-support, sticky-post, threaded-comments, translation-ready, auto-loading-homepage, full-site-editing
+Tags: one-column, flexible-header, accessibility-ready, custom-colors, custom-menu, custom-logo, editor-style, featured-images, footer-widgets, rtl-language-support, sticky-post, threaded-comments, translation-ready, auto-loading-homepage, full-site-editing, global-styles
 
 This theme, like WordPress, is licensed under the GPL.
 Use it to make something cool, have fun, and share what you've learned with others.
--- a/maywood/sass/style-child-theme.scss
+++ b/maywood/sass/style-child-theme.scss
@@ -5,7 +5,7 @@ Author: Automattic
 Author URI: https://automattic.com/
 Description: Maywood is a refined theme designed for restaurants and food-related businesses seeking a modern look.
 Requires at least: WordPress 4.9.6
-Version: 1.1.4
+Version: 1.2.4
 License: GNU General Public License v2 or later
 License URI: LICENSE
 Template: varia
--- a/maywood/sass/style-child-theme.scss
+++ b/maywood/sass/style-child-theme.scss
@@ -5,7 +5,7 @@ Author: Automattic
 Author URI: https://automattic.com/
 Description: Maywood is a refined theme designed for restaurants and food-related businesses seeking a modern look.
 Requires at least: WordPress 4.9.6
-Version: 1.2.4
+Version: 1.2.0
 License: GNU General Public License v2 or later
 License URI: LICENSE
 Template: varia
--- a/maywood/sass/style-child-theme.scss
+++ b/maywood/sass/style-child-theme.scss
@@ -10,7 +10,7 @@ License: GNU General Public License v2 or later
 License URI: LICENSE
 Template: varia
 Text Domain: maywood
-Tags: one-column, flexible-header, accessibility-ready, custom-colors, custom-menu, custom-logo, editor-style, featured-images, footer-widgets, rtl-language-support, sticky-post, threaded-comments, translation-ready, auto-loading-homepage, full-site-editing, global-styles
+Tags: one-column, flexible-header, accessibility-ready, custom-colors, custom-menu, custom-logo, editor-style, featured-images, footer-widgets, rtl-language-support, sticky-post, threaded-comments, translation-ready, auto-loading-homepage, full-site-editing, a8c-global-styles
 
 This theme, like WordPress, is licensed under the GPL.
 Use it to make something cool, have fun, and share what you've learned with others.
--- a/maywood/style-rtl.css
+++ b/maywood/style-rtl.css
@@ -582,6 +582,7 @@ html {
 html {
 	font-size: 16.66667px;
 	font-family: "IBM Plex Sans", -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
+	font-family: var(--font-base, "IBM Plex Sans", -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif);
 	line-height: 1.6;
 }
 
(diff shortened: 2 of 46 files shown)


maywood/style-rtl.css
```diff
@@ -1634,6 +1640,7 @@ ol {
 
 dt {
 	font-family: "IBM Plex Sans", -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
+	font-family: va(--font-headings, "IBM Plex Sans", -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif);
```
allancole: Noticed a small typo here `va(` should be `var(`.

varia/sass/blocks/list/_style.scss
```diff
@@ -21,6 +22,7 @@ li > ol {
 
 dt {
 	font-family: #{map-deep-get($config-global, "font", "family", "primary")};
+	font-family: va( --font-headings, #{map-deep-get($config-global, "font", "family", "primary")} );
```
allancole: Found another `va(` typo. May wanna do a find/replace on these :-)

allancole:
Love where this is headed @nosolosw! Also, nice to see the fallback for browsers that don’t support css-custom-properties too. I’m gonna do some testing with this shortly.

maywood/style-rtl.css
nosolosw: :eagle: :eye: Thank you!

allancole:
This works really well, Andrés! Only noticed a few unexpected things. 

1.) On the frontend the `#a8c-global-styles-frontend-inline-css` inline stylesheet outputs an `@import:…;` rule that appears to import _all_ the Google fonts referenced by the plugin. On first glance this seems fine, but I worry that it could slow down the page load speed if all those fonts are actually being loaded by the browser. On the backend, its fine to import all the fonts at once, but for the frontend, the import should only include the fonts selected by the user.

2.) The `'enqueue_styles' => true,` parameter name sounds similar to other hooks/functions that us themers are already used to (ie: `wp_enqueue_style`). I think if we rename it to something like `'enqueue_theme_global_styles' => true,` it’ll be a bit more unique and easier to understand for the context of the plugin. 

Other than that, this PR works REALLY well! I definitely think the changes in Varia could get merged to master sometime soon since the css-custom-property fallbacks are already in place. Great work!

nosolosw:
Those are great suggestions! Implemented them at https://github.com/Automattic/global-styles/pull/1 and https://github.com/Automattic/global-styles/pull/2. I've also rebased this from `master`, so it's ready for a second round of review.

allancole:
Sweet! I’ll take a look shortly 👍

allancole:
Crap, looks like I just ran into this issue while testing again: https://github.com/WordPress/gutenberg/issues/14457

Do you know of anyway to get the panel back @nosolosw ?

allancole:
Nevermind, I figured it out and I have a couple more questions/notes. 

Would something like this be possible?

image

The Publish button feels kind of hidden at the bottom and I’ve twice forgotten to save before loading the page. I’m hoping that if the Save button is a different color it won’t get confused for the Main Publish button. I’m also thinking the button wording could be more explicit and say something like “Save this Design”, “Save Global Styles” or “Save Design Styles.” 

**Saving UX**

Is there a way to tie the **Editor Publish** button to the **Global Styles Publish** button so that if I leave the GS panel, edit a block, and then click the **Editor Publish** button, the current Global Styles get saved so that I don’t accidentally undo what I changed? Are there any other alerts or warnings we can use in the UI that say something like: “if Global Styles have been changed but not saved, and then a user clicks the main Publish button, alert them that they have unsaved Global Styles and ask if they want to keep or undo them”?

nosolosw:
> Crap, looks like I just ran into this issue while testing: WordPress/gutenberg#14457

I added the Global Styles to the Options menu at https://github.com/Automattic/global-styles/pull/3 so if a user ever unpins the sidebar it can always be found in there.

> The Publish button feels kind of hidden at the bottom and I’ve twice forgotten to save before loading the page

As for the reset/publish buttons: I've got https://github.com/Automattic/global-styles/pull/4 to experiment. We may want to comment there on the approach, change copy, etc.

> Is there a way to tie the Editor Publish button to the Global Styles Publish button 

I think there are some things we could explore here. One is investigating the work the core team has been doing around entities - that may be exacttly what you mention. Other could be to use the pre-publish sidebar and show something there. My thinking is that it's a great addition that can be done post-launch.

iamtakashi:
Thanks @nosolosw for working on this PR. It works well in my testing, and I don't see why we should wait for adding the support to all the Varia child themes. Especially, there is nothing breaking without the plugin. As it's been mentioned, there might be some room to improve in terms of the UI, but that shouldn't hold adding the support to the themes, should it?

When are we looking at the launch?

iamtakashi:
When we merge this and add the support to other template-first themes, we'll be deploying the themes for both Simple sites and Atomic sites. Would that be a problem? I assume no because I didn't see any errors when I deactivated the plugin in my local install, but I just wanted to make sure :) 

If deploying themes would be an issue for some reason, we could also wait for the plugin to be available in both Simple and Atomic sites.

nosolosw:
@iamtakashi yeah, I think this is safe to land first, it'll work as it is now if the Global Styles plugin is not present. One this PR is merged, I'll see to open the plugin for testing.

---

For Global Styles to work it needs these changes in WordPress.com as well, so I was wondering how should we update it after this lands. I don't know how do you usually update WordPress.com after a change to the themes, but I see two options. The 1st:

1) Merge this PR.
2) Sync WordPress.com with the latest from themes.
3) Apply the WordPress.com patch that adds the Global Styles plugin (I'd remove the Maywood changes first). I'll do this.

The 2nd option I see:

1) Merge this PR.
2) Apply the WordPress.com patch with the corresponding Maywood changes included (`functions.php`, `print.css`, `style-editor.css`, `style-rtl.css`, `style.css`). I'll do this.

I'm happy with whatever works best for you.

allancole:
@nosolosw, Yeah I think option #1 sounds most like our theme deployment process. The code on GitHub is slightly ahead of wpcom, so I think the process might look more like this if we were to add support for all Varia child themes: 

1. Move `add_theme_support( 'a8c-global-styles' )` from Maywood to Varia in this PR.
2. Merge this PR to master.
3. Recompile all Varia child themes with these changes.
4. For each Varia child theme, add the new `css-custom-property` and fallback for any `font-family` declarations in the Sass files of the `varia-child-theme-name/sass` directory. Recompile again.
5. Test each Varia child theme.
6. Deploy to `wpcom`.
7. Apply the WordPress.com patch that adds the Global Styles plugin.

nosolosw:
After talking to Allan and Takashi this are the steps forward:

- Need to bump the minor version of the affected themes in package.json and style-child-theme.scss.
- Need to add the global-styles tag in the style-child-theme.scss.
- We'd like to release these changes for Maywood only in WordPress.com. However, due to how things are setup it may be easy to actually updating all varia-children at once. We're still not sure this is the way we want to go, but eventually, we'll need global styles support in all of them. So, we're going to try that approach in a different PR to assess what's our best shot.

maywood/sass/style-child-theme.scss
```diff
@@ -5,7 +5,7 @@ Author: Automattic
 Author URI: https://automattic.com/
 Description: Maywood is a refined theme designed for restaurants and food-related businesses seeking a modern look.
 Requires at least: WordPress 4.9.6
-Version: 1.1.4
+Version: 1.2.4
```
iamtakashi: I'm not an expert in semver, but don't we need to reset the patch version when we increment the minor version? ie `1.2.0` instead of `1.2.4`?

nosolosw: fixed at 8e2e524

nosolosw:
https://github.com/Automattic/themes/pull/1516/ is ready for review

nosolosw:
We're favoring https://github.com/Automattic/themes/pull/1516 so closing this for the time being. We can reopen if necessary.

Issue #1409, opened 2019-09-13
Morden: Avatars overlapping text on small devices


kyleaparker:
## Steps to replicate

[Please include screenshots, browser information, and anything else that can help us recreate the problem.]

- Navigate to the site using a small device size (iPhone 5/SE)
- Select a post and scroll down to the comments section ([example](https://doctorsofficedotblog.wordpress.com/2019/09/12/hakuna-matata-an-essential-way-of-life/)) 
- Notice the avatar moves to the right and overlaps the text

## Result

![IMG_5240](https://user-images.githubusercontent.com/712873/64877024-8594df80-d61e-11e9-93b2-d2e09fa578bd.jpg)

## Expected

I wouldn't expect the avatar to increase in size on mobile devices. I would also expect it to remain on the left side. 

* User report: #2346294-zd 
* Follow-up ticket:

PR #1444, opened 2019-09-28
Varia: Remove auto width for image

### LINKED ISSUE #1442: Varia: Image width attribute gets ignored
This becomes apparent with smaller images, but the width attribute is ignored in Varia. 

Screen Shot 2019-09-27 at 18 08 18

It's possibly caused by this change: https://github.com/Automattic/themes/commit/3e526d1b9eaac01b96f813eaf7d4ca23de3d125c

`width: auto;` 😨

iamtakashi:
Fixes #1442 

**After:**
The size matchs in both editor and front end.

Screen Shot 2019-09-28 at 16 48 46

### DIFF
--- a/varia/sass/blocks/image/_style.scss
+++ b/varia/sass/blocks/image/_style.scss
@@ -26,5 +26,4 @@ img {
 	height: auto;
 	max-width: 100%;
 	vertical-align: middle;
-	width: auto;
 }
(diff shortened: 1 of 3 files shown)


iamtakashi:
> where I know the auto width is respected (like avatar images in comments)

@allancole, Avatar is appearing the double size in WP.com because of this though See #1409 or https://iamtakashi.wordpress.com/2013/05/23/comments-with-embeds/

In self-hosted, removing width: auto doesn't change the size of the avatar.

Screen Shot 2019-09-30 at 16 59 48

allancole:
Ahhh, I see! Good catch and I noticed the same improvement to the avatar size on my end, @iamtakashi. I think this is good to be merged 👍

PR #1445, opened 2019-09-28
Varia: Change list style position and indentation

### LINKED ISSUE #1372: Redhill: Lists Style Mismatch Between Editor vs. Published Site
## Steps to replicate

1. Create a numerical and a bulleted list in a post/page.
2. Preview and/or publish the post/page.
3. Notice that the list styling in the previewed/published post/page does not match the list result in the editor.

## Result

Editor screenshot:

Block Editor

Published Post screenshot:

Published post

## Expected

That the list style in the editor should match the final published product.

* User report: https://en.forums.wordpress.com/topic/lists-shifted-to-the-edge-of-the-screen/
* Follow-up ticket: https://en.forums.wordpress.com/topic/lists-shifted-to-the-edge-of-the-screen/

### LINKED ISSUE #1441: Varia: Improve list style position
I understand the intention, but the inside list-style-position is hard to read, and it degrades the point of a bullet list. Also the style doesn't match in the editor and the front-end.

Screen Shot 2019-09-28 at 18 18 01

We should make it outside, and add a side padding to the right.

iamtakashi:
Fixes #1441.  This should fix #1372 as well.

**After**
Screen Shot 2019-09-28 at 18 13 06

This makes the list item indented to inside, but it's much easier to read. Eventually, we'd want to introduce a new spacing config for lists with `em` unit so it works relatively with the font size.

### DIFF
--- a/varia/sass/blocks/list/_editor.scss
+++ b/varia/sass/blocks/list/_editor.scss
@@ -2,6 +2,6 @@
 	ul,
 	ol {
 		margin: #{map-deep-get($config-global, "spacing", "vertical")} 0;
-		padding-left: calc(2 * #{map-deep-get($config-global, "spacing", "unit")});
+		padding-left: #{2 * map-deep-get($config-global, "spacing", "unit")};
 	}
 }
--- a/varia/sass/blocks/list/_editor.scss
+++ b/varia/sass/blocks/list/_editor.scss
@@ -2,6 +2,6 @@
 	ul,
 	ol {
 		margin: #{map-deep-get($config-global, "spacing", "vertical")} 0;
-		padding-left: #{2 * map-deep-get($config-global, "spacing", "unit")};
+		padding-left: #{2 * map-deep-get($config-global, "spacing", "horizontal")};
 	}
 }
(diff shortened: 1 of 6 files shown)


varia/sass/blocks/list/_editor.scss
```diff
@@ -0,0 +1,7 @@
+.block-library-list {
+	ul,
+	ol {
+		margin: #{map-deep-get($config-global, "spacing", "vertical")} 0;
+		padding-left: calc(2 * #{map-deep-get($config-global, "spacing", "unit")});
```
allancole: Let’s get rid of the `calc()` here and just use `#{2 * map-deep-get($config-global, "spacing", "horizontal")}`. That way its tied to same the horizontal/x-axis spacing metric used in gutters and a few other areas. 

Same should happen for the frontend here: https://github.com/Automattic/themes/pull/1445/files#diff-2ef3dcf0689f1f654e28f9549dcd415dR5

iamtakashi: 👍

iamtakashi:
@allancole The code style has been fixed.

PR #1455, opened 2019-09-30
FIX #1396: Align more block

nielslange:
Fixes #1396

@iamtakashi It seems that in #1420 you've only changed the stylesheets of `Varia` but missed to render the stylesheets for the child theme `Rockfield` which comes with its own stylesheets inherited by its parent theme.😀

I just rendered the stylesheets for `Rockfield` with this PR. I also saw that there are much more changes in the stylesheets when running `npm build` and `npm run build:rtl`. 

<table>
<tr>
<td>Before:
<br><br>

#1396 - before

</td>
<td>After:
<br><br>

#1396 - after

</td>
</tr>
</table>

PR #1457, opened 2019-09-30
FIX #1409: Limit Gravatar size

### LINKED ISSUE #1409: Morden: Avatars overlapping text on small devices

nielslange:
Fixes #1409 

Before the fix, all Gravatars were visible in their original size. This PR ensures that the Gravatars are visible with a maximum height of 32px.

<table>
<tr>
<td>Before I:
<br><br>

#1409 - before I

</td>
<td>Before II:
<br><br>

#1409 - before II

</td>
<td>After:
<br><br>

#1409 - after

</td>
</tr>
</table>

While  #1409 affects the `Morden` theme, this PR contains the styles of its parent theme `Varia`  as according to https://github.com/Automattic/themes/pull/1455#issuecomment-536551295 the process should be automated in a way that styles of child themes are created automatically.

### DIFF
--- a/varia/sass/components/comments/_comments.scss
+++ b/varia/sass/components/comments/_comments.scss
@@ -96,6 +96,8 @@
 			display: block;
 			position: absolute;
 			right: 0;
+			height: $avatar-size;
+			width: $avatar-size;
 
 			@include media(mobile) {
 				margin-right: #{map-deep-get($config-global, "spacing", "horizontal")};
--- a/varia/style-rtl.css
+++ b/varia/style-rtl.css
@@ -2937,6 +2937,8 @@ table th,
 	display: block;
 	position: absolute;
 	left: 0;
+	height: 32px;
+	width: 32px;
 }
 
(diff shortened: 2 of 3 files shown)


iamtakashi:
@nielslange Thanks for the PR.

A part of the issue #1409 (the avatar appearing larger than should be) will be fixed in #1444 — Varia has accidentally set all image width to be `auto`, and the value of the width parameter has been ignored.

#1444 should be merged today, and after that, we'll recompile Morden's CSS and push to WP.com.

PR #1459, opened 2019-10-01
Varia and Maywood: bump theme versions

vindl:

Bumps Varia to version 1.2.2.
Bumps Maywood to version 1.1.2.

### DIFF
--- a/maywood/languages/maywood.pot
+++ b/maywood/languages/maywood.pot
@@ -2,7 +2,7 @@
 # This file is distributed under the GNU General Public License v2 or later.
 msgid ""
 msgstr ""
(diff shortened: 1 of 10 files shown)


vindl:
It seems that it have pulled in some misc changes as a result of `npm run build`. My guess is that someone hasn't regenerated the assets during one of the previous merges?

iamtakashi:
Yes. I'm familiar with those changes pulled in from some recent updates in Varia. (we really want to streamline this workflow :/)

LGTM!

vindl:
Thanks for the review @iamtakashi! 🙇

Issue #1461, opened 2019-10-01
Varia: Remove featured image support on pages


iamtakashi:
See discussion in #1443.

The gist is that we should remove the featured image support from pages since the theme doesn't display on the front of the site, and having the UI in the editor is confusing.

allancole:
Fixed in #1462 👍

iamtakashi:
Thanks @liviopv, that's a good point that we forgot about. 

We should have a separate feature for that in WP.com that works regardless of the activated theme so that the customers can have a clear and consistent experience.

Having said that, in the meantime, we probably need to reenable it for pages and display it on the front-end :/

@allancole, any opinion?

eduardozulian:
> I think we need to reopen this issue and consider reverting it because it ends up creating a collateral problem of not allowing users to have control over the image that gets passed to social media.

We have a similar case on #1502 where the customer has an old logo showing up on social media and there's no way for us to force a new one.

It feels harder to set a thumbnail for FB and Twitter especially with logos extremely compressed on those themes.

Apparently, the only way would be to ask them to change to another theme, which we already did.

cecoates:
> We should have a separate feature for that in WP.com that works regardless of the activated theme so that the customers can have a clear and consistent experience.

That would be excellent, but things like podcasts, Facebook, etc. also pull featured images per post/page as well.

So having just 1 site image wouldn't cover our customer's needs.

Is it possible to have the cover block pass the necessary tags to work with Twitter, FB, etc.?

iamtakashi:
> It would still need to be individually customizable on each page/post.

Yes, absolutely. If we were making this feature, it'd be another panel in the sidebar of the editor, and it's always there regardless of the activated theme. But I'm not sure if that's going to happen anytime soon.

gracie:
How long should we expect to have this feature restored?

Another case in #6867587-HC

tijosh:
Another report of this in #2442698-zen

In this case, the OG image is being selected from other posts on the site—not the home page—which is producing a very undesired effect for the user.

In the short term, I'd say reverting this ASAP would be ideal, and in the long term, building out a universal Social Media Preview module for all themes if possible.

JoshuaGoode:
> > We should have a separate feature for that in WP.com that works regardless of the activated theme so that the customers can have a clear and consistent experience.
> 
> That would be excellent, but things like podcasts, Facebook, etc. also pull featured images per post/page as well.
> 
[...]

PR #1462, opened 2019-10-01
Varia: Remove featured image support for Pages

### LINKED ISSUE #1443: Shawburn: Featured Image in Page not Displayed
## Steps to replicate

- Open a page and add a featured image, then Update or Publish page. 

Screen Shot 2019-09-28 at 3 28 34 PM

## Result

- Featured image is not be displayed in the page. 

- Even with _Enable the Jetpack Mobile theme > Show featured images_ enabled or not, the featured image still won't be displayed in the page.

Screen Shot 2019-09-28 at 3 31 52 PM

## Expected

The featured image in the page should be displayed. 

* User report: #15476272-hc
* Follow-up ticket: #2384811-zd

allancole:
After some further discussion, we decided to go in the opposite of the direction of the solution suggested in #1452 for a couple of reasons: 

1. A featured image on a Page is _not_ currently displayed anywhere in Varia based themes. Some users may have had featured images on Pages before switching to a Varia, so if we add support their sites may be disrupted with new featured images that they’ll have to remove page-by-page. 
2. We want to ultimately push customers toward editing with Gutenberg more directly and right now they’re isn’t a _direct_ way to control how or when they appear. 

More background info here: https://github.com/Automattic/themes/issues/1443#issuecomment-536743360

- Remove featured image support for Pages and only allow them for Posts.

#### Related issue(s):
#1452 
Fixes #1443 #1461

### DIFF
--- a/varia/functions.php
+++ b/varia/functions.php
@@ -50,7 +50,7 @@ if ( ! function_exists( 'varia_setup' ) ) :
 		 *
 		 * @link https://developer.wordpress.org/themes/functionality/featured-images-post-thumbnails/
 		 */
-		add_theme_support( 'post-thumbnails' );
+		add_theme_support( 'post-thumbnails', array( 'post' ) );
 		set_post_thumbnail_size( 1568, 9999 );
 
 		// This theme uses wp_nav_menu() in two locations.


iamtakashi:
LGTM! 👍

PR #1463, opened 2019-10-01
Varia: Center image block by default.

### LINKED ISSUE #1407: Varia: Manually resized image become centred align on front of site
Originally reported here: https://github.com/Automattic/wp-calypso/pull/36026#issuecomment-528900980

Image is left aligned if you don't set the alignment both in the editor and on the front of the site.

Screen Shot 2019-09-06 at 16 06 06

But if we manually resize the image, it becomes centre aligned, because the width of the `figure` becomes the size of the image. 

Screen Shot 2019-09-06 at 20 06 56

We'll need to exclude this from the rue that adds auto margins to the sides.

```
.entry-content > *:not(... {
    max-width: 100%;
    margin-left: auto;
    margin-right: auto;
}

```

allancole:

- Center image block by default in the editor.
- Corrects caption placement on small, resized image blocks.
- This is a temporary solution that allows images to display consistently between the editor and the frontend. The permanent solution needs to happen in Gutenberg. See here: https://github.com/WordPress/gutenberg/issues/13684

## After

**Editor** | **Frontend**
------------ | ------------
image | image

## Related issue(s): 

Fixes #1407

### DIFF
--- a/varia/sass/blocks/image/_editor.scss
+++ b/varia/sass/blocks/image/_editor.scss
@@ -1,0 +1,13 @@
+/* Center image block by default in the editor */
+
+.wp-block-image {
+	text-align: center;
+}
+
+.wp-block-image > div {
+	text-align: center;
+}
+
+[data-type="core/image"] .block-editor-block-list__block-edit figure.is-resized {
+	margin:0 auto;
+}
--- a/varia/sass/blocks/image/_editor.scss
+++ b/varia/sass/blocks/image/_editor.scss
@@ -1,9 +1,5 @@
 /* Center image block by default in the editor */
 
-.wp-block-image {
-	text-align: center;
-}
-
 .wp-block-image > div {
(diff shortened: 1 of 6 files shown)


varia/sass/blocks/image/_editor.scss
```diff
@@ -0,0 +1,13 @@
+/* Center image block by default in the editor */
+
+.wp-block-image {
```
iamtakashi: Sorry, this bit was supposed to be for the front of the site.

iamtakashi:
@allancole my instruction was a bit short, the first rule was supposed to be for the front of the site. because a non-resized image is still left aligned at the moment.

Screenshot 2019-10-01 at 22 28 30

iamtakashi:
Thanks for the further fix. It looks good, and I'll merge it shortly.

PR #1501, opened 2019-10-08
Varia: Add more specificity to the `.children` selector for comments …

allancole:

- Add more specificity to the `.children` selector to make it only effect comments. This will prevent conflicts with other nested lists that also happen to use the `.children` selector. 

#### Related issue(s):

Discovered while testing the Categories Widget Block with the nested option turned on.

### DIFF
--- a/varia/sass/components/comments/_comments.scss
+++ b/varia/sass/components/comments/_comments.scss
@@ -56,7 +56,7 @@
 	}
 }
 
-.children {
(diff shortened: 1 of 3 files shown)


nielslange:
@allancole I already had a quick chat with @iamtakashi and clarified that I merged your PR accidentally as a result of some merging mess I created. 😬

PR #1500, opened 2019-10-08
Varia / Maywood: Allow Full Site Editing blocks to be styled with Gutenberg attributes

Copons:
#### Changes proposed in this Pull Request:

Remove most hardcoded styles from Site Title, Site Description, and Navigation Menu blocks, allowing the user to customize them in the Block Editor.

Unfortunately, to prevent the normal site branding styles overriding the FSE blocks `font-size`, I had to break one of my self-imposed rules for adding FSE support on themes, and I had to modify a non-FSE file: https://github.com/Automattic/themes/pull/1500/files#diff-5179757d99782aab3a850b1bf94594fa
I hope this will remain an extremely rare exception.

---

See https://github.com/Automattic/wp-calypso/pull/36588 for more details and testing instructions.

PR #1516, opened 2019-10-14
Add/global styles support to all varia-based themes

nosolosw:
This PR super-seedes https://github.com/Automattic/themes/pull/1422 by adding support for Global Styles to all varia-based themes:

- Alves
- Balasana
- Barnsbury
- Brompton
- Coutoire
- Dalston
- Exford
- Hever
- Leven
- Mayland
- Maywood
- Morden
- Redhill
- Rivington
- Rockfield
- Shawburn
- Stow
- Stratford
- Varia

For each of the above, this PR:

- Bumps the minor version.
- Adds the a8c-global-styles tag.
- Pulls all the changes from varia (current changes from master, and CSS variables to font-family and font rules).
- Updates the child theme SCSS files (adding CSS variables to font-family and font rules).

### DIFF
--- a/alves/sass/_config-child-theme-deep.scss
+++ b/alves/sass/_config-child-theme-deep.scss
@@ -323,6 +323,7 @@ $config-header: (
 		// Fonts
 		"font": (
 			"family": map-deep-get($config-global, "font", "family", "secondary"),
+			"family-css-variable": '--font-base',
 			"size": map-deep-get($config-global, "font", "size", "sm"),
 			"weight": bold,
 			"line-height": 1,
--- a/alves/sass/_config-child-theme-deep.scss
+++ b/alves/sass/_config-child-theme-deep.scss
@@ -20,10 +20,19 @@ $config-global: (
 	"font": (
 		/* Font Family */
 		"family": (
-			"primary": "\"Lora\"\, Georgia\, sans-serif",
-			"secondary": "\"Karla\"\, Arial\, sans-serif",
+			"primary": (
+				"fallback": "\"Lora\"\, Georgia\, sans-serif",
+				"css-var": '--font-headings',
+			),
+			"secondary": (
+				"fallback": "\"Karla\"\, Arial\, sans-serif",
+				"css-var": '--font-base',
+			),
 			"code": "Monaco\, Consolas\, \"Andale Mono\"\, \"DejaVu Sans Mono\"\, monospace",
-			"ui": "-apple-system\, BlinkMacSystemFont\, \"Segoe UI\"\, \"Roboto\"\, \"Oxygen\"\, \"Ubuntu\"\, \"Cantarell\"\, \"Fira Sans\"\, \"Droid Sans\"\, \"Helvetica Neue\"\, sans-serif",
+			"ui": (
+				"fallback": "-apple-system\, BlinkMacSystemFont\, \"Segoe UI\"\, \"Roboto\"\, \"Oxygen\"\, \"Ubuntu\"\, \"Cantarell\"\, \"Fira Sans\"\, \"Droid Sans\"\, \"Helvetica Neue\"\, sans-serif",
+				"css-var": '--font-base',
+			),
 		),
 		/* Font Size */
 		"size": (
@@ -140,7 +149,10 @@ $config-elements: (
 
 		// Fonts
 		"font": (
-			"family": map-deep-get($config-global, "font", "family", "secondary"),
+			"family": (
+				"fallback": map-deep-get($config-global, "font", "family", "secondary", "fallback"),
+				"css-var": map-deep-get($config-global, "font", "family", "secondary", "css-var"),
+			),
 			"line-height": map-deep-get($config-global, "font", "line-height", "md"),
 			"size": map-deep-get($config-global, "font", "size", "md"),
 			"weight": bold,
@@ -169,7 +181,10 @@ $config-button: (
 	),
 	// Fonts
 	"font": (
-		"family": map-deep-get($config-global, "font", "family", "secondary"),
+		"family": (
+			"fallback": map-deep-get($config-global, "font", "family", "secondary", "fallback"),
+			"css-var": map-deep-get($config-global, "font", "family", "secondary", "css-var"),
+		),
 		"size": map-deep-get($config-global, "font", "size", "sm"),
 		"weight": bold,
 		"line-height": 1,
@@ -202,7 +217,10 @@ $config-heading: (
 	// Fonts & Typography
 	"font": (
 		// Family
-		"family": map-deep-get($config-global, "font", "family", "primary"),
+		"family": (
+			"fallback": map-deep-get($config-global, "font", "family", "primary", "fallback"),
+			"css-var": map-deep-get($config-global, "font", "family", "primary", "css-var"),
+		),
 		// Size
 		"size": (
 			"h6": map-deep-get($config-global, "font", "size", "base"),
@@ -241,7 +259,10 @@ $config-heading: (
 $config-list: (
 	// Fonts
 	"font": (
-		"family": map-deep-get($config-global, "font", "family", "secondary"),
+		"family": (
+			"fallback": map-deep-get($config-global, "font", "family", "secondary", "fallback"),
+			"css-var": map-deep-get($config-global, "font", "family", "secondary", "css-var"),
+		),
 	),
 );
 
@@ -251,7 +272,10 @@ $config-list: (
 $config-pullquote: (
 	// Font
 	"font": (
-		"family": #{map-deep-get($config-heading, "font", "family")},
+		"family": (
+			"fallback": #{map-deep-get($config-heading, "font", "family", "fallback")},
+			"css-var": #{map-deep-get($config-heading, "font", "family", "css-var")},
+		),
 	),
 	// Border
 	"color": (
@@ -270,7 +294,10 @@ $config-pullquote: (
 $config-quote: (
 	// Font
 	"font": (
-		"family": #{map-deep-get($config-heading, "font", "family")},
+		"family": (
+			"fallback": #{map-deep-get($config-heading, "font", "family", "fallback")},
+			"css-var": #{map-deep-get($config-heading, "font", "family", "css-var")},
+		),
 	),
 );
 
@@ -297,7 +324,10 @@ $config-header: (
 		"title": (
 			// Fonts
 			"font": (
-				"family": map-deep-get($config-global, "font", "family", "primary"),
+				"family": (
+					"fallback": map-deep-get($config-global, "font", "family", "primary", "fallback"),
+					"css-var": map-deep-get($config-global, "font", "family", "primary", "css-var"),
+				),
 				"size": map-deep-get($config-global, "font", "size", "lg"),
 				"weight": bold,
 				"line-height": 1,
@@ -307,7 +337,10 @@ $config-header: (
 		"description": (
 			// Fonts
 			"font": (
-				"family": map-deep-get($config-global, "font", "family", "secondary"),
+				"family": (
+					"fallback": map-deep-get($config-global, "font", "family", "secondary", "fallback"),
+					"css-var": map-deep-get($config-global, "font", "family", "secondary", "css-var"),
+				),
 				"size": map-deep-get($config-global, "font", "size", "sm"),
 			),
 		),
@@ -322,8 +355,10 @@ $config-header: (
 		),
 		// Fonts
 		"font": (
-			"family": map-deep-get($config-global, "font", "family", "secondary"),
-			"family-css-variable": '--font-base',
+			"family": (
+				"fallback": map-deep-get($config-global, "font", "family", "secondary", "fallback"),
+				"css-var": map-deep-get($config-global, "font", "family", "secondary", "css-var"),
+			),
 			"size": map-deep-get($config-global, "font", "size", "sm"),
 			"weight": bold,
 			"line-height": 1,
@@ -353,7 +388,10 @@ $config-footer: (
 	),
 	// Fonts
 	"font": (
-		"family": map-deep-get($config-global, "font", "family", "primary"),
+		"family": (
+			"fallback": map-deep-get($config-global, "font", "family", "primary", "fallback"),
+			"css-var": map-deep-get($config-global, "font", "family", "primary", "css-var"),
+		),
 		"size": map-deep-get($config-global, "font", "size", "sm"),
 		"line-height": map-deep-get($config-global, "font", "line-height", "sm"),
 	),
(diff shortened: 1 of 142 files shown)


nosolosw:
It looks like there were some changes to the font-size related to FSE in varia and maywood that weren't yet ported to other themes. Because I've pulled changes from `master` in varia and ran `npm run build` to update each child theme, I've also got those changes in this PR (see, for example, [alves](https://github.com/Automattic/themes/pull/1516/files)).

I don't see how we can avoid this kind of conflicts, or how would we select which changes to pull from varia (only apply the font-family/GlobalStyles ones but not font-size/FSE-related). So I guess to make this easier for everyone we should update all themes at once, at least in this repository -- we don't have to update them immediately in WordPress.com, though.

I see two paths to move forward.

1) Apply these changes to automattic/themes but only update varia and maywood in WordPress.com. How this would work:

- Merge this PR in automattic/themes. This is to avoid running into conflicts and future changes to themes already have the latest from global styles.
- Declare global styles support in maywood, not varia. This way only sites with maywood will get the global styles sidebar at the moment.
- Test varia and maywood. Update them in WordPress.com.
- Release Global Styles.

Progresively, we can test and sync to WordPress.com every other themes, as we see fit. We should declare support for global styles individually. When all of them are released, we can move the global styles support from the children to varia itself.

2) Update all of them at once. How this would work:

- Merge this PR as it is.
- Test them all. Update them in WordPress.com
- Release Global Styles.

I'm comfortable with this approach when it comes to global styles changes (the font-family changes), but I don't know if the other unrelated things are ready (the font-size/FSE). If we aren't sure about this, we can go with 1) for the time being.

What do you think?

iamtakashi:
Thank you so much for working on this, @nosolosw!

> I don't know if the other things are ready (the font-size/FSE).

The recent change #1500 seems only affect to FSE enabled sites, and that is limited to Maywood at the moment. So it should be safe.

I've been a bit indecisive between **1)** & **2)** before, but I'm more comfortable with **2)** now. I'll go through each theme to test with this branch.

- [ ] Alves
- [ ] Balasana
- [ ] Barnsbury
- [ ] Brompton
- [ ] Coutoire
- [ ] Dalston
- [ ] Exford
- [ ] Hever
- [ ] Leven
- [ ] Mayland
- [ ] Maywood
- [ ] Morden
- [ ] Redhill
- [ ] Rivington
- [ ] Rockfield
- [ ] Shawburn
- [ ] Stow
- [ ] Stratford
- [ ] Varia

iamtakashi:
This is a kind of unique case, but, for Coutoire, I think the site description would look better with Heading font.

Screenshot 2019-10-15 at 14 16 46

iamtakashi:
Actually, Dalston too, I expect the site description to be with the heading font.

Screenshot 2019-10-15 at 14 22 43

iamtakashi:
(Sorry, I'm adding a comment as I find) I realised buttons use the heading font in most (if not all) of the themes, and I'd expect it to be body font.

iamtakashi:
In Redhill, the body is actually in the heading font, and the widget title is the body font.

Screenshot 2019-10-15 at 15 01 25

iamtakashi:
In Rockfield, the widget titles stay unchanged.

Screenshot 2019-10-15 at 16 46 23

nosolosw:
e6a9b1f5789ebc51901a2b9af29289f129580938 updates Coutoire's site description to use the `--font-headings` variable instead. When Global Styles is not enabled it uses the same font as before.

f8fa7d1c945c5ef1a42b508fc31e5586625cd7ed updates Dalston's site description to use the `--font-headings` variable instead. When Global Styles is not enabled it uses the same font as before.

nosolosw:
https://github.com/Automattic/themes/pull/1516/commits/17307943835723590306b96aea6f177401b30b16 fixes Redhill. This theme probably uses more font-family rules in its child CSS files that the other varia children, so it's probably best to give it a second thorough testing round.

nosolosw:
https://github.com/Automattic/themes/pull/1516/commits/28bcd91305f6457f98a38f456bb42997bbc3d8f4 fixes widgets' title in Rockfield. This was a tricky case. I hadn't added any css variable because this was using "font family **ui** from $config-global". The other cases I left were mostly related to `code` and `pre`, so these are fine until we decide [how we deal with those](https://github.com/Automattic/global-styles/issues/10). I'm going to review all the themes for any potential miss like this one.

nosolosw:
In the last bunch of commits, I've set the button font-family to font-base.

I've also added a CSS variable alternative to font-family rules that didn't have one. This was mostly for Jetpack code judging for the names (`.widget_jp_blogs_i_follow li`, and `.wp-block-jetpack-business-hours dt`). The only font-family rules that don't have an equivalent CSS variable are:

- everything in `varia/sass/base/_normalize.scss`
- code and preformatted tags and blocks ([see](https://github.com/Automattic/global-styles/issues/10))

iamtakashi:
Thank you for the fixes, @nosolosw! All the aforementioned issues have been fixed 🎉

There is one thing I'm not sure about. When Global Styles provides two choices as Base and Headings, should the main navigation take the base font or the heading font? My gut feeling is the customers expect it to be base font, but I'm not sure. @allancole do you have any opinion on this?

allancole:
> When Global Styles provides two choices as Base and Headings, should the main navigation take the base font or the heading font? My gut feeling is the customers expect it to be base font, but I'm not sure.

Interesting. Yeah, that does make things a bit tricky. This actually makes me think we should change it in Varia, but thats a separate discussion. 

Each child-theme sets which blocks/components/elements get a Heading or Base font. When you use Global Styles, I think those decisions should stay in tact even though it might make sense in some instances to swap Heading for Base (like in your main nav example). There are some themes that use the same font for Heading & Base, and so if a customer starts using Global Styles, it’ll initially be unclear which elements will change until they start swapping fonts. Another thing that makes that tricky is the header/footer aren’t visible in the editor so customers won’t see those changes until after they save/publish the Global Styles and visit the frontend. 

That does make me think that this will be kind of an awkward experience, but I think I’m okay with all this for a V1 and then we can let feedback inform how to improve the awkwardness.

Thinking toward the future with FSE, I imagine there would be a _Main Navigation Block_, and it could come with it’s own block-specific font settings that customers could use to override the Global Styles if they wanted to.

allancole:
To clarify, here’s a few examples of how the CSS structure cold work.

**Child theme with two fonts**:
```
$global-primary: Helvetica;
$global-secondary: Georgia;

--font-heading: $global-primary;
--font-base: $global-secondary;

h1 { 
    font-family: $global-primary; 
    font-family: var( --font-heading, $global-primary); 
}
p { 
    font-family: $global-secondary; 
    font-family: var( --font-base, $global-secondary); 
}
```

**Child theme with only one font**:
```
$global-primary: Helvetica;
$global-secondary: Helvetica;

--font-heading: $global-primary;
--font-base: $global-secondary;

h1 { 
    font-family: $global-primary; 
    font-family: var( --font-heading, $global-primary); 
}
p { 
    font-family: $global-secondary; 
    font-family: var( --font-base, $global-secondary); 
}
```

iamtakashi:
Thanks for chiming in @allancole. That (Global Styles should follow the pattern that the child-theme defines) makes sense. It makes sense to use the body font for `ui` doesn't it?

allancole:
> It makes sense to use the body font for ui doesn't it?

Ahhh, that’s a good catch. Yeah, the UI font should probably shift to the `body` font once the Global Style is changed. Later we might want to add UI as specific option since it’s specific to browser elements like buttons and forms. A similar thing might need to happen for `code` fonts but I think we can leave that as-is for now.

nosolosw:
I was working with this mapping:

varia | global styles
--- | ---
primary | headings
secondary | base
ui | base
buttons | base

Does this look like what we want? It should already be more or less like that, with a few exceptions (some intentional, perhaps others don't and I'll fix these tomorrow). A few that come to mind that are intentional:

* redhill/sass/style-child-theme-editor.scss (see `html` selector)
* rockfield/sass/_extra-child-theme.scss (see `widgets-title`)

I'll review tomorrow for more and will update accordingly. Wanted to post quickly for @allancole to see -- timezones are against us! :)

iamtakashi:
All the themes except Redhill use `primary` for headings.

I've noticed that, in the following themes, the main navigation uses the heading font with Global Styles while the originals specify `secondary` for it.

* Alves
* Balasana
* Coutoire
* Redhill
* Shawburn
* Stow
* Stratford

How can we specify to use `--font-base` for the navigation in these themes in a way that doesn't cause a regression next time we run the build command? What am I missing?

nosolosw:
Thanks for the clarification! Just pushed the fixes for the font in the main-nav.

The good news is that I've now grokked how the `_config-child-theme-deep.scss` works for each child. With this new knowledge I suspect that there may be other issues that we need fixed. I'm going to review that and will report back with my findings.

iamtakashi:
Thanks, @nosolosw. I can confirm the issue with the main navigation has been fixed 🎉

Would it be worth to do the same for all the configs with `font => family` in `_config-child-theme-deep.scss` (such as form, button, headings, list and so on?)

nosolosw:
Yep, exactly what I was thinking. I've taken a step back to reflect what needs to be done here and came up with a better system: half the code, less error-prone as the CSS variable is colocated within the child config file, and more future-proof as it gives us the flexibility to change code in only one place.

I've just pushed the new system for varia headers and footer, so you can have a look early.  The idea is to use a new mixin that takes care of embedding the rules.

We use it like this:

```scss
.site-info {
    @include font-family( map-deep-get($config-footer, "font", "family") );
}
```

being the config file:

```scss
$config-footer: (
	"font": (
		"family": (
			"fallback": map-deep-get($config-global, "font", "family", "primary"),
			"css-var": '--font-headings',
		),
	),
);
```

and the mixin:

```scss
@mixin font-family( $config_map ) {
	font-family: #{map-deep-get( $config_map, 'fallback' )}; // For browsers without CSS custom properties support.
	font-family: var( #{map-deep-get( $config_map, 'css-var')}, #{map-deep-get( $config_map, 'fallback' )} );
}
```

Working now through all the other use cases. It's quite easy now to make the changes :)

nosolosw:
Done migrating Varia to the new system and now the style choices are intact. The migration caught an error that we had: the buttons were using `--font-base` instead of `--font-headings` (because the `$config-button` partial sets the font family to `primary`).

nosolosw:
@iamtakashi @allancole this is ready for the final review.

After migrating to use a mixin, everything is simpler and matches exactly what the theme was using before. In case you wanted something different, it can be changed easily by editing the child's `sass/_config-child-theme-deep.scss`. Hopefully, it's self-explanatory!

There are a few places in the children's `sass/_extra-child-theme.scss` where `font-family` was set to `inherit` so those places merit your review. Please, leave a comment in the line you want changed and I'll do it.

iamtakashi:
Awesome, this is much better and follows the theme pattern. So far so good. I'll keep reviewing, but as we discussed, we should be able to merge and deploy on Monday!

Copons:
I've tested this on an FSE site (based on Maywood), and it seems to work as expected (and it's super cool @nosolosw ✨)!

Screenshot 2019-10-18 at 17 33 34

There are two elements that, in my personal opinion, didn't use the font I was expecting:

- The Main Navigation uses the heading font, which it's very awkward to me.
Here the backtracking:
https://github.com/Automattic/themes/blob/62e30c58d024b408cca5aa7be7786eee4a869b1e/varia/sass/components/header/_site-main-navigation.scss#L149
https://github.com/Automattic/themes/blob/62e30c58d024b408cca5aa7be7786eee4a869b1e/varia/sass/components/header/_config.scss#L38
https://github.com/Automattic/themes/blob/62e30c58d024b408cca5aa7be7786eee4a869b1e/varia/sass/abstracts/_config-global.scss#L14

- The Site Info, in the footer, uses the heading font as well. I don't have a strong opinion on this, but it kinda stuck out.
https://github.com/Automattic/themes/blob/62e30c58d024b408cca5aa7be7786eee4a869b1e/varia/sass/components/footer/_footer-branding.scss#L6
https://github.com/Automattic/themes/blob/62e30c58d024b408cca5aa7be7786eee4a869b1e/varia/sass/components/footer/_config.scss#L10
https://github.com/Automattic/themes/blob/62e30c58d024b408cca5aa7be7786eee4a869b1e/varia/sass/abstracts/_config-global.scss#L14

nosolosw:
I'll defer to Takashi and Allan about editorial choices, but after 

Screenshot from 2019-10-18 18-56-24

I couldn't resist to check whether what you're pointing out are actually pre-existing theme preferences before preparing myself for the weekend. So, I've checked Maywood and, indeed, it sets the main navigation to primary [here](https://github.com/Automattic/themes/pull/1516/files#diff-6b6d576e17fc0b11ca617099bc96b0c0R357), and so the site info at [here](https://github.com/Automattic/themes/pull/1516/files#diff-6b6d576e17fc0b11ca617099bc96b0c0R357). I can go in peace :sweat_smile:

Copons:
> I'll defer to Takashi and Allan about editorial choices, but after
> 
> Screenshot from 2019-10-18 18-56-24
> 
> I couldn't resist to check whether what you're pointing out are actually pre-existing theme preferences before preparing myself for the weekend. So, I've checked Maywood and, indeed, it sets the main navigation to primary [here](https://github.com/Automattic/themes/pull/1516/files#diff-6b6d576e17fc0b11ca617099bc96b0c0R357), and so the site info at [here](https://github.com/Automattic/themes/pull/1516/files#diff-6b6d576e17fc0b11ca617099bc96b0c0R357). I can go in peace 😅

Ahaha yes, sorry for not being very clear @nosolosw: all my links to the code are at current master, so yes, you totally made the right call! 😄

iamtakashi:
Let's focus on the implementation in this PR :)

As @Copons pointed out, some themes that use a single font could use tweaking. But let's do that in separate PR. (Nothing wrong with this implementation, we just left some primary because it didn't matter at that time :) )

iamtakashi:
I've reviewed all the cases of `font-family: inherit` and they seem fine. As the implementation is solid, let's merge this first, and we can tweak some cases like @Copons pointed out on a theme by theme basis in separate PRs.

iamtakashi:
@Copons I thought about the use of the heading font in the navigation and the footer in Maywood, but I think it's actually ok, and no need to change. @allancole this is subjective, but what's your opinion?

**Header:**
Screenshot 2019-10-21 at 20 27 43

**Footer:**
Screenshot 2019-10-21 at 20 27 31

Copons:
@iamtakashi it's not bad looking per se, just I think it's rather unconventional to use the same font for the site title and the site navigation.
(It also might just be a matter of perception 🤔)

In any case, I'm so happy with this new addition that I'm good either way!

Issue #1519, opened 2019-10-17
Child configurations


nosolosw:
While working on https://github.com/Automattic/themes/pull/1516 I've noticed a couple of configurations that are inconsistent. Not sure how relevant it is but want to point it out:

- `config-separator.width` is present in `blocks/separator/config` as well as the `_config-child-theme-deep.scss` file of these themes: coutoire, leven, morden, redhill, rockfield, shawburn, stow, stratford - it's not defined in the rest of the themes.
- `config-footer.font.line-height` is not present in redhill's `_config-child-theme-deep.scss`

nosolosw:
cc @allancole @iamtakashi

allancole:
WOW, what a catch! You’re right that the variable arrays should match between `child-theme-name/sass/_config-child-theme-deep.scss` and `varia/sass/abstracts/_config-global.scss`(+ its includes). 

It looks like when we added `config-separator.width` we added it to the partial but forgot to add it to `varia/sass/child-theme/_config-child-theme-deep.scss` which is the starter array that we copy into _newly_ generated child themes with the latest Varia improvements. 

If child themes were overriding either the `height` or `width` variable, they’d throw an error while compiling. But since they aren’t, Varia just skips the child-theme’s `config-separator.width` array entirely and uses it’s own so no errors pop up. 

This is definitely worth a fix, just in case a future theme tries to override the separator variable. I’ll sort it out. 

If this were done entirely with `css-custom-properties` instead of Sass variables + css-custom properties, we wouldn’t need to break up and separate the Sass-variable arrays and their partials like this. Hoping we can move in that direction sooner than later.

PR #1520, opened 2019-10-17
Varia + various child-themes: Add missing config variables

allancole:

- Adds `config-separator.width` variable to the base Varia child-theme variables array so it gets copied into new child-themes properly.
- Adds a width variable to the `config-separator` array for various child-themes to prevent potential compiling errors.
- Add `config-footer.font.line-height` default variable to Redhill.

#### Related issue(s):

Fixes: #1519

### DIFF
--- a/alves/sass/_config-child-theme-deep.scss
+++ b/alves/sass/_config-child-theme-deep.scss
@@ -280,6 +280,7 @@ $config-quote: (
 
 $config-separator: (
 	"height": #{0.25 * $baseline-unit},
+	"width": #{6 * map-deep-get($config-global, "spacing", "horizontal")},
 );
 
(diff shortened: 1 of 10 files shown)


allancole:
I _did not_ recompile the child themes intentionally. Since the bug is benign as it is, and other unrelated Varia improvements would get tossed into this PR. After this is merged, we can recompile against the master branch the next time we do a mass child-theme deploy to wpcom.

nosolosw:
This looks fine to me. Perhaps we can land after https://github.com/Automattic/themes/pull/1516 ?

allancole:
Now that #1516 is merged, I’ll resync this with master and merge it.

PR #1532, opened 2019-10-21
Varia: Fix float clearing

### LINKED ISSUE #1531: Varia: Clearing with pseudo element is not working
I think this is an oversight, but currently, the float clearing with pseudo element isn't working properly. This is raised in #1499

We see this in `/sass/base/_clearings.scss`, but these elements need actually clear floating with ` clear: both`.

```
.clear:after,
.entry-content:after,
.comment-content:after,
.site-header:after,
.site-content:after,
.site-footer:after {
	content: "";
	display: table;
	table-layout: fixed;
}
```

iamtakashi:
Fixes #1531

### DIFF
--- a/varia/sass/base/_clearings.scss
+++ b/varia/sass/base/_clearings.scss
@@ -21,7 +21,5 @@
 .site-header:after,
 .site-content:after,
 .site-footer:after {
-	content: "";
-	display: table;
-	table-layout: fixed;
+	clear: both;
 }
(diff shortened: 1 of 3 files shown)


varia/sass/base/_clearings.scss
```diff
@@ -21,7 +21,5 @@
 .site-header:after,
 .site-content:after,
 .site-footer:after {
-	content: "";
-	display: table;
```
allancole: Haven’t tested this with Chrome on Android 10 yet but I think we may need to keep these 2 `content` and `display` rules. See here: https://css-tricks.com/snippets/css/clear-fix/

Issue #1488, opened 2019-10-03
Varia: offer the skip cropping option for logos


eduardozulian:
## Steps to replicate

1. Activate one of the Varia-based themes
2. Go to Customize > Site Identity
3. Upload a logo

## Result

The theme will force you to crop the logo, even if you don't need to.

## Expected

To see the `Skip Cropping` button.

The workaround, for now, is to drag all the cropping borders to the edge of the image, then crop it.

Another reason for that is when you try to upload an SVG logo using a custom plugin. It breaks the functionality since you cannot crop the SVG even if you use the above workaround. Being able to skip cropping would fix that.

cc @danieldudzic and @allancole, per request :)

* User report: 13242987-hc

PR #1565, opened 2019-10-23
Varia: Restore the “Skip cropping” option to custom-logo setup.

allancole:

- Sets the _custom-logo_ `flex-height` and `flex-width` parameters to true so that customers are not forced to crop their logo and/or manually remove the crop area. 
- Customers can still crop their logos if they choose to.
- This change may require some logos to be resized manually before uploading to the customizer.

**Note**
- This change should be tested against a site that already has a logo setup with the previous logo size settings to make sure there’s no disruption when/if this change gets deployed.

#### Related issue(s):

Fixes: #1488

### DIFF
--- a/varia/functions.php
+++ b/varia/functions.php
@@ -88,7 +88,7 @@ if ( ! function_exists( 'varia_setup' ) ) :
 				'height'      => 96,
 				'width'       => 100,
 				'flex-width'  => true,
-				'flex-height' => false,
+				'flex-height' => true,
 				'header-text' => array( 'site-title', 'site-description' ),
 			)
 		);


iamtakashi:
I've tested this, and it didn't affect to a logo that was already uploaded.

When I add a new logo with this, the logo appears in the size of the image (no overflow, no restriction in height). That's also what we're aiming, correct?

allancole:
> When I add a new logo with this, the logo appears in the size of the image (no overflow, no restriction in height). That's also what we're aiming, correct?

@iamtakashi, yeah thats a side effect of removing the cropping constraints. According to HE’s its easier for them this way, as they can either recommend custom CSS to control the size or have the customer regenerate their logo at the exact size they want it to appear. 

FSE will replace this entirely eventually. Until then, whatever makes this easier to explain for HE’s to explain is probably best. We should probably have an HE confirm this before merging.

allancole:
Gonna go ahead an merge this since it doesn’t appear to cause disruption.

eduardozulian:
Thank you!

PR #1572, opened 2019-10-23
Varia: Restore featured images for Pages

allancole:

1. Revert this feature bringing back the Featured Image section for Pages in the editor.
2. Add the template tag to Page templates to display the image.
3. Connect the Featured Image to Jetpack Content Options 
4. Add meta link Content Options support for date, categories, tags, author, and comment.
5. Sets featured Images on Pages to `false` by default to prevent disruption on homepages. This way customers can select a Featured Image for the `og:image` and have an option to display or hide the image on the page (or home page) using the customizer.

#### Related issue(s):

Fixes: #1461


varia/inc/template-tags.php
```diff
@@ -212,16 +211,16 @@ function varia_post_thumbnail() {
 		if ( is_singular() ) :
 			?>
 
-			<figure class="post-thumbnail">
+			<figure class="post-thumbnail responsive-max-width">
```
iamtakashi: @allancole I remember the feedback saying the featured image can be too big if it goes full width. But adding this css class now changes a lot of sites, and the users can't change it back. Are we ok with that?

allancole: Ahhh, good catch! I was thinking it should have just been the `.aligndefault` column width from the beginning so I just changed it. But you’re right, if we add that class now it‘ll disrupt things for customers. I’ll remove it…

varia/inc/template-tags.php
```diff
@@ -220,8 +219,8 @@ function varia_post_thumbnail() {
 		else :
 		?>
 
-			<figure class="post-thumbnail">
-				<a class="post-thumbnail-inner alignwide" href="<?php the_permalink(); ?>" aria-hidden="true" tabindex="-1">
+			<figure class="post-thumbnail responsive-max-width">
```
iamtakashi: Here as well.

allancole: Done 👍

allancole:
#1579 is related to this issue. Seems like we should probably support featured images on all post-types supported by wpcom which also includes `portfolio`.

Issue #1508, opened 2019-10-09
Stratford: mobile menu is partly covered on small screens


caimin-j:
## Steps to replicate

On smaller screens, the *Menu* is partly covered by the div below it:

<img width="307" alt="Screenshot 2019-10-09 at 11 41 56" src="https://user-images.githubusercontent.com/37620825/66476144-c6044380-ea8c-11e9-8061-128b7d97cb13.png">

Checked with Chrome 77 on a Mac and Chrome 77 on an iPhone. I'm just using the out-of-the-box theme and menu settings, nothing fancy.

* User report: [#2409474-zd](https://woothemes.zendesk.com/agent/tickets/2409474)

I sent the user this quick CSS fix:

`.main-navigation {
    height: 30px;
}`

iamtakashi:
I believe this issue should be resolved in #1544. In my local install, I can't replicate the issue.

caimin-j:
Looks fixed to me :) The issue is resolved on my test site; I'll close this.

PR #1573, opened 2019-10-24
Stratford: Refactor navigation styling to fix various issues

danieldudzic:
While working on #1508 I have noticed a number of issues with the Startford navigation, like:

**1. Inconsistent borders**

borders

**2. Margin issues**

nav

---------------------------

The update addresses all that:

strat

stratford

### DIFF
--- a/stratford/sass/_extra-child-theme.scss
+++ b/stratford/sass/_extra-child-theme.scss
@@ -87,6 +87,7 @@ a {
 	-moz-transition: all .25s ease-in-out;
 	transition: all .25s ease-in-out;
 	width: 100%;
+
 	.site-header-wrapper {
 		display: flex;
 		flex-wrap: wrap;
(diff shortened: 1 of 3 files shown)


stratford/sass/_extra-child-theme.scss
```diff
@@ -168,6 +171,62 @@ a {
...
+						border-style: solid;
+						border-width: 0 0 .125em;
+						width: auto;
+
+					}
+					ul li a	{
+						border-color: $color_background_light;
+						border-style: solid;
+						border-width: 0 0 .125em;
+						padding-left: 0.75em;
```
iamtakashi: I'm not sure if we need the side padding to indent the sub-menu items. When we have the `-`, we might not need it? Varia doesn't add the indentation.

Screenshot 2019-10-24 at 12 00 04

Screenshot 2019-10-24 at 12 04 17

Issue #1576, opened 2019-10-24
Stratford: Category Pages and Tag Pages missing Continue Reading link after post excerpts


lizswafford:
## Steps to replicate

- Activate the theme Stratford (or Maywood) on a simple site. 
- Add a menu button to a specific category or tag to create the Category Page or Tag Page.
- View the new Category or Tag Page on the published view of the site. 
- The link will be something like: https://myawesomesite.com/tag/excellent/

## Result

- When viewing the Category or Tag Page you will see the post excerpts. 
- However, there is no Read More or Continue Reading link to go to the full post. 
- This happens even though there is a Read More Block in each post.

## Expected

- Expect to see a Continue Reading link below the post excerpt as shown on the blog page assigned from the Customzier > Homepage Settings. 

## Screenshots
Tag Archive Page - https://d.pr/R7VgwX
Blog Page - https://cld.wthms.co/CRP9QC

* User report:  #2440638-zen
* Follow-up ticket:  #2440638-zen

@allancole

allancole:
Thanks @lizswafford 

As you mentioned earlier this actually effects multiple Varia-based themes, so I’ll fix it in Varia so all the others can benefit from this fix.

carladoria:
Thanks @allancole for working on this. I had this issue today with Redhill theme. Do you know when this fix will be updated?

PR #1577, opened 2019-10-24
Varia: Add `Continue reading ->` link filter for excerpts and read more blocks.

allancole:

- Adds a filter to add a “Continue reading ->” link to excerpts and full-post content when using the read-more block or the read-more classic comment.

**Note:** 
This change should be _non-destructive_, so all Varia child-themes can be re-compiled on this PR or on master branch after this is approved and committed. 

#### Related issue(s):

Fixes: #1576


iamtakashi:
@allancole If More block is placed before 55th words the link doesn't show up, but I think that's a core issue. Otherwise, it works 👍

allancole:
@iamtakashi Yeah, that sounds like a core issue, I’ll go ahead and merge this then.

PR #1509, opened 2019-10-09
Varia: Add WooCommerce Support

allancole:
#### Changes proposed in this Pull Request:

- This PR primarily adds basic support for WooCommerce when the plugin exists.
- Adds a new `alert` color group to the variable config file. Child themes can overwrite these if they choose.
- Adds a new `style-woocommerce.scss` file that needs to be connected to the [build tool](https://github.com/Automattic/theme-dev-utils/) for automatic inclusion in future Varia-based child themes.
- Fixes `.children` selector in comments so that WC list widgets display properly.
- Introduces a new `_extends.scss` file as an abstract for unifying buttons across the theme.

Once merged, all Varia child themes should be able to receive instant WC support simply by recompiling styles :-). 

#### ToDo:

- [ ] Clean up all WC Sass stylesheets to remove redundancy and improve clarity compared to base Varia styles.
- [ ] Add a Cart UI to the header of Varia (Can FSE support this?).
- [ ] Add woocommerce support tag for `wpcom`.
- [ ] Review [WooCommerce docs](https://docs.woocommerce.com/document/third-party-custom-theme-compatibility/) to ensure efficient compatibility.
- [ ] Add new `alert` color group to all child themes to prevent compiling errors.
- [ ] Test with all child themes.

PR #1578, opened 2019-10-24
Varia: Add WooCommerce support (with refactored mobile mini-cart UI)

allancole:

- This PR is an improvement on #1509 and simply introduces a better mini-cart UI for Mobile and Desktop screens. 
- Adds a separate Cart button for mobile screens so you can open the mini-cart and the main-menu independently.
- Desktop views should remain unchanged.
- So far its been tested with Brompton, Redhill, and Coutoire. More to come. 

**Before Closed (no cart link at all!?)** | **Before Open (Cart is hidden below the full menu)**
------------|------------
image|image

**After Closed** | **After Open**
------------|------------
image|image

#### Todo:

- [ ] Test with all child themes. 
- [ ] Once approved, merge into `#1509` (not master) 

#### Related issue(s):

#1509


pmcpinto:
Everything seems to be working properly, there's just one minor issue with the icon that's displayed in the `add to cart` button after it's clicked: https://cld.wthms.co/EPlCWx

Btw, it seems to only happen with products displayed within a block, so probably it might be an issue on our side.

PR #1592, opened 2019-10-25
Rockfield: Unset measuere variable default

iamtakashi:


iamtakashi:
> We can either fix it on this PR, or merge this as-is and introduce the Block Style in a separate PR.

Let's do that in a separate issue and PR.

allancole:
> Let's do that in a separate issue and PR.

Sounds good. Let’s get this merged then 👍

PR #1599, opened 2019-10-25
Varia: Unset measure variable default for child themes

### LINKED ISSUE #1598: Varia: Unset measuere variable default for child themes
We've fixed in Varia https://github.com/Automattic/themes/commit/7c0a36aa61ab9721441222301d1f3fdba330d02a a while ago, but we forgot to change it in the config file for child themes, so all of them had the issue :) Let's change it so that child theme in future doesn't need to change it manually.

Related issue #1515

iamtakashi:
Fixes #1598


allancole:
Good catch, LGTM! 👍

Issue #1579, opened 2019-10-24
Varia: Featured image not supported on testimonial post type


samiff:
## Steps to replicate

1. Activate the Varia theme on a WordPress.com site.
2. Make sure Testimonials are turned on in `Settings → Writing → Content Types`.
3. Create a new Testimonial.

## Result

* When creating a new testimonial, there is not an option to set a featured image.
* If editing an existing Testimonial that had a featured image previously set (while using a different theme), then you are unable to edit that featured image as well.
* This behavior is the same when using a Varia child theme. 

## Expected

To be able to set the featured image for a Testimonial post type.

#user-report
2445199-zen

PR #1609, opened 2019-10-30
Varia: Restore featured images for Pages

allancole:

- Turn on featured images for _all_ posts types to allow customers to use it for social media (`og:image`) on posts, pages, portfolios, testimonials and anything else. 
- Featured images are displayed on posts-only to prevent disruption. 

#### Related issue(s): 

Fixes: #1461 
Fixes: #1579 
Replaces: #1572


iamtakashi:
LGTM! I can see the panel in the editor, and it does not show up on the front-end (This is what we want at this moment.)

PR #1612, opened 2019-10-30
(DON'T MERGE) Only display default credit info if FSE is not active.

noahtallen:
We'll be using the FSE footer credit block instead.

**Note: This should NOT be deployed until the footer credit is included on WordPress.com.**

### Changes:
- Move footer credit structure inside if statement which runs if FSE is not active. 
- (EDIT): Removed manual text align setting for credit if FSE is active. We want the user to set that with blocks (tm)

### Testing:
1. Pull these changes to your local FSE setup.
2. The footer credit will no longer show up if FSE is active.
3. The footer credit will show up if FSE is inactive.
4. You can use the footer credit block from the below PR. If you put it in the footer, it should look identical to the implementation we are moving in this PR.

### DIFF
--- a/maywood/style.css
+++ b/maywood/style.css
@@ -2832,6 +2832,13 @@ body:not(.fse-enabled) .footer-menu a {
 .entry-content .more-link {
 	display: block;
 	color: inherit;
+	margin-top: 16px;
+}
+
+@media only screen and (min-width: 560px) {
+	.entry-content .more-link {
+		margin-top: 32px;
+	}
 }
 
 .entry-content .more-link:after {
@@ -3650,6 +3657,13 @@ body.admin-bar .widget_eu_cookie_law_widget.widget.top {
 }
 
 /**
+ * Business Hours
+ */
+.jetpack-business-hours dd {
+	padding-left: 0;
+}
+
+/**
  * Child Theme Extra Styles
  */
 /**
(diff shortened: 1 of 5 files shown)


Copons (changes requested):
I think you have moved the `site-info` into the wrong conditional block 😅 

It's inside this:
https://github.com/Automattic/themes/blob/dde2d7c8af88a24252a9793f6583c89e4f68701a/varia/footer.php#L26-L38

But instead it should go at the end of this:
https://github.com/Automattic/themes/blob/dde2d7c8af88a24252a9793f6583c89e4f68701a/varia/footer.php#L23-L39

noahtallen:
Ah great catch! Does 34c3287 look right?

maywood/style.css
```diff
@@ -3649,6 +3656,13 @@ body.admin-bar .widget_eu_cookie_law_widget.widget.top {
 	font-size: 0.83333rem;
 }
 
+/**
+ * Business Hours
```
noahtallen: These changes must not have been rebuilt at some point

Copons (approved):
Changes in `footer.php` LGTM!

Though, there are some SCSS changes going on at the moment with the Varia themes, that might break stuff unexpectedly.
Since this change does not have any styling changes, do you mind removing the CSS files from the PR before merging?

noahtallen:
@copons those are expected style changes (I updated the PR description and left a comment) which simply removes an FSE override for the footer credit area. This shouldn’t break anything since that style only existed if FSE was active.

However you do bring up a good point. Should we just wait to merge until the Dotcom integration of the footer credit block is ready? If we merge this and then deploy varia, current FSE sites won’t see a footer credit area.

Copons:
@noahtallen Ah, you're right, there are a few unrelated CSS changes but also related ones, sorry about it.

Anyway, yes let's wait for the Site Credit block, and deploy in sync! 👍

vindl (approved):

noahtallen:
this will be superseded by proper FSE themes.

PR #1630, opened 2019-10-31
Varia: Fix FSE button-style `%extend` compiling error.

allancole:

- Fixes button style compiling errors in FSE after `%extends` moved to a new location in Sass directory.
- Code tidying.


allancole:
@Copons, @iamtakashi This is another small change that was necessary for WC but also effects FSE. I needed to move `%button-style` out of `sass/button/style` and into its own `_extends.scss` file so that WC and other vendors achieve the global appearance of the button without repeating the Button *Block* styles as well. It’s small but it’ll prevent a lot of duplicated CSS :-).

Copons:
Thank you so much, I really appreciate this fix! ✨

PR #1641, opened 2019-11-04
Bump Varia to 1.5.0 and Maywood to 1.4.0

Copons:


Copons:
@iamtakashi @allancole just as a heads up: could you please avoid merging stuff for Varia and Maywood before I merge this (and the two related WPCOM diffs)?
I was supposed to do a Full Site Editing release but for various reasons it got very late, and I need to deploy the themes and the plugin in sync.
Thanks in advance! 🙇

PR #1652, opened 2019-11-11
Only run dalston_block_extends() for editor

westonruter:

I was checking Dalston for AMP-compatibility and I was surprised to see 46 separate scripts being added to the frontend. This was especially surprising because the other Varia child themes enqueue nothing. It seems there is a typo where `dalston_block_extends()` should be running at the `enqueue_block_editor_assets` action rather than `enqueue_block_assets`. This fixes that apparent bug, as well as improving AMP-compatibility while also improving performance by not adding a lot of unused editor scripts to the frontend.


michaeldcain:
Great catch! I'm not sure why the decision was made to enqueue individual scripts and styles for each block, but it does seem like each of the CSS files [have front-end styles](https://github.com/Automattic/themes/blob/master/dalston/block-extends/extend-cover-block.css#L86) which will need to be enqueued.

It'd probably be better to split those out into a single separate enqueue on the `enqueue_block_assets` hook, or to just add them to the theme styles, since this functionality is bundled specifically with Dalston.

@allancole might have some thoughts about it.

allancole:
Wow, great catch @westonruter! 

> I'm not sure why the decision was made to enqueue individual scripts and styles for each block, but it does seem like each of the CSS files have front-end styles which will need to be enqueued.

@michaeldcain This was maybe my second time writing a custom block style in a theme. Gonna chalk it up to learning-in-progress ;-) 

It might be better to do an `@import` on the frontend block styles in the main Dalston Sass stylesheet. Will borrow this fix and spin up a new PR that adds the frontend block styles to the main stylesheet.

westonruter:
@allancole I'm not sure about `@import`. Last I knew it had performance problems: https://csswizardry.com/2018/11/css-and-network-performance/#avoid-import-in-css-files

westonruter:
Oh, if if this is only in the Sass then it would be compiled, right? Nevermind then 😊

sixhours:
A very belated thanks, Weston! I've moved this to #3830. :)

PR #1653, opened 2019-11-11
Varia: Skip skip-link-focus-fix in AMP

westonruter:

The skip-link-focus-fix involves custom JavaScript which is not allowed in AMP (in this way). In any case, the AMP framework has internalized this IE11 a11y fix per https://github.com/ampproject/amphtml/issues/18671 so there is no reason to add it on AMP pages.


michaeldcain:
This was deploy to WordPress.com in r55046-wpcom-themes.

PR #1654, opened 2019-11-11
Varia, Maywood: Fix FSE Navigation Menu width when alignwide or alignfull

Copons:
Fix a regression on Maywood, that caused `alignwide` and `alignfull` Navigation Menu blocks not having the expected width.

| Before (`alignwide`) | After (`alignwide`) |
| - | - |
| Screenshot 2019-11-11 at 20 24 42 | Screenshot 2019-11-11 at 20 24 28 |

### DIFF
--- a/maywood/style-editor.css
+++ b/maywood/style-editor.css
@@ -1344,6 +1344,10 @@ b, strong {
  */
 }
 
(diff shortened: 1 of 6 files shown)


Copons:
Thanks @noahtallen, I'll merge and deploy first thing tomorrow to avoid a deploy-and-run tonight!

Issue #1597, opened 2019-10-25
Varia: text-align: center in aligncenter class


iamtakashi:
@allancole What was the reason for the CSS class `aligncenter` to have `text-align: center`? 

I'm asking this because I've seen blocks having the class as the default such as Newspack Homepage Articles block, and I remember some Jetpack blocks used to have that too. 

I'm not too sure why those blocks have the class, but because of `text-align: center` in Varia, the text inside appears as centre-aligned on the front-end while it's left-aligned in the editor.

![localhost_wordpress_wp-admin_post php_post=2123 action=edit](https://user-images.githubusercontent.com/908665/67586892-97ec5800-f74a-11e9-9172-d8bb3376e6f0.png)

![localhost_wordpress__p=2123](https://user-images.githubusercontent.com/908665/67586908-a2a6ed00-f74a-11e9-9cd3-884980a77d9e.png)

Twenty Nineteen doesn't have it, and with auto side margins, it's probably not necessary. What do you think?

allancole:
@iamtakashi Hmmm, I think this is an artifact from adapting to some of the changes that have happened in Gutenberg over time. Early on, you couldn’t always center text contained in a Block (like the Cover Block, for example), so centering the text with `.aligncenter` was there as an assumption on what a customer might be trying to do. 

Now that we can more explicitly set text-align rules for text in Blocks and nested Blocks, I’m pretty sure we can safely remove it, without causing a major regression. But it’s definitely worth testing with other Blocks to be sure.

iamtakashi:
@allancole I saw widgets blocks are benefitting from the rule. The position of the bullets is strange though.

<img width="822" alt="Screenshot 2019-10-28 at 12 29 53" src="https://user-images.githubusercontent.com/908665/67678845-86d85c80-f97f-11e9-9b45-8bcec1018ef0.png">

Twenty Nineteen and Twenty Twenty don't have that rule in the class, and that makes me think what the expectation on this.

Should we keep it for now?

allancole:
Hmmm, on second thought, I wonder if it’s worth keeping `.aligncenter` as a utility class so we can use it in templates. Sort of like what we have for `.margin-top-none`.

iamtakashi:
> I wonder if it’s worth keeping .aligncenter as a utility class

Oh definitely. The question was whether we keep `text-align: center` in the class or not.

allancole:
[...]

PR #1660, opened 2019-11-12
Varia: Adds support for left, center and right aligned lists.

allancole:

- Allow list-item `li` text and bullets to respect text-alignment rules in the editor and frontend when it makes sense.
- Applies to both unordered `ul` and ordered `ol` lists whenever a text-align utility class is added to them.
- Also applies to core widget blocks like archives, categories and other lists that support text alignment options. 

**Before:**
image

**After:**
image

**Editor After:**
image

#### Related issue(s):

Fixes: #1597

### DIFF
--- a/varia/sass/blocks/list/_editor.scss
+++ b/varia/sass/blocks/list/_editor.scss
@@ -6,11 +6,13 @@ ol {
 	// Utility classes
 	&.aligncenter {
 		list-style-position: inside;
+		padding: 0;
 	}
 
 	&.alignright {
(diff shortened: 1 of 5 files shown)


iamtakashi:
Nice! Although we want to remove the left padding from the centre- and the right-aligned list items.

michaeldcain:
@iamtakashi @allancole: I removed the padding for center and right-aligned lists (the commit message is wrong). This should be read for a final +1.

iamtakashi:
LGTM 👍

Issue #1659, opened 2019-11-12
Mayland: Color scheme causes key prop warning in Editor


obenland:
## Steps to replicate

With Mayland activated, select two block and transform them into a group. Check the browser console to find:

> Warning: Encountered two children with the same key, `#000000`. 

## Cause

In `functions.php`, Mayland defines `#000000` twice, once as the primary color and again as Black.

PR #1661, opened 2019-11-12
Mayland: Fix key prop warning in Editor.

allancole:
Simply removing the duplicate colors might cause disruptions for customers who have selected one of the two `#000000` colors in their editor content. As an alternative solution, I changed the Black color to `#010101` which isn’t perfect black but it’s close enough to _not_ cause an obvious disruption.

The difference is virtually indistinguishable as you can see here: 
image

- Change duplicate `#000000` theme color to `#010101` to fix key prop console errors.
- Replaces some instances of `black` and `white` keyword colors with their hex color equivalents to help distinguish between theme-specific colors, and similarly named general colors that come from Varia.
- Recompiles all stylesheets with the color change.

#### Related issue(s):

Fixes: #1659

### DIFF
--- a/mayland/functions.php
+++ b/mayland/functions.php
@@ -75,7 +75,7 @@ if ( ! function_exists( 'mayland_setup' ) ) :
 				array(
 					'name'  => __( 'Black', 'mayland' ),
 					'slug'  => 'foreground',
-					'color' => '#000000',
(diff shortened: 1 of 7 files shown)


allancole:
Gonna go ahead and merge this one after running a rebase 👍

PR #1680, opened 2019-11-26
Stow: Add FSE support

noahtallen:
Adds FSE support to stow.

### Test:
1. On a site without FSE, everything should look as it did before this PR
2. With FSE, test many combinations of changes to the header/footer blocks including but not limited to:
a) Changing the font size
b) Changing the text color
c) Changing the background color
d) Putting things inside a cover/group block
3. Test that the default template part content matches the theme demo as much as possible.

Closes https://github.com/Automattic/wp-calypso/issues/37799


stow/sass/style-child-theme-editor.scss
```diff
@@ -32,6 +32,12 @@
  */
 @import "../../varia/sass/blocks/editor";
 
+/**
+ * Full Site Editing
+ * - Full Site Editing overrides
+ */
+ @import "full-site-editing-editor";
+
```
Copons: I've generally added the FSE overrides at the very bottom of the child theme files, after the extras, in the off-chance that those extras affect the FSE styles making it a pain to debug.

stow/sass/_full-site-editing.scss
```diff
@@ -0,0 +1,60 @@
...
+.fse-enabled {
+	#masthead {
+		.site-description {
+			color: unset;
+		}
+		.site-title {
+			font-size: unset;
+		}
+	}
+}
```
Copons: These also unset the blocks style attributes.
Unfortunately, this is one of those annoying cases where we need to work outside the FSE files, and into the original theme:

https://github.com/Automattic/themes/blob/3834391965040259a8fe546a375153b385ec59dc/stow/sass/_extra-child-theme.scss#L60-L65

Moving those rules into something like (right after the `#masthead` block) this would suffice:

```scss
// Prevent overruling the user defined font size and color set inside Gutenberg
// for Full Site Editing's Site Title and Site Description blocks.
body:not( .fse-enabled ) #masthead {
	.site-title {
		font-size: #{map-deep-get($config-global, "font", "size", "xxl")};
	}
	.site-description {
		color: #333;
	}
}
```

Copons: Oh, just noticed that this happens for the navigation menu as well.
Please replace the unsets with a `body:not( .fse-enabled )` where needed.

Copons:
@noahtallen apologize for the messy reviews!
It seems a lot of stuff, but I think it should all be simple enough to fix (easy for me to say, when I've been doing it for a while now 😄).
This said, the theme looks almost perfect to me! ✨

stow/sass/_full-site-editing.scss
noahtallen: Damn, I was pretty sure the block attributes were still working :/

ianstewart:
I think we might need a bit more work on Stow …

image

I made a quick illustration of the Editor and the Front-End views over top of the Theme Demo at 50% opacity …

image

apeatling:
@noahtallen Removing the unset of the site title font size, and setting the site title size to a custom 40 matches the site title with the demo.

apeatling:
Changing the site description from font size in the editor from 14.6 to 15 also helps match the demo site.

noahtallen:
@Copons I think this is ready for another look.

@ianstewart I think most of this should be improved now, thanks for taking a look!

PR #1681, opened 2019-11-26
Alves: Add Full Site Editing support

Copons:
* Added FSE support to Alves.
  * Note: the theme `#masthead` style were preventing the FSE header to be full width as intended. Instead of needlessly overriding the whole `header.php`, I've opted to add a simpler `body:not( .fse-enabled ) #masthead`.
* Update Varia (and the other FSE themes) with FSE improvements:
  * Hide the (+) icons when rendering the Template Part block inside the Page editor (especially useful for the Columns block).
  * Remove the top margin if the Site Title block is the first child of an editor layout.
  * Make sure each column in a Columns block is max width, so that nested FSE blocks can behave as expected.

| Template | Page | Front End |
| - | - | - |
| Screenshot 2019-11-26 at 14 53 51 | Screenshot 2019-11-26 at 14 53 35 | Screenshot 2019-11-26 at 14 54 01 |

#### Testing instructions

Smoke test Alves on an FSE environment!

### DIFF
--- a/alves/footer.php
+++ b/alves/footer.php
@@ -15,6 +15,12 @@
 
 	</div><!-- #content -->
 
+	<?php if ( class_exists( 'A8C\FSE\WP_Template' ) ) : // If the FSE plugin is active, use the Header template for content. ?>
+	<footer class="fse-template-part fse-footer entry-content">
+	<?php
+		$template = new A8C\FSE\WP_Template();
+		$template->output_template_content( A8C\FSE\WP_Template::FOOTER );
+	else : // Otherwise we'll fallback to the default Varia footer below. ?>
 	<footer id="colophon" class="site-footer">
 		<?php get_template_part( 'template-parts/footer/footer', 'widgets' ); ?>
 
@@ -37,7 +43,8 @@
 				);
 				?>
 			</nav><!-- .footer-navigation -->
-		<?php endif; ?>
+			<?php endif;
+	endif; ?>
 
 		<div class="site-info">
 				<?php $blog_info = get_bloginfo( 'name' ); ?>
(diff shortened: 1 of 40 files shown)


noahtallen:
This also demonstrates it. In the template part editor and on the front end, the tops of the description and nav block are aligned, but in the page editor they are not:

Screen Shot 2019-11-26 at 3 53 07 PM

Copons:
@noahtallen ah!
To fix the Column block messing up with the top margin in the page editor, I added a Spacer block (20px) in the default Header template.
The Spacer is there in the FSE site, but looking at your screens I'm pretty I haven't snapshotted the change.

It's a smelly fix, but I honestly don't know how to work around the Columns block quirks. 🙁

noahtallen:
> It's a smelly fix, but I honestly don't know how to work around the Columns block quirks. 🙁

as long as it can get us somewhere for now, it should be ok 😬 I'll recheck in a bit! I'll copy the code from the demo site if needed

apeatling:
@Automattic/cylon If support for FSE is in Alves, why is it not showing up in the signup step?

andrewserong:
@apeatling for the FSE themes to show up in the design selection step they need to be added to `themes-data.js` in Calypso: https://github.com/automattic/wp-calypso/blob/d1be047e543cb010e05a56fd1db25b038e749c51/client/lib/signup/themes-data.js#L428 — I'll whip up a PR to add in Alves.

noahtallen:
Alves has not been deployed on dotcom yet, so it won't be available until D36095-code is landed

andrewserong:
Thanks for confirming @noahtallen. I've put together a draft PR for adding it in when it's ready: https://github.com/Automattic/wp-calypso/pull/38157 — it looks a bit odd with 4 themes, so we can always keep this PR open until we have enough themes (e.g. 6) to merge in that'll look good on that step.

noahtallen:
makes sense! hopefully we have more landing soon :D

PR #1684, opened 2019-11-28
Morden: Add Full Site Editing support

Copons:
* Added FSE support to Morden.

Note: since Morden has a multi-column header like Alves, I've checked out this branch off `add/alves-fse-support` to obtain the new Varia columns fixes.
They help a bit, but definitely not enough. Unfortunately working with the Columns block is just very wonky.

| Template | Page | Front End |
| - | - | - |
| Screenshot 2019-11-28 at 16 38 51 | Screenshot 2019-11-28 at 16 39 07 | Screenshot 2019-11-28 at 16 39 18 |

#### Testing instructions

Smoke test Morden on an FSE environment!

### DIFF
--- a/morden/header.php
+++ b/morden/header.php
@@ -23,6 +23,17 @@
 <div id="page" class="site">
 	<a class="skip-link screen-reader-text" href="#content"><?php _e( 'Skip to content', 'varia' ); ?></a>
 
+	<?php if ( class_exists( 'A8C\FSE\WP_Template' ) ) : // If the FSE plugin is active, use the Header template for content. ?>
+
+		<header class="fse-template-part fse-header entry-content">
+			<?php
+				$template = new A8C\FSE\WP_Template();
+				$template->output_template_content( A8C\FSE\WP_Template::HEADER );
+			?>
+		</header>
+
+	<?php else : // Otherwise we'll fallback to the default Varia header below. ?>
+
 		<header id="masthead" class="site-header">
 
 			<div class="site-header-wrap">
@@ -70,5 +81,7 @@
 			</div>
 
 		</header><!-- #masthead -->
+
+	<?php endif; ?>
 
 	<div id="content" class="site-content">
--- a/morden/languages/morden.pot
+++ b/morden/languages/morden.pot
@@ -2,7 +2,7 @@
 # This file is distributed under the GNU General Public License v2 or later.
 msgid ""
 msgstr ""
-"Project-Id-Version: Morden 1.0\n"
+"Project-Id-Version: Morden 1.5.0\n"
 "Report-Msgid-Bugs-To: http://wordpress.org/support/theme/morden\n"
 "POT-Creation-Date: 2019-08-08 09:07:01+00:00\n"
 "MIME-Version: 1.0\n"
--- a/morden/package-lock.json
+++ b/morden/package-lock.json
@@ -1472,8 +1472,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -1494,14 +1493,12 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
-          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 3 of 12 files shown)


Addison-Stavlo:
I am testing things locally and most of the things look like the demo site and are consistent across the different sections of editor, preview, and front end.

One thing I notice that is missing from the demo is the line under the currently visited section of the nav bar.  

**From Theme Demo**
Screen Shot 2019-11-28 at 10 04 07 PM
**From local FSE front end**
Screen Shot 2019-11-28 at 10 05 35 PM


I also notice that the menu button on mobile width is a bit wider and there is no visible text at first.

**From Theme Demo**
Screen Shot 2019-11-28 at 10 08 04 PM
**From local FSE front end**
Screen Shot 2019-11-28 at 10 08 09 PM

And the last thing I am not quite sure of is getting the page Title out of the way to test how this would look as a home page with this header.  I am having this problem with all FSE on my local setup currently.  I have tried setting this as the Front Page but it still renders the title in all modes.  Which makes the homepage look like:
Screen Shot 2019-11-28 at 10 13 27 PM

The best I can do is delete all text in the title which gives us this:
Screen Shot 2019-11-28 at 10 14 58 PM

Perhaps I am missing a way of setting this locally to get rid of the title like on the homepages for FSE on dotcom?  So I am wondering if that white gap will disappear if I had it set up properly.. 🤔

Copons:
@Addison-Stavlo thanks for looking at this!

> One thing I notice that is missing from the demo is the line under the currently visited section of the nav bar.

The Nav Menu block doesn't have that capability unfortunately. :(
I don't think it would be very hard to add, but I'm not sure at this point how valuable would be improving our block instead of Core's.

> I also notice that the menu button on mobile width is a bit wider and there is no visible text at first.

Good catch!
That's caused by the fact that the text color is forced white, which conflicts with the original theme menu background color on mobile.
About the width: that's caused by the `alignfull` width of the Nav Menu block, that is kept at all screen sizes.

> Perhaps I am missing a way of setting this locally to get rid of the title like on the homepages for FSE on dotcom?

Unless the theme overrides this (and yes, [it does](https://github.com/Automattic/themes/blob/3834391965040259a8fe546a375153b385ec59dc/morden/sass/_extra-child-theme.scss#L348-L360), and I didn't notice! 😓), the original style that hides the home on wpcom is defined in Varia and inherited by the child themes.

https://github.com/Automattic/themes/blob/3834391965040259a8fe546a375153b385ec59dc/varia/inc/style-wpcom.css#L5-L14

Whenever I remember to test for it, I usually just add those styles directly in the browser dev tools.

Addison-Stavlo:
Nice!

So with those styles I can see it looks great.  
Screen Shot 2019-11-29 at 12 03 49 PM

And with some small changes to FSE settings on the mobile menu it looks like the demo as well:
Screen Shot 2019-11-29 at 12 07 31 PM

Side question:  Is there a way to select what screen width we are editing inside the FSE or template-part editors?  To change styles on the above mobile menu I basically just squished my browser window until it showed up in the editor, but there must be some better way?

ianstewart:
image

There are a few differences between the front-end output and the theme demo styles (without the header template).

I think it's mainly …

- font-sizes for the site-title
- font-sizes for site-description
- and the padding on fse-header vs site-header

ianstewart:
Illustration:

image

apeatling:
If we can include https://github.com/Automattic/themes/pull/1693 and https://github.com/Automattic/themes/pull/1692 then add the social links block, we can get it looking like this:

Screen Shot 2019-12-03 at 1 10 19 PM

apeatling:
I've updated the header template part to include the social icons block.

apeatling:
With a rebase to include those changes in master, I think this one is ready to go.

apeatling:
With the homepage title hidden there is still a gap between the bottom of the header and the start of the content. Adjusting the padding top on main to 32px from 64px fixes this.

Screen Shot 2019-12-03 at 3 05 15 PM

noahtallen:
Hope you don't mind, @Copons, I "rebased" this on master and re-built with the latest changes.

Copons:
> with the homepage title hidden there is still a gap between the bottom of the header and the start of the content. Adjusting the padding top on main to 32px from 64px fixes this

@apeatling I'm testing the same thing locally and I think it should be ok?

To be precise, I'm manually adding the classes and styles that Varia would add on WPCOM into the browser dev tools:

https://github.com/Automattic/themes/blob/e2bd8cc15cecf65b1f2135da0d75d1cb870347ba/varia/inc/style-wpcom.css#L5-L14

- Add the `hide-homepage-title` class to the body (and `home` too in case I'm not testing the actual front page).
- Add `display: none` to the `.entry-header` element.
- Add `margin-top: 0` to the `.entry-content` element.

It looks like this to me:

Screenshot 2019-12-04 at 15 08 26

noahtallen:
I've noticed this gap appears in the page editor, but not on the front end. (on wpcom)

PR #1686, opened 2019-12-01
Lodestar: Fix image block background color issue

### LINKED ISSUE #1685: Lodestar: Remove white background for images added via image block
## Steps to replicate

Add an image to the image block and it will have white background, even though the image is transparent.

Screen Shot 2019-11-28 at 18 55 46

This doesn't happen when an image is added via URL.

But it happens again if the image block is linked to another URL. The workaround is to either add custom CSS:

.entry-content a img {
background: none;
box-shadow: none;
}

or add the link to the caption instead.

But overall, I'm not sure why a white background is added by default.

## Expected

Expected not to have a white background on transparent images added via image block.

* User report: 16976068-hc

cc @catkit

nielslange:
Fixes #1685

<table>
<tr>
<td>Before:
<br><br>

#1685-before

</td>
<td>After:
<br><br>

#1685-after

</td>
</tr>
</table>

### DIFF
--- a/lodestar/style.css
+++ b/lodestar/style.css
@@ -4,7 +4,7 @@ Theme URI: http://theme.wordpress.com/themes/lodestar
 Author: Automattic
 Author URI: https://www.wordpress.com
 Description: Lodestar is a trendy one-page theme designed with startups and small business ventures in mind.
-Version: 1.0.14-wpcom
+Version: 1.0.15-wpcom
 License: GNU General Public License v2 or later
 License URI: http://www.gnu.org/licenses/gpl-2.0.html
 Text Domain: lodestar
@@ -2718,7 +2718,7 @@ object {
 		padding: 0;
 		text-align: left;
 	}
-	
+
 	.main-navigation li {
 		display: inline-block;
 		position: relative;
--- a/lodestar/style.css
+++ b/lodestar/style.css
@@ -794,16 +794,6 @@ a:active {
 	border-bottom-color: transparent;
 }
 
-.entry-content a img,
-.widget a img {
-    background: #fff;
-    box-shadow: 0 0 0 6px #fff;
-}
-
-.site-footer .widget-area a img {
-	box-shadow: 0 0 0 6px #eee;
-}
-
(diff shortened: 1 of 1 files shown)


davoraltman:
LGTM!

One thing though - for some reason, I can't reproduce the original issue anymore. Looks like this was removed even though this PR wasn't merged?

nielslange:
> LGTM!
> 
> One thing though - for some reason, I can't reproduce the original issue anymore. Looks like this was removed even though this PR wasn't merged?

@davoraltman Have you linked the image? The issue only appears for linked images, not for unlinked images.

davoraltman:
@nielslange In my defense, it was really late when I tried this yesterday 🤦‍♂ 

Reproduced it and checked your fix, works like a charm :)

michaeldcain:
Although the patch functionally works, I wonder if this was an intentional design decision, and the transparent background issue is more an unfortunate side-effect. Personally, I'd prefer a non "destructive" patch in situations like this for live themes, where the issue might not be an explicit defect. A workaround could be a `.no-frame` CSS rule that removes the background and box shadow, and then have the user add the class as needed for transparent images.

I've added a "Needs Theam Reply" label to get @Automattic/theam to weigh in on this.

nielslange:
> Although the patch functionally works, I wonder if this was an intentional design decision, and the transparent background issue is more an unfortunate side-effect. Personally, I'd prefer a non "destructive" patch in situations like this for live themes, where the issue might not be an explicit defect. A workaround could be a `.no-frame` CSS rule that removes the background and box shadow, and then have the user add the class as needed for transparent images.
> 
> I've added a "Needs Theam Reply" label to get @Automattic/theam to weigh in on this.

Personally, I don't see how this issue could be based on an intentional design decision. That said, let's see what the @Automattic/theam thinks on this issue. 😀

pbking (approved):
Change works and makes sense.  I also cannot discern why the original author would have intended for the white background.
👍

scruffian:
Deployed in r57303-wpcom-themes

PR #1690, opened 2019-12-03
Hever: Add FSE support. 

apeatling:
Add CSS column and block margin changes to Varia so that columns can be used to layout header and footer template parts more accurately.

Screen Shot 2019-12-03 at 9 52 04 AM

To test:

1. Run a copy of WordPress locally with the Full Site Editing plugin installed, and Gutenberg plugin installed.

2. Install the hever and varia theme from this PR

3. Activate hever, and the two plugins

4. Create a new page and use the Hever homepage template, confirm that you see the header template as it appears in the visual above. Confirm this works correctly on the front end.

### DIFF
--- a/hever/package-lock.json
+++ b/hever/package-lock.json
@@ -2998,8 +2998,7 @@
             "ansi-regex": {
               "version": "2.1.1",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "aproba": {
               "version": "1.2.0",
@@ -3020,14 +3019,12 @@
             "balanced-match": {
               "version": "1.0.0",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "brace-expansion": {
               "version": "1.1.11",
               "bundled": true,
               "dev": true,
-              "optional": true,
               "requires": {
                 "balanced-match": "^1.0.0",
                 "concat-map": "0.0.1"
@@ -3042,20 +3039,17 @@
             "code-point-at": {
               "version": "1.1.0",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "concat-map": {
               "version": "0.0.1",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "console-control-strings": {
               "version": "1.1.0",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "core-util-is": {
               "version": "1.0.2",
@@ -3172,8 +3166,7 @@
             "inherits": {
               "version": "2.0.3",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "ini": {
               "version": "1.3.5",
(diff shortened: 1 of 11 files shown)


ianstewart:
Some minor front-end issues with Hever …

image

apeatling:
@ianstewart This should be fixed now, but to test you'll need to remove the `hever-fse-template-data-v1` entry from wp_options, then deactivate and reactivate hever.

apeatling:
I think this is ready to deploy now pending review.

Copons:
I gave this PR a very quick spin on Maywood, by modifying the header to have a column block, with logo, title, and description on the left and the nav menu on the right (a common use case in other themes).

First thing first: I've run `npm run build` and for some reasons I've built some additional changes from the 2 Varia PRs. I think we should re-run it on all affected themes just to be sure. (a rather common thing... this is not the case, but sometimes the RTL style updates only happens after the second `run build`).

Then, this is how the Maywood header looks like on this PR:

| Template | Page | Front end |
| - | - | - |
| Screenshot 2019-12-04 at 00 26 40 | Screenshot 2019-12-04 at 00 26 58 | Screenshot 2019-12-04 at 00 27 11 |

There are some obvious things that I'm noticing:

1. With #1693 we are basically removing the block margins inside the columns block on the front end, while they still keep the usual margins in the editor. This is not necessarily an issue, but just something to keep in mind.
To keep as much parity as possible, I've also intentionally tried to have the same editor wider margins in the front end too, because there isn't any good way to reduce them in the editor.
But this is just my opinion, and I don't have any strong feelings about it. 🙂 

2. With #1692 we are adding an additional top margin top any blocks inside the template editor. For some reasons this only applies to the _first_ `[data-block]`, while the subsequent ones' margins are overridden by Gutenberg. I don't know why off the top of my head...

Screenshot 2019-12-04 at 00 36 05

3. The columns content in the template editor is a few pixel _above_ the frame, while in the page editor is a few pixel under.
In the screenshot is kinda visible with the logo, but it's also noticeable when the columns block outline appears on focus.

Screenshot 2019-12-04 at 00 40 00

Copons:
This looks legit @apeatling! Thanks for taking care of this so quickly and with such dedication 🙇 

Aside the things I've pointed out above, I think there's only one more inconsistency that I discovered on Alves that we need to sort out.
Apparently, Hever and Alves are basically the same theme, and they both define a non-100% width to `#masthead`.
This could be solved long term by updating the Varia `header.php` file to NOT use that ID (which would also entail making sure there are no regressions on other FSE themes), or, as I've done with Alves, a quicker and dirtier solution of only defining the `#masthead` styles when FSE is not enabled (`body:not(.fse-enabled)`):

This is how I changed Alves:
https://github.com/Automattic/themes/blob/f7d45d11194cddbf891bf251213d3e417d769e3f/alves/sass/_extra-child-theme.scss#L149-L162

This is where that selector should be applied to Hever (notice that it might affect the footer as well):
https://github.com/Automattic/themes/blob/f7d45d11194cddbf891bf251213d3e417d769e3f/hever/sass/_extra-child-theme.scss#L25-L40

---

Last but not least: I've mentioned it above, but I'll mention it again.
We need to `npm run build` all affected themes at least a couple of times! 😄
I've only run this on Maywood and Hever, and both output more compiled styles!

PR #1691, opened 2019-12-03
Varia: Update margins for better header design support

apeatling:
This change updates block margins to that header and footer template parts render close or identical to how they render when viewed when editing a page.

This will improve consistency when using full site editing, and also allow us to compress excess margins on blocks like the social icons block.

The main thing to test is Maywood, since this is the only live FSE theme. Apply this change to Varia, and then test Maywood to make sure that the header renders correctly when editing a page, and also renders correctly when editing the header template part.

### DIFF
--- a/alves/footer.php
+++ b/alves/footer.php
@@ -15,6 +15,12 @@
 
 	</div><!-- #content -->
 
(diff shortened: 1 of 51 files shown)


apeatling:
Mixed up different changes here, closing this and starting a new PR.

PR #1702, opened 2019-12-10
Maywood: Fix strong-italic font-weight issue

### LINKED ISSUE #1700: Maywood: Paragraph text can't be both bold and italicized
## Steps to replicate

1. Highlight some text in a paragraph block on the Maywood theme.
2. Click the Bold option in the toolbar (text turns bold).
3. Click the Italics option in the toolbar (text loses bolding but shows italicized).

It has the same issue the other was as well.

1. Highlight some text in a paragraph block on the Maywood theme.
2. Click the Italics option in the toolbar (text is italicized).
3. Click the Bold option in the toolbar (text stays italicized, no bolding).

## Result

The text in both the editor and on the front end shows the italics, but not the bolding:

The HTML `<strong>` tag is there, so it seems to be an issue with the styling:

## Expected

For the text to be both bold and italicized.

User Report:
17153077-hc

nielslange:
Fixes #1700 

The issue is caused by https://github.com/Automattic/themes/blob/86ea415132f62d9893d4404b4f7ac3d789486cfc/varia/sass/base/_normalize.scss#L92-L99 and does not take into consideration that when `bold/strong` and `italic/emphasis` are combined, the bolder font-weight is no longer visible. To fix this issue, I've added additional statements to correct the font-weight in when these elements are combined.

Other than in #1700 reported, this issue affects both Varia and it's child-themes and potentially all other themes that use the `_normalize.scss` file.

<table>
<tr>
<td>Before:
<br><br>

#1700-editor-before

</td>
<td>After:
<br><br>

#1700-editor-after

</td>
</tr>
</table>

<table>
<tr>
<td>Before:
<br><br>

#1700-site-before

</td>
<td>After:
<br><br>

#1700-site-after

</td>
</tr>
</table>

### DIFF
--- a/alves/functions.php
+++ b/alves/functions.php
@@ -132,7 +132,7 @@ add_action( 'widgets_init', 'alves_widgets_init', 12 );
  * Filter the content_width in pixels, based on the child-theme's design and stylesheet.
  */
 function alves_content_width() {
-	return 700;
+	return 750;
 }
 add_filter( 'varia_content_width', 'alves_content_width' );
 
--- a/alves/package-lock.json
+++ b/alves/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "alves",
-  "version": "1.4.0",
+  "version": "1.4.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -2998,7 +2998,8 @@
             "ansi-regex": {
               "version": "2.1.1",
               "bundled": true,
-              "dev": true
+              "dev": true,
+              "optional": true
             },
             "aproba": {
               "version": "1.2.0",
@@ -3019,12 +3020,14 @@
             "balanced-match": {
               "version": "1.0.0",
               "bundled": true,
-              "dev": true
+              "dev": true,
+              "optional": true
             },
             "brace-expansion": {
               "version": "1.1.11",
               "bundled": true,
               "dev": true,
+              "optional": true,
               "requires": {
                 "balanced-match": "^1.0.0",
                 "concat-map": "0.0.1"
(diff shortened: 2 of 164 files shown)


michaeldcain:
I'm unable to reproduce this in Varia:

Screen Shot 2019-12-19 at 12 49 09 PM

As reported, it only seems to be a bug in Varia child themes like Maywood, that declare a lighter than `400` body font-weight. The browser then renders `font-weight: bolder` as `400`, which isn't enqueued in the theme's functions, so the browser renders the closest enqueued weight.

@allancole @iamtakashi: should we be using `font-weight: bold` instead of `bolder` in the normalize rule or fixing this on a per-theme basis?

michaeldcain:
> should we be using font-weight: bold instead of bolder in the normalize rule or fixing this on a per-theme basis?

I think I'd advocate for fixing this on a per-theme basis. Using `font-weight: bold` removes the subtlety of normalize.css, and using `font-weight: 700` does the same (and additionally introduces a similar enqueued font issue).

I believe that Maywood is the only TFT that uses a body weight under `300`. @nielslange: would you be able to patch it in `maywood/sass/_extra-child-theme.scss`?

nielslange:
> I believe that Maywood is the only TFT that uses a body weight under `300`. @nielslange: would you be able to patch it in `maywood/sass/_extra-child-theme.scss`?

@michaeldcain Sure. I'll update my patch accordingly.

iamtakashi:
I tested the fix in Maywood, but the issue still remains in the editor. Can we fix this in the editor also?

nielslange:
> I tested the fix in Maywood, but the issue still remains in the editor. Can we fix this in the editor also?

@iamtakashi I must have missed that and just updated the editor styles.

michaeldcain:
@nielslange: thanks for the fix! 

I went ahead and removed the paragraph specificity, so that this will work in any instance of bolded+italicized text. On closer inspection, I also noticed that Maywood uses the `500` weight of IBM Plex Sans for its bold weight, and we weren't enqueuing the italicized variant of the font (`500i`), so I made that switch too.

Front-end | Editor
------------ | -------------
Screen Shot 2019-12-23 at 1 32 38 PM | Screen Shot 2019-12-23 at 1 32 22 PM

michaeldcain:
Clicked the wrong button 😬

michaeldcain:
This has been deployed in 54964-wpcom-themes

PR #1709, opened 2019-12-16
Add ‘wp_body_open’ hook.

zzap:
Starting from WordPress 5.2 a new function is added – `wp_body_open()` – that is used to trigger a `wp_body_open` action. The function should be placed inside of the <body> tag immediately after it is opened.


michaeldcain:
Deployed to WP.com in r55056-wpcom-themes

Issue #1708, opened 2019-12-16
Varia: Entry meta in Blog Posts block stays in grey with background colour


iamtakashi:
## Steps to replicate

* Add a Blog posts block to display some posts.
* Add a background colour to a container block.
* Notice the colour of entry meta on the front-end.

<img width="605" alt="Screenshot 2019-12-16 at 14 15 45" src="https://user-images.githubusercontent.com/908665/70915979-c8e25e00-2012-11ea-92a5-b3933b58383a.png">

All the child themes likely share the issue, and they should also be updated.

michaeldcain:
Fixed in https://github.com/Automattic/themes/pull/1710

PR #1710, opened 2019-12-16
Template-First Themes: add styling for Blog Posts (Newspack) block.

michaeldcain:
The Newspack Blog Posts block has been ported to WordPress.com, and is now the recommended upgrade from the previous a8c-posts-list block. The block comes with a number of default styles, which need to be overwritten to allow the template-first theme blog styling to work.

Theme | Old Block | New Block (before) | New Block (after
------------ | ------------ | ------------- | -------------
Varia | old-block | new-block-before | new-block-after
Barnsbury | barnsbury-old-block | barnsbury-new-block-before | barnsbury-new-block-after

ref: p7DVsv-7Rw-p2
additionally fixes https://github.com/Automattic/wp-calypso/issues/38389

### DIFF
--- a/varia/sass/blocks/blog-posts/_style.scss
+++ b/varia/sass/blocks/blog-posts/_style.scss
@@ -25,6 +25,12 @@
 			&:hover {
 				color: #{map-deep-get($config-global, "color", "primary", "hover")};
 			}
+
+			.has-background:not(.has-background-background-color) &,
+			[class*="background-color"]:not(.has-background-background-color) &,
+			[style*="background-color"] & {
+				color: currentColor;
+			}
 		}
 	}
 
@@ -51,6 +57,12 @@
 	.cat-links {
 		color: #{map-deep-get($config-global, "color", "foreground", "light")};
 		font-size: #{map-deep-get($config-global, "font", "size", "sm")};
+
+		.has-background:not(.has-background-background-color) &,
+		[class*="background-color"]:not(.has-background-background-color) &,
+		[style*="background-color"] & {
+			color: currentColor;
+		}
 
 		> span {
 			display: inline-block;
@@ -98,4 +110,12 @@ button[data-load-more-btn] {
 	// Extend button style
 	@extend %button-style;
 	display: inline-block;
+
+	.has-background:not(.has-background-background-color) &,
+	[class*="background-color"]:not(.has-background-background-color) &,
+	[style*="background-color"] & {
+		background-color: transparent;
+		border: 2px solid currentColor;
+		color: currentColor;
+	}
 }
--- a/varia/sass/blocks/blog-posts/_style.scss
+++ b/varia/sass/blocks/blog-posts/_style.scss
@@ -1,0 +1,100 @@
+.wp-block-newspack-blocks-homepage-articles article {
+	display: block;
+
+	/* Vertical margins logic between posts */
+	margin-top: calc(3 * #{map-deep-get($config-global, "spacing", "vertical")});
+	margin-bottom: calc(3 * #{map-deep-get($config-global, "spacing", "vertical")});
+
+	&:first-child {
+		margin-top: 0;
+	}
+
+	&:last-child {
+		margin-bottom: calc(3 * #{map-deep-get($config-global, "spacing", "vertical")});
+	}
+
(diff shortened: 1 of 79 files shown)


iamtakashi:
Awesome work, @michaeldcain !

I understand that it'd be ideal for the existing users if the styles match between the old and the new blocks, I'm not sure if we should override the heading text alignment in the theme like Maywood that has a centre-aligned title. It's the sort of thing that the user should be able to control in the block, and when the block adds the text alignment setting, that wouldn't work if the theme overrides it in the CSS.

Another thing that I've noticed is Type scale setting in the block. We'd also need to specify each size (1- 10, the default is 4). At the moment in Varia, the default size 4 is larger than size 5 that comes from the block. 

Screeny Video 17 Dec 2019 at 11 05 56 2019-12-17 11_09_52

I'll defer to @allancole 's decision on these, but I thought it'd be worth to consider these details.

michaeldcain:
> I'm not sure if we should override the heading text alignment in the theme like Maywood that has a centre-aligned title. It's the sort of thing that the user should be able to control in the block, and when the block adds the text alignment setting, that wouldn't work if the theme overrides it in the CSS.

Yeah, I'm not sure if there's a plan to extend the Newspack block (or any reason why we can't add alignment controls) - @obenland might know - so I opted to do the styling in the themes, to at least give some visual variation. I'm happy to undo it though.

/cc @ianstewart 

> Another thing that I've noticed is Type scale setting in the block. We'd also need to specify each size (1- 10, the default is 4). At the moment in Varia, the default size 4 is larger than size 5 that comes from the block.

Let me undo the font-size changes and see if that works.

michaeldcain:
Takashi created https://github.com/Automattic/wp-calypso/issues/38470 to deal with the alignment controls.

ianstewart:
> when the block adds the text alignment setting

Are there plans for this?

obenland:
> Yeah, I'm not sure if there's a plan to extend the Newspack block (or any reason why we can't add alignment controls) - @obenland might know 

There is no plan on Ajax' end to extend it at this time. Equally though, there is no reason why it couldn't be done.

michaeldcain:
I've updated the themes to remove alignment and text size changes. @Automattic/theam it'd be great to get a review on this so we can get it merged soon.

allancole:
This looks really good @michaeldcain :-) I tested it with various configurations of Varia child-themes, and the Blog Posts Block options to see if anything was off and all seemed to check out. The code looks good as well. 

I think this is good to merge :-)

iamtakashi:
Nice update, @michaeldcain ! 

One thing I've noticed is that the different bottom margin for each article when they are in the grid mode. On the front of the site, the margin is 1em that's set by the block, but in the editor, that's 96px that's overridden by the theme. I think the margin should be the same between in the editor and the front of the site.

**Editor**
Screenshot 2019-12-18 at 11 20 12

**The front of the site**
Screenshot 2019-12-18 at 11 20 19

iamtakashi:
Would it be possible to fix [this issue](https://github.com/Automattic/themes/issues/1708) as well? 

As soon as we specify the colour of an element, we need to change it to `currentColor` so that it's always visible to avoid this situation — the blue title is invisible on a blue background, and the grey entry meta never changes its colour.

Screenshot 2019-12-18 at 13 43 10

It can be easily done by something like this: https://github.com/Automattic/themes/blob/master/varia/sass/blocks/quote/_style.scss#L34

varia/sass/blocks/blog-posts/_style.scss
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	.more-link {
+		margin-top: #{map-deep-get($config-global, "spacing", "unit")};
+	}
+
+	.entry-meta,
+	.entry-footer,
+	.cat-links {
+		color: #{map-deep-get($config-global, "color", "foreground", "light")};
```
michaeldcain: @iamtakashi for the `.entry-meta` issue, can I just use `currentColor` here, or does it require the `has-background-color` class logic?

michaeldcain:
2b63d8b fixes #1710 1708

michaeldcain:
2b63d8b fixes #1708

PR #1714, opened 2019-12-18
Convert line-height and font-size into CSS custom properties

nosolosw:
This PR adds support for tweaking the line-height and font-size using Global Styles. Goes with https://github.com/Automattic/wp-calypso/pull/38754

2001-global-styles-font-settings-14-1218

**Some thoughts:**

- I've found a few instances where invalid values were used and changed them to use the body line-height instead (see comments on code).
- There are a few other places that use a number instead of a SASS variable. I've left those untouched, but I wonder whether we need to modify them to react to user choices.
- I was thinking that, for release, we should also add font sizes as a minimum and potentially letter-spacing as well. What do you think? Is tweaking line-height useful enough?

#### Testing instructions

* Follow instructions on https://github.com/Automattic/wp-calypso/pull/38754
* From the root directory, run `cd varia && npm install && npm run build && npm run build` (twice, so rtl stylesheet compile as well).
* Copy the `varia` dir into the `wp-content/themes` directory of a WordPress instance.
* Switch to varia as the theme active and open a post.
* Open Global Styles and interact with the controls in the "Font Settings" panel. Publish the changes.
* Publish the post and check that the front-end also has the changes.


varia/print.scss
```diff
@@ -34,7 +34,6 @@ Andreas Hecht in https://www.jotform.com/blog/css-perfect-print-stylesheet-98272
   /* Fonts */
 
   body {
-    font: 13pt Georgia, "Times New Roman", Times, serif; // Fallback for browsers without CSS variables support.
```
nosolosw: The postcss-custom-properties plugin will take care of adding the fallback rules for us.

varia/sass/abstracts/_mixins.scss
```diff
@@ -97,6 +97,5 @@
  * It also adds the proper fallback for browsers without support.
  */
 @mixin font-family( $config_map ) {
-	font-family: #{map-deep-get( $config_map, 'fallback' )}; // For browsers without CSS custom properties support.
```
nosolosw: The postcss-custom-properties plugin will take care of generating the fallback rules for us. One thing I'd like to do is refactoring how this works and remove the mixin. This is perhaps best accomplished in a subsequent PR as to keep the focus on line-heights in this one.

varia/sass/blocks/heading/_config.scss
```diff
@@ -19,15 +19,6 @@ $config-heading: (
 			"h1": map-deep-get($config-global, "font", "size", "xxxxl"),
 		),
 		// Letter spacing
-		"line-height": (
```
nosolosw: All themes use the heading `line-height` for every `h*` element, so I thought we could remove these mappings. If we wanted to keep these variables, it's still possible, although they should be provided by Global Styles.

varia/sass/blocks/image/_style.scss
```diff
@@ -4,7 +4,7 @@
 	figcaption {
 		color: #{map-deep-get($config-global, "color", "foreground", "light")};
 		font-size: #{map-deep-get($config-global, "font", "size", "xs")};
-		line-height: #{map-deep-get($config-global, "font", "line-height", "xs")};
```
nosolosw: I couldn't find the `line-height.xs` mapping anywhere, so I changed it to `--line-height-body`. Is it ok for this instance?

varia/sass/child-theme/_config-child-theme-deep.scss
```diff
@@ -228,15 +228,6 @@ $config-heading: (
 			"h1": map-deep-get($config-global, "font", "size", "xxxxl"),
 		),
 		// Letter spacing
-		"line-height": (
```
nosolosw: See comment at https://github.com/Automattic/themes/pull/1714/files#r359437116

varia/sass/elements/_media.scss
```diff
@@ -5,7 +5,7 @@ figure {
 figcaption {
 	color: #{map-deep-get($config-global, "color", "foreground", "light")};
 	font-size: #{map-deep-get($config-global, "font", "size", "xs")};
-	line-height: #{map-deep-get($config-global, "font", "line-height", "xs")};
```
nosolosw: I couldn't find the `line-height.xs` mapping anywhere, so I changed it to `--line-height-body`. Is it ok for this instance?

varia/sass/child-theme/_config-child-theme-deep.scss
```diff
@@ -152,7 +152,7 @@ $config-elements: (
 				"fallback": map-deep-get($config-global, "font", "family", "secondary", "fallback"),
 				"css-var": map-deep-get($config-global, "font", "family", "secondary", "css-var"),
 			),
-			"line-height": map-deep-get($config-global, "font", "line-height", "md"),
```
nosolosw: `$config-global` doesn't have this value, so I've changed it to `body` instead. Is that ok in this instance?

varia/sass/child-theme/_config-child-theme-deep.scss
```diff
@@ -390,6 +381,6 @@ $config-footer: (
 			"css-var": map-deep-get($config-global, "font", "family", "primary", "css-var"),
 		),
 		"size": map-deep-get($config-global, "font", "size", "sm"),
-		"line-height": map-deep-get($config-global, "font", "line-height", "sm"),
```
nosolosw: `$config-global` doesn't have this value, so I've changed it to `body` instead. Is that ok in this instance?

varia/sass/components/footer/_config.scss
```diff
@@ -12,6 +12,6 @@ $config-footer: (
 			"css-var": map-deep-get($config-global, "font", "family", "primary", "css-var"),
 		),
 		"size": map-deep-get($config-global, "font", "size", "sm"),
-		"line-height": map-deep-get($config-global, "font", "line-height", "sm"),
```
nosolosw: `$config-global` doesn't have this value, so I've changed it to `body` instead. Is that ok in this instance?

varia/sass/elements/_config.scss
```diff
@@ -15,7 +15,7 @@ $config-elements: (
 				"fallback": map-deep-get($config-global, "font", "family", "secondary", "fallback"),
 				"css-var": map-deep-get($config-global, "font", "family", "secondary", "css-var"),
 			),
-			"line-height": map-deep-get($config-global, "font", "line-height", "md"),
```
nosolosw: `$config-global` doesn't have this value, so I've changed it to `body` instead. Is that ok in this instance?

varia/sass/abstracts/_mixins.scss
allancole: Interesting. Originally, I made Varia (previously called `_dsgnsystm`) with _only_ CSS-variables (see [here](https://github.com/allancole/_dsgnsystm/tree/master/_dsgnsystm-theme)). 

Are you basically suggesting we go back to that? I’d definitely prefer that direction 👍

varia/sass/blocks/heading/_config.scss
allancole: Interesting. Didn’t realize that but if they aren’t being used anywhere, happy to have these removed :-)

varia/sass/blocks/image/_style.scss
allancole: Hmmm, that’s probably a typo. `--line-height-body` should be fine although, I think `--line-height-heading` might actually look nicer since it’s a caption and not regular paragraph.

varia/sass/abstracts/_mixins.scss
allancole: Wait, I just answered my own question. Nevermind ;-)

varia/sass/elements/_media.scss
allancole: As with the other instance of this, I think `--line-height-heading` is best for captions so they take up less space.

varia/sass/abstracts/_mixins.scss
nosolosw: Go back to the roots! Yeah, I think this is clearer than the mixin approach. Although I still need to figure out how the overrides would work (global configs VS blocks/components configs), it is promising.

varia/sass/blocks/image/_style.scss
nosolosw: Done at 0a48e84e4ec57ba50ea3121259b77a780113bd77

varia/sass/elements/_media.scss
nosolosw: Done at b185f0b55e9361c0bdb70e5f3dcdd91223bc4306

PR #1718, opened 2019-12-19
Varia: only inherit top/bottom margin with cover images.

### LINKED ISSUE #1670: Varia: Inherited margins around cover block
@allancole Do you remember why we have these inherited margins?https://github.com/Automattic/themes/blob/master/varia/sass/blocks/cover/_style.scss#L6

This adds an extra side margin to the cover block when it's inserted in a column cell that has a side margin. Can we remove it?

Screenshot 2019-11-18 at 14 35 27

michaeldcain:
Fixes #1670


iamtakashi:
LGTM 👍

PR #1716, opened 2019-12-18
Template-First Themes: Blog Posts block style updates

michaeldcain:
Initial theme styles were added for the new Blog Posts (Newspack) block in #1710. This additionally fixes some layout issues and bugs.

**Included:**
- fixes: https://github.com/Automattic/wp-calypso/issues/38389
- rebuilds child themes from: #1704

PR #1721, opened 2019-12-20
Template-First Themes: recompile from master

michaeldcain:
This bumps the version numbers and recompiles the template-first themes after: #1716, #1718, #1660


michaeldcain:
I've tested these in each of the TFTs. Since the PRs were approved, I'm going to merge to keep things in sync.

Issue #1687, opened 2019-12-02
Redhill: Global styles not working properly


sandymcfadden:
## Steps to replicate

- Use Redhill theme
- Edit a page
- Edit fonts with global styles option

Here is a screen recording of the changes being made
https://cloudup.com/c98WS0r3Rzc

## Result

Changing the header font doesn't change all header options, changes some body fonts.
Chaging the base font changes some header options.

## Expected

For the heading option to change only the heading fonts and the base font to change the rest.

* User report: 17020773-hc
* Follow-up ticket: 2531525-zd

nosolosw:
I'm preparing a fix for this at https://github.com/Automattic/themes/pull/1731

zdenys:
This is still an issue. I remember I noticed this oddity on other themes but was not able to reproduce consistently (I have tried all Global Styles Themes just now, but no luck, so I will keep looking out for this).

Currently, I can reproduce consistently with Redhill on a Simple Site:

- `Heading Font` changes font of Page Title, Paragraph Block, Quote
- `Base Font` changes font of Heading Block

I also noticed this issue in 2834861-zen during a Quick Start session.

PR #1731, opened 2019-12-30
Redhill: fix body/headings fonts mapping

nosolosw:
This PR fixes https://github.com/Automattic/themes/issues/1687
Corresponding wpcom patch https://code.a8c.com/D37152

The body and lists were set as `font-heading` (see the [corresponding](https://github.com/Automattic/themes/pull/1731/commits/c56ffcd08f9f4117df29b8947958356c4a0a7476) [commit](https://github.com/Automattic/themes/pull/1731/commits/d573708934dd63d8266bfeb2e14ad11397208cfb)) while `h*` tags were set as `font-base` (see the [corresponding commit](https://github.com/Automattic/themes/pull/1731/commits/e2c6bb96a9b2fe2087a2f9858300b58477f0b833)).

A few notes:

- While this change fixes the issue, it also inverts which font is displayed when the user selects _Theme Font_.
  - `Palatino, Palatino Linotype, Palatino LT STD, Book Antiqua, Times, Times New Roman, serif` was used for body and lists, and now is used for `h*`.
  - `-apple-system, BlinkMacSystemFont, Segoe UI, Roboto, Oxygen, Ubuntu, Cantarell, Fira Sans, Droid Sans, Helvetica Neue, sans-serif` was used for `h*` and now it's used for body and lists.

- There are a few items that still use the `primary`/`font-heading` fonts, notably: config-elements.form.font.family, config-pullquote.font.family, config-quote.font.family. I've left those untouched as those may be considered stylistic choices.

### DIFF
--- a/redhill/sass/_config-child-theme-deep.scss
+++ b/redhill/sass/_config-child-theme-deep.scss
@@ -266,8 +266,8 @@ $config-list: (
 	// Fonts
 	"font": (
 		"family": (
-			"fallback": map-deep-get($config-global, "font", "family", "primary", "fallback"),
-			"css-var": map-deep-get($config-global, "font", "family", "primary", "css-var"),
+			"fallback": map-deep-get($config-global, "font", "family", "secondary", "fallback"),
+			"css-var": map-deep-get($config-global, "font", "family", "secondary", "css-var"),
 		),
 	),
 );
--- a/redhill/sass/_config-child-theme-deep.scss
+++ b/redhill/sass/_config-child-theme-deep.scss
@@ -224,8 +224,8 @@ $config-heading: (
 	"font": (
 		// Family
 		"family": (
-			"fallback": map-deep-get($config-global, "font", "family", "secondary", "fallback"),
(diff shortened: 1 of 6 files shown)


iamtakashi:
This theme is confusing because the original theme (Radcliffe 2) had a bit unique mapping and doesn't follow the usual "heading and base" pattern.

In this theme, the primary font is Palatino, and the secondary font is the system font.

Screenshot 2020-01-06 at 16 19 59

What I remember from the [conversation](https://github.com/Automattic/themes/pull/1516#issuecomment-542755620) was that Global Styles should keep the theme's font mapping intact. And with this change, the original font mapping will break.

Screenshot 2020-01-06 at 16 39 53

I wonder if the labels in Global Styles should be _primary_ and _secondary_ so they are not too specific to the kind of elements.

nosolosw:
Let's close this and try a different approach. I'll post the details shortly.

nosolosw:
Put together an alternative to this at https://github.com/Automattic/themes/pull/1744

It doesn't modify Redhill style choices -- at least, the blocks that I've tried, would welcome a more knowledgable pair of eyes.

PR #1735, opened 2020-01-04
Varia: Fix iframe width issue

### LINKED ISSUE #1732: Varia: iframes with style attributes have a max-width of 32px when window size is at least 560px wide
## Steps to replicate
Embed a Google Calendar on an AT site with a Varia-type theme as the active theme and a window size of 560px wide or greater. Below is Stow; I also tried on Dalston, Maywood, Shawburn, and Mayland.

image

This is the theme code in question in Stow:
```
@media only screen and (min-width: 560px) {
	.entry-content > iframe[style] {
		max-width: 32px !important;
	}
}
```

A specific override for Google Calendar is as follows:
```
@media only screen and (min-width: 560px) {
  .entry-content > iframe[src^="https://calendar.google.com"] {
    max-width: unset !important;
  }
}
}
```

* User report: 17642074-hc
* Follow-up ticket: 2600091-zen

nielslange:
Fixes #1732 

<table>
<tr>
<td>Before:
<br><br>

#1732-before

</td>
<td>After:
<br><br>

#1732-after

</td>
</tr>
</table>

### DIFF
--- a/varia/sass/components/content/_entry-content.scss
+++ b/varia/sass/components/content/_entry-content.scss
@@ -40,7 +40,12 @@
 	// Overwrite iframe embeds that have inline styles.
 	> iframe[style] {
 
-		display: block;
+		margin: #{map-deep-get($config-global, "spacing", "vertical")} 0 !important;
+		max-width: 100% !important;
+
+		@include media(mobile) {
+			max-width: #{map-deep-get($config-global, "spacing", "measure")} !important;
+		}
 
 	}
 
--- a/varia/sass/components/content/_entry-content.scss
+++ b/varia/sass/components/content/_entry-content.scss
@@ -43,9 +43,6 @@
 		margin: #{map-deep-get($config-global, "spacing", "vertical")} 0 !important;
 		max-width: 100% !important;
 
-		@include media(mobile) {
-			max-width: #{map-deep-get($config-global, "spacing", "vertical")} !important;
-		}
 	}
 
 	// Page links
--- a/varia/sass/components/content/_entry-content.scss
+++ b/varia/sass/components/content/_entry-content.scss
@@ -40,8 +40,7 @@
 	// Overwrite iframe embeds that have inline styles.
 	> iframe[style] {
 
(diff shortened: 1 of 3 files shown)


danieldudzic:
@allancole I have done some investigating, and the bug seems to be present since the first commit.

```
// Overwrite iframe embeds that have inline styles.
	> iframe[style] {

		margin: #{map-deep-get($config-global, "spacing", "vertical")} 0 !important;
		max-width: 100% !important;

		@include media(mobile) {
			max-width: #{map-deep-get($config-global, "spacing", "vertical")} !important;
		}
	}
```

I think [@tbradsha's theory](https://github.com/Automattic/themes/issues/1732#issuecomment-570812320) sounds plausible - however since I haven't done much work on Varia recently, I think it's best if you greenlight this fix.

allancole:
I think replacing the max-width rules with `display: block;` may negatively affect the responsive behavior here. [@tbradsha's theory](https://github.com/Automattic/themes/issues/1732#issuecomment-570812320) is close but its actually the variable thats wrong, not the parameter :-). 

The solution should look like this:
```
// Overwrite iframe embeds that have inline styles.
	> iframe[style] {

		margin: #{map-deep-get($config-global, "spacing", "vertical")} 0 !important;
		max-width: 100% !important;

		@include media(mobile) {
			max-width: #{map-deep-get($config-global, "spacing", "measure")} !important;
		}
	}
```

That’ll output `max-width: unset` which _should_ fix this without effecting the responsiveness.

nielslange:
@allancole As promised in Slack yesterday, I've updated this PR based on your feedback in https://github.com/Automattic/themes/pull/1735#issuecomment-593470957.

allancole:
@nielslange, on second thought, my suggestion was a bit incomplete the other day. Looks like I forgot to consider how the responsive refactor from a while ago might effect these iframe. I’m gonna work out a separate PR that fixes this properly and apply it to the all the child themes 👍.

nielslange:
@allancole Sounds good. Please go ahead with a separate PR. 😉

PR #1744, opened 2020-01-07
Redhill: Switch base/headings css variable so user expectations are met

oandregal:
* Corresponding wpcom patch https://code.a8c.com/D37361
* Alternative to https://github.com/Automattic/themes/pull/1731

This PR addresses https://github.com/Automattic/themes/issues/1687 while maintaining Redhill style choices. I've tried with a post containing paragraph, quote, list, and button blocks.

Note, however, that Redhill makes some style choices (post title uses the same font than base text, not that headings) that some users might find unexpected. I believe this is ok but wanted to surface it anyway. The only solution to that would be either 1) converge the themes into some kind of defaults as to what base/headings mean or 2) adding a mechanism for the user to change the mapping at the block level -- ex: in this case the user could say "Let's change Post Title to use font headings", which is doable, but not in the near-term roadmap.


oandregal:
It's been a while since this PR, so I'm closing it.

PR #1756, opened 2020-01-20
Shoreditch: Remove box shadow from image link

### LINKED ISSUE #1752: Shoreditch: border-bottom for a text link appears even for image links
## Steps to replicate

- Activate Shoreditch theme on your site.
- Create/edit a post/page and add a transparent image with a link

## Result

There's a border-bottom for the link, showing up as a line overlapping with the image.
SWCywbxm

## Expected

No border-bottom since there's no text.

* User report: [#17972805-hc]

If the image has an opaque background, the border-bottom gets hidden behind it. This is the workaround I've shared for this particular user. Folks with a Premium Plan can probably use a CSS workaround as well.

nielslange:
Fixes #1752

<table>
<tr>
<td>Before:
<br><br>

#1752-before

</td>
<td>After:
<br><br>

#1752-after

</td>
</tr>
</table>

### DIFF
--- a/shoreditch/style.css
+++ b/shoreditch/style.css
@@ -1069,10 +1069,10 @@ a:hover {
 	text-decoration: underline;
 }
 
-.comment-content a.no-shadow,
-.entry-content a.no-shadow,
-.entry-summary a.no-shadow {
-	box-shadow: none;
+.comment-content a:hover,
+.entry-content a:hover,
+.entry-summary a:hover {
+	text-decoration: none;
 }
 
 .comment-content a:focus,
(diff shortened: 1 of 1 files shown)


nielslange:
> Please add a fix in style.css for regular (non-block) images. This should also fix the issue with block images.

@danieldudzic I'm afraid, this will not work based on the underlying HTML structure:

```
<figure class="wp-block-image size-large">
  <a href="...">
    <img src="...">
  </a>
</figure>
```

The child element `img` should affect the CSS of its parent element `a`, which is not possible with CSS3. It seems that this issue might be addressed with CSS4, but at the moment I'm not able to find a solution that works for all browsers. 

Would you mind looking up this issue to see if you can come up with a solution that addresses the regular (non-block) images?

danieldudzic:
I have consulted with the theme author.

`box-shadow: 0 1px 0 0;` is the culprit here.

It should've never been used for the underlines.

Would you mind replacing it across all links with the standard `text-decoration: underline` ?

This should resolve the issue with the images.

nielslange:
@danieldudzic Replacing `box-shadow: 0 1px 0 0;` with `text-decoration: underline;` did the trick! 🙌

PR #1757, opened 2020-01-20
Varia: Fix Faux button style

### LINKED ISSUE #1751: Varia: Faux button style is broken
While I was testing the new Calendly block that adds a link with `.button` class (faux button). Varia tries to style it like a button, but it's broken on the front of the site. It looks fine in the editor.

Screenshot 2020-01-17 at 00 18 47

Twenty Twenty style it correctly.

Screenshot 2020-01-17 at 00 19 38

It should be displayed as an inline-block element. Calendly block is soon to be released so it'd be great if we can fix this in Varia and all the child themes quickly.

nielslange:
Fixes #1751 

<table>
<tr>
<td>Before:
<br><br>

#1751-before
</td>
<td>After:
<br><br>

#1751-after
</td>
</tr>
</table>

### DIFF
--- a/varia/sass/base/_extends.scss
+++ b/varia/sass/base/_extends.scss
@@ -14,6 +14,7 @@
 	background-color: #{map-deep-get($config-button, "color", "background")};
 	border-radius: #{map-deep-get($config-button, "border-radius")};
 	border-width: 0;
+	text-decoration: none;
 	padding: #{map-deep-get($config-button, "padding", "vertical")} #{map-deep-get($config-button, "padding", "horizontal")};
 
 	&:hover,
--- a/varia/sass/blocks/blog-posts/_style.scss
+++ b/varia/sass/blocks/blog-posts/_style.scss
@@ -150,7 +150,7 @@
 }
 
 @import "../../base/extends";
-button[data-load-more-btn] {
+button[data-load-more-btn], .button {
 	// Extend button style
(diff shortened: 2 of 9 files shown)


iamtakashi:
@nielslange, thanks for working on the fix! 

I know this wasn't mentioned in the original issue, but can we make sure that the link won't have an underline at all? It'd be good to test with a parent block that has a background colour or/and image. For example, Group or Cover block.

nielslange:
> @nielslange, thanks for working on the fix!
> 
> I know this wasn't mentioned in the original issue, but can we make sure that the link won't have an underline at all? It'd be good to test with a parent block that has a background colour or/and image. For example, Group or Cover block.

Sure, @iamtakashi! I'll adjust the PR according to your feedback later today and will create more test cases.

nielslange:
## Editor

<table>
<tr>
<td>Before:
<br><br>

#1757-editor-before
</td>
<td>After:
<br><br>

#1757-editor-after
</td>
</tr>
</table>

## Frontend

<table>
<tr>
<td>Before:
<br><br>

#1757-frontend-before
</td>
<td>After:
<br><br>

#1757-frontend-after
</td>
</tr>
</table>

cc: @iamtakashi

iamtakashi:
Thanks for the improvement! LGTM!!

PR #1769, opened 2020-01-28
Shoreditch: Make default color AA/AAA compliant

### LINKED ISSUE #1768: Shoreditch: Make default font color darker
We have a user request to make the default font color darker.

The current default font color is #7a7c84 and the user claims the contrast ratio of text on background is 4.16 for that color, while an acceptable ratio is 4.5

Trying to help them with CSS instead.

* User report: https://wordpress.org/support/topic/default-font-color/

nielslange:
Fixes #1768

<table>
<tr>
<td>Before:
<br><br>

#1768-contrast-ratio-before
</td>
<td>After:
<br><br>

#1768-contrast-ratio-after
</td>
</tr>
</table>

<table>
<tr>
<td>Before:
<br><br>

#1768-editor-before
</td>
<td>After:
<br><br>

#1768-editor-after
</td>
</tr>
</table>

<table>
<tr>
<td>Before:
<br><br>

#1768-frontend-before
</td>
<td>After:
<br><br>

#1768-frontend-after
</td>
</tr>
</table>

### DIFF
--- a/shoreditch/css/blocks.css
+++ b/shoreditch/css/blocks.css
@@ -527,7 +527,7 @@ p.has-drop-cap:not(:focus)::first-letter {
 .has-medium-gray-color:focus,
 .has-medium-gray-color:active,
 .has-medium-gray-color:visited {
-	color: #7a7c84;
+	color: #767676;
 }
 
 .has-medium-gray-background-color,
@@ -535,7 +535,7 @@ p.has-drop-cap:not(:focus)::first-letter {
 .has-medium-gray-background-color:focus,
 .has-medium-gray-background-color:active,
 .has-medium-gray-background-color:visited {
-	background-color: #7a7c84;
+	background-color: #767676;
 }
 
 .has-light-gray-color,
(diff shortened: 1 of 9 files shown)


KokkieH:
I'm not sure how to go about testing something like this. I suspect it requires a sandbox, so a theme dev would probably be better to review.

kjellr:
Design-wise, this seems great. 👍

nielslange:
> I'm not sure how to go about testing something like this. I suspect it requires a sandbox, so a theme dev would probably be better to review.

@KokkieH The best way to test this change if by checking it out via Git. However, there's also a second, slightly more complicated approach. 

1. Head over to https://github.com/Automattic/themes/tree/update/%231768, which contains that change, and download the repo as ZIP file
2. Unzip the ZIP file on your computer and open it
3. Compress the Shoreditch folder as ZIP file and upload it to an atomic site

It requires a few more steps, but this is how you can test this change without version control. 😉

mendezcode:
Deployed changes in wpcom.

PR #1778, opened 2020-02-03
Independent Publisher 2: Fix table cell width issue

### LINKED ISSUE #1383: Independent Publisher 2: Table Block Column Width Does Not Adjust
## Steps to replicate

1. Create a new table with cell data of varying widths
2. Publish

## Result

The columns in the table do not respect the widths of the data inside them, even with 'fixed width table cells' disabled in the table block settings.

See this screenshot:  
Annotation on 2019-09-03 at 09-48-46

compared to the same table on a block-editor theme (Maywood):
Screen Shot on 2019-09-03 at 09_46_36

## Expected

I expected the column widths to conform to the data inside.

* User report: #2316265-zen

nielslange:
Fixes #1383

<table>
<tr>
<td>Before:
<br><br>

#1383-before
</td>
<td>After:
<br><br>

#1383-after
</td>
</tr>
</table>

The `Table Setting` option `Fixed width table cells` remained untouched from this fix.


donalirl (approved):
This change works well when I checked it in Chrome, Firefox, Edge, and IE.

Mobile looks good too, but I did notice a different behavior compared to how the block themes (like Maywood) handle wide tables.

With this update to Independent Publisher 2, a scroll bar gets added like so:
https://share.getcloudapp.com/yAuLd7qQ

On Maywood, there is no left to right scroll. Instead the table cells are compressed to fit everything: https://share.getcloudapp.com/WnuEo2nP

nielslange:
@allancole & @iamtakashi We could need your input here:

> With this update to Independent Publisher 2, a scroll bar gets added like so:
https://share.getcloudapp.com/yAuLd7qQ

> On Maywood, there is no left to right scroll. Instead the table cells are compressed to fit everything: https://share.getcloudapp.com/WnuEo2nP

Would you rather avoid horizontal scrolling on mobile devices and having a very narrow table in which some data is hard to read or would you rather have a readable table but horizontal scrolling? Even though both options are not ideal, personally, I'd prefer the second option.

allancole:
@nielslange, I think that’s happening because the table cell needs `word-break: break-word` to deal with that extra long “straganofffffffffffffffffff” edge case text. Pretty sure that’ll prevent the horizontal scrolling and still allow the tables to match the width of the content :-)

nielslange:
@allancole & @donalirl I adjusted the table cells according to Allan's feedback in https://github.com/Automattic/themes/pull/1778#issuecomment-584174079. Personally, I prefer the option with the horizontal scroll, as the updated version makes the texts very hard to read, as seen in the screenshots.

<table>
<tr>
<td>Before:

#1778-before
</td>
<td>After:

#1778-after
</td>
</tr>
</table>

mendezcode (commented):
@nielslange The fix looks legit, but when testing I am a bit doubtful to deploy these changes. For example, not sure if this is expected behavior:

Screen Shot 2020-02-13 at 11 30 58 AM

_Note: I just added a table on Gutenberg, didn't change any setting. Did I miss something?_

allancole:
Yeah, I view this as an edge-case which means there’s really no perfect solution here. The table is constrained by the `$content_width`, and if a users adds words that are longer than that width inside a table, there simply isn’t enough room to make that display nicely without breaking the word up, or allowing it scroll. 

The text appears to gets cut off in the scrollable solution, so I lean toward the `word-break` solution but both are acceptable solutions in my opinion :-)

nielslange:
@allancole Based on the following screenshots, I'd rather not use `word-break: break-word;`:

<table>
<tr>
<td>With `word-break: break-word;`
<br><br>

#1778-with-word-break
</td>
<td>Without `word-break: break-word;`
<br><br>

#1778-without-word-break</td>
</tr>
</table>

cc: @mendezcode

allancole:
> Without `word-break: break-word;`

Yup, this looks about as best as it can. Good catch 👍

allancole (approved):
LGTM!

PR #1796, opened 2020-02-12
Independent Publisher 2: Correct submenu offset

### LINKED ISSUE #272: Independent Publisher 2: Submenu items cut off at right margin
Not sure if you will choose to address this issue or not after a Slack chat with @crunnells in Themes channel, but thought I would report it just in case. @crunnells mentioned this would require JS to fix.

## Steps to reproduce

1. Set up a submenu item in IP2 and view site
2. Adjust browser window so that menu item is last item on right
3. When you hover the top level menu item so the submenu appears, it is partially cut off at edge of container

If you have more than two levels, the additional levels are not accessible unless you scroll to the right.

Screenshot:
https://cldup.com/axSz5LHZJk.png

* User report: https://en.forums.wordpress.com/topic/submenus-5/#post-3187231
* User's site: https://thebeadenmillerfamilyhistory.wordpress.com/

nielslange:
Fixes #272

<table>
<tr>
<td>Before:

#272-before
</td>
<td>After:

#272-after
</td>
</tr>
</table>

### DIFF
--- a/independent-publisher-2/functions.php
+++ b/independent-publisher-2/functions.php
@@ -209,6 +209,8 @@ function independent_publisher_2_scripts() {
 
 	wp_enqueue_script( 'independent-publisher-2-navigation', get_template_directory_uri() . '/js/navigation.js', array(), '20170317', true );
 
+	wp_enqueue_script( 'independent-publisher-2-subnavigation', get_template_directory_uri() . '/js/subnavigation.js', array('independent-publisher-2-navigation'), '20190212', true );
+
 	if ( ! is_active_sidebar( 'sidebar-1' ) ) {
 		wp_enqueue_script( 'independent-publisher-2-images', get_template_directory_uri() . '/js/independent-publisher-2.js', array( 'jquery' ), '20170406', true );
 	}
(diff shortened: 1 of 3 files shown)


danieldudzic (changes requested):
We want to avoid adding new scripts and libraries to reduce the weight and maintenance burden.

I'd recommend exploring a CSS-only fix.

nielslange:
@danieldudzic I wouldn't know how to address this issue with a CSS-only fix if the submenu should only be affected when it overlaps. The only CSS-only solution I can come up with would affect all submenus and not only the one that would potentially overflow. Please note that due to a flexible width, it's not necessarily the last submenu item that would overflow. It could also be the 2nd-last, 3rd-last, etc. which makes a CSS-only solution only possible if the CSS for all submenus will be changed. Do you want me to update the PR in a way that all submenus are affected by this change or do you have a CSS-only solution in mind that I'm currently overseeing?

danieldudzic:
@allancole After some further investigation, it seems that this issue is prevalent across most (if not all) of our themes (with dropdowns expanding to the right).

I think it would be smart to make a standardized fix and apply it to all affected themes. Not sure pure CSS here will do.

What are your thoughts on this?

PR #1807, opened 2020-02-14
Maywood: Add a "No Text Shadow" block style for the cover block.

kjellr:
This PR adds a "No Text Shadow" block style to the cover block on the Maywood theme. In cases where the background is light, this allows folks to avoid having a really weird drop shadow behind their text: 

Default Block Style: 
Screen Shot 2020-02-14 at 8 45 12 AM

New Block Style:
Screen Shot 2020-02-14 at 8 44 45 AM

Block Style Options:
Screen Shot 2020-02-14 at 8 44 55 AM


maywood/style-editor.css
```diff
@@ -475,6 +475,7 @@ object {
 	background-color: #897248;
 	border-radius: 5px;
 	border-width: 0;
+	text-decoration: none;
```
kjellr: There are a few extra things that showed up when the Sass compiled. I'm guessing these are supposed to be there? But I can prune them out if they're not.

maywood/style.css
```diff
@@ -194,8 +194,7 @@ $grid-configuration: map-extend($grid-configuration-default, $grid-configuration
  *   let’s use a placeholder to keep them all
  *   in-sync
  */
-button[data-load-more-btn], button,
-.button,
+button[data-load-more-btn], .button, button,
```
kjellr: These changes also showed up on compile, but weren't modified in the Sass.

mendezcode: I think this happens because the internal Sass build has been updated in a recent version and it's now grouping selectors differently. Should be fine.

pablohoneyhoney:
It might help to use bold on images?

kjellr:
> It might help to use bold on images?

Definitely, but that's something the user would set themselves in the editor — so I don't think it makes sense to specify that here by default.

maywood/style.css
allancole: Yeah, this (and the other stuff that pops up after recompiling) is a result of a fix getting added to Varia on `master`, but not yet applied to the child themes. It’s pretty tedious to deploy each theme individually, so sometimes Varia fixes won’t get into a child theme until a theme-specific issue is worked on like in this issue. 

Technically, when we do a fix in Varia we should be auto-recompiling _all_ the child themes on the same branch but that part of the `theme-dev-util` hasn’t got merged to `master` yet so this kind of thing has been happening occasionally. As far as, I’ve seen these kinds of “upstream”, Varia fixes haven’t caused problems, so I think this PR is good to go.

pablohoneyhoney:
It's important we provide the best design defaults and then let the users define what they need. 
The light font won't be legible over noisy images.

kjellr:
> It's important we provide the best design defaults and then let the users define what they need.
The light font won't be legible over noisy images.

Fully agreed — In general, my only hesitancy here is that this theme is already live on hundreds of sites. For all we know, users chose the theme because they like that light font weight there... it's featured prominently [in the theme demo](https://maywooddemo.wordpress.com) even. I'm not really comfortable with us changing that on so many websites all of a sudden. Themes on WP.com are auto-updated, so it's not possible for users to avoid updating if they wanted to keep things the way they are. 

With that in mind, we tend not to make noticeable design updates after launch — we don't have a good way to iterate on design decisions for themes after they're in the wild. 

This PR is a bit of a workaround to this problem. It implements an optional Cover style for folks to choose if the default shadow doesn't work for them. We could theoretically change the font weight in here too, but we'd run into the UI issues noted above... all text would show as bold by default without being noted as bold in the Gutenberg text formatting options. I think global styles will hopefully offer a better framework for this.

In this case, I'm most comfortable just ensuring that we take this to heart moving forward. We should be more careful about this sort of readability and text treatment moving forward, and try to find better ways for text to work inside of cover blocks. If we did fully revisit the Maywood theme as a brand new, separate theme in the future (as we've done with themes like [Independent Publisher](https://en.blog.wordpress.com/2017/04/25/independent-publisher-2-is-here/)), then we could include an update on the font weight in there.

allancole:
This is fixed in #1949 which I just merged to master and deployed to Dotcom.

Closing this issue 👍

PR #1810, opened 2020-02-17
Varia: Adjust menu logic

### LINKED ISSUE #1808: Varia: Menu button appears for primary menu on mobile even when menu is empty.
## Steps to replicate

1. Set site to use Varia or one of the Varia child themes (tested with Maywood, Mayland, and Exford).
2. Set an empty menu in the "Primary" menu location.
3. View site on mobile.

## Result

A menu button appears for the empty primary menu, allowing you to click to expand—nothing :).
exford mobile empty menu button

## Expected

For the page to display without a menu button on mobile—because the menu is empty.

## Temporary Fix/Workaround

Making sure that nothing is set to display in the "Primary" menu location removes the button.

## Comment
The fix/workaround that exists for this is somewhat less than intuitive—users do not generally understand that this is an option, and assume that if they empty out the menu, the button will disappear. This was opened previously in https://github.com/Automattic/themes/issues/1331 and then closed for priority. However, this is the fifth time I've encountered this issue as an HE, and with more folks using our Varia-based themes, it seems like it *might* be worth looking at. But feel free to tell me I'm wrong :).

* User report: #18676186-hc is my most recent example.

### LINKED ISSUE #1813: Maywood: empty menu button shows on mobile
## Steps to replicate

1 - Create an empty menu
2 - Assign it to the main location
3 - Save your changes

## Result

The menu doesn't appear on the desktop as expected. However, a **Menu +** toggle button appears on mobile.

## Expected

I would expect the menu not to show on mobile if it's empty. Or at least, I expected the same behavior both on mobile and desktop.

* User report: #18589958-hc
* Follow-up ticket: None - we deactivated menu in chat

nielslange:
Fixes #1808
Fixes #1813 (added later)

---

By changing `has_nav_menu( 'menu-1' )` to `has_nav_menu( 'menu-1' ) && wp_get_nav_menu_items( 'menu-1' )` the underlying `<nav>` part only gets loaded if the menu it not empty. Otherwise, the `<nav>` part itself does not even get loaded, which leads to less HTML code and should therefore load the site a bit faster.

<table>
<tr>
<td>Before:
<br><br>

#1808-before
</td>
<td>After:
<br><br>

#1808-after
</td>
</tr>
</table>

### DIFF
--- a/varia/header.php
+++ b/varia/header.php
@@ -45,7 +45,7 @@
 
 			<?php get_template_part( 'template-parts/header/site', 'branding' ); ?>
 
-			<?php if ( has_nav_menu( 'menu-1' ) ) : ?>
+			<?php if ( has_nav_menu( 'menu-1' ) && wp_get_nav_menu_items( 'menu-1' ) ) : ?>
 				<nav id="site-navigation" class="main-navigation" aria-label="<?php esc_attr_e( 'Main Navigation', 'varia' ); ?>">
 					<input type="checkbox" role="button" aria-haspopup="true" id="toggle" class="hide-visually">
 					<label for="toggle" id="toggle-menu" class="button">


danieldudzic (changes requested):
This doesn't seem to work for me correctly.

It hides the menu when there are no items, but it fails to display it when menu items are present.

scruffian:
I think this has since been fixed another way. Closing.

PR #1835, opened 2020-03-03
Dyad 2 - Remove color inhertiance from input/textarea/button fields

### LINKED ISSUE #1833: Dyad 2: Contact Form text is white.
## Steps to replicate

1. Add Contact Form (either template or Form block) to a page
2. View the Contact Form and try to input information.

## Result
The text in the input fields are in white on a white background and thus not visible.

[What happened.]

## Expected
The text in the input fields should be visible.

[What you expected to happen.]

* User report: [shortlink to chat/forum thread/ticket, ie #12345-hc]
* Follow-up ticket: [shortlink to ticket, ie #12345-zd]
#2755796-zen

achyuthajoy:
Removed `color: inherit` from `style.css` since it was conflicting with the default styles of cover block and rendering the input fields text color as white.

Proposed fix for #1833

### DIFF
--- a/dyad-2/style.css
+++ b/dyad-2/style.css
@@ -698,7 +698,6 @@ button,
 input,
 textarea {
 	border: 1px solid #ddd;
-	color: inherit;
 	font-family: inherit;
 	font-size: 1.6rem;
 	line-height: 1.5;


dyad-2/style.css
```diff
@@ -698,7 +698,6 @@ button,
 input,
 textarea {
 	border: 1px solid #ddd;
-	color: inherit;
```
pbking, line 701: color: #6a6c6e;

pbking (changes requested):
Removing the 'inherit' value drops the color of input/textarea elements to the default black in all instances.
Rather if we set the value to the defined text color value explicitly then it remains the expected color in existing situations as well as changes it to the expected value in the broken use case.

pbking:
Closing in deference to #2778

PR #1839, opened 2020-03-05
Add a No Text Shadow style to cover block.

### LINKED ISSUE #1837: Maywood: Add a new Cover block style variation that's without text shadow
Maywood adds text-shadow for the text on Cover block to help the text stands out a bit, but it's been come up as an issue more than once. `p9Jlb4-1g4-p2` and https://github.com/Automattic/wp-calypso/issues/39350.

As it's been suggested by @kjellr, it'd be great if we can add a new cover block style that's without the shadow, and make it as the default style so that the change wouldn't be so disruptive

alaczek:
By default text in cover block gets a dark shadow. 

image

This PR adds a No Text Shadow style to cover block.

image

image

I'm not well versed with Varia child themes, so let me know if I missed anything.

### DIFF
--- a/maywood/block-extends/extend-cover-block.css
+++ b/maywood/block-extends/extend-cover-block.css
@@ -1,0 +1,45 @@
+/**
+ * Editor
+ */
+ .editor-styles-wrapper .wp-block-cover.is-style-no-text-shadow p, 
+ .editor-styles-wrapper .wp-block-cover-image.is-style-no-text-shadow p {
+ 	text-shadow: none;
+ }
+
+.editor-styles-wrapper .wp-block-cover.is-style-no-text-shadow h1, 
+.editor-styles-wrapper .wp-block-cover.is-style-no-text-shadow h2, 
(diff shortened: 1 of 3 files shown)


nielslange:
@alaczek Wouldn’t it be better to add this functionality directly to the Varia theme, so that all its child themes can benefit from this change?

alaczek:
> Wouldn’t it be better to add this functionality directly to the Varia theme, so that all its child themes can benefit from this change?
As far as I can tell this default text shadow style in cover block is present only in Maywood, not in Varia itself. Am I missing something?

nielslange:
> As far as I can tell this default text shadow style in cover block is present only in Maywood, not in Varia itself. Am I missing something?

My bad, @alaczek! I just double-checked this and the Maywood theme is the only theme that shows a `text-shadow` on the cover block. Varia and all Maywood siblings do not use `text-shadow`. Sorry for the confusion.

pbking:
Noting that original issue is closed; leaving this closed.

PR #1902, opened 2020-04-06
Varia: Ensure full-width items inside of columns do not overlap.

kjellr:
Fixes #40719

This PR adds some editor styles to Varia that ensure that full-width blocks inside of columns do not expand beyond the width they're allowed. It accomplishes this by asking those blocks to inherit the standard margins instead of the full-width margins. 

Before:

Screen Shot 2020-04-06 at 2 31 00 PM
Screen Shot 2020-04-06 at 2 37 27 PM

After:

Screen Shot 2020-04-06 at 2 35 03 PM

Screen Shot 2020-04-06 at 2 37 06 PM

### DIFF
--- a/varia/sass/blocks/_editor.scss
+++ b/varia/sass/blocks/_editor.scss
@@ -10,6 +10,7 @@
 @import "blog-posts/editor";
 @import "button/editor";
 @import "code/editor";
+@import "columns/editor";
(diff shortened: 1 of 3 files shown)


allancole (approved):
LGTM along with #1921 right after ;-)

PR #1917, opened 2020-04-13
[all themes] Declare HTML5 scripts and styles support for better HTML5 compliance

audrasjb:
Hi,

Given all WordPress.com themes are written in HTML5, they should declare HTML5 support for styles and script to avoid using `type` attributes in styles and scripts.

Otherwise, W3C validator will throw a warning because of the `type` attribute.

This option was introduced in WordPress 5.3. For reference, see:
- https://make.wordpress.org/core/2019/10/15/miscellaneous-developer-focused-changes-in-5-3/
- https://core.trac.wordpress.org/ticket/42804#comment:32

Cheers,
Jb

### DIFF
--- a/affinity/functions.php
+++ b/affinity/functions.php
@@ -79,6 +79,8 @@ function affinity_setup() {
 		'comment-list',
 		'gallery',
 		'caption',
+		'style',
+		'script',
 	) );
 
 	// Set up the WordPress core custom background feature.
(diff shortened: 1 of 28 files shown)


pbking:
Noting that these are being re-opened on a per-theme basis.  This will remain closed.

audrasjb:
@pbking yes. I already opened PR for other (most used) themes. See related PRs above.

PR #1935, opened 2020-04-22
Dara: Prevent that homepage can be selected as featured page

### LINKED ISSUE #1932: Dara: featured pages can lead to duplicated content
## Steps to replicate

1. set static front page
2. open customizer and set a featured page 
3. assign the page set as static front page as featured.

## Result

Homepage content is mysteriously duplicated 
image

## Expected

User should not be able to feature homepage to avoid duplicate content. 

* User report: #16865785-hc

nielslange:
Fixes #1932 

<table>
<tr>
<td>Before:
<br><br>

#1932-before
</td>
<td>After:
<br><br>

#1932-after
</td>
</tr>
</table>

### DIFF
--- a/dara/inc/customizer.php
+++ b/dara/inc/customizer.php
@@ -77,7 +77,7 @@ add_action( 'customize_register', 'dara_customize_register' );
  * @return interger.
  */
 function dara_sanitize_dropdown_pages( $input ) {
-	if ( is_numeric( $input ) ) {
+	if ( is_numeric( $input ) && $input != get_option( 'page_on_front' ) ) {
 		return intval( $input );
 	}
 }


dara/inc/customizer.php
```diff
@@ -77,7 +77,7 @@ function dara_customize_register( $wp_customize ) {
  * @return interger.
  */
 function dara_sanitize_dropdown_pages( $input ) {
-	if ( is_numeric( $input ) ) {
+	if ( is_numeric( $input ) && $input != get_option( 'page_on_front' ) ) {
```
MaggieCabrera, line 80: This function is meant to sanitize the input from the dropdown, not to use it as a filter, when the user selects the home page as an option the error message is hard to understand. I tried filtering what's on the control but it felt like redoing the whole "dropdown-pages" type by hand for this issue.

scruffian:
@nielslange are you able to update this?

PR #1940, opened 2020-04-24
Fix incorrect font name

sgomes:
Using a `+` instead of a space here will cause the font name to be doubly encoded (as `PT%2BSans`), resulting in an error 400 when loading from Google Fonts.

@josephscott kindly validated the fix by testing on his sandbox.

Fix Google Font loading. Validate by checking in DevTools that it no longer produces an error 400 when attempting to load the font CSS.


allancole:
LGTM!

sgomes:
Thank you for the review + merge, @allancole! 👍

PR #1944, opened 2020-04-27
Varia: Add Color Annotations

allancole:

This PR attempts to add color-annotation support to Varia.

#### Screenshot: 

varya-color-annotations-1

#### ToDo: 

- [x] Add support for Primary, Secondary, Foreground, Background, and Accent colors where approriate.
- [x] Improve the color-annotations plugin to support the Gutenberg editor styles and color palettes.
- [x] Add featured color palettes to Varia.
- [x] Apply this solution to all Varia-based child themes.


allancole:
Still working through all the subtle details but after the last batch of commits here’s where things are: 

varya-color-annotations-2

Works pretty well so far and it looks like we’ll be able to make and custom-colors configuration a11y friendly.

kjellr (commented):
This looks pretty good! I tested it out and had only one really minor observation: 

I think the SVG icons here need a `fill: curentColor` rule applied. Might as well bundle that into this PR. 

Screen Shot 2020-04-29 at 12 40 14 PM

_EDIT: I pushed a quick fix for this in_ 1151c03

varia/functions.php
```diff
@@ -145,19 +145,25 @@ function varia_setup() {
 		// $lightness       = varia_get_default_lightness();
 		// $lightness_hover = varia_get_default_lightness_hover();
 
+		$colors_array = get_theme_mod( 'colors_manager' );
```
kjellr, line 148: Minor, but can you add a code comment explaining what this is?

allancole: Added some comments in a28de42

kjellr: Thanks!

varia/style.css
```diff
@@ -2436,35 +2440,35 @@ table th,
 }
 
 .has-primary-color[class] {
-	color: blue !important;
```
kjellr, line 2439: Will the removal of these `!important` rules automatically flow down to the child themes when we recompile their CSS? It doesn't look like these rules are defined in the child theme CSS in the first place.

varia/functions.php
```diff
@@ -140,10 +140,16 @@ function varia_setup() {
...
-		// $saturation      = varia_get_default_saturation();
-		// $lightness       = varia_get_default_lightness();
-		// $lightness_hover = varia_get_default_lightness_hover();
+		/*
+		 * Get customizer colors and add them to the editor color palettes
+		 *
+		 * - if the customizer color is empty, use the default
+		 */
+		$colors_array = get_theme_mod('colors_manager'); // color annotations array()
+		$background   = $colors_array['colors']['bg'];   // $config-global--color-background-default;
```
kjellr, line 149: One thing I noticed is that this throws errors when I run the theme locally. For example: 

```
Notice: Trying to access array offset on value of type bool in ../wp-content/themes/dotcom/alves/functions.php on line 62

Notice: Trying to access array offset on value of type null in ../wp-content/themes/dotcom/alves/functions.php on line 62
```

Can we wrap these in a conditional to prevent that?

allancole: Figured out a solution for this and revised the the conditionals in all the child themes so they don’t throw an error on non-dotcom setups, and so they’re a little easier to digest. Also switched to `! empty()` from `! isset()` which is a better [kind of check](https://phppot.com/php/isset-vs-empty-vs-is_null/).

allancole:
Just adding a not that this PR fixes #2103

allancole:
Merging this PR and syncing the changes to wpcom.

PR #1949, opened 2020-04-29
Maywood: Only apply cover image text shadow if there's a background image or video.

kjellr:
As suggested in #1837, this PR changes Maywood's behavior so that it only applies a text shadow to inner blocks if there's a background image in the color block.

Before: 

Screen Shot 2020-04-29 at 1 04 49 PM
Screen Shot 2020-04-29 at 1 12 43 PM

After: 

Screen Shot 2020-04-29 at 1 03 52 PM
Screen Shot 2020-04-29 at 1 12 51 PM

### DIFF
--- a/maywood/sass/_extra-child-theme.scss
+++ b/maywood/sass/_extra-child-theme.scss
@@ -228,8 +228,9 @@ b, strong {
 		font-weight: 300;
 	}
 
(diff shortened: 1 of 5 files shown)


iamtakashi:
LGTM. Thank you!

PR #1950, opened 2020-04-29
Balasana: Add Color Annotations

kjellr:
Follows the mold of #1944 to add basic color annotations for Balasana. 

Screen Shot 2020-04-29 at 14 05 46

Still needs: 


balasana/inc/wpcom-colors.php
```diff
@@ -0,0 +1,340 @@
...
+  */
+function balasana_custom_colors_extra_css() { ?>
+	::selection {
+		background-color: currentColor;
+		mix-blend-mode: hard-light;
+	}
+	::-moz-selection {
+		background-color: currentColor;
+		mix-blend-mode: hard-light;
+	}
```
allancole, line 333: So these selection colors don’t always come out that great. Plus browser support for `mix-blend-mode` isn’t great either. Wondering if we actually _need_ selection colors at all. It’s a nice subtle design detail, but without access to color opacity, it’s tough to really get it right without doing it manually for each theme. 

Varia now uses the primary and background colors for selected text like this:
https://github.com/Automattic/themes/pull/1944/files#diff-aa2f4bd1a0647c2da27f37eca952a82fR375-R392

It looks okay but it’s also kind of aggressive depending on the selected color options: 
image

kjellr: Yeah, maybe we should just set all of the selection styles back to `initial`.

kjellr: FWIW, `mix-blend-mode` doesn't seem to work at all in Firefox, making the text unreadable.

PR #1955, opened 2020-04-30
Leven: Add Color Annotations

kjellr:
Follows the mold of #1944 to add basic color annotations for Leven. 

Screen Shot 2020-04-30 at 14 19 06

Still needs: 


allancole (approved):
This looks good on both desktop & mobile :-) 
Should be good to merge once #1944 is in.

kjellr:
Found a handful of small bugs: 

**1. The default button text color is always white, no matter what colors are chosen.**

Screen Shot 2020-08-13 at 1 46 28 PM

**2. When a "Background" background color is assigned, with no text color, the link color doesn't match between the front and back end. ** 

Front End|Editor
---|---
Screen Shot 2020-08-13 at 1 42 59 PM|Screen Shot 2020-08-13 at 1 42 50 PM

**3. The default "outline" button variant always appears to be red, regardless of the palette chosen. In addition, it appears incorrectly in the editor.** 

Front End|Editor
---|---
Screen Shot 2020-08-13 at 1 47 56 PM|Screen Shot 2020-08-13 at 1 47 52 PM

**4. In general, the default Leven homepage colors (the red, blue, and yellow) do not change, no matter what palette you choose.** 

I _think_ this is intended, but it may confuse folks. (cc @ianstewart in case this is something that needs to be updated in the Leven template itself).

Screen Shot 2020-08-13 at 1 51 11 PM
Screen Shot 2020-08-13 at 1 49 48 PM

ianstewart:
Colours seem to be working as intended in my test site …

image

Taking a look at that issue @kjellr noted above.

ianstewart:
Hmm. This is not what I expected to see in the Editor …

image

ianstewart:
Here's what I see, un-sandboxed, on the theme demo site …

image

ianstewart:
> I think this is intended, but it may confuse folks.

I think it's OK for now. It's part of Leven's homepages charm. But that's a weakly held opinion. If it we get reports of confusion I'll recant and we'll fix it. :)

allancole:
Those last couple of commits should have fixed the above issues sans `4. In general, the default Leven homepage colors (the red, blue, and yellow) do not change, no matter what palette you choose.`. 

If I remember correctly that template might have been made before the color utility classes were setup properly. Because of it, some of the blue, red, and yellow color settings use inline CSS instead of the utility classes. Best way to fix these is directly in the template.

allancole:
This PR should be good to merge now, cc: @kjellr @ianstewart :-)

kjellr:
@allancole This looks good! I'm still seeing this issue, but the other are all taken care of. 

> 1. The default button text color is always white, no matter what colors are chosen.

kjellr:
Those last two commits take care of the final issue for me. This is good to go!

allancole:
Sweet! Added one other fix to make sure the outline looks right on frontend and editor. LGTM!

PR #1956, opened 2020-04-30
Hever: Add Color Annotations

kjellr:
Follows the mold of #1944 to add basic color annotations for Hever. 

Screen Shot 2020-04-30 at 14 34 01

Still needs: 


scruffian:

PR #1960, opened 2020-04-30
Morden: Add Color Annotations

kjellr:
Follows the mold of #1944 to add basic color annotations for Morden. 

Screen Shot 2020-04-30 at 16 17 28

Still needs: 

- [ ] Add featured color palettes.


scruffian:

PR #1967, opened 2020-05-01
Maywood: Add Color Annotations

kjellr:
Follows the mold of #1944 to add basic color annotations for Maywood. 

Screen Shot 2020-05-01 at 09 09 56

Still needs: 


kjellr:
@allancole I just did some testing and found a handful of small bugs: 

**1. The menu has no hover state on desktop:** 

hovers

**2. Many of the default homepage's featured areas look pretty ugly, or don't match up in the front and back end:** 

Frontend|Editor
---|---
Screen Shot 2020-08-26 at 12 10 12 PM|Screen Shot 2020-08-26 at 12 10 27 PM

Frontend|Editor
---|---
Screen Shot 2020-08-26 at 12 10 38 PM|Screen Shot 2020-08-26 at 12 10 46 PM

Frontend|Editor
---|---
Screen Shot 2020-08-26 at 12 11 19 PM|Screen Shot 2020-08-26 at 12 11 28 PM

**3. The select field in the Form block appears mostly unreadable in the editor when you have a dark color scheme applied:** 

Screen Shot 2020-08-26 at 12 11 00 PM

allancole:
> 1. The menu has no hover state on desktop 

Fixed:



> 2. Many of the default homepage's featured areas look pretty ugly, or don't match up in the front and back end 

I ended up removing the text-shadow rules entirely for this. Text-shadow only really looks good on black background-dim colors with white text and since we won’t always know what the color combinations will be, removing text-shadow seemed approapriate. 

Frontend|Editor
------------|------------
image|image
image|image
image|image
image|image

> 3. The select field in the Form block appears mostly unreadable in the editor when you have a dark color scheme applied:

This needs to be fixed by adding proper jetpack-form block support to Varia before we can deal with implementing color-annotations in child themes. Will have to do that in a separate PR, unfortunately so that all Varia child-themes support this correctly.

Let me know what you think, @kjellr

kjellr:
Thanks @allancole — that all sounds reasonable. 

> This needs to be fixed by adding proper jetpack-form block support to Varia before we can deal with implementing color-annotations in child themes. Will have to do that in a separate PR, unfortunately so that all Varia child-themes support this correctly.

Does that mean the readability of those field options is an issue in all Varia child themes with color annotations right now? If so, it's weird we didn't notice before now.

allancole:
> Does that mean the readability of those field options is an issue in all Varia child themes with color annotations right now? If so, it's weird we didn't notice before now.

Sadly yes, although it seems to mostly effect the dark color scheme(s). I thought an alternative fix might be to reduce the text-color specificity in Jetpack itself, but it turns out that text color rule comes from a [general wp-admin stylesheet for forms](https://github.com/WordPress/WordPress/blob/af523447f46c351cb27432813273049777fbb890/wp-admin/css/forms.css#L58). That pretty much forces us to fix it in Varia, recompile the child themes and then update the annotations again :-/ .

allancole:
LGTM!

PR #1970, opened 2020-05-01
Rockfield: Add Color Annotations

kjellr:
Follows the mold of #1944 to add basic color annotations for Rockfield. 

Screen Shot 2020-05-01 at 09 43 43

Still needs: 


rockfield/inc/wpcom-colors.php
```diff
@@ -95,6 +95,91 @@
...
 
+	/**
+	 * Grays
+	 */
+	// Text Color (Slightly less opaque color)
+	array( '.main-navigation a:hover,
+			.site-title a:hover,
+			.social-navigation a:hover,
+			.site-title:not(:empty) + .site-description:not(:empty):before,
+			#site-navigation > div > ul > li:not(:first-child):before', 'color', 0.8 ),
```
allancole, line 106: Ahhh, good idea to use opacity here 👍

kjellr: 🙇 Thanks! Since these are just hovers, that seemed to make sense.

ianstewart:
Colour annotations seem to be working, generally, in this one. This …

image

… felt a little off but I'm not sure if it's something we can handle in the annotation.

kjellr:
Agreed with @ianstewart — in general the annotations looked good. Aside from the issue Ian noted above, I noticed just a couple issues: 

**1. The "Foreground Dark" Background color isn't appearing correctly in the editor**

Front end|Editor
---|---
kjellreigstad wordpress com_2020_06_19_custom-background-colors_ (1)|kjellreigstad wordpress com_wp-admin_post php_post=1514 action=edit

**2. The default button styles don't match between the front end and editor.**

(I think the editor is the correct one)

Front end|Editor
---|---
Screen Shot 2020-09-04 at 1 37 09 PM|Screen Shot 2020-09-04 at 1 37 04 PM

scruffian:

PR #1971, opened 2020-05-01
Stratford: Add Color Annotations

kjellr:
Follows the mold of #1944 to add basic color annotations for Stratford. 

Screen Shot 2020-05-01 at 09 53 46

Still needs: 


stratford/inc/wpcom-colors.php
```diff
@@ -0,0 +1,394 @@
...
+			.has-background-dark-background-color[class],
+			.has-background-light-background-color[class]', 'color' ),
+	// Background-color
+	array( '.has-background-background-color[class]', 'background-color' ),
+	// Background-color darkened
+	array( '.has-background-dark-background-color[class]', 'background-color', '-1' ),
+	// Background-color lightened
+	array( '.has-background-light-background-color[class]', 'background-color', '+1' ),
+
+	/**
```
kjellr, line 86: @allancole In a27c7cc I added some annotations to get grays working better — they're just slightly darker or lighter variants of existing colors. 

The only issue is that if the user chooses a black background color, the "darker" version of that color is obviously black too. Is there a contrast setting that would swap it out to a lighter version in that case? If not, I think it's okay — this is still better than nothing.

allancole: > Is there a contrast setting that would swap it out to a lighter version in that case?

Yeah, we can choose between using brightness (which you currently have) or just contrast which would look like this `'background-color', 'bg', '12' )`. If you use contrast, then it’ll go dark or light depending on the `'bg'` color. But this is also what causes the colors to not match the users expectations sometimes. Ideally, we could combine them both like this`'background-color', 'bg', '+2' )`, but thats not possible with brightness.

kjellr: Ok, got it. I looked into a bunch of options, and I think this works the best. It falls apart if the palette is super-dark (like with a black background), but In general, these are all dividers, not text... so I think it's okay.

ianstewart:
I tested the dark colour scheme for Stratford this morning.

- The editor was not picking up the dark colour scheme.
- The front-end and Editor previews worked as expected.

kjellr:
> The editor was not picking up the dark colour scheme.

@ianstewart: Just clarifying this (for this issue, and the other one you tested): are you saying that the editor styles did not appear at all when using the dark color palette? Or are you saying that Gutenberg didn't activate its ["dark mode"](https://developer.wordpress.org/block-editor/developers/themes/theme-support/#dark-backgrounds) when using a dark color palette?

ianstewart:
> editor styles did not appear at all when using the dark color palette?

I've been following up in Slack but to close this loop, yes. This. :D

kjellr:
👋 Noticed a few issues in testing: 

**1. Default button appearance doesn't match between the editor and the front end:** 

Editor|Front end
---|---
Screen Shot 2020-09-11 at 8 35 29 AM|Screen Shot 2020-09-11 at 8 35 14 AM

**2. On the default homepage, the font color for these items don't match in the editor vs. the front end:** 

Editor|Front end
---|---
Screen Shot 2020-09-11 at 8 37 15 AMScreen Shot 2020-09-11 at 8 40 01 AM|Screen Shot 2020-09-11 at 8 37 05 AMScreen Shot 2020-09-11 at 8 40 16 AM


**3. The quote block gets a left border color when a custom palette is active (In the front end and the editor too). The default does not include a left border color.**

Default Palette|Custom Palette
---|---
Screen Shot 2020-09-11 at 8 34 15 AM|Screen Shot 2020-09-11 at 8 34 09 AM

**4. A couple of the color palette colors don't seem to work correctly in the front end or editor. Additionally, the link color inside these doesn't match up between the front end and the editor.** 

Editor|Front end
---|---
Screen Shot 2020-09-11 at 8 41 26 AM|Screen Shot 2020-09-11 at 8 41 32 AM

scruffian:
We did this using CSS variables instead.

PR #1972, opened 2020-05-01
Varia: Remove negative margins for full-width items within full-width group blocks

kjellr:
Fixes the editor overflow portion of https://github.com/Automattic/wp-calypso/issues/41719 (for just the Varia parent theme, but this effects all child themes too). 

This removes the negative margins on full-width child blocks that live inside of a full-width group block. To be honest, this doesn't feel like something that should _need_ to be fixed, since those margins come from Gutenberg. I'm also seeing that issue only on WP.com, not on my local install. So we should do some investigation before we commit this — I worry that this will break again someday if this isn't the root of the issue. 

Before: 

Screen Shot 2020-05-01 at 15 54 42

After:

Screen Shot 2020-05-01 at 15 43 25

### DIFF
--- a/varia/sass/blocks/group/_editor.scss
+++ b/varia/sass/blocks/group/_editor.scss
@@ -12,3 +12,10 @@
 	margin: 0;
 	width: 100%;
 }
+
+// Remove negative margins for full-width children of Group blocks.
+.wp-block-group__inner-container .wp-block.alignfull, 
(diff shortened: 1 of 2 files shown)


simison:
Make sure to test with Gutenberg 8.0 as there are some related fixes.

kjellr:
I can confirm that this specific issue doesn't appear in v8.0, so I'll close this for now. 👍

Issue #2009, opened 2020-05-13
Varia & Varia Child Themes: Comments author avatar size issue


deepaklalwani97:
## Steps to replicate

Check the comments section of any post in AMP mode and check the author's avatar.

## Result

Comment author avatar size is enlarged on the AMP page. Also, the reply link is also not working because of the overlying image.

![81372719-ae6f6500-9118-11ea-81b1-7c48c1fe8fa5](https://user-images.githubusercontent.com/1139853/81587707-0a661200-93d5-11ea-9e31-f244daa967a2.png)

## Expected

Gravatar size should be the same as the Non-AMP version. Refer to the screenshot below.

![81372899-19b93700-9119-11ea-8e79-ff4e6751d4bb](https://user-images.githubusercontent.com/1139853/81587726-1225b680-93d5-11ea-93ec-5056e7b7fe1b.png)

## Status of theme compatibility
Issue exists on all the following themes. 
- [x] Varia
- [x] Maywood
- [x] Exford
- [x] Shawburn
- [x] Alves
- [x] Balasana
- [x] Barnsbury
- [x] Coutoire
- [x] Dalston
- [x] Hever

allancole:
Same as #2011, In the future, when you find a systematic Varia issues like this, it’s a little cleaner to solve these in a single PR with separate commits that address each theme. Thanks!

PR #2010, opened 2020-05-13
Varia: Fix comment avatar image/position size on AMP pages

mangesh:
Fixes #2009

#### Changes proposed in this pull request:
Varia Theme: Fix comment avatar size for AMP pages.

Before on iPad AMP | After on iPad AMP
------------ | -------------
image | image

Before on iPad Non AMP | After on iPad Non AMP
------------ | -------------
image | image

### DIFF
--- a/varia/sass/components/comments/_comments.scss
+++ b/varia/sass/components/comments/_comments.scss
@@ -144,7 +144,7 @@
 			.avatar {
 				margin-right: #{map-deep-get($config-global, "spacing", "horizontal")};
 				display: inherit;
-				position: inherit;
+				position: relative;
 				right: inherit;
 			}
 		}
--- a/varia/style.css
+++ b/varia/style.css
@@ -3418,7 +3418,7 @@ body:not(.fse-enabled) .footer-menu a {
 	.comment-meta .comment-author .avatar {
 		margin-right: 16px;
 		display: inherit;
-		position: inherit;
+		position: relative;
 		right: inherit;
 	}
 	.comment-meta .comment-metadata {


varia/sass/components/comments/_comments.scss
```diff
@@ -144,7 +144,7 @@
 			.avatar {
 				margin-right: #{map-deep-get($config-global, "spacing", "horizontal")};
 				display: inherit;
-				position: inherit;
+				position: relative;
```
westonruter, line 147: I think the `position` can be just removed entirely, no?

westonruter: Nevermind. I see that otherwise it gets `position:absolute`, so that has to be overridden.

It's important for `amp-img` to get `position:relative` (or `position:absolute`) because the nested `img` element needs the parent `amp-img` to have a new stacking context so that it can fill its parent when the `img` does `position:absolute; top:0; right:0; bottom:0; left:0;`. 

The `amp-img` has `position:relative` by default for images (with intrinsic layout, the default as output by the AMP plugin): 

https://github.com/ampproject/amphtml/blob/e1a66bce763efff45522ab77e15cdc7ff2aaebc7/css/ampshared.css#L78-L84

So that `position` needs to not be overridden with a value that removes the new CSS stacking context.

pradeep910: Removing position entirely somehow creates different issue. Somehow it takes absolute position and hides the author name behind the avatar. Having relative position seems to be fixing this theme issue for AMP. 
 
Screenshot 2020-05-14 at 11 54 28 PM

PR #2015, opened 2020-05-14
Update Dyad 2 Editor Styles

kjellr:
Dyad 2's editor styles were targeting outdated selectors. This PR updates them to use `editor-styles-wrapper` and `wp-block`. (Really though, we should be enquing the editor styles via `add_editor_style()` instead of the way we're doing it here, but that can be left to a separate PR). 

Here's a front end screenshot for reference: 

Screen Shot 2020-05-14 at 8 10 06 AM

Here's how that looked in the editor before this PR: 

Screen Shot 2020-05-14 at 8 21 35 AM

... and here's how it looks _after_ this PR: 

Screen Shot 2020-05-14 at 8 21 04 AM

### DIFF
--- a/dyad-2/css/editor-blocks.css
+++ b/dyad-2/css/editor-blocks.css
@@ -18,14 +18,11 @@ Description: Used to style Gutenberg Blocks in the editor.
 1.0 General Typography
 --------------------------------------------------------------*/
 
-.editor-styles-wrapper .wp-block,
-.editor-styles-wrapper .wp-block p,
+.editor-styles-wrapper,
+.editor-styles-wrapper p,
 .editor-default-block-appender textarea.editor-default-block-appender__content {
 	font-family: "Lato", "Helvetica Neue", Helvetica, Arial, sans-serif;
 	font-size: 17px;
-}
-
-.editor-styles-wrapper .wp-block {
 	color: #6a6c6e;
 }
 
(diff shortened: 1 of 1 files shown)


kjellr:
This could use a bit of a wider test before deploy — I mostly tested on text-based posts since those constitute most of these editor styles anyway.

allancole:
Found an issue in backwards compatibility for the Buttons Block:

image

allancole:
Posts List Block missing fonts and button styles:

Editor|Frontend
------------|------------
image|image

allancole:
Image alignment inconsistencies and caption fonts missing.

Editor|Frontend
------------|------------
image|image

allancole:
@kjellr I’m finding tons of issues in Dyad that mostly point back to this theme not using `add_editor_style()`. Most of the issues I’m finding have to do with overrides not working properly because of it. I’m getting the sense that fixing all the issues block-by-block will take longer if we don’t switch to `add_editor_style()` first and then when we do switch to use `add_editor_style()` we’ll have to undo these fixes anyway. Gonna do a quick PR to see if I can get further, faster that way. If not, I’ll come back and do the block-by-block thing on this PR.

kjellr:
Ok, sounds good. Thanks @allancole!

kjellr:
Closing in favor of #2034.

Issue #2019, opened 2020-05-14
Dyad 2: Social Links


jeffikus:
## Steps to replicate

Create a page using the Contact page template or the About page template.

## Result

Social icons are vertically stacked https://cloudup.com/cE8afU7icFx

## Expected

Social icons should be horizontally aligned.

kjellr:
Looks like Dyad 2 adds a `display: inline` rule to `alignleft` blocks. This overrides the `flex` style for the social icons block. I'm not sure that the rule is needed.

kjellr:
Fixed in #2034

Issue #2018, opened 2020-05-14
Dyad 2: Block alignment for buttons


jeffikus:
## Steps to replicate

I'm not sure if this is just a button issue, but if you create a button and add center alignment, it doesn't center.

## Result

https://cloudup.com/cQR3D6b0HAW

## Expected

I expected the button to be in there center of the page. This happens on the Services and Team page templates.

kjellr:
Fixed in #2034

PR #2034, opened 2020-05-15
Dyad 2: Refactor editor styles implementation

allancole:

This PR revises the solution in #2015 to add Gutenberg support via the recommended `add_editor_style()` function. In doing so, the editor selectors can be simplified and easier to maintain with changes in Gutenberg. 

ToDo:

- [x] Add color utility classes for the editor color palette ([see here](https://github.com/Automattic/themes/blob/12d7dec6190c917d9522801fd2fec13435863f37/dyad-2/functions.php#L103-L140))
- [x] More testing to make sure all the issues mentioned on #2015 are corrected.
- [x] Code cleanup

#### Related issue(s):
#2015

### DIFF
--- a/dyad-2/css/blocks.css
+++ b/dyad-2/css/blocks.css
@@ -34,12 +34,34 @@ Description: Used to style Gutenberg Blocks.
 		margin-right: -20%;
 		width: auto;
 	}
+
+	body:not(.has-post-thumbnail) .alignfull .wp-block-group__inner-container .alignwide {
+		margin-left: auto;
+		margin-right: auto;
+		max-width: calc(700px + 40%);
+	}
+
+	body:not(.has-post-thumbnail) .wp-block-group__inner-container .alignwide {
+		margin-left: auto;
+		margin-right: auto;
+	}
+
+	body:not(.has-post-thumbnail) .wp-block-group__inner-container > *:not(.alignwide):not(.alignfull) {
+		margin-left: auto;
+		margin-right: auto;
+		max-width: 700px;
+	}
 }
 
 body:not(.has-post-thumbnail) .alignfull {
 	margin-left: calc(50% - 50vw);
 	margin-right: calc(50% - 50vw);
 	width: auto;
+}
+
+body:not(.has-post-thumbnail) .wp-block-group__inner-container .alignfull {
+	margin-left: 0;
+	margin-right: 0;
 }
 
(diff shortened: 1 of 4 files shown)


kjellr:
Migrating these over from the other ticket: 

> Found an issue in backwards compatibility for the Buttons Block:
> 
> Editor|Frontend
> ------------|------------
> image|image
> 

I don't think we actually need to solve for this one, as the broken code there was only present in the 7.9.1 release, which was never deployed to .com. 

---

> Posts List Block missing fonts and button styles:
> 
> Editor|Frontend
> ------------|------------
> image|image
> 

I'm not seeing those buttons in my testing, but the fonts are definitely fixed in this PR. 

---

> Image alignment inconsistencies and caption fonts missing.
> 
> Editor|Frontend
> ------------|------------
> image|image
> 

I think these are still issues. I also noticed something similar with the Gallery Block on the front end.

dyad-2/css/blocks.css
```diff
@@ -34,6 +34,23 @@ Description: Used to style Gutenberg Blocks.
 		margin-right: -20%;
 		width: auto;
 	}
+
+	body:not(.has-post-thumbnail) .alignfull .wp-block-group__inner-container .alignwide {
```
kjellr, line 38: Do we need similar styles for cover + columns blocks?

kjellr:
Just tested, and this fixes the button issue in #2018

kjellr:
Just pushed a93cb72, which fixes #2019, and doesn't appear to cause any unintended issues in my testing.

dyad-2/css/blocks.css
kjellr: Answer: Yes. 😄 I'll work on adding that in.

allancole: Oh shoot! Forgot to change the selector here. `.wp-block-group__inner-container` should be `*[class="__inner-container"]` so any new Blocks with nested styles gets solved here.

kjellr: Ah, ok cool. I can make that change. Columns are structured a little differently, but I'll sort that out too.

kjellr: Fixed in 94ee61e

allancole:
This PR is in a good spot and addresses the following issues. 

- Fixes #2015 
- Fixes #2019 
- Fixes #2018 

Gonna go ahead get this merged!

Issue #2045, opened 2020-05-19
Varia & Varia Child Themes: Latest post block post titles are overlapping in Grid View


itowhid06:
## Steps to replicate

If the **Latest Posts** block is set to Grid view and the column number is high like 5/6, the post titles are overlapping in both AMP & non-AMP versions.

## Result

**"Latest Posts" block in Grid view(AMP Paired browsing mode)**

![image](https://user-images.githubusercontent.com/12055657/82325360-f2a61380-99fc-11ea-82c6-cf60013ea93d.png)

## Expected

Post titles should not overlap each other in any view.

PR #2046, opened 2020-05-19
Varia: Fix overlapping issue in latest post block

itowhid06:
Fixes https://github.com/Automattic/themes/issues/2045

If the **Latest Posts** block is set to Grid view and the column number is high like 5/6, the post titles are overlapping in both AMP & non-AMP versions. This PR aims to fix this by wrapping the post titles.

**"Latest Posts" block in Grid view in AMP Paired browsing mode (Before Fix)**

image

**"Latest Posts" block in Grid view in AMP Paired browsing mode (After Fix)**

image

### DIFF
--- a/varia/sass/blocks/latest-posts/_style.scss
+++ b/varia/sass/blocks/latest-posts/_style.scss
@@ -2,6 +2,7 @@
 	padding-left: 0;
 
 	& > li {
+		word-wrap: break-word;
 		/* Vertical margins logic */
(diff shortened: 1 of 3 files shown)


kjellr:
Just a planning note for @allancole: To save us some time spent re-compiling all the child themes, I'm going to hold off on deploying this for all the child themes until #2145 and #2077 are finalized.

Issue #2103, opened 2020-06-05
Rivington: Table with Stripes style texts are not visible


itowhid06:
## Steps to replicate

1. Go to Block Editor and add a Table block, Select `Stripes` style and see the text is not visible in white background

## Result
**Table with `Stripes` style in Block Editor**

![image](https://user-images.githubusercontent.com/12055657/83749105-ff756900-a684-11ea-9bef-45e264673290.png)

**Table with `Stripes` style in Frontend**

![image](https://user-images.githubusercontent.com/12055657/83749199-26339f80-a685-11ea-9793-de795e1163c6.png)

## Expected

Text should be visible regardless of the style.

kjellr:
Thanks, @itowhid06.

@allancole I think this will be an issue for _all_ Varia themes once custom color annotations land.

PR #1966, opened 2020-05-01
Rivington: Add Color Annotations

kjellr:
Follows the mold of #1944 to add basic color annotations for Balasana. 

Screen Shot 2020-04-30 at 20 08 21
Still needs: 

- [x] Refinement around contrast. 
- [x] Double checking mobile styles.
- [x] Add featured color palettes.

PR #2104, opened 2020-06-05
Rivington: Fix text showing issue in tables with Stripes style

itowhid06:
Fixes https://github.com/Automattic/themes/issues/2103

This PR adds styling for tables with `Stripes` style in both Block editor and Frontend, so that the texts are visible.

**After applying Fix in Block Editor**
image

**After applying Fix in frontend**
image

#### Related issue(s):
#2103

### DIFF
--- a/rivington/sass/_extra-child-theme.scss
+++ b/rivington/sass/_extra-child-theme.scss
@@ -345,6 +345,22 @@ a {
 	}
 }
 
+// Table with Stripes style
+.wp-block-table.is-style-stripes {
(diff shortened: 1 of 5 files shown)


allancole:
Hi @itowhid06. Thanks for contributing again! It looks like this issue is going to get fixed in a related PR (See #1966 and 91962ae) so that it can address both the stripes option and the extra background table colors in Gutenberg 5.4.2. 

Going to close this in favor of that PR but thanks for the help as always :-)

PR #2117, opened 2020-06-11
Canard: Remove reference to HEADER_TEXT

### LINKED ISSUE #2116: Canard: Stop checking for HEADER_TEXTCOLOR in in/custom-header.php
This fails the ThemeCheck plugin's check, and prevents the theme from being updated on WordPress.org. 

```
REQUIRED: HEADER_TEXTCOLOR was found in the file custom-header.php. Use add_theme_support( 'custom-header' ) instead.
Line 37: // get_header_textcolor() options: HEADER_TEXTCOLOR is default, hide text (returns 'blank') or any hex value
Line 38: if ( HEADER_TEXTCOLOR == $header_text_color ) {
```

https://github.com/Automattic/themes/blob/773d1d70623ee503a677a1e18ab0315d22132a26/canard/inc/custom-header.php#L36-L40

The recommendation there doesn't look right, but we need to stop checking for `HEADER_TEXTCOLOR` there and find another route for verifying if that no custom header text options are set if we want to pass this check.

jffng:
This PR addresses #2116. There should be no visual changes.

**How to Test**

- Install Canard.
- Change the header color and verify that it works.
- Install ThemeCheck plugin.
- Run the check and verify that there are no errors.

Tested on:
WordPress 5.4.2
PHP 7.3.5

### DIFF
--- a/canard/inc/custom-header.php
+++ b/canard/inc/custom-header.php
@@ -33,10 +33,8 @@ if ( ! function_exists( 'canard_header_style' ) ) :
 function canard_header_style() {
 	$header_text_color = get_header_textcolor();
 
-	/*
-	 * If no custom options for text are set, let's bail.
-	 * get_header_textcolor() options: Any hex value, 'blank' to hide text. Default: add_theme_support( 'custom-header' ).
-	 */
+	// If no custom options for text are set, let's bail.
+	// get_header_textcolor() options: Any hex value, 'blank' to hide text. Default: add_theme_support( 'custom-header' ).
 	if ( get_theme_support( 'custom-header', 'default-text-color' ) === $header_text_color ) {
 		return;
 	}
--- a/canard/inc/custom-header.php
+++ b/canard/inc/custom-header.php
@@ -33,10 +33,11 @@ if ( ! function_exists( 'canard_header_style' ) ) :
 function canard_header_style() {
 	$header_text_color = get_header_textcolor();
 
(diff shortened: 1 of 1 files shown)


canard/inc/custom-header.php
```diff
@@ -33,9 +33,10 @@ function canard_custom_header_setup() {
 function canard_header_style() {
 	$header_text_color = get_header_textcolor();
 
+	preg_match( '/^(?:[0-9a-fA-F]{3}){1,2}$/', $header_text_color, $hex_match );
```
kjellr, line 36: Would you mind adding a comment here to say what this does?

kjellr (approved):
This seems to work fine on my end. Just left a very minor comment. @allancole would you mind taking a quick look too just to be sure? 

The general concept of this is that if the only other potential returned values of `get_header_textcolor()` are a hex color or `blank`, we can flip this conditional so that it checks against those and not against `HEADER_TEXTCOLOR`.

allancole:
The regex might work but this solution looks a little cleaner and should still pass the checks,  @kjellr @jffng :-)  

https://github.com/Automattic/themes/blob/465f1d0de2f07eeb3d82f365ad372fe5bbd3f201/dara/inc/custom-header.php#L30-L39

kjellr:
I updated to use the same method as [the example](https://github.com/Automattic/themes/pull/2117#issuecomment-642785987) that @allancole shared. 

Would you two mind taking another look?

allancole (approved):
Yup, that works for me! 

LGTM!

Issue #2142, opened 2020-06-17
Seedlet: Default hover colors aren't correct unless you have the Customizer open


kjellr:
I think we’ve still got some weirdness in our Seedlet custom colors defaults logic. It looks like the default hover states aren’t right unless you have the customizer open. :thinking_face: 

GIF: 

![colors](https://user-images.githubusercontent.com/1202812/84945485-53d10d80-b0b5-11ea-9f62-7a5fc0e49ccb.gif)

Note that the default for the theme should be green text on hover. This is right when the Customizer is open. But when the customizer is closed, the hover color appears to be gray instead.

cc @allancole @jffng

Issue #2141, opened 2020-06-17
Seedlet: Fix array to string conversion error in class-seedlet-custom-colors.php


kjellr:
On a fresh activation with `wp_debug` active, Seedlet throws the following error: 

> <b>Notice</b>:  Array to string conversion in <b>/Users/kjell/Sites/valet/dotcomthemes/wp-content/themes/dotcom/seedlet/classes/class-seedlet-custom-colors.php</b> on line <b>216</b>

Here's the line in question: 

https://github.com/Automattic/themes/blob/30d597b799075d5d9643eb279d43710ad9308f50/seedlet/classes/class-seedlet-custom-colors.php#L216

![Screen Shot 2020-06-17 at 4 03 51 PM](https://user-images.githubusercontent.com/1202812/84944673-23d53a80-b0b4-11ea-94b4-8dbf74c1202f.png)

jffng:
I should have a fix for this + #2142 up shortly.

PR #2143, opened 2020-06-17
Seedlet: fixes logic and warnings around custom colors

jffng:
Addresses #2141 and #2142.

### DIFF
--- a/seedlet/classes/class-seedlet-custom-colors.php
+++ b/seedlet/classes/class-seedlet-custom-colors.php
@@ -231,7 +231,6 @@ class Seedlet_Custom_Colors {
 				}
 
 				if ( $theme_mod_variable_name === '--global--color-secondary' && $theme_mod_default_color !== $theme_mod_custom_color ) {
-					$theme_css .= "--global--color-primary-hover: " . $adjusted_color . ";"; // Since the default palette uses --global--color-secondary as its primary hover color, we need to change the primary hover color when a user selects a different secondary color
 					$theme_css .= "--global--color-secondary-hover: " . $adjusted_color . ";";
 				}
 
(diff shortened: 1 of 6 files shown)


kjellr:
This doesn't _quite_ work in my testing. It does get hovers working correctly by default, and also surpasses the warnings, but it also keeps the green hovers even when custom colors are active. Here's an example where I've hovered over the site title:  

Screen Shot 2020-06-18 at 8 43 36 AM

jffng:
> it also keeps the green hovers even when custom colors are active

@kjellr I _think_ I addressed this with my latest commit, could you give it another test? 

Basically, what's happening is: 
- user chooses a custom **secondary** color and expects the link hover color to change. 
- our link hover styles use `--global--color-primary-hover`, not `--global--color-secondary-hover`.
- since the primary color has not changed, its hover color didn't either (according to our existing custom color logic). 

I can think of two ways to address this:
1. Change the custom color logic to also change `--global--color-primary-hover` when a user selects a custom secondary color. [I tried this first](https://github.com/Automattic/themes/pull/2143/commits/631c10d540cf51f0b7be5463bbe0359f99db76bb).
2. Change the `--global--color-primary-hover` to be the value of `--global--color-secondary`. This is what's reflected in the latest commit.

I like the second option because it reflects how we actually use the secondary color in the default scheme. I may be missing something here though.

seedlet/classes/class-seedlet-custom-colors.php
```diff
@@ -212,14 +212,15 @@ function seedlet_generate_custom_color_variables( $context = null ) {
 			$theme_css = ':root {';
 		}
 
+		$theme_mod_bg_color = empty( get_theme_mod( "seedlet_--global--color-background" ) ) ? '#FFFFFF' : get_theme_mod( "seedlet_--global--color-background" );
```
kjellr, line 215: Would you add a code comment here to clarify what's happening?

PR #2153, opened 2020-06-18
Restore foreground color rules

kjellr:
#1966 and #1944 made some changes to foreground color rules when a preset background color was selected. This led to existing sites changing appearance, even when custom colors were not active. 

This PR restores some of those rules, in order to restore the previous appearance. It also modifies the custom color rules to properly override these changes. 

**If this looks right, we should apply these same changes to Varia as well as to Coutoire and Alves. I can spin up PRs for those after this one lands.**

## Screenshot

The following screenshots were taken with the Rivington default homepage.

Before (Default Colors)

kjellreigstad wordpress com_ (6)

After (Default Colors)

kjellreigstad wordpress com_ (2)

---

Before (Custom Colors)

kjellreigstad wordpress com_ (5)

After (Custom Colors)

kjellreigstad wordpress com_ (7)

### DIFF
--- a/rivington/inc/wpcom-colors.php
+++ b/rivington/inc/wpcom-colors.php
@@ -278,6 +278,11 @@ add_color_rule( 'txt', '#f2f2f2', array(
 	array( '.has-foreground-color[class]', 'color' ),
 	array( '.has-foreground-background-color[class]', 'background-color' ),
 
+	// Text color
+	array( '.has-background-background-color[class],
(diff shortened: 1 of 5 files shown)


ianstewart:
I tested a patch on my sandbox that fixes this …

image

And it's now working as expected.

kjellr:
Thank you! This is deployed, and I'll work on fixing up Varia, Coutoire, and Alves too.

PR #2163, opened 2020-06-22
Seedlet: Add focus styles

### LINKED ISSUE #2140: Seedlet: Add focus styles
It appears that the Seedlet theme is missing focus styles. Currently in testing, the theme just falls back to the default browser styles: 

Screen Shot 2020-06-17 at 3 56 25 PM

We should replace those with the version from the theme mockups: 

Screen Shot 2020-06-17 at 3 56 51 PM

(It looks like we're also missing the double-tall border for hover, as well as the active state. 😄 So we should consider adding those too)

kjellr:
Fixes #2140.

Seedlet was missing custom focus styles. This PR adds those in, using the design in the original design comps. 

84944006-2be0aa80-b0b3-11ea-9f05-ec61ad19f2d6

It also cleans up the active styles.

Still todo: 

- ~~Add the 2px bottom border on hover/focus.~~

## Screenshots:

Before (Default browser focus styles):

focus-before

After: 

focus-after

### DIFF
--- a/seedlet/assets/css/style-editor.css
+++ b/seedlet/assets/css/style-editor.css
@@ -113,8 +113,12 @@ a {
 	text-decoration: none;
 }
 
-a:hover {
+a:hover, a:focus {
 	color: var(--global--color-primary-hover);
+}
+
+a:active {
+	color: var(--global--color-primary);
 }
 
 button,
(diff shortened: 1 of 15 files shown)


kjellr:
> Still todo: 
> - ~~Add the 2px bottom border on hover/focus.~~

Update: I tried this out, and it seemed _really_ weird in practice. 😂 So this PR should be ready to go without it.

PR #2145, opened 2020-06-18
Morden: Fix Gallery Alignment Issue causing from setting max-width in desktop

itowhid06:
Fixes https://github.com/Automattic/themes/issues/2144.

#### Changes proposed in this Pull Request:
This PR fixes the image gallery alignment issue in Morden theme after applying https://github.com/Automattic/themes/pull/2077.

**Gallery with left alignment Before Fix**

image

**Gallery with Left alignment After Fix**
image

This PR also fixes the similar issue with right aligned gallery.

#### Related issue(s):
https://github.com/Automattic/themes/issues/2065

PR #2077, opened 2020-06-02
Varia & Child Themes: Fix Image Gallery alignment issues

itowhid06:
Fixes https://github.com/Automattic/themes/issues/2065
#### Changes proposed in this Pull Request:
This PR fixes the image gallery alignment issues.

**Gallery with Right alignment Before Fix**

image

**Gallery with Left alignment Before Fix**
image

**Gallery with Right alignment After Fix**

image

**Gallery with Left alignment After Fix**
image

This PR also fixes the alignment issues whenever a alignful Gallery is followed by a left or right aligned Gallery.

**A Post with different Gallery Alignment after applying this fix**

**In the Block Editor**

image

**In the Frontend**

image

#### Related issue(s):
https://github.com/Automattic/themes/issues/2065

PR #2166, opened 2020-06-22
Varia Child Themes: Recompile child themes with the latest changes to Varia.

allancole:
This PR recompiles all Varia child themes against #2046 and all the latest changes to Varia.

Before merging this, we need #2145 and #2077 to be merged first and then run another recompile against those.


allancole:
Merging this now that #2145 and #2077 have landed :-)

PR #2167, opened 2020-06-22
Seedlet Blocks: Add basic global styles support

kjellr:
Adds basic global styles support for colors and font sizes to Seedlet Blocks. It accomplishes this by: 

- Adding an `experimental-theme.json` file, [using the spec here](https://developer.wordpress.org/block-editor/developers/themes/theme-json/). 
- It adds presets for the colors and font sizes used in the theme. 
- It adds CSS to the editor and to the front-end to map the Seedlet variables to the ones generated by `experimental-theme.json`. 

At this point, nothing at all should change visibly when this patch is applied. 😄 

To verify this is working, modify one of the values in `experimental-theme.json`, and verify that the update is reflected in the editor and on the front end.

### DIFF
--- a/seedlet-blocks/experimental-theme.json
+++ b/seedlet-blocks/experimental-theme.json
@@ -1,0 +1,58 @@
+{
+	"global": {
+		"presets": {
+			"font-size": [
+				{
+					"slug": "tiny",
+					"value": "14px"
+				},
+				{
+					"slug": "small",
+					"value": "16px"
+				},
+				{
+					"slug": "medium",
+					"value": "90px"
+				},
+				{
+					"slug": "large",
+					"value": "24px"
+				},
+				{
+					"slug": "extra-large",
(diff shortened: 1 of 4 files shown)


seedlet-blocks/experimental-theme.json
```diff
@@ -0,0 +1,58 @@
...
+					"slug": "tiny",
+					"value": "14px"
+				},
+				{
+					"slug": "small",
+					"value": "16px"
+				},
+				{
+					"slug": "medium",
+					"value": "90px"
```
jffng, line 15: Did you mean for this to be `90px`? 😄 

```suggestion
					"value": "18px"
```

kjellr:
> Left one small comment but this works as advertised.

Thanks for finding that — I was testing to make sure it worked. 😂

mtias:
Cool to see this :)

Issue #2171, opened 2020-06-23
Seedlet: Build IE11 fallback for CSS variables


kjellr:
In the mold of https://github.com/Automattic/themes-workspace/pull/15. 

Will most likely use a postcss plugin to find all custom properties and turn them into static values, placing a "duplicate" style rule before the custom property.

jffng:
After doing some more research, I can think of two approaches to doing this.

1. Have a single stylesheet that contains fallbacks for IE11. The benefit to this is that the front-end styles remain a single stylesheet; there is no conditional loading of a secondary script necessary. The drawback is that the stylesheet grows substantially in size. 

2. Have an IE11 specific stylesheet. The benefit to this is that we can maintain a much smaller stylesheet for most users. The drawback is that we have to conditionally load an IE11-specific stylesheet. I haven't come across a clean example doing this yet.

kjellr:
> Have an IE11 specific stylesheet. The benefit to this is that we can maintain a much smaller stylesheet for most users. The drawback is that we have to conditionally load an IE11-specific stylesheet. I haven't come across a clean example doing this yet.

This seems like a much better solution if we can swing it — there's no need for most folks to load all those extra styles.

kjellr:
Reopening this issue, because I _think_ we need to also bundle in an IE-specific version of `style-editor.css`.

jffng:
>  I _think_ we need to also bundle in an IE-specific version of style-editor.css.

To do this, we have to bundle the editor styles + variables together so that the plugin can process the stylesheet correctly. I'm not sure how this will impact our custom color logic, since [we enqueue the editor variables separately in our custom color class](https://github.com/Automattic/themes/blob/a52a6aa758e06aeaefc7ea51c1cdfbce13a7bf2c/seedlet/classes/class-seedlet-custom-colors.php#L267-L272), but I can try it out.

scruffian:
We did this with a pony fill in the end, so I think we can close this.

PR #2177, opened 2020-06-24
Seedlet: CSS Variables fallback support for IE11

jffng:
Addresses #2171. 

There should be no visual changes to theme as a result of this PR. 

We do need a way to verify that it works on IE11 — how have we done this in the past? Via Browserstack?


seedlet/package.json
```diff
@@ -36,17 +38,15 @@
...
+    "build:variables-editor": "node-sass assets/sass/variables-editor.scss assets/css/variables-editor.css --output-style expanded --indent-type tab --indent-width 1",
+    "build:style": "node-sass assets/sass/style.scss style.css --output-style expanded --indent-type tab --indent-width 1",
+    "build:style-editor": "node-sass assets/sass/style-editor.scss assets/css/style-editor.css --output-style expanded --indent-type tab --indent-width 1",
+    "build:woocommerce": "node-sass assets/sass/style-woocommerce.scss assets/css/style-woocommerce.css --output-style expanded --indent-type tab --indent-width 1",
     "build:woocommerce-rtl": "rtlcss assets/css/style-woocommerce.css assets/css/style-woocommerce-rtl.css",
     "build:rtl": "rtlcss style.css style-rtl.css",
-    "build:print": "node-sass assets/sass/print.scss assets/css/print.css --output-style expanded --indent-type tab --indent-width 1 && postcss -r assets/css/print.css",
-    "build": "run-p \"build:*\"",
+    "build:print": "node-sass assets/sass/print.scss assets/css/print.css --output-style expanded --indent-type tab --indent-width 1",
+    "ie": "postcss style.css -o style.css",
```
jffng, line 48: This script should probably be renamed.

seedlet/assets/sass/style.scss
```diff
@@ -69,6 +69,9 @@ Source: https://www.si.edu/object/hymenoxys-odorata-dc:nmnhbotany_13045011
 Included in theme screenshot.
 */
 
+// Variables
+@import "variables";
```
jffng, line 73: For the postcss plugin to work, the variables and styles need to live in the same file.

It made me wonder — why are the front-end variables and stylesheet enqueued separately?

kjellr:
> We do need a way to verify that it works on IE11 — how have we done this in the past? Via Browserstack?

Yeah, browserstack. Or else you can load up a VM using one of these: https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/

seedlet/assets/sass/style.scss
allancole: > why are the front-end variables and stylesheet enqueued separately?

I was thinking it might be easier for child themes to _replace_ the variables file that way, but it may not be necessary.

jffng:
Ok, I think this is in a good spot for feedback!

I was able to conditionally load the IE-specific stylesheet, as well as flatten the stylesheet so the responsive styles are mostly maintained. I tested this using BrowserStack and overall it looks pretty good:

Screen Shot 2020-06-29 at 1 15 41 PM

seedlet/assets/sass/_variables.scss
```diff
@@ -1,6 +1,4 @@
-/**
- * These styles should be loaded on the frontend-only
- */
+// These styles should be loaded on the frontend-only
```
jffng, line 1: I changed the style of comments here (`/**` to `//`) so they wouldn't be included in the compiled stylesheets.  
 Including the variables along with the comments made the beginning of the stylesheet look quite cluttered / difficult to read.

seedlet/functions.php
```diff
@@ -345,11 +345,14 @@ function seedlet_scripts() {
 	// Enqueue Google fonts
 	wp_enqueue_style( 'seedlet-fonts', seedlet_fonts_url(), array(), null );
 
-	// Theme variables
-	wp_enqueue_style( 'seedlet-variables-style', get_template_directory_uri() . '/assets/css/variables.css', array(), wp_get_theme()->get( 'Version' ) );
-
 	// Theme styles
-	wp_enqueue_style( 'seedlet-style', get_template_directory_uri() . '/style.css', array(), wp_get_theme()->get( 'Version' ) );
+	if (preg_match('~MSIE|Internet Explorer~i', $_SERVER['HTTP_USER_AGENT']) || (strpos($_SERVER['HTTP_USER_AGENT'], 'Trident/7.0; rv:11.0') !== false)) {
```
kjellr, line 349: Have you looked into one of the options suggested over here? https://gist.github.com/wpscholar/4947518#gistcomment-1669140

It looks like some of the Twenty- themes use a similar approach:

https://github.com/WordPress/twentyseventeen/blob/fa09211f54799dd03e8acda9d19108aafb1a8c77/functions.php#L286-L288

jffng: Unfortunately IE10+ don't support conditional comments, otherwise we could use that approach.

kjellr: Ah ok, I see. What we have here is a bit lengthy, but I understand why it's necessary then.

jffng:
Great, let's see how this works in practice. Two things to keep our eyes on are:

1. if bundling the variables with the main stylesheet has any implications for child-theme development
2. if the development scripts should be modified because running postcss on every build could be annoying

PR #2196, opened 2020-06-29
Seedlet: Add a "hide homepage title" option to the customizer

kjellr:
Adds a "Hide homepage title" option to the customizer. I mostly copied/pasted this from Varia. This ended up being quite a lot of code — is this how you usually do it, @allancole?

You'll need to test this in a WP.com sandbox.

## Screenshots

Screen Shot 2020-06-29 at 3 28 29 PM
Screen Shot 2020-06-29 at 3 27 52 PM
Screen Shot 2020-06-29 at 3 28 16 PM


kjellr:
Also, this update adds a number of these new files in `/inc`: 

- `customize-preview-wpcom.js`
- `style-editor-wpcom.css`
- `style-wpcom.css`
- `wpcom.php`

Some include the `-wpcom` suffix, which is helpful. But at a glance, it’s still hard to tell which files are for WP.com, and which aren’t. I wonder if it would be possible to move some of these into their own `/inc/wpcom` folder (and/or change that suffix to a prefix). @allancole do you have any thoughts on that? Would either of those break anything on the dotcom of things?

seedlet/inc/wpcom.php
```diff
@@ -0,0 +1,123 @@
...
+ */
+function seedlet_wpcom_body_classes( $classes ) {
+
+	$hide = get_theme_mod( 'hide_front_page_title', false );
+
+	if ( true === $hide ) {
+		$classes[] = 'hide-homepage-title';
+	}
+
+	$credit_option = get_option( 'footercredit' );
```
jffng, line 78: This is if the user is on a business plan and has selected "Hide", we add a class to hide the comma?

Screen Shot 2020-06-30 at 10 29 35 AM

allancole:
After discussing this briefly with Kjell, I think it makes a lot of sense to move all the wpcom specific files in a single `/wpcom` directory. This might require changing some of the dependencies on wpcom to look inside of that directory, but I think it would be worth it. I also think the “change that suffix to a prefix” route could be a good interim solution as we could keep   `inc/wpcom.php` where it is and have all of its related files listed next to it.

seedlet/inc/wpcom.php
kjellr: Yep, that's right.

kjellr:
> I also think the “change that suffix to a prefix” route could be a good interim solution as we could keep inc/wpcom.php where it is and have all of its related files listed next to it.

Ok, this should be ready for a final review. I've added a `wpcom-` prefix to all fo these files so that the WP.com-specific files are all easy to spot.

PR #2208, opened 2020-07-01
Add support for custom link color.

### LINKED ISSUE #2187: Seedlet: Add support for the custom link color
As added in Gutenberg: https://github.com/WordPress/gutenberg/pull/22722

kjellr:
Closes #2187 

Screen Shot 2020-07-01 at 9 55 05 AM


seedlet/assets/sass/abstracts/_config.scss
```diff
@@ -65,4 +65,7 @@ $typescale-ratio: 1.2; // Run ratio math on 1em == $typescale-base * $typescale-
 	--global--border-radius-lg: #{$typescale-root};
 	--global--border-radius-pill: #{10 * $typescale-root};
 	--global--border-radius-none: 0;
+
+	/* WordPress Variables */
+	--wp--style--color--link: var(--global--color-primary);
```
kjellr, line 70: Defining this here seems a little weird to me since it doesn't match our other rule naming, but I wasn't sure of how to handle this otherwise — we need to define this value otherwise we'll end up with the WordPress default. 

I also needed to switch to using the `var(--wp--style--color--link)` color in our `a` rules to get this to work. I guess that's ok, since we'll need to do that in the future anyway.

seedlet/assets/sass/base/_editor.scss
```diff
@@ -19,7 +19,7 @@ body {
 // Links styles
 a {
 	border-bottom: 1px solid var(--global--color-secondary);
-	color: var(--global--color-primary);
+	color: var(--wp--style--color--link);
```
jffng, line 22: As an alternative to defining it in our abstracts, something like this should work?

```suggestion
	color: var( --wp--style--color--link, var(--global--color-primary) );
```

Not sure how much cleaner this is, but maybe it feels more contained and consistent?

seedlet/assets/sass/abstracts/_config.scss
jffng: Left a suggestion below that wouldn't impact this file.

But yeah, this gets at a discussion of how we will eventually handle mapping of presets from the theme config to the set of design tokens.

kjellr: Yeah. For the block-based version of this, I bundled those in a totally separate stylesheet like this: 

https://github.com/Automattic/themes/blob/feaf746c0f6d1ef9297cce069ff7a22c4eb79e90/seedlet-blocks/style.css#L33-L49

That was really clean, but probably doesn't make sense here.

seedlet/assets/sass/base/_editor.scss
kjellr: Yeah, that's probably cleaner. I'll give it a shot.

kjellr:
Ok this is ready for another look. 👍

seedlet/assets/sass/abstracts/_config.scss
jffng, line 70: Sorry my previous comment wasn't clear.  By changing to `var( --wp--style--color--link, var(--global--color-primary) );` in the actual link styles, I believe this mapping becomes unnecessary.
```suggestion
```

kjellr: Ah, you're right. That does make sense for now.

kjellr:
> When you apply a background color via the editor, the paragraph styles are too specific and override the link color

Yuck, you're right. Hmm. I did some light digging into fixing this, but it involves a _lot_ of extra `:not()` rules. I'll need to give it a fresh look tomorrow.

kjellr:
Ok, @jffng — I got the background color bug sorted out. Rather than write a lot of extra `:not()` rules throughout the theme, I elected to write just [one very specific rule](https://github.com/Automattic/themes/blob/d08db9960b88f5e1e9d9650d461d3ef6caa3e1a9/seedlet/assets/sass/base/_reset.scss#L96-L100). That seems to take care of it in my testing: 

Screen Shot 2020-07-09 at 7 27 40 PM

Mind giving this another spin?

jffng:
Nice! Works on my end. 🚢

Issue #2125, opened 2020-06-15
Seedlet: Add starter content


jffng:
The theme will need starter + headstart content.

ianstewart:
@jffng ping me when you have the seedletdemo.wordpress.com site set up just right and I can walk through what I remember knowing about how Headstart works. (It's relatively simple to generate annotations.)

jffng:
With the help of @ianstewart and some good docs, I figured out how to write / test the headstart annotations. 

But it brings up some questions:
- What dotcom flows today trigger headstart? In other words, which users will see this content?
- With headstart, should we re-create the demo site content, or do something more generic like populate the site with existing dotcom page layouts e.g. About, Contact.

ianstewart:
> What dotcom flows today trigger headstart?

- Signup from theme logged-out theme showcase
- The current "update your homepage on theme switch" feature (which we intend to pause and improve on dotcom)
- HE tools for setting up sites to look like the demos
- At one time signup used it as well

> With headstart, should we re-create the demo site content, or do something more generic like populate the site with existing dotcom page layouts e.g. About, Contact.

My personal opinion is that there's an art to it. A theme demo site should have "just enough" content to show off the theme and look like a real website. If sub-pages help that, yes. I think it's better than sub-pages that show off various elements like a test of the stylesheet.

With the "botany" demo I might have a nice blog on the homepage, maybe an about, and contact page.

PR #2209, opened 2020-07-01
Seedlet: Update headstart annotations

jffng:
This PR partially addresses #2125 to include headstart annotations. It adds a home page layout and one blog post to the user's site, as well as setting the primary and social navigation. 

To test:
- From your sandbox, run: `wp headstart run --user=[dotcom username] --url=[your Headstart test site] --theme=pub/seedlet`
- Visit the site and verify that the headstart content has populated your test site


seedlet/inc/headstart/en.json
```diff
@@ -74,7 +60,7 @@
             "hs_custom_meta": "_hs_front_page",
             "hs_sharing": 1,
             "hs_like_status": 1,
-            "attachment_url": "https:\/\/headstartdata.files.wordpress.com\/2016\/06\/skyline-buildings-new-york-skyscrapers.jpg",
+            "attachment_url": "https://seedletdemo.files.wordpress.com/2020/06/ec867-plants_2x2.png",
```
kjellr, line 63:
```suggestion
            "attachment_url": "https://headstartdata.files.wordpress.com/2020/07/envelope.jpg",
```

I've uploaded a version of this to the headstartdata site. 👍

seedlet/inc/headstart/en.json
```diff
@@ -60,12 +46,12 @@
             "hs_custom_meta": "_hs_first_post",
             "hs_sharing": 1,
             "hs_like_status": 1,
-            "attachment_url": "https:\/\/headstartdata.files.wordpress.com\/2016\/06\/pexels-photo-30732.jpg",
+            "attachment_url": "https://seedletdemo.files.wordpress.com/2020/07/envelope.png",
```
kjellr, line 49:
```suggestion
            "attachment_url": "https://headstartdata.files.wordpress.com/2020/07/envelope.jpg",
```

I've uploaded a higher-res version of this to the headstartdata site.

kjellr:
Cool — this is looking pretty good. I generated higher-resolution versions of a couple of these images, but I think I need to do more (I noticed they're low-res on the theme demo too 😕). I'll try to find some time to get those ready later today, so let's hold onto this PR for now.

ianstewart:
The homepage content "headstarted" as expected for me.

image

Looks like only [a social menu](https://github.com/Automattic/themes/blob/9b7a7ede6ca0a772971dfa644193d51b59bfd87e/seedlet/inc/headstart/en.json#L81) was added. I don't think it's a concern or blocker for how we currently use Headstart today though.

seedlet/inc/headstart/en.json
kjellr: Actually, I'm not sure it even makes a difference where the images come from. 🤷‍♂️

I've updated the theme demo, and will update this PR in just a moment.

kjellr:
Just pushed an update where I updated the demo content with fresh data from the demo site (including higher res, cleaned up images). Please note that I also removed the bottommost promo on the homepage from the demo, since I think the page flows a little more nicely without it.  

@jffng if you give this one more test, it should be ready to go.

jffng:
Nice, thanks @kjellr and @ianstewart 🚀

PR #2217, opened 2020-07-02
Seedlet: Try implementing Jetpack's Global Styles

kjellr:
This adds theme support for [Jetpack Global Styles](https://github.com/Automattic/wp-calypso/tree/master/apps/full-site-editing/full-site-editing-plugin/global-styles#how-it-works-and-how-themes-can-use-it), so users can customize fonts. 

This generally works alright in the front end, but it has an issue in there editor, because if variables are nested (for example, when the value of a variable is another variable) it appears that it won't update in real-time. So currently with this PR, you can make changes in the editor, but you won't see anything change at all until you visit the front end. 

I was able to _sort of_ fix this by doing a global find & replace where I swapped out stuff like `var(--global--font-primary)` for `var(--font-headings, --global--font-primary)`. But that was just the beginning... we'd also need to replace (for example) `var(--global--font-headings)`, since that maps to `var(--global--font-primary)` too. We'd been making pretty liberal use of nested variables, so it gets pretty messy. I didn't commit it yet, and I'm interested to see if @jffng or @allancole have any other ideas first. 

--

Screen Shot 2020-07-02 at 6 29 49 PM


seedlet/assets/css/ie.css
```diff
@@ -1279,7 +1279,7 @@ button {
 	color: #FFFFFF;
 	cursor: pointer;
 	font-weight: normal;
-	font-family: 'Fira Sans', Helvetica, Arial, sans-serif;
+	font-family: --global--font-secondary;
```
kjellr, line 1282: (The approach I have in here now also breaks this for IE, which isn't great)

jffng:
> if variables are nested (for example, when the value of a variable is another variable) it appears that it won't update in real-time. 

Could we write some JS to update the values of `--global--font-*` if a user selects a custom font?

kjellr:
I suppose it's possible. 🤔 I'm not sure where to get started with that, but if you have an idea, then feel free to explore.

jffng:
Hmm this is what I get when I check out this PR onto my test site: 

Screen Shot 2020-07-07 at 2 23 55 PM

Am I missing a step?

kjellr:
Hmm... I was getting that for a while, but then it started working. I'll give it another look...

kjellr:
I just re-tested, and it's working fine on my end. Do you have the newest version of [the plugin installed](https://wordpress.org/plugins/full-site-editing/)?

Screen Shot 2020-07-07 at 2 44 44 PM

kjellr:
Hmm interesting. Why do the variables need to be scoped to the `body` in the editor to work? Now that I'm looking at [the code here](https://github.com/Automattic/themes/blob/a52a6aa758e06aeaefc7ea51c1cdfbce13a7bf2c/seedlet/classes/class-seedlet-custom-colors.php#L267-L272) it seems really weird that we enqueue _all_ of the variables only in this custom colors class. 🤔 

Anyway, it sounds like if we can find a workaround for the custom colors, then we can just bundle the variables with our normal stylesheet, thus making the work in https://github.com/Automattic/themes/issues/2171 possible too. Is that right?

jffng:
> Why do the variables need to be scoped to the body in the editor to work? 

Allan explains it well here: https://github.com/Automattic/themes-workspace/issues/140#issuecomment-642165829

> it sounds like if we can find a workaround for the custom colors, then we can just bundle the variables with our normal stylesheet, thus making the work in #2171 possible too. Is that right?

I think so, I can give this a shot.

jffng:
I pushed a handful of changes that should unblock this:

- Bundle the editor variables with the editor stylesheet.
- Create an empty stylesheet to append custom color variables to. (an existing stylesheet is necessary to use `wp_add_inline_style`, which I believe is why you had enqueued the styles along with the custom color variable implementation) 

Can you give it a test @kjellr ?

kjellr:
That seems to work on my end, @jffng! Twenty Nineteen uses a similar [empty file](https://core.trac.wordpress.org/browser/trunk/src/wp-content/themes/twentynineteen/style-editor-customizer.css) approach for colors, so this isn't totally weird. 

I think the only potential blocker is that this breaks the font rules for IE. Any idea how to get those working?

jffng:
Looks like it was actually just a bug with the `var()` syntax. Fixed in [65d50ec](https://github.com/Automattic/themes/pull/2217/commits/65d50ec4d3d0f5a64ad487b1c4063996e7293efe), so this should be ready to land?

Note that this won't actually allow custom font selections in IE, but that's not supported by the plugin anyhow.

kjellr:
Hmm. I'm getting [this error again](https://github.com/Automattic/themes/pull/2217#issuecomment-655042280) when I run this on my test site. 🤔 Once I can get this up and running we should be good to go. Was there something you did to get it working for you before?

kjellr:
> I'm getting [this error again](https://github.com/Automattic/themes/pull/2217#issuecomment-655042280) when I run this on my test site.

I've tried on a fresh WP.com test site, and I'm seeing the same error. @allancole do you know if there's a step we're missing here? It's working fine locally.

kjellr:
> I'm getting [this error again](https://github.com/Automattic/themes/pull/2217#issuecomment-655042280) when I run this on my test site.

@nosolosw do you happen to know what could be going wrong here? This runs fine locally when using the full-site editing plugin, but when we test it on WordPress.com, it gives us that `You are customizing undefined / Your active theme doesn't support font pairings` error. Is there a WP.com-specific step we're missing?

nosolosw:
:wave: I know _what_ is happening but I don't know _why_. 

_What_ is happening is that the data the global styles sidebar uses doesn't get there. The request either 1) doesn't return any data or 2) the user doesn't have access to the endpoint. Would you go to your browser's tools, visit the network requests tab, search for `jetpack-global-styles` there and inspect the result of the request? When is functioning properly it looks like this (note how the data is present, for example, for the `blogname`):

Screenshot from 2020-07-16 12-01-26

If you share what you get, it'd be useful data to debug. In any case, this bug is unrelated to this PR and something specific to the dotcom setup and/or the testing sandbox. Any chance anyone from @Automattic/cylon can take a look at this?

kjellr:
😄 I figured it out — the request was giving a 403, because the WP.com API wasn't sandboxed. All set now. Thanks for the help, @nosolosw!

I'm going to go ahead and get this change deployed to WP.com. 👍

mtias:
Great to see this!

Issue #2212, opened 2020-07-01
Seedlet Blocks: Margins missing inside Layout Grid block


alaczek:
## Steps to replicate
1. With Seedlet Blocks active add Layout Grid block to a post or a page. Pick any number of columns.
2. Inside add a heading, followed by another heading or paragraph.
3. Save and preview. Notice the vertical margins on headings and paragraphs are missing.

Here's an example pattern using the Layout Grid block:
* editor view

![image](https://user-images.githubusercontent.com/4550351/86301073-33be5400-bc48-11ea-82dc-b771c004ede3.png)

* preview

![image](https://user-images.githubusercontent.com/4550351/86301105-4afd4180-bc48-11ea-9cd3-f2b45aa82205.png)

Tested in Chrome 83.0 / macOS 10.15.5

Issue #2211, opened 2020-07-01
Seedlet Blocks: Layout Grid columns not lining up horizontally in the editor


alaczek:
## Steps to replicate
1. With Seedlet Blocks active add Layout Grid block to a post or a page. Pick two or more columns configuration.
2. Notice how in the initial empty state of the block the first column outline is a little higher than others.
3. Add content to each column. The first column is still sitting a bit higher than others.
4. Save and preview. Notice that the issue does not affect the front end.

Editor - initial state of the Layout Grid block

![image](https://user-images.githubusercontent.com/4550351/86299613-12f3ff80-bc44-11ea-8787-cf938e7ffe05.png)

Editor - Layout Grid block with content added to each column

![image](https://user-images.githubusercontent.com/4550351/86299671-2e5f0a80-bc44-11ea-803b-c3e68935411b.png)

Page preview - no issues.

![image](https://user-images.githubusercontent.com/4550351/86299753-5babb880-bc44-11ea-9389-a65f2d40c7cd.png)

Tested in Chrome 83.0 / macOS 10.15.5

jffng:
This issue is caused by these spacing overrides:

https://github.com/Automattic/themes/blob/7d84a62a77863e35cab835d3ad34d1a8458f285b/seedlet/assets/sass/blocks/utilities/_editor.scss#L158-L174

@kjellr do you remember what these are for? Is it primarily to add some extra margins to every block in the editor to make them easier to select?

kjellr:
That's the regular top/bottom spacing supplied for all blocks. Gutenberg provides one, and Seedlet overrides it. We make that present on the front end too so the spacing matches up nicely: 

https://github.com/Automattic/themes/blob/9f03a09e56f0ccf50d8c38f06a17f4c472a7cf5d/seedlet/assets/sass/structure/_vertical-margins.scss#L131-L152

In this case, we may need to write special rules to make sure these apply properly in the layout grid block. Do we have this issue in the Columns block too? If not, we should look into why.

alaczek:
Nope, this is not happening in the Columns block.

PR #2219, opened 2020-07-03
Seedlet: Fix layout grid margins

jffng:
Addresses #2211 and #2212.

### Before

_Editor_
Screen Shot 2020-07-03 at 9 30 36 AM

_Front-end_
Screen Shot 2020-07-03 at 9 32 57 AM

### After

_Editor_
Screen Shot 2020-07-03 at 9 31 01 AM

_Front-end_
Screen Shot 2020-07-03 at 9 32 43 AM

### DIFF
--- a/seedlet/assets/css/ie.css
+++ b/seedlet/assets/css/ie.css
@@ -1112,32 +1112,32 @@ hr.wp-block-separator.is-style-wide {
  * - Sets spacing-vertical margin logic
  */
 .site-footer > * {
-	margin-top: 19.98px;
-	margin-bottom: 19.98px;
+	margin-top: 20px;
+	margin-bottom: 20px;
 }
 .site-main > article > * {
-	margin-top: 19.98px;
-	margin-bottom: 19.98px;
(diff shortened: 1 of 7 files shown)


alaczek:
I didn't dive into the code, but from my testing, it looks like this PR address both issues.

seedlet/assets/css/ie.css
```diff
@@ -6535,6 +6535,26 @@ button[data-load-more-btn],
 	padding-right: 20px !important;
 }
 
+.wp-block-jetpack-layout-grid .wp-block-jetpack-layout-grid-column > * {
+	margin-top: 19.98px;
```
kjellr, line 6539: What's up with these partial pixel measurements? Is that mirroring something from the Grid block?

jffng: Ah there is a postCSS plugin to transform `calc()` to static values wherever possible for IE. This is coming from a `calc( 0.666 * var(--global--spacing-horizontal) );`.

Issue #2126, opened 2020-06-15
Seedlet: Try removing .entry-content specificity for alignment rules


jffng:
Many of Seedlet's alignment rules only apply to blocks immediately within `.entry-content`.  Here's an example: 

https://github.com/Automattic/themes-workspace/blob/ed33753beb90d9c26e4385af9d103c813a91686b/seedlet/assets/sass/blocks/utilities/_style.scss#L87-L94

These all need to be rebuilt for use in block-based themes, where blocks appear outside of just `entry-content`. We should see if there's a more global way to declare these rules.

PR #2222, opened 2020-07-03
Seedlet: try removing .entry-content specificity for alignment rules

jffng:
This PR tries out what's proposed in #2126. The alignment classes seem to work just fine without the parent selector specificity.

I couldn't get away with removing `.entry-content` on the default width implementation, because the selector would become too generic:

https://github.com/Automattic/themes/blob/2b61bca40864fb757e68e664c9fb725e948e60e6/seedlet/assets/sass/blocks/utilities/_style.scss#L51-L57

Two questions:
- What was the purpose of using the parent selector for alignments?    
- Would we want to remove the `.entry-content` + alignment specificity in block styles as well, for example in the [video block](https://github.com/Automattic/themes/blob/seedlet/try/remove-alignment-parent-class/seedlet/assets/sass/blocks/video/_style.scss#L11-L13)?


allancole:
> I couldn't get away with removing .entry-content on the default width implementation, because the selector would become too generic

You might be able to get away with something like `*[class^="wp-block]:not(…)`. Might also be too generic, though. 

> What was the purpose of using the parent selector for alignments?

It’s mostly just an artifact of how themes’ content areas have traditionally been styled but it _I think_ it also helps with backwards compatibility for non-Gutenberg-enabled sites. If we can figure out how to remove it, I’d be all for replacing this with something specific for Gutenberg and then move the less complex logic for `.entry-content > *:not(…)` to a separate place for backwards compatibility.

allancole:
> Would we want to remove the .entry-content + alignment specificity in block styles as well, for example in the video block?

I definitely think this is worth a try—especially if it means lightening the CSS load overall.

seedlet/assets/sass/blocks/utilities/_style.scss
```diff
@@ -41,16 +38,13 @@
...
-.entry-content > .alignright {
 	@extend %responsive-alignright-mobile;
 	@extend %responsive-alignright;
 }
 
 // Make sure siblings of floated elements are top-aligned when nested
-.entry-content * > .alignleft + *,
-.entry-content * > .alignright + * {
+* > .alignleft + *,
+* > .alignright + * {
```
allancole, line 47: I think we can replace that `*` selector with the nested block wrapper since this style is specific to nested blocks. Otherwise, it might effect top level blocks as well. 

```suggestion
[class*="inner-container"] > .alignleft + *,
[class*="inner-container"] > .alignright + * {
```

seedlet/assets/sass/blocks/utilities/_style.scss
```diff
@@ -38,13 +41,16 @@
 	margin-bottom: var(--global--spacing-vertical);
 	/*rtl:ignore*/
 	margin-left: var(--global--spacing-horizontal);
+}
+
+.entry-content > .alignright {
```
jffng: I spoke / committed too soon. I think using need to use `.entry-content > .alignleft/right`. Otherwise we'd have to use a lot of `:not` selectors to determine when to apply the `%responsive-alignleft/right-width`s.

jffng:
After some exploration, all this PR does is enable the use of `.alignwide` and `.alignfull` without a parent `.entry-content` class. 

`.alignright`, `.alignleft`, and `.aligncenter` could already be used outside of `.entry-content`. 

@kjellr if there are specific use cases in mind that this PR doesn't cover yet, I'm happy to take a look.

seedlet/assets/sass/blocks/utilities/_style.scss
jffng: Good call, done!

kjellr:
> @kjellr if there are specific use cases in mind that this PR doesn't cover yet, I'm happy to take a look.

That should cover it. To test this, try [removing the extra alignment rules in Seedlet Blocks](https://github.com/Automattic/themes/blob/b5225140b29e553316d300b77f351612672ac956/seedlet-blocks/style.css#L51-L69) and see if things work generally well with this PR active. ([These lines](https://github.com/Automattic/themes/blob/b5225140b29e553316d300b77f351612672ac956/seedlet-blocks/style.css#L71-L77) might also be able to be removed?)

seedlet/assets/sass/blocks/utilities/_style.scss
```diff
@@ -1,5 +1,13 @@
 /* Block Alignments */
 
+/**
+ * These selectors set the default max width for content appearing inside a post or page. 
+ */
+.entry-content > *:not(.alignwide):not(.alignfull):not(.wp-block-separator):not(.entry-attachment):not(.woocommerce),
```
jffng, line 6: I removed `:not(.alignleft):not(.alignright)` from these selectors. Otherwise, blocks aligned left or right will overflow from the main column if their content is wider than the default width.

jffng:
I made the following changes:
- Removed Seedlet Blocks alignment classes as suggested by @kjellr . This worked 👍 in my testing
- Removed some extra, highly-targeted alignment classes that don't seem to be doing anything in the pullquote, file, and video blocks.  
- Fixed a bug for applying the default width to `.alignright` and `.alignleft` blocks. 

To test this, it's worth trying out a few different block alignments (both nested and not) and ensure that they behave as expected / didn't break behavior prior to this PR.

jffng:
I rebased this against master and fixed a Seedlet Blocks issue with the help of @kjellr. 

While supporting alignments will change as FSE matures, I still think this PR is worth merging because it:

1. Allows the alignwide and alignfull classes to work without the need for an `.entry-content` parent class, without affecting existing alignment behavior.
2. Removes unneeded alignment-specific CSS. 

@kjellr do you want to give it another spin? Thanks.

PR #2231, opened 2020-07-06
Dalston: Enqueue Block Scripts only in non-AMP

### LINKED ISSUE #2230: Dalston: AMP validation errors due to Block Scripts loading in frontend
## Steps to replicate

Go to Block Editor and add a post with `Cover block`, `Column block` or `Media & Text block`. Validate the post with [AMP plugin](https://wordpress.org/plugins/amp/).

## Result
Some Arbitrary JavaScript validation errors will be generated as JS <script> tags cannot be used in AMP version unless they are for loading AMP components. In order for a page to be served as AMP, the custom JS codes must be removed from the page. In this case the `enqueue_block_assets` hook used to load the block scripts in both admin and frontend which is the cause of the validation errors

## Expected

In ideal cases, AMP version of a page should not produce any warnings.

itowhid06:

As JS <script> tags cannot be used in AMP version unless they are for loading AMP components. In order for a page to be served as AMP, the custom JS codes must be removed from the page. In this case the enqueue_block_assets hook is used to load the block scripts in both admin and frontend which is causing validation errors. This PR checks for whether is on AMP endpoint and if not then enqueues the Block scripts.

#### Related issue(s):
Fixes #2230

### DIFF
--- a/dalston/functions.php
+++ b/dalston/functions.php
@@ -196,37 +196,36 @@ add_action( 'enqueue_block_editor_assets', 'dalston_editor_styles' );
  */
 function dalston_block_extends() {
 
-	// Cover Block Tweaks
-	if ( ! dalston_is_amp() ) {
-		wp_enqueue_script( 'dalston-extend-cover-block',
-			get_stylesheet_directory_uri() . '/block-extends/extend-cover-block.js',
-			array( 'wp-blocks' )
-		);
+	// Bail out early while in AMP endpoint.
+	if ( dalston_is_amp() ) {
+		return;
 	}
+
+	// Cover Block Tweaks
+	wp_enqueue_script( 'dalston-extend-cover-block',
+		get_stylesheet_directory_uri() . '/block-extends/extend-cover-block.js',
+		array( 'wp-blocks' )
+	);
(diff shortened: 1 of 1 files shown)


itowhid06:
Hi @allancole, could you please review this PR and merge it if all looks good? This will fix AMP validation errors logged in the WordPress dashboard.

dalston/functions.php
```diff
@@ -197,33 +197,50 @@ function dalston_editor_styles() {
 function dalston_block_extends() {
```
allancole, line 198: @itowhid06, I think it might be better if we remove both the block style scripts _and_ styles from AMP. Without those JS files, the custom block classes (`.is-style-full-bleed` for example) will be missing so the accompanying CSS will be useless anyway. Let’s remove all the `if ( ! dalston_is_amp() ) {…}` conditionals and add something like this at the top of the function:

```
if ( dalston_is_amp() ) {
    return;
}
```

itowhid06: Hi @allancole, Thank you very much for your input. I have updated the PR as per your suggestion. Please check.

Issue #2169, opened 2020-06-23
Rivington: Custom colour schemes are not applied in the editor


iamtakashi:
I hope I'm not missing anything obvious. In my several test sites with Rivington, the custom colour schemes are applied correctly on the front of the site, but the editor remains the default.

As an example, this is what I see on the front-end:
<img width="1411" alt="Screenshot 2020-06-23 at 12 20 04" src="https://user-images.githubusercontent.com/908665/85398281-e73e8e80-b54c-11ea-9afd-2f4e8123e169.png">

But the page appears like this in the editor:
<img width="1410" alt="Screenshot 2020-06-23 at 12 28 05" src="https://user-images.githubusercontent.com/908665/85398398-135a0f80-b54d-11ea-8cbe-67a0a15ac204.png">

I've also seen some weirdness in Gutenberg UI. The custom colours might be bleeding to the UI.
<img width="282" alt="Screenshot 2020-06-23 at 12 29 13" src="https://user-images.githubusercontent.com/908665/85398705-8d8a9400-b54d-11ea-9e8c-8518b82f624a.png">

Tested with Mac Chrome, Simple site on the premium plan.

cc @kjellr @ianstewart @alaczek

kjellr:
That's super weird. I can reproduce, but only when using this color scheme: 

<img width="279" alt="Screen Shot 2020-06-23 at 7 49 55 AM" src="https://user-images.githubusercontent.com/1202812/85400287-2efdef80-b526-11ea-991e-da99a7fbf0b6.png">

@iamtakashi can you test the other two and see if they show up in the editor?

iamtakashi:
Thanks for looking into this so quickly, @kjellr!

The issue is happening to me with all the pre-defined palettes and a totally custom one.

![Screenshot 2020-06-23 at 13 02 02](https://user-images.githubusercontent.com/908665/85401641-418e1e00-b552-11ea-817b-ee36586089df.png)

kjellr:
Hmm. Ok, that's odd. Are Coutoire and Alves working fine for you?

In any case, @allancole can take a look as soon as he gets on. 👍

iamtakashi:
> Are Coutoire and Alves working fine for you?

Not for me. Both themes share the same issues including the bleeding to the UI. A colour scheme gets applied on the front of the site, but the editor.

kjellr:
Ok got it. That sounds like it may be an issue with editor styles in the Color Annotations plugin then. We'll dig into it. 👍

[...]

PR #2232, opened 2020-07-06
#2169 - fixes the issue with the cover block styles mismatch.

jeffikus:
- fix for the cover block color priority issue in #2169


allancole (approved):
Yup, this works well. 
LGTM!

Issue #2127, opened 2020-06-15
Seedlet: Figure out if we need to build in Color Annotations support for WP.com 


jffng:
Figure out if we need to build in Color Annotations support for .com, or if we can rely on the support already built into the theme.

The reasons we might need to add it are:

- The current approach uses CSS variables, which means your site's colors won't be different in IE11. I don't expect that will change with #15.
- The current approach allows everyone to change colors, not just those on WordPress.com premium plans.

jffng:
Closed by #2233.

PR #2233, opened 2020-07-06
Seedlet: Begin adding color annotations

kjellr:
Addresses #2127. 

This PR adds initial WP.com color annotations to Seedlet. 

Light|Medium|Dark
---|---|---
kjellreigstad wordpress com_wp-admin_customize php (1)|kjellreigstad wordpress com_wp-admin_customize php|kjellreigstad wordpress com_wp-admin_customize php (3)

Still todo: 

- [x] Add styles values for the border color (Should just mirror the tertiary color, for simplicity). 
- [x] Implement editor stylesheet.
- [x] Update the editor color presets to reflect the colors chosen here. 
- [x] Unload `classes/class-seedlet-custom-colors.php` when this is active. 
- [x] [Refresh `seedlet_custom_colors_extra_css()`](https://github.com/Automattic/themes/pull/2233#issuecomment-656114965) to reflect updates in the Customizer.


kjellr:
Works pretty well so far. One frustrating bug is that none of our custom gradients (or the site title shadows) will update automatically in the Customizer, because it won't refresh `seedlet_custom_colors_extra_css()`. I wonder if there's a way we can force that? Here's how that looks currently: 

gradients

Those errors go away when the page reloads and `seedlet_custom_colors_extra_css()` runs again.

kjellr:
There are a few things I could use some extra dev help with here: 

**Re-run `seedlet_custom_colors_extra_css()` to reflect user's updates in the Customizer.** 

Not sure how possible this is, but [as described above](https://github.com/Automattic/themes/pull/2233#issuecomment-654416367), we need this function to regenerate and inject CSS whenever the user makes a color edit in the customizer. Otherwise it'll be out of sync with the user's selections. 

**Unload classes/class-seedlet-custom-colors.php when this is active.**

On WP.com, we need to unload everything in `inc/class-custom-colors.php`, since we don't want those controls showing up in addition to the controls here. I'm not sure how to do that myself, but my understanding is that we'd want to make those color controls [filterable](https://developer.wordpress.org/reference/functions/add_filter/), and then likely [filter them out](https://developer.wordpress.org/reference/functions/remove_filter/) in `inc/wpcom.php`. 

**Update the editor color presets to reflect the colors chosen**

Currently we [modify the custom Gutenberg colors](https://github.com/Automattic/themes/blob/b5225140b29e553316d300b77f351612672ac956/seedlet/functions.php#L152-L168) if someone's chosen colors in the customizer panel. On WP.com, we'll want to instead use the values chosen in the WP.com customizer panel ([like we do in Varia](https://github.com/Automattic/themes/blob/b5225140b29e553316d300b77f351612672ac956/varia/functions.php#L143-L152)). So the logic here is likely going to have to get a bit more complicated.

cc @jffng, @allancole, @danieldudzic, or @jeffikus if any of you have any time to look into these.

jffng:
This makes sense, as a heads up I've started to look into these two things first:

> **Unload classes/class-seedlet-custom-colors.php when this is active.**

> **Update the editor color presets to reflect the colors chosen**

allancole:
> Re-run seedlet_custom_colors_extra_css() to reflect user's updates in the Customizer.
 
For this part, [Selective Refresh](https://make.wordpress.org/core/2016/02/16/selective-refresh-in-the-customizer/) might be the way to get the colors reprocessed when color options change :-) 
 
This may be helpful: https://wordpress.stackexchange.com/questions/247234/how-do-i-implement-selective-refresh-with-a-customizer-setting.

jffng:
I pushed an update that should address editor palette + unloading Seedlet's built-in custom color implementation when wpcom color annotations exist. It doesn't feel very elegant.

That being said, I didn't use a filter because it seemed more complicated than necessary.

seedlet/functions.php
```diff
@@ -150,22 +150,23 @@ function seedlet_setup() {
...
+			$primary          = ( ! empty( $colors_theme_mod ) && 'default' === $colors_theme_mod || empty( get_theme_mod( 'seedlet_--global--color-primary' ) ) ) ? '#000000' : get_theme_mod( 'seedlet_--global--color-primary' );
+			$secondary        = ( ! empty( $colors_theme_mod ) && 'default' === $colors_theme_mod || empty( get_theme_mod( 'seedlet_--global--color-secondary' ) ) ) ? '#3C8067' : get_theme_mod( 'seedlet_--global--color-secondary' );
+			$foreground       = ( ! empty( $colors_theme_mod ) && 'default' === $colors_theme_mod || empty( get_theme_mod( 'seedlet_--global--color-foreground' ) ) ) ? '#333333' : get_theme_mod( 'seedlet_--global--color-foreground' );
+			$tertiary         = ( ! empty( $colors_theme_mod ) && 'default' === $colors_theme_mod || empty( get_theme_mod( 'seedlet_--global--color-tertiary' ) ) ) ? '#FAFBF6' : get_theme_mod( 'seedlet_--global--color-tertiary' );
+			$background       = ( ! empty( $colors_theme_mod ) && 'default' === $colors_theme_mod || empty( get_theme_mod( 'seedlet_--global--color-background' ) ) ) ? '#FFFFFF' : get_theme_mod( 'seedlet_--global--color-background' );
+		} else {
+			// Use wpcom color annotations
```
jffng, line 163: If we go with this approach, these lines may need to be stripped out when we submit updates to the dotorg repo.

seedlet/functions.php
```diff
@@ -396,7 +397,9 @@ function seedlet_skip_link_focus_fix() {
 /**
  * Custom colors class.
  */
-require get_template_directory() . '/classes/class-seedlet-custom-colors.php';
+if ( empty( get_theme_mod( 'colors_manager' ) ) ) {
```
kjellr, line 400: Can you add a code comment here? I can imagine someone not familiar with WordPress.com color annotations wondering what this is for.

seedlet/functions.php
kjellr: Hmm yes, you're probably right. 🤔 That may get a little confusing. @allancole do you have any thoughts?

allancole: Hmmm, I think we can roll all of this back to the wporg-specific method. Then, in `inc/wpcom.php` we can use `remove_theme_support()`([link](https://developer.wordpress.org/reference/functions/remove_theme_support/)) to remove _this_ method. Then later in that same file we can use `add_theme_support()` again to replace the wporg method with this color-annotation based, wpcom-specfic method.

allancole: Not 100% sure that will work but, 🤞🏽

jffng: I'm giving this a try, thanks for the suggestion @allancole!

seedlet/inc/wpcom-colors.php
```diff
@@ -539,4 +539,107 @@ function seedlet_custom_colors_extra_css() {
...
+				array(
+					'name'     => __( 'Stripe', 'seedlet' ),
+					'gradient' => 'linear-gradient(to bottom, transparent 20%, ' . $gradient_color_a . ' 20%, ' . $gradient_color_a . ' 80%, transparent 80%)',
+					'slug'     => 'stripe',
+				),
+			)
+		);
+	}
+}
+add_action( 'after_setup_theme', 'seedlet_replace_editor_palette' );
```
jffng, line 645: This approach isn't working in my testing, but I wanted to push this to see if anyone had any ideas.

kjellr: I think this needs to go in `wpcom.php`, not `wpcom-colors.php`. The latter file is normally used just for the color annotations themselves, so I think it may be pulled in too late to make this work.

jffng: Ah good call! Moving this code to `wpcom.php` loads the editor palette correctly.

jffng:
>  Implement editor stylesheet.

How should this be done? Does `add_color_rule` work for the editor too?

seedlet/functions.php
```diff
@@ -396,7 +386,9 @@ function seedlet_skip_link_focus_fix() {
 /**
  * Custom colors class.
  */
-require get_template_directory() . '/classes/class-seedlet-custom-colors.php';
+if ( empty( get_theme_mod( 'colors_manager' ) ) ) { // If the theme is on wpcom, we bypass the theme's built in custom colors, because wpcom uses a different custom color implementation.
```
jffng, line 389: Is this an acceptable way to "unload" the theme's default custom color implementation?

kjellr:
> How should this be done? Does add_color_rule work for the editor too?

Yep, we need to add a `wpcom-editor-colors.php` file for the editor annotations. 

Oddly enough, this becomes _way_ easier to do once we have an [IE version of `style-editor.css`](https://github.com/Automattic/themes/issues/2171#issuecomment-654891435)(there's a script we can use to auto-pull and format the selectors for `add_color_rule` from that file). Do you think you can make that happen first? Once we have an `ie-editor.css` file, I can probably make this happen in ~20 minutes. 😄

seedlet/functions.php
kjellr: Yeah, I think it's okay.

jffng:
@kjellr [0361b28](https://github.com/Automattic/themes/pull/2233/commits/0361b283b5d9e90692ca99a3fd82e5bdf00fccbe) adds support for `ie-editor.css`. There are a handful of undefined CSS variables that the postcss plugin generates warnings for, which I will take a look at fixing.

kjellr:
I added a first pass at editor annotations in 3f0dffc — this could definitely use some testing though. 👍

kjellr:
I added a couple more commits: 

- Tidying up the color annotations for cover blocks and custom background colors. 
- Added the Varia Customizer warning about color contrast.

kjellr:
> One thing that's definitely missing is the code that powers the gradients. In the front-end annotations, I'm adding those as extra CSS, but I don't think we have an equivalent function available for the editor. Is that right, @allancole?

Turns out this was super easy — it works the same way it works on the front end. 🙌 Fixed in b4a4e07. @jffng would you mind testing the custom colors a bit to see if anything feels off to you?

sirbrillig:
It's been a long time since I worked on anything having to do with colors, but I found some historical reference that sounds related and perhaps may help y'all.

[This issue](https://github.com/Automattic/customization/issues/157) involved the custom colors plugin not applying changed styles to the customizer preview during full refresh.

> Ok, what seems to be happening is specifically that custom colors is rendering its changes via postMessage directly to the site preview, but then when any other setting is changed that uses refresh to reload the whole preview, the colors changes are not being re-applied to the new preview markup. The changed colors settings are still changed, however, so when the Customizer is saved, they appear on the front-end. They are simply not being applied in the preview correctly after a refresh. There is a built-in cache of the generated colors CSS which is disabled on sandboxes. We should be able to disable the cache for previews, and thus prevent the bug in the Customizer entirely.

If I'm understanding your issue correctly, and I probably am not, you're creating a new system in this theme that does color-related things and you want those things to be reflected in the preview. There are three ways I know of to do that:

1. When a relevant setting changes, cause the full preview iframe to reload. This is how settings in the customizer worked prior to the next two options (and I'm sure some still do). I don't recall the commands needed to do it, but I think there's some special `postMessage` command you can send to the iframe that causes it to request the preview again from the PHP on the server with the new settings applied.
2. When a relevant setting changes, use `postMessage` to send data from the controls javascript to the preview iframe and then have a second script listening inside the preview where it will apply the updated CSS via javascript directly. This was how settings in the customizer worked before selective refresh was available. 
3. When a relevant setting changes, tell the customizer to make a request to the PHP on the server to regenerate _just_ certain changed sections of the page using the new settings and then dynamically replace those sections inside the preview pane. This is "selective refresh", but I'm not sure how it actually works. It also might not be what you want here since I imagine your color changes will affect more than a few specific page sections.

jffng:
Thank you @sirbrillig , very helpful background and recommendations!

I went with option 2 and it appears to be working:

Kapture 2020-07-27 at 13 09 01

seedlet/inc/wpcom-customize-preview.js
```diff
@@ -18,4 +17,74 @@
 			}
 		} );
 	} );
+
+	// The wpcom colors plugin does not handle live updates of extra CSS.
+	// This function provides the CSS updates via JavaScript.
+	wp.customize( 'colors_manager[colors]', function( value ) {
+		value.bind( function( to ) {
+			const { bg, fg1, fg2 } = to;
+			const extraCSS = `/*
```
jffng, line 26: It's unfortunate that some version of this CSS has to live in three places (here, `inc/wpcom-colors.php`, and `inc/wpcom-editor-colors.php`), but I'm not sure how to get around this if we want to support custom colors as the plugin is written right now.

kjellr: Agreed — it's less than ideal. I think this might be the sort of thing we implement for now, and then re-evaluate later on. 

For `inc/wpcom-colors.php`, and `inc/wpcom-editor-colors.php`, could we write a single function for it in `wpcom.php`, and then call it in each of those files? We could use some sort of context parameter to fill in the appropriate classnames.

jffng: The extra css is added via `add_theme_support` here: https://github.com/Automattic/themes/blob/try/seedlet/color-annotations/seedlet/inc/wpcom-colors.php#L517

Can this function be called passing an argument for the context?

kjellr: I'm not really sure, to be honest. 😄 `add_theme_support` [can generally take extra arguments](https://developer.wordpress.org/reference/functions/add_theme_support/), but it looks like the function name here _is_ the extra argument.

jffng: Yeah not totally sure what to do here as the argument in this case refers to callback's handle. I don't think this optimization should be a blocker though.

I'll add some comments to clarify what's going on, because, as we spoke about, these functions are rather opaque.

kjellr:
This is great! I ran it through some tests and it appears to be working well. If someone can give this one last code review (particularly the JS, which I shouldn't review myself), I think it's good to go. @jeffikus or @allancole would you mind taking a look?

Also want to note that merging this will close #2171.

seedlet/inc/wpcom-customize-preview.js
allancole: Yeah, I wish we didn’t have to put this extra CSS in here, but it does work well and child themes won’t have to touch it either which is a plus. 

I agree we should keep it and iterate on it later :-)

seedlet/inc/wpcom.php
```diff
@@ -112,3 +252,12 @@ function seedlet_wpcom_editor_scripts() {
 	wp_enqueue_style( 'seedlet-wpcom-editor-style', get_template_directory_uri() . '/inc/wpcom-style-editor.css', array(), '20200629' );
 }
 add_action( 'enqueue_block_editor_assets', 'seedlet_wpcom_editor_scripts' );
+
+/**
+ * Enqueue CSS and JS for customizer pane.
+ */
+function seedlet_enqueue_message_scripts() {
+	wp_enqueue_script( 'seedlet-customize-message-wpcom-script', get_template_directory_uri() . '/inc/wpcom-customize-message.js', array( 'customize-controls' ), wp_get_theme()->get( 'Version' ) );
```
allancole, line 260: We can remove `'seedlet-customize-message-wpcom-script'` and its related JS file `/inc/wpcom-customize-message.js` entirely. 

The color a11y warning message is added with PHP (see below) and so we only need to keep the `'seedlet-customize-message-wpcom-style'` CSS to style the message. 

https://github.com/Automattic/themes/blob/b2b6229bf897596c11b35900c02bcb0ab90937dc/seedlet/inc/wpcom.php#L157-L169

kjellr: Thank you! Should be all set as of 7cec571.

kjellr:
> I noticed on dotorg sites, the custom colors are no longer showing up on the frontend or the customizer. They do show up in the editor, though:
> 
> I think we just need to add `:root:root` to the output of `seedlet-style-inline-css` so it overrides `:root:root` in `styles.css` . 

@allancole: Ah, interesting. I _think_ this only happens when you have Seedlet Blocks active, right? It's [these styles](https://github.com/Automattic/themes/blob/42bbe12cb50c333f46c85f9fd017be5195ba8a10/seedlet-blocks/style.css#L31-L49) that need to be overridden. Part of me wonders if we should write something special for Seedlet Blocks, rather than unnecessarily double-up on selectors in the main theme.

allancole:
🤦‍♂️ … yup, that issue is only happening for Seedlet Blocks! Somehow I blocked out my own site-title telling me what theme was active, lol. 

> Part of me wonders if we should write something special for Seedlet Blocks, rather than unnecessarily double-up on selectors in the main theme.

Yeah, there’s probably something we can do to change _where_ the stylesheet is loaded in Seedlet blocks so we can rely on cascading over stricter selectors here.

kjellr:
This issue was present before this PR then, so I think it's fine to merge this and then follow up with that later — this is already getting to be a pretty big PR. 😄 Would you mind giving this one more review?

jffng:
Sweet. @kjellr should we hold off merging this until we ship #2305 and resubmit to the themes directory, since this PR has a pretty wide footprint?

kjellr:
> should we hold off merging this until we ship #2305 and resubmit to the themes directory, since this PR has a pretty wide footprint?

@jffng sure, that sounds fine to me. Is that the last PR that needs to land before we resubmit?

jffng:
Yes.

kjellr:
Now that #2305 has been merged, I'm going to go ahead and get this one merged and deployed. Thanks, all!

PR #2236, opened 2020-07-07
Seedlet: Remove top margin of all layout grid columns in the editor

Copons:

- Remove the top margin of all Layout Grid's Column blocks in the editor in the Seedlet theme.

Seedlet applies custom vertical margins to all blocks in the editor, then it removes it for first-childs.
This is generally ok, as blocks are displayed vertically, but misaligns columns, displayed horizontally at the same level in the DOM.

Both Core Columns and Layout Grid Columns take care of removing the top margin, but the Seedlet CSS is specific enough to override the latter style.

| Before | After |
| - | - |
| Screenshot 2020-07-03 at 01 12 23 | Screenshot 2020-07-07 at 15 41 14 |

#### Related issue(s):

Fixes https://github.com/Automattic/wp-calypso/issues/43882


seedlet/assets/css/style-editor.css
```diff
@@ -1519,3 +1519,8 @@ pre.wp-block-verse {
 	padding-left: 0;
 	padding-right: 0;
 }
+
+/* Remove columns top margin */
+.wp-block-jetpack-layout-grid .wp-block[data-type="jetpack/layout-grid-column"] {
```
noahtallen, line 1524: Is there any way to do it so that we don't have to add overrides for the specific block 😬

noahtallen: (i imagine there might be other blocks which might end up having the same bug?)

Copons: @noahtallen I've tested with Core Columns and Media & Text, and they work fine, as the Core CSS is more specific.
We _could_ fix this by increasing the specificity of the [block's style](https://github.com/Automattic/block-experiments/blob/d60797dcd0731add1641ec8f63f62fcc82dd5e3b/blocks/layout-grid/editor.scss#L25-L30), but in this case the misalignment is caused by the theme.

To be clear, both approaches sound good to me.
Maybe updating the block would make it more bullet-proof, though. 🤔

noahtallen: Yeah, I guess I wonder if there is a way to fix it in the theme without having to be specific about the block..

> Seedlet applies custom vertical margins to all blocks in the editor, then it removes it for first-childs.

it sounds like this is _not_ expected behavior in any case where you have grid/columns lining up like this, so maybe that behavior should change more broadly?

Copons: That's a good point. 🤔 

AFAICS the [theme](https://github.com/Automattic/themes/blob/55051ac7b0904213034713c8d8c56ba20a5766ec/seedlet/assets/sass/blocks/utilities/_editor.scss#L158-L174) is applying a custom margin to all blocks, and then removing it to first childs, to avoid having double margins on nested blocks.

It's also done in a sensible way, with a very low specificity. The issue here is only caused the wrapper class automagically added to the editor CSS, that increases it just enough to override the block style.

The problem here is that I don't think we can specifically target horizontally-aligned child blocks, so we would risk removing the margin to "vertical"  blocks as well (which would affect the inserter positioning I think).

Solving this in the block is probably better, though, and it would be consistent enough with how Core itself handles the Core Column top margin.

Copons:
Block-first approach that would replace this PR: https://github.com/Automattic/block-experiments/pull/112

Copons:
Fixed in the plugin: https://github.com/Automattic/block-experiments/pull/112

PR #1962, opened 2020-04-30
Alves: Add Color Annotataions.

allancole:
Follows the mold of #1944 to add basic color annotations for Alves.

image

To Do:
- [x] Refinement around contrast.
- [x] Double checking responsive styles.
- [x] Add featured color palettes.

PR #2243, opened 2020-07-08
All Varia-themes: Fix form text color issue in dark color palette.

allancole:

When using the dark color palette, the form text-color can sometimes disappear because it’s associated with a `$foreground` variable which which is too light. Since there isn’t a background-color variable for form elements in Varia, it defaults to a browser-set, `white` background-color. This means the form text needs to always be dark to remain legible. To solve this simply, I replaced the `$foreground` text-color for forms with a standard `black` color that persists regardless of color palette. 

To test, choose any Varia-theme and set it to use the Dark color palette. 
Then visit any form (search, contact, etc.), enter some text and make sure its legible. 

#### Related issue(s):

Discovered and reported here: https://github.com/Automattic/themes/pull/1962#issuecomment-655086545


kjellr:
Is it best practice to set a form field background color, if we're setting the text color here too?

allancole:
> Is it best practice to set a form field background color, if we're setting the text color here too? 
 
@kjellr Hmmm… Good question. That makes me wonder if it might be better to remove the `$form-text-color` variable altogether in this PR. That way input forms take on the browser’s defaults. 
 
My thinking is that we use this PR as a quick-fix solution out that corrects this for the dark palette. Then we can revisit a more robust solution at a later time.

allancole:
> the form field color annotations still overrode this new black value:

Ahhh, right!. I’ve gotta remove the form color rules from all the annotations on this PR as well.

allancole:
@kjell, I’ve added these fixes to Alves, Coutoire and Rivington. However, since the other color-annotations are not merged to `master` yet, I’ll have to add [this detail](https://github.com/Automattic/themes/pull/2243#pullrequestreview-445825588) to each individual PR. That way we can get this merged and deployed now for those 3 themes and then deploy the others normally as they get merged to master.

kjellr:
Cool — I'll test those three themes out, and if they're working right I'll add a ✅ here. 👍

kjellr:
Ok, this generally works, but I noticed a couple things: 

1. In all these themes, the jetpack address block fields are unreadable in the "Dark" palette. This doesn't seem to be a direct result of this PR, but should we fix it here? 

Screen Shot 2020-07-09 at 6 22 30 PM

2. Similarly, in Rivington when you choose a palette with a white background, the form fields appear invisible. This existed before this PR too, so I wonder if it should be tackled separately too?

Screen Shot 2020-07-09 at 6 27 49 PM

allancole:
> In all these themes, the jetpack address block fields are unreadable in the "Dark" palette. This doesn't seem to be a direct result of this PR, but should we fix it here?

@kjellr Pretty sure this should be solid now :-)

Needed to add some vendor supports to allowing Jetpack styles to appear in the default editor stylesheet to get this to work properly. 

> Similarly, in Rivington when you choose a palette with a white background, the form fields appear invisible. This existed before this PR too, so I wonder if it should be tackled separately too?

Rivington set the border to 0px, so I reversed that and connected it to the border-color variable so its always visible. May need to check other themes to see if they did a similar thing but maybe we can do that separately.

rivington/sass/_config-child-theme-deep.scss
```diff
@@ -166,7 +166,7 @@ $config-elements: (
 
 		// Borders
 		"border": (
-			"width": 0,
+			"width": 1px,
```
kjellr, line 169: Question: Should we add this to the custom colors extra css? 

It's minor enough that folks probably won't notice, but on the other hand, we don't _need_ it unless custom colors are active.

allancole: Ahh, yeah I agree. 
I’ll do this with the extra CSS.

PR #2244, opened 2020-07-08
Use display=swap for Google Fonts

josephscott:
Using `display=swap` allows text to show up right away while waiting for the font to load, then swaps once it is finished.

You can see the type of difference this can make: https://www.webpagetest.org/video/compare.php?tests=200708_64_a5c3e57fdad44be3f5e98a618c78aca6%2C200708_7Y_27106d14550289c4ccc296ee0d819bbd&thumbSize=200&ival=100&end=visual

### DIFF
--- a/hever/functions.php
+++ b/hever/functions.php
@@ -141,6 +141,7 @@ function hever_fonts_url() {
 		$query_args = array(
 			'family' => urlencode( implode( '|', $font_families ) ),
 			'subset' => urlencode( 'latin,latin-ext' ),
+			'display' => 'swap'
 		);
 
 		$fonts_url = add_query_arg( $query_args, 'https://fonts.googleapis.com/css' );


allancole:
Interesting. Not sure what our stance is on “FOUT” these days, but if this is better for performance, we might need to do this in all Varia-themes and probably Seedlet too, @kjellr.

kjellr:
Yeah, it makes sense to have things load this way. We should also consider bundling the fonts rather than loading from Google if at all possible, to avoid the external request.

josephscott:
> Yeah, it makes sense to have things load this way. We should also consider bundling the fonts rather than loading from Google if at all possible, to avoid the external request.

Big +1 - I would love to see themes that use web fonts ship with them.

josephscott:
> Interesting. Not sure what our stance is on “FOUT” these days, but if this is better for performance, we might need to do this in all Varia-themes and probably Seedlet too, @kjellr.

FOUT is definitely not great, but I'd argue that it is still better than showing a blank screen.

josephscott:
@allancole / @kjellr - what is the path forward for this?

josephscott:
@kjellr thank you.  Is there a plan for bumping Hever to 1.4.2 and releasing that version?

kjellr:
Yep! Hever v1.4.2 has been released and includes this change. Thanks again!

PR #2245, opened 2020-07-09
Remove full width block margins in the editor

jeyip:
This PR removes the margins for full-width blocks in Seedlet.

The Seedlet frontend has [styling](https://github.com/Automattic/themes/blob/master/seedlet/style.css#L613-L621) that removes top and bottom margins from full-width blocks within entry content. The `.entry-content` class, however, does not exist in the block editor. In order to implement the same behavior in the editor, we target full-width blocks by leveraging the the `[data-align="full"]` selector.

One thing to note is that the selectors to remove margins are different between `seedlet` and `seedlet blocks`. This is meant to accommodate full site editing. We don't want to remove the margins for _all_ full width blocks in the site editor (otherwise the header, content, and footer would have no spacing between them)

### Block Editor Before:
wordpress com_block-editor_page_bigmacsiteeditor5 wpcomstaging com_1541

### Block Editor After:
localhost_8888_wp-admin_post php_post=11 action=edit

## Related issue(s):
Fixes https://github.com/Automattic/wp-calypso/issues/43883

### DIFF
--- a/seedlet-blocks/style-editor.css
+++ b/seedlet-blocks/style-editor.css
@@ -16,3 +16,8 @@ body {
 	--global--color-secondary: var(--wp--preset--color--secondary);
 	--global--color-tertiary: var(--wp--preset--color--tertiary);
 }
+
+.site-content .wp-block[data-align="full"] > [data-block], .site-content .wp-block.alignfull > [data-block] {
+    margin-top: 0;
+    margin-bottom: 0;
+}
--- a/seedlet-blocks/style-editor.css
+++ b/seedlet-blocks/style-editor.css
@@ -18,6 +18,6 @@ body {
 }
 
 .site-content .wp-block[data-align="full"] > [data-block], .site-content .wp-block.alignfull > [data-block] {
-    margin-top: 0;
-    margin-bottom: 0;
+	margin-top: 0;
+	margin-bottom: 0;
 }
--- a/seedlet/assets/sass/blocks/utilities/_editor.scss
+++ b/seedlet/assets/sass/blocks/utilities/_editor.scss
@@ -233,6 +233,19 @@
 	}
 }
 
+// Selects the post editor and not the site editor
+.block-editor-block-list__layout:not(.edit-site-block-editor__block-list){
+	.wp-block {
+		&[data-align="full"],
+		&.alignfull {
+			&>[data-block] {
+				margin-top: 0;
+				margin-bottom: 0;
+			}
+		}
+	}
+}
+
 .alignleft {
 	margin: 0;
 	margin-right: var(--global--spacing-horizontal);
--- a/seedlet/assets/css/style-editor.css
+++ b/seedlet/assets/css/style-editor.css
@@ -1163,6 +1163,11 @@ pre.wp-block-verse {
 	max-width: none;
 }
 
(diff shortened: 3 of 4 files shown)


seedlet-blocks/style-editor.css
```diff
@@ -15,4 +15,9 @@ body {
 	--global--color-primary: var(--wp--preset--color--primary);
 	--global--color-secondary: var(--wp--preset--color--secondary);
 	--global--color-tertiary: var(--wp--preset--color--tertiary);
-}
\ No newline at end of file
+}
+
+.site-content .wp-block[data-align="full"] > [data-block], .site-content .wp-block.alignfull > [data-block] {
+    margin-top: 0;
+    margin-bottom: 0;
```
Copons, line 22: These should be [indented with tabs](https://www.youtube.com/watch?v=SsoOG6ZeyUI)! 🙂

Copons:
> We don't want to remove the margins for all full width blocks in the site editor (otherwise the header, content, and footer would have no spacing between them)

@jeyip Is there a particular reason for wanting to keep the margins between site sections?

What if the user has a full-width image in the header, and another one in the content.
Should there be a space between them?

As far as I can see, an argument in favour of keeping the margins is that the block appender kinda relies on them.
On the other hand, the appender in the site editor works... a bit funny? It shows up in some cases, remains hidden in others.
(Note that this is unrelated to this PR though, as I didn't see any big differences between master and this branch.)

I'm still not familiar enough with the site editor, so I'd probably move this discussion somewhere else and involve other people. 🙂

seedlet/assets/sass/blocks/utilities/_editor.scss
```diff
@@ -233,6 +233,19 @@
 	}
 }
 
+// Selects the post editor and not the site editor
+.block-editor-block-list__layout:not(.edit-site-block-editor__block-list){
```
jffng, line 237: Is there a good reason to also not remove these margins in the site editor?

seedlet-blocks/style-editor.css
jeyip: Good catch! Thank you.

jeyip: LOL 😂

jeyip:
> What if the user has a full-width image in the header, and another one in the content.
Should there be a space between them?

Good question. My intention was to narrow the scope of the changes. I aimed to match existing visual styles as much as possible while still fixing unnecessary margins between full-width blocks in site content. 

I'd 100% defer to Automattic designers and the team though. If we feel like we want to remove the margins for _all_ full-width blocks, I'm on board (it makes the css selectors cleaner as well!)

kjellr:
> I'd 100% defer to the team and Automattic designers though. If we feel like we want to remove the margins for all full-width blocks in the site-editor, I'm on board (it makes the css selectors cleaner as well!)

Since this PR gets things in sync between the front and back end, let's move forward with this in the near term, and then reassess this question later on. 👍

seedlet/assets/sass/blocks/utilities/_editor.scss
jeyip: See https://github.com/Automattic/themes/pull/2245#issuecomment-656380777

I was trying to keep the visual styling close to what already exists today. Instead of removing _all_ margins for all full-width blocks in the side editor, I used [more specific selectors](https://github.com/Automattic/themes/pull/2245/files#diff-4d8badb34f37a5e11b5e97c01b7db662R20-R22) in `seedlet-blocks` to remove margins from site content (and not site header / site footer)

PR #2246, opened 2020-07-09
Seedlet Blocks: Use spacer blocks instead of css margins for the site header and footer

### LINKED ISSUE #2240: Seedlet Blocks: Make the header more compact
Seeldet Blocks is the future base theme of our FSE compatible designs. To give ourselves a running start we're already working on layouts using Seedlet Blocks, but we've run into some issues with header design.

If you look at our [current starter designs](https://www.figma.com/file/1Esa2uHH5jeIgI8JinEv3a/Default-Homepage-Layouts-for-Gutenboarding-March-2020?node-id=0%3A1), 7 out of 10 are using a theme with a compact, two-column header as a base. This is because we're aiming to show as much of the design above the fold as possible to create an appealing "wine label" (at 1280 x 960px resolution). 

image

For comparison, here's a new design I'm working on, with default Seedlet header (sans logo):

image

...and this is the kind of header I had in mind for this particular design:

image

The secondary goal is to include as much content as possible in the preview in [Gutenboarding](https://wordpress.com/new). Tight spacing in the header goes a long way. While FSE will eventually improve enough to allow us to create all kinds of headers, having a header that fits our needs out of the box would bridge the current gap.

image

I have two ideas on how we could address this:

1. Reduce the top and bottom padding in the header. Current 90 px padding is quite generous, maybe we could reduce it to say 30px? Could we examine the spacings between the header element too and reduce as much as possible?
2. Change the header design to a single-line, two-column similar to Mayland. This could be done in Seedlet Blocks, or in another block-based child theme if those are on the map (à la Varia child-themes). 

cc @ianstewart

kjellr:
Fixes #2240. 

Currently, the header in Seedlet Blocks uses a `site-header` class, to bring in some top and bottom margins from the original Seedlet theme. This isn't fully editable (or discoverable) in the site editor, so this PR removes that class and uses spacer blocks in the `header.html` template instead. 

This should have no effect on the header layout whatsoever, as the spacer blocks and native margins mimic what the `site-header` class did before. 

This PR does the same thing for the `site-footer` class too. In that case, it'll get just a little bit more margin than it did before this PR, because of the way Seedlet applies auto-margins. But that doesn't seem like a dealbreaker. 👍 

Theoretically, we shouldn't need those classnames for any semantic reasons, since these group blocks should be getting the `header` and `footer` tags anyway. (They don't appear to be doing that, but I think that's a standing issue)

---

Header Before| Header After
---|---
Screen Shot 2020-07-09 at 8 33 02 AM|Screen Shot 2020-07-09 at 8 33 02 AM

Footer Before| Footer After
---|---
Screen Shot 2020-07-09 at 8 36 49 AM|Screen Shot 2020-07-09 at 8 36 58 AM

### DIFF
--- a/seedlet-blocks/block-template-parts/footer.html
+++ b/seedlet-blocks/block-template-parts/footer.html
@@ -1,3 +1,11 @@
+<!-- wp:spacer -->
+<div style="height:30px" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- /wp:spacer -->
+
 <!-- wp:paragraph {"align":"center","fontSize":"small"} -->
 <p class="has-text-align-center has-small-font-size">Proudly powered by <a href="https://wordpress.org">WordPress</a></p>
 <!-- /wp:paragraph -->
+
+<!-- wp:spacer -->
+<div style="height:30px" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- /wp:spacer -->
--- a/seedlet-blocks/block-template-parts/footer.html
+++ b/seedlet-blocks/block-template-parts/footer.html
@@ -1,4 +1,4 @@
-<!-- wp:spacer -->
+<!-- wp:spacer {"height":30} -->
 <div style="height:30px" aria-hidden="true" class="wp-block-spacer"></div>
 <!-- /wp:spacer -->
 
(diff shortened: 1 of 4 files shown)


kjellr:
Whoops. Turned out the markup was mysteriously missing an attribute. Should be all set after 52cf8b7.

PR #2249, opened 2020-07-10
Seedlet Blocks: Replace header image with site logo block

kjellr:
Now that there's a site logo block in Gutenberg 8.5, this PR swaps out the image at the top of our header template part with the actual site logo block. 

Screen Shot 2020-07-10 at 9 26 49 AM


kjellr:
Whoops — I guess it's on track for 8.6, so we should probably let this PR sit until then.

PR #2251, opened 2020-07-13
Seedlet: bundle fonts with the theme

jffng:
This PR attempts to bundle the fonts instead of loading them externally from Google fonts, as mentioned in #2244 and elsewhere. 

Three `@font-face` declarations — regular, italic, and bold — are used per font: Playfair and Fira Sans. This should be equivalent of the font-faces we were fetching via Google Fonts. But if there are other ones we want to add, now might be a good time to consider doing so.

**To do**
- [x] Fix the unused `--wp--style--color--link` unused var
- [x] Remove unused fonts


seedlet/functions.php
```diff
@@ -345,9 +298,6 @@ function seedlet_content_width() {
  * Enqueue scripts and styles.
  */
 function seedlet_scripts() {
-	// Enqueue Google fonts
-	wp_enqueue_style( 'seedlet-fonts', seedlet_fonts_url(), array(), null );
```
kjellr, line 349: I think we may actually want to separately enqueue these, so that child themes will be able to de-queue them when they add their own fonts. If we use a standard `@import`, child themes will always have these loaded, even if they don't use them.

jffng: That's a good point!

[b48e544](https://github.com/Automattic/themes/pull/2251/commits/b48e544d3f0ea495b3ffb371e64f7c6e1656c40f) enqueues the fonts separately for the front end, and enables them to be filterable for the editor.

What do you and @allancole think of this approach?

seedlet/assets/css/style-editor.css
```diff
@@ -322,7 +322,7 @@ p.has-background.has-link-color:not(.has-background-background-color) a {
 }
 
 .has-link-color a {
-	border-bottom: 1px solid var(--wp--style--color--link);
+	border-bottom: 1px solid var(--wp--style--color--link, var(--global--color-primary));
```
jffng, line 325: This ensures that IE stylesheet can compile without an undefined variable.

seedlet/functions.php
kjellr: I think it works — my only question is whether or not we need two separate sass files for `fonts` and `fonts-editor`. I can't think of any standard situations where these would be different. Can we just use a single `fonts.scss` file somewhere and import it into both contexts?

jffng: I'm not sure how to get around having two sass files because the paths for importing the font are different for editor vs front-end, so two different stylesheets are required. I tried to minimize the code duplication on the sass side by using a [mixin](https://github.com/Automattic/themes/blob/b48e544d3f0ea495b3ffb371e64f7c6e1656c40f/seedlet/assets/sass/base/fonts/_mixin.scss). 

I'm open to suggestions, we could maybe do something clever where we pass a variable to `node-sass` at compilation.

seedlet/assets/sass/base/fonts/fonts.scss
```diff
@@ -0,0 +1,6 @@
+/**
+ * Font imports for the editor
```
kjellr, line 2:
```suggestion
 * Font imports for the front end
```

seedlet/functions.php
kjellr: Ah, yes ok I see what you mean. I think this seems like the best solution for now then. I'll give it a quick test before I do a final review. 👍

kjellr (commented):
I just tested, and I'm actually not seeing these fonts load on the front end. It's defaulting to other fonts for me: 

Screen Shot 2020-07-21 at 8 46 15 AM

I'm not seeing any console errors, and the fonts are loading correctly in the editor.

jffng:
Ah okay, I think I fixed the paths and consolidated to a single `font.css/scss` file! Also tested and verified on my wpcom sandbox.

Mind giving it one more spin @kjellr? Will wait to merge in case @allancole's has thoughts on the approach, specifically how the fonts are enqueued separately to make it easier for child themes to overwrite.

kjellr (approved):
This is working for me now. 👍

seedlet/assets/css/fonts.css
```diff
@@ -0,0 +1,40 @@
...
+  font-display: swap;
+  font-style: bold;
+  font-weight: 500 900;
+  src: url("../fonts/Fira_Sans/FiraSans-Bold.ttf") format("truetype"); }
+
+@font-face {
+  font-family: 'Fira Sans';
+  font-display: swap;
+  font-style: italic;
+  font-weight: 100 900;
```
allancole, line 39: Should this be 400?: 
```suggestion
  font-weight: 100 400;
```

seedlet/functions.php
```diff
@@ -353,8 +311,8 @@ function seedlet_content_width() {
...
+	// Fonts 
+	wp_enqueue_style( 'seedlet-fonts', get_template_directory_uri() . '/assets/css/fonts.css', array(), wp_get_theme()->get( 'Version' ) );
```
allancole, line 315: This is a little trickier than I thought. Didn’t realize the URI’s were different for the editor but I think its easier to override a singular font url function. Otherwise, I worry we’ll be forcing child themes to bundle fonts which isn’t really something we can do with the build tool, and can be cumbersome to do repetitively in child themes when we may want simply load external Google fonts. 

Can we do this like we do in Varia themes or is that bad for the TRT? 
https://github.com/Automattic/themes/blob/36c24e223f03a166e0548a2c3ed6bb0ee9464cce/alves/functions.php#L218-L233

That way we can override a singular function making it easier for us to override with a Google font _or_ a bundled font.

seedlet/assets/css/fonts.css
jffng: I was trying to account for the editor allowing **_bold and italic_**.

However it looks like this combination isn't working, and we may have to include an additional font-face for declaration for the bold+italic use case: 

Screen Shot 2020-08-04 at 5 51 09 PM

seedlet/functions.php
jffng: I don't see why we can't do it the way you're suggesting, I will try it out.

jffng: @allancole check out https://github.com/Automattic/themes/pull/2251/commits/fa4cb62801afb293de1e0048101c9e1cab2a47d9 — is this what you had in mind?

Issue #2218, opened 2020-07-03
Varia: Comment reply title skips heading hierarchy


AtrumGeost:
## Steps to replicate

Create a new blog post and check the comment reply title at the end of the blog post. Then use an accessibility tool to evaluate the post:

![Screen Shot on 2020-07-02 at 21:36:30](https://user-images.githubusercontent.com/3696121/86425972-25a21f00-bcac-11ea-8540-37e8c1be1869.png)

## Result

The comment reply title is using an H3 heading level.

## Expected

The comment reply title should use an H2 heading level.

* Dogfooding
* Slack convo: p1593446135037700-slack-C0FGK24F9

PR #2259, opened 2020-07-15
#2218: Varia: Comment reply title skips heading hierarchy

nielslange:

Change comment form title from h3 to h2 due to accessibility reason.

#### Related issue(s):

#2218 

#### Before:

#2218-before

#### After:

#2218-after

### DIFF
--- a/varia/comments.php
+++ b/varia/comments.php
@@ -69,7 +69,12 @@ if ( post_password_required() ) {
 
 	endif; // Check for have_comments().
 
-	comment_form();
+	comment_form(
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
I'm not sure about the reason behind this change because this is putting the comment form title at the same level as the headings inside the post itself. Also, this is something that is common to many other themes since it's the default for the WordPress function

scruffian:
I agree with @MaggieCabrera this seems like something to address in core.

PR #2260, opened 2020-07-15
Varia: Add left/right margins to iframes and other embeds to better s…

### LINKED ISSUE #1732: Varia: iframes with style attributes have a max-width of 32px when window size is at least 560px wide

allancole:

- Fixes the left/right margins on iframes and other embeds to better support responsiveness.
- Recompiles all Varia child themes with this change

#### Related issue(s):

Fixes #1732

### DIFF
--- a/alves/style-rtl.css
+++ b/alves/style-rtl.css
@@ -3177,15 +3177,13 @@ body:not(.fse-enabled) .footer-menu a {
 	text-decoration: none;
 }
 
-.entry-content > iframe[style] {
-	margin: 32px 0 !important;
-	max-width: 100% !important;
-}
-
-@media only screen and (min-width: 560px) {
-	.entry-content > iframe[style] {
-		max-width: 32px !important;
-	}
+.entry-content > iframe,
+.entry-content > iframe[style],
+.entry-content > fb\:post {
+	display: block !important;
+	margin-right: auto !important;
+	margin-left: auto !important;
+	overflow: scroll;
 }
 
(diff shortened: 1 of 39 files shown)


kjellr (approved):
This seems to work. I'm ok with it, but also not sure why we need so many `!important` rules? Is there a way around using those here?

allancole:
Yeah, I hate that I had to add those, but embeds/iframes use all kinds of unpredictable inline-CSS rules that simply can’t be controlled without an `!important` rule, ie:

image

I also looked into using a filter that would add a custom wrapper around all embedded (oEmbed) content but [Jetpack overrides and removes it](https://github.com/Automattic/jetpack/issues/1571), so this was the only way to control it.

kjellr:
Ok, makes sense. Thanks!

Issue #2179, opened 2020-06-25
AltoFocus: Problems with Older Posts button


KokkieH:
## Steps to replicate

1. Activate theme on a site containing 20+ posts.
2. Make sure Infinite Scroll is switched off
3. View the front/posts page, and scroll to the bottom
4. Click the Older Posts button to load more posts.

## Result

Additional posts loaded, but the Older Posts button remained stuck in its original position below the first batch of posts, instead of remaining at the bottom of the page:

<img width="984" alt="Screenshot_2020_06_25__06_20" src="https://user-images.githubusercontent.com/11873759/85653075-c7be7780-b6ac-11ea-82f2-2d50e7e414f7.png">

This can be observed on the demo site.

## Expected

For the button to always appear at the bottom of the posts page, below the last visible post.

* User report: Reported [via Twitter](https://twitter.com/vossbrink/status/1275878181237673991)

chaitanyamsv:
7441092-hc

donalirl:
Another case in #15451250-hc. I also want to note that when older posts are loaded, they stack on top of existing posts instead of showing underneath:
<img width="662" alt="altofocus" src="https://user-images.githubusercontent.com/30727666/86185988-25563680-bb7b-11ea-90be-a551ddaa81b9.PNG">

JessBoctor:
We had two more users report this issue in #3107734-ZEN and #3106720-ZEN

asheliza29:
Another user reported this problem to us today. Ticket #3110662-ZEN

formosattic:
+1 in 3106378-zen

KristinaKay:
Another report here: #3103983-zen

eamcgill90:
Reported here 22455777-hc

marutim:
Another user reported the issue here 22487968-hc

RealCarFax:
Another user has reported this issue here: 16369178-hc

fresatomica:
Another report: https://wordpress.com/forums/topic/overlapping-blog-posts/

kcswong:
Followed up user report after chat here: 3158830-zen

tavilla61:
Another instance here: 22790749-hc
Screenshot from the user showing the overlapping posts: https://www.dropbox.com/sh/fxl4jknrxtfdztu/AAAHAVTMrul4gqbLr_-CT_6Aa?dl=0&preview=Screen+Shot+2020-07-21+at+12.24.04+PM.png

[...]

PR #2264, opened 2020-07-16
Fix older posts button positioning after loading posts.

mendezcode:
This PR fixes one of the issues reported in #2179

There is currently another issue after clicking the "Older Posts"
button, which causes the posts to overlap each other. This will be fixed
in another PR.

### DIFF
--- a/altofocus/style.css
+++ b/altofocus/style.css
@@ -3189,10 +3189,16 @@ body {
   clear: both;
   display: block;
   padding: 1.75em 10%;
-  position: absolute;
   text-align: center;
   width: 100%;
   z-index: 1;
+  position: relative;
+  float: left;
+  top: 100%;
+}
+
+.infinity-success #infinite-handle {
+	top: auto !important;
 }
 
 #infinite-handle span {
--- a/altofocus/style.css
+++ b/altofocus/style.css
@@ -1,14 +1,15 @@
+@charset "UTF-8";
 /*
 Theme Name: AltoFocus
 Theme URI: https://wordpress.com/themes/altofocus/
-Description: AltoFocus is a theme for photographers, artists, and other creative types in search of a simple and easy way to display their work.
-Version: 1.0.9-wpcom
 Author: Automattic, Inc
 Author URI: http://automattic.com
+Description: AltoFocus is a theme for photographers, artists, and other creative types in search of a simple and easy way to display their work.
+Version: 1.0.0-wpcom
 License: GNU General Public License v2 or later
 License URI: http://www.gnu.org/licenses/gpl-2.0.html
 Text Domain: altofocus
-Tags: featured-images, grid-layout, one-column, portfolio, rtl-language-support, theme-options, translation-ready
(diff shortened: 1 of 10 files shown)


mendezcode:
The button was absolutely positioned and it was making it overlap with the posts.
My solution involves floating the button and making it relatively positioned.
There's a `top: 100%` property too, which I'm overriding when `.infinity-success` is added to the body.

altofocus/style.css
```diff
@@ -3189,10 +3189,16 @@ body {
   clear: both;
   display: block;
   padding: 1.75em 10%;
-  position: absolute;
   text-align: center;
   width: 100%;
   z-index: 1;
+  position: relative;
```
kjellr, line 3195: This needs to be changed in `altofocus/assets/stylesheets/shared/_infinitescroll.scss`, and then compiled here.

mendezcode:
Since there is no build tooling integrated into Altofocus, I installed Sass globally with `npm install --global sass`, and then compiled with `sass --no-source-map assets/stylesheets`. 

I then moved the `assets/stylesheets/style.css` to the theme's root directory.

As can be noticed in the [changed files](https://github.com/Automattic/themes/pull/2264/files), there are a lot of changes added, which make me a bit hesitant about pushing this change forward due to the amount of changes in the compiled sources.

@allancole since you worked some time ago with Altofocus, do you mind sharing the app/tool (and its configuration) used to compile the Sass for the theme?

Also, as an alternative method, you could also add [the changes](https://github.com/Automattic/themes/files/4933483/altofocus.diff.zip) and provide me with the compiled diiff, so I can use this as a patch for this PR:

danieldudzic:
@kjellr I have investigated the reason for the new posts overlap. This is caused by a failure inside of the re-layout function.
 
A list of freshly loaded posts should be set in:

https://github.com/Automattic/themes/blob/9306ee271fba16605ec673966ffaa6bea4b34539/altofocus/inc/jetpack.php#L180

But for some reason, when the re-layout function runs that global variable is not set.

So the whole thing stops on this check:

https://github.com/Automattic/themes/blob/9306ee271fba16605ec673966ffaa6bea4b34539/altofocus/assets/js/grid.js#L103

Tried debugging further but got stuck a bit.

Perhaps @jeffikus will have an idea as to why this is happening.

allancole:
Did some testing with this and I’m wondering if this is ultimately a Jetpack issue. When I test locally, and look at the console logs for `loadedPosts`, it properly returns an array of post ids and gathers them together as expected. But on wpcom it returns as an `undefined` variable. From what I can tell, this is happening because Jetpack on dotcom isn’t rendering  `altofocus_infinite_scroll_render` ([here](https://github.com/Automattic/themes/blob/b9983eb8217d12a5f07f9138a624e1f8dd14832a/altofocus/inc/jetpack.php#L22)) the same way it does locally with the latest version of the Jetpack plugin which shouldn’t be the case.

danieldudzic:
@allancole Not sure why this isn't working for you on WP.com - I have asked @jffng to test it on his sandbox and he confirmed, that it resolves the issues for him.

Since it also works fine on my sandbox - I'll merge and deploy the fix.

PR #2266, opened 2020-07-16
Seedlet: Opt in to line custom-line-height

### LINKED ISSUE #2265: Seedlet-blocks custom-line-height support opt-in now required.
I ran into this change while working on some FSE blocks and testing with the Seedlet-Blocks theme.  Perhaps this is already on your radar but I figured it would be worth a heads-up.  (@kjellr - I hope you don't mind but since you authored the Seedlet p2 post I figured you would be the right person to ping.)

## Steps to replicate
* Run Gutenberg master with seedlet blocks theme.
* In the editor select a post and look to customize the Line Height.
* Verify this option is no longer available.

Recent changes in Gutenberg have made the block level custom line hight options require a theme level opt-in.  That is, the "Line Height" option we used to have for blocks such as Paragraph:

Screen Shot 2020-07-16 at 2 28 57 PM

Is no longer present on Gutenberg master while using Seedlet Blocks.  For this to be available again, we will need to add `add_theme_support( 'custom-line-height' )` to the theme (https://github.com/WordPress/gutenberg/pull/23044#issuecomment-659475699).

kjellr:
Fixes #2265. [Seedlet already supports this](https://github.com/Automattic/themes-workspace/pull/44), but custom line height has turned into an opt-in rather than a default. 

Before:

Screen Shot 2020-07-16 at 6 42 46 PM

After: 

Screen Shot 2020-07-16 at 6 21 39 PM

### DIFF
--- a/seedlet/functions.php
+++ b/seedlet/functions.php
@@ -255,6 +255,9 @@ if ( ! function_exists( 'seedlet_setup' ) ) :
 		// Add support for responsive embedded content.
 		add_theme_support( 'responsive-embeds' );
 
+		// Add support for custom line height controls.
+		add_theme_support( 'custom-line-height' );
+
 		// Add support for experimental link color control.
 		add_theme_support( 'experimental-link-color' );
     


jffng:
Side note — it feels like theme authors will have two ways of controlling the editor and site appearance — via `theme.json` and `add_theme_support`.

PR #2267, opened 2020-07-17
Canard: Add wporg required fixes

danieldudzic:
This PR addresses WP.org requirements listed here: https://themes.trac.wordpress.org/ticket/85232


kjellr:
This is looking great! Just found two bugs: 

1. The focus does get trapped in the search field when its open, which is great. But the menu modal on small screens doesn't work the same way. When you've finished tabbing through each of the menu items, we need the focus to circle back to the menu close button. Currently, you just exit the menu and continue down the page:

tab

2. On small screens the tab order for the menu items doesn't match the visual order. I'm not sure if this is a requirement or not, but we should see if it's something we can fix? Currently, you tab to the title, then the menu, and then back to the search: 

tab

kjellr:
This is really confusing. 🙃 I'm seeing very different behavior depending on the browser I use. Do you see this too? 

Safari:

- Desktop: When tabbing forward, I can’t focus on the search button (or anything on the page after it). https://cloudup.com/cqlll9uR08t
- Mobile: Generally works pretty well, but when shift-tabbing backwards, I can’t proceed beyond the search button. https://cloudup.com/c1dqLeuWYmq

Firefox: 

- Desktop: Focus does not get trapped in the search modal. https://cloudup.com/cvv4xwl0E6e
- Mobile: Focus does not get trapped in the search modal (https://cloudup.com/cCHPt_bxYhL), or in the navigation modal (https://cloudup.com/c_DX_BuGODx).

Chrome: 

- Desktop: Works perfectly! 🎉
- Mobile: Focus works for the search modal, but does not get trapped in the navigation modal. https://cloudup.com/cXV8T0zKEY4

danieldudzic:
Safari:
* Desktop: Focus stops on the last navigation item before the Search icon
* Mobile: Focus gets correctly trapped for both Navigation and Search

Firefox:
* Desktop: Focus gets trapped in the Search fine: http://g.recordit.co/2EklNNC0TO.gif - but doesn't go further down
* Mobile: Focus get's trapped for search but not for Navigation, and also doesn't continue correctly below

Chrome:
* Desktop: Works fine for both Search and Navigation
* Mobile: Seems to be working fine for me for both Search and Navigation: http://g.recordit.co/DIF8MkxWe6.gif

kjellr:
Any idea why it's behaving so differently in different browsers? Is there something about the JS that doesn't work consistently across them or do you think it's rooted in the way each one handles tabbing or something?

kjellr:
@danieldudzic take a look at #2305 — @jffng sorted out a solution for Seedlet that may come in handy here too.

PR #2275, opened 2020-07-20
Seedlet: Allow all blocks with a link color to override the text color value set by custom background colors

Copons:

- When using a custom background color that changes the text color (e.g. picking a black background sets the text white), only change the link color if it's not specified by the block.
- Remove the override that only worked for blocks with `p` as root element (e.g. the Paragraph block).
- Add a missing fallback color for the `--wp--style--color--link` variable.

This change originates from a [discussion in Gutenberg](https://github.com/WordPress/gutenberg/pull/23945#discussion_r455596699), where we noticed that the theme's style specificity was getting in the way with the link colors on non-Paragraph blocks.

The missing fallback color was throwing a `postcss-css-variable` working when building the theme style.
The `--wp--style--color--link` variable is defined by Gutenberg, and it's supported by all current browsers.
For older browsers, the build system uses a fallback color.
Without providing one, the `ie.css` style showed something like this:
```css
a:active { color: undefined; }
```

It's not very clear by the diff because the change was mixed up with the `p` override removal, but basically I just did this:
```diff
&:active {
-	color: var(--wp--style--color--link);	
+	color: var( --wp--style--color--link, var(--global--color-primary) );	
}	
```

### DIFF
--- a/seedlet/assets/css/ie.css
+++ b/seedlet/assets/css/ie.css
@@ -1520,15 +1520,19 @@ a:focus {
 }
 
 a:active {
-	color: undefined;
-}
-
-p.has-background.has-link-color:not(.has-background-background-color) a {
 	color: #000000;
 }
 
 .has-link-color a {
 	border-bottom: 1px solid #000000;
+}
+
+.has-background:not(.has-background-background-color) .has-link-color a {
+	color: #000000;
+}
+
+.has-background:not(.has-background-background-color).has-link-color a {
+	color: #000000;
 }
 
 *:focus {
@@ -2194,6 +2198,30 @@ object {
 .wp-block-cover-image .wp-block-cover-image-text a,
 .wp-block-cover-image .wp-block-cover-text a {
 	color: currentColor;
+}
+
+.wp-block-cover .wp-block-cover__inner-container .has-link-color a {
+	color: #000000;
+}
+
+.wp-block-cover .wp-block-cover-image-text .has-link-color a {
+	color: #000000;
+}
+
+.wp-block-cover .wp-block-cover-text .has-link-color a {
+	color: #000000;
+}
+
+.wp-block-cover-image .wp-block-cover__inner-container .has-link-color a {
+	color: #000000;
+}
+
+.wp-block-cover-image .wp-block-cover-image-text .has-link-color a {
+	color: #000000;
+}
+
+.wp-block-cover-image .wp-block-cover-text .has-link-color a {
+	color: #000000;
 }
 
 .wp-block-cover:not([class*='background-color']) .wp-block-cover__inner-container {
@@ -2874,6 +2902,14 @@ dd {
 }
 
 .wp-block-media-text {
+	/*&[class*="background-color"]:not(.has-background-background-color),
+	&[style*="background-color"] {
+		.wp-block-media-text__content {
+			a {
+				color: currentColor;
+			}
+		}
+	}*/
 	/**
 	 * Block Options
 	 */
@@ -2907,10 +2943,6 @@ dd {
 
 .wp-block-media-text .wp-block-media-text__content > *:last-child {
 	margin-bottom: 0;
-}
-
-.wp-block-media-text[class*="background-color"]:not(.has-background-background-color) .wp-block-media-text__content a, .wp-block-media-text[style*="background-color"] .wp-block-media-text__content a {
-	color: currentColor;
 }
 
 @media only screen and (min-width: 482px) {
@@ -2995,6 +3027,10 @@ p {
 
 p.has-background {
 	padding: 20px;
+}
+
+p.has-text-color a {
+	color: #000000;
 }
 
 .a8c-posts-list__listing {
--- a/seedlet/assets/css/ie.css
+++ b/seedlet/assets/css/ie.css
@@ -2902,14 +2902,6 @@ dd {
 }
 
 .wp-block-media-text {
-	/*&[class*="background-color"]:not(.has-background-background-color),
-	&[style*="background-color"] {
-		.wp-block-media-text__content {
-			a {
-				color: currentColor;
-			}
-		}
-	}*/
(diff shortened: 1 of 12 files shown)


Copons:
> In my testing, this works for Headings but not Paragraphs:

@jffng my bad, I've missed one more override 🤦

kjellr:
Regarding the approach here, I originally tried using extra `:not()` rules instead of [the extra specificity being deleted here](https://github.com/Automattic/themes/blob/3dd2e8222f42b5885e7d5784c2e71505b1f73ff9/seedlet/assets/sass/base/_editor.scss#L34-L38), but at the time I found it to be pretty complicated and unwieldy to get right. Looking at that code fresh, I see I missed some use cases outside of the paragraph block though. 

In any case, I just tested this PR briefly, but noticed that this doesn't cover all cases (as noted above, I don't think the old styles did either so I'm not sure that it's a regression). But [this sample content](https://cloudup.com/cm958If1yWS) should come in handy for testing purposes.

In all of these cases, the links labeled "custom blue link" should be blue:

dotorgthemes test_wp-admin_post php_post=2793 action=edit (2)

Copons:
@kjellr thanks for sharing the sample content!
The combinations to test here are just so many and I really appreciate it! 🙇

Copons:
Compared to your screen, headers with background look fine to me. 🤔 

Screenshot 2020-07-21 at 15 41 28

On the other hand, all nested blocks fail when the container has a background but no link color, which is logic given how the selector works.

Maybe giving the double-not a more "positive" spin could be the answer.

jeyip:
I went ahead and recreated all of the blocks @kjellr made. Some of my results are different. 🤔

---
Screen Shot 2020-07-21 at 3 42 22 PM

---
Screen Shot 2020-07-21 at 3 43 56 PM

--- 
Screen Shot 2020-07-21 at 3 45 56 PM


---
Screen Shot 2020-07-21 at 3 46 30 PM

---
Screen Shot 2020-07-21 at 3 46 22 PM

---

Copons:
Thanks for double checking @jeyip!
I should have clearly marked this PR as WIP.
I have a clear idea on how to solve almost all the issues, although I can't understand why in "unstyled" blocks, Guten's default link color (dark blue in the screenshots) manages to override the theme's link color.
Or to put it better, I understand why, but I haven't found a clean way to make the theme color's win over Guten's.

Copons:
It seems everything is working fine with my latest change.
(Filter by `.scss` when reviewing to conserve your sanity!)

One quirk I've noticed when setting a **text** (not link) color.
If the block doesn't have a background color (or it's the theme's background color), the link will use the actual link color specified by the theme.
Otherwise, it will inherit the text color.

Screenshot 2020-07-22 at 12 13 40

I'm pretty sure the idea there is to automatically adapt the text color according to the background (e.g. picking a black background will turn the text, link included, to white).

This creates a little inconsistency, but I'm kinda inclined to disregard it.
Any thoughts?

jeyip:
### Testing
- For Chrome, Firefox, Safari, and Edge, everything works as advertised except for the inconsistency that Jacopo has pointed out. 
- Side note: Styles don't seem to work out in IE11, but to be fair, _everything_ looks borked there. If we are actively supporting IE11 for seedlet, then it seems like those issues might be better addressed in a separate PR. See example below.

---

Screen Shot 2020-07-23 at 12 39 49 AM

Copons:
> Side note: Styles don't seem to work out in IE11, but to be fair, everything looks borked there. If we are actively supporting IE11 for seedlet, then it seems like those issues might be better addressed in a separate PR. See example below.

@jeyip Yeah, that's expected unfortunately.
These colors work with CSS variables which are not supported by IE11.

If you look at the generated `ie.css` file, all rules using a variable defined in Gutenberg (as in this case `--wp--style--color--link`) automatically use their fallback (typically a variable defined in the theme itself) instead of `var( ... )`).

jeyip:
> @jeyip Yeah, that's expected unfortunately.
> These colors work with CSS variables which are not supported by IE11.

👍 Thanks for the clarification.

@Copons This might be a silly question, but could we add `currentColor` into the fallback assignment for the paragraph text? It seems like it would make sense given the class name `has-text-color`. I also did a quick verification. Adding the fallback _does not_ affect automatic adaptation of the text color according to the background (e.g. black background will turn the text, link included, to white).

```
p.has-text-color a {
	color: var(--wp--style--color--link, currentColor, var(--global--color-primary));
}
```

seedlet/assets/css/ie.css
```diff
@@ -2874,6 +2902,14 @@ dd {
 }
 
 .wp-block-media-text {
+	/*&[class*="background-color"]:not(.has-background-background-color),
```
jffng, line 2905: Leftover commented code.

seedlet/style-rtl.css
```diff
@@ -1848,6 +1858,14 @@ dd {
 }
 
 .wp-block-media-text {
+	/*&[class*="background-color"]:not(.has-background-background-color),
```
jffng, line 1861: Leftover commented code.

seedlet/style.css
```diff
@@ -1856,6 +1866,14 @@ dd {
 }
 
 .wp-block-media-text {
+	/*&[class*="background-color"]:not(.has-background-background-color),
```
jffng, line 1869: Leftover comments.

Copons:
Good catch @jffng! I've rebased and rebuilt, and those comments disappeared. 🙂

jffng:
Thank you @Copons @jeyip. I'm going to merge this because we want to more fully support custom link colors, and the addition of this PR covers more use cases than what we have now.

Also mentally bookmarking this PR as a good example of how challenging it is right now for a theme to support an increasing number of styling combinations provided by the editor. Hopefully the amount of highly targeted CSS provided by the theme can be reduced as global styles progresses.

PR #2298, opened 2020-07-27
Seedlet Blocks - Update header site title block text alignment attribute

Addison-Stavlo:
Updates the site-title block used in the header to use `textAlign` for the text alignment attribute.

**Background**
FSE blocks in core have been updated so that text alignment settings will be saved under the `textAlign` attribute instead of `align`.  This was for compatibility and clarity since the `align` hook uses this attribute to save block alignment settings. The site-title block seems to be the only block used in seedlet Template and Template Parts that was effected.

As a result this attribute was not being set and the header looked like:

Screen Shot 2020-07-27 at 12 25 15 PM

After this update:

Screen Shot 2020-07-27 at 5 09 18 PM


kjellr:
Thank you!

Issue #2284, opened 2020-07-22
Seedlet: keyboard navigation issues


jffng:
There are a handful of keyboard navigation issues that need addressed before the theme can be accepted in the dotorg themes directory.

- [x] The wp admin bar styles should not be affected by your styles (focus for example)
- [x] Mobile: Visual order is wrong, burger button focus should be before the logo focus (mobile)
- [x] Mobile: Within the menu, focus should be trapped inside the modal. Once you tab past the last menu item, focus should go on the close button or close the menu. Once you shift+tab on the first element focus should go on the close button (and if you shift tab on this, it should focus the last element in the menu).

kjellr:
This is all set, as per #2305

PR #2305, opened 2020-07-29
Seedlet: fix keyboard navigation

jffng:
**Summary**

This PR addresses #2284, implementing the following changes:

- Traps keyboard navigation in the mobile modal menu when its open.
- Removes theme focus styles from wpadmin bar.

Kapture 2020-07-29 at 15 09 27

**How to test**
- Check out this PR
- Add a Primary menu and a WooCommerce product or two to your cart
- Resize to the mobile viewport
- Use keyboard navigation 
- Verify that the keyboard navigation is trapped in the menu when you open it

I've tested on Safari and Chrome so far.

**To-do**
- [x] Change the tabbing order on mobile so that the menus buttons are reached first via keyboard. The visual order does not follow the page markup, and altering the tabindex doesn't seem like a straightforward solution.
- [x] Test thoroughly.
- [x] Consider refactoring this JS to a class.


seedlet/assets/sass/base/_reset.scss
```diff
@@ -110,10 +110,12 @@ a {
 }
 
 // Focus styles
-*:focus {
-	outline-width: 1px;
-	outline-style: dotted;
-	outline-color: var(--global--color-secondary);
+.site {
```
kjellr, line 113: Super-minor comment, but since this wrapper only contains one set of rules, I don't think it's necessary to nest these rules. Easier to just write `.site *:focus { ...`

jffng:
I pushed some changes that add a `tabindex` to the menu header elements so they tab through in the correct visual order on mobile. 

There is one issue with this approach: when a user is logged in, the site's menu buttons are reached via keyboard navigation before the toolbar elements in the admin bar. 

I'm not sure how to solve for this and solve the tabbing order issue on mobile. 

@danieldudzic do you have any ideas or think this acceptable?

allancole:
This works really well on desktop but on mobile, the tab key doesn’t display the menu. I’m also unable to open the menu by clicking it. 

seedlet-ally-menu-demo

I’m getting a missing jQuery issue which may be causing the problem.

image

Maybe try adding `jquery` as a dependency on this line?: 
https://github.com/Automattic/themes/blob/55a0c2af33b98db5a6442a9bdbc124fc04eb63c8/seedlet/functions.php#L383

allancole:
> I pushed some changes that add a tabindex to the menu header elements so they tab through in the correct visual order on mobile.

Is this something required by the TRT or just recommended? If it conflicts with the admin-bar maybe it’s worth reverting back to how it worked previously?

jffng:
> Is this something required by the TRT or just recommended? 

The TRT requires it, based on [this comment](https://themes.trac.wordpress.org/ticket/85664#comment:6).

jffng:
I removed the code that was causing the errors @allancole ran into.

It still faces this issue on mobile, where the tab order does not follow the visual order:

Kapture 2020-08-06 at 10 32 07

This is because on mobile, we move the position of the menu out of the order that it appears in the markup. Currently, my only idea for fixing this is:

- Rewriting the markup so that the buttons for **opening** the mobile menus exist outside the menu containers themselves, and can be placed in the correct order in the markup.
- Refactoring `primary-navigation.js` to **handle open and close events separately**, and focus the correct element when the menu opens.
- Make any associated CSS additions and refactors to display the open and close buttons correctly.

I can explore this next unless there are other ideas?

kjellr:
Your suggestion above seems like the best path forward: bringing the buttons outside of the menu itself, placing them in the correct semantic order, and then hiding the buttons on desktop.

jffng:
I implemented the recommendation above, refactoring the JS, markup, and CSS. @kjellr want to give this a spin?

Kapture 2020-08-11 at 17 43 19

kjellr:
@jffng this is working well for me, but I'm seeing one bug: On mobile if the Woo cart is empty, I'm able to tab out of the menu without closing it: 

tab

I _think_ in this case, the user should be prevented from tabbing at all, and the focus should stay on the close button (since there's nothing left to tab into). 

I also see that there's one small conflict in `header.php`. Once those are fixed, I think this is good to go.

jffng:
Ah good catch. Fixed and rebased.

PR #2313, opened 2020-07-30
Ibis: Editor Style and Font Fixes

ianstewart:
This PR has a few fixes.

- Correctly loads the Editor Styles for Ibis _after_ the Seedlet ones. Props @alaczek for pointing me in the right direction with her debuggin'.
- Adds some of the new front-end styles to the editor styles. Layout and fonts'll match now.
- Fixes a typo in the code I stole from Mayland. :)

PR #2315, opened 2020-07-30
Seedlet Blocks: Load theme setup with correct priority

ianstewart:
These styles should load _after_ the parent theme in order to correctly override the parent. Hence the priority. I'm not sure if it'll have a noticeable effect on the Editor? I only noticed this when I stole this code to work on the Ibis theme. Worth fixing if only for future code, ahem, borrowers, like myself.

Ref:

https://github.com/Automattic/themes/pull/2313/files#diff-0140ea09b827571eed7227805bc9a4d3R28

### DIFF
--- a/seedlet-blocks/functions.php
+++ b/seedlet-blocks/functions.php
@@ -25,7 +25,7 @@ if ( ! function_exists( 'seedlet_blocks_setup' ) ) :
 		add_editor_style( 'style-editor.css' );
 	}
 endif;
-add_action( 'after_setup_theme', 'seedlet_blocks_setup' );
+add_action( 'after_setup_theme', 'seedlet_blocks_setup', 999 );
 
 /**
  * Enqueue scripts and styles.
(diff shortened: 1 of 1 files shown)


seedlet-blocks/functions.php
```diff
@@ -25,7 +25,7 @@ function seedlet_blocks_setup() {
 		add_editor_style( 'style-editor.css' );
 	}
 endif;
-add_action( 'after_setup_theme', 'seedlet_blocks_setup' );
+add_action( 'after_setup_theme', 'seedlet_blocks_setup', 999 );
```
jffng, line 28: Could we change the priority to something that appears more intentional, like 1 after the default?

ianstewart:
> Is the intention here to get the child theme styles to load after the parent's?

Yep. Otherwise the child theme editor styles don't "stick". They're overridden by the parent theme styles that will wind up getting loaded later making them further down the cascade.

> If so, would we also want to change the priority of the front-end stylesheet enqueueing as well?

I haven't run into any issues yet on the front-end … but that doesn't mean much. :D Definitely maybe. :)

jffng:
Changed the priority for enqueuing the stylesheet on front-end as well, and lowered the priority to 11 (the default is 10). 

Thanks @ianstewart !

ianstewart:
🚀

PR #2316, opened 2020-07-30
Ibis: Update header defaults and screenshot to match

ianstewart:
- Update header to be a bit more compact with a custom font size.
- Add a screenshot to match the front end.

This should get Ibis to a state where it's good enough to start testing on WP.com with our Starter Designs and Patterns.


ianstewart:
This one's also pretty simple. I'll just merge it.

Issue #2129, opened 2020-06-16
Button 2: More Block padding


otpidusprime:
## Steps to replicate

- Activate Button 2
- Set the front page to show Latest Posts in Customizer
- Create a new Post with More Block

The **"Continue Reading →"** link is moved below the post divider in the theme, like this:

![Screen Shot on 2020-06-16 at 12:07:10](https://user-images.githubusercontent.com/29238672/84740044-530f7d00-afca-11ea-81cc-c850db0550bf.png)

- As a workaround, when we use the **Insert more tag** option in Classic Block instead, it renders correctly:

![Screen Shot on 2020-06-16 at 12:09:53](https://user-images.githubusercontent.com/29238672/84740058-586cc780-afca-11ea-95b9-ab5bcd6ae97b.png)

## Result

![Screen Shot on 2020-06-16 at 12:07:10](https://user-images.githubusercontent.com/29238672/84740044-530f7d00-afca-11ea-81cc-c850db0550bf.png)

## Expected

**"Continue Reading →"** link should be above the post divider, like this:

![Screen Shot on 2020-06-16 at 12:11:05](https://user-images.githubusercontent.com/29238672/84740188-936efb00-afca-11ea-8d06-7e2c91ea371a.png)

Upon further inspection, found that the **More Block** has this CSS `padding-top: 100px` which moves it below the post divider:

![Screen Shot on 2020-06-16 at 12:16:37](https://user-images.githubusercontent.com/29238672/84740569-3fb0e180-afcb-11ea-9bda-9f87a1993a7a.png)

* User report: #21420459-hc
* Follow-up ticket:

PR #2331, opened 2020-08-06
Button 2: More Block padding

swoodipto:

Wrap `more-link` _(added from Block Editor)_ inside a paragraph so that **"Continue Reading →"** does not shift below the post divider.

#### Related issue(s):

#2129 

#### BEFORE:

#2129-before-site
#2129-before-html

#### AFTER:

#2129-after-site
#2129-after-html

### DIFF
--- a/button-2/functions.php
+++ b/button-2/functions.php
@@ -116,6 +116,14 @@ function button_2_adjusted_content_width() {
 }
 add_action( 'template_redirect', 'button_2_adjusted_content_width', 0 );
 
+/** 
+ * Wrap more-link inside a paragraph 
+ */
+function button_2_more_link($more) {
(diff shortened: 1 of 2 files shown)


pbking (dismissed):
I believe this could be done with a css-only solution without introducing additional functions or wrapping the element in a `<p>` tag.

The `<a .more-link>` seems that it shouldn't be absolute positioned and eliminating the forced height and padding-top positions the element where it should be.

Therefore I suggest instead that we delete CSS rather than adding PHP to address.

pbking:
Made the proposed changes and requesting review by another.

MaggieCabrera (approved):
LGTM!

MaggieCabrera:
D52718-code

PR #2342, opened 2020-08-10
Spearhead: add basic structure and core styles

jffng:
This PR adds the initial project structure and base styles for Spearhead, using Seedlet as the parent theme to get the latest Gutenberg support out of the box.

Screen Shot 2020-08-10 at 2 01 21 PM


jeffikus:
Nice work @jffng will give this a full test and review tomorrow!

kjellr:
This is a great start! Seems to get the base styles in place nicely. I noticed a few obvious bugs right away: 

Screen Shot 2020-08-13 at 2 00 47 PM

Since this PR is primarily about the structural styles though, I _think_ the WooCommerce menu placement seems like the only one that would likely fit within this PRs scope.

spearhead/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "node-sass": "^4.13.1",
+    "npm-run-all": "^4.1.5",
+    "minimist": "^1.2.2",
+    "rtlcss": "^2.4.0"
+  },
+  "scripts": {
+    "start": "npm run watch",
+    "build:style": "node-sass assets/sass/style.scss style.css --output-style expanded --indent-type tab --indent-width 1",
+    "build:rtl": "rtlcss style.css style-rtl.css",
+    "build": "run-p \"build:*\"",
```
scruffian, line 25: this should be `run-s`

spearhead/assets/js/extend-blocks.js
```diff
@@ -0,0 +1,20 @@
+/* global wp */
+( function( blocks, domReady ) {
+    var styles = [
+        {
+            name: "core/group",
```
scruffian, line 5: What do we need this for?

spearhead/assets/sass/_block-extends.scss
```diff
@@ -0,0 +1,3 @@
+// Import the block extends
+@import "block-extends/button";
```
scruffian, line 2: Do we need this?

spearhead/assets/sass/_responsive.scss
```diff
@@ -0,0 +1,137 @@
...
+$horizontal_space: var(--global--spacing-horizontal);
+
+$breakpoint_sm: 482px;
+$breakpoint_md: 592px;
+$breakpoint_lg: calc( 780px + 48px ); // default width + the margins
+$breakpoint_xl: calc( 1247px + 48px ); // wide width + the margins
+$breakpoint_xxl: 1440px;
+
+// Responsive breakpoints mixin
+@mixin media( $res ) {
```
scruffian, line 18: I have a few questions about this:
1. Do we need to define this mixin again? Could we reuse the code from the parent theme and just redefine the variables?

2. Could we use the mixin from core?
https://github.com/WordPress/gutenberg/blob/master/packages/base-styles/_mixins.scss

spearhead/assets/js/extend-blocks.js
jffng: I left this in here as a pattern to follow for adding block styles. It's unused at the moment.

spearhead/assets/sass/_responsive.scss
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Required Variables
+ */
+$flexwidth: 100%;
+$horizontal_space: var(--global--spacing-horizontal);
+
+$breakpoint_sm: 482px;
+$breakpoint_md: 592px;
+$breakpoint_lg: calc( 780px + 48px ); // default width + the margins
+$breakpoint_xl: calc( 1247px + 48px ); // wide width + the margins
```
scruffian, line 14: should the margins be in a variable?

spearhead/style.css
```diff
@@ -0,0 +1,170 @@
...
+	--responsive--spacing-horizontal: calc(2 * var(--global--spacing-horizontal));
+	--responsive--aligndefault-width: 100%;
+	--responsive--alignwide-width: 100%;
+	--responsive--alignfull-width: 100%;
+	--responsive--alignwide-width-multiplier: calc(16 * var(--global--spacing-horizontal));
+	--responsive--alignright-margin: var(--global--spacing-horizontal);
+	--responsive--alignleft-margin: var(--global--spacing-horizontal);
+}
+
+@media only screen and (min-width: 482px) {
```
scruffian, line 34: why do we have media queries here that aren't using the mixin?

scruffian: oh that's because it's compiled! 🙄

scruffian:
This is looking good.

I'm not totally convinced about the way we are doing responsive widths. It seems like we are writing a lot of code, which I am not sure of the benefits of. Why won't we want to allow the content to scale up to a max width?

spearhead/assets/sass/_responsive.scss
jffng: 1. I think so, we would just need to extract the mixin itself from the [`_responsive_logic.scss`](https://github.com/Automattic/themes/blob/master/seedlet/assets/sass/structure/_responsive-logic.scss#L18-L79).

2. I could see a benefit to doing this if we used the same breakpoints as core also.

jffng (commented):
Thanks for the review @scruffian! I committed and addressed what I could of your feedback.

I'm also not convinced of the responsive styles, especially after [revisiting them in Twenty Twenty One](https://github.com/WordPress/twentytwentyone/pull/100). 

Do you think it's in a good enough place that we can merge and keep working on it?

scruffian (approved):
I think this is in a good place to merge and then iterate on

PR #2365, opened 2020-08-18
Ibis: New Templates, Template Parts, Typescale, and Grid

ianstewart:
This PR should add new Templates and Template Parts, with a new Typescale and Grid, to allow more "robust" site design options. It's a work-in-progress so I'll be adding to this draft.


ibis/style.css
```diff
@@ -100,8 +118,8 @@ GNU General Public License for more details.
 }
 @media only screen and (min-width: 1024px) {
 	:root:root {
-		--responsive--aligndefault-width: calc(750px - var(--responsive--spacing-horizontal));
-		--responsive--alignwide-width: calc(990px - var(--responsive--spacing-horizontal));
+		--responsive--aligndefault-width: calc(610px - var(--responsive--spacing-horizontal));
```
iamtakashi, line 121: I haven't tested the theme in-depth, but we probably should override `--responsive--spacing-horizontal` to `64px` as we worked on the grid with 32px gutters. If that's correct, the default will be `624` and the wide will be `1216`.

iamtakashi: Alternatively, we can worked out a grid with gutters of 30px that is equivalent to the vertical spacing and the height of the normal paragraph. If that makes more sense:

`--global--spacing-horizontal: 30px;`
`--responsive--aligndefault-width: calc(625px - var(--responsive--spacing-horizontal));`
`--responsive--alignwide-width: calc(1220px - var(--responsive--spacing-horizontal));`

alaczek: I tested this a little bit and the `--responsive--aligndefault-width: calc(610px - var(--responsive--spacing-horizontal));` feels very narrow - it amounts to 560px. I like Takashi's suggestion better, but fear it will still be too narrow. I _think_ we discussed making it wider - on par with Mayland?

iamtakashi: I'm hesitant to make the default much wider as the body font size is smaller (18px) than Mayland. I think a wider `wide` is useful, but the default width should be a comfortable width to read with a bare paragraph.

PR #2366, opened 2020-08-18
Seedlet: support filtering archive page titles

### LINKED ISSUE #2317: Seedlet: give users the ability to partially hide archive page headings
As of WordPress 5.5, theme authors can [filter of archive page headings](https://make.wordpress.org/core/2020/07/30/filtering-archive-pages-headings-in-wordpress-5-5/).

We should filter the header output in Seedlet so that we can give users an easy way to hide the prefix via CSS, if so desired.

Screen Shot 2020-07-30 at 3 26 07 PM

cc @kathrynwp

jffng:
This PR addresses #2317.

**To test**
- Activate seedlet and visit any page that uses the archive template
- Verify that the prefix is wrapped in a span with the CSS class `archive-prefix`


scruffian:
The `$prefix` and `$original_title` params were only added in 5.5.0, so we'll need to change the minimum version supported if we make this change.

scruffian (approved):
LGTM

jffng:
> we'll need to change the minimum version supported if we make this change.

Ah that's a good point. @jeffikus do you think this is acceptable to support #2317?

jeffikus:
Being a future focused theme, I think it's reasonable to start support for it at 5.5.0.

pbking:
Has a conflict and needs to updated minimum WP version to 5.5

seedlet/functions.php
```diff
@@ -106,20 +106,11 @@ function seedlet_setup() {
 		// Add support for editor styles.
 		add_theme_support( 'editor-styles' );
 
-		$editor_stylesheet_path = './assets/css/style-editor.css';
-
-		// Note, the is_IE global variable is defined by WordPress and is used
-		// to detect if the current browser is internet explorer.
-		global $is_IE;
-		if ( $is_IE ) {
-			$editor_stylesheet_path = './assets/css/ie-editor.css';
```
jffng, line 115: Not sure why we are still checking this here because we don't use or have IE specific stylesheets anymore.

jffng:
I brought this PR up-to-date and bumped the minimum required WP version of Seedlet to 5.5.0.

MaggieCabrera (approved):
This looks good to go

pbking:
reviewed and rebased against trunk

PR #2370, opened 2020-08-20
Update Seedlet foreground color

kjellr:
This PR addresses recent feedback about the hover states throughout the theme: 

> the contrast ratio between the default state and the hover state for several links, 
> that are missing a text decoration, is not enough.
> 
> `#333333` and `#3C8067` only has a contrast ratio of 2.69:1.
> 
> The easiest way to indicate the different link states is with a text decoration, and not only rely on color. For example if a link is not underlined in its default state, you can add an underline for the hover state. Or a border, outline, box shadow, background color etc.

The [WCAG guidelines suggest](https://webaim.org/blog/wcag-2-0-and-link-colors/) that text must have: 

> - A 4.5:1 contrast between the non-link text color and the background.
> - A 4.5:1 contrast between the link text color and the background.
> - A 3:1 contrast between the link text color and the surrounding non-link text color.

Our current text has: 

- ✅ A 12.63:1 contrast between the non-link text color and the background.
- ✅ A 4.69:1 contrast between the link text color and the background.
- ❌ A 2.69:1 contrast between the link text color and the surrounding non-link text color.

This PR updates the `--global--color-foreground` value from `#333333` to `#292929` which provides a 3.1:1 contrast ratio.

@melchoyce, should this satisfy the guidelines, or do we need something visual too?

### DIFF
--- a/seedlet/assets/css/ie.css
+++ b/seedlet/assets/css/ie.css
@@ -1510,7 +1510,7 @@ body {
 	font-size: 1em;
 	/* 1em; */
 	font-weight: normal;
-	color: #333333;
+	color: #292929;
 	text-align: left;
 	background-color: #FFFFFF;
 }
@@ -1577,7 +1577,7 @@ a {
 	box-shadow: 0 0 2px 2px rgba(0, 0, 0, 0.6);
 	clip: auto !important;
 	clip-path: none;
-	color: #333333;
+	color: #292929;
 	display: block;
 	font-size: 18px;
 	font-weight: bold;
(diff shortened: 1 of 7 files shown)


melchoyce:
I believe this would satisfy the guideline. Could you add a screenshot?

kjellr:
Actually, looks like a 3:1 ratio _and_ a visual indicator is required: 

https://www.w3.org/TR/2008/NOTE-WCAG20-TECHS-20081211/G183

I'll close this and merge it into the other PR I have going, which works with underlines.

kjellr:
Hmm nevermind. Let's hold off until we have definite clarification.

kjellr:
In the meantime, here are screenshots. It's pretty subtle:

Default Before|Default After
---|---
Screen Shot 2020-08-20 at 10 49 42 AM|Screen Shot 2020-08-20 at 10 49 22 AM

Hover Before|Hover After
---|---
Screen Shot 2020-08-20 at 10 49 46 AM|Screen Shot 2020-08-20 at 10 49 27 AM

kjellr:
Yeah, sounds like this isn't the issue. #2373 should meet their requirements instead.

PR #2373, opened 2020-08-20
Seedlet: Add a visual hover indicator for all links

kjellr:
This PR adds a visual hover indicator for all links, [addressing this feedback](https://themes.trac.wordpress.org/ticket/85664#comment:17):

> But the contrast ratio between the default state and the hover state for several links, 
> that are missing a text decoration, is not enough.
> 
> #333 and #3C8067 only has a contrast ratio of 2.69:1
> 
> The easiest way to indicate the different link states is with a text decoration, and not only rely on color.
> For example if a link is not underlined in its default state, you can add an underline for the hover state.
> Or a border, outline, box shadow, background color etc.

[Based on further clarification](https://themes.trac.wordpress.org/ticket/85664#comment:19), it sounds like all link hover states need a visual indication of change in addition to (or instead of) a color change. This PR makes sure that all links that are underlined lose their underlines on hover, and links that do not have underlines gain one on hover. 

In many cases, I felt like color change _and_ the visual indicator change seemed too prominent of a change on hover, so I used _only_ the visual indicator (This seems to be within the TRT guidelines, and is what the a11y project does, so I figure it's ok? Thanks @enriquesanchez for a gut check). 

This required a bit of padding re-working in the main menu, so please be sure to check that out. 

## Screenshots

**Before**
before-hover

**After**
after-hover

### DIFF
--- a/seedlet/assets/css/ie.css
+++ b/seedlet/assets/css/ie.css
@@ -4503,35 +4503,6 @@ nav a {
 	}
 }
 
-@media only screen and (min-width: 482px) {
-	.primary-navigation > div > ul > li > a {
-		line-height: 1;
-	}
-	.woo-navigation > div > ul > li > a {
-		line-height: 1;
-	}
-	.primary-navigation > div > ul > li > a:before, .primary-navigation > div > ul > li > a:after,
-	.woo-navigation > div > ul > li > a:before,
-	.woo-navigation > div > ul > li > a:after {
-		content: '';
-		display: block;
-		height: 0;
-		width: 0;
-	}
-	.primary-navigation > div > ul > li > a:before {
-		margin-bottom: -calc(1em + 0);
-	}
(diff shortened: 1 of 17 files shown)


melchoyce:
I think the post titles and "continue reading" would be more effective if they didn't have the underline by default, and instead gained it on hover. Generally I think underlining links be default is most successful in contexts where links appear within blocks of non-linked text (like a single link within a paragraph). (The title/continue reading link would need to be a different color, though.)

kjellr:
> (The title/continue reading link would need to be a different color, though.)

Yeah, this is the difficulty there. I think it looks somewhat unsophisticated to use the green that way. And the lighter text changes the hierarchy in a way that makes the title seem less prominent than it did before: 

Screen Shot 2020-08-20 at 1 02 06 PM

melchoyce:
Ah, I see what you mean :/

PR #2385, opened 2020-08-21
Seedlet: Clean up footer navigation

kjellr:
Cleans up the appearance of the footer menu, for both desktop and mobile screens.

Desktop 

Before|After
---|---
Screen Shot 2020-08-21 at 2 22 09 PM|Screen Shot 2020-08-21 at 2 15 43 PM

Mobile

Before|After
---|---
Screen Shot 2020-08-21 at 2 21 55 PM|Screen Shot 2020-08-21 at 2 15 57 PM

### DIFF
--- a/seedlet/assets/css/ie.css
+++ b/seedlet/assets/css/ie.css
@@ -4847,14 +4847,10 @@ nav a {
 
 .site-footer {
 	overflow: hidden;
-}
-
-@media only screen and (min-width: 822px) {
-	.site-footer {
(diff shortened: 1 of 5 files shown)


scruffian:
This test well for me. I tried removing the CSS altogether and it still seems to look fine. Do you think we need this code at all?

kjellr:
😄 You're right! Turns out that flexbox stuff isn't necessary.

PR #2387, opened 2020-08-21
Seedlet: Use WordPress global to detect IE support

### LINKED ISSUE #2386: Seedlet Child Themes: undefined variable is_ie
cc @ianstewart @kjellr 

## Steps to replicate
Activate a seedlet child them on a fse-experiment enabled local site. (e.g. ibis or seedlet-blocks)

## Result
View PHP notices polluting the front end on pretty much any page:

image

## Expected

No notices

## Cause

This line of code is at fault: https://github.com/Automattic/themes/blob/7d597af99c9a55e2ca1ce17c960507828b1aed2a/seedlet/functions.php#L117

I believe the problem is that `is_ie` is defined in the upper scope, rather than in the same scope as the function itself.

noahtallen:
Fixes #2386 by removing the `is_ie` variable in favor of the global `is_IE` variable exposed by WordPress here: https://github.com/WordPress/WordPress/blob/001ffe81fbec4438a9f594f330e18103d21fbcd7/wp-includes/vars.php#L107

### DIFF
--- a/seedlet/functions.php
+++ b/seedlet/functions.php
@@ -16,14 +16,6 @@ if ( version_compare( $GLOBALS['wp_version'], '4.7', '<' ) ) {
 	return;
 }
 
-/**
- * Determine whether the site is being requested from IE.
- */
-$is_ie = false;
-if ( preg_match( '~MSIE|Internet Explorer~i', $_SERVER['HTTP_USER_AGENT']) || (strpos($_SERVER['HTTP_USER_AGENT'], 'Trident/7.0; rv:11.0') !== false ) ) {
-	$is_ie = true;
-}
-
 if ( ! function_exists( 'seedlet_setup' ) ) :
 	/**
 	 * Sets up theme defaults and registers support for various WordPress features.
@@ -114,7 +106,9 @@ if ( ! function_exists( 'seedlet_setup' ) ) :
 		add_theme_support( 'editor-styles' );
 
 		$editor_stylesheet_path = './assets/css/style-editor.css';
-		if ( $is_ie ) {
+
+		global $is_IE;
+		if ( $is_IE ) {
 			$editor_stylesheet_path = './assets/css/ie-editor.css';
 		}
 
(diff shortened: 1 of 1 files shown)


seedlet/functions.php
```diff
@@ -19,9 +19,8 @@
 /**
  * Determine whether the site is being requested from IE.
  */
-$is_ie = false;
-if ( preg_match( '~MSIE|Internet Explorer~i', $_SERVER['HTTP_USER_AGENT']) || (strpos($_SERVER['HTTP_USER_AGENT'], 'Trident/7.0; rv:11.0') !== false ) ) {
-	$is_ie = true;
+function seedlet_get_is_ie() {
```
vindl, line 22: It turns out that WP core already declares `$is_IE` global that we should be able to reuse here.

https://github.com/WordPress/WordPress/blob/001ffe81fbec4438a9f594f330e18103d21fbcd7/wp-includes/vars.php#L93

The browser detection there seems to be more robust and a bit more performant since it uses `strpos` on `MSIE` instead of `preg_match` which is not really required (I mean, if we disregard the fact that it's already computed there 😄).

noahtallen: That sounds very nice! Do we support win_ie or mac_ie?? :P

I'm in favor of that approach instead, but I'm not sure how to test it out in the theme

david-szabo97: TBH, looks like it just had a typo. AFAIK, PHP variables are case sensitive.
In this case rather than `$is_ie`, we should just change it to `$is_IE`, don't we?

Since these are being used in functions, you will need to add `global $is_IE;` as well.

I don't think we need to care about whether it's `win` or `mac`. It's used for IE specific stylesheet loading so it should apply to both win and mac. (that's what `$is_IE` doing)

noahtallen: Good point, I will change this to point at `$is_IE`. I'm not sure that `$is_ie` was a typo since it is duplicating the same logic here, so it seems like it was just added as a way to check IE support without intending to rely on the logic in core

noahtallen:
cc @Automattic/theam What do you think about this approach (using the `is_IE` global var from wordpress?)

kjellr:
That seems like a more solid solution to me — can we perhaps add a little code comment to note that the variable is coming from WP and not from the theme?

noahtallen:
Cool, will add that comment and then merge :)

PR #2395, opened 2020-08-24
Seedlet: fix landmark roles

jffng:
This PR addresses the following Theme Review accessibility [feedback](https://themes.trac.wordpress.org/ticket/85664#comment:23): 

- The UL items inside the nav, in items_wrap, should not have the navigation role:
- Aside must not be contained in another landmark ("Aside is used for the widget area inside the footer landmark")
- The role="main" should be placed on the `main` element only, one per document

The one thing I'm unsure of is whether to remove the `aria-label` completely from the `ul` in the nav:

https://github.com/Automattic/themes/blob/b3c4428a023a737c0e23b37fde8058d7130ab8ce/seedlet/header.php#L61

### DIFF
--- a/seedlet/404.php
+++ b/seedlet/404.php
@@ -12,7 +12,7 @@ get_header();
 ?>
 
 	<section id="primary" class="content-area">
-		<main id="main" class="site-main default-max-width">
+		<main id="main" class="site-main default-max-width" role="main">
 
 			<div class="error-404 not-found">
 				<header class="page-header">
--- a/seedlet/archive.php
+++ b/seedlet/archive.php
@@ -12,7 +12,7 @@ get_header();
 ?>
 
 	<section id="primary" class="content-area">
-		<main id="main" class="site-main">
+		<main id="main" class="site-main" role="main">
 
 		<?php if ( have_posts() ) : ?>
 
--- a/seedlet/footer.php
+++ b/seedlet/footer.php
@@ -14,7 +14,7 @@
 
 	</div><!-- #content -->
 
-	<footer id="colophon" class="site-footer default-max-width" role="contentinfo">
+	<footer id="colophon" class="site-footer default-max-width" role="contentinfo" aria-label="<?php esc_attr_e( 'Footer', 'seedlet' ); ?>">
 
 		<?php get_template_part( 'template-parts/footer/footer-widgets' ); ?>
 
--- a/seedlet/header.php
+++ b/seedlet/header.php
@@ -58,7 +58,7 @@
 							'theme_location'  => 'primary',
 							'menu_class'      => 'menu-wrapper',
 							'container_class' => 'primary-menu-container',
-							'items_wrap'      => '<ul id="%1$s" class="%2$s">%3$s</ul>',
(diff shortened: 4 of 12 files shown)


kjellr:
> The one thing I'm unsure of is whether to remove the `aria-label` completely from the ul in the nav

I'm not totally sure myself. @enriquesanchez,  we'd love your input here.

jffng:
I think we should remove `aria-label="submenu"` from the ul in this case; given where it appears in the markup tree, adding an aria-label seems redundant:

Screen Shot 2020-08-25 at 9 11 18 AM

enriquesanchez:
>The one thing I'm unsure of is whether to remove the aria-label completely from the ul in the nav

@jffng The ARIA role should be applied only to the containing element. For example:

```
<nav role="nav">
  <ul>
    <li>...</li>
   ...
  </ul>
</nav>
```

Also important to note that `role=nav` can appear more than once in the document. So if there's a main navigation section in the header, but also a secondary navigation section in the footer, both sections can have the role applied.

This is opposite to other ARIA roles, such as `main`, which there can only be one per document.

kjellr:
Thanks @enriquesanchez. Once Jeff makes that change, would you be able to download and run through a few pages to make sure it's all looking right to you? Or else maybe @jffng can put it up on a demo site real quick. It would be good to just make sure we have our bases covered here. 

Thanks again!

jffng:
@enriquesanchez 

Since the `main` role should only be added once per document, would `aria-label="Main"` potentially be confusing? Could we do `aria-label="Main menu"` instead, as suggested [here](https://www.washington.edu/accessibility/web/landmarks/)?

enriquesanchez:
>would you be able to download and run through a few pages to make sure it's all looking right to you?

@kjellr Sure thing. Can I download it from this repo?

>Since the main role should only be added once per document, would aria-label="Main" potentially be confusing?

@jffng The `role` is different from the `aria-label`. They are interpreted and announced differently.

On an element marked up as:

```
<nav aria-label="Main">
    <ul><li>...</li></ul>
</nav>
```

... a screen reader when entering the element will announce it as: _"Navigation, Main"_ or _"Main Navigation"_ (depending on the screen reader/OS).

If you had `aria-label="Main navigation"` instead, it'll be announced as _"Navigation, Main navigation"_ or _"Main navigation Navigation ._

It's not the end of the world, but these little things can add up to the experience of the user.

kjellr:
> @kjellr Sure thing. Can I download it from this repo?

Yes! You can [download this branch here](https://github.com/Automattic/themes/archive/fix/seedlet-landmark-roles.zip). The download will include all the other themes in this repo, but all you need is the `seedlet` folder. Just drag that into your `/wp-content/themes` folder and you should be all set to activate it.

jffng:
@enriquesanchez I see, thank you for explaining this! So helpful. I've updated accordingly in this commit: https://github.com/Automattic/themes/pull/2395/commits/e4d46143a4667bdde275da01750c9a62cf4c37d2

enriquesanchez:
@jffng @kjellr 

Just downloaded and tested locally. The added landmark roles work great and make it so much easier to navigate the pages 👍

kjellr:
Excellent, thank you @enriquesanchez!

PR #2399, opened 2020-08-25
Seedlet: Add Alternating Grid block style for the Latest Posts block

kjellr:
This provides an "Alternating Grid" block style to match up with the original theme design. It displays posts in an alternating grid on desktop, but keeps them stacked and left aligned on tiny screens. 

It also displays the exact same way whether or not you have the "Grid" setting set up for the Latest posts block.

## Screenshots

dotorgthemes test_wp-admin_post php_post=2846 action=edit (1)

### DIFF
--- a/seedlet/assets/css/ie.css
+++ b/seedlet/assets/css/ie.css
@@ -2826,6 +2826,34 @@ img {
 	padding-right: 0;
 }
 
+@media only screen and (min-width: 592px) {
+	.wp-block-latest-posts.is-style-seedlet-alternating-grid {
+		overflow: hidden;
+	}
+	.wp-block-latest-posts.is-style-seedlet-alternating-grid > li {
+		width: calc(50% - 13px);
+		max-width: calc(50% - 13px);
+		text-align: right;
+	}
+	.wp-block-latest-posts.is-style-seedlet-alternating-grid > li:nth-child(2n + 1) {
+		float: right;
+		text-align: left;
+	}
+	.wp-block-latest-posts.is-style-seedlet-alternating-grid.is-grid {
+		display: inherit;
+	}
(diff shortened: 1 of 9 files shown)


kjellr:
It technically works with all of the standard latest posts block options, except for number of columns (if you have the grid block set up... but I think that's fine). It _does_ work when you show "full post", but it's pretty ugly. Luckily that's something the user can turn on and off.

kjellr:
Just added a block pattern. Might be overkill, but it does make this a little easier to find: 

Screen Shot 2020-08-25 at 12 46 43 PM

jffng:
We will need to open up a follow-up PR to update the headstart content, which will require a few placeholder posts to display correctly...

kjellr:
Thank you! I'll merge this into .com and work on updating the demo.

kjellr:
> We will need to open up a follow-up PR to update the headstart content, which will require a few placeholder posts to display correctly...

Yeah, we do want to update the headstart content — I'll open a new PR for that. I think we _might_ want to avoid adding a bunch of fake posts in the headstart content though, since those will be things that the user will have to manually delete.

allancole:
Ah, great solution!

kathrynwp:
I'm so pleased to see a special style being added, thank you!

I wonder if you'd consider also adding this special style to the **Blog Posts block** in addition to (or instead of) the Latest Posts block.

The Blog Posts block has some super useful settings that aren't in the Latest Posts block, like allowing allow users to pull in posts with a specific tag, or even select certain posts individually. It also _doesn't_ have an option to display the full post, so you'd avoid the potential problem you're discussing above. Thanks for considering!

kjellr:
> I wonder if you'd consider also adding this special style to the Blog Posts block in addition to (or instead of) the Latest Posts block.

Yeah, I'll look into adding it here too — it hopefully works the same way.

PR #2401, opened 2020-08-25
Add Alternating Grid block style to the Blog Posts block.

kjellr:
Followup to https://github.com/Automattic/themes/pull/2399. This PR adds an alternating grid block style to the Automattic Blog Posts block as well. Looks & works basically the same. 

This PR also removes the erroneous [double underline](https://cloudup.com/cb5lAfepVxi) that was showing up for post meta in the Blog Posts block. 

## Screenshot

dotorgthemes test_wp-admin_post php_post=2864 action=edit

### DIFF
--- a/seedlet/assets/sass/vendors/jetpack/blocks/blog-posts/_editor.scss
+++ b/seedlet/assets/sass/vendors/jetpack/blocks/blog-posts/_editor.scss
@@ -32,7 +32,9 @@
 	}
 
 	.article-section-title {
-		font-size: var(--global--font-size-base);
+		font-size: var(--heading--font-size-h4);
+		letter-spacing: var(--heading--letter-spacing-h4);
+		line-height: var(--heading--line-height-h4);
 		margin-top: 0;
 		margin-bottom: calc(0.5 * var(--global--spacing-vertical));
 	}
@@ -153,7 +155,7 @@
 		// Necessary so that the block boundaries are respected.
 		overflow: hidden;
 
-		h2 {
+		.article-section-title {
 			margin-left: calc(50% + (0.5 * var(--global--spacing-horizontal)));
 		}
 
@@ -176,6 +178,10 @@
 				.entry-meta {
 					justify-content: flex-start;
 				}
+			}
+
+			.more-link {
+				display: inline-block;
 			}
 		}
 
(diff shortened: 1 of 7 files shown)


seedlet/assets/sass/vendors/jetpack/blocks/blog-posts/_editor.scss
```diff
@@ -146,6 +145,51 @@
...
+// Block Style
+
+@include media(tablet) {
+	.wp-block-a8c-blog-posts.is-style-seedlet-alternating-grid {
+
+		// Necessary so that the block boundaries are respected.
+		overflow: hidden;
+
+		h2 {
+			margin-left: calc(50% + (0.5 * var(--global--spacing-horizontal)));
```
jffng, line 157: Should this be applied as an `nth-of-type` rule?

This is what the headings look like to me in the editor and front-end: 

Screen Shot 2020-08-26 at 8 13 16 AM

kjellr: Hmm, that's weird — I didn't see that before. I'll re-test. If anything, I can probably make that a direct-child rule.

kjellr:
> when I use the Blog Posts block, the editor uses the wp-block-a8c-blog-posts class to target this block, but the front end uses wp-block-newspack-blocks-homepage-articles:

Good catch! That's the old classname for this block. I've added it in. 

This should be ready for another test!

seedlet/assets/sass/vendors/jetpack/blocks/blog-posts/_editor.scss
kjellr: I think I fixed this in 974b049.

jffng:
LGTM! 🚢

kjellr:
@jffng, what do you think about replacing the "Latest Posts" block on the theme demo + headstart with this one? The Blog Posts block has more options, and I think will be more useful for folks. 

If you agree, I can get that started and update the headstart file in this PR too.

jffng:
Hmm while the theme demo + headstart is wpcom specific, I'm not sure about that because the Blog Posts block isn't a core block. I would imagine there is an audience of non-wpcom users coming to the demo site.

kjellr:
> Hmm while the theme demo + headstart is wpcom specific, I'm not sure about that because the Blog Posts block isn't a core block. I would imagine there is an audience of non-wpcom users coming to the demo site.

Yeah, I considered this too. The other weird part is the Block Pattern: If the demo and headstart use the Blog Posts block, we should probably update the block pattern to do so too. But then we can't really bundle that pattern for .org. We could strip it out and replace it with the latest posts version, but that gets a little complicated. 

In any case, let's leave this as is for now — we can always change it later. I'll go ahead and merge this in. 👍

PR #2414, opened 2020-08-31
Seedlet: Remove duplicate variables from style-editor.css

kjellr:
The editor variables were being imported twice:

https://github.com/Automattic/themes/blob/017e6f2572217d9849919fec88a6e0bd45be292a/seedlet/assets/sass/style-editor.scss#L5-L9

This PR removes the extra call. 

(Thanks @carolinan for finding this!)


jeffikus (approved):
LGTM!

PR #2417, opened 2020-08-31
Seedlet: Correct nested color rules

kjellr:
Today, Seedlet's background color rules sometimes override text color when it is applied to button text. 

To test this: 

1. Create a cover block OR a group block, and set its background to one of the default palette colors. 
2. Inside that block, add a button block that uses theme palette colors for the text and background. 
3. Observe that it appears as intended in the editor. 
4. On the front end however, notice that the text color is not correct. 

This PR ensures that custom text color always override the background-imposed colors. 

## Screenshots

Editor: 

Screen Shot 2020-08-31 at 3 37 15 PM

Frontend (Before):

Screen Shot 2020-08-31 at 3 37 07 PM

Frontend (After):

Screen Shot 2020-08-31 at 3 39 36 PM

### DIFF
--- a/seedlet/assets/css/ie.css
+++ b/seedlet/assets/css/ie.css
@@ -3804,58 +3804,88 @@ table th {
 	color: black;
 }
 
-.has-background:not(.has-background-background-color) a,
+.has-background:not(.has-background-background-color) a:not(.has-text-color),
 .has-background p, .has-background h1, .has-background h2, .has-background h3, .has-background h4, .has-background h5, .has-background h6 {
 	color: currentColor;
 }
 
 .has-primary-background-color[class] {
 	background-color: #000000;
+}
+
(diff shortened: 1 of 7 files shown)


kjellr:
Oof that is not a fun bug. Even if I remove the theme styles here, the default link color provided by Core overrides the button color. I'll need to dig in a little deeper.

kjellr:
@jffng I did more digging into the issue, and I think it's something that needs to be addressed in Gutenberg first. I opened [an issue here](https://github.com/WordPress/gutenberg/issues/24981) to hopefully get that started.

In the meantime, from what I can tell, that issue existed before this PR, so I think we should probably merge this for now. Does that sound alright to you?

jffng:
Sounds good, thanks for opening the issue. The number of custom color combinations (background * text * link) is increasingly challenging for the theme to support.

Issue #2416, opened 2020-08-31
Seedlet: load default editor palette on site creation


jffng:
When I create a new site on wpcom and set its theme to Seedlet, the site does not load Seedlet's default editor palette:

<img width="1904" alt="Screen Shot 2020-08-31 at 2 49 04 PM" src="https://user-images.githubusercontent.com/5375500/91755907-e142da80-eb99-11ea-9625-6e943eaca3ce.png">

jffng:
cc @jeffikus @kjellr if you're able to replicate, that would be helpful.

jeffikus:
@jffng tested the PR, LGTM!

jffng:
closed by #2418

PR #2418, opened 2020-08-31
Seedlet: de-register default editor palette correctly

jffng:
This PR addresses #2416. 

This moves Seedlet's default editor palette de-registration into the conditional that verifies the presence of a custom color palette loaded via the colors plugin.

**To Test**
- Spin up a fresh site on wpcom
- Activate Seedlet
- Open up a post and verify that the default Seedlet palette is loaded


jeffikus (approved):
Followed the testing steps and it works for me!

kjellr:
Good catch! This definitely makes sense.

PR #2439, opened 2020-09-09
Fix seedlet font translator comments

scruffian:
I think this is just a simple typo :)


kjellr (approved):
Good catch!

scruffian:
merged in r57041-wpcom

PR #2448, opened 2020-09-10
Update Seedlet to Version 1.0.8.

kjellr:
Version bump for Seedlet. 

This is relatively routine, but with one small adjustment: [In an earlier commit](https://github.com/Automattic/themes/commit/50957b5480c2d4d2c05429c09f1d1b2a7252c13b), the minimum PHP version was removed from `style.css`. This needs to be in there in order to pass ThemeCheck. 

This reinstates the PHP version requirement, but uses `5.2.4` instead of `7.3`. We'd had reports that folks were frustrated that they couldn't install this on servers running below `7.3`, and some light testing showed no errors. I used `5.2.4` because it is the same minimum PHP version used for Twenty Nineteen and Twenty Twenty, and I think it's fine here, but it could use another gut check.

### DIFF
--- a/seedlet/assets/css/ie.css
+++ b/seedlet/assets/css/ie.css
@@ -6,7 +6,7 @@ Author URI: https://automattic.com/
 Description: Seedlet is a free WordPress theme. A two-column layout and classically elegant typography creates a refined site that gives your works and images space to breathe - and shine. Seedlet was built to be the perfect partner to the block editor, and supports all the latest blocks. Writing, audio, illustrations, photography, video - use Seedlet to engage and direct visitors' eyes, without your theme getting in the way. And the responsive design shifts naturally between desktop and mobile devices. Seedlet is a great option for professionals and creatives looking for a sophisticated vibe. Whether you're looking to create a blog or a robust site promoting your business, do with simplicity, style, and Seedlet.
 Requires at least: 4.9.6
 Tested up to: 5.5
-Requires PHP: 7.3
+Requires PHP: 5.2.4
 Version: 1.0.8-wpcom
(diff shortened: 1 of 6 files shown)


allancole (commented):
Would love to just approve this but I’m not sure how to do a test with PHP 5.2.4, either :-/. 

Out of curiosity does the Gutenberg plugin have a minimum PHP version and if so, could we just sync it with that? 🤷🏽‍♂️

kjellr:
> Out of curiosity does the Gutenberg plugin have a minimum PHP version and if so, could we just sync it with that? 🤷🏽‍♂️

According to [the plugin page](https://wordpress.org/plugins/gutenberg/), the Gutenberg Plugin requires 5.6 or higher.

kjellr:
I ran the [PHPCompatibility linter](https://github.com/PHPCompatibility/PHPCompatibility) against Seedlet, and it appears to be compatible with PHP version 5.5 and above. Given that, I think the `5.2.4` version above is fine. Though I'm happy to just mirror what WordPress uses (`5.6.2`) if we'd like to be more conservative. 

@jeffikus would you mind offering up one last thought? 

Thanks!

jeffikus (approved):
LGTM! Ship it!

kjellr:
I'll update this on WP.org as soon as it's possible to upload a theme that uses the `block-patterns` tag. Just waiting for the automated ThemeCheck for the upload form to be updated there. 

I'll update this comment once it's uploaded there and in sync. 👍

kjellr:
This update was submitted to WP.org, minus the `block-patterns` tag. https://themes.trac.wordpress.org/ticket/89864

PR #2459, opened 2020-09-14
Add a rebuild tool for varia themes

scruffian:
This adds a shell script to make it easy to rebuild all Varia child themes in one go.

Just run `npm run children`

### DIFF
--- a/varia/rebuild-child-themes.sh
+++ b/varia/rebuild-child-themes.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+declare -a ChildThemes=("alves" "balasana" "barnsbury" "brompton" "coutoire" "dalston" "exford" "hever" "leven" "mayland" "maywood" "morden" "redhill" "rivington" "rockfield" "shawburn" "stow" "stratford" )
+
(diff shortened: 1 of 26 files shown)


varia/rebuild-child-themes.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+declare -a ChildThemes=("alves" "balasana" "barnsbury" "brompton" "coutoire" "dalston" "exford" "hever" "leven" "mayland" "maywood" "morden" "redhill" "rivington" "rockfield" "shawburn" "stow" "stratford" )
```
scruffian, line 3: This is the file that does the magic. Did I miss any?

allancole: Is it possible to explicitly select which themes to be recompile? 

Like if I wanted to just recompile **hever**, **alves**, **coutoire** and **dalston**, could I do that too?

scruffian: Sure, we could pass a param. Of course, if nothing changed in a child theme then the diff would be blank anyway...

Issue #2415, opened 2020-08-31
Coutoire: Headings bolded inside the Editor are not shown as bold on the live site


silvio-meter:
## Steps to replicate

Add the Heading Block to a post or a page inside the Editor.

![Optional alternative text if image doesn't load](https://testsite257com.files.wordpress.com/2019/12/heading-bold.png)
--

## Result

The headings aren't bold on the live site.

![Optional alternative text if image doesn't load](https://testsite257com.files.wordpress.com/2020/08/heding-2.png)

I also tested this with the Twenty Twenty theme and it works as expected.

## Expected

Heading should be bold on the live site as well. I managed to fix it with this CSS:

```
/*Making the titles blod -23319836-hc - SM*/
strong {
font-weight: bold;
}
```

* User report: [ #23319836-hc]

MaggieCabrera:
This was fixed in #2462

PR #2462, opened 2020-09-16
#2415 Coutoire: Fixed missing bold font on headings 

MaggieCabrera:

Added the bold font weight inside functions.php `coutoire_fonts_url` function because we only had 400 and 400i for Garamond. 

I also had to default the headings to a weight of 400 for the `bolder` rule to pick up the correct weight.

#### Related issue(s):

#2415

### DIFF
--- a/coutoire/style-rtl.css
+++ b/coutoire/style-rtl.css
@@ -1709,7 +1709,7 @@ h5, .h5,
 h6, .h6 {
 	font-family: "EB Garamond", serif;
 	font-family: var(--font-headings, "EB Garamond", serif);
-	font-weight: 200;
+	font-weight: 400;
 	clear: both;
 }
 
@@ -1837,7 +1837,7 @@ img {
 	font-family: "EB Garamond", serif;
 	font-family: var(--font-headings, "EB Garamond", serif);
 	font-size: 1.728rem;
-	font-weight: 200;
+	font-weight: 400;
 	line-height: 1;
 }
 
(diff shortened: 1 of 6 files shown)


coutoire/style-rtl.css
```diff
@@ -1662,11 +1662,6 @@ button[data-load-more-btn], .button {
 	font-size: 0.69444rem;
 }
 
-.wp-block-gallery .blocks-gallery-image,
-.wp-block-gallery .blocks-gallery-item {
-	width: calc( (100% - 16px) / 2);
-}
-
```
MaggieCabrera, line 1669: This is coming from a previous merge?

jffng: Seems like it.

jffng (approved):
Thanks @MaggieCabrera, this fixes the bug and works as expected.

It makes me wish though that the theme could already control the editor at this point, and disable the ability to bold a heading.

MaggieCabrera:
> It makes me wish though that the theme could already control the editor at this point, and disable the ability to bold a heading.

Can we actually do that? I agree that this theme wasn't intended to have bold headings as it is.

jffng:
At this point, our only option to avoid bold headings would be to add styles to "strong" that do nothing in the editor so that at least the behavior is consistent between front-end and editor. But we can't disable the UI for bolding a heading in the block itself yet.

scruffian:
> It makes me wish though that the theme could already control the editor at this point, and disable the ability to bold a heading.

Sounds like a good candidate for a core patch

PR #2470, opened 2020-09-17
Run build sequentially and recompile IE styles.

jffng:
This PR fixes the same issue as #2469. It also recompiles the IE editor styles that apparently hadn't been checked in.

To test:

- Make a change to Seedlet
- Run `npm run build`
- Confirm that the style.css, style-rtl.css, and ie.css files change.

### DIFF
--- a/seedlet/assets/css/ie-editor.css
+++ b/seedlet/assets/css/ie-editor.css
@@ -122,21 +122,37 @@ body {
 	font-size: 18px;
 }
 
-a {
+.wp-block a {
 	border-bottom: 1px solid #3C8067;
 	color: #000000;
 	text-decoration: none;
 }
 
-a:hover {
+.wp-block a:hover, .wp-block a:focus {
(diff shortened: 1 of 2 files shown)


jffng:
It's worth noting that the postcss mod used to generate the IE styles increases the compile time significantly. Might be worth splitting the build scripts out further to decrease compilation time during development.

scruffian:
Nice! Would there be a way that child themes can use the parent package.json, so if we want to make changes in the future we only have to update this one file?

jffng:
Could we move `package.json` to the root directory? This would have the added benefit of consolidating our development dependencies, and making it easier to package the themes for download on dotorg.

scruffian:
The problem is that different parent themes have different package.json files - so the Seedlet one is different from the varia ones. I'll think about a solution for Spearhead...

In the meantime I think you can merge this 👍

Issue #2382, opened 2020-08-21
Seedlet: mobile menu button flash


jffng:
A flash happens when the mobile menu is closed.

![mobile-menu](https://user-images.githubusercontent.com/5375500/90908121-f3569900-e3a1-11ea-9eff-7986e2709228.gif)

cc @kriskorn

PR #2473, opened 2020-09-17
#2382 swaped the z-index of the menu buttons to avoid flash

MaggieCabrera:

The flash was caused by the buttons changing state faster than the menu was disappearing. First I thought about having the buttons change state with the same transition time as the menu but their state changes instantly because the property affected is display: none;

So I thought having the Menu button over the actual menu but behind the close button would do the trick and avoid the flickering with a minimal change.

Screen Capture on 2020-09-17 at 17-30-12

#### Related issue(s):

#2382


scruffian:
LGTM 🚢

PR #2484, opened 2020-09-18
Remove width and height attributes from SVG icons as they are added dynamically

### LINKED ISSUE #2483: Seedlet: Remove duplicate width/height attributes from SVG icons
_Note: this issue was also reported on Twenty Twenty One bundled theme GitHub repository._

We shouldn’t use width and height attributes in the SVG icons worker class.

This currently results in duplicated attributes in SVG icons.

Worth noting that there is a DocBlock with this information in the related PHP file:

> Place each source on its own array key, without adding the
> both width and height attributes, since these are added dnamically,
> before rendering the SVG code.

(aside: the PR also fixes some typos in this DocBlock ^^)

audrasjb:
Fixes #2483 

Note: this issue was also reported on Twenty Twenty One bundled theme GitHub repository.

We shouldn’t use width and height attributes in the SVG icons worker class.

This currently results in duplicated attributes in SVG icons.

Worth noting that there is a DocBlock with this information in the related PHP file:

> Place each source on its own array key, without adding the
> both width and height attributes, since these are added dnamically,
> before rendering the SVG code.

(aside: the PR also fixes some typos in this DocBlock ^^)


kjellr (approved):
Great catch, thank you!

kjellr:
_Deployed to WordPress.com. Will be bundled with the next Seedlet release (1.0.9)._

PR #2494, opened 2020-09-22
remove the redundant if statement

### LINKED ISSUE #2413: Seedlet: Duplicate if statements
In the footer widget file, there are two checks to see if the widget area is active:
See
https://github.com/Automattic/themes/blob/master/seedlet/template-parts/footer/footer-widgets.php#L9
and https://github.com/Automattic/themes/blob/master/seedlet/template-parts/footer/footer-widgets.php#L13

## Expected
It seems unnecessary to check twice.

nunyvega:
Remove the [second]( https://github.com/Automattic/themes/blob/master/seedlet/template-parts/footer/footer-widgets.php#L13) `if` statement, as it's redundant. 

#### Related issue(s):
fixes #2413

### DIFF
--- a/seedlet/template-parts/footer/footer-widgets.php
+++ b/seedlet/template-parts/footer/footer-widgets.php
@@ -9,9 +9,9 @@
 if ( is_active_sidebar( 'sidebar-1' ) ) : ?>
 
 	<div class="widget-area">
-				<div class="widget-column footer-widget-1">
-				<?php dynamic_sidebar( 'sidebar-1' ); ?>
-				</div>
+		<div class="widget-column footer-widget-1">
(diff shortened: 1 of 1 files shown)


scruffian (approved):
LGTM

jeffikus:
[Recommended use of the sidebar has 3 options](https://developer.wordpress.org/reference/functions/dynamic_sidebar/#comment-424), we want to make sure the container div is not output then we should wrap it inside an if. With this approach, if dynamic_sidebar() returns false, output markup is an empty div:
`<div class="widget-column footer-widget-1"></div>`
Wrapped in an if statement, there will be no div unless the area has content.

MaggieCabrera:
I went ahead and merged this

Issue #2485, opened 2020-09-19
Morden: Pullquote block background color not affecting frontend


nielslange:
## Steps to replicate

1. Create a new page or post
2. Add the [Pullquote Block](https://wordpress.com/support/wordpress-editor/blocks/pullquote-block/)
3. Set the background color to blue

## Result

The background color of the pullquote is red, as defined in https://github.com/Automattic/themes/blob/18f104e1f33a2b51a81289a44115a845299cd5e0/morden/style-editor.css#L882-L884

## Expected

The background color of the pullquote is blue. It seems that due to CSS Specificity, the color selected by the customer is overwritten by the default color of the pullquote block.

* User report: 24249027-hc
* Follow-up ticket: 3331007-zen

MaggieCabrera:
This is affecting other Varia children (I have tested Coutoire. Looking into this.

MaggieCabrera:
This is affecting more than one block from Varia and its children. I'm looking into making a PR to fix this directly in Gutenberg, like this PR: https://github.com/WordPress/gutenberg/pull/25291

rinazrina:
Another report here #3326590-zen 

Theme: Stow

PR #2502, opened 2020-09-24
Fix/varia children pullquote color issue

MaggieCabrera:

Changed the inheritance of the pullquote block background color so that the selected color from the editor is more specific.

#### How to test 

Create a new Pullquote block, make it solid color, change the main color and see if the frontend has the same color as the editor.

#### Screenshots

Varia: 

Stow: (Has a color annotation issue still in progress so one of the colors is not showing correctly due to that)

Morden: 

#### Related issue(s):
#2485

### DIFF
--- a/alves/package-lock.json
+++ b/alves/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "alves",
-  "version": "1.4.6",
+  "version": "1.4.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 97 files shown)


scruffian (approved):
This looks good to me. I wondered if you needed to make the same changes in the _editor.scss file, but the more I look into that the more confused I am, so I think lets merge this as is and I'll follow up with another PR to look into that.

MaggieCabrera:
> This looks good to me. I wondered if you needed to make the same changes in the _editor.scss file, but the more I look into that the more confused I am, so I think lets merge this as is and I'll follow up with another PR to look into that.

Well, since the editor adds these changes inline, they were showing correctly, but I see your point.

PR #2503, opened 2020-09-24
Varia: Simplify pullquote styles

scruffian:
This simplifies the CSS for pullquotes. I think a lot of this CSS has moved to the block itself so it's not needed here anymore. It also ensures that the layout of the pullquote in the editor matches the front end.

#### Related issue(s):
From #2502


varia/sass/blocks/pullquote/_style.scss
```diff
@@ -27,44 +27,15 @@
...
 		color: #{map-deep-get($config-global, "color", "background", "default")};
 
-		&:not(.alignleft):not(.alignright) blockquote {
-			@extend %responsive-aligndefault;
-		}
-
-		blockquote {
-			padding-left: #{map-deep-get($config-global, "spacing", "unit")};
-			padding-right: #{map-deep-get($config-global, "spacing", "unit")};
-			max-width: inherit;
```
MaggieCabrera, line 65: This is changing how it looks on the frontend now compared to before this PR. It is closer to how it looks in the editor but I don't know which of the two options is the correct one.

scruffian: We discussed this on slack and decided to go with this  for now.

MaggieCabrera: LGTM!

MaggieCabrera (approved):
LGTM

PR #2507, opened 2020-09-25
Seedlet: Update the color annotations

scruffian:
I noticed an issue in Seedlet in the dark color scheme:

In the dark color scheme some of the contrast isn't very strong:
Screenshot 2020-09-25 at 18 17 51
Screenshot 2020-09-25 at 18 17 45

Rather than creating more selectors to deal with this, this PR proposes that we just update the CSS variables for this color scheme.

To test, apply this patch to your wpcom sandbox, set the site to Seedlet and then open the customizer.

When you switch to the dark color, you should see text with higher contrast, like this:

Screenshot 2020-09-25 at 18 29 16
Screenshot 2020-09-25 at 18 29 07

Also check that this looks the same when you edit the homepage.

### DIFF
--- a/seedlet/inc/wpcom-colors.php
+++ b/seedlet/inc/wpcom-colors.php
@@ -91,7 +91,7 @@ add_color_rule( 'bg', '#FFFFFF', array(
 ), __( 'Background Color' ) );
 
 // Foreground Color
-// --global--color-foreground-light
+// --global--color-background
 add_color_rule( 'txt', '#444444', array(
 
 	// Text-color
@@ -102,6 +102,7 @@ add_color_rule( 'txt', '#444444', array(
 			.has-background-dark-background-color[class],
 			.has-foreground-color[class],
 			.has-tertiary-background-color[class],
+			.has-tertiary-background-color[class]:not(.has-text-color),
 			.navigation,
 			.pagination .nav-links > *,
 			.post-navigation .meta-nav,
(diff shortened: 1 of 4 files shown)


MaggieCabrera:
https://github.com/Automattic/themes/blob/e9ca2b01ac5d307bb6d9f367c93dd08c73003859/seedlet/inc/wpcom-editor-colors.php#L70

Do we need to change it here too?

kjellr:
> Rather than creating more selectors to deal with this, this PR proposes that we just update the CSS variables for this color scheme.

I haven't tested, but my hesitation around this approach is just that it will fix this bug everywhere except in IE. (IE is basically the only thing preventing us from switching to CSS variables for all of this).

scruffian:
Right. I wonder if we should change the way we support IE11. I don't think global styles will work in IE11, so maybe its ok that this doesn't?

kjellr:
Yeah, I think the main concern is how we'd handle IE gracefully. It's super weird if someone customizes the colors of their website, but then it appears in a totally different color scheme for someone using IE. 🤔 

As you noted, this is something that core will need to figure out for Global Styles. Our approach had been to maintain the status quo for now (with IE support), but then reassess once core has their path forward 100% figured out.

scruffian:
@kjellr this now takes a different approach which should also work in IE11.

MaggieCabrera (approved):
It works fine for me on IE11. I suppose you know about [this polyfill](https://github.com/jhildenbiddle/css-vars-ponyfill) for css variables? I honestly don't know which approach I prefer personally.

scruffian:
Yeah, I assume we've decided not to do it that way for a reason....

kjellr:
> It works fine for me on IE11. I suppose you know about this polyfill for css variables? I honestly don't know which approach I prefer personally.

Yeah, I'm not sure — @jffng did that polyfill come up while we were exploring IE support?

PR #2515, opened 2020-09-28
Add PHPCS and PHPCBF to fix linting issues

scruffian:
This adds a PHPCS check before you commit a file. It will also try to automatically fix any issues it finds in the file for you.

It doesn't block you from committing if there are errors. It also doesn't work unless you run `composer install` in your root directory. I hope this means that contributors won't be preventing from committing, while enabling us to improve code quality.

To test:
- run `npm i` in the root of the repo
- then run `composer install`
- make a change to a file
- commit the file
- notice that any problems with the coding standards in the file are reported to you.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,6 @@ node_modules
 twentynineteen/
 theme-dev-utils/
 theme-dev-utils
+vendor/
 *.DS_Store
+
--- a/README.md
+++ b/README.md
@@ -26,10 +26,10 @@ By contributing code to our free themes, you grant its use under the [GNU Genera
 
 ## Testing pull requests
 
-### Using GitHub 
+### Using GitHub
 
 1. Clone repository locally `git clone git@github.com:Automattic/themes.git`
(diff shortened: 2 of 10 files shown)


jffng (commented):
I encountered a couple issues when running this for the first time:

- `lodash` is required by husky. I resolved this by running `npm i --save-dev lodash` at the root of the repo.
- Upon introducing a change to `seedlet/functions.php` that would result in a coding standards error to be automatically fixed, I received the following output: 

Screen Shot 2020-10-07 at 10 23 31 AM

MaggieCabrera:
I also had to install lodash but I got this when changing `varia/functions.php`:

Screenshot 2020-10-07 at 19 47 02

scruffian:
Thanks for testing. I've fixed those two points.

jffng:
Works for me! As a follow on, should we add something to the readme about this bit of tooling? It doesn't prevent someone from committing so I'm okay to merge as is.

scruffian:
I've added a note to the README about this. I'll p2 it too...

MaggieCabrera:
Thank you for this!

kjellr:
This is excellent, thank you!

PR #2528, opened 2020-09-29
fix for mobile scroll on alignfull galleries

### LINKED ISSUE #2471: Seedlet: Full width gallery triggers a horizontal scroll on mobile
## Steps to replicate

Create a new page with a full-width gallery (a portfolio template works. 
Try to scroll horizontally on a phone or small screen.

## Result

IMG_4919

## Expected

Expected no scroll and some padding on the right side of the gallery

MaggieCabrera:

The margin value for galleries was affecting the responsive style for full width galleries and causing a scroll. I omitted the margin in the case where the gallery is full width so that the `.alignfull` rules will apply to them.

#### To test:

Create a full width gallery and check that you can't scroll horizontally on mobile. The gallery should be full width.

#### Related issue(s):

Closes #2471

### DIFF
--- a/seedlet/assets/css/ie.css
+++ b/seedlet/assets/css/ie.css
@@ -2406,7 +2406,7 @@ object {
 	padding: 12px 13px;
 }
 
-.wp-block-gallery {
+.wp-block-gallery:not(.alignfull) {
 	margin: 0 auto;
 }
 
--- a/seedlet/assets/sass/blocks/gallery/_style.scss
+++ b/seedlet/assets/sass/blocks/gallery/_style.scss
@@ -1,6 +1,8 @@
 .wp-block-gallery {
 
-	margin: 0 auto;
+	&:not(.alignfull) {
+		margin: 0 auto;
+	}
 
 	.blocks-gallery-image,
 	.blocks-gallery-item {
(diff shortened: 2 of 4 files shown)


scruffian:
What happens if we just remove that rule altogether?

MaggieCabrera:
> What happens if we just remove that rule altogether?

it would affect non aligned galleries only but on my tests those would always feel the container, so I think it will be safe to remove

scruffian:
👍  to removing :)

MaggieCabrera:
> What happens if we just remove that rule altogether?

@scruffian Upon further testing I do believe we need it because the align left/right options don't add css for margin left/right to auto, so that would be the rule to apply to them. I think my fix is the way to go in this case

jffng (commented):
I gave this a spin and it fixes the issue. 

I also tried removing the margin rule, and it also fixed the issue, and doesn't appear to negatively impact the left and right aligned galleries.

MaggieCabrera:
> I also tried removing the margin rule, and it also fixed the issue, and doesn't appear to negatively impact the left and right aligned galleries.

Weird, it breaks right aligned galleries for me: https://d.pr/i/Iek5xA

I'm investigating more.

MaggieCabrera:
So I was about to make a PR to gutenberg to add the margin rule for the gallery block but when I tested I found out that that would actually break again the full width galleries. I don't think that setting the margin globally for the block is a good idea in this case because it may break other themes. I don't see how we could make a change to the block in this way without testing how older themes work and if they'd be affected.

scruffian:
Ok that makes sense. Lets merge this as is for now. I think it would still be good to understand why those margin were added to the alignright and alignleft classes. Are they in Seedlet too?

PR #2531, opened 2020-09-30
Seedlet: Compress the SASS build and add source maps

scruffian:
 - Use `clean-css` to compress the css files that we build, to make them smaller for production
- Add source maps to CSS files so that we can see what files the original rules live in

To test:
- Check that you can see the source file in the inspector when looking at CSS rules.
- Check that RTL ignore rules are preserved.


MaggieCabrera:
I think the rtl ignore rules are not working. 

```
/**
 * .alignright
 */
.alignright {
	/*rtl:ignore*/
	float: right;
	margin-top: 0;
	/*rtl:ignore*/
	margin-left: var(--global--spacing-horizontal);
}
```

float right is working when the website is in arabic:

Screenshot 2020-10-01 at 11 24 10

scruffian:
I think that's the expected behaviour. The rule is not being changed, so it is being ignored... unless I am missing something?

MaggieCabrera:
Yeah sorry, I was not understanding how that worked!

MaggieCabrera (approved):
I compared the RTL site with and without the changes and it looked the same. LGTM.

jffng:
> Use clean-css to compress the css files that we build, to make them smaller for production

Out of curiosity, do we have a way of measuring the performance benefit of this change on wpcom?

scruffian:
I decided to remove the CSS minification. I was concerned that removing the comment from the top of style.css might have unintended consequences, and it only saves 20KB. We can look at doing it again in the future, but for me the priority is to have the sourcemap, so I'm merging!

PR #2532, opened 2020-10-01
Spearhead: Add a dark color scheme

scruffian:
This adds a dark mode to Spearhead. To test, set your system to use a dark mode. The theme should look like this:
Screenshot 2020-10-01 at 12 54 33

cc @melchoyce if you want to tweak any of the colors.

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/2520


MaggieCabrera:
I think the contrast of the red may be a bit too low fro accesibility, I'm not sure how we handle that. https://webaim.org/resources/contrastchecker/ says it's a 2.46:1 ratio and apparently 4.5:1 would be the target.

It was really cool how the website changed instantly when I switched on dark mode, I didn't even have to reload.

The only issues with the dark mode styles that I've found are with the calendar widget. Here it is on the frontend:


Screenshot 2020-10-01 at 15 51 45


Without dark mode:

Screenshot 2020-10-01 at 15 51 38


Here on the editor:


Screenshot 2020-10-01 at 15 55 03

MaggieCabrera:
When creating a new page or post, the placeholders on the editor are very dark:

Screenshot 2020-10-01 at 16 32 31

MaggieCabrera:
When I select text on the editor (to make a link for example) the color of the text is the same as the highlighted text's background

Screenshot 2020-10-01 at 16 33 56

melchoyce:
Let's grab the colors from https://spearhead.co/'s original dark mode, if we haven't already:

image

scruffian:
Thanks for testing @MaggieCabrera. I added some code to deal with the calendar block and the selected text. The placeholder text needs a patch in Gutenberg: https://github.com/WordPress/gutenberg/pull/25796

jffng (approved):
Since https://github.com/WordPress/gutenberg/pull/25796 was merged, I tested this against Gutenberg master and can verify this is working well. Nice work @scruffian!

jffng:
Hmm I just tested the primary foreground color against the background and these were the results:

Screen Shot 2020-10-02 at 12 58 03 PM

While this works to validate dark mode, I'm wondering if the dark scheme colors need to be revisited from a design perspective as a follow up.

scruffian:
> While this works to validate dark mode, I'm wondering if the dark scheme colors need to be revisited from a design perspective as a follow up.

I added a followup issue here: https://github.com/Automattic/themes/issues/2538

PR #2545, opened 2020-10-05
Spearhead: Add color annotations

### LINKED ISSUE #2519: Add color annotations for WordPress.com (~12hrs).

scruffian:
This adds three new color pallettes for Spearhead.

Screenshot 2020-10-06 at 11 37 28
Screenshot 2020-10-06 at 11 37 16
Screenshot 2020-10-06 at 11 37 01

 

Closes #2519


kjellr:
👋 Just gave this a test, and the main thing I'm noticing is that the main text color when using the first palette is unreadable against the background: 

Screen Shot 2020-10-05 at 10 42 21 AM

In general though, where did these palettes come from? I'd recommend we just use the same three palettes we're using for all of the Varia themes (and Seedlet).

jffng (commented):
The color palettes are not loading for me in the editor. I am looking into why.

_Edit_
Color palettes in the editor are working fine for Seedlet. Could this be a cacheing issue?

scruffian:
I've changed how this works. I didn't realize that we used the same colors for all themes. It now works in the editor too. I also have simplified how the code works.

kjellr:
The colors themselves look good, and it's working great. I'm noticing just one weird thing: After I revert back to the default palette, the colors stop updating. Is that just me? 

colors

jffng (commented):
Looking good, but I ran into the same issue as @kjellr — whenever the "default" is selected it breaks the custom color selection.

scruffian:
@kjellr @jffng are your sandboxes up to date? I think I resolved that issue...

jffng:
Ah mine wasn't! That issue goes away, thanks.

`wpcom-colors-editor.php` still needs to be renamed to `wpcom-editor-colors.php` though for editor colors to work for me.

kjellr:
> Ah mine wasn't! That issue goes away, thanks.

+1! 🙇

PR #2550, opened 2020-10-05
Seedlet: fix alignment for galleries

### LINKED ISSUE #2544: Seedlet: Gallery alignment issues
## Steps to replicate

Add 3 galleries to a post and align them to the left, right and full width respectively. Add images that are not very big.

## Result

For the left/right aligned galleries on desktop they are not aligned inside the content column anymore. For the full width one, on big screens it ends up aligned to the left instead of stretching full width:

Screenshot 2020-10-05 at 09 31 23

On mobile the left/right aligned galleries are flush with the edge of the viewport:

Screenshot 2020-10-05 at 09 32 03

## Expected

The left/right aligned galleries should be contained within the main column, then aligned left/right. On mobile the left/right aligned galleries should have a margin with the edge of the viewport.
The full width gallery should stretch to the edges of the viewport and the caption should go below it.

MaggieCabrera:

- Made the full width galleries stretch to mirror the behavior of images with the same alignment (if we merge this to Seedlet we would need to make a PR for this to TT1, because that has the same issue):  https://github.com/Automattic/themes/pull/2550/commits/4edfa4eb0e8674a2bb561f1bbfe93e99ea19f794
- Ported [this PR](https://github.com/WordPress/twentytwentyone/pull/250) from TT1 to solve the align left/right issues

#### How to test:

Add 3 galleries to a post and align them to the left, right and full width respectively. Add images that are not very wide and open in a browser where they would need to stretch when aligned full width.

Screenshot 2020-10-05 at 23 32 10
Screenshot 2020-10-05 at 23 34 57
Screenshot 2020-10-05 at 23 35 09

#### Related issue(s):

Closes #2544


seedlet/assets/sass/blocks/utilities/_style.scss
```diff
@@ -14,22 +14,33 @@
...
+		/*rtl:ignore*/
+		float: left;
+
+		/*rtl:ignore*/
+		margin-right: var(--global--spacing-horizontal);
+		margin-bottom: var(--global--spacing-vertical);
+	}
+
+	.entry-content > .alignleft {
+		max-width: calc(50% - var(--responsive--alignleft-margin));
```
jffng, line 40: We had a convo about this in Slack, but wanted to post it here for good measure: 

What do you think about renaming this variable to `--responsive--alignleft-width`, and removing the unused `responsive-alignleft/right` mixins?

Can either be in this PR or a follow up.

jffng (commented):
This improves things on my end. I am curious to see how much more we can cut back on the alignment styles. They really are one of the most challenging aspects to support, so reducing their complexity on the theme's end is a good idea.

seedlet/assets/sass/blocks/utilities/_style.scss
MaggieCabrera: yeah, I'm going to work on this and maybe afterwards we can PR TT1 with the same changes

seedlet/assets/sass/structure/_responsive-logic.scss
```diff
@@ -86,14 +86,12 @@ $breakpoint_xxl: 1024px;
 	--responsive--alignwide-width: calc(100vw - var(--responsive--spacing-horizontal));
 	--responsive--alignfull-width: 100%;
 	--responsive--alignwide-width-multiplier: calc(16 * var(--global--spacing-horizontal));
-	--responsive--alignright-margin: var(--global--spacing-horizontal);
-	--responsive--alignleft-margin: var(--global--spacing-horizontal);
+	--responsive--alignrightleft-width: var(--global--spacing-horizontal);
```
MaggieCabrera, line 89: @jffng I decided to merge them into one to make it simpler since they are always the same anyway.

scruffian:
This looks like a good step forward 👍

seedlet/style-rtl.css
```diff
@@ -2400,16 +2396,24 @@ table th,
...
 	max-width: var(--responsive--aligndefault-width);
 }
 
+@media only screen and (min-width: 482px) {
+	.alignright {
+		float: right;
+		margin-left: var(--global--spacing-horizontal);
+	}
+	.entry-content > .alignright {
+		max-width: calc(50% - var(--responsive--alignrightleft-width));
```
jffng, line 2413: Do you think it's strange that we are doing another `calc` here? Could we just run one `calc` in the responsive styles, so that this line could look like this instead:

```suggestion
		max-width: var(--responsive--alignrightleft-width);
```

jffng: (Whoops, this should be a suggestion on the sass file)

MaggieCabrera: If I do that I'd have to make a calculation here: https://github.com/Automattic/themes/blob/2a25779dd771b417d5e95a273bdddf9365ac9aaf/seedlet/assets/sass/structure/_responsive-logic.scss#L146

jffng: Hmm why not here? 

https://github.com/Automattic/themes/blob/2a25779dd771b417d5e95a273bdddf9365ac9aaf/seedlet/assets/sass/structure/_responsive-logic.scss#L94

jffng (approved):
This fixes in the issue in my testing and has the benefit of simplifying the alignment styles. Thanks @MaggieCabrera!

PR #2551, opened 2020-10-05
Stratford: homepage headings in RTL do not align to the right

ramonjd:

Some of the headings on the stratford homepage do no align right in an RTL locale. This commit ensures that all headings within a block group are right-aligned.

**Before**

image

**After**

image

### Testing

Apply D49467-code patch:

In the site options, change the language to `he` or `ar` (or any RTL language)

Open https://stratforddemo.wordpress.com/

Check that the headings are right aligned. (The content won't be translated of course :) )

## Related issue(s):

First reported p58i-9ow-p2#comment-47478

### DIFF
--- a/stratford/style-rtl.css
+++ b/stratford/style-rtl.css
@@ -1567,6 +1567,7 @@ button[data-load-more-btn], .button {
 	padding: 0;
 }
 
+.wp-block-group h2.has-text-align-left,
 .wp-block-cover h2.has-text-align-left,
 .wp-block-cover-image h2.has-text-align-left {
 	text-align: right;


ramonjd:
@andrewserong said:

> I'm not sure but this might even be a CSS rule generated by the Varia -> child theme build process? From looking at the last commit on that file in Github: https://github.com/Automattic/themes/commit/709fbc5596b6ba50e31ee4db798ed02f178c0743, so perhaps it needs to be changed in Varia and then the build script needs running again. At the very least, hopefully an issue or PR in that repo will get the ball rolling.


👋 Reviewers, could you confirm if this is true? If so, what's the best way to fix this issue on Stratford? Thank you!!

andrewserong:
@scruffian I thought you might have a good idea about how to get this change in, since it looks like you last touched this file via https://github.com/Automattic/themes/pull/2503 ?

scruffian:
@jeffikus can we keep this in mind for a future planning session?

scruffian:
I think we should do this a different way, see https://github.com/Automattic/themes/pull/2761

scruffian:
Actually looking at this some more, it seems to be a deliberate decision in Gutenberg that if a left align is set like this we respect it. See https://github.com/WordPress/gutenberg/commit/916cbb0817f3c60b54067b361f56475acf511cee

I don't think we should change anything here. If we want the headings to change their alignment then we should align them with CSS rather than using the `.has-text-align-left` class.

PR #2555, opened 2020-10-06
Seedlet: Editor titles

scruffian:
The display of Post titles in Seedlet are the same in the frontend and the editor, but in Spearhead they aren't. This PR tidies the way that post titles are styled in Seedlet which also has the effect that in Spearhead the editor now matches the frontend:

To test
- switch your blog to Spearhead
- edit a post in the editor
- Check that the post title looks like this:

Screenshot 2020-10-06 at 14 22 22

Looking at this makes me wonder if we should just use the main styles in the editor too, rather than having a different stylesheet.


seedlet/assets/sass/components/entry/_header.scss
```diff
@@ -1,7 +1,24 @@
-.entry-title {
+/**
+* Editor Post Title
+* - Needs a special styles
+*/
+/*
+color: var(--global--color-primary);
```
jffng, line 6: Can these commented styles be removed?

jffng (commented):
~~I see how the PR sets up the post titles to be changed in the editor — but is supposed to actually change them for Spearhead?~~

~~I don't see changes to Spearhead checked in, so the editor titles look the same as before.~~

**Edit**
This is working for me, something was wrong in my local testing setup before but is fixed now :)

PR #2557, opened 2020-10-06
Spearhead: match the navigation to the design

scruffian:
This makes some changes the the menu so that it matches the design.

Screenshots:
Screenshot 2020-10-09 at 10 57 22
Screenshot 2020-10-08 at 15 29 45
Screenshot 2020-10-08 at 15 29 32

This has required some changes to Seedlet:
- I moved the markup for the open and close buttons on the menu, so that they could be positioned within the flow of the document. This change shouldn't impact Seedlet, since the buttons there are absolutely positioned.
- I removed the responsive mixins from Spearhead - we can import them from Seedlet directly
- We are also reusing the primary-navigation CSS from Seedlet
- I changed the primary-navigation breakpoints to use a SASS variable so that we could have the mobile menu use a different breakpoint on Spearhead
- I updated this file (spearhead/assets/sass/_header.scss) to use tabs not spaces


kjellr:
Just did a quick test — in this PR it doesn't look like tab focus is kept within the mobile menu. 

Before|After
---|---
tab|tab2

@jffng can confirm, but I _think_ that's the reason the close button was located in a different spot before.

MaggieCabrera:
The close button on mobile dissappears for Seedlet. I have to leave for a bit but I will test more later:

scruffian:
Thanks for the feedback on this PR. I have fixed both the issues mentioned, and made a few more tweaks to make it look better.

seedlet/header.php
```diff
@@ -28,30 +28,34 @@
...
+						<span class="hide-visually collapsed-text"><?php echo esc_html__( 'collapsed', 'seedlet' ); ?></span>
 					</button>
 				<?php endif; ?>
 				<?php if ( has_nav_menu( 'primary' ) ) : ?>
 					<button id="primary-open-menu" class="button open">
-						<span class="dropdown-icon open"><?php _e( 'Menu', 'seedlet' ); ?> <?php echo seedlet_get_icon_svg( 'menu' ) ?></span>
+						<span class="dropdown-icon open"><?php _e( 'Menu', 'seedlet' ); ?> <?php echo seedlet_get_icon_svg( 'menu' ); ?></span>
 						<span class="hide-visually expanded-text"><?php _e( 'expanded', 'seedlet' ); ?></span>
 					</button>
+					<button id="primary-close-menu" class="button close">
```
jffng, line 43: Moving the close button here has a side effect on the keyboard navigation in both Spearhead and Seedlet: the close button is no longer accessible by tab after you navigate away from it. You can use escape to exit, but I believe the close button also needs to be reachable by keyboard. 

Master | This PR
------- | ------
Kapture 2020-10-09 at 10 02 30 | Kapture 2020-10-09 at 10 03 24

Also in Seedlet, the site link which is not supposed to be reachable is now the first element you reach after tabbing away from the close button. 

We might be able to hack around this in the JS because we're already trapping the focus that way, but I'm not sure that's a great approach.

jffng (commented):
This is good except for a tricky keyboard navigation issue.

Also I'm not sure if this is a browser emulation issue, but on mobile Safari for me the top margin of the menu container appears off: 

Screen Shot 2020-10-09 at 10 11 21 AM

seedlet/header.php
scruffian: thanks for point this our, I added a commit to fix it.

scruffian:
> Also I'm not sure if this is a browser emulation issue, but on mobile Safari for me the top margin of the menu container appears off:

Please could you explain how to reproduce this?

jffng (commented):
The keyboard navigation when the menu is open is trapped correctly. 

I noticed three issues: 

1. The position of the close button should be in the top right, but it appears contained in the menu now.
2. The padding on the site title changes when the menu opens. It's also too close to the menu:
Kapture 2020-10-12 at 11 06 15
3. The keyboard navigation is a bit out of order. The site title should be reached before the menu. 

I'm using the mobile emulator on Chrome.

jffng:
@scruffian I made some changes that moves the position of the social navigation on mobile, and addresses 1 and 2 of my points above. 

It looks like this for me: https://cloudup.com/cdizfDOGYsU

scruffian:
This is looking good, just one small bug at this resolution:
Screenshot 2020-10-12 at 20 16 00

jffng:
I fixed one regression with the Seedlet mobile menu's close button. I'll merge this and open up two follow up issues to track the bug you mentioned and this one: 

> The keyboard navigation is a bit out of order. The site title should be reached before the menu.

Issue #2523, opened 2020-09-28
Adapt single page template, including related posts functionality (~6hrs).


scruffian:

jffng:
Closed by #2559

PR #2559, opened 2020-10-06
Spearhead: Add related posts block styles

jffng:
The design for a post suggests that at the end of the post, it shows related posts:

Screen Shot 2020-10-06 at 11 16 46 AM

From a templating perspective, this will be a very easy thing to support when block templates / FSE are stable.

In the interim, my proposal for addressing this (#2523) is by adding styles to the Jetpack related posts block. That way we avoid baking this functionality into a template, and can easily add related posts to a page or post as needed via blocks. This PR adds those styles.

Before | After
------- | -------
Screen Shot 2020-10-06 at 11 14 19 AM | Screen Shot 2020-10-06 at 11 13 30 AM

**To test**
- Check this PR out to a site that has Jetpack enabled, e.g. your sandbox
- Edit a page or post and add a related posts block
- Verify that the styles match the design


scruffian:
This looks good. The only thing I noticed is that on spearhead.co the titles are white (in dark mode) not red.

MaggieCabrera:
Looking at [Mel's figma](https://www.figma.com/file/N0vAffi3hSFufveaQMXBGA/naval) file I see some differences (I think this is an edited version that differs from what is online because Mel made some modifications):

* The title of the post should be:
```
font-family: BentonSans;
font-size: 20px;
line-height: 28px;
color: #000000; /* I believe it should be white on dark mode */
```
* The date of the post:
```
font-family: Nitti;
font-size: 13px;
text-transform: uppercase;
```
* There's a margin to the left of the first related post, the design doesn't have it:
Screenshot 2020-10-08 at 11 47 30

MaggieCabrera:
The "Related" title and the separator are missing too, maybe we should make this a block pattern to add the 3 things in one go?

jffng:
Thanks for the reviews!

@MaggieCabrera 

I believe [this is the Figma file we should be referencing](https://www.figma.com/file/qGQspfKbJmooNwhlT9DkG6/Spearhead?node-id=497%3A2006&viewport=416%2C762%2C0.27915963530540466); the one you shared are the original design comps but Mel made some modifications. So I'm matching the styles to this file. 

> There's a margin to the left of the first related post, the design doesn't have it:

Your screenshot doesn't look like the PR is being applied because neither the fonts nor margins look correct, here's what it looks like on my sandbox:

Screen Shot 2020-10-08 at 9 24 19 AM

> maybe we should make this a block pattern to add the 3 things in one go?

Good idea!

MaggieCabrera:
Sorry, my bad, I didn't have the sandbox properly updated, and I thought the figma file was the other one. The only comments I have are the post title color and the block pattern idea then :D

MaggieCabrera (approved):
LGTM, great work! :boat:

jffng:
Thanks! Note I'm going to leave the link colors as is — I think it makes more sense to keep them consistent than what's implemented on spearhead.co.

Issue #2560, opened 2020-10-06
Varia: New gutenberg options for search bar are broken on frontend


MaggieCabrera:
## Steps to replicate

* Install Gutenberg 9.1
* Insert a Search Bar block
* Choose "toggle search label"
* Change button position to inside 
* Click "Use button with icon"

## Result

Editor:

<img width="850" alt="Screenshot 2020-10-06 at 19 56 41" src="https://user-images.githubusercontent.com/3593343/95246779-43d06b80-0815-11eb-893e-6a0d2def9ae7.png">

Frontend:

<img width="942" alt="Screenshot 2020-10-06 at 19 53 32" src="https://user-images.githubusercontent.com/3593343/95246789-48951f80-0815-11eb-87c3-a8ac38bfccca.png">

## Expected

I expected the frontend to look like the editor

MaggieCabrera:
This is a gutenberg issue that I reported here: https://github.com/WordPress/gutenberg/issues/25902

scruffian:
This is now fixed for Seedlet and Spearhead.

scruffian:
Possibly connected to #1936

Issue #2524, opened 2020-09-28
Adapt archive / category page template (~4hrs).


scruffian:

MaggieCabrera:
For this issue I have:

* Styled the post excerpts (title, body, read more link): https://github.com/Automattic/themes/pull/2575
* Created a block pattern that has a styled Search block (but only partially works because of Gutenberg bugs): https://github.com/Automattic/themes/pull/2564

What's missing:

* The post subtitle
* The audio block after the excerpt for category pages
* Decide how we want to do the rest of the content of the Archive page (potentially adding blocks to the block pattern I created). We currently don't have blocks that lets us have Top posts and List posts by year.

MaggieCabrera:
I'm going to close this issue since what is missing has already issues created for them

PR #2564, opened 2020-10-07
Spearhead: archive block pattern

MaggieCabrera:

This PR is broken on the frontend because of this issue with Gutenberg (https://github.com/Automattic/themes/issues/2560) but I went ahead and finished the styles for the search block on the editor to at least have that. When the gutenberg issue is fixed, the frontend should look like the editor

* Search block
* Styles for the archive/category pages (the excerpt post styles go on a different PR: https://github.com/Automattic/themes/pull/2575 )

Screenshot 2020-10-09 at 19 39 28

* Added the category block + a title to the block pattern:

Screenshot 2020-10-13 at 12 24 07

* Cleaned up a bit the Search Block styles for other posible positions of the button with text/icon. This will probably need some design @kjellr :

Screenshot 2020-10-13 at 12 22 59

#### Related issue(s):

#2524


jffng (commented):
The search block looks great in the editor, nice work!

I wonder if we should add some padding to the button when the icon style is not active? 

Screen Shot 2020-10-09 at 3 00 27 PM

MaggieCabrera:
Waiting on [this PR](https://github.com/WordPress/gutenberg/pull/26052) to be merged for the search block fixes

spearhead/assets/sass/_archive.scss
```diff
@@ -0,0 +1,16 @@
+.archive {
+
+    .page-header {
+        line-height: 56px;
```
scruffian, line 4: I wonder if this should be an `em` value so it adapts to the font size variable

kjellr:
> Cleaned up a bit the Search Block styles for other posible positions of the button with text/icon. This will probably need some design @kjellr

Are these block options? I don't see them. 🤔

MaggieCabrera:
> Are these block options? I don't see them. 🤔

Yes, these are all new to Gutenberg 9.1, hence why they are buggy :D

kjellr:
Ah, got it. Ok well here's how I think these variations _should_ look based on the other mockups: 

search

That's all in Figma too if you want to take a look there.

MaggieCabrera:
> Ah, got it. Ok well here's how I think these variations _should_ look based on the other mockups:

Wow, so fast! Thank you, I'll get to it!

spearhead/assets/sass/block-extends/_categories.scss
```diff
@@ -0,0 +1,24 @@
...
+.wp-block-categories {
+	.cat-item {
+		line-height: 2.2em;
+		margin: 0 0 5px 0;
+	}
+
+	a {
+		color: var(--entry-header--color);
+		font-size: var(--latest-posts--title-font-size);
+		font-weight: var(--heading--font-weight);
```
MaggieCabrera, line 17: Looking at the figma these are `font-size: 20px` and `font-weight: 500`.

spearhead/inc/block-patterns.php
```diff
@@ -35,6 +35,19 @@ function spearhead_register_block_patterns() {
...
 				)
 			);
+
+			register_block_pattern(
+				'spearhead/archive-page',
+				array(
+					'title'      => __( 'Archive page', 'spearhead' ),
+					'categories' => array( 'spearhead' ),
+					'content'    => '<!-- wp:search {"label":"Search","showLabel":false,"placeholder":"Search...","buttonText":"Search","buttonPosition":"button-inside","buttonUseIcon":true} /-->
+					<!-- wp:spacer {"height":50} --><div style="height:50px" aria-hidden="true" class="wp-block-spacer"></div><!-- /wp:spacer -->
```
MaggieCabrera, line 45: I think the spacer could be quite smaller according to the spacing changes suggested by @kjellr. Right now the space between the line of the input and the title is 120px, on the figma file it's 80px.

MaggieCabrera:
We still need to address the styles for the [possible versions of the search block](https://github.com/Automattic/themes/pull/2564#issuecomment-707878367) but I think it would be fair to open a new issue/PR for them

spearhead/assets/sass/block-extends/_search.scss
```diff
@@ -0,0 +1,69 @@
...
+
+	// wrong class on gutenberg editor, we'll need to change it after that's fixed
+	&.wp-block-search__icon-button.wp-block-search__button-inside {
+		.wp-block-search__button {
+			padding: var(--button--padding-vertical) var(--button--padding-horizontal);
+		}
+	}
+
+	// wrong class on gutenberg editor, we'll need to change it after that's fixed
+	&.wp-block-search__text-button.wp-block-search__button-inside {
```
MaggieCabrera, line 16: Lines 9 and 16 need to swap after the gutenberg bug was solved

scruffian:
I updated the layout based on the figma file:
Screenshot 2020-10-15 at 12 17 26

I decided to make the "titles" 18px as that's the base font size, and then use 22px for the links so that they are noticeably bigger than the headings. This also means we can reuse existing variables which makes the theme more consistent generally.

I think we should follow up with the search block styles in a different PR.

MaggieCabrera:
Looks good to me!

scruffian:
We'll follow up with the search block changes in https://github.com/Automattic/themes/issues/2606.

PR #2565, opened 2020-10-07
Seedlet: Separate the navigation CSS into a different file

scruffian:
This separates the navigation CSS into a separate file, so we can dequeue it in child themes if necessary.

To test, check that the navigation works as before.


jffng (approved):
Works as advertised. We can address rtl styles in a follow up if you'd prefer to get the Spearhead menu unblocked by this.

scruffian:
I added the navigation RTL to this PR.

Issue #2538, opened 2020-10-02
Spearhead: Dark mode contrast


scruffian:
The color contrast in Spearhead's dark mode isn't very good for accessibility (see https://github.com/Automattic/themes/pull/2532). We should look at making some improvements to it.

MaggieCabrera:
Using https://app.contrast-finder.org We could have:

`#080B11` for the background color
`#F00048` for the accent color
`#080B11` apparently works fine for the body text
And white for the titles, hover on the links, etc like on https://spearhead.co/

I think we weren't using the same red/dark gray colors on our dark mode and the ones in the original site have better contrast, but still needed some adjustment.

kjellr:
I noticed that the background color used in dark mode is actually not as dark as the one used on Spearhead.co. 

Our Theme|Spearhead.co
---|---
<img width="1439" alt="Screen Shot 2020-10-05 at 10 40 53 AM" src="https://user-images.githubusercontent.com/1202812/95094808-4e491300-06f8-11eb-9416-29428640d7bc.png">|<img width="1440" alt="Screen Shot 2020-10-05 at 10 40 48 AM" src="https://user-images.githubusercontent.com/1202812/95094815-51440380-06f8-11eb-9237-6b1016214518.png">

The color scheme used there does pass contrast checks. 

`#1E1F21` for the page background.
`#B2B2B2` for body text.
And white for the titles, hover on the links, etc. 

We'll need to lighten the accent color though. Perhaps to `#FF2667`.

kjellr:
I'd still recommend some updates here. I took a look at Spearhead's existing color scheme, and mocked this up. It uses the same background colors that are present on the Spearhead site, and makes some adjustments to make sure the text passes (mostly the primary red/pink color). All text here passes AA with at least 4.5:1 contrast. 

![Dark Mode](https://user-images.githubusercontent.com/1202812/95897086-f1ce9f00-0d5a-11eb-9d2e-1c45fd618053.png)

![Dark Mode Article](https://user-images.githubusercontent.com/1202812/95897085-f1360880-0d5a-11eb-9500-1609b772796f.png)

scruffian:
The issue here is that we are using a same color for the background of pinned posts and single posts, but in light mode we use the normal background color for single posts and a darker version for pinned posts. This isn't insurmountable but it makes things more complex. Could we switch the colors round?

kjellr:
Yeah, we should swap it around to mimic the same treatment that the original theme uses.

[...]

PR #2567, opened 2020-10-08
#2538: Spearhead: Dark mode contrast

otpidusprime:

Update dark mode colors with the following values to improve contrast _(for accessibility reasons)_:

`#080b11` for background
`#b2b2b2` for the foreground _(body text)_
`#f00048` for the accent color

As proposed [here](https://github.com/Automattic/themes/issues/2538#issuecomment-703129124)

Furthermore, updated entry header color to white: `--global--color-foreground-light`

#### Here's a before-after comparison:

| BEFORE  | AFTER |
| before_spearhead_contrast | after_spearhead_contrast |

#### Related issue(s):

#2538

### DIFF
--- a/spearhead/variables.css
+++ b/spearhead/variables.css
@@ -41,7 +41,7 @@
 	--heading--line-height-h3: 1.4;
 
 	--entry-header--font-size: var(--heading--font-size-h1);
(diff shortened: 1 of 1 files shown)


scruffian (approved):
LGTM

PR #2573, opened 2020-10-09
Update/seedlet file block

### LINKED ISSUE #2500: Seedlet: Improve wrapping elements for the File block.
When the file block wraps onto multiple lines (or generally doesn't leave enough room for the button), things can look pretty odd: 

Screen Shot 2020-09-23 at 2 39 06 PM

Screen Shot 2020-09-23 at 2 40 22 PM

We should enforce that the button should always be on the right on desktop, and appear properly left-aligned below the text on smaller screens.

scruffian:
- Updates the file block CSS to push the button onto a new line
- Creates a new mixin to allow us to extend button styles.

#### Screenshots:
Screenshot 2020-10-09 at 14 45 07

#### Related issue(s):
Fixes #2500

If this is the right approach we should consider pushing these styles back to Gutenberg


seedlet/assets/sass/blocks/file/_style.scss
```diff
@@ -9,10 +8,17 @@
...
 
+	a:after {
+		content: '';
+		display: block;
+	}
+
 	.wp-block-file__button {
 		// Extend button style
-		@extend %button-style;
+		@include button-style;
```
scruffian, line 18: When this is an `extend` the padding declaration below doesn't work. That's why we need a mixin.

kjellr:
This is looking great in the front end: 

Screen Shot 2020-10-09 at 9 52 56 AM

But can we bring these changes to the editor as well? I'm still seeing this there. 

Screen Shot 2020-10-09 at 9 52 52 AM

jffng:
I pushed a commit applying this in the editor. 

I'm not sure we can overcome the underline extending all the way in the editor on filenames that extend two lines:

Screen Shot 2020-10-09 at 2 29 40 PM

kjellr:
> I'm not sure we can overcome the underline extending all the way in the editor on filenames that extend two lines

Yeah, since it's a textfield I think we're out of luck getting the `border-bottom` to wrap. Maybe we switch to using this here, and only in the editor?
 
```
text-decoration: underline; 
text-decoration-thickness: 1px; 
text-decoration-color: var(--global--color-primary);
```

It's not 100% exact, but at least it's similar. Alternatively, we could just remove the underline in the editor view.

jffng:
Looks good in Safari, Chrome doesn't support `text-underline-offset` yet but I think we can live with this for now:

Safari | Chrome
------- | -------
Screen Shot 2020-10-09 at 3 04 51 PM | Screen Shot 2020-10-09 at 3 06 43 PM

kjellr (approved):
LGTM!

jffng (approved):

PR #2574, opened 2020-10-09
Spearhead: fix unregister block pattern bug

jffng:
This PR addresses the following notice: 

```
Notice: WP_Block_Patterns_Registry::unregister was called incorrectly. 
Pattern "seedlet/latest-posts-alternating-grid" not found. 
Please see Debugging in WordPress for more information. (This message was added in version 5.5.0.) in /Users/maggie/A8c/repos/wp/wp-includes/functions.php on line 5225
```

The unregister calls are not working at the moment and I'm not sure why. Perhaps a child theme cannot unregister a block pattern from its parent theme? This removes them for now.

### DIFF
--- a/spearhead/functions.php
+++ b/spearhead/functions.php
@@ -5,7 +5,7 @@
  * @link https://developer.wordpress.org/themes/basics/theme-functions/
  *
  * @package WordPress
- * @subpackage Sk8prk
+ * @subpackage Spearhead
  * @since 1.0.0
  */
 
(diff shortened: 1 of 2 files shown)


MaggieCabrera (approved):
Yay! no more notices for me!

jffng:
@MaggieCabrera would you mind giving this another test? I followed your suggestion :)

The notices should be fixed, and when you try inserting a block pattern now, Seedlet's should not show up.

MaggieCabrera:
Yeah, it looks good, I don't see the patterns from Seedlet and the notices are gone.

PR #2575, opened 2020-10-09
Spearhead posts excerpts styles

MaggieCabrera:

This provides styles for the excerpt of the posts both on the front page and the archives/category pages.

* Changed the "More" link to fit the design
* Removed the Seedlet footer links for the excerpts.

#### Screenshots:

Screenshot 2020-10-09 at 17 24 18

#### Related issue(s):

#2524


jffng (commented):
This looks good to me!

My only comment is I wonder we should place an ellipsis between the excerpt text and the "More" link? It's a little abrupt to go straight from the text to the "More" button.

Screen Shot 2020-10-09 at 2 53 27 PM

scruffian:
I found a couple of issues with this approach:

Screenshot 2020-10-12 at 17 59 55
Screenshot 2020-10-12 at 17 59 36

I have a solution in mind for this, I'm just testing it

scruffian (changes requested):
See screenshots above

scruffian:
This is ready for another review

spearhead/template-parts/content/content-excerpt.php
```diff
@@ -0,0 +1,29 @@
+<?php
+/**
+ * Template part for displaying post archives and search results
+ *
+ * @link https://developer.wordpress.org/themes/basics/template-hierarchy/
+ *
+ * @package Seedlet
```
jffng, line 7:
```suggestion
 * @package Spearhead
```

jffng (commented):
This is looking good! I found one issue with the hover state of the more link:

Kapture 2020-10-13 at 12 10 26

Looks like it briefly gets an underline and changes the height of the element when the mouse leaves the target hover area.

kjellr:
I may have missed the discussion, but I see that this isn't inlined anymore. Was that idea abandoned? 

In any case, can we shrink the text size of the read more link to match the excerpt? We should also add an underline on hover to this link. 

Not sure if it's just that this PR is out of date, but I see this error when loading a post that uses the Latest Posts block. I don't see it on Master: 

Screen Shot 2020-10-13 at 12 20 44 PM

MaggieCabrera:
> Looks like the height of the element changes when the mouse enters the target hover area, and briefly gets an underline when the mouse leaves the target area.

This is actually a border, I believe: 

https://github.com/Automattic/themes/blob/68196181c6efc7d0488a9b5b69ca8b4aa7b1f7dd/spearhead/assets/sass/_text.scss#L2

jffng:
> I may have missed the discussion, but I see that this isn't inlined anymore. Was that idea abandoned?

It should still be inline, only when you add a "Read More" block would it appear on its own line.

> Not sure if it's just that this PR is out of date, but I see this error when loading a post that uses the Latest Posts block. I don't see it on Master:

I'm unable to reproduce this error. I did have to delete the branch and check it out fresh before I could view the latest changes.

jffng:
> can we shrink the text size of the read more link to match the excerpt?

The text sizes should match when a read more block is used, and it's actually a little smaller for the category / archive pages. Should they match? 

Screen Shot 2020-10-13 at 3 19 28 PM

> We should also add an underline on hover to this link.

👍

scruffian:
This is looking good to me. @kjellr if you have more changes we'll follow up in a future PR. Thanks!

kjellr:
@jffng: 

> The text sizes should match when a read more block is used, and it's actually a little smaller for the category / archive pages. Should they match?

Yeah, I think we should use the same size everywhere. For consistency's sake, maybe we try using the same font size as the post meta?

PR #2586, opened 2020-10-12
Seedlet: Define colorline functions in the theme

scruffian:
This moves the colorline functions into the theme, so that we don't have to use the library, which isn't on Atomic sites.

To test:
- Check out this branch and move the files to your sandbox
- Switch color pallettes in the customizer
- Check that the colors work on the front end too.


jeffikus (approved):
LGTM, ship it!

Issue #2590, opened 2020-10-13
Spearhead: Widgets need some visual refinement


kjellr:
- [x] The search box and button are gigantic: 

<img width="866" alt="Screen Shot 2020-10-13 at 9 49 19 AM" src="https://user-images.githubusercontent.com/1202812/95869671-a1931500-0d39-11eb-819f-ffc465db143c.png">

------

- [x] List-based widgets shouldn't inherit the default `ul` left padding. If it's removed, they look much nicer: 

Before|After
---|---
<img width="674" alt="Screen Shot 2020-10-13 at 9 52 06 AM" src="https://user-images.githubusercontent.com/1202812/95869790-c38c9780-0d39-11eb-9999-804510dedfd5.png">|<img width="545" alt="Screen Shot 2020-10-13 at 9 50 11 AM" src="https://user-images.githubusercontent.com/1202812/95869793-c4bdc480-0d39-11eb-8d9f-e01647c14b10.png">

MaggieCabrera:
The first item I'm working on it myself on https://github.com/Automattic/themes/issues/2576

scruffian:
The second item is done here: https://github.com/Automattic/themes/pull/2607

kjellr:
Awesome, this one's all set then. Thank you!

PR #2595, opened 2020-10-13
Spearhead: Search Results page

### LINKED ISSUE #2576: Spearhead: Search Results page
Style the Search Results page:

Screenshot 2020-10-09 at 17 47 17
 
Screenshot 2020-10-09 at 17 48 31

### LINKED ISSUE #2588: Spearhead: Post meta should be adjusted to match comps
Current post meta: 

Screen Shot 2020-10-13 at 9 45 36 AM

Recommended: 

Screen Shot 2020-10-13 at 9 46 06 AM

On single post pages, we can display the categories/tags/comments if we'd like. But they should be hidden on archive pages, to more closely match the design. Might also consider changing the date format to match the original theme if it's not too much hassle ("Sept 14 2020" Instead of "September 14, 2020")

MaggieCabrera:

This PR ended a bit on the bigger side, please do tell if it's better if I divided into parts. This PR styles the Search Results page.

* Styled the search form to fit the design. I had to make a decision about the icon on the button and went with the css route. It's not super clean but I think it's better than creating our own `searchform.php` file just to add an image. I reused the svg that Gutenberg uses and added it as a background. I also added a media query for dark mode. I also made use of seedlet's accesibility class to hide the text of the button.
* Simplified what is showed on the search results loop according to the design
* Created our own footer meta function for seedlet according to @kjellr design input in #2588

To test, add to the site url: `/?s=test`

#### Screenshots:

Screenshot 2020-10-13 at 17 31 01

#### Related issue(s):

Closes #2576 
Closes #2588 

Partially addresses #2590


spearhead/search.php
```diff
@@ -0,0 +1,60 @@
...
+						<?php
+						if ( is_sticky() && is_home() && ! is_paged() ) {
+							printf( '<span class="sticky-post">%s</span>', _x( 'Featured', 'post', 'seedlet' ) );
+						}
+						the_title( sprintf( '<h2 class="entry-title"><a href="%s" rel="bookmark">', esc_url( get_permalink() ) ), '</a></h2>' );
+						?>
+					</header><!-- .entry-header -->
+
+					<footer class="entry-footer default-max-width">
+						<?php spearhead_entry_meta_footer(); ?>
```
jffng, line 38: I'm getting an error here unless I change this to `seedlet_entry_meta_footer`. Is that correct, or should the filter functions be renamed instead?

jffng (commented):
Thanks for tackling this, I don't mind that it's one large PR as it's all related to getting the search styles in place. 

Here's what I found on first review: 

- The "no results" text appears full-width: https://cloudup.com/cOebfDJ0IrN 
- The search field itself feels tight and could use some extra padding on the text:   https://cloudup.com/cOebfDJ0IrN
- The bottom margin on the search form feels unnecessarily large and could be reduced to a single unit rather than `3 * global--spacing-vertical`: https://cloudup.com/cn6grS1q0D8

cc @kjellr as those last two comments are design-related

spearhead/assets/sass/_accesibility.scss
```diff
@@ -0,0 +1,38 @@
+/* Text meant only for screen readers. */
+.screen-reader-text {
```
scruffian, line 2: Rather than copy/paste this from Seedlet, can we import it as a helper, like we do for the navigation CSS?

spearhead/assets/sass/_forms.scss
```diff
@@ -0,0 +1,19 @@
...
+input[type="month"], 
+input[type="week"], 
+input[type="time"], 
+input[type="datetime"], 
+input[type="datetime-local"], 
+input[type="color"], 
+textarea {
+    background: transparent;
+    box-shadow: none;
+}
```
scruffian, line 19: I wonder if it would be better to add this to Seedlet as variables that we can override?

kjellr:
> The search field itself feels tight and could use some extra padding on the text: https://cloudup.com/cOebfDJ0IrN

Yeah, you can refer to [the mockups here](https://github.com/Automattic/themes/pull/2564#issuecomment-707878367) for the preferred spacing.

> The bottom margin on the search form feels unnecessarily large and could be reduced to a single unit rather than 3 * global--spacing-vertical: https://cloudup.com/cn6grS1q0D8

Agreed. 1 * the global spacing variable seems fine to be.

spearhead/assets/sass/_accesibility.scss
MaggieCabrera: oh you are totally right, I was unsure if this was the correct way to do it. I will fix!

spearhead/assets/sass/_search.scss
```diff
@@ -0,0 +1,62 @@
...
+        color: var(--entry-header--color);
+        @extend .screen-reader-text;
+    }
+}
+
+@media ( prefers-color-scheme: dark ) {
+    .search-form {
+        & > label {
+            &:before {
+                background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' id='search-icon' class='search-icon' viewBox='0 0 24 24' width='24' height='24'%3E%3Cpath fill='%23ffffff' d='M13.5 6C10.5 6 8 8.5 8 11.5c0 1.1.3 2.1.9 3l-3.4 3 1 1.1 3.4-2.9c1 .9 2.2 1.4 3.6 1.4 3 0 5.5-2.5 5.5-5.5C19 8.5 16.5 6 13.5 6zm0 9.5c-2.2 0-4-1.8-4-4s1.8-4 4-4 4 1.8 4 4-1.8 4-4 4z'%3E%3C/path%3E%3C/svg%3E");
```
scruffian, line 58: Can we just change the fill color with CSS rather than having a different icon?

spearhead/functions.php
```diff
@@ -192,3 +192,61 @@ function( $classes ) {
...
+	function seedlet_entry_meta_footer() {
+
+		seedlet_posted_on();
+
+		// Edit post link.
+		edit_post_link(
+			sprintf(
+				wp_kses(
+					/* translators: %s: Name of current post. Only visible to screen readers. */
+					__( 'Edit <span class="screen-reader-text">%s</span>', 'seedlet' ),
```
scruffian, line 213:
```suggestion
					__( 'Edit <span class="screen-reader-text">%s</span>', 'spearhead' ),
```

spearhead/search.php
```diff
@@ -0,0 +1,60 @@
...
+			// Start the Loop.
+			while ( have_posts() ) :
+				the_post();
+				?>
+
+				<article id="post-<?php the_ID(); ?>" <?php post_class(); ?>>
+					<header class="entry-header default-max-width">
+						<?php
+						if ( is_sticky() && is_home() && ! is_paged() ) {
+							printf( '<span class="sticky-post">%s</span>', _x( 'Featured', 'post', 'seedlet' ) );
```
scruffian, line 31:
```suggestion
							printf( '<span class="sticky-post">%s</span>', _x( 'Featured', 'post', 'spearhead' ) );
```

spearhead/variables.css
```diff
@@ -80,6 +80,12 @@
 	--sticky-posts--color-background: #F9F9F9;
 	--sticky-posts--entry-title-font-size: 24px;
 	--sticky-posts--entry-content-font-size: 20px;
+
+	/* Forms */
```
scruffian, line 84: I think these should be part of Seedlet

MaggieCabrera:
>     * The "no results" text appears full-width: https://cloudup.com/cOebfDJ0IrN

Good catch! Seedlet had the same issue and I went ahead and fixed it for both: https://github.com/Automattic/themes/pull/2595/commits/cd980a4f2e61ece9ee037e49c9c9e397233bd0bf

spearhead/variables.css
MaggieCabrera: Those variables do come from Seedlet, I'm just overriding their values here

spearhead/assets/sass/_search.scss
scruffian: Turns out, no we can't because the SVG is set in the CSS.

seedlet/assets/sass/elements/_forms.scss
```diff
@@ -19,6 +19,8 @@ textarea {
 	color: var(--form--color-text);
 	line-height: var(--global--line-height-body);
 	padding: var(--form--spacing-unit);
+    background: var(--form--background-color);
```
scruffian, line 22:
```suggestion
	background: var(--form--background-color);
```

seedlet/assets/sass/elements/_forms.scss
```diff
@@ -19,6 +19,8 @@ textarea {
...
+    box-shadow: var(--form--box-shadow);
```
scruffian, line 23:
```suggestion
	box-shadow: var(--form--box-shadow);
```

spearhead/assets/sass/_accesibility.scss
MaggieCabrera: As we discussed, I turned it into a mixin within seedlet and imported it on spearhead

kjellr (approved):
LGTM. Thanks @MaggieCabrera!

spearhead/functions.php
```diff
@@ -245,3 +245,61 @@ function spearhead_the_excerpt( $excerpt ) {
...
+
+/*
+ * Post footer meta
+ */
+
+if ( ! function_exists( 'seedlet_entry_meta_footer' ) ) :
+	/**
+	 * Prints HTML with meta information for the categories, tags and comments.
+	 */
+	function seedlet_entry_meta_footer() {
```
scruffian, line 257: Should this be called `spearhead_entry_meta_footer`?

spearhead/functions.php
```diff
@@ -245,3 +245,61 @@ function spearhead_the_excerpt( $excerpt ) {
...
+			'</span>'
+		);
+	}
+endif;
+
+if ( ! function_exists( 'seedlet_posted_on' ) ) :
+	/**
+	 * Prints HTML with meta information for the current post-date/time.
+	 */
+	function seedlet_posted_on() {
```
scruffian, line 285: Should this be called `spearhead_posted_on`?

MaggieCabrera: I thought so initially but if we rename it, we need to call again the new functions on their respective templates. Unless there's a better way of doing that that I'm not aware of.

scruffian (approved):
Few small comments but this LGTM

Issue #2593, opened 2020-10-13
Spearhead: Spacing adjustments


kjellr:
There are a handful of areas of Spearhead where the spacing doesn't match the comps: 

- [ ] Horizontal spacing between menu items: Should be 16px, is currently 12px. 

<img width="304" alt="Screen Shot 2020-10-13 at 9 29 08 AM" src="https://user-images.githubusercontent.com/1202812/95867134-d2257f80-0d36-11eb-85ed-6ec72c849d07.png">

------

- [ ] Top padding above the header should be 35px, is currently 60. 

<img width="1267" alt="Screen Shot 2020-10-13 at 9 37 42 AM" src="https://user-images.githubusercontent.com/1202812/95867997-c4bcc500-0d37-11eb-8a7d-9aff2ef74dfe.png">

------

- [ ] Site title bottom margin is 17.5, should be 4px. 

<img width="319" alt="Screen Shot 2020-10-13 at 9 39 48 AM" src="https://user-images.githubusercontent.com/1202812/95868315-15342280-0d38-11eb-9bb6-8c174e7f5b65.png">

------

- [ ] Spacing above post titles on single post pages is far too much.

Current|Spearhead Site
---|---
<img width="1266" alt="Screen Shot 2020-10-13 at 9 36 31 AM" src="https://user-images.githubusercontent.com/1202812/95868504-49a7de80-0d38-11eb-8c1f-962b2af3c334.png">|<img width="1266" alt="Screen Shot 2020-10-13 at 9 36 19 AM" src="https://user-images.githubusercontent.com/1202812/95868514-4d3b6580-0d38-11eb-95a2-7541fa3c8349.png">

------

- [x] "Continue Reading" link should be left aligned.

<img width="846" alt="Screen Shot 2020-10-13 at 10 07 33 AM" src="https://user-images.githubusercontent.com/1202812/95871728-fafc4380-0d3b-11eb-814f-07a1cdc5f80b.png">

jffng:
>  "Continue Reading" link should be left aligned.

This should be addressed by #2575

PR #2596, opened 2020-10-13
Spearhead: refine spacing

jffng:
This PR addresses most of #2593, making the following changes:

- Horizontal spacing between menu items to 16px
- Reduces top padding above header
- Reduces site title bottom margin
- Reduces spacing between single posts and header
- Fixes a bug in button spacing


spearhead/assets/sass/_posts-and-pages.scss
```diff
@@ -25,7 +25,10 @@
 
 .site-main > * {
 	margin: 0 auto;
-	padding: calc(3 * var(--global--spacing-vertical)) 0;
```
jffng, line 28: @scruffian do you know why this was set to such a large value and on such a generic selector?

MaggieCabrera: This is exactly what I was comenting on slack today. Even buttons are using calculations with this variable as a base (and they look huge because of it). This is inherited from Seedlet, and on Seedlet it's value is 30px.

spearhead/variables.css
```diff
@@ -52,6 +52,8 @@
 	--button--font-weight: bold;
 	--button--font-family: var(--global--font-secondary);
 	--button--font-size: var(--global--font-size-sm);
+	--button--padding-vertical: 24px;
+	--button--padding-horizontal: 28px;
```
MaggieCabrera, line 56: These look like variables that will only work well with that size for the button on the header menu, not the rest of the site's buttons (like the search one)

jffng: Open to suggestions on better values here. But I used the search block's button and main button block as references to determine them. Are there instances where these are breaking button appearances?

Screen Shot 2020-10-13 at 2 56 28 PM

MaggieCabrera: Yeah, there’s a work in progress PR for the search block, @kjellr designed some of the possibilities, maybe we can use them for reference? https://github.com/Automattic/themes/pull/2564#issuecomment-707878367

MaggieCabrera: I should have opened with that, sorry!

spearhead/assets/sass/_posts-and-pages.scss
scruffian: Yeah we do something similar in [Seedlet](https://github.com/Automattic/themes/blob/master/seedlet/assets/sass/structure/_vertical-margins.scss#L45), so this is following that pattern. I agree its problematic.

spearhead/variables.css
jffng: No worries, thanks! I took a look at those mockups and adjusted the variables to get close. We can iterate as needed in a follow up.

Issue #2594, opened 2020-10-13
Spearhead: Remove previous/next posts links


kjellr:
For Spearhead, we should remove the previous/next posts links: 

<img width="879" alt="Screen Shot 2020-10-13 at 9 53 44 AM" src="https://user-images.githubusercontent.com/1202812/95871989-4ca4ce00-0d3c-11eb-83ee-72eee669edc7.png">

The theme's aesthetic is pretty minimal, and there's no way for users to turn these on or off (plus, adding that capability seems like overkill), so I think they're ok to just remove.

PR #2600, opened 2020-10-14
Spearhead + Seedlet: make it possible to remove the previous / next post navigation (#2594)

jffng:
This PR addresses #2594. 

Previously in Seedlet, the post navigation was hardcoded into the template. This PR makes the following changes: 

- Refactors Seedlet's post navigation into a template tag so it can be overwritten in a child theme.
- Overwrites the template tag in Spearhead (disabling previous / next links). 
- Rename's Seedlet's post _pagination_ template tag function to be consistent with the actual WP function it calls, and updates its use respectively in the templates

Test
- Go to a post in Spearhead and confirm the Previous/Next links are gone.
- Go to a post and category / archive page in Seedlet and confirm the pagination and navigation work as they did before.

### DIFF
--- a/seedlet/search.php
+++ b/seedlet/search.php
@@ -42,8 +42,8 @@ get_header();
 				// End the loop.
 			endwhile;
 
-			// Previous/next page navigation.
-			seedlet_the_posts_navigation();
+			// Numbered pagination.
+			seedlet_the_posts_pagination();
 
 			// If no content, include the "No posts found" template.
 		else :
--- a/seedlet/archive.php
+++ b/seedlet/archive.php
@@ -37,8 +37,8 @@ get_header();
 				// End the loop.
 			endwhile;
 
-			// Previous/next page navigation.
-			seedlet_the_posts_navigation();
+			// Numbered pagination.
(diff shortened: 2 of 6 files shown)


seedlet/search.php
```diff
@@ -42,8 +42,8 @@
 				// End the loop.
 			endwhile;
 
-			// Previous/next page navigation.
-			seedlet_the_posts_navigation();
+			// Numbered pagination.
+			seedlet_the_posts_pagination();
```
MaggieCabrera, line 46: Should we keep this also on Spearhead? I think we should but maybe @kjellr has some input about this?

MaggieCabrera: "This" being the numbered navigation on the search results page

kjellr: Yeah, we should keep the numbered pagination on the search results and archive pages. Just the prev/next post nav from single post pages should be removed in Spearhead.

MaggieCabrera: then ⛵

MaggieCabrera (approved):

PR #2601, opened 2020-10-14
Spearhead: fix import error

jffng:
On master, I'm getting an import error when trying to build Spearhead because of a typo here:

https://github.com/Automattic/themes/blob/49eead899692a426a2ad97caf0665f04c7f5bb8b/spearhead/assets/sass/style.scss#L17

This PR fixes it and rebuilds the stylesheets.


spearhead/assets/sass/style.scss
```diff
@@ -14,7 +14,7 @@ Tags: one-column, flexible-header, accessibility-ready, custom-colors, custom-me
 */
 
 @import "../../../seedlet/assets/sass/abstracts/mixins";
-@import "../../../seedlet/assets/sass/base/accesibility";
+@import "../../../seedlet/assets/sass/base/accessibility";
```
scruffian, line 17: I think we should import these where they are actually needed

MaggieCabrera: That would be here, sorry for breaking things!

https://github.com/Automattic/themes/blob/49eead899692a426a2ad97caf0665f04c7f5bb8b/spearhead/assets/sass/_search.scss#L1

spearhead/assets/sass/_search.scss
```diff
@@ -1,3 +1,6 @@
+// Import screen reader classes from Seedlet.
+@import "../../../seedlet/assets/sass/base/accessibility";
```
jffng, line 2: @scruffian @MaggieCabrera is this a more logical place to import these styles?

spearhead/assets/sass/style.scss
jffng: No worries, thanks!

Issue #2592, opened 2020-10-13
Spearhead: Typography updates


kjellr:
There are a handful of areas of Spearhead where the font sizes don't match the intended ones from the comps: 

- [x] Menu: Font size should be 17px, is currently 16px. 

<img width="304" alt="Screen Shot 2020-10-13 at 9 29 08 AM" src="https://user-images.githubusercontent.com/1202812/95866990-9c809680-0d36-11eb-862d-cdbe1c997b3a.png">

---

- [x] Sticky post: Body font color should be white. 

<img width="871" alt="Screen Shot 2020-10-13 at 9 32 47 AM" src="https://user-images.githubusercontent.com/1202812/95867349-1022a380-0d37-11eb-8f82-162d0a533742.png">

---

- [x] Body line height should be 1.6.

<img width="824" alt="Screen Shot 2020-10-13 at 9 35 05 AM" src="https://user-images.githubusercontent.com/1202812/95867675-6a236900-0d37-11eb-8fb3-e599b4143271.png">

---

- [x] Single post date font size + color is too incorrect. Should be larger, white, and spaced out more. (This is a change from the comps, and will match the original theme design more closely). 

Current|Recommended
---|---
<img width="1266" alt="Screen Shot 2020-10-13 at 9 36 31 AM" src="https://user-images.githubusercontent.com/1202812/95868816-a99e8500-0d38-11eb-8d1f-616e4c11f286.png">|<img width="1266" alt="Screen Shot 2020-10-13 at 9 36 19 AM" src="https://user-images.githubusercontent.com/1202812/95868830-ac997580-0d38-11eb-85b7-28e56b44054b.png">

---

- [x] Site title in footer should be red with no underline. Imprint in footer should be set in Libre Baskerville.  

Before|After
---|---
<img width="398" alt="Screen Shot 2020-10-13 at 9 55 09 AM" src="https://user-images.githubusercontent.com/1202812/95870445-7e1c9a00-0d3a-11eb-8832-def7ba9b6778.png">|<img width="365" alt="Screen Shot 2020-10-13 at 9 56 49 AM" src="https://user-images.githubusercontent.com/1202812/95870452-807ef400-0d3a-11eb-9e56-165b4a142c0a.png">

PR #2602, opened 2020-10-14
Spearhead: refine typography styles

jffng:
This PR addresses #2592.


MaggieCabrera:
On the sticky post not all html elements are white. The `ul` tag and the heading are still gray (the inherit the color from the body)

Screenshot 2020-10-15 at 09 15 18

MaggieCabrera:
The navigation links font size is still 16px for me, not sure if I'm missing something:

Screenshot 2020-10-15 at 09 17 59

jffng:
Good catches, thanks @MaggieCabrera!

spearhead/variables.css
```diff
@@ -25,7 +25,7 @@
 	--global--content-width: 900px;
 	/* Line Height */
 	--global--line-height-base: 1;
-	--global--line-height-body: 1.7;
+	--global--line-height-body: 1.6;
```
kjellr, line 28: Sorry — I must've mistyped in the original issue. This should be `1.8`, not `1.6`.

kjellr (commented):
This looks good! Left one small comment above. Besides that, the only additional note I have is that we should increase the font size of the post date on single post pages to match `--global--font-size-base`: 


Current:
Screen Shot 2020-10-15 at 1 23 54 PM

Recommended: 
Screen Shot 2020-10-15 at 1 26 23 PM

(If at all possible, it would be great to change the date format so that the month isn't capitalized too — `Jun` instead of `JUN` — but that can happen separately.)

PR #2605, opened 2020-10-15
Spearhead block patterns for podcast links

### LINKED ISSUE #2556: Spearhead: Podcast links block
We have a block of links to the different sources for the podcast that we need to decide how we want to handle. It shows both on the home page and the Podcast category/archive page:

Home Page:

Screenshot 2020-10-06 at 16 36 54

Archive page:

Screenshot 2020-10-06 at 16 37 06

MaggieCabrera:

* Created two block patterns with the styles for the Podcast links

Screenshot 2020-10-15 at 13 07 51

#### Related issue(s):

Closes #2556


scruffian:
My concern with this approach is that if the user switches themes they would lose the custom CSS that makes this look nice.

What do you think about using a navigation block? The advantage of this approach is that we get all the block controls for free...

MaggieCabrera:
> My concern with this approach is that if the user switches themes they would lose the custom CSS that makes this look nice.
> 
> What do you think about using a navigation block? The advantage of this approach is that we get all the block controls for free...

Yeah, I considered that at first too. I thought that the navigation block would maybe be too much for this simple case (being all external links, not that that really matters a lot), and the css added by the theme wouldn't be that much that when swaped out it would look broken but the spacing would look off indeed. The navigation block certainly gives more control to the user on how it looks.

scruffian:
@MaggieCabrera this now uses the navigation block. What do you think?

@jffng I removed the CSS for the navigation block in Seedlet as they seemed to make the block unusable. Does this look OK to you?

spearhead/assets/sass/block-extends/_navigation.scss
```diff
@@ -0,0 +1,13 @@
+.wp-block-navigation {
+	.wp-block-navigation__container {
+		.wp-block-navigation-link {
+			margin: 0 calc( 0.5 * var(--global--spacing-unit) );
```
MaggieCabrera, line 4: I think this is a bit too much compared to what's on the original site (it goes into two lines)

scruffian:
@MaggieCabrera I added back the list pattern, and also made sure the editor styles match

MaggieCabrera:
LGTM! Again I can't review my own PR :D ⛵

jffng (commented):
> I removed the CSS for the navigation block in Seedlet as they seemed to make the block unusable. Does this look OK to you?

It would be good to keep Seedlet's nav block styles because they make it so the nav block matches the primary menu styling cc @kjellr. Is there a way we can approach this without removing them entirely? 

Master | This PR
------ | ------
Screen Shot 2020-10-15 at 2 29 31 PM | Screen Shot 2020-10-15 at 2 29 01 PM


Also we should address styling the nav block (e.g. the submenus) more thoroughly in a follow-up:

Screen Shot 2020-10-15 at 2 27 41 PM

kjellr:
> It would be good to keep Seedlet's nav block styles because they make it so the nav block matches the primary menu styling

That's right — removing it from Seedlet would also remove it from Seedlet blocks, which would be unfortunate.

scruffian:
Ok, I changed the approach here. I had to make a few tweaks to the way the Seedlet navigation block CSS works, but things are looking good in both Seedlet and Spearhead now. Ready for another review :)

MaggieCabrera:
This looks good to me, navigation seems ok for both themes and the block looks the way it should

PR #2608, opened 2020-10-15
Spearhead: Update dark mode colors

scruffian:
This updates the colors for Spearhead's dark mode to match those in https://github.com/Automattic/themes/issues/2538#issuecomment-708343956

Screenshot 2020-10-16 at 11 37 44


jffng (commented):
This works from a code perspective.

From a design perspective, these colors feel washed out / faded, especially when there's a sticky post and the lighter grey background dominates above the fold:

Before | After
------ | ------
Screen Shot 2020-10-15 at 12 04 27 PM | Screen Shot 2020-10-15 at 12 02 55 PM

I don't mind it though, especially since it actually passes accessibility. But it does feel a departure from the original palette.

kjellr:
This generally works well, but what I'm seeing on the homepage is flipped compared to the screenshots in https://github.com/Automattic/themes/issues/2538#issuecomment-708343956:  

It should use the darker color for the top sticky post, and the lighter one for the rest of the page. (Mirroring the behavior on [spearhead.co](https://spearhead.co)). 

Using the darker color for all other page backgrounds is working right here. 👍

scruffian:
Sorry my understanding from [this](https://github.com/Automattic/themes/issues/2538#issuecomment-708343956) was that we were going to switch it round to keep the code simple.

kjellr:
Ah sorry about that. I meant that we should mimic what the original theme does.

scruffian:
@kjellr I've switch those colors around now, if you wouldn't mind giving me another review. Thanks!

MaggieCabrera (approved):
LGTM!

PR #2610, opened 2020-10-15
Spearhead: Style all possible search block combinations

### LINKED ISSUE #2606: Spearhead: Update the search block to match designs
Design from @kjellr in https://github.com/Automattic/themes/pull/2564:

MaggieCabrera:

Added styles for all the possible combinations of the search block.

* To test this you need to have the latest version of the Gutenberg repo that has the latest bug fixes. Otherwise, it will be broken on both editor and frontend for you.
* Create a search block and try all the combinations of button with icon/text, outside, inside and with/without label

Editor:

Screenshot 2020-10-15 at 18 01 19

Frontend:

Screenshot 2020-10-15 at 18 01 06

#### Related issue(s):

Closes #2606


kjellr:
These look pretty good! I'm seeing just a couple bugs with the "Button inside" configuration, both with and without the search icon: 

Front end: 

Screen Shot 2020-10-15 at 1 17 13 PM

Editor: 

Screen Shot 2020-10-15 at 1 17 20 PM

spearhead/assets/sass/block-extends/_search.scss
```diff
@@ -5,53 +5,64 @@
...
-		.wp-block-search__button {
-			&.has-icon {
-				margin-left: 0;
-				background: transparent;
-				color: var(--entry-header--color);
-				padding: 0;
+		&.wp-block-search__icon-button {
+
+			.wp-block-search__inside-wrapper{
+				flex-direction: row-reverse;
```
MaggieCabrera, line 30: > These look pretty good! I'm seeing just a couple bugs with the "Button inside" configuration, both with and without the search icon:

@kjellr  the buttons should always be on the right side (my screenshots are on my local install with the gutenberg plugin downloaded from the repo directly installed), this bit of the code is what moves them to the right. Maybe you don't have the latest version? The old Gutenberg version didn't have the `.wp-block-search__icon-button` class correctly showing up

kjellr: Ok, it's mostly working now. Using the `master` for Gutenberg, the only bug I'm seeing is that I'm unable to get the Button inside with search icon configuration working: 

search-button

Is that due to the bug you're mentioning?

MaggieCabrera: No, that's the intended style for button inside, icon instead of text. If you are looking for this, then this is button **outside**, button with an icon (and I believe that you have the correct gutenberg version now):

Screenshot 2020-10-15 at 20 22 58

kjellr: Ha, you're right! Thanks for explaining. It's a little confusing, but it makes sense now.

kjellr (approved):
Design-wise, this is all set. 👍

scruffian (approved):
This looks good to me. Do you think we should move the code to Seedlet or does the search block look OK in Seedlet already?

MaggieCabrera:
> This looks good to me. Do you think we should move the code to Seedlet or does the search block look OK in Seedlet already?

Good catch, Seedlet needs some tweaks too since the changes to the search block (the ability to change the button to an icon) are quite new. I'll update this PR with that and simplify it for spearhead since it will inherit that

MaggieCabrera:
Updated Seedlet to support the new options on the Search Block too:

Screenshot 2020-10-16 at 11 20 23

seedlet/assets/sass/blocks/search/_editor.scss
```diff
@@ -1,24 +1 @@
...
-        &:focus {
-            border-color: var(--form--border-color);
-        }
-    }
-    .wp-block-search__button {
-        @extend %button-style;
-    }
-}
\ No newline at end of file
+@import "style";
```
scruffian, line 1: nice :)

seedlet/assets/sass/blocks/search/_style.scss
```diff
@@ -1,21 +1,58 @@
...
         padding: var(--form--spacing-unit);
+        background: transparent;
 
         &:focus {
             color: var(--form--color-text);
             border-color: var(--form--border-color);
         }
     }
+	.wp-block-search__button {
+        @extend %button-style;
```
scruffian, line 43: I think your editor has gone back to spaces here. We can also do this with a mixin rather than an extend.

MaggieCabrera: this was done like this by seedlet before, I can change that of course

seedlet/assets/sass/blocks/search/_editor.scss
scruffian: could we just change the import in the parent _editor.scss file?

seedlet/assets/sass/blocks/search/_style.scss
scruffian: Yeah I added the mixin to Seedlet, but I didn't change every use of the `extends` but we should change them as we encounter them.

scruffian (approved):
LGTM

Issue #2591, opened 2020-10-13
Spearhead: tab order of the header elements is incorrect keyboard navigation


jffng:
When navigating elements via keyboard tab, the menu button is reached before the site title. It should be the inverse.

Screen recording of the current behavior: 

![Kapture 2020-10-13 at 09 52 20](https://user-images.githubusercontent.com/5375500/95869898-e0c16600-0d39-11eb-8e30-a5f6a11868de.gif)

PR #2621, opened 2020-10-16
Seedlet: Fix keyboard navigation in the menu

scruffian:
This fixes https://github.com/Automattic/themes/issues/2591 by reordering the markup so that it's in the right order for keyboard navigation:

keyboard

This was also an issue in Seedlet, but it's fixed there now too:

keyboard2


jffng (commented):
There is a potential issue with the Seedlet order: when submitting Seedlet to the dotorg themes directory, we [got feedback](https://themes.trac.wordpress.org/ticket/85664#comment:6) that the focus of the menu needed to happen before the site title.

This is why the markup is so all over the place — because the menu in Seedlet changes position visually from desktop to mobile. This is not the case for Spearhead's design. 

That said, this feedback seems like a nitpick, and I'm comfortable merging this change to both Seedlet and Spearhead, and facing the consequences.

scruffian:
I think I disagree with [that feedback](https://themes.trac.wordpress.org/ticket/85664#comment:6). Semantically the site title should come before the menu button, and visually when reading left to right, you could argue that the site title comes first...

PR #2622, opened 2020-10-16
fixed width of contact form

### LINKED ISSUE #2613: Spearhead: Contact form
Looking at the demo site, I think the contact form page needs some work.

Screenshot 2020-10-16 at 10 09 20

MaggieCabrera:

Made the jetpack form `form` tag extend to 100% of the width of the parent. Fixed for Seedlet and Spearhead.

Screenshot 2020-10-16 at 16 25 19

Screenshot 2020-10-16 at 16 25 38

#### Related issue(s):

Closes #2613


scruffian:
This seems to be an issue with the form block itself. When it's in a post on its own it seems to stretch to 100%, and when its in the editor it also seems to stretch 100%, but then in the frontend it isn't. I wonder if we should fix this in the Jetpack block itself?

MaggieCabrera:
> This seems to be an issue with the form block itself. When it's in a post on its own it seems to stretch to 100%, and when its in the editor it also seems to stretch 100%, but then in the frontend it isn't. I wonder if we should fix this in the Jetpack block itself?

The editor markup is quite different and there's no form tag actually. Do you know where this block lives? is it in the Gutenberg repo or is there a separate Jetpack repo?

scruffian:
It's in [Jetpack](https://github.com/Automattic/jetpack/tree/master/extensions/blocks/contact-form). There's a good guide on getting started with [developing Jetpack](https://github.com/Automattic/jetpack/blob/master/docs/development-environment.md) and one [specifically for the blocks](https://github.com/Automattic/jetpack/tree/master/extensions).

MaggieCabrera:
> It's in [Jetpack](https://github.com/Automattic/jetpack/tree/master/extensions/blocks/contact-form). There's a good guide on getting started with [developing Jetpack](https://github.com/Automattic/jetpack/blob/master/docs/development-environment.md) and one [specifically for the blocks](https://github.com/Automattic/jetpack/tree/master/extensions).

Thanks! Should we keep this fix until we get a PR for them going/merged?

scruffian:
I'm tempted so say no. The problem is once we merge this then we'll put the Jetpack fix on the back burner, whereas if we put the fix there directly we'll be more motivated to see it through. I also worry that we'll forget to come back and remove this when the Jetpack fix is merged.v:)

MaggieCabrera:
This is not needed anymore as the issue is no longer reproducible

Issue #2561, opened 2020-10-06
Spearhead: Dark mode block issues


MaggieCabrera:
## Steps to replicate

Create one of each block:

* Verse
* Code Block
* Navigation Horizontal
* Navigation Vertical

## Result

On the editor all of them have a very dark text:

<img width="875" alt="Screenshot 2020-10-06 at 21 15 44" src="https://user-images.githubusercontent.com/3593343/95249723-68c6dd80-0819-11eb-8f9c-78837a5e6c44.png">

On the frontend only the navigation blocks have dark text, Code and Verse are ok:

<img width="848" alt="Screenshot 2020-10-06 at 21 17 27" src="https://user-images.githubusercontent.com/3593343/95249753-70868200-0819-11eb-86a1-c82167bf5d69.png">

## Expected

I expected all of them to have a white/light grey on both frontend and editor.

PR #2623, opened 2020-10-16
Spearhead: Fixes the code and verse blocks in the editor

scruffian:
This fixes the colors of the code and verse blocks in the editor:

Screenshot 2020-10-16 at 15 47 34

Fixes https://github.com/Automattic/themes/issues/2561


MaggieCabrera:
I thought about doing this but isn't `--global--color-foreground-light` the wrong color for Seedlet?. It's the right one for Spearhead, but Seedlet's body text is actually `--global--color-foreground`. They are extremely similar for Seedlet but it may not be the case for other themes.

scruffian:
Ok, so I'll set it to `--global--color-foreground` in Seedlet and the override it in Spearhead if we need to....

MaggieCabrera:
> Ok, so I'll set it to `--global--color-foreground` in Seedlet and the override it in Spearhead if we need to....

On Spearhead dark mode there is a big difference between them:

```
--global--color-foreground-light: #ffffff;
--global--color-background: #1e1f21;
```

I do agree that defining them in Seedlet too is the best approach for future child themes

scruffian:
I think the root of the problem here is that `foreground-light` is being abused in Spearhead. I think foreground should actually be `ffffff` and foreground-light should be `b2b2b2`, which of course doesn't make sense! I think we need to rename these variables...

I updated this PR to deal with the problem in Seedlet. I'll make a followup one to rename the variable which is going to be a bit more involved!

MaggieCabrera (approved):

PR #2624, opened 2020-10-16
Seedlet: add a script to package for dotorg

jffng:
This PR adds a shell script that creates a zip processes and packages the theme so that it is ready for submission to the dotorg themes directory. 

To test
- `cd seedlet` and `./package-dotorg.sh`
- Verify that it generates a zip of the theme that does not include wpcom-specific files and features

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,4 @@ theme-dev-utils/
 theme-dev-utils
 vendor/
 *.DS_Store
-
+seedlet.zip
--- a/seedlet/package-dotorg.sh
+++ b/seedlet/package-dotorg.sh
@@ -3,7 +3,7 @@ find assets/sass/*.scss -type f -exec sed -i '' 's/-wpcom//g' {} \;
 find assets/sass/*.scss -type f -exec sed -i '' 's/auto-loading-homepage, //g' {} \; 
 npm run build;
 mkdir seedlet;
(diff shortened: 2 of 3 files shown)


seedlet/package-dotorg.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/zsh
+find assets/sass/*.scss -type f -exec sed -i '' 's/-wpcom//g' {} \; 
+find assets/sass/*.scss -type f -exec sed -i '' 's/auto-loading-homepage, //g' {} \; 
+npm run build;
+mkdir seedlet;
+rsync -avz --exclude-from 'dotorg-exclude.txt' ./ seedlet
+find seedlet -type f -name '*.map' -delete # for some reason rsync won't exclude map files
+zip -r -X seedlet.zip seedlet
+rm -rf seedlet
```
jffng, line 9: It might also be worth adding this line at the end: 
```suggestion
rm -rf seedlet
git restore ./
```

.gitignore
```diff
@@ -4,4 +4,4 @@ theme-dev-utils/
 theme-dev-utils
 vendor/
 *.DS_Store
-
+seedlet.zip
```
scruffian, line 7: Maybe this should just be *.zip

scruffian:
This looks good. It would be great to extend it to other themes in time 🚢

PR #2625, opened 2020-10-16
Add stylelint

scruffian:
This adds a stylelint config and also adds stylelint to the pre commit hook so we catch and fix issues in our Sass files.

If you don't install the hook nothing will happen, so this is just a bonus for people who want to catch and fix these issues on commit.

To test, try committing a change to sass file...


MaggieCabrera:
This LGTM, it will be very helpful. Maybe we should add a mention to this on the readme file too?

jffng (commented):
This works in my testing.

Where did the rules come from? What do you think about using the [WordPress coding standards stylelint config](https://github.com/WordPress-Coding-Standards/stylelint-config-wordpress)?

scruffian:
The rules came from Calypso - I didn't know we had one for WordPress - we should definitely use that...

scruffian:
I think we should do this in the same way that GB does it...

PR #2627, opened 2020-10-19
Seedlet: Update theme to work with the new marketing bar

MaggieCabrera:
#### Changes proposed in this Pull Request:

Updated styles on Seedlet to fit with the new marketing bar.

#### How to test:

* Update your sandbox with the contents of this PR.
* Modify the following file on your sandbox:
`https://opengrok.a8c.com/source/xref/trunk/wp-content/blog-plugins/marketing-bar.php`
* If your site is older than 2018-05-16 change this line:
`const SHOW_MARKETING_BAR_FOR_BLOGS_REGISTERED_ON_OR_AFTER = '2018-05-16';`
* If modulo of your user id is not 7, change the `!===` to `==`
`if ( is_user_logged_in() && 7 !== (int) $blog_owner % 10 ) {`

#### Screenshots:

Seedlet, with wp admin bar:

Seedlet, without wp admin bar:

#### Related issue(s):

#2629

PR #2635, opened 2020-10-19
Update the way we position the nav in seedlet

scruffian:
This is an alternative approach to https://github.com/Automattic/themes/pull/2627 which aims to simplify the current way we handle the admin bar and also allow support for other modifications to the header at the same time.


MaggieCabrera:
I really like this solution and it works fine with just the marketing bar, but not when we have both the marketing + admin bar:

scruffian:
Oh yeah you're right I forgot that the close button was position fixed. This is not going to work then...

Issue #2629, opened 2020-10-19
Multiple themes: Update themes to work with the new marketing bar


MaggieCabrera:
This issue is here to centralize the work needed on multiple themes to fit the new marketing bar. This is a work in progress, the list of themes affected:

- [x] ‘seedlet’ -> https://github.com/Automattic/themes/pull/2627 **DEPLOYED**
- [x] ‘twentyfourteen’ -> D51501-code **DEPLOYED**
- [x] ‘twentyseventeen’ - D51630-code **DEPLOYED**
- [x] ‘[twentytwenty](p7DVsv-9gv-p2#comment-31049)’ - D51555-code **DEPLOYED**
- [x] ‘boardwalk’ -> D51394-code **DEPLOYED**
- [x] ‘gateway’ - D51568-code **DEPLOYED**
- [x] ‘dyad-2’,  -> #2637 // D51402-code **DEPLOYED**
- [x] ‘cubic’ -> D51466-code **DEPLOYED**
- [x] ‘coherent’ -> D51471-code **DEPLOYED**
- [ ] ‘rockfield’ - I don't see issues with this theme
- [ ] ‘stratford’ - I don't see issues with this theme
- [x] ‘pique’ - #2641 **DEPLOYED**
- [x] ‘[scrawl](https://href.li/?https://github.com/Automattic/themes/issues/2476)’ -> D51502-code **DEPLOYED**

#### How to test:

* Update your sandbox with the contents of the specific PR or diff for the theme you want to test
* Modify the following file on your sandbox:
`https://opengrok.a8c.com/source/xref/trunk/wp-content/blog-plugins/marketing-bar.php`
* If your site is older than 2018-05-16 change this line:
`const SHOW_MARKETING_BAR_FOR_BLOGS_REGISTERED_ON_OR_AFTER = '2018-05-16';`
* If modulo of your user id is not 7, change the `!===` to `==`
`if ( is_user_logged_in() && 7 !== (int) $blog_owner % 10 ) {`

#### Screenshots:

Seedlet, with wp admin bar:

![](https://d.pr/i/yKzTY2+)

Seedlet, without wp admin bar:

![](https://d.pr/i/0o4xTD+)

MaggieCabrera:
Boardwalk:

Diff: D51394-code

Boardwalk with wp admin bar:

![](https://d.pr/i/x1K7PC+)

Boardwalk without wp admin bar:

![](https://d.pr/i/Jq5qX8+)

[...]

PR #2641, opened 2020-10-21
Pique: fixes for sticky navigation + marketing bar

MaggieCabrera:

Fixed the Sticky navigation to work with the marketing bar.

Check the home page with posts and with a page, the navigation bar should work on both.

#### Related issue(s):

#2629


scruffian:
This solves the issue, although the theme seems to be behaving in an unexpected way - I think the menu should be sticky when the breakpoints are between 600px and 782px, which it isn't, but that's an unrelated issue...

scruffian (approved):

Issue #2257, opened 2020-07-14
Varia: Premium Content block content doesn't have correct margins


edequalsawesome:
## Steps to replicate

1. Activate Varia, or any Varia-based theme
2. Add a Premium Content block to a post or page
3. Add content to the Premium Content block
4. Publish the page or post

## Result

The spacing for the elements in the Premium Content block is incorrect:

[![Premium and Regular Content](https://d.pr/i/8WSpWL+)](https://d.pr/i/8WSpWL)

This happens because the Premium Content block doesn't have an `inner-container` class, so the margins from this CSS don't apply:

```
.site-footer > *, 
.site-main > article > *, 
.site-main > .not-found > *,
.entry-content > *, 
[class*="inner-container"] > *, 
.widget-area > * {
	margin-top: 21.312px;
	margin-bottom: 21.312px;
}

@media screen and (min-width: 560px) {
	.site-footer > *, 
	.site-main > article > *, 
	.site-main > .not-found > *,
	.entry-content > *, 
	[class*="inner-container"] > *, 
	.widget-area > * {
		margin-top: 32px;
		margin-bottom: 32px;
	}
}

```

## Expected

The content in the Premium Content block would be spaced like other elements in the post or page content:

[![With CSS Applied](https://d.pr/i/OeYa6c+)](https://d.pr/i/OeYa6c)

This CSS will resolve the issue:

```
/* fix content margins in premium content block */
.entry-content .wp-block-premium-content-container [class*="wp-block-premium-content-"] > * {
	margin-top: 21.312px;
	margin-bottom: 21.312px;
}

/* unset button margins in premium content block */
.entry-content .wp-block-premium-content-container [class*="wp-block-premium-content-"] > .wp-block-buttons > * {
	margin-top: 0;
	margin-bottom: unset;
}

/* fix content margins in premium content block for larger screens */
@media screen and (min-width: 560px) {
	/* fix content margins in premium content block */
	.entry-content .wp-block-premium-content-container [class*="wp-block-premium-content-"] > * {
		margin-top: 32px;
		margin-bottom: 32px;
	}
}
```

Originally reported here: #21827539-hc

iamtakashi:
+1. This is not a blocker for patterns directly, but they will surface the issue with it. cc @ianstewart

ianstewart:
> This happens because the Premium Content block doesn't have an inner-container class, so the margins from this CSS don't apply

[...]

PR #2654, opened 2020-10-21
Premium Content : Ensuring correct vertical margin

pbking:
The paragraph elements inside of Premium Content blocks weren't
included in the entry-content margin definitions.  This ensures
that any paragraph (including those in Premium Content blocks)
have the appropriate styles assigned.

Add new selector for entry-content paragraphs when defining margins.

#### Steps to Reproduce:
Follow this guide to allow for usage of Premium Content block: https://dotcom.wordpress.com/2020/04/15/call4testpremium-content-block/

Ensure a premium content block has been added to a page/post with multiple paragraph blocks.

View the page with premium content (not the editor) and note that the margins between paragraphs does not match the margins between paragraphs outside of the content block.

image

#### Verify Fix:

Note that once the fix is in place the vertical margins of content in the Premium Content block match the rest of the page.

image

#### Related issue(s):
#2257

### DIFF
--- a/coutoire/package.json
+++ b/coutoire/package.json
@@ -1,6 +1,6 @@
 {
   "name": "coutoire",
-  "version": "1.3.6",
+  "version": "1.3.5",
   "description": "Coutoire",
   "bugs": {
     "url": "https://github.com/Automattic/themes/issues"
--- a/coutoire/package.json
+++ b/coutoire/package.json
@@ -1,6 +1,6 @@
 {
   "name": "coutoire",
-  "version": "1.3.5",
+  "version": "1.3.6",
   "description": "Coutoire",
   "bugs": {
     "url": "https://github.com/Automattic/themes/issues"
--- a/coutoire/package.json
+++ b/coutoire/package.json
@@ -1,6 +1,6 @@
 {
   "name": "coutoire",
-  "version": "1.3.5",
+  "version": "1.3.7",
   "description": "Coutoire",
   "bugs": {
     "url": "https://github.com/Automattic/themes/issues"
--- a/coutoire/package.json
+++ b/coutoire/package.json
@@ -1,6 +1,6 @@
 {
   "name": "coutoire",
-  "version": "1.3.7",
+  "version": "1.3.5",
   "description": "Coutoire",
   "bugs": {
     "url": "https://github.com/Automattic/themes/issues"
--- a/varia/sass/layout/_structure.scss
+++ b/varia/sass/layout/_structure.scss
@@ -87,7 +87,7 @@
 .site-main > article > *, // apply vertical margins to article level
 .site-main > .not-found > *, // apply vertical margins to article level
 .entry-content > *,
-.entry-content p, // ensure all paragraphs, including nested ones, carry a margin
+.entry-content-child,
 [class*="inner-container"] > *,
 .widget-area > * {
 
--- a/varia/sass/layout/_structure.scss
+++ b/varia/sass/layout/_structure.scss
@@ -87,6 +87,7 @@
 .site-main > article > *, // apply vertical margins to article level
 .site-main > .not-found > *, // apply vertical margins to article level
 .entry-content > *,
+.entry-content p, //ensure proper margins in Premium Content block
 [class*="inner-container"] > *,
 .widget-area > * {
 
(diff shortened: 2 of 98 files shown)


varia/sass/layout/_structure.scss
```diff
@@ -87,6 +87,7 @@
 .site-main > article > *, // apply vertical margins to article level
 .site-main > .not-found > *, // apply vertical margins to article level
 .entry-content > *,
+.entry-content p, //ensure proper margins in Premium Content block
```
MaggieCabrera, line 90: Do you think we could make this change more specific to the Premium content in particular? since the theme seems to be wanting to target only the direct siblings of `.entry-content` maybe our fix could live in the `sass/blocks/`

scruffian: I think this approach is good. It is quite heavy handed, but I do think that all paragraph blocks should behave like this even if they are in other blocks. I think a second opinion from @kjellr would be useful though.

jffng: I think the intuition to want to target the premium content block is good, but in this case I agree with @scruffian that all paragraph blocks would benefit from applying the margins this way. 

Also worth noting that doing so means that any paragraph which is the first/last child child of its container will have a `margin-top/bottom: 0;` applied.

https://github.com/Automattic/themes/blob/6a6fbeb8111e08a8f2a4c89ca8daf81d1e6c8b13/varia/sass/layout/_structure.scss#L102

The comment should probably be updated though because it's not specific to the Premium Content block.
```suggestion
.entry-content p, // ensure all paragraphs, including nested ones, carry a margin
```

coutoire/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "coutoire",
-  "version": "1.3.5",
+  "version": "1.3.6",
```
scruffian, line 3: Somehow this is out of sync with the other version numbers. I guess we just fix it manually...

MaggieCabrera:
I've found that this is not only affecting paragraphs but also headings. It may be affecting other blocks too. 

Screenshot 2020-10-22 at 09 43 01

pbking:
Yes @MaggieCabrera I see that there could be many elements effected by this; not just the <p>.

Looking closer it actually looks to me that the Premium Content block should have the `.*inner-container` class applied to it for the existing CSS to properly cover it (my preferred solution as it's more generic and seems more correct) OR something like this would need to be used (instead of my proposed change)
```
[class*="premium-content-subscriber-view"] > *, 
```

I am hesitant to add that change to _structure.scss (as it's a block-specific change) however the alternative is to duplicate the margin definitions in the block-specific area which I am also not keen on doing.

Thoughts on how to proceed?

scruffian:
I wonder if we should fix this in the block itself by adding an inner-container class to the block.

MaggieCabrera:
> I wonder if we should fix this in the block itself by adding an inner-container class to the block.

Jason commented this too, I wonder if this is the case for more than one block, since he pointed out that a workaround for a similar issue was already in place for other blocks such as the Group block:

https://github.com/Automattic/themes/blob/5e3b8754ddcd81c3bd5c30e93329661bcbb28e69/varia/sass/blocks/group/_style.scss#L7

pbking:
@scruffian That was exactly the fix that I wanted to make when I sat down this morning.  However after poking around I saw the same block of margin code applied to the [group block ](https://github.com/Automattic/themes/blob/trunk/varia/sass/blocks/group/_style.scss#L7) as well as the columns block.  So while I'm a big fan of that Premium Content block having the inner-container class that isn't the only instance of the situation.

What I'm proposing instead (currently committed) is to just [define that margins logic in a class](https://github.com/Automattic/themes/pull/2654/files#diff-77681300429009e75afc346c67263e608a508a7229bbccd102793935cbcb62bfR90) and [extend that](https://github.com/Automattic/themes/pull/2654/files#diff-1e32b2c294dce648323414cd283ea037cf1c8eb7bca2bd17253764373ee28dc1R6) where appropriate to avoid code duplication.

I would also argue that (at least) the group and columns block should follow the same pattern.

scruffian (approved):
This reveals some of the shortcomings of Varia. I think this is the best solution given those shortcomings :)

MaggieCabrera (approved):
I agree with @scruffian and we should look into how Seedlet is mimicking this behavior and improve it there too

pbking:
Cool.  I'll reserve the group/column changes for when either of those assets need to be touched and move forward with this fix.  Thank you @MaggieCabrera and @scruffian!

Issue #2466, opened 2020-09-17
Radcliffe 2: Blog Posts block grid view broken


sandymcfadden:
## Steps to replicate

- Create a page and add the Blog Posts block
- Choose grid view
- Set multiple columns 
- Use posts with longer titles.

## Result

- The titles force each post in the grid to be too wide so they appear one over the other instead of in a grid layout.

## Expected

- The posts to show in a grid instead of stacked on top of each other.

* User report: #3319559-zd

hacchism:
Another report: #23938515-hc

It seems this happens no matter if post titles are long or not.

rinazrina:
Similar report: #3427940-zen

The Blog Posts block is displayed as stacked though it's set in a grid view. Also, the featured images are showing on the editor, but missing on the preview and on the published version.

I suggested using the Latest Posts block for now as a workaround.

MaggieCabrera:
Reported this to the block repo in order to get a more global fix https://github.com/Automattic/newspack-blocks/issues/616

scruffian:
Since this is an issue in the Newspack block, I'm going to close this issue.

Issue #1926, opened 2020-04-17
Canard: Blog Posts blog shows black boxes in grid view


yamphoenix:
## Steps to replicate

1. Activate the Canard theme.
2. Publish some posts and add the featured image.
3. Create a page and add the Blog Posts block. Select Grid View for the block.
4. In the grid view, black boxes display below the features images. And the post excerpt is jumbled up.

## Result

Black boxes below featured image - 
![Black box](https://d.pr/i/3zYhiV+)

Post excerpt jumbled up-
![jumbled](https://d.pr/i/kbuTfQ+)

[Link to video screencast of the issue](https://d.pr/v/kFdpi0)

## Expected

The Blog Posts Block to appear properly.

* Site: http://shrimpchips.blog
* Follow-up ticket: #2882677-zen

goblinartificer:
I've been working with this user. The reason this happens is that Canard has this: 

```
/* Post Thumbnail */
.post-thumbnail {
	background: #000;
	display: block;
	height: 100%;
	position: relative;
	width: 100%;
```


That `height: 100%` sets the thumbnail in the Blog Posts block to 100% of the allocated height for the entire `.post-has-image` element in the Blog Posts block, causing the black box and the gnarly overlap.

A quick fix (not for this user as they're not on Premium) is `.post-thumbnail {
    height: auto;
}`

AtrumGeost:
Got another report here: 2938357-zen (The CSS fix is not an option here since they have the Personal plan)

mikeicode:
Same issue 3202552-hc they are also on Personal so can't use CSS.

JoshuaGoode:
Encountered on 21445670-hc

kspilarski:
Same case as 23068267-hc. Follow-up ticket at 3190375-zd.

jonnyruzek:
25110116-hc

MaggieCabrera:
Related to https://github.com/Automattic/themes/issues/2504

MaggieCabrera:
Reported this to the block repo in order to get a more global fix https://github.com/Automattic/newspack-blocks/issues/618

scruffian:
Since this is an issue in the Newspack block, I'm going to close this issue.

PR #2657, opened 2020-10-21
Update Radcliffe 2 block.css for blog posts block

sandymcfadden:

    Adding some CSS to fix broken layout for the Blog Posts block when used in grid view with long titles.

Related issue(s):

#2466

### DIFF
--- a/radcliffe-2/assets/css/blocks.css
+++ b/radcliffe-2/assets/css/blocks.css
@@ -924,6 +924,10 @@ ul.wp-block-gallery li {
 	text-decoration: none;
 }
 
+.wp-block-a8c-blog-posts.is-grid .entry-title {
+	width: auto;
+}
+
 @media all and (min-width: 600px) {
 	.a8c-posts-list-item__featured,
 	.a8c-posts-list-item__meta {


scruffian:
I think this fix should go in the block itself. That way it will be fixed for all themes, not just this one.

arcangelini:
> I think this fix should go in the block itself. That way it will be fixed for all themes, not just this one.

There is an update in the latest version of Newspack that fixes this however it has hit some snags and may never see the light of day.

The fix ⇢ https://github.com/Automattic/newspack-blocks/pull/592
The push to WPCOM ⇢ https://github.com/Automattic/wp-calypso/pull/45924
The fix for my PR breaking the staging ⇢ https://github.com/Automattic/wp-calypso
The roadblock ⇢ https://github.com/Automattic/wp-calypso/pull/46018

Not sure if there is any way around the roadblock, but it does not look like my *one line of code* is not going to reach the light of day any time soon haha.

scruffian:
One way round the road block could be to move the block itself into Jetpack...

MaggieCabrera:
Hey! I didn't notice this PR and was about to offer another solution from the theme's perspective but I noticed the discussion about moving the fix to the block. I'm not sure what you mean by moving the block to Jetpack @scruffian Also if we consider that the solution for https://github.com/Automattic/themes/issues/1926 should also live in the block we should handle that as well at the same time, right? How can we unblock this?

scruffian:
Since we are hoping to ship this fix in the block, we can close this.

Issue #2440, opened 2020-09-09
Apostrophe 2: Blog posts block lacks padding for full width and no sidebar


sandymcfadden:
## Steps to replicate

- Create a page with the default page template
- Ensure there are no sidebar widgets
- Add a blog posts block

## Result

You can see the blocks go to the far edge of the screen.
<img width="1577" alt="Screen Shot 2020-09-09 at 10 02 43 AM" src="https://user-images.githubusercontent.com/1326294/92602094-0dd6b080-f284-11ea-82d6-3011f9ed7a75.png">

## Expected

Other blocks have padding assigned to the left and right of 2em, but it is not set for the posts block.

* User report: #3292344-zd

PR #2658, opened 2020-10-21
Add padding for blog posts block

sandymcfadden:

Edits blocks.css to add class for the blog posts block so that it gets padding like other blocks when no sidebar is active.
Related issue(s):

Fix for issue #2440


scruffian:
Thanks for the fix. This seems a lot more complex than it appears on the surface. Some concerns I have:

1. The fix doesn't work in the editor, only in front end - we should at least add it to the editor CSS too.
2. This will only fix the issue for the file and homepage articles blocks. What if we add more blocks in the future? We don't want to keep updating this list
3. I wonder if this fix should go in the blocks themselves rather than this theme, since the same behaviour seems to apply in other themes...
4. I notice that the verse and code blocks don't let you set their alignment to full - maybe that's a better way to deal with these blocks too...

PR #2659, opened 2020-10-21
Update theme.json to follow latest spec 

scruffian:
Updates the `experimental-theme.json` file to follow the latest [spec](https://developer.wordpress.org/block-editor/developers/themes/theme-json/). 

This mostly just ensures that the variables are generated — the `add_theme_support` settings in Seedlet's functions.php still ensure that these settings show up in the Global Styles sidebar.


seedlet-blocks/experimental-theme.json
```diff
@@ -1,58 +1,103 @@
...
-				{
-					"slug": "huge",
-					"value": "32px"
-				},
-				{
-					"slug": "gigantic",
-					"value": "48px"
+		"settings": {
+			"color": {
+				"gradients": [
```
jffng, line 5: I noticed that the gradient previews are broken when we supply the gradients via the theme.json:

Screen Shot 2020-10-26 at 2 37 15 PM

Other than that, this PR looks good to me.

nosolosw: Ouch, this is a bug that needs fixing. It's a bit convoluted and can take a few days of back/forth, so I'd like to offer an alternative to unblock this PR: use raw values to define the gradient presets instead of pointing to the color preset CSS variables.

I'll ping here when the fix lands in Gutenberg.

nosolosw: Actually, I've created an [issue](https://github.com/WordPress/gutenberg/issues/26511) to better track this. You can subscribe over there!

kjellr: Thanks @nosolosw! I'll keep an eye on that. @jffng would you mind re-testing? I think this should be all set now.

jffng:
Looks good, merging for now and will follow up to replace gradients with presets when there's resolution there.

PR #2660, opened 2020-10-21
Seedlet: add a script to package for dotorg

jffng:
This PR adds a shell script that creates a zip processes and packages the theme so that it is ready for submission to the dotorg themes directory.

**To test**
- `cd seedlet` and `./package-dotorg.sh`
- Verify that it generates a zip of the theme that does not include wpcom-specific files and features


jffng:
Going to merge this since it was approved already in #2624 .

PR #2675, opened 2020-10-22
Compiled Varia Child Themes for #2257.

pbking:
Fix for Varia was previously put in place with this change: https://github.com/Automattic/themes/pull/2654

This change compiles that into all of the Varia child themes.

Extends a new .entry-content-child class (leveraging existing attribute definitions)to properly apply margins to premium content blocks.

#### Related issue(s):
#2257

Wee #2654 for more details regarding the fix.

### DIFF
--- a/alves/package-lock.json
+++ b/alves/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "alves",
-  "version": "1.4.7",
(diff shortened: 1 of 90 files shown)


scruffian (approved):
LGTM

Issue #2630, opened 2020-10-19
Seedlet: Change the way "light" colors work


scruffian:
In Seedlet we have a “foreground-light” color. “Light” usually means “a bit closer to the same color as the background”. However in dark mode, a “lighter” foreground color actually means the opposite – “a bit further away from the background color”. This is problematic when you start to define a dark mode color scheme, you either have to make a “lighter” color mean “darker”, which is confusing, or you have to change the contrast ratios. I propose we change the name to “low-contrast”.

PR #2678, opened 2020-10-23
Seedlet: Update light and dark color names

scruffian:
This is a fix for https://github.com/Automattic/themes/issues/2630.

Nothing should change functionally. This will break Spearhead though so we need to be careful how we merge it.


MaggieCabrera (approved):
LGTM

PR #2682, opened 2020-10-26
Spearhead: Update RTL config

scruffian:
This just copies the RTL config from Seedlet to Spearhead


jffng (commented):
LGTM, though I don't understand why the package-lock had a huge chunk of it removed and what that has to do with the RTL config.

Also looks like there are some unrelated marketing bar updates that weren't checked in yet.

scruffian:
I wonder if the RTL config meant that some dependencies changed...

PR #2684, opened 2020-10-26
Spearhead: Update colors

scruffian:
This updates Spearhead so that the color names match Seedlet. This should also fix this issue properly: https://github.com/Automattic/themes/pull/2623

This is the other part of: https://github.com/Automattic/themes/pull/2678

To test, make sure that the colors look correct in both normal and dark modes.


spearhead/variables.css
```diff
@@ -103,15 +103,15 @@
...
+		--global--color-foreground-low-contrast: #b2b2b2;
 		--global--color-background: #1e1f21;
 		--global--color-text-selection: #000000;
-		--primary-nav--color-background: var(--global--color-foreground-light);
+		--primary-nav--color-background: var(--global--color-foreground-low-contrast);
 		--button--padding-vertical: 24px;
 		--button--padding-horizontal: 32px;
 		--sticky-posts--color-background: var(--global--color-background);
-		--sticky-posts--color-text: var(--global--color-foreground-light);
+		--sticky-posts--color-text: var(--global--color-foreground-low-contrast);
```
jffng, line 114: Should this be `--global--color-foreground`? The dark mode text color for sticky posts used to be `#ffffff`.

Before | This PR
------ | -------
Screen Shot 2020-10-26 at 11 10 14 AM | Screen Shot 2020-10-26 at 11 10 31 AM

scruffian: Yes, but maybe we can remove that line altogether now...

scruffian: I removed these lines as the fallback color is the same.

Issue #2648, opened 2020-10-21
Spearhead: Author's "More posts" link should be smaller, include arrow. 


kjellr:
The "View more posts" link in the author bio should be visually a little more like the "Read more" links. 

- Smaller text.
- Arrow after it. 

Current:

![Screen Shot 2020-10-21 at 9 51 10 AM](https://user-images.githubusercontent.com/1202812/96729488-3245a280-1383-11eb-8cba-e9886085254f.png)

Recommended:

![Screen Shot 2020-10-21 at 9 51 49 AM-2](https://user-images.githubusercontent.com/1202812/96729752-805aa600-1383-11eb-836f-4140677b51bb.png)

scruffian:
Will be closed by https://github.com/Automattic/themes/pull/2685

scruffian:
Fixed by #2685

PR #2685, opened 2020-10-26
Seedlet: Make it possible to hide the author bio

### LINKED ISSUE #2683: Spearhead: Add support for hiding author bio.
See: https://wordpress.com/support/customizer/content-options/#author-bio

We should do this in Seedlet so all child themes get the benefit

scruffian:
This adds a section to the customizer to turn off the Author bio:

Screenshot 2020-10-26 at 13 52 14

I decided not to add the rest of the Jetpack options as it would be more work. We can add them as and when we need to.

Fixes #2683


kjellr:
Does this need some CSS paired with it or something? I see the option, but unchecking it has no effect: 

Screen Shot 2020-10-26 at 10 07 46 AM

scruffian:
Sorry, you're right there was a bunch of code missing. This should work now...

kjellr:
Thanks — It's working now, but the design has changed. Before: 

Screen Shot 2020-10-26 at 3 35 35 PM

After:

Screen Shot 2020-10-26 at 3 25 54 PM

Can we more closely replicate what we had before? (I'm not sure if having the image there is new in this PR, but that's what I was seeing on my end. 🤔)

scruffian:
I have updated this so that the styles match as closely to the old way of doing it as possible. There are some constraints as we have to use the markup provided by Jetpack, but hopefully this is good enough:
Screenshot 2020-10-27 at 10 14 17


I also adapted Spearhead and did #2648 at the same time:
Screenshot 2020-10-27 at 10 11 18

MaggieCabrera:
When I try this without the jetpack plugin on my local install I see this:

Screenshot 2020-10-27 at 11 51 53

spearhead/assets/sass/_author-bio.scss
```diff
@@ -0,0 +1,22 @@
+.author-link {
+	font-size: var(--global--font-size-xs);
+	position: relative;
+
+	&:after {
```
MaggieCabrera, line 5: Maybe we should make this a mixin since we are starting to reuse it?

scruffian: I'm not sure there are quite a few differences. The position is bottom right not top left for example...

kjellr:
This looks great. Any chance we can add a tiny amount of top padding to the `author-name`? Something like 4px or so should be fine. Once that's in place, this should be all set to go. 

Before|After
---|---
Screen Shot 2020-10-27 at 8 46 40 AM|Screen Shot 2020-10-27 at 8 46 30 AM

Issue #2646, opened 2020-10-21
Spearhead: Buttons have a bottom border on hover


kjellr:
Buttons gain a bottom border on hover, which pushes the rest of the page content down slightly. This is likely an inherited style, and should be removed. 

Screenshot: 

![buttons_bottom-border](https://user-images.githubusercontent.com/1202812/96728591-450ba780-1382-11eb-91a6-2b03e27a0725.gif)

PR #2686, opened 2020-10-26
Spearhead: fix button border bug (#2646)

jffng:
This PR addresses #2646.


independent-publisher-2/functions.php
```diff
@@ -6,138 +6,154 @@
  */
 
 if ( ! function_exists( 'independent_publisher_2_setup' ) ) :
-/**
```
jffng, line 9: It looks like some phpcs + phpcbf changes are getting checked into this PR as well.

spearhead/assets/sass/_text.scss
```diff
@@ -2,7 +2,7 @@
     border-bottom: 1px solid transparent;
     transition: all 0.1s ease;
 
-    &:hover {
+    &:hover:not(.wp-block-button__link) {
```
jffng, line 5: This is the only real change in this PR.

scruffian (approved):
LGTM

PR #2687, opened 2020-10-26
Fix/seedlet spearhead body font weight

### LINKED ISSUE #2643: Spearhead: Default font weight should be 500.
The theme defines the default font weight as `Normal`, but it should be `500`. This will result in slightly heavier text, and better match the original Spearhead site. 

Before|After
---|---
Screen Shot 2020-10-21 at 9 27 02 AM|Screen Shot 2020-10-21 at 9 27 12 AM

MaggieCabrera:

Created a new variable on Seedlet that handles body font-weight and changed its value for Spearhead.

* Seedlet should look the same as it did before
* Spearhead's body text should look heavier (changed from 400 to 500):

Screenshot 2020-10-26 at 16 45 23

#### Related issue(s):

Closes #2643


scruffian (approved):
LGTM

Issue #2670, opened 2020-10-22
Spearhead: Gallery block not showing images on click


MaggieCabrera:
## Steps to replicate

Gallery block – when clicking to view an image, nothing appears. 

## Result

[Video](https://d.pr/v/J3PPCw/k94GxKPPMv)

jffng:
I'm unable to reproduce this locally or on theamdemo.wordpress.com. 

Can you verify that this is still occurring for you @MaggieCabrera?

MaggieCabrera:
This was reported by team Athena [here](pNEWy-dfi-p2#comment-52094), I only made issues for each item but the issue does happen on the Jetpack tiled gallery here: pcrugM-6M-p2

jffng:
Closing this as the issue is gone through no intervention of ours.

PR #2688, opened 2020-10-26
Seedlet: fix image template link spacing

jffng:
In looking into #2670, I discovered a bug in the layout of the image template previous / next links. This PR fixes it.

Before | After
------ | ------
Screen Shot 2020-10-26 at 12 03 36 PM | Screen Shot 2020-10-26 at 12 03 29 PM


scruffian (approved):
Just noting that the page numbers also use the `nav-links` class, but I don't think this change negatively impacts them.

jffng:
Oh good note. Confirmed that this does not impact the page numbers:

Screen Shot 2020-10-26 at 2 25 54 PM

Issue #2645, opened 2020-10-21
Spearhead: Submenu padding feels cramped. 


kjellr:
The current submenu padding in Spearhead feels cramped on the left/right sides. THere's more padding on the top and bottom of the menu than there is on the sides: 

![Screen Shot 2020-10-21 at 9 38 10 AM](https://user-images.githubusercontent.com/1202812/96727796-759f1180-1381-11eb-89c5-a251356739d0.png)

We should even that out by setting the left/right padding for the menu to `calc(2 * var(--primary-nav--padding))` instead of just `var(--primary-nav--padding)`:

![Screen Shot 2020-10-21 at 9 39 37 AM](https://user-images.githubusercontent.com/1202812/96727875-8cddff00-1381-11eb-8a0d-cfb6a798ed88.png)

melchoyce:
It's also weird that the "Page Title" there is wrapping onto two lines.

kjellr:
Yeah, I remember seeing that issue in Seedlet too. There's something about the width calculation that makes that happen sometimes.

PR #2681, opened 2020-10-26
fixed archive headers margins to fit design

MaggieCabrera:
#### Changes proposed in this Pull Request:

Changed the margin for article's headers to fit the original design.

Screenshot 2020-10-26 at 11 00 20

#### Related issue(s):

Closes #2653

PR #2689, opened 2020-10-26
Spearhead: adjust submenu padding

jffng:
This PR addresses #2645.  

Before | After
------ | ------
Screen Shot 2020-10-26 at 12 19 40 PM | Screen Shot 2020-10-26 at 12 18 50 PM


spearhead/assets/sass/navigation.scss
```diff
@@ -8,3 +8,16 @@ $navigation-min-break-point: 'laptop';
 $navigation-max-break-point: 'laptop-only';
 
 @import "../../../seedlet/assets/sass/components/navigation/style";
+
+.primary-navigation,
+.woo-navigation {
+	.sub-menu {
+		@include media( $navigation-min-break-point ) {
+			padding: calc(0.5 * var(--primary-nav--padding)) calc( 2 * var(--primary-nav--padding) );
```
jffng, line 16: It is unfortunate that all of these styles need to be added to change one value.

scruffian: Yeah, there was a similar thing earlier with margins: https://github.com/Automattic/themes/pull/2681/files

I wonder if we could simplify these styles in Seedlet to make them easier to override in the child themes? cc @kjellr

jffng: I timeboxed an exploration into simplifying the styles, and it deserves some deeper consideration than a quick refactor to make this change easier.

Going to merge this with a comment that we should keep use cases like this in mind when we get around to simplifying Seedlet's nav styles.

PR #2705, opened 2020-10-27
Seedlet: Remove subscription block CSS

### LINKED ISSUE #2702: Spearhead: Jetpack Subscription block input field is too large
The Jetpack subscription input field has a much too large padding: 

Screen Shot 2020-10-27 at 9 34 27 AM

This isn't a blocker for launch, but it will appear on every headstarted site's "Home" page.

scruffian:
This removes the custom CSS for the Jetpack subscriptions block. Its better to let the block handle its own styles, as these rules end up looking broken.

Fixes #2702


MaggieCabrera (approved):
I think this is the best we can do here

PR #2706, opened 2020-10-27
Add spearhead files from WPCOM

scruffian:
This just syncs back some files from WPCOM to keep the repos in sync.


jffng (commented):
LGTM, though I'm not sure that the the map file should be checked in?

PR #2707, opened 2020-10-27
Added header padding when there is no menu

### LINKED ISSUE #2697: Spearhead: Header looks broken when there is no menu
When there's no menu the header looks kind of broken:
Screenshot 2020-10-27 at 12 33 18

pbking:
There was no padding value assigned to the header when there was not a
menu available.  This change uses the same padding values as are used
WITH a menu for when there is none.

#### Testing
* Install Spearhead theme.
* Setup a menu and assign it to 'Primary Navigation'
* Note the location of the title (in the header).  (This is the expected location)
* Remove the 'Primary Navigation' menu (via the customizer)
* Note the location of the title.
* The expected location is the same location as when there is a 'Primary
Navigation' menu assigned.

Menu:
image

No Menu:
image

Mobile Menu:
image

Mobile No Menu:
image

#### Related issue(s):
Fixes #2697


spearhead/assets/sass/_header.scss
```diff
@@ -28,6 +28,11 @@ $site-branding-height: 68px;
 	display: flex;
 	justify-content: space-between;
 	max-width: var(--responsive--alignfull-width);
+	padding: var(--global--spacing-horizontal);
```
scruffian, line 31: This looks good. I think this is already declared higher in the file but wrapped in `.has-main-navigation`. I think we can remove that one, which will make this simper and dry-er.

scruffian (changes requested):
Remove duplicate code

spearhead/assets/sass/_header.scss
pbking: To do so seems that it will also require changes to Seedlet.  Otherwise the element defers to the more specific selector there which has the 3x margins defined.

Is is less dry but is the smaller change.

scruffian: Oh I see what you mean. In that case 🚢

scruffian (approved):

Issue #2712, opened 2020-10-27
Spearhead: Site logo in customizer


scruffian:
We don't have any CSS to handle a site logo in Spearhead, so it looks like this:
<img width="676" alt="Screenshot 2020-10-27 at 17 46 58" src="https://user-images.githubusercontent.com/275961/97341301-cbd9ec00-187c-11eb-848b-28b76be6833b.png">

We should probably just align this left...

cc @kjellr

scruffian:
When this issue is complete we need to add a screenshot to the theme showcase page.

kjellr:
Yeah, just align the logo left and I think that's fine for now. 👍

kjellr:
On second thought... it looks pretty bad when the logo is there (at least... when it's paired with the site title and tagline). If possible, I'd prefer we borrowed an approach from Twenty Twenty-One. In that theme, the logo moves up to the top if it's displayed along with a title and tagline. 

Is this something we can pursue for Spearhead? 

## Screenshots:

Site Title/Tagline: 

![Light Mode-1](https://user-images.githubusercontent.com/1202812/97349532-f7051100-1865-11eb-93b7-099e7f85e522.png)

![Frame 2](https://user-images.githubusercontent.com/1202812/97350021-962a0880-1866-11eb-91e2-511f9de14a38.png)

Site Logo Only: 

![Light Mode-2](https://user-images.githubusercontent.com/1202812/97349560-fec4b580-1865-11eb-842e-58edb5a7d867.png)

![Frame 3](https://user-images.githubusercontent.com/1202812/97350024-9924f900-1866-11eb-8191-d11f2550e167.png)

Site Title/Tagline _and_ Logo: 

![Light Mode-3](https://user-images.githubusercontent.com/1202812/97349579-06845a00-1866-11eb-91db-e56f71f449d7.png)

![Frame 4](https://user-images.githubusercontent.com/1202812/97350033-9b875300-1866-11eb-843e-8dd23723fccb.png)

PR #2713, opened 2020-10-27
Spearhead: refine navigation design

### LINKED ISSUE #2676: Spearhead: Design improvements to the social menu
We should add some enhancements to the navigation design: 

## On Desktop

- [Ensure that the site title baseline is aligned with the menu](https://cloudup.com/cPTOG38l-vd).
- If active, place the social menu underneath the main menu. 
- If active, place the cart to the right of the main menu. 

Before: 

Screen Shot 2020-10-22 at 2 11 06 PM

After: 

recommended

## On Mobile

- Place the social menu inside of the opened menu. 

Before: 

Screen Shot 2020-10-22 at 2 15 16 PM

After: 

Closed|Opened
---|---
recommended-closed|recommended-opened

---

I'm a little torn about how to handle the Woo shopping cart on mobile. @melchoyce/@beafialho do you think one of these options is stronger? Folks like to view their cart at a glance, so I don't think it makes sense to hide it inside the Menu button on mobile. 

Option 1|Option 2
---|---|---
cart-mobile|two

jffng:
This PR attempts to address #2676 and #2712. I don't think it's successful, but I'm stuck and could use some help. 

Before | After
------ | ------
Screen Shot 2020-10-27 at 4 13 50 PM | Screen Shot 2020-10-27 at 4 13 20 PM

On desktop, we want the primary and social menus to exist in a single row with the site title + branding, [as suggested in this comp](https://user-images.githubusercontent.com/1202812/96914714-57194300-1473-11eb-9558-8a13a1f24b56.png). I believe the most straightforward way to do this is to wrap both navigations in a single container element — meaning we need to overwrite Seedlet's header template, which is what I did in this PR. I guess you could use a negative margin on the social navigation but that feels hacky and imprecise. 

On mobile, the design suggests the social menu lives _within_ the primary navigation / menu. I don't know how to do this without rewriting the existing navigation's markup, styles, and logic. I'm wondering if we can come up with an alternative that doesn't require the social menu to live inside the primary menu, or gets rid of the social menu for Spearhead altogether in favor of just adding links directly to the primary navigation.

I didn't touch the cart yet because it would add another element of complexity and the design wasn't resolved from what I could gather in the issue.


jffng:
Closing in favor of #2720

Issue #2718, opened 2020-10-28
Remove any padding attribute overrides from Spearhead Dark Mode variables.


pbking:
At some point in development some padding variables were added to the 'dark mode' media query block duplicating values that were since changed.

Rather than bringing these into sync these padding-related values should be removed from the dark mode query.

```
--button--padding-vertical: 24px;
--button--padding-horizontal: 32px;
```

scruffian:
This was closed by #2719

PR #2719, opened 2020-10-28
Remove padding attributes from Dark Mode variables

pbking:
At some point in development some padding variables were added to the 'dark mode' media query block duplicating values that were since changed.

Rather than bringing these into sync these padding-related values are being removed from the dark mode query.

#### Testing
There shouldn't be any testing necessary, or any that would be obvious; no change is expected.

#### Related issue(s):
#2718


MaggieCabrera (approved):
This looks good to me

jffng (approved):
Thanks @pbking !

PR #2720, opened 2020-10-28
Spearhead: Try a different approach for Spearhead navigation

scruffian:
This is another approach to https://github.com/Automattic/themes/pull/2713.

On desktop the menu looks like this:
Screenshot 2020-10-28 at 15 41 36

On mobile:
Screenshot 2020-10-28 at 15 42 36

I haven't done anything for the cart and the site logo yet.

To test, make sure this also works with the admin bar enabled.


spearhead/header.php
```diff
@@ -0,0 +1,121 @@
...
+							array(
+								'theme_location'  => 'primary',
+								'menu_class'      => 'menu-wrapper',
+								'container_class' => 'primary-menu-container',
+								'items_wrap'      => '<ul id="%1$s" class="%2$s">%3$s</ul>',
+							)
+						);
+						?>
+						<?php if ( has_nav_menu( 'social' ) ) : ?>
+							<nav class="social-navigation" role="navigation" aria-label="<?php esc_attr_e( 'Social Links Menu', 'seedlet' ); ?>">
```
jffng, line 53: Is it necessary or acceptable to add a `nav` element within a `nav` element? 

I tested out removing this extra nav element and adding the class `social-navigation` directly to the container generated by `wp_nav_menu`, and it seems to work about the same:

Screen Shot 2020-10-28 at 11 59 51 AM

jffng (commented):
Nice exploration, there are details to be refined, but in general I like this approach for how little additional sass is required! 

I do wonder how this will work with the Woo cart, but that's broken at the moment anyways so I'm not sure it should block us from refining the details of this PR and merging.

jffng:
We will need a solution for how to handle the logo on mobile: 

Screen Shot 2020-10-28 at 12 08 09 PM

jacklenox:
We could just remove the `top: 68px` margin on the nav container. You get a free preview of this by looking at the demo site on a small screen where the margin is negated by the "Use this theme" banner! :trollface:

Screenshot from 2020-10-28 16-21-01

spearhead/header.php
scruffian: good idea

jacklenox:
Looks like this on a vanilla install, which seems fine to me:

Peek 2020-10-28 16-28

jffng:
I made a handful of changes: 

- Reworked header template so that there's a single `nav` element present: https://github.com/Automattic/themes/pull/2720/commits/bd2747a89d1df00051bf40f31e2a628a942bb582
- Added a variable in seedlet to adjust the primary nav border color so that it can be adjusted in child themes: https://github.com/Automattic/themes/pull/2720/commits/295d97a5da13cae39a7ca9dbc2ceaade155ea464
- Cleaned up styles

I think it's ready for another test. Be sure to:
- Activate a primary and social menu 
- Try light mode vs dark mode hover styles
- Look at the mobile menu

There's still a known [visual bug with the mobile menu](https://github.com/Automattic/themes/pull/2720#issuecomment-718039171) if a logo is present.

kjellr:
This looks pretty well. 

The social menu should be moved up a bit to match the mockups currently it seems a little far away:

Screen Shot 2020-10-28 at 2 20 19 PM

---

I'm not sure what this is about, but in Chrome I'm seeing some weird hover behavior. Soon after I mouse out of the submenus, the submenus reactivate when I try to hover on the social menu items: 

hover


---

Additionally, I'm noticing a weird mobile animation bug in this PR: 

menu

(I slowed that down so you can see what's happening.) Here's how the animation looked before: 

menu

jffng:
> I'm not sure what this is about, but in Chrome I'm seeing some weird hover behavior. Soon after I mouse out of the submenus, the submenus reactivate when I try to hover on the social menu items:

This is caused by the hover transition in Seedlet's navigation styles: https://github.com/Automattic/themes/blob/trunk/seedlet/assets/sass/components/navigation/_style.scss#L298

I think we should remove it for Spearhead — it doesn't feel as appropriate for the theme's design. What do you think @kjellr?

kjellr:
> I think we should remove it for Spearhead — it doesn't feel as appropriate for the theme's design. What do you think @kjellr?

I think we should just speed it up. 😄 `0.5s` is really slow. `0.15` or `0.2` at the most should be fine.

jffng:
> 0.5s is really slow. 0.15 or 0.2 at the most should be fine

In Spearhead, or both? 

Also out of curiosity, how do you justify adding an animation here?

kjellr:
Just in Spearhead — I think it's fine to leave Seedlet alone for now. I think it's worthwhile to include a (subtle) animation here. These menus can be pretty large pieces of UI, so it's much smoother to have them transition in versus having a large portion of the screen change all of a sudden.

spearhead/assets/sass/navigation.scss
```diff
@@ -21,3 +35,25 @@ $navigation-max-break-point: 'laptop-only';
...
+	visibility: visible;
+	opacity: 1;
+	top: 50px;
+	right: 0;
+	bottom: 30px; // Not sure why this div is offset by 30px
+	left: 0;
+}
+
+.admin-bar.primary-navigation-open .extra-navigation-wrapper {
+	top: calc(50px + var(--wpadmin-bar--height));
```
jffng, line 57: It's something about this line that's triggering the animation bug. Removing it and the animation returns to the previous "slide up" effect. 

I thought it maybe had to do with the calc or use of a variable, but adding a static value here (e.g. `96px`) does not resolve the issue.

spearhead/assets/sass/navigation.scss
```diff
@@ -21,3 +35,25 @@ $navigation-max-break-point: 'laptop-only';
...
+}
+
+.primary-navigation-open .extra-navigation-wrapper {
+	background: var(--primary-nav--color-background);
+	padding: var(--global--spacing-unit) var(--global--spacing-horizontal);
+	visibility: visible;
+	opacity: 1;
+	top: 50px;
+	right: 0;
+	bottom: 30px; // Not sure why this div is offset by 30px
```
jffng, line 52: I believe this is offset by 30px because the navigation contains a translation that shifts it up by ~30px: 

https://github.com/Automattic/themes/blob/trunk/seedlet/assets/sass/components/navigation/_style.scss#L166

spearhead/assets/sass/navigation.scss
```diff
@@ -11,13 +11,54 @@ $navigation-max-break-point: 'laptop-only';
...
+$header-height: 80px;
+div.extra-navigation-wrapper {
+	@include media( $navigation-max-break-point ) {
+		background: var(--primary-nav--color-background);
+		padding: var(--global--spacing-unit) var(--global--spacing-horizontal);
+		top: 0;
+		right: 0;
+		bottom: $header-height;
+		left: 0;
+		transform: translateY( 200px ); // This can be any value > $header-height
```
jffng, line 56: I think the animation looks a little better if this is a smaller value, e.g. `90px`.

jffng:
Since we addressed @kjellr's feedback, I'm going to merge this and we can tackle remaining issues like how to handle the logo in follow ups.

PR #2723, opened 2020-10-28
Spearhead: fixed paddings after sticky post

### LINKED ISSUE #2703: Spearhead: Spacing between sticky post and first post should match
The padding between the bottom of the sticky post and the top of the first post should match: 

Screen Shot 2020-10-27 at 9 35 22 AM

MaggieCabrera:

Fixed the padding of the post after the sticky post.

The original approach was using a combination of classes and was only working on Safari (!), apparently [according to W3c](https://developer.mozilla.org/en-US/docs/Web/CSS/:not)

> The ability to list more than one selector is experimental and not yet widely supported. 

So I just changed it to target non sticky articles, which should be enough.

#### Related issue(s):

Closes #2703


kjellr (approved):
> The ability to list more than one selector is experimental and not yet widely supported.

😄 Ha, that's wild. Thanks for figuring this out!

jffng (approved):
Wow, interesting. Fixes the issue in my testing, thanks @MaggieCabrera!

PR #2729, opened 2020-10-28
Spearhead: Update text color of post excerpts

### LINKED ISSUE #2724: Spearhead: In archive pages, body post text should be --global--color-secondary
In archive pages (and the blog posts block), post text excerpts should be `--global--color-secondary`, to match spearhead.co. 

The body text for the sticky post should remain darker (the way it is now). 

Current|Recommended
---|---
Screen Shot 2020-10-28 at 12 03 04 PM|Screen Shot 2020-10-28 at 12 04 23 PM

kjellr:
Fixes #2724.

Makes the color of excerpts (and their "More" links) on the homepage + archive pages lighter to match the original comps. Does not effect sticky posts, and should have no effect when these pages are set to show full posts. 

Before|After
---|---
Screen Shot 2020-10-28 at 3 48 33 PM|Screen Shot 2020-10-28 at 3 44 53 PM

Before|After
---|---
Screen Shot 2020-10-28 at 3 48 20 PM|Screen Shot 2020-10-28 at 3 45 07 PM


jffng (approved):
LGTM!

PR #2731, opened 2020-10-29
Spearhead: Add support for site logo

### LINKED ISSUE #2712: Spearhead: Site logo in customizer

scruffian:
This updates the design for the site logo to match #2712. There are a lot of cases to test here. I was going to list them all but 
there are too many. Here are all the different things you need to consider:

- This change affects both Seedlet and Spearhead
- Different browser widths (this works differently for mobile, tablet and desktop)
- Site title can be empty
- Site description can be empty
- Title and tagline can also be enabled/disabled
- Site logo can be enabled/disabled
- Primary menu can be enabled/disabled
- Social menu can be enabled/disabled
- The menu can be open or closed

The total number of combinations is hard to calculate! Here are a few screenshots:

Screenshot 2020-10-29 at 10 31 47
Screenshot 2020-10-29 at 10 32 01
Screenshot 2020-10-29 at 10 32 24
Screenshot 2020-10-29 at 10 32 32
Screenshot 2020-10-29 at 10 32 43

Fixes #2712


spearhead/style-rtl.css
```diff
@@ -363,6 +364,16 @@ Tags: one-column, flexible-header, accessibility-ready, custom-colors, custom-me
...
 }
 
+.has-title-and-tagline .site-logo {
+	flex-basis: 100%;
+	margin-bottom: var(--global--spacing-vertical);
+	text-align: center;
+}
+
+.primary-navigation-open .has-title-and-tagline .site-logo {
+	display: none;
```
scruffian, line 374: This makes the open animation quite janky, but the alternative is to have the header be really large which looks odd...

kjellr: Seems fine for now — we can always spend some time revising the animation post-launch if we decide to.

kjellr (changes requested):
This is looking good! Everything appears as expected in Spearhead. 

I'm seeing an un-centered logo in Seedlet though: 

Desltop|Mobile
---|---
Screen Shot 2020-10-29 at 8 03 38 AM|Screen Shot 2020-10-29 at 8 03 47 AM

Once that's cleaned up I think this is a solid improvement. 👍

MaggieCabrera:
I have a big menu + Social links, is this expected? 

Screenshot 2020-10-29 at 13 10 15

kjellr:
> I have a big menu + Social links, is this expected?

Good catch — the menu should ideally wrap. Something like this: 

wrap

scruffian:
@kjellr @MaggieCabrera these issues should now be fixed!

MaggieCabrera (approved):
Looks awesome on my tests!

kjellr (approved):
Looks great, thanks @scruffian!

PR #2732, opened 2020-10-29
Spearhead: show audio block on the excerpt

### LINKED ISSUE #2578: Spearhead: Podcast Footer for post excerpts
Add the audio blocks to the excerpt footer like in the design/original site:

Screenshot 2020-10-09 at 17 51 30

Screenshot 2020-10-09 at 17 51 35

Screenshot 2020-10-09 at 17 51 51

MaggieCabrera:

This is my attempt at showing the audio block on the excerpt of a post. 

To test:

* Create a post and add an audio block
* Set the home page and/or archive pages to show only summary of posts
* The audio block should show up after the content of the post

Also note that:

* If there is no other content but the audio block, the block should show up on the excerpt
* If there is multiple blocks on the post (audio or otherwise) the only block that should show up on the excerpt is the first audio block from the post.
* If no audio block is present the excerpt should look just like before.

Screenshot 2020-10-29 at 11 32 07

#### Related issue(s):

Fixes #2578

### DIFF
--- a/spearhead/functions.php
+++ b/spearhead/functions.php
@@ -227,13 +227,26 @@ function spearhead_excerpt_more() {
 }
 
 function spearhead_the_excerpt( $excerpt ) {
+
+	$audio_block = '';
+	if ( has_block( 'audio' ) ) {
+		$post   = get_post();
+		$blocks = parse_blocks( $post->post_content );
+		foreach ( $blocks as $block ) {
+			if ( 'core/audio' === $block['blockName'] ) {
+				$audio_block .= '<div class="excerpt-audio-block">' . $block['innerHTML'] . '</div>';
+				break;
+			}
+		}
+	}
+
 	// For cases where the post excerpt is empty
 	// (but the post might have content)
 	if ( 0 === strlen( $excerpt ) ) {
-		return $excerpt;
+		return $excerpt . $audio_block;
 	}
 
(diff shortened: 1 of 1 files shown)


scruffian:
Do you think we should add this to Seedlet so it works for all child themes?

scruffian:
I wonder if the audio block should show first? It seems strange to have the More > link before the block.

MaggieCabrera:
> Do you think we should add this to Seedlet so it works for all child themes?

I think it makes sense to have it available for all of them but we are already overriding Seedlet's excerpt function to alter the read more link so we will need to keep this one here

MaggieCabrera:
> I wonder if the audio block should show first? It seems strange to have the More > link before the block.

It's what spearhead does:

Screenshot 2020-10-29 at 11 45 23

scruffian:
Code wise this looks great. I'll leave @kjellr to give some input on the design.

jeffikus:
@MaggieCabrera perhaps add a conditional inside the function, and a second argument to the function that allows the block to be displayed before or after the excerpt. Default that argument to after.

kjellr:
This is a cool solution! I like it and I think it's fine to display after the "More" link. 

My only bit of feedback is that we should consider adding a Customizer option for this — I can imagine it might take some users by surprise, and it would be great to have a way to turn it off. (We can do that as a post-launch enhancement if we need to though).

PR #2733, opened 2020-10-29
Spearhead: Update theme screenshot

kjellr:
Updates spearhead's theme screenshot after the changes in https://github.com/Automattic/themes/pull/2729.

Screen Shot 2020-10-29 at 9 54 45 AM


kjellr:
Let's hold off until #2732 is merged to .com – I think this will need a refresh then too.

kjellr:
Should be ready for a review now!

screenshot

PR #2736, opened 2020-10-29
Spearhead: adds customizer notice about dark mode.

### LINKED ISSUE #2642: Spearhead: Add dark mode note or checkbox to customizer panel
Spearhead currently uses a dark mode palette if the user has the default palette selected. This is not reflected in the Customizer unfortunately, and it can be jarring to select the default palette, but then have the site appear using totally different colors when you close the panel: 

dark-mode

Suggestions: 

- At the very least, it would be great to display a message in the Customizer panel when the user has the default palette active. Something like: "This theme's default palette will display a dark mode automatically for users who have dark mode enabled on their devices. The preview here is only showing the light mode. [Learn more]" (with a link to a new support page about the feature). 
- Alternatively, we could display a checkbox to enable/disable dark mode if they have the default palette active.

jffng:
This PR addresses the first part of #2642.

Screen Shot 2020-10-29 at 2 08 46 PM

**To Test**
- Go to the customizer
- Verify the notice appears as expected


kjellr:
Thanks, @jffng! A couple notes: 

- I think the `.pot` reject file is here erroneously, right?
- This message is only necessary if the user has the Default palette selected, right? Otherwise, their chosen colors are displayed all the time. Given that, I wonder if it makes sense to mention that in the message, and also to display it as a notification (like the contrast one above it), since the message may be appearing and disappearing?

jffng:
> I think the .pot reject file is here erroneously, right?

Yes, I just wanted to remove it but I guess it should be done separately if we're being clean about things.

> This message is only necessary if the user has the Default palette selected, right? Otherwise, their chosen colors are displayed all the time. Given that, I wonder if it makes sense to mention that in the message, and also to display it as a notification (like the contrast one above it), since the message may be appearing and disappearing?

In theory, but there's a bug that I think needs to be resolved before we can implement showing and hiding the message depending on the palette: https://github.com/Automattic/themes/issues/2737

Also, any logic that we implement should work on dotorg and wpcom.

MaggieCabrera (approved):
This looks fine to me!

jffng:
This should be ready for another test. Diff id is D52226.

I made two changes:
- Updated the design so it looks like the contrast notice
- Only shows the dark mode notice when the default palette is active

Screen Shot 2020-11-03 at 11 57 59 AM

cc @kjellr

MaggieCabrera:
LGTM!

kjellr (approved):
Cool, that's much better! Thanks.

scruffian:
Deployed in r57292-wpcom-themes

Issue #2715, opened 2020-10-28
Spearhead: Adding WooCommerce support


MaggieCabrera:
## Steps to replicate

* Activate WooCommerce plugin and add products.

### Product page

Light Mode | Dark mode
--- | ---
<img width="1630" alt="Screenshot 2020-10-28 at 09 46 10" src="https://user-images.githubusercontent.com/3593343/97412886-6f211480-1902-11eb-86ad-56433b7e81d0.png"> | <img width="1650" alt="Screenshot 2020-10-28 at 09 37 55" src="https://user-images.githubusercontent.com/3593343/97412906-7516f580-1902-11eb-8f15-fd1248f770db.png">
<img width="506" alt="Screenshot 2020-10-28 at 09 47 47" src="https://user-images.githubusercontent.com/3593343/97413174-be674500-1902-11eb-8730-1516b266aeb4.png"> | <img width="481" alt="Screenshot 2020-10-28 at 09 38 27" src="https://user-images.githubusercontent.com/3593343/97413195-c4f5bc80-1902-11eb-90e5-93826b0a3041.png">
<img width="532" alt="Screenshot 2020-10-28 at 09 48 02" src="https://user-images.githubusercontent.com/3593343/97413386-05edd100-1903-11eb-8bba-c5a48616c6d7.png"> | <img width="538" alt="Screenshot 2020-10-28 at 09 48 10" src="https://user-images.githubusercontent.com/3593343/97413397-08e8c180-1903-11eb-8e79-6474e082448d.png">
<img width="498" alt="Screenshot 2020-10-28 at 09 48 59" src="https://user-images.githubusercontent.com/3593343/97413249-d8a12300-1902-11eb-9e03-fc76a434a074.png"> | <img width="493" alt="Screenshot 2020-10-28 at 09 40 42" src="https://user-images.githubusercontent.com/3593343/97413259-db9c1380-1902-11eb-991e-8c6771a3f977.png">

### Product list

Light Mode | Dark mode
--- | ---
<img width="1646" alt="Screenshot 2020-10-28 at 09 52 52" src="https://user-images.githubusercontent.com/3593343/97413905-9cba8d80-1903-11eb-926d-5ea1aa9953f9.png"> | <img width="1655" alt="Screenshot 2020-10-28 at 09 53 00" src="https://user-images.githubusercontent.com/3593343/97413913-9e845100-1903-11eb-9f22-c42269f5cefb.png">

[...]

PR #2745, opened 2020-10-30
Spearhead: improved styles for product list page

MaggieCabrera:

On the Shop products list page:

* Changed color of prices
* Changed the color of the Sale button
* Changed font family of the Sale button

Can be tested on https://theamshopdemo.wpcomstaging.com/shop/

Light mode | Dark mode
--- | ---
Screenshot 2020-10-30 at 10 43 15 | Screenshot 2020-10-30 at 10 43 21

For the WooCommerce blocks that show a grid of products:

* Changed color and font of prices
* Changed styles and position of Sale button

Light mode | Dark mode
--- | ---
Screenshot 2020-10-30 at 11 07 51 | Screenshot 2020-10-30 at 11 07 43

Can be tested on https://theamshopdemo.wpcomstaging.com/

#### Related issue(s):

Partially addresses #2715


kjellr:
Thanks Maggie! 

A few minor notes: 

1. It looks like the "Sale" badge is not always circular now. We should ensure the width and height are equal to each other. 

2. On hover, the Sale badge becomes unreadable. 

hover

3. Can we remove all of the margin between the product title and its price? And then make sure the margin between the price and the "Add to Cart" button equals the margin between the image and the title:

Current|Recommended
---|---
Screen Shot 2020-10-30 at 9 39 24 AM|Screen Shot 2020-10-30 at 9 40 12 AM

Thank you!

MaggieCabrera:
Hey @kjellr thanks for the feedback, I'm working on it. Just two things:

* How come the sale text shows up in red for you? Are you checking the block or the list page (they have a few differences)?
* What I see is that the "Sale" badge has min-width and min-height but it can and will change shape depending on the text size. I wonder if that's something we should consider design-wise, since in other languages the word will be bigger.

spearhead/assets/sass/_woocommerce.scss
```diff
@@ -0,0 +1,24 @@
+#content .wc-block-grid .wc-block-grid__product-onsale {
+	font-family: var(--global--font-secondary);
+}
+
+body[class*="woocommerce"] #page span.onsale,
+#content .wc-block-grid .wc-block-grid__product .wc-block-grid__product-onsale {
```
Aljullu: I know this is not specific to this PR but something 'inherited' from Seedlet, but I wonder what's the rationale to add the `#content` ID selector here. In the future it might be possible to add these blocks in other areas like the footer, in that case these styles would not be applied, correct? Was that intentional?

kjellr:
> How come the sale text shows up in red for you? Are you checking the block or the list page (they have a few differences)?

I see what you mean. I saw that _somewhere_, but I can't find it anymore! 😩 Let's ignore for now, and we can always open a separate bug once we can reproduce.

> What I see is that the "Sale" badge has min-width and min-height but it can and will change shape depending on the text size. I wonder if that's something we should consider design-wise, since in other languages the word will be bigger.

Ah, good point. Let's have that not be circular then: 

Screen Shot 2020-10-30 at 11 25 57 AM

I achieved that with this override in the inspector: 

```
min-height: 0;
min-width: 0;
border-radius: 0;
line-height: 1.6;
min-height: 0;
padding: 1em;
```

Aljullu (commented):
@MaggieCabrera thanks for assisting me getting the theme up and running so I could test it. :muscle: 

I reviewed this PR since it's related to WC Blocks, but take my review with a pinch of salt since I'm not used to the codebase. Also, I didn't review the end result but only how it related to WC Blocks.

IMO all changes make sense and I don't think there is anything that should be upstreamed to WC Blocks, so it should be good to go from my side. I added a comment related to the usage of ID selectors, but it's not specific to this PR so don't consider it as blocking. :slightly_smiling_face: 

> Can we remove all of the margin between the product title and its price? And then make sure the margin between the price and the "Add to Cart" button equals the margin between the image and the title:

Not sure if that's specific to the blocks, but if it is and you think the issue comes from WC Blocks CSS, we would appreciate if you could upstream the fix, we have an open issue related to that (https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2565). Feel free to ping me if you think I can help in any way.

spearhead/assets/sass/_woocommerce.scss
MaggieCabrera: Thanks for pointing that out, it's worth looking into this to improve Seedlet for the future

scruffian:
I also added a commit to fix #3005

scruffian (approved):
LGTM

Issue #2737, opened 2020-10-29
Spearhead: wpcom custom colors not working


jffng:
On wpcom, custom colors do not update when selecting individual custom colors. Here is a video recording attempting to update the primary color: 

https://cloudup.com/cBzwUc1aqeG

Only after selecting one of the palette presets do the colors update.

PR #2756, opened 2020-11-02
Spearhead: fix custom colors

jffng:
This PR addresses #2737. 

Currently, some of our code that converts RGB to HSL causes a php warning to be output to a style tag where the CSS variable overrides are set:

Screen Shot 2020-11-02 at 1 57 38 PM

This would result in invalid style rules, thus custom colors picked via the customizer would not appear.

**To test**
- Activate Spearhead
- Change an individual color from the default palette
- Verify that the color is changed in the view, and that no warnings are generated in the style tag `id="custom-colors-css"`

### DIFF
--- a/seedlet/inc/wpcom-colors-utils.php
+++ b/seedlet/inc/wpcom-colors-utils.php
@@ -97,7 +97,7 @@ function rgb_to_hsvl( $rgb ) {
 	$min_rgb           = min( $r, $g, $b );
 	$chroma            = $max_rgb - $min_rgb;
 	$v                 = 100 * $max_rgb;
-	if ( 0 == $chroma ) {
+	if ( 0 === (int) $chroma ) {
 		return array( 0, 0, $v, $l );
 	}
 	$s = 100 * ( $chroma / $max_rgb );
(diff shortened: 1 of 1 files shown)


seedlet/inc/wpcom-colors-utils.php
```diff
@@ -97,7 +97,7 @@ function rgb_to_hsvl( $rgb ) {
 	$min_rgb           = min( $r, $g, $b );
 	$chroma            = $max_rgb - $min_rgb;
 	$v                 = 100 * $max_rgb;
-	if ( 0 === $chroma ) {
+	if ( 0 == $chroma ) {
```
jffng, line 100: Using a less strict comparison causes the warning to go away, but I had to commit with `--no-verify` to avoid the strict comparison check.

scruffian: Maybe the way round that would be to do `if ( 0 === (int) $chroma )`?

jffng:
This should be good for another test — diff created D52190.

MaggieCabrera:
This works on my test site and I see the correct colors even when I have black mode enabled

PR #2759, opened 2020-11-03
fixed height of blog post thumbnail

### LINKED ISSUE #1926: Canard: Blog Posts blog shows black boxes in grid view

MaggieCabrera:

This PR addresses an issue with post thumbnails on the Blog posts block. 
I have considered making this a PR for the actual block but I'm unsure if that is the right way to go about this since the issue seems to be caused by the theme's css. The `.post-thumbnail {height: 100%; }` rule is valid in other cases that are not this block (I haven't found a case that would break if we removed this rule but I see the `.post-thumbnail` class used a lot throughout the stylesheet), and the block itself does declare `height: auto;` for the image inside the post-thumbnail container, so I don't think the block should be responsible for what the theme is breaking (for other purposes).

Before | After 
--- | ---
Screenshot 2020-11-03 at 11 29 26 | Screenshot 2020-11-03 at 11 29 21

#### Related issue(s):

Closes #1926

### DIFF
--- a/canard/style.css
+++ b/canard/style.css
@@ -2172,6 +2172,10 @@ a.post-thumbnail:hover img {
 	transform: translateY(-50%);
 }
 
+.wpnbha .post-thumbnail {
+	height: auto;
+}
+
 /* Page Links */
 .page-links {
 	border-top: 1px solid #eee;


MaggieCabrera:
Closing as this will be better fixed on a block level (https://github.com/Automattic/newspack-blocks/issues/616)

PR #2761, opened 2020-11-03
Add heading rules to varia based themes

scruffian:
This is a different approach to https://github.com/Automattic/themes/pull/2551 which fixes all Varia based themes.

### DIFF
--- a/varia/style-editor.css
+++ b/varia/style-editor.css
@@ -727,6 +727,14 @@ object {
 	margin-bottom: 0;
 }
 
+.wp-block-group {
+	/* Treating H2 separately to account for legacy /core styles */
+}
+
+.wp-block-group h2.has-text-align-left {
+	text-align: left;
+}
+
 .wp-block-group.has-background {
 	padding: 21.312px;
 }
--- a/varia/style-editor.css
+++ b/varia/style-editor.css
@@ -727,12 +727,16 @@ object {
 	margin-bottom: 0;
 }
 
-.wp-block-group {
-	/* Treating H2 separately to account for legacy /core styles */
-}
-
 .wp-block-group h2.has-text-align-left {
 	text-align: left;
+}
+
+.wp-block-group h2.has-text-align-center {
+	text-align: center;
+}
+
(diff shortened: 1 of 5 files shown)


varia/style-editor.css
```diff
@@ -727,6 +727,14 @@ object {
 	margin-bottom: 0;
 }
 
+.wp-block-group {
+	/* Treating H2 separately to account for legacy /core styles */
+}
```
MaggieCabrera, line 732: Are we getting this empty rule because of the comment?

scruffian: seems like it. interestingly switching to a `//` comment removes it...

MaggieCabrera:
I was trying to review but I'm not sure what the correct behavior is. A heading with no defined alignment goes to the right on RTL site, the rest goes to the alignment that was selected in the editor, is that correct? If so, then this looks fine to me:

Screenshot 2020-11-03 at 18 38 17

varia/style-editor.css
allancole: Yeah, I think the Sass compiler completely removes `//` but it keeps `/* … */`. 

_I believe_ there’s a separate, later script on wpcom that minifies each theme stylesheet which removes all code comments in CSS.

scruffian:
Looking at this some more I think it's a bad idea, see https://github.com/Automattic/themes/pull/2551#issuecomment-721419680

allancole:
This is what I’m getting on my end with **RTL turned on** inside of a group block:

image

Looks like the alignment is still getting reversed in RTL, and _I think_ its supposed to stay persistent regardless of the RTL setting.

image

For other text-align utility class rules we’re using a Sass flag to ignore the RTL so the floating position isn’t automatically flipped, which breaks the expected behavior. See here:

https://github.com/Automattic/themes/blob/96c77a2a4b3e38e6a2ae1581459845525c6fc23e/varia/sass/blocks/utilities/_style.scss#L6-L10

scruffian:
I wonder if we can just remove these rules and let core handle it all?

PR #2762, opened 2020-11-03
Spearhead: add missing dotorg files + scripts

jffng:
This PR makes the following changes in preparation for submission to dotorg: 

- Adds a license to Spearhead
- Adds a readme.txt to Spearhead
- Moves and amends the list of files to exclude from dotorg theme bundles to the top level so it can be shared.


scruffian:
Do you think we could also move `package-dotorg.sh` to the top level too and pass the theme slug to it?

scruffian:
This looks great. I think we should also add some documentation to explain how to use it :)

jffng:
Added a bit to the readme about how to use this script. Going to merge this PR and submit Spearhead to dotorg along with an update to Seedlet.

PR #2763, opened 2020-11-03
Seedlet/Spearhead: Fix mobile heading word-breaks and spacing rules in Jetpack layout block.

### LINKED ISSUE #2666: Spearhead: Bigger headings on mobile break words
## Steps to replicate

Add a heading (h1) inside a cover block and check on a smaller screen

allancole:

1. Remove word-breaks in Layout Grid block.
2. Clean up small, medium, large, and huge grid spacing option settings. 
3. Some light code tidying.

This solution should be tested in both Seedlet and Spearhead to make sure it works properly in both. 

#### Related issue(s):

Fixes #2666

### DIFF
--- a/seedlet/assets/css/style-editor.css
+++ b/seedlet/assets/css/style-editor.css
@@ -121,7 +121,7 @@
 	--layout-grid--gutter-medium: var(--global--spacing-unit);
 	--layout-grid--gutter-large: calc( var(--global--spacing-unit) * 2);
 	--layout-grid--gutter-huge: calc( var(--global--spacing-unit) * 3);
-	--layout-grid--background-offset: calc( var(--global--spacing-unit));
+	--layout-grid--background-offset: var(--global--spacing-unit);
 	--list--font-family: var(--global--font-secondary);
 	--definition-term--font-family: var(--global--font-primary);
 	--pullquote--font-family: var(--global--font-primary);
--- a/seedlet/assets/sass/vendors/jetpack/blocks/layout-grid/_config.scss
+++ b/seedlet/assets/sass/vendors/jetpack/blocks/layout-grid/_config.scss
@@ -6,5 +6,5 @@ $layout-gutter-sizes: "none", "small", "medium", "large", "huge";
 	--layout-grid--gutter-medium: var(--global--spacing-unit);           // 16px
 	--layout-grid--gutter-large: calc( var(--global--spacing-unit) * 2); // 32px
 	--layout-grid--gutter-huge: calc( var(--global--spacing-unit) * 3);  // 48px
-	--layout-grid--background-offset: calc( var(--global--spacing-unit)); // 16px
+	--layout-grid--background-offset: var(--global--spacing-unit);       // 16px
 }
--- a/seedlet/assets/sass/vendors/jetpack/blocks/layout-grid/_style.scss
+++ b/seedlet/assets/sass/vendors/jetpack/blocks/layout-grid/_style.scss
@@ -1,9 +1,34 @@
 @import 'config';
 
 .wp-block-jetpack-layout-grid {
+	/* Grid gutter size options */
 	grid-gap: var(--layout-grid--gutter-large) !important;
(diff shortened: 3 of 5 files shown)


MaggieCabrera (approved):
LGTM! But definitely, these changes belong on the block

scruffian:
The reason for adding this is here: https://github.com/Automattic/block-experiments/pull/34

Are we sure that letting the text overflow the block is preferable to wrapping if it doesn't fit?

MaggieCabrera:
But the issue with the padding still persists, right? they are a bit too big as they are. I just run an iphone 5 simulator and it looks like this:

Screenshot 2020-11-04 at 10 55 23

scruffian:
> But the issue with the padding still persists, right? they are a bit too big as they are. I just run an iphone 5 simulator and it looks like this:

Right, but this is an issue with the starter content itself, not the block. I think we should fix it here: https://github.com/Automattic/wp-calypso/issues/44151

allancole:
> But the issue with the padding still persists, right? they are a bit too big as they are. 

Ahh, I forgot to mention that in the description, but yes this needs to be fixed in the starter content itself :-). 

I think this is good to go ahead and merge then, right? @scruffian

scruffian:
I don't this so. I think most of these fixes are already in the block: https://github.com/Automattic/block-experiments/blob/master/blocks/layout-grid/style.scss

If there are more fixes we need to make, they should also go in the block.

allancole:
Got it. So I agree the word-break fixes can be removed from this PR since thats fixed in the block itself. 

However, the spacing fixes on [this commit](https://github.com/Automattic/themes/pull/2763/commits/c5bb8334d4c13377bd6627128f98c947b4d2ca33) are a legitimate bug in the theme. Without it, the CSS-variables for spacing (padding/margins) fall out of sync with the Seedlet global variables used in the Layout Grid block (and many other Blocks). 

See here:
https://github.com/Automattic/themes/blob/e77e1d78d0383d8a8a2d2e459fbf4d3c4d5f4a04/seedlet/assets/sass/abstracts/_config.scss#L57-L58

and here:
https://github.com/Automattic/themes/blob/c5bb8334d4c13377bd6627128f98c947b4d2ca33/seedlet/assets/sass/vendors/jetpack/blocks/layout-grid/_config.scss#L3-L10

If we leave this unchanged, and a user sets the grid space setting to anything other than default, the actual spacing value will be determined by the Layout Block instead of by Seedlet.

scruffian:
Ah I see, thanks for explaining! What is the advantage of Seedlet defining the global spacing unit?

allancole:
So the idea behind Seedlet is that it overrides most font, color, and spacing rules that come from Gutenberg and replaces them with CSS-variables that we can then reset in Seedlet child themes. 

This gives us a quick and easy way to make global or block-specific design changes in child-themes without having to write a ton of custom CSS. That gets tedious and hard to maintain very quickly when you need to consistently do the same change for all the blocks manually. Instead, we can just reset a particular variable and the change will take effect in all the places we expect it to. 

The global spacing unit specifically would come in handy if we needed to build a theme design that was more “airy” or spacious than Seedlet is by default. We’d be able to write something like `:root { --global--spacing-unit: 30px; }` (20px is default) and know that the entire feel of Seedlet would immediately be more open. Or we could target a specific Block and say `:root { --layout-grid--gutter-medium: 12px; }` to only influence the medium gutter size in the Layout Grid block. The purpose is to make Seedlet child themes really quick and easy to produce with very little code, while the heavy lifting is done with Seedlet. 

There’s a separate question to be asked about whether its actually worth it to do overrides this way. Especially with actual Gutenberg global styles on the way. But for the purposes of Seedlet, this approach already exists for core and most other Jetpack Blocks, so I figured we should treat this one the same for the time being.

scruffian:
Ok that makes sense. I removed the word break changes so this should be good to merge.

scruffian (approved):

pbking:
Rebased to trunk to pick up svn<>git syncing changes

MaggieCabrera:
D52721-code

PR #2766, opened 2020-11-04
Seedlet: Bump version to 1.1.2

jffng:
This PR bumps the Seedlet version so we can submit an update to dotorg along with Spearhead


scruffian (approved):
LGTM

PR #2773, opened 2020-11-05
Update the query loop for Seedlet blocks.

kjellr:
Updates the Query loop for Seedlet Blocks so that each post is more in line with what Seedlet shows by default. 

- Each post title is now clickable. 
- Each post shows the post date beneath the content. 
- There is now appropriate spacing beneath each post in the feed.

Before|After
---|---
Screen Shot 2020-11-05 at 9 07 47 AM|Screen Shot 2020-11-05 at 9 06 50 AM

### DIFF
--- a/seedlet-blocks/block-templates/index.html
+++ b/seedlet-blocks/block-templates/index.html
@@ -3,11 +3,21 @@
 <!-- /wp:group -->
 
 <!-- wp:group {"align":"full","className":"site-content","tagName":"main"} -->
-<div class="wp-block-group alignfull site-content"><div class="wp-block-group__inner-container"><!-- wp:query-loop -->
-<!-- wp:post-title /-->
+<div class="wp-block-group alignfull site-content"><div class="wp-block-group__inner-container">
+<!-- wp:query {"queryId":1,"query":{"perPage":10,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":""}} -->
+<!-- wp:query-loop -->
+<!-- wp:post-title {"isLink":true} /-->
 
 <!-- wp:post-content /-->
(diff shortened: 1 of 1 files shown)


seedlet-blocks/block-templates/index.html
```diff
@@ -3,11 +3,21 @@
...
+<!-- wp:query {"queryId":1,"query":{"perPage":10,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":""}} -->
+<!-- wp:query-loop -->
+<!-- wp:post-title {"isLink":true} /-->
 
 <!-- wp:post-content /-->
-<!-- /wp:query-loop --></div></div>
+
+<!-- wp:post-date {"fontSize":"tiny"} /-->
+
+<!-- wp:spacer {"height":60} -->
```
jffng, line 15: Clever use of the spacer block inside the query block. This is how it has to be done because the query block is layout agnostic?

jffng (commented):
Looks good, just two comments which don't seem to be blockers for this PR:

Why don't alignments work within the query block in the site editor? Everything appears the default width: 
 
Screen Shot 2020-11-05 at 10 26 20 AM

seedlet-blocks/block-templates/index.html
kjellr: The query block is basically just a dynamic wrapper — so anything you add in there will repeat for each post. It doesn't leave much padding between posts by default, so this helps with that.

kjellr:
> Why don't alignments work within the query block in the site editor? Everything appears the default width:

Alignments in general are really weird in the Site Editor. I think in this case, it might be because the query block isn't full-width? I'm not entirely sure. In any case, I believe this is something to be solved in Gutenberg itself.

PR #2774, opened 2020-11-05
Seedlet Blocks: Remove extra group wrapper around template parts.

kjellr:
These group wrappers were added a long time ago, before template parts had controls like alignment and `tagName`. They shouldn't be necessary anymore. This PR should have no visual effect on anything. 

This PR also make a small addition to Seedlet's CSS: It takes our default alignment rules and applies them to direct children of template parts too. This rule _could_ live in Seedlet Blocks instead, but I figured it's easier to have that all defined in one place for now.


jffng (approved):
I can confirm this fixes an issue where a full-width block is placed inside a template part and doesn't appear full-width:

Before | After
------- | -------
Screen Shot 2020-11-05 at 10 04 47 AM | Screen Shot 2020-11-05 at 9 59 09 AM

Issue #2701, opened 2020-10-27
Spearhead: Custom colors not working


scruffian:
Custom colors in the customizer don't work in Spearhead
<img width="1440" alt="Screenshot 2020-10-27 at 13 35 10" src="https://user-images.githubusercontent.com/275961/97308866-3ed16b80-1859-11eb-9ec5-84cf008fcd63.png">

jffng:
This will only be for dotorg sites since dotcom uses its own custom color mechanism, right?

scruffian:
yes

pbking:
In Seedlet's [functions.php](https://github.com/Automattic/themes/blob/trunk/seedlet/functions.php#L449) the custom colors values are added via `wp_enqueue_script` by way of the [class-seedlet-custom-colors](https://github.com/Automattic/themes/blob/trunk/seedlet/classes/class-seedlet-custom-colors.php#L40) class.

Later, in the child theme, spearhead adds css variables.  Since these are loaded AFTER Seedlet's custom variables those take precedence.

PR #2776, opened 2020-11-05
Spearhead & Seedlet: Re-sort stylesheets for proper variable cascading. 

allancole:

- Seedlet: Enqueue an empty stylesheet to decouple parent styles and the CSS variables.
- Spearhead: Re-sort the empty stylesheet to allows parent styles/variables and child CSS-variables to cascade properly.
- Makes the custom color options work properly on the frontend and customizer.

Before|After
------------|------------
|

#### To Do: 

- [ ] Replace the Seedlet default colors with the Spearhead defaults. 

See here: 
https://github.com/Automattic/themes/blob/fb71c3713afe7cd962cb0bdb62e82c672a82d5bb/seedlet/classes/class-seedlet-custom-colors.php#L23-L30

We may need to make this array in the Seedlet custom-colors class filterable by child themes. I have no idea if that’s actually possible though. 

#### Related issue(s):

#2701

### DIFF
--- a/seedlet/classes/class-seedlet-custom-colors.php
+++ b/seedlet/classes/class-seedlet-custom-colors.php
@@ -14,14 +14,14 @@
 class Seedlet_Custom_Colors {
 
 	private $seedlet_custom_color_variables = array();
-	private $default_background             = '#FFFFFF';
-	private $default_foreground             = '#333333';
-	private $default_primary                = '#000000';
-	private $default_secondary              = '#3C8067';
-	private $default_tertiary               = '#FAFBF6';
-	private $default_border                 = '#EFEFEF';
+	private $default_background;
+	private $default_foreground;
+	private $default_primary;
+	private $default_secondary;
+	private $default_tertiary;
+	private $default_border;
 
-	function __construct( $background = $default_background, $foreground = $default_foreground, $primary = $default_primary, $secondary = $default_secondary, $tertiary = $default_tertiary, $border = $default_border ) {
+	function __construct( $background, $foreground, $primary, $secondary, $tertiary, $border ) {
 
 		/**
 		 * Define defaults
@@ -341,5 +341,15 @@ class Seedlet_Custom_Colors {
 
 // Instantiate Custom Colors if this is a parent theme
 if ( get_template_directory() === get_stylesheet_directory() ) {
-	new Seedlet_Custom_Colors;
+	/**
+	 * Set Seedlet Custom Color Defaults
+	 */
+	new Seedlet_Custom_Colors(
+		'#FFFFFF', // background
+		'#333333', // foreground
+		'#000000', // primary
+		'#3C8067', // secondary
+		'#FAFBF6', // tertiary
+		'#EFEFEF'  // border
+	);
 }
--- a/seedlet/classes/class-seedlet-custom-colors.php
+++ b/seedlet/classes/class-seedlet-custom-colors.php
@@ -263,7 +263,9 @@ class Seedlet_Custom_Colors {
 	 */
 	function seedlet_custom_color_variables() {
 		if ( 'default' !== get_theme_mod( 'custom_colors_active' ) ) {
-			wp_add_inline_style( 'seedlet-style', $this->seedlet_generate_custom_color_variables() );
+			wp_register_style( 'seedlet-variable-overrides', false );
+			wp_enqueue_style( 'seedlet-variable-overrides' );
+			wp_add_inline_style( 'seedlet-variable-overrides', $this->seedlet_generate_custom_color_variables() );
 		}
 	}
 
--- a/seedlet/classes/class-seedlet-custom-colors.php
+++ b/seedlet/classes/class-seedlet-custom-colors.php
@@ -14,19 +14,35 @@
 class Seedlet_Custom_Colors {
 
 	private $seedlet_custom_color_variables = array();
+	private $default_background             = '#FFFFFF';
+	private $default_foreground             = '#333333';
+	private $default_primary                = '#000000';
+	private $default_secondary              = '#3C8067';
+	private $default_tertiary               = '#FAFBF6';
+	private $default_border                 = '#EFEFEF';
 
-	function __construct() {
+	function __construct( $background, $foreground, $primary, $secondary, $tertiary, $border ) {
 
 		/**
-		 * Define color variables
+		 * Define defaults
+		 */
+		$this->default_background = $background;
+		$this->default_foreground = $foreground;
+		$this->default_primary    = $primary;
+		$this->default_secondary  = $secondary;
+		$this->default_tertiary   = $tertiary;
+		$this->default_border     = $border;
+
+		/**
+		 * Define color variables with defaults
 		 */
 		$this->seedlet_custom_color_variables = array(
-			array( '--global--color-background', '#FFFFFF', __( 'Background Color', 'seedlet' ) ),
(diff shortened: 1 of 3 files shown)


allancole:
I found a tricky issue with this solution thats related to the _To Do_ note above. On the frontend, any color that I don’t change in the customizer, resolve back to the defaults set in Spearhead instead of the colors actually set in the customizer in some cases. 

Customizer|Frontend
------------|------------
image|image

This is happening because on the frontend, we’re doing a check to see if the custom color setting matches the default color settings before outputting the overrides. It only adds the CSS-variables override when it _doesn’t_ match the defaults set by Seedlet (not Spearhead). 

However, since Spearhead is a child theme and the color defaults are still hinged to the parent theme [here](https://github.com/Automattic/themes/blob/fb71c3713afe7cd962cb0bdb62e82c672a82d5bb/seedlet/classes/class-seedlet-custom-colors.php#L23-L30), the conditional check is checking against Seedlet’s defaults instead of Spearhead’s which means that _some_ Spearhead defaults won’t get overridden unless a user changes all the custom color options to not match the defaults in Seedlet. Kind of confusing to grasp, I know. 

The solution is to figure out a way for child themes to redefine the color option defaults that are set in the parent color options class here: 
https://github.com/Automattic/themes/blob/fb71c3713afe7cd962cb0bdb62e82c672a82d5bb/seedlet/classes/class-seedlet-custom-colors.php#L23-L30

MaggieCabrera:
I also see that when we change colors on the customizer after having used them in the editor, the editor is not showing the correct colors:

Customizer | Editor | Frontend
--- | --- | ---
Screenshot 2020-11-06 at 09 51 46 | Screenshot 2020-11-06 at 09 51 52 | Screenshot 2020-11-06 at 09 51 56

allancole:
So I think this PR is good to go but there are a couple of things that need a deeper look before merging.

### Seedlet doesn’t support darkmode but Spearhead does.

When setting custom colors, the darkmode rules get overwritten by the custom colors which effectively disable darkmode without a warning. We’ll probably want to add a warning about this to prevent confusion (maybe in a separate PR?). 

This probably doesn’t have to happen now or anytime soon, but the darkmode solution in Spearhead should ideally live in Seedlet. This comes with two benefits:

1. Gives us an easy way to support darkmode in potentially all Seedlet child themes.
2. With a more robust solution that lives in Seedlet, we could have options for default, dark and/or light mode custom colors in the customizer.

### Some Spearhead snippets that would need to be added to all Seedlet child-themes

**Connect editor palette colors to child theme’s CSS-variables / theme options:**

https://github.com/Automattic/themes/blob/ba85f09d19dbfa0226407ea1db17e8b9d86833b8/spearhead/functions.php#L78-L116

**Pass the child-theme’s color variables to the Seedlet custom colors class:**

https://github.com/Automattic/themes/blob/ba85f09d19dbfa0226407ea1db17e8b9d86833b8/spearhead/functions.php#L121-L131

### Seedlet’s inline style get reorganized.

**Re-organize seedlet CSS-variable files to respect cascading properly**

https://github.com/Automattic/themes/blob/ba85f09d19dbfa0226407ea1db17e8b9d86833b8/spearhead/functions.php#L150

https://github.com/Automattic/themes/blob/ba85f09d19dbfa0226407ea1db17e8b9d86833b8/spearhead/functions.php#L159-L160

Maybe we can do this in Seedlet itself? 

Also, the inline color variable overrides are no longer attached to the main Seedlet style sheet. Instead an empty stylesheet is registered and the inline color variable overrides get attached to that, which make it easier to resort without causing regressions in the variable cascading.

https://github.com/Automattic/themes/blob/ba85f09d19dbfa0226407ea1db17e8b9d86833b8/seedlet/classes/class-seedlet-custom-colors.php#L277-L286

seedlet/classes/class-seedlet-custom-colors.php
```diff
@@ -263,7 +279,9 @@ function seedlet_generate_custom_color_variables( $context = null ) {
 	 */
 	function seedlet_custom_color_variables() {
```
jffng: Unrelated to this PR, but I think this function should be renamed to be more specific to what it does, since there is already a private variable of this class called `seedlet_custom_color_variables`. 

```suggestion
	function seedlet_enqueue_custom_color_variables() {
```

seedlet/classes/class-seedlet-custom-colors.php
```diff
@@ -263,7 +279,9 @@ function seedlet_generate_custom_color_variables( $context = null ) {
...
 		if ( 'default' !== get_theme_mod( 'custom_colors_active' ) ) {
-			wp_add_inline_style( 'seedlet-style', $this->seedlet_generate_custom_color_variables() );
+			wp_register_style( 'seedlet-variable-overrides', false );
```
jffng, line 282: Do we need to do the same re-ordering for enqueuing the editor custom colors?

seedlet/classes/class-seedlet-custom-colors.php
```diff
@@ -14,19 +14,35 @@
...
 	private $seedlet_custom_color_variables = array();
+	private $default_background             = '#FFFFFF';
+	private $default_foreground             = '#333333';
+	private $default_primary                = '#000000';
+	private $default_secondary              = '#3C8067';
+	private $default_tertiary               = '#FAFBF6';
+	private $default_border                 = '#EFEFEF';
 
-	function __construct() {
+	function __construct( $background, $foreground, $primary, $secondary, $tertiary, $border ) {
```
jffng, line 24: I think this needs to be:

```suggestion
	function __construct( $background = $default_background, $foreground = $default_foreground, $primary = $default_primary, $secondary = $default_secondary, $tertiary = $default_tertiary, $border = $default_border ) {
```

jffng (commented):
Thanks @allancole for a thorough explanation!

Re: moving dark mode to Seedlet, agree that it should be a different exploration.

> Some Spearhead snippets that would need to be added to all Seedlet child-themes

I'm comfortable with the snippets that need to be added to each child theme; they seem necessary at least until global styles replaces this mechanism. 
 
> Seedlet’s inline style get reorganized.

I also can't find a quick way around this, but I still am a bit confused as to why the stylesheets aren't already appended in the correct order.

seedlet/classes/class-seedlet-custom-colors.php
allancole: That’s a good question! It’s seems like we _don’t_ because editor colors are already working as expected on this PR, but still worth a look anyway.

seedlet/classes/class-seedlet-custom-colors.php
allancole: Agreed. 

I kind of lean towards just doing it in this PR unless there’s a good reason not to.

allancole:
> I'm comfortable with the snippets that need to be added to each child theme; they seem necessary at least until global styles replaces this mechanism.

Yeah, thats probably the better way forward, all things considered.

seedlet/classes/class-seedlet-custom-colors.php
allancole: Yeah, so it looks like [this file](https://github.com/Automattic/themes/blob/fix/spearhead-custom-colors/seedlet/assets/css/custom-color-overrides.css) is completely useless, lol. We can go ahead and borrow the same enqueuing technique for the editor. Good catch!

scruffian:
The issue seems to be fixed by https://github.com/Automattic/themes/pull/3026. Please reopen if you think there's more we need to do.

PR #2779, opened 2020-11-05
Spearhead: fix dotorg submission issues

### LINKED ISSUE #2767: Spearhead: dotorg submission errors
Spearhead has the following issues that need to be addressed. 

These were reported by the Theme Review Team automated theme check.  

- [ ] REQUIRED: Tested up to: is missing from your style.css header. Also, this should be numbers only, so <em>5.0</em> and not <em>WP 5.0</em>
- [ ] REQUIRED: Requires PHP: is missing from your style.css header.
- [ ] REQUIRED: Found wrong tag, remove jetpack-global-styles from your style.css header.~~
- ~~[ ] REQUIRED: The theme must not use the <title> tags. Found the tag in header.php.~~
- ~~[ ] REQUIRED: The theme must not call to wp_title(). Found wp_title() in header.php.~~
- ~~[ ] REQUIRED: The theme must not use the <title> tags. Found the tag in block-patterns.php.~~
- ~~[ ] REQUIRED: The theme must not call to wp_title(). Found wp_title() in block-patterns.php.~~
- ~~[ ] REQUIRED: The theme must not use the <title> tags. Found the tag in functions.php.~~
- ~~[ ] REQUIRED: The theme must not call to wp_title(). Found wp_title() in functions.php.~~
- ~~[ ] REQUIRED: The theme must not use the <title> tags. Found the tag in content-excerpt.php.~~
- ~~[ ] REQUIRED: The theme must not call to wp_title(). Found wp_title() in content-excerpt.php.~~
- ~~[ ] REQUIRED: The theme must not use the <title> tags. Found the tag in search.php.~~
- ~~[ ] REQUIRED: The theme must not call to wp_title(). Found wp_title() in search.php.~~

A full list can be found here in [this trac ticket](https://themes.trac.wordpress.org/ticket/91317), but from what I can tell most of them don't take into account the fact that this is a child theme.

jffng:
This PR addresses #2767 and adds a minor version bump to accommodate the re-submission to dotorg.


package-dotorg.sh
```diff
@@ -2,6 +2,7 @@
 if [[ "$1" != "" ]]; then
     THEME="$1"
 	find $THEME/assets/sass/*.scss -type f -exec sed -i '' 's/-wpcom//g' {} \; 
+	find $THEME/assets/sass/*.scss -type f -exec sed -i '' 's/, auto-loading-homepage, jetpack-global-styles//g' {} \; 
 	find $THEME/assets/sass/*.scss -type f -exec sed -i '' 's/auto-loading-homepage, //g' {} \;
```
scruffian: do we still need this line?

jffng: We do but there is probably a more elegant solution.. I will think about it.

scruffian: There should be a way to do this with a regex like this:
`find $THEME/assets/sass/*.scss -type f -exec sed -i '' 's/jetpack-global-styles(,)? //g' {} \;`
sed doesn't seem to like the `?` though...

jffng: I tried using the `-E` flag for sed to support extended regex, but couldn't figure it out.

For now I went with a solution that covers cases where the tags are not first.

MaggieCabrera (changes requested):
> REQUIRED: Found wrong tag, remove jetpack-global-styles from your style.css header

~~I think we need to remove this tag from here?~~

Nevermind, I see how you did it already :S

https://github.com/Automattic/themes/blob/da967ec3496964fbe5145e5833bf061941c2f4ac/spearhead/style.css#L13

MaggieCabrera (approved):
I checked the generated zip and it looks good to me

PR #2783, opened 2020-11-06
Add spearhead workflow

scruffian:
See p9ue0y-iF-p2

### DIFF
--- a/.github/workflows/blank.yml
+++ b/.github/workflows/blank.yml
@@ -1,0 +1,31 @@
+name: Spearhead Theme Check
+
+on:
+  pull_request:
+    paths:
+    - 'spearhead/**'
+    
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+        - uses: actions/checkout@v2
+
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
I made a PR based on this branch and the workflow seems to work but I don't think we want all of this to happen on PR, maybe when it's merged?

https://github.com/Automattic/themes/pull/2785/checks

.github/workflows/blank.yml
```diff
@@ -0,0 +1,31 @@
...
+          id: theme-check
+          with:
+            accessible-ready: true
+            root-folder: spearhead 
+        
+        - uses: actions/upload-artifact@v2
+          if: ${{ always() }}
+          with:
+            name: Screenshots
+            path: ${{ steps.theme-check.outputs.screenshots }}
```
StevenDufresne, line 31: Worth noting that the action default saves screenshots and storage does accumulate even though this project most likely will not come close to billable figures. You could consider explicitly adding the `retention-days` property.

Ref: https://docs.github.com/en/free-pro-team@latest/actions/guides/storing-workflow-data-as-artifacts#configuring-a-custom-artifact-retention-period

PR #2784, opened 2020-11-06
Illustratr: Remove margins on left and right alignments

### LINKED ISSUE #2463: Illustratr: Social Icon block overshoots column block when right aligned
## Steps to replicate

* Create a test site and install Illustratr theme
* Create a page, add a columns block with 50-50 proportion
* In the left column, add a social icons block
* Add a couple of social media handles in the block and right-align the whole block . 
* Add a image to the right column. It should look like this in the editor

image

## Result

When you browse to the front-end view of the page, you do not see the social icons. They are actually hidden behind the image on the right side. Screenshot:
image

On screens narrower than 1200px, we see the social icons piled vertically:
image

## Expected

Expect to see the icons right-aligned within the same left column like this:
image

## Cause and Workaround

The issue seems to be happening due to this CSS:

```
@media screen and (min-width: 1220px) {
.alignright {
    margin-right: -130px;
}}
```

There is also a `display: inline;` property of generic alignright selector that makes the icons go vertical. 

The following specific CSS workaround seems to work for now:

```
/* Correct right align for social links block within column block */

.wp-block-social-links.alignright {
    margin-right:  unset;
    display: flex;
}
```

* User report: #15448584-hc
* Follow-up ticket: None since user was happy center-aligning the block for the moment.

scruffian:
This removes the negative margins for alignments. These cause problems when inside column blocks - see https://github.com/Automattic/themes/issues/2463

- Remove negative margins on left and right alignments

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/2463

### DIFF
--- a/illustratr/style.css
+++ b/illustratr/style.css
@@ -4,7 +4,7 @@ Theme URI: https://wordpress.com/themes/illustratr/
 Author: Automattic
 Author URI: https://wordpress.com/themes/
 Description: A minimalist portfolio theme.
-Version: 1.3.8-wpcom
+Version: 1.3.9-wpcom
 License: GNU General Public License
 License URI: license.txt
 Text Domain: illustratr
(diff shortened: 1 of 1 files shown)


MaggieCabrera (approved):
LGTM!

PR #2798, opened 2020-11-11
Morden and Varia: fixed conditional class not working corectly

MaggieCabrera:
#### Changes proposed in this Pull Request:

Fix on #2782 where the condition was never being met for the headers with tagline + site title. This affects Morden and Varia.

### Morden:

Before | After
--- | ---
Screenshot 2020-11-11 at 09 28 14 | Screenshot 2020-11-11 at 09 54 23

#### Related issue(s):

#1467

PR #2799, opened 2020-11-11
Morden, Varia: changed default value for get_theme_mod

MaggieCabrera:
Changed default value for `get_theme_mod` from true to 1 since 1 is what we are expecting as a truthy value from this function. Re-fix on #2798


scruffian (approved):
LGTM. Thanks for fixing

MaggieCabrera:
D52598-code

PR #2821, opened 2020-11-16
Monticello: block-based boilerplate

jffng:
WIP


monticello/experimental-theme.json
```diff
@@ -52,5 +56,12 @@
 			"typography": { 
 			}
 		}
+	},
+	"core/site-title": {
+		"styles": {
+			"typography": {
+				"fontFamily": "var(--wp--custom--font-family--primary)"
```
jffng: This isn't being applied as of 9.3, but I left it in here anyways.

jffng:
Added a block pattern in af770a3:

Screen Shot 2020-11-17 at 3 53 59 PM

Once the font-sizes are in, we can use those in the pattern as well.

PR #2833, opened 2020-11-18
Try/monticello block menu patterns

pbking:
A first pass a creating re-usable elements that can be used to build menus for Monticello block-based theme.

In block-patterns.php I registered a NEW pattern called 'Menu Item'.  
This was mostly gutenberg markup I coped out of a page editor and tweaked by hand a little bit as I worked through it.  It's essentially a GROUP at the top-level wrapping a COLUMNS container with three pieces of content (name, description, price).  The top-level GROUP has the `is-style-menu-list-item` class applied out-of-the-gate so that the block style (outlined later) is picked up as soon as it's dropped in.  The other P elements all have helpful custom classes assigned for further styling.

In block-styles.php (added and called from functions.php) the custom block styles are initialized and pulled in from block-styles.css and the `menu-list-item` block style is defined.

block-styles.css provides the custom CSS that the `menu-list-item` needs to lay out properly in this theme.

The above registers everything and allows me to drop reusable elements into my page that I can pretty easily change to suit the needs outlined in the comp.  The result, after some work, looks something like this:

image

image

I noticed a couple of limitations while leveraging these elements to build out a page:

* While it's simple to add these elements onto a PAGE it's not possible to easily add them into a CONTAINER.  When attempting to add menu items to my two-column section I was NOT offered the options of choosing one of the patterns.
image

* Using the Gutenberg editor to edit a POST I was able to leverage this Block Pattern.  However, using the Full Site Editor I had no option to choose to insert a Block Pattern.
image

### DIFF
--- a/monticello/style.css
+++ b/monticello/style.css
@@ -82,18 +82,23 @@ represented in .json
 }
 
 /* END BUTTON STYLES */
+
 /*
-.theme-menu-item {
+These are styles for the 'menu-list-item' block pattern.
+I'm not sure that these style should live here but I don't have a better place for them yet.
+*/
+.menu-list-item {
 	font-family: var(--wp--custom--font-family--secondary);
 	font-weight: 700;
+    margin: 0 0 12px;
+}
+
+.menu-list-item p,
+.menu-list-item .wp-block-columns {
 	margin: 0;
 }
 
-.theme-menu-item p {
-	margin: 0;
-}
-
-.theme-menu-item .theme-menu-item-description {
+.menu-list-item .menu-list-item-description,
+.menu-list-item .menu-list-item-price {
 	font-weight: 400;
 }
-*/
--- a/monticello/style.css
+++ b/monticello/style.css
@@ -82,3 +82,18 @@ represented in .json
 }
 
 /* END BUTTON STYLES */
+/*
+.theme-menu-item {
+	font-family: var(--wp--custom--font-family--secondary);
+	font-weight: 700;
+	margin: 0;
+}
+
+.theme-menu-item p {
+	margin: 0;
+}
+
+.theme-menu-item .theme-menu-item-description {
+	font-weight: 400;
+}
+*/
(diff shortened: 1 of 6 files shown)


pbking:
I also had a question as to the appropriateness of adding this new "group style" to a theme.
image

These tools gave me a pretty good method of styling a section of elements, and a pretty easy way (mostly) of adding that section of elements to a page such that they will be styled as expected.  But in doing so it feels like we're "muddying the waters" for "group" themes.  Meaning: This thing we're adding, it isn't something that you could/would apply to ANY 'group' to change the style to suit.  It's more about changing things SPECIFICALLY for this use-case and I'm not sure that using a 'block-style' in this instance is actually helpful.  Simply including "custom CSS for this stamp-able block of content in styles.css" seems more appropriate to me.

monticello/style.css
```diff
@@ -81,4 +81,19 @@ represented in .json
 	opacity: 1;
 }
 
-/* END BUTTON STYLES */
\ No newline at end of file
+/* END BUTTON STYLES */
+/*
```
jffng, line 86: Can we remove these commented styles?

jffng (approved):
> While it's simple to add these elements onto a PAGE it's not possible to easily add them into a CONTAINER. When attempting to add menu items to my two-column section I was NOT offered the options of choosing one of the patterns.

It's true, the pattern insert does not appear when it's in a container. You can search for it however: 

Screen Shot 2020-11-18 at 8 40 59 PM

> Using the Gutenberg editor to edit a POST I was able to leverage this Block Pattern. However, using the Full Site Editor I had no option to choose to insert a Block Pattern.

Certain blocks are allowed / disallowed depending on the context (full site editor vs page/post). Do you think we should be able to able to insert block patterns in the site editor? 

> It's more about changing things SPECIFICALLY for this use-case and I'm not sure that using a 'block-style' in this instance is actually helpful. 

I think I agree with you, it seems like an abuse of the block styles system to apply such specific menu styles to a group when it's such a generic block.

monticello/style.css
pbking: I think (based on the conversation above) I'm about to UNcomment them and pull out the block styles.

PR #2841, opened 2020-11-21
Seedlet: Remove side padding from Group block

### LINKED ISSUE #2781: Seedlet: Unexpected side paddings added to inner container on the front end
## Steps to replicate

* Add a group block with a background colour
* Add some blocks in the group block

## Result

**Front of the site:**
Screenshot 2020-11-05 at 22 28 01

See the side paddings in the group block is much larger than the vertical ones. This is what I see in the editor.

**Editor:**
Screenshot 2020-11-05 at 22 35 56

The extra padding seems to come from here.

Screenshot 2020-11-05 at 22 30 36

## Expected

Screenshot 2020-11-05 at 22 32 13

arcangelini:
Editor styles for the Group block do not appear to match live site styles. There padding appeared to be coming from

```
seedlet/assets/sass/blocks/group/_style.scss
```

But turns out it was just in: 'style.css`

Removed from style.css and rtl.css and tested with or without background of div.

#### After

[Screenshot](https://d.pr/i/xtcrB1)

#### Before

[Screenshot](https://d.pr/i/TtGW2X)
#### Related issue(s):

Fixes #2781


scruffian:
@arcangelini I found where I think those rules are coming from in the SASS. Could you take a look to confirm that this fixes the issue. Thanks!

arcangelini:
Yes, I tested both. Removing it in the `style.css` is what fixed it. The rules in the SASS were not affecting this with or without background.

I still have the test live here if you want to check: https://business-plan.blog/sample-page/

scruffian:
Did you run the build script after you changed the SASS?

arcangelini:
> Did you run the build script after you changed the SASS?

Sorry, I think I misunderstood how this works. I am updating the SASS file and committing again. Thanks for your patience @scruffian

seedlet/style-rtl.css
```diff
@@ -7,7 +7,7 @@ Description: Seedlet is a free WordPress theme. A two-column layout and classica
 Requires at least: 4.9.6
 Tested up to: 5.5
 Requires PHP: 5.6.2
-Version: 1.1.3-wpcom
+Version: 1.1.2-wpcom
```
scruffian: This should match the numbers above

scruffian (changes requested):

seedlet/style-rtl.css
arcangelini: OK gotcha, these get updated whenever a change is made to them? I updated both to `1.1.3`

seedlet/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "seedlet",
-  "version": "1.1.3",
+  "version": "1.1.4-wpcom",
```
jffng, line 3: Is the `-wpcom` necessary in the package.json?

jffng:
This looks good to me, just have one small comment above about the package.json.

seedlet/package.json
scruffian: I don't this so, I just thought maybe they should match?

jffng:
Thanks again @arcangelini! Committed in r57365.

Issue #1832, opened 2020-03-02
Varia: Can Jetpack Testimonials archive show full content instead of excerpt?


jordesign:
## Steps to replicate

This affects Testimonials on any site using Varia or its child themes:

1. Add 1 or more Testimonials (CPT provided by Jetpack)
2. View the single Testimonial as compared to the Testimonial archive page

eg:
* https://envisionnaturalhealing.com/testimonial/michelle-m/
* https://envisionnaturalhealing.com/testimonial/

## Result

No styling of text (bold italics) comes through in the archive view. All styling is stripped.
<img width="781" alt="Screen Shot 2020-03-02 at 1 26 51 pm" src="https://user-images.githubusercontent.com/1842363/75640561-9c40f880-5c89-11ea-85b2-03b1353e7841.png">

## Expected

I expected styling of the text (bold, italics) to come through in the Archive view as they do in the single/detail view.
<img width="778" alt="Screen Shot 2020-03-02 at 1 26 46 pm" src="https://user-images.githubusercontent.com/1842363/75641005-f68e8900-5c8a-11ea-97c7-fe1a65f1edcf.png">

* User report: #19040743-hc
* Slack thread: https://a8c.slack.com/archives/C03TY6J1A/p1583113512414100

jordesign:
I'm aware this is in some ways a discussion of whether the Jetpack Testimonials deserves some kind of special treatment. The archive template currently shows `the_excerpt()` when not specifically defined for a post type.

In the instance of Testimonials I can see there being value in archives outputting `the_content()` instead as they are usually short and self contained posts.

I'm also happy to work on a PR for this - but figured it needs discussion first...

nielslange:
@jordesign As discussed in Slack, this issue is yours and you can create a PR for it. 😉

PR #2843, opened 2020-11-22
Varia: use the_content() for Testimonials archives

jordesign:
Currently Varia themes user the_excerpt() for in the archives for all posts. This strips out any styling from the content.

This seems appropriate for many instances. With Jetpack-Testimonials it is *much* more likely that the full content of the post is likely to be designed to be shown in the archive view as well. 

This Pull Request adds templates for the Jetpack-Testimonials archive in order to use the_content() instead - retaining any styling.

#### Screenshots

[link to image](https://user-images.githubusercontent.com/1842363/75648363-fcdd2f00-5ca3-11ea-8a7a-971f074fdc8a.png)

#### Related issue(s):

#1832

### DIFF
--- a/varia/archive-jetpack-testimonial.php
+++ b/varia/archive-jetpack-testimonial.php
@@ -1,0 +1,54 @@
+<?php
+/**
+ * The template for displaying Jetpack Testomonial archive pages
+ *
+ * @link https://developer.wordpress.org/themes/basics/template-hierarchy/
+ *
+ * @package WordPress
+ * @subpackage Varia
+ * @since 1.0.0
+ */
+
+get_header();
+?>
+
+	<section id="primary" class="content-area">
+		<main id="main" class="site-main">
+
+		<?php if ( have_posts() ) : ?>
+
+			<header class="page-header responsive-max-width">
+				<?php
+					the_archive_title( '<h1 class="page-title">', '</h1>' );
+				?>
+			</header><!-- .page-header -->
+
+			<?php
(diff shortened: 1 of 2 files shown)


scruffian:
@jordesign I'm not sure how to test this. I've enabled Jetpack Testimonials and added some to my site, but I'm not sure how to see the archive page. Can you help? Thanks!

jordesign:
I was testing this by visiting [site_url]/testimonials/

kathrynwp:
> I was testing this by visiting [site_url]/testimonials/

I think the automatically generated testimonials archive page has no "s" at the end - so [site_url]/testimonial/ - right @jordesign? 

reference: https://wordpress.com/support/testimonials/#viewing-testimonials

jordesign:
Thanks @kathrynwp - yes that's correct. Thanks heaps.

scruffian (approved):
This looks good and works as expected. My only concern is that it duplicates the markup for the archive page, but it seems like that's the case in all our themes, so we shouldn't block this because of it.

scruffian:
r57366-wpcom-themes

jordesign:
> This looks good and works as expected. My only concern is that it duplicates the markup for the archive page, but it seems like that's the case in all our themes, so we shouldn't block this because of it.

I had been wondering a little about that TBH. I wasn't sure of the scale/point at which to break the template out as different but ended up deciding doing it in the structure/files for the sake of clarity.

PR #2852, opened 2020-11-23
Dont try to sync TT1 blocks

scruffian:
This excludes TT1 blocks so we don't try to sync it with SVN


jffng:
Will we want to sync TT1 blocks once it's added been added via https://github.com/orgs/Automattic/projects/215#card-49898611?

scruffian:
I don't think so, we don't want TT1 blocks to live in this repo...

jffng (approved):

Issue #2853, opened 2020-11-23
Seedlet: Unable to Activate


kspilarski:
## Steps to replicate

**In a simple site:** 

- activating Seedlet in Calypso shows _"Unable to activate theme. Contact support."_ 
- In WP-admin > Appearance, activating Seedlet shows _"Error: Current WordPress version does not meet minimum requirements for Seedlet."_ 
- when previewing the theme it shows "Cannot Activate" in the Customizer. 

<img width="284" alt="Screen Shot 2020-11-23 at 5 09 45 PM" src="https://user-images.githubusercontent.com/36863633/99986202-3106fb80-2daf-11eb-8d50-b76e331aa5c5.png">

**In an AT site:** Seedlet can be activated. 

## Result

Unable to activate Seedlet theme in a simple site. 

## Expected

Clicking "Activate" should activate the theme on simple sites as well. 

* User report: #23208137-hc 
* Follow-up ticket: #3514179-zd

jffng:
Thanks for reporting! Closed in #2854.

PR #2854, opened 2020-11-23
Seedlet: fix minimum version bug

jffng:
Addresses #2853.

D53193 on phabricator.

### DIFF
--- a/seedlet/assets/sass/style.scss
+++ b/seedlet/assets/sass/style.scss
@@ -7,7 +7,7 @@ Description: Seedlet is a free WordPress theme. A two-column layout and classica
 Requires at least: 5.4
 Tested up to: 5.5
 Requires PHP: 5.6.2
-Version: 1.1.2-wpcom
+Version: 1.1.3-wpcom
 License: GNU General Public License v2 or later
 License URI: LICENSE
 Text Domain: seedlet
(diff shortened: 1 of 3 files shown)


seedlet/assets/sass/style.scss
```diff
@@ -4,10 +4,10 @@ Theme URI: https://wordpress.com/theme/seedlet
 Author: Automattic
 Author URI: https://automattic.com/
 Description: Seedlet is a free WordPress theme. A two-column layout and classically elegant typography creates a refined site that gives your works and images space to breathe - and shine. Seedlet was built to be the perfect partner to the block editor, and supports all the latest blocks. Writing, audio, illustrations, photography, video - use Seedlet to engage and direct visitors' eyes, without your theme getting in the way. And the responsive design shifts naturally between desktop and mobile devices. Seedlet is a great option for professionals and creatives looking for a sophisticated vibe. Whether you're looking to create a blog or a robust site promoting your business, do with simplicity, style, and Seedlet.
-Requires at least: 5.5.0
+Requires at least: 5.4
```
jffng, line 7: Spearhead works and that has its "requires at least" set to 5.4.

scruffian:
Thanks for fixing this. According to [this comment](https://github.com/Automattic/themes/pull/2366#issuecomment-677779666) the minimum version should be 5.5.0, so maybe we should look into why that isn't working on wordpress.com

PR #2858, opened 2020-11-24
Require theme files without assuming the theme is installed

scruffian:
Spearhead was assuming that it was activated when including this file, but there are some cases where this is loaded without the theme being active. This change will load the correct files regardless of the current active theme.


jffng (approved):

jffng:
Committed in r57364.

PR #2861, opened 2020-11-25
Seedlet + Spearhead: correct required version numbers

jffng:
This PR corrects the required version numbers in Seedlet and Spearhead.


scruffian (approved):
LGTM. I guess we need to wait for the version numbers on wpcom to be correct before we ship this

jffng:
I believe we can ship this now because the wpcom version is now `5.6-wpcom-49518`. 

Committed in r57377.

PR #2865, opened 2020-11-26
fix spearhead fatal

scruffian:
This require was causing a fatal on Atomic sites. By duplicating the file we avoid the fatal :)

### DIFF
--- a/spearhead/inc/wpcom-colors.php
+++ b/spearhead/inc/wpcom-colors.php
@@ -1,7 +1,6 @@
 <?php
 
-// We can't assume that the theme is activated when this is called, so we can't rely on get_template_directory().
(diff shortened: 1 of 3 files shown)


spearhead/inc/wpcom-colors.php
```diff
@@ -1,7 +1,6 @@
 <?php
 
-// We can't assume that the theme is activated when this is called, so we can't rely on get_template_directory().
-require_once get_theme_root() . '/pub/seedlet/inc/wpcom-colors-utils.php';
+require_once 'wpcom-colors-utils.php';
```
aidvu: Always use something like `__DIR__ . '/wpcom-color-utils.php'`.

PR #2891, opened 2020-12-05
Update mobile nav to be on the side

enejb:

This PR update the mobile nav on the following themes. The update is to make the theme more streamlines by moving the menu to the right side of the site title and description.

For example: 
Before:
Screen Shot 2020-11-30 at 2 26 28 PM

After:
Screen Shot 2020-11-30 at 2 28 32 PM

Since the themes are already available on site on WordPress.com and we don't want to change the navigation on those sites. We will have to overwrite the `function varia_mobile_nav_on_side( $classes )` in a mu-plugin by defining that function before the one that is provided by the theme. 

**Alves**
<details><summary>Screenshots</summary>
Screen Shot 2020-12-04 at 2 20 14 PM
Screen Shot 2020-12-04 at 2 20 09 PM

</details>

**Barnsbury**
<details><summary>Screenshots</summary>
Screen Shot 2020-12-04 at 2 19 24 PM
Screen Shot 2020-12-04 at 2 19 14 PM
</details>

**Dalson**
<details><summary>Screenshots</summary>
Screen Shot 2020-12-04 at 3 13 40 PM
Screen Shot 2020-12-04 at 3 13 35 PM

</details>

**Hever**
<details><summary>Screenshots</summary>
Screen Shot 2020-12-04 at 2 54 22 PM
Screen Shot 2020-12-04 at 2 54 11 PM

</details>

**Mayland**
<details><summary>Screenshots</summary>
Screen Shot 2020-12-04 at 2 43 27 PM
Screen Shot 2020-12-04 at 2 43 20 PM

</details>

**Morden**
<details><summary>Screenshots</summary>
Screen Shot 2020-12-04 at 3 24 10 PM
Screen Shot 2020-12-04 at 3 24 05 PM

</details>

**Rivington**
<details><summary>Screenshots</summary>
Screen Shot 2020-12-04 at 2 49 30 PM
Screen Shot 2020-12-04 at 2 49 24 PM

</details>

#### Related issue(s):
See pNEWy-doZ-p2


kathrynwp:
It would be great if you could please make sure this also looks OK when the site has a logo set. Thanks!

kathrynwp:
This change could also affect users with custom CSS, so would be good to give Happiness and the CSS guild a heads-up once deployed.

enejb:
Thanks for raising the issues regarding site logos and custom css @kathrynwp. 

I haven' tested the change with a custom logo yet. I will do so today.

The plan for the roll out is to not change sites that have the theme applied already on .com but to only apply the change to new sites. I am not sure what we can can do for .org sites or atomic sites. 🤔 So the issue with custom css shouldn't be a problem since sites should maintain the same css as before.

scruffian:
The CSS looks the same for every child theme - is that correct? If so then i'd add the CSS to Varia even though its not used in that theme, and that way we reduce the duplication - each child will inherit the CSS anyway...

alves/functions.php
```diff
@@ -231,3 +233,10 @@ function alves_editor_styles() {
 	}
 }
 add_action( 'enqueue_block_editor_assets', 'alves_editor_styles' );
+
+// This makes it possible to define the function in earlier to alter if the class should be applied or not.
+if ( ! function_exists( 'varia_mobile_nav_on_side' ) ) {
```
scruffian: I think we can also add this to Varia, then the child theme only needs to add the one liner for the filter.

enejb: I was thinking about this as well. Initially I had the function defined in `varia/functions.php` the issue that I ran into is that if the site doesn't update the parent theme (varia) to the latest version. We end up with the site creating PHP warning which is no a issue on .com but maybe atomic and .org sites. Since as far as I know don't have a good way to also update the parent theme when you update the child theme. 

So I decided to move the functions to the theme's functions.php file.

enejb:
> The CSS looks the same for every child theme - is that correct? If so then i'd add the CSS to Varia even though its not used in that theme, and that way we reduce the duplication - each child will inherit the CSS anyway...

@scruffian I updated the sass files to that they are the code that repeats itself is just in once place.

alves/functions.php
scruffian: Varia themes aren't on .org. In theory it should be fine on Atomic too, as the parent theme should get updates when the child does. You might want to double check that with the Atomic team though.

barnsbury/sass/_extra-child-theme.scss
```diff
@@ -393,3 +393,6 @@ a {
 		margin-top: $spacing_unit;
 	}
 }
+// Updates the Mobile Navigation to be next to the site title.
+$mobile-nav-side-right-margin: 100px;
+@import "../../varia/sass/components/header/_site-mobile-nav-side";
```
scruffian, line 397: This approach also relies on the parent theme being updated.

Do we need to import the CSS here? Can't we just add the CSS itself to the parent theme and then add one rule to change the right margin for each child?

alves/functions.php
scruffian: Also we can just wrap the filter in a check for the function and avoid the warning that way.

scruffian:
I think it would be worth making this an option in the Customizer, so that users who have it set up the old way can change it. 

You can see an example of how to do that here: https://github.com/Automattic/themes/blob/0100e61255bc8175d354426db440006b75e4eb2c/spearhead/inc/wpcom-colors.php#L30

barnsbury/sass/_extra-child-theme.scss
enejb: The imported css here gets expanded and and combined into the css that ships with the child themes. 
So it does rely on the parent theme but only during the development. So I think it is fine in this case.

scruffian: Oh good point :)

MaggieCabrera:
I tried this and it worked fine for me, but I wonder if we want to tweak this a bit when a logo is active on the site (this is Barnsbury Theme): 

Screenshot 2020-12-09 at 11 39 47

Edit: tagging @kjellr for design input

barnsbury/style-rtl.css
```diff
@@ -4390,3 +4390,32 @@ body.admin-bar .widget_eu_cookie_law_widget.widget.top {
 .wp-block-latest-posts .wp-block-latest-posts__post-full-content {
 	margin-top: 16px;
 }
+
+@media only screen and (max-width: 559px) {
+	.mobile-nav-side .has-menu .site-logo,
+	.mobile-nav-side .has-menu .site-title,
+	.mobile-nav-side .has-menu .site-description {
+		margin-left: 100px;
```
MaggieCabrera: I'm worried if this value will work if the button changes sizes. I'm not sure if other languages will change the width of the button.

varia/sass/components/header/_site-mobile-nav-side.scss
```diff
@@ -0,0 +1,31 @@
+// Updates the Mobile Navigation to be next to the site title.
```
scruffian, line 4: nitpick: Looks like this file has spaces not tabs

scruffian:
I added a commit to move the option to the menus section as that made more sense in retrospect.

alves/functions.php
```diff
@@ -102,6 +102,21 @@ function alves_setup() {
...
 		);
+
+		// Add .mobile-nav-side body class.
+		if ( function_exists( 'varia_mobile_nav_on_side' ) ) {
+			add_filter( 'body_class', 'varia_mobile_nav_on_side' );
+		}
+
+		// Enable the mobile nav on side on theme switch.
+		if ( function_exists( 'varia_enable_mobile_nav_on_side' ) ) {
+			add_action( 'after_switch_theme', 'varia_enable_mobile_nav_on_side' );
```
scruffian: This is great

scruffian:
This is looking good. I think @MaggieCabrera  is right though we can't use absolute positioning for the menu button as we can't be sure how wide it will be. I'd use flex box. Let us know if you need any help with it. Thanks for all your work on this.

kjellr:
> I wonder if we want to tweak this a bit when a logo is active on the site (this is Barnsbury Theme):

> This is looking good. I think @MaggieCabrera is right though we can't use absolute positioning for the menu button as we can't be sure how wide it will be. I'd use flex box. Let us know if you need any help with it. Thanks for all your work on this.

Yeah, if we can use flexbox (or grid) there, it should take care of the problem.

varia/sass/components/header/_site-mobile-nav-side.scss
```diff
@@ -0,0 +1,86 @@
...
+				font-size: #{map-deep-get($config-global, "font", "size", "xs" )};
+			}
+			#toggle-cart[for="woocommerce-toggle"] {
+				margin-top: $baseline-unit;
+			}
+			#toggle-cart[for="woocommerce-toggle"].button svg {
+				margin-top: -2px;
+			}
+
+			.menu-main-menu-container {
```
scruffian: This class is different depending on the name of the menu. I think we'll need to target it with `> div`

enejb: This is now fixed

MaggieCabrera:
I have been looking into this issue and I went ahead and pushed a few changes to the original approach. After talking with @scruffian it seemed simpler to go with `grid-template-areas` and the solution lends itself to adaptation for the themes that have a slightly different html.

I have found some problems with this approach that are not easily solvable without changing the html. The first one is that the menu needs to use position absolute and moved out of the flow. This causes the open menu to cover the social menu and sometimes the site tagline if the font size is big (it happens in Alves). I tried to maintain the general design of the dropdown menu but I believe we need some design input before we go any further. @kjellr if you can have a look at this, it would be very helpful, thank you.

Lastly, since we are applying a background color to the menu that wasn't needed before, we need to take it into account for color variations. I have left that out of my commits and will get to it after we are done with any design changes in case that interferes in any way. I'm pointing this out here as a reminder to do that before we are done with this PR.

### Screenshots

There are several possibilities that come into play here:

Only logo | Only title + tagline
--- | ---
Screenshot 2020-12-11 at 12 20 56 | Screenshot 2020-12-11 at 12 20 39

No logo, no title | Logo + Title and tagline
--- | ---
Screenshot 2020-12-11 at 12 21 41 | Screenshot 2020-12-14 at 11 25 49


Here we can see the problem with the overlap of the menu in **Alves**:


Screenshot 2020-12-14 at 11 15 42

varia/sass/components/header/_site-mobile-nav-side.scss
```diff
@@ -0,0 +1,107 @@
...
+				font-size: #{map-deep-get($config-global, "font", "size", "xs" )};
+			}
+			#toggle-cart[for="woocommerce-toggle"] {
+				margin-top: $baseline-unit;
+			}
+			#toggle-cart[for="woocommerce-toggle"].button svg {
+				margin-top: -2px;
+			}
+
+			.main-menu-container {
```
scruffian: I don't think we can use class names here as they change for each menu. The way we do it in other themes is with a `>div` selector.

MaggieCabrera: Changed it

kjellr:
> I have found some problems with this approach that are not easily solvable without changing the html. The first one is that the menu needs to use position absolute and moved out of the flow. This causes the open menu to cover the social menu and sometimes the site tagline if the font size is big (it happens in Alves). I tried to maintain the general design of the dropdown menu but I believe we need some design input before we go any further. @kjellr if you can have a look at this, it would be very helpful, thank you.

Thanks for digging into this, @MaggieCabrera. Outside of changing the HTML (which I understand we don't want to do), I think the next-best solution would be to have the expanded menu take up the full viewport. This gets a little weird when a header logo are involved, but it still works: 

No Site Logo|With Site Logo
---|---
Kapture 2020-12-14 at 13 28 41|Kapture 2020-12-14 at 13 35 31

We would need to prevent the background from scrolling when this menu is open, but I _think_ that should be doable.

MaggieCabrera:
> We would need to prevent the background from scrolling when this menu is open, but I think that should be doable.

I think we can do that, thanks for the input! In the cases where the woocommerce cart button is present, I assume we want to hide it, right? In fact we should swap the order from left to right the way the two buttons are now, so that the cart one is to the left of the menu button, does that make sense?

kjellr:
Yeah, the WooCommerce button should be hidden when the menu is open. And I agree that it makes sense to show the Menu button to the right of the Woo Cart. 👍

enejb:
Thanks for all the changes @MaggieCabrera ! 

It is starting to look really nice. 

I found that 2 themes. Rivinstong and Mayland didn't quite follow the same layout. 

See 
| Mayland   | Rivinston      |  
|----------|:-------------:|
|  Screen Shot 2020-12-14 at 3 28 18 PM | Screen Shot 2020-12-14 at 3 35 37 PM |

Notice that the site title and the site description take up the 100% of the column. 

On the  Barnsbury, Daltson, Hever and Morden themes the Woo menus ends below the current Menu button. For example. 

Screen Shot 2020-12-14 at 3 28 56 PM Morden theme. 

Would it be possible but can we try to align them to be on the same line. I think this might not be possible with all the different language combination out there though.

Thanks again for your help with this!

MaggieCabrera:
Working on the issue pointed by @enejb for Rivington and Mayland I'm a bit blocked by the wrapper div (.site-branding) to make it work with the grid approach. I have found out a very nifty rule: `display: contents;` This rule lets the children of the element act as if they were direct children of their "grandparent". The drawback: [it doesn't work in IE](https://caniuse.com/css-display-contents). My latest commits include this solution for testing.


Before | After 
--- | ---
Screenshot 2020-12-16 at 10 59 05 | Screenshot 2020-12-16 at 10 58 58


The options I see to go around this are:

- [ ] Keep the `display: contents;` solution and don't support IE. It doesn't look broken, we can keep the "before" option for IE and the "after" for the rest of the browsers. 
- [ ] Solve this with JS. I don't like this idea.
- [ ] Remove the wrapper and adapt the CSS on desktop so the styles aren't affected by that. This is potentially complicated.

scruffian:
This looks great. I noticed is that the menu button moves slightly when it's opened, like this:
variamenu

kjellr:
Thanks Maggie! I tested in Alves, and noticed a couple bugs: 

Kapture 2020-12-16 at 09 03 16

- The first is the same thing that @scruffian noted — the menu button jumps a little but when it's open. 
- And the second is that the cart menu appears to be clipped and scrolling in a tiny area. 

---

On Rivington, I noticed that the cart button appears taller than the menu button for some reason, but only if I have a site logo: 

Screen Shot 2020-12-16 at 9 07 02 AM

---

> Keep the display: contents; solution and don't support IE. It doesn't look broken, we can keep the "before" option for IE and the "after" for the rest of the browsers.

Wow, `display: contents` is weird! If we can fallback to the old way of doing this for IE (left-aligning the menu), then I think we're ok to keep using this rule here. It does seem to solve things in my testing.

MaggieCabrera:
> Wow, display: contents is weird! If we can fallback to the old way of doing this for IE (left-aligning the menu), then I think we're ok to keep using this rule here. It does seem to solve things in my testing.

I had an aha moment when I noticed that this is a non-issue, since this rule will only be applied for mobile, we don't need to support IE there.

I will take care of the other issues and if @kjellr thinks we are good on the design side then I think this should be good to go.

enejb:
This is how rivington is looking now. 


Screen Shot 2020-12-16 at 2 25 05 PM
Screen Shot 2020-12-16 at 1 21 21 PM
Screen Shot 2020-12-16 at 1 21 16 PM


Dalston 
Screen Shot 2020-12-16 at 2 26 19 PM
Screen Shot 2020-12-16 at 2 26 14 PM
Screen Shot 2020-12-16 at 2 26 10 PM

dalston/style.css
```diff
@@ -4339,7 +4339,6 @@ a {
 		display: flex;
 		align-items: flex-start;
 		justify-content: end;
-		white-space: nowrap;
```
MaggieCabrera: I believe we need this to avoid the buttons from stacking on top of each other if the site title is very long

scruffian (approved):
I've tested all the themes and it looks good in all of them. Great work 👏

MaggieCabrera:
Merged and deployed (D54426-code). @kjellr if you find any adjustments that need to be made to the styles do let me know and I'll get to them. @enejb and @scruffian thanks for your work

PR #2909, opened 2020-12-14
fix editor overflow on mobile for nested full width elements

roo2:

When group blocks have a background set, they are displayed in the editor with 32px of padding, this rule is **only applied on larger screens** (`min-width: 560px;`). There is a rule in [gutenberg's editor.css#32](https://github.com/WordPress/gutenberg/blob/master/packages/block-library/src/group/editor.scss#L32) which makes "full width" elements expand outside of this padding so that they are touching the sides of the window with `margin-left: -30px;`. 

But the `margin-left: -30px;` rule in gutenberg applies on Mobile devices while the padding we apply here is **only being applied on desktop**. So the end result is that on mobile, full width elements inside groups with backgrounds are being pushed outside of the screen.

Screen Shot 2020-12-14 at 6 20 02 PM

This PR applies consistent padding to the group and offset to the full width element, rather than relying on the offset that is set in [gutenberg](https://github.com/WordPress/gutenberg/blob/master/packages/block-library/src/group/editor.scss#L32). For desktop the offset is 32px, and mobile it is 16px. 

### Context

We originally noticed this with [the vesta homepage on wordpress.com](https://github.com/Automattic/wp-calypso/issues/47293) which has a "Layout Grid" inside a group with a background color. I can also reproduce the issue with a full-width image and other blocks. 
98731971-3162c800-23a7-11eb-9c3a-1e8c45c05048

**To reproduce**
* On a site with gutenberg installed and one of the affected themes active 
  * if using a wordpress.com sandbox, you can follow [these repro instructions](https://github.com/Automattic/wp-calypso/issues/47293) using a new site created with the vesta theme through the https://wordpress.com/new flow
* Add a group block
* Set a background color on the group
* Add an image block to the group
* Set the image to be "Full Width"
* Resize the screen to below 560px

The image will extend off the screen and cause horizontal scrolling in the editor.

Apply this change by running sandbox push and following the prompts
```
npm run sandbox:push
```

The image will be constrained to full width in the editor and no horizontal scrolling or clipping will occur

### DIFF
--- a/alves/package-lock.json
+++ b/alves/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "alves",
-  "version": "1.5.7",
+  "version": "1.5.8",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/alves/package.json
+++ b/alves/package.json
@@ -1,6 +1,6 @@
 {
   "name": "alves",
-  "version": "1.5.7",
+  "version": "1.5.8",
   "description": "Alves",
   "bugs": {
     "url": "https://github.com/Automattic/themes/issues"
--- a/alves/sass/style-child-theme.scss
+++ b/alves/sass/style-child-theme.scss
@@ -5,7 +5,7 @@ Author: Automattic
 Author URI: https://automattic.com/
 Description: Convincing design for your charity or organization’s online presence. Highlight your actions, causes and projects, Alves is versatile enough to be your personal site too.
 Requires at least: WordPress 4.9.6
-Version: 1.5.7
+Version: 1.5.8
 License: GNU General Public License v2 or later
 License URI: LICENSE
 Template: varia
--- a/alves/style-editor.css
+++ b/alves/style-editor.css
@@ -724,13 +724,30 @@ object {
 }
 
 .wp-block-group.has-background {
-	padding: 21.312px;
+	padding: 16px;
 }
 
 @media only screen and (min-width: 560px) {
 	.wp-block-group.has-background {
 		padding: 32px;
 	}
+}
+
+[data-align=full] .wp-block-group.has-background > .wp-block-group__inner-container > [data-align=full] {
+	margin-left: -16px;
+	width: calc(100% + 32px);
+}
+
+@media only screen and (min-width: 560px) {
+	[data-align=full] .wp-block-group.has-background > .wp-block-group__inner-container > [data-align=full] {
+		margin-left: -32px;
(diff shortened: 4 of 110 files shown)


roo2:
Since this PR touches files in the base theme "varia" I had to do a rebuild of child themes, I noticed that the npm script in varia `npm run child-theme` referenced a script which no longer exists.

I actually found a `rebuild-child-themes.sh` script in the version of varia that runs on wpcom and included that in this PR

scruffian:
Do you think it would make more sense to fix this in Gutenberg?

ramonjd:
What's the priority on this one? Could we add an issue over at [Gutenberg](https://github.com/WordPress/gutenberg/issues) if folks think that'd be better/easier?

scruffian:
cc @jeffikus

roo2:
@scruffian 
> Do you think it would make more sense to fix this in Gutenberg?

I had a look at creating a fix in gutenberg but it doesn't make sense because gutenberg's behaviour is actually working correctly on mobile for other themes:
#### Core with blank canvas theme
Screen Shot 2021-03-31 at 1 10 55 PM

#### One of the affected themes (Alves)
Screen Shot 2021-03-31 at 1 11 52 PM

#### Affected theme with this fix applied
Screen Shot 2021-03-31 at 3 56 27 PM

I rebased this PR and updated the selector that's used

scruffian:
If the problem is only on Varia themes rather than in Gutenberg would it be possible to fix it by removing code from Varia rather than adding more?

roo2:
> would it be possible to fix it by removing code from Varia rather than adding more?

hmm good point I'll have a look at that

roo2:
ok, actually, on the live site, we have a 16px padding around the images on mobile and 32px on desktop.

The other themes that inherit from `seedlet` already display the images with matching padding in the editor.

I removed code and copied the aproach from seedlet so that the editor matches the live site for these themes that inherit from `varia`

### Live site 16px padding on mobile
Screen Shot 2021-04-07 at 10 07 25 PM

### Editor 16px padding on mobile
Screen Shot 2021-04-07 at 10 07 00 PM

roo2:
Actually, this issue has been fixed in core [with this commit](https://github.com/WordPress/gutenberg/commit/50cdfbeae59cf0fe91faa2183877b3b7fc60dd3e#diff-1a9996754d0946dc5f83e38ec6b5e6bc139ec4d16690ec8c9d4b5402175bc17b), the editor experience for these themes now matches what shows up on the live site

PR #2954, opened 2021-01-01
Stow: Button Alignment Fix

ghost:
Issue:

Block button is misaligned, the button should appear in the center.  

Fix:

Added the CSS code below to make sure buttons with class attribute aligncenter will always be displayed in the center. 

`.entry-content > .wp-block-button:not(.alignleft):not(.alignright):not(.alignwide):not(.alignfull) {
    margin-right: auto;
    margin-left: auto;
}`

Before:

Button Misaligned

After:

Button Fixed

### DIFF
--- a/stow/style-rtl.css
+++ b/stow/style-rtl.css
@@ -3605,6 +3605,11 @@ img#wpstats {
 	margin-left: auto;
 }
 
+.entry-content > .wp-block-button:not(.alignleft):not(.alignright):not(.alignwide):not(.alignfull) {
+    margin-right: auto;
+    margin-left: auto;
+}
+
 @media only screen and (min-width: 560px) {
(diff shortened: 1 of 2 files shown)


wprhunt:
This seems to be the case for at least the Maywood theme as well if not all Varia themes.

In a previous slack conversation we found this CSS can work as a work around:

/* fix alignment of centered button blocks | #XXXXXX-hc eD */
.entry-content > .wp-block-button.aligncenter:not(.alignleft):not(.alignright) {
	margin-right: auto;
	margin-left: auto;
}

The issue seems to step from a margin-left being added to the button block when its not set to left or right alignment:

 .entry-content > .wp-block-button:note(.alignleft):not(.alignright){
 margin-left: calc( 0.5 * (100vw - calc( 782px - 32px));
 }
 
 Not sure why margin to the left would have to be added to a centered element.

image

ghost:
You are probably right, I see the same issue in [Hever](https://github.com/Automattic/themes/tree/trunk/hever) and [Red Hill](https://github.com/Automattic/themes/tree/trunk/redhill)

hacchism:
Another case with Alves theme: 31507935-hc

PR #2960, opened 2021-01-04
Fix quote block alignment in the editor

### LINKED ISSUE #2947: Varia + Seedlet & children: Quote block misaligned in the editor
For Varia + children and Seedlet + children, the alignment of the quote block is broken in the editor (the frontend looks fine). This is caused by some editor only styles that remove the margin auto values. Those styles are not necessary and should be removed.

**Seedlet:** 

Screenshot 2020-12-29 at 12 01 17

**Varia:**

Screenshot 2020-12-29 at 12 01 02

* User report: p1609175603455100-slack-C029FM1EH
* Trac issue for Default Themes: https://core.trac.wordpress.org/ticket/52009

jffng:
This PR address #2947. Diff ID is D54837 for testing on wpcom.

**Before**
Screen Shot 2021-01-04 at 1 20 12 PM

**After**
Screen Shot 2021-01-04 at 1 19 57 PM

### DIFF
--- a/alves/sass/style-child-theme.scss
+++ b/alves/sass/style-child-theme.scss
@@ -5,7 +5,7 @@ Author: Automattic
 Author URI: https://automattic.com/
 Description: Convincing design for your charity or organization’s online presence. Highlight your actions, causes and projects, Alves is versatile enough to be your personal site too.
 Requires at least: WordPress 4.9.6
-Version: 1.4.12
+Version: 1.4.14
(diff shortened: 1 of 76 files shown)


scruffian (approved):
LGTM. The only strange thing is that the version numbers are bumped twice. Not a big deal...

jffng:
I noticed that... maybe because I ran the `npm children` script twice?

PR #2961, opened 2021-01-04
Try: Add hide site header/footer Customizer options to Varia

kjellr:
This PR tries out some new options in Varia: 

- Hide Site Header
- Hide Site Footer (But just for the Menu... we still want the footer credit to appear.)

This could come in handy for single-page sites like those in https://github.com/Automattic/wp-calypso/issues/48381. 

For now I placed these checkboxes in the existing Hopepage Options area, but they could be expanded if we want them to be global (or on specific pages). It's worth noting that you can already hide most of this stuff globally if you manually remove menus from the menu locations, and also check the "Display Site Title and Tagline" option under Site Identity.

_This shouldn't be merged before we test it out in all the child themes!_ 😉

## Screenshots

Default: 

Screen Shot 2021-01-04 at 2 44 39 PM

Site Header hidden:

Screen Shot 2021-01-04 at 2 44 49 PM

Site Header & Footer Menu hidden: 

Screen Shot 2021-01-04 at 2 44 58 PM


kjellr:
I tested this out on child themes, and it's looking pretty good:

Screen Shot 2021-01-06 at 14 31 43|Screen Shot 2021-01-06 at 14 32 42|Screen Shot 2021-01-06 at 14 33 17
---|---|---
Screen Shot 2021-01-06 at 14 34 21|Screen Shot 2021-01-06 at 14 35 52|Screen Shot 2021-01-06 at 14 36 34
Screen Shot 2021-01-06 at 14 39 05|Screen Shot 2021-01-06 at 14 45 52|Screen Shot 2021-01-06 at 14 40 27


In most cases, the theme just needs some padding adjustments (if anything). Hiding the header on Brompton doesn't work at all though, so that would need some extra work.

kjellr:
@scruffian: I did the initial work to get this PR working in a similar way to #2985. I do a concern with the approach though: 

For Seedlet, we needed to make edits to both the child themes to get this working there. This wasn't a big deal because there was only one of them (Spearhead). It's going to be much more work to edit and test each of the 19 Varia child themes. I imagine some will be pretty straightforward, while others will need a bit of header/footer refactoring. This turns a relatively simple PR into something quite massive. 

While the CSS-Only approach wasn't great in that it relied on `display:none`, one positive thing is that it was so simple in code + execution.  

Any thoughts on how to make this easier?

scruffian:
From looking at the code for the Varia child themes headers and footer I think they could benefit from some refactoring anyway, so I think it's worth doing it to keep these themes maintainable. What do you think?

kjellr:
That sounds about right. 😄 I'll take another look tomorrow — if someone else on the team has spare time they're welcome to dig in sooner.

kjellr:
Marking this as "To do", and removing my assignment since I don't think I'll have a chance to get to it this week.

MaggieCabrera:
I'm working on this now too, I added support for Alves.

Things you need to add to your footer when testing:

- A privacy policy link
- A footer menu needs to be defined
- Widgets on the footer area (each theme may have them different)
- Blog name and description defined

After we hide the footer, only the footer credit must show:

Before | After
--- | ---
Screenshot 2021-01-19 at 12 24 36 | Screenshot 2021-01-19 at 12 19 31

brompton/template-parts/footer/footer-content.php
```diff
@@ -0,0 +1,29 @@
...
+		<?php if ( ! empty( $blog_info ) ) : ?>
+			<a class="site-name" href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a><span class="comma">,</span>
+		<?php endif; ?>
+		<a href="<?php echo esc_url( __( 'https://wordpress.org/', 'varia' ) ); ?>" class="imprint">
+			<?php
+			/* translators: %s: WordPress. */
+			printf( __( 'proudly powered by %s.', 'varia' ), 'WordPress' );
+			?>
+		</a>
+		<?php get_template_part( 'template-parts/footer/privacy-policy', '' ); ?>
```
scruffian: Technically this is a change - previously the privacy policy was inside the wordpress.org link, but I assume that is a bug.

varia/header.php
```diff
@@ -42,63 +30,6 @@
...
-						)
-					);
-					?>
-				</nav><!-- .social-navigation -->
-			<?php endif; ?>
-
-		</header><!-- #masthead -->
-
-	<?php endif; ?>
+	<?php get_template_part( 'template-parts/header/header', 'content' ); ?>
```
scruffian: Can we move this code here:
```
<?php 
if ( ( true === get_theme_mod( 'hide_site_header', false ) && is_front_page() && is_page() ) ) : return; // Return if this is the homepage and the hide-header setting is active. ?>
```
Then we don't need to repeat it in every template part.

MaggieCabrera: yes, I believe for the header we will be able to do that, I will go ahead with that, thx

scruffian:
For Brompton the footer is still quite obtrusive:

Screenshot 2021-01-20 at 10 27 07

I wonder if we should scale it down if the footer is hidden?

MaggieCabrera:
> For Brompton the footer is still quite obtrusive:
> 
> Screenshot 2021-01-20 at 10 27 07
> 
> I wonder if we should scale it down if the footer is hidden?

Thought so too. Maybe remove the background and keep the text black when the footer is hidden? This will affect color annotations by the way :D Maybe design input for this? @kjellr

scruffian:
> Thought so too. Maybe remove the background and keep the text black when the footer is hidden? This will affect color annotations by the way :D Maybe design input for this?

Probably this makes sense in a followup PR anyway. And maybe only if we get that feedback...

spearhead/variables.css
```diff
@@ -102,7 +103,7 @@
 }
 
 @media ( prefers-color-scheme: dark ) {
-	:root {
+	:root :not(.custom-colors) {
```
MaggieCabrera: this is coming from somewhere else :D

scruffian: oops

scruffian (approved):

PR #2962, opened 2021-01-04
Try: Adding a "Blank" Theme

kjellr:
Trying out a blank theme, tentatively called “Blankie.” It (mostly) just outputs the post content. Something like this could come in handy for single-page sites, like those shown in https://github.com/Automattic/wp-calypso/issues/48381.

It's a Seedlet child theme, and though it could still could use some refinement, it's generally working alright so far. 

Archive pages _do_ work, even though they probably wouldn't really be used here. They do look a little weird, since there's no page header though. Also, the theme removes menus entirely, so navigation to other pages requires a direct link. 

Todo: 

- [x] Remove post meta on archive pages.
- [x] Adjust color scheme.
- [x] Add a theme screenshot.
- [x] Come up with a better name(?)

## Screenshots

Screen Shot 2021-01-04 at 16 10 32

Same page, with the Seedlet Theme (for comparison):
Screen Shot 2021-01-04 at 16 03 22

cc @rickybanister @ianstewart @scruffian


kjellr:
A few screenshots of archive pages, for reference: 

Default Posts List|Category|Search
---|---|---
Screen Shot 2021-01-04 at 16 27 26|Screen Shot 2021-01-04 at 16 26 54|Screen Shot 2021-01-04 at 16 27 13

We could easily add a site title or something to these pages if we felt it made sense to (I think it might, personally).

scruffian:
I removed the navigation assets as they won't be needed anymore.

In terms of the color scheme I think the one from Seedlet looks fine.

In terms of a name, what about "Single Page Site"?

kjellr:
> In terms of a name, what about "Single Page Site"?

I think that might work for our purposes, (Maybe just "Single Page" for short?) but it's a little deceptive since the theme _technically_ supports more than that. 

BTW, it occurs to me that we'll definitely need some clear docs for this one. 

@ianstewart would the default head start content for this theme be one of the starter designs from https://github.com/Automattic/wp-calypso/issues/48381? Or would we really want it to be 100% blank and empty to begin with? (If it's the latter, I'm not sure what would be on the screenshot. 😅)

scruffian:
The screenshot could just be:

Screenshot 2021-01-05 at 14 17 39

scruffian:
What about "Blank Page" for a name?

kjellr:
I like "Blank Page". It's simple and gets the single page point across. 

Using a totally blank screenshot is weird, but I guess it works for now. 😄 I'll add that shortly.

kjellr:
I pushed a few updates. 

- The theme name has been changed to "Blank Page"
- The theme includes a blank screenshot.
- The color has been changed to blue. 
- I added wpcom color annotations, copied over from Spearhead. (Though I don't know of a way to actually test these before the theme is actually committed to wpcom, right @scruffian?)
- [As I suggested above](https://github.com/Automattic/themes/pull/2962#issuecomment-754229939), I tidied up the archive pages so that they include a logo/title/tagline _on archive pages only_. This seems reasonable to me, and it also provides a link back to the homepage. 

Here's an updated set of screenshots: 

Single Page/Post|Posts Index|Archive|Search Results
---|---|---|---
Screen Shot 2021-01-05 at 12 41 18|Screen Shot 2021-01-05 at 12 41 05|Screen Shot 2021-01-05 at 12 41 41|Screen Shot 2021-01-05 at 12 41 48

blank-page/readme.txt
```diff
@@ -0,0 +1,33 @@
...
+Tested up to: 5.6
+Requires PHP: 5.6.2
+License: GPLv2 or later
+License URI: http://www.gnu.org/licenses/gpl-2.0.html
+
+A blank starting point for building your site.
+
+== Description ==
+
+Seedlet is a blank starting point for building your site.
```
jffng:
```suggestion
Blank Page is a blank starting point for building your site. 
```

scruffian (approved):
This looks good. I think we can go ahead and merge it.

It might be quite nice to move some of the duplicated code from the headers and footers into template parts, but not a high priority :)

blank-page/inc/wpcom-colors-utils.php
```diff
@@ -0,0 +1,192 @@
+<?php
```
jffng, line 1: @scruffian was it determined that we cannot reuse / reference Seedlet's `wpcom-colors-utils.php`?

blank-page/template-parts/content/content-singular.php
```diff
@@ -0,0 +1,43 @@
...
+
+	<?php seedlet_post_thumbnail(); ?>
+
+	<div class="entry-content">
+		<?php
+		the_content(
+			sprintf(
+				wp_kses(
+					/* translators: %s: Name of current post. Only visible to screen readers */
+					__( 'Continue reading<span class="screen-reader-text"> "%s"</span>', 'seedlet' ),
```
jffng: I don't think it matters since we won't be submitting to the theme directory, but should these text domains be `blank-page`? Or are they left as Seedlet in order to reference Seedlet's translation file?

jffng (approved):
Left some minor comments.

blank-page/inc/wpcom-colors-utils.php
scruffian: Yeah it will throw an error in Atomic sites

blank-page/readme.txt
kjellr: Good catch, thank you!

blank-page/template-parts/content/content-singular.php
kjellr: Yeah, they should probably be blank page. I'll update.

kjellr:
Thank you both! I'll get this committed today.

PR #2977, opened 2021-01-06
Blank Canvas: Only load the header if there's content in it

kjellr:
Changes the conditionals a bit in the `header.php` file so that it doesn't load in a bunch of empty HTML elements like this:

image

Instead, it'll only load the whole `<header>` element if the page is not `singular`.


scruffian (approved):
LGTM

PR #2984, opened 2021-01-07
Blank Canvas: Customizer Fixes

### LINKED ISSUE #2980: Blank Canvas: Deregistered menu locations are still appearing in the Customizer
These are de-registered here, but are still showing up in the Customizer: 

https://github.com/Automattic/themes/blob/38f4d9a546cb550e1b451dc35ac746621f03b048/blank-canvas/functions.php#L79-L82

Screen Shot 2021-01-07 at 8 06 07 AM

kjellr:
This PR tidies up customizer options for the Blank Canvas theme: 

- It increases the priority of the `blank_canvas_remove_parent_theme_features` function, to ensure that the nav menu areas are actually removed (Fixes #2980)
- It removes the "Navigation" panel from the customizer entirely, since no menu areas are present in this theme. (I guess this doesn't disappear automatically?)  
- It removes Seedlet's widget area, so the Widgets panel should no longer appear.
- It adds a message to the top of the "Site Identity" panel, to clarify that the title and tagline will never appear on single posts and pages. 

---

## Screenshots

Before|After
---|---
Screen Shot 2021-01-07 at 2 07 51 PM|Screen Shot 2021-01-07 at 2 06 50 PM

Before|After
---|---
Screen Shot 2021-01-07 at 2 09 10 PM|Screen Shot 2021-01-07 at 2 10 54 PM

### DIFF
--- a/blank-canvas/functions.php
+++ b/blank-canvas/functions.php
@@ -91,6 +91,19 @@ function blank_canvas_dequeue_parent_scripts() {
 }
 add_action( 'wp_enqueue_scripts', 'blank_canvas_dequeue_parent_scripts', 11 );
 
+function blank_canvas_remove_customizer_settings( $wp_customize ) {
+
+	// Remove the navigation menus Customizer panel.
+	$wp_customize->get_panel( 'nav_menus' )->active_callback = '__return_false';
+
+	// Remove Jetpack's Author Bio setting.
+	if ( function_exists( 'jetpack_author_bio' ) ) {
+		$wp_customize->remove_control( 'jetpack_content_author_bio_title' );
+		$wp_customize->remove_control( 'jetpack_content_author_bio' );
+	}
+}
+add_action( 'customize_register', 'blank_canvas_remove_customizer_settings', 11 );
+
 /**
  * Remove Meta Footer Items.
  */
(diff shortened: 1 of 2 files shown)


blank-canvas/functions.php
```diff
@@ -71,25 +71,55 @@ function blank_canvas_setup() {
...
+}
+add_action( 'widgets_init', 'blank_canvas_remove_widgets_area', 11 );
+
+/**
+ * Remove unused custmizer settings.
+ */
+function blank_canvas_remove_customizer_settings( $wp_customize ) {
+
+	// Remove the navigation menus Customizer panel.
+	$wp_customize->get_panel( 'nav_menus' )->active_callback = '__return_false';
```
kjellr, line 110: I expected to be able to use something like `$wp_customize->remove_panel( 'nav_menus' )`, but it throws an error:

> Notice: WP_Customize_Manager::remove_panel was called incorrectly. Removing nav_menus manually will cause PHP warnings. Use the customize_loaded_components filter instead. Please see Debugging in WordPress for more information. (This message was added in version 4.5.0.)

The approach I used instead [comes from this thread](https://wordpress.stackexchange.com/a/225234), and seems to work in my testing.

scruffian:
This looks good, but I note that it also removes Content Options which I think could be useful.

kjellr:
Oh right, I forgot to mention that part! This PR doesn't explicitly remove that panel, it just removes Jetpack's "Author Bio" setting in that panel, since the theme won't display an author bio. If that is the only setting in the panel, the panel disappears on its own when it's empty.

scruffian:
In that case I'll 🚢

PR #2985, opened 2021-01-07
Seedlet: Add hide site header/footer Customizer options for the homepage

kjellr:
Same as https://github.com/Automattic/themes/pull/2961, but for Seedlet. This also applies to its child themes. 

This PR adds customizer settings for hiding the page header and all footer elements except for the credit line. It also ensures that these new options do not show up for the Blank Canvas theme, since these elements do not exist in that theme. 

It appears to work well in Seedlet, and also in Spearhead. 

## Screenshots

Seedlet:

Kapture 2021-01-07 at 15 30 16

Spearhead:

Kapture 2021-01-07 at 15 31 48


scruffian:
I'm not sure about this approach. Hiding them with CSS will mean they are still in the source and therefore read by machines, which might have unintended impacts on SEO. Could we switch the template that the theme uses when these options are selected, and use the same header/footer that we use in Blank Canvas?

kjellr:
Yeah, this PR just duplicates the current "Hide title" functionality, which has that same issue. I'll take a look at doing what you've suggested instead though — that would definitely be a cleaner approach.

kjellr:
@scruffian This is ready for a re-test when you have a chance. It switches to using conditionals within `header.php` and `footer.php` to remove the selectors entirely when these options are active.

There's still some CSS, but it's just for some minor margin/padding and visual updates. The update also moves this stuff out of the `wpcom`-only files, which seems like a positive move anyway.

If this seems to work better, I'll update #2961 to use the same approach.

scruffian:
I've done a big refactor here. If Blank Canvas and Seedlet are both working as expected then I think we can merge.

scruffian:
I also added a fix from https://github.com/Automattic/themes/pull/2999

kjellr:
@scruffian this works well. I pushed two changes: 

1. These options should only impact the front page, so I had to adjust the conditional a bit. I'm not 100% sure this is the best way to write that, but it works: ba0a48c

2. Since these options are no longer wpcom specific, I moved Empty Canvas's customizer remove functions into the main `functions.php` file. 

Let me know what you think.

scruffian:
LGTM

kjellr:
Hmm actually I'm getting a fatal error when I test Spearhead... taking a look.

kjellr:
Ok, so that fatal error occurs in `master` too — I'll open a separate issue for it (_Edit:_ https://github.com/Automattic/themes/issues/3005). But we do need to adjust spearhead's header files to use these partials now. 😕 The footer options work fine, but the header checkbox currently has no effect.

kjellr:
These options should work for Spearhead as of 57fdf9c. This could use a final review!

Kapture 2021-01-11 at 14 18 46

scruffian (approved):
LGTM

scruffian:
Let's leave this deploy until the end of the week and do all the Seedlet changes together.

kjellr:
Cool. I'll revise #2961 to use this same approach for Varia too.

PR #2992, opened 2021-01-08
Blank Canvas: Improve top margin CSS rules

kjellr:
There are just a handful of instances where we want Blank Canvas to remove margins from the first block used in the post/page. In these cases, we want the block to appear full-bleed, right to the top edge of the page.

- Full-width Images
- Full-width Cover Blocks
- Full-width Media & Text Blocks
- Full-width Group Blocks (When they have a background color, and thus supply their own padding).

The theme previously used a faulty approach to determine when and when not to apply these margins: It removed the margin by default, and only added it back in for certain blocks. It led to some cases when top margins were removed unnecessarily. 

This PR takes the opposite approach: it includes the top padding by default all the time, and removes it only in the use cases listed out above. If/when we find other use cases we can 

## Screenshots

Here how the theme previously displayed a post that starts out with a standard-width cover block: 

Screen Shot 2021-01-08 at 10 53 52 AM

This PR corrects that and adds top margin: 

Screen Shot 2021-01-08 at 10 54 08 AM

It also retains the zero'd out top margin if that cover block were set to full-width: 

Screen Shot 2021-01-08 at 10 58 53 AM

### DIFF
--- a/blank-canvas/style.css
+++ b/blank-canvas/style.css
@@ -30,14 +30,6 @@ GNU General Public License for more details.
 
 @import "variables.css";
 
-/* Remove outside padding on single post pages. */
-
-.single .site-main,
-.page .site-main {
-	padding-top: 0;
-	padding-bottom: 0;
-}
-
 /* Remove extra margin from articles on single post pages. */
 
 .single .site-main > article,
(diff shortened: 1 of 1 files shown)


blank-canvas/style.css
```diff
@@ -47,9 +39,15 @@ GNU General Public License for more details.
 
 /* Add some top padding if the first block on the page is not full-width or a spacer. */
```
jffng:
```suggestion
/* Add some top padding if the first block on the page is a full-width image, cover, media & text, or group block. */
```

blank-canvas/style.css
```diff
@@ -47,9 +39,15 @@ GNU General Public License for more details.
...
 
-.single .entry-content > *:first-child:not(.alignfull):not(.wp-block-spacer),
-.page .entry-content > *:first-child:not(.alignfull):not(.wp-block-spacer) {
-	padding-top: var(--global--spacing-vertical);
+.single .entry-content > .wp-block-image.alignfull:first-child,
```
jffng, line 42: Could this just be a single `.entry-content > ...` rule, or is it not specific enough?

jffng: I also ask because the specificity of these rules makes it so they do not apply to posts — is that the intention?

kjellr: Yeah, it should apply to individual posts and pages, but not posts within a feed. So that's why the extra specificity is necessary: `.single` is a body class added to single posts, `.page` is for single pages.

PR #2999, opened 2021-01-10
Varia: fix to add logic to prevent empty menu button

### LINKED ISSUE #2918: Mayland: Menu button appears in header even when menu empty.
This may extend to all Varia themes. A screenshot of the desktop view and mobile view of a site with no menu, using the Mayland theme.

image

arcangelini:
Currently the menu logic checks if the primary menu is set using `has_nav_menu( 'menu-1' )`. This does not check to make sure the menu has links in it. I added another check to the logic.

```
$has_menu_items = wp_nav_menu(
	array(
		'theme_location'    => 'menu-1',
		'fallback_cb' 	    => false,
		'echo' 	            => false 
	)
);
```

This returns `false` if the menu is empty preventing the button from being added.

#### Screenshots
[Screenshot](https://d.pr/i/Qwureb)

## Related issue(s):
Fixes #2918

### DIFF
--- a/varia/header.php
+++ b/varia/header.php
@@ -11,6 +11,12 @@
  * @since 1.0.0
  */
 $has_primary_nav = has_nav_menu( 'menu-1' );
+$has_menu_items = wp_nav_menu(
+	array(
+		'theme_location' => 'menu-1',
+		'fallback_cb' => false,
+		'echo' => false 
+	));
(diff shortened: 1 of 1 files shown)


scruffian:
This looks good, thanks for fixing. I changed the variable name. If you're happy with it we can merge!

PR #3004, opened 2021-01-11
Varia themes: fix cover block + layout grid bug

### LINKED ISSUE #2976: Alves: Footer Credit Migrates To The Middle Of Page
## Steps to replicate
1. Create a new page with a blank template. 
2. Add a Group Block
3. Add a 1 Column Layout Grid Block with the following settings: https://d.pr/i/YbpRCD
4. Add 2 blocks within the Layout Grid block: a Cover Block, then a Media And Text Block with content (order of which block is added first doesn't matter).  
Video: https://d.pr/v/08u8yG
5. Publish the page and mark it as your homepage in Design > Customize > Homepage Settings. Choose "Select" for the posts page. 
6. View the homepage.

## Result

The theme's footer credit migrates from the bottom of the page to the second block. 
Video: https://d.pr/v/v7iwnG

When Group and Layout Grid Blocks are not used on a page (so just using a Cover Block and Media/Text Block on the page), the footer credit stays at the bottom of the page. 

## Expected

The footer credit to remain at the bottom of the page. CSS can "fix" this by hiding the credit but that isn't ideal especially on the user's current plan. 

* User report:  #26825302-hc
* Follow-up ticket: #3627711-zd

jffng:
This PR addresses #2976. 

To describe what's going on in more detail: 

- The cover block's default CSS contains `height: 100%;`
- When you place a cover block inside a layout grid column, the cover takes up 100% of the calculated height of the layout grid column, pushing whatever else inside that same column _beyond_ the height the column takes up in the document flow. This is why the footer credit appears in the middle as reported in 2976. 
- It happens to be that the Doyle layout uses this pattern, but any theme that implements the pattern would experience this issue.

This PR targets all direct descendants of layout grid columns to set their `height: auto`. It feels like a bandaid, and I'm not attached to this solution but wanted to put it out there to get a fix in place.

**To test**
- Follow the steps to recreate the pattern described in #2976. You can use any varia theme on a free plan.
- Verify that any elements that are placed after the group block appear to overlap the layout grid
- Apply this patch (D55158), and verify that the issue is resolved

### DIFF
--- a/alves/package-lock.json
+++ b/alves/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "alves",
-  "version": "1.4.14",
+  "version": "1.4.15",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/alves/package.json
+++ b/alves/package.json
@@ -1,6 +1,6 @@
 {
   "name": "alves",
-  "version": "1.4.14",
+  "version": "1.4.15",
   "description": "Alves",
   "bugs": {
     "url": "https://github.com/Automattic/themes/issues"
--- a/alves/sass/style-child-theme.scss
+++ b/alves/sass/style-child-theme.scss
@@ -5,7 +5,7 @@ Author: Automattic
 Author URI: https://automattic.com/
 Description: Convincing design for your charity or organization’s online presence. Highlight your actions, causes and projects, Alves is versatile enough to be your personal site too.
 Requires at least: WordPress 4.9.6
-Version: 1.4.14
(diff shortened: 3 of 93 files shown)


scruffian:
Do you think we should make this fix in the block itself?

jffng:
I think so. I will propose it.

jffng:
https://github.com/WordPress/gutenberg/pull/28114

jffng:
The above PR has been merged and fixes this issue.

PR #3008, opened 2021-01-12
Seedlet: Only hide site header/footer if the homepage is set to a static page

kjellr:
As per the checkbox labels, the options added in #2985 should only apply if the homepage is set to display a static page: 

Screen Shot 2021-01-12 at 9 15 25 AM

This mirrors wpcom's "Hide Homepage Title" functionality. Currently, the options hide the title even when an index is set to the homepage. This PR corrects that.

## Screenshots

Before

Kapture 2021-01-12 at 09 13 28

After

Kapture 2021-01-12 at 09 14 13


scruffian (approved):
LGTM

PR #3010, opened 2021-01-12
Blank Canvas: Add block patterns

kjellr:
This PR adds a set of single-page site block patterns to the Blank Canvas theme. 

Currently included: 

About Me|Invitation
---|---
Screen Shot 2021-01-12 at 11 25 53 AM|Screen Shot 2021-01-12 at 1 11 28 PM

Links|Card
---|---
Screen Shot 2021-01-12 at 1 09 58 PM|Screen Shot 2021-01-12 at 1 09 13 PM

These still need some tidying up style-wise.


ianstewart:
The the alignment of the photo subject and the buttons seems to clash a bit with the left alignment  of the heading and text in the card pattern. I wonder if playing around with a different image, cropping or alignment there could brighten it up? The photo also a bit of an apocalyptic look with the pattern being overall more "dark" in contrast with the other patterns.

Perhaps the invitation pattern could be have the text or overall height reduced a bit? The two, tall rectangles next to each other just feels a bit static or not as dynamic as it perhaps could be?

It might be nice for "theme cohesion" if the photos in the patterns could all get the same "feel" but this may be difficult if we're looking to use GPL-compatible images. Along the same vein, did you try out using the same heading weights?

ianstewart:
Looks like a right single quote wound up missing in this one …

image

melchoyce:
Let's update that copy entirely:

> You're Invited
> 
> Pier 38
> San Francisco, CA
> 
> 8pm–12am
> February 24th
> 
> Join us for a celebratory party featuring creative cocktails and mocktails, seasonal chef-crafted snacks and appetizers, and a range of desserts and pastries from local bakeries. 

image

kjellr:
Here are a few updates. I've:

- Adjusted the text for the invitation pattern.
- Synced these up better from a colors + photos perspective. 
- Added a full-screen "Split" pattern. 
- Moved the "Card" one to be WP.com-only, since I couldn't get it to work nicely without the layout grid block.
- Widened the wide alignment width so it's more clearly differentiated from the standard width.

Screen Shot 2021-01-13 at 12 16 28 PM
Screen Shot 2021-01-13 at 12 16 34 PM
Screen Shot 2021-01-13 at 12 16 48 PM
Screen Shot 2021-01-13 at 12 17 51 PM
Screen Shot 2021-01-13 at 11 55 20 AM

kjellr:
@melchoyce is also looking into adding one more pattern that would include a contact form. 👍

kjellr:
The latest commit adds an email signup pattern: 

Screen Shot 2021-01-14 at 9 34 50 AM

@melchoyce, I started with one of your layouts, but changed the topic a bit to fit a colorful photo like the other ones we're using here.

This seems like a decent set of patterns to start with — I think this should be ready for a review!

scruffian:
This looks good. What would be amazing is if the patterns could live in separate files and then the function pulls from them, so that they are easier to maintain...

kjellr:
Are you envisioning a separate "block-patterns" directory, that would house each of these as individual files? That could be cool. I'd suggest we explore that in a followup PR though.

blank-canvas/inc/wpcom.php
```diff
@@ -23,7 +23,7 @@ function blank_canvas_register_wpcom_block_patterns() {
 		
 		if ( function_exists( 'register_block_pattern' ) ) {
 
-			/* This pattern is WP.com-Only Beacuse it relies on the Layout Grid block. */
+			/* This pattern is WP.com-only because it relies on the Layout Grid block. */
```
ianstewart, line 26: @kjellr is there some additional function_exists() to check if Jetpack Layout Grid is present that could be added here? Or are there any WordPress plugin or block check functions that could be checked here for throwing notices in /wp-admin?

kjellr: I did look into that briefly, but I didn't find anything obvious. There don't seem to be any functions I could check for in [the plugin source](https://plugins.svn.wordpress.org/layout-grid/trunk/). 

That could use another set of eyes though if someone else has a moment.

kjellr: So it _really_ looks like this should work: 

```
if ( class_exists( 'WP_Block_Type_Registry' ) && WP_Block_Type_Registry::get_instance()->is_registered( 'jetpack/layout-grid' ) ) {
    register_block_pattern ( ... )
}
```

But for some reason I can't get it to. 😩

kjellr: @MaggieCabrera and I took a look at getting that working on Friday, but we didn't make any progress. 

Since the current solution (load these only on WP.com) technically works ok, I suggest we merge and follow up with a better fix later on. 

The only effected parties here are folks who use this theme outside of WP.com and also have the Layout Grid plugin installed. And even in that case, they'll currently just not see this pattern, which seems fine. 

Does that work for folks? 

If so, would someone mind giving this a final review?

pbking:
I notice that for the "Links" Block Pattern the links alignment is different for .com and .org.
image

kjellr:
@pbking did you have the Gutenberg plugin activated there? The pattern uses a 100% width control that is currently only available in the plugin (which is loaded by default on .com). 

If the plugin is turned off, things are left aligned and they wrap: 

Screen Shot 2021-01-19 at 9 47 27 AM

I think that's ok for now though, since we're only launching on WP.com to start, and I don't think it makes sense to write custom code to fix something that'll work fine out of the box once this version of the plugin is merged into core.

pbking:
Indeed, installing the Gutenberg plugin addressed the alignment issue.  If that's OK then it all LGTM.

kjellr:
Thanks!

PR #3013, opened 2021-01-12
Change layout direction of Seedlet Submenu

### LINKED ISSUE #2790: Spearhead: page titles truncated in some child (drop down) menus
When adding pages to menus, there can be an issue with longer page titles displaying correctly if entered in a child (drop down) menu in the right-most position, but happens any time the title page in the child menu overruns the right margin of the theme.  The text is truncated, and there is a (new) option to scroll right to see the full text.  

## Steps to replicate

1. set up menu with child menu on the far right, and add a page with a sufficiently long title
2. publish changes and visit live site
3. Full text of the page in the child menu is not shown unless you "scroll right"

See video of the issue in action:  https://d.pr/v/uDuJCA

## Result

Child menu page title is not visible and text overruns the page.  You can scroll right to see the full text.  As a side issue, even if the child menu is not hovered over, the page has a horizontal scroll option that is awkward and makes it appear that margins on all pages are broken

## Expected

Child menu to more gracefully handle long titles

* User report: https://wordpress.com/forums/topic/horizonal-scrolling-with-spearhead-theme/

pbking:

This changes the alignment of submenus.  It will align the container to
the RIGHT of the parent (rather than the left as it was doing) and RIGHT
ALIGN the sub-menu items (whereas previously they were left aligned).

This is to address #2790.

Before:
image
image

After:
image
image

RTL:
image

#### Related issue(s):
https://github.com/Automattic/themes/issues/2790

### DIFF
--- a/spearhead/assets/sass/navigation.scss
+++ b/spearhead/assets/sass/navigation.scss
@@ -29,6 +29,9 @@ $navigation-max-break-point: 'laptop-only';
 		@include media( $navigation-min-break-point ) {
 			padding: calc(0.5 * var(--primary-nav--padding)) calc( 2 * var(--primary-nav--padding) );
 			transition: all 0.15s ease;
+			text-align: right;
+			left: unset;
+			right: 0;
 
 			> .menu-item.menu-item-has-children {
 				padding: calc(0.5 * var(--primary-nav--padding)) calc( 2 * var(--primary-nav--padding) ) 0 0;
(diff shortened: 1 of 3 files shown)


scruffian:
This looks good. With a very long link this will still overflow:

Screenshot 2021-01-12 at 22 20 31

One option to overcome this would be to set a max width on the submenu, but it would be difficult to know exactly what that should be. I think this is fine to merge as is 🚢

scruffian (approved):

PR #3017, opened 2021-01-13
Try: Make spacing overrides more specific in Seedlet

### LINKED ISSUE #2727: Seedlet: First item in Social Icons does not align with the rest in editor
## Steps to replicate

* Add Social Icons block

## Result

Screenshot 2020-10-28 at 16 54 09

## Expected

It should align with the rest. This is what we see with Varia themes.

Screenshot 2020-10-28 at 16 55 36

### LINKED ISSUE #2914: Spearhead/Seedlet: Visual issues with Social Links Block in Editor
There are a few issues with the Social Links Block in the Editor in Spearhead.

Here's the block in a column inside a group.

image

Note that the individual icons are no longer wrapped in circles.

When selected all the circles stretch to form strange pill shapes …

image

### LINKED ISSUE #3015: Seedlet: Buttons in the editor are improperly spaced.
Using Seedlet, buttons in the editor appear improperly aligned. The front-end appearance is correct. 

Horizontal: 

Editor|Front End
---|---
Screen Shot 2021-01-13 at 8 04 07 AM|Screen Shot 2021-01-13 at 8 06 11 AM

Vertical:

Editor|Front End
---|---
Screen Shot 2021-01-13 at 8 04 35 AM|Screen Shot 2021-01-13 at 8 06 14 AM

jacklenox:
This PR solves #2727 by rewriting some CSS rules that are unnecessarily global, [as was done in Twenty Twenty One](https://github.com/WordPress/twentytwentyone/pull/859).

If this is a goer, we'll want to propagate this change to Seedlet's child themes.

Unlike Twenty Twenty One, I think there's just one instance where the specificity needs to be added because Seedlet has fewer editor tweaks than Twenty Twenty One. I used the same mixin approach as before anyway in case I'm wrong, and to ease adding this specific override anywhere else that needs it.

### DIFF
--- a/seedlet/assets/sass/blocks/cover/_editor.scss
+++ b/seedlet/assets/sass/blocks/cover/_editor.scss
@@ -6,6 +6,8 @@
 	margin-top: inherit;
 	margin-bottom: inherit;
 
+	@include innerblock-margin-clear(".wp-block-column");
+
 	.wp-block-cover__inner-container,
 	.wp-block-cover-image-text,
 	.wp-block-cover-text,
--- a/seedlet/assets/sass/blocks/cover/_editor.scss
+++ b/seedlet/assets/sass/blocks/cover/_editor.scss
@@ -6,8 +6,6 @@
 	margin-top: inherit;
 	margin-bottom: inherit;
 
-	@include innerblock-margin-clear(".wp-block-column");
-
 	.wp-block-cover__inner-container,
 	.wp-block-cover-image-text,
 	.wp-block-cover-text,
--- a/seedlet/assets/sass/blocks/utilities/_editor.scss
+++ b/seedlet/assets/sass/blocks/utilities/_editor.scss
@@ -149,15 +149,6 @@
 [data-block] {
 	margin-top: var(--global--spacing-vertical);
 	margin-bottom: var(--global--spacing-vertical);
-
-	[data-block]:first-child {
-		margin-top: 0;
-	}
-
-	// Needs to be the second-last child to avoid applying this to the appender.
-	[data-block]:nth-last-child(2) {
-		margin-bottom: 0;
-	}
 }
 
 /*
--- a/seedlet/assets/css/style-editor.css
+++ b/seedlet/assets/css/style-editor.css
@@ -534,23 +534,6 @@ div[data-type="core/button"] {
 	margin-bottom: inherit;
 }
 
-.wp-block-cover > .wp-block-column > *:first-child,
-.wp-block-cover-image > .wp-block-column > *:first-child {
-	margin-top: 0;
-}
-
-.wp-block-cover > .wp-block-column > *:last-child:not(.block-list-appender),
-.wp-block-cover-image > .wp-block-column > *:last-child:not(.block-list-appender) {
-	margin-bottom: 0;
-}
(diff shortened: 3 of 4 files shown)


jacklenox:
This PR also solves #3015, I just tested it. :slightly_smiling_face:

seedlet/assets/sass/blocks/cover/_editor.scss
```diff
@@ -6,6 +6,8 @@
 	margin-top: inherit;
 	margin-bottom: inherit;
 
+	@include innerblock-margin-clear(".wp-block-column");
```
jffng: I'm not sure I'm following the use case here, but this is what it looks like when I place a paragraph inside a columns block inside a cover block: 

Screen Shot 2021-01-13 at 1 42 02 PM

Should this mixin remove that top margin from the paragraph?

seedlet/assets/sass/blocks/utilities/_editor.scss
```diff
@@ -149,15 +149,6 @@
 [data-block] {
 	margin-top: var(--global--spacing-vertical);
 	margin-bottom: var(--global--spacing-vertical);
-
-	[data-block]:first-child {
```
jffng, line 153: I can confirm that removing these lines solves the social icons + buttons alignment issue in the editor.

seedlet/assets/sass/blocks/cover/_editor.scss
jacklenox: I'm not sure how well I'm following all this either, but I don't think so, and it doesn't seem to. If @kjellr wouldn't mind taking a glance at this, I'd much appreciate it.

jacklenox: Update: I've noticed that in viewing the post itself, there wouldn't be a top margin on those paragraphs, although there seem to be quite a few discrepancies when it comes to vertical spacing on the editor vs the post itself (both with, and without this PR), so I'm not sure how far we want to go in trying to rectify that as part of this PR.

jffng: Gotcha. In that case, is this mixin needed at all?

jacklenox: Haha. :thinking: Maybe not? I copied [this implementation from Twenty Twenty One](https://github.com/WordPress/twentytwentyone/blob/4c83b2781ae9cc9002ef8e0a5c275b2cea8a7f08/assets/sass/05-blocks/cover/_editor.scss) because I had assumed this was, for some reason, desirable.

I have to confess I couldn't actually work out exactly what this achieves on Twenty Twenty One either, but presumed I was missing something and we needed to keep it.

scruffian: If it's not needed let's leave it out. Keep it simple!

jacklenox: :+1: I'll make this change.

jacklenox:
PR has been amended to reflect that we're just going with dropping the global margin settings as we don't think they're needed. So I think this is now ready to land subject to a final approval.

scruffian (approved):
LGTM

PR #3020, opened 2021-01-13
Remove jQuery from Independent Publisher 2

sgomes:
Rewrites existing `jQuery` code in Independent Publisher 2 as vanila, IE11-compatible ES5.
The only exception is the customizer code, which is only used in wp-admin (which always loads `jQuery`, and isn't an important target for performance improvement).

Sites that are not making use of other `jQuery`-dependent functionality should see a nice reduction in the amount of loaded JS in end-user browsers.

- Rewrite existing "Independent Publisher 2" scripts without `jQuery`
- Remove `jQuery` from enqueue dependency list

#### Related issue(s):

See D55154-code.


josephscott:
I looked at the before and after using the `update/remove-jquery-from-independent-publisher-2` branch and confirmed that jQuery is no longer loading.  I specifically tested the condition where there is no sidebar.

Issue #2991, opened 2021-01-08
Color Annotations: Try a different approach


scruffian:
We are experiencing many problems with color annotations. Let's try using the Seedlet approach with variables to see if that fixes some of these issues:

https://github.com/Automattic/themes/issues/2895
https://github.com/Automattic/themes/issues/2896

(There are a lot more)

scruffian:
Closing in favour of #3038

PR #3025, opened 2021-01-14
Hever: css variable approach to color annotations

MaggieCabrera:

This PR implements color annotations for Hever using Seedlet's method. We are using css variables in an attempt to improve the performance of color annotations in general and to prepare this theme for Global Styles in the future.

Screen Capture on 2021-01-14 at 15-45-07(4)

To Do: 

- [x] Implement IE ponyfill to support css variables on that browser (in the same way this PR does: https://github.com/Automattic/themes/pull/2542)

#### Testing:

- Make sure that the colors are consistent between the customizer view, the frontend of the site, and the editor.
- The block color palette should reflect the changes made on the customizer (if we insert a button, the color options for Primary, Secondary, etc should match the ones we have customized for the theme instead of the default ones)
- Check that going back to the default palette has the desired effect.

#### Related issue(s):

#2991


hever/inc/customize-preview-wpcom.js
```diff
@@ -0,0 +1,103 @@
+/**
```
jffng, line 1: I think this file should be renamed to `wpcom-customize-preview.js` for consistency.

hever/functions.php
```diff
@@ -53,49 +54,67 @@ function hever_setup() {
...
-					'slug'  => 'background-dark',
-					'color' => '#C5C5C5',
+					'name'  => __( 'Background', 'hever' ),
+					'slug'  => 'background',
+					'color' => $background,
 				),
 				array(
-					'name'  => __( 'Subtle Gray', 'varia' ),
+					'name'  => __( 'Background High Contrast', 'hever' ),
 					'slug'  => 'background-light',
```
jffng: Should this be `background-dark` so it's not the same slug as the Background Low Contrast color?

```suggestion
					'slug'  => 'background-dark',
```

hever/inc/wpcom.php
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * @package Hever
+ */
+
+
+/**
+ * Bind JS handlers to instantly live-preview changes.
+ */
+function hever_wpcom_customize_preview_js() {
+	wp_enqueue_script( 'hever_wpcom_customize_preview', get_theme_file_uri( '/inc/customize-preview-wpcom.js' ), array( 'customize-preview' ), '1.0', true );
```
jffng, line 15: Just noting that if the `customize-preview-wpcom.js` filename changes, this line would change too.

hever/sass/style-child-theme-editor.scss
```diff
@@ -8,6 +8,15 @@
  */
 @import "../../varia/sass/abstracts/imports";
 
+/**
+ * Global variables
+ */
+ @import "global-variables";
+
+:root, body {
```
jffng, line 16: Is it still required by Gutenberg to target the `:root` and `body` to get variables to register in the editor?

varia/inc/color-utils.php
```diff
@@ -0,0 +1,46 @@
+<?php
```
jffng, line 1: Might make sense to call this file `wpcom-color-utils.php` for consistency. But we never submitted Varia to the themes directory so I could go either way.

jffng (commented):
This is working great for me, and lays the foundation to transition to Global Styles-based custom colors. 

I left some minor comments. Nice work @MaggieCabrera!

hever/inc/customize-preview-wpcom.js
MaggieCabrera: I went for this because varia, and other default themes use the same naming convention (https://github.com/Automattic/themes/blob/trunk/varia/inc/customize-preview-wpcom.js), it's just seedlet that does it differently

hever/sass/style-child-theme-editor.scss
MaggieCabrera: yes, they don't work unless we do it like that :( If we only target `:root `then it doesn't work on the editor, if we target both on both stylesheets, then the customizer doesn't work.

varia/inc/color-utils.php
MaggieCabrera: I didn't go that route because these functions are not wpcom specific and they can be used in other places. In fact they are used inside `functions.php` and even though Varia + children are wpcom only I wanted to keep them consistent. If they were to be released to .org, these functions would be used for the color customizer that core uses instead of the color annotation plugin

jffng: Got it!

hever/inc/customize-preview-wpcom.js
scruffian: Since a lot of this is going to be shared with other child themes can we put it as a lib in varia?

hever/inc/wpcom-editor-colors.php
```diff
@@ -0,0 +1,2 @@
+<?php
+require_once 'wpcom-colors.php';
```
scruffian, line 2: do we need to do it this way? What's the purpose of having two files?

hever/sass/style-child-theme.scss
```diff
@@ -27,71 +27,81 @@ Nicolas Gallagher and Jonathan Neal https://necolas.github.io/normalize.css/
...
+  * Child Theme Extra Styles
+  */
+ @import "extra-child-theme";
+ 
+ /**
+  * Full Site Editing
+  * - Full Site Editing overrides
+  */
+  @import "full-site-editing";
+
```
scruffian, line 107: Why does git think so much changed in this file?!!?

varia/functions.php
```diff
@@ -357,6 +357,11 @@ function varia_mobile_nav_on_side_setup() {
 	add_action( 'wp_enqueue_scripts', 'varia_add_mobile_nav_on_side_scripts' );
 }
 
+/**
+ * Color palette related utilities
+ */
+require get_template_directory() . '/inc/color-utils.php';
```
scruffian, line 455: just need to make sure this doesn't break things on Atomic

scruffian: This was the Spearhead fatal on Atomic: https://github.com/Automattic/themes/pull/2865/files

Now I wonder why I didn't just just use `get_template_directory()`. hmmm...

scruffian: So I think the problem with this is that on Atomic there is no guarantee that the parent theme is installed which means that calls to functions like `change_color_luminescence` in Hever could end up causing fatal errors on Atomic.  For full context see p1606403131274200-slack-C7YPW6K40

There are two ways round this:
1. add these to the colors plugin
2. duplicate them in every theme.

I think option 1 is better but neither are great.

scruffian: Actually [on further discussion](p1610662510061000-slack-C7YPW6K40) I think this approach will be fine, but let's just make sure to test in Atomic.

hever/inc/customize-preview-wpcom.js
MaggieCabrera: > Since a lot of this is going to be shared with other child themes can we put it as a lib in varia?

This file already exists for Varia and if we add it right now it will be adding those css variables to varia too, which may affect Varia in some way, since the css variables have the same as the ones gutenberg creates (not a 100% sure on how that would affect it). I'd say do this on a separate PR that migrates Varia to the new system too, to be sure we aren't breaking stuff and allow for the rest of the children to inherit everything as well

hever/inc/wpcom-editor-colors.php
MaggieCabrera: From what I understand, the color annotation plugin looks for bothe these files (`wpcom-colors.php` and `wpcom-editor-colors.php`) and includes them respectively. Without `wpcom-editor-colors.php` the new selected colors don't show within the editor, just the frontend and the customizer. Seedlet does this as well.

varia/functions.php
MaggieCabrera: Can confirm this works on Atomic (provided that Varia is installed too, for sure)

Screenshot 2021-01-15 at 12 39 10

hever/sass/style-child-theme.scss
MaggieCabrera: I honestly can't say, I compared to what's on trunk and the only change is the import for the global variables

hever/inc/customize-preview-wpcom.js
MaggieCabrera: I opened the issue for it and added it to the board: https://github.com/Automattic/themes/issues/3038

varia/style.css
```diff
@@ -140,6 +140,12 @@ $grid-configuration: map-extend($grid-configuration-default, $grid-configuration
  * Add font-family using CSS variables.
  * It also adds the proper fallback for browsers without support.
  */
+body {
+	display: none;
+	/* Legacy */
+	display: var(--skip, block);
```
scruffian, line 146: Should/could this go in Varia?

MaggieCabrera: it's in Varia, haha

scruffian: Oh so it is. I guess we'll need to rebuild all child themes then...

MaggieCabrera:
After adding the ponyfill we now have a working page on the frontend (the customizer is breaking on IE11 but we are going to leave that unsupported as long as the frontend shows up fine). This is ready for a final review.

**IE11:**

Screenshot 2021-01-29 at 11 14 33

hever/inc/wpcom-colors.php
```diff
@@ -0,0 +1,164 @@
...
+		--wp--preset--color--border-low-contrast: <?php echo $border_low_contrast; ?>;
+		--wp--preset--color--border-high-contrast: <?php echo $border_high_contrast; ?>;
+	}
+
+	<?php
+}
+add_theme_support( 'custom_colors_extra_css', 'hever_custom_colors_extra_css' );
+
+/**
+ * Featured Hever Palettes
```
MaggieCabrera, line 120: This file could be split and leave only the part where we define the palettes to the child theme, the rest up top can stay just on Varia once we replace the way Varia does color annotations to this method.

hever/sass/_config-child-theme-deep.scss
```diff
@@ -70,37 +70,37 @@ $config-global: (
 	/* Colors */
 	"color": (
 		"primary": (
-			"default": #1279BE,
-			"hover": #303030,
+			"default": var(--wp--preset--color--primary),
```
MaggieCabrera, line 73: It's worth looking into if this code will be repeated for every theme that we apply this method to or there will be any differences and if we can avoid that repetition.

varia/sass/abstracts/_imports.scss
```diff
@@ -6,3 +6,4 @@
 @import "functions";
 @import "config-global";
 @import "mixins";
+@import "ie11-ponyfill";
```
MaggieCabrera, line 9: I don't like having this compiled to every child theme and loaded on every browser (even if it's harmless). I think I should move this to a separate stylesheet conditionally loaded for IE only and have that on the function that has the ponyfill too

MaggieCabrera:
I keep thinking about ways to refactor this approach so it's easier to extend it from the child themes and is DRYer, it's still open for reviews but I'm going to leave it on standby for deployment because I already have ideas to improve it.

hever/inc/wpcom-colors.php
scruffian: Yeah this should definitely be refactored to Varia at some point.

MaggieCabrera:
After discussing this, we've decided to deploy this PR after review and move the refactor to a new one that encompasses Varia and the rest of the children so that the approach can be tested for Hever in isolation before we move on with the rest of the themes. This is ready for review again.

scruffian (approved):
LGTM. Let's 🚢 !

hever/inc/customize-preview-wpcom.js
```diff
@@ -0,0 +1,103 @@
...
+
+			var backgroundLowContrast = changeColorLuminescence( background, -10 );
+			var backgroundHighContrast = changeColorLuminescence( background, 10 );
+			var foregroundLowContrast = changeColorLuminescence( foreground, 10 );
+			var foregroundHighContrast = changeColorLuminescence( foreground, -10 );
+			var borderLowContrast = changeColorLuminescence( border, 10 );
+			var borderHighContrast = changeColorLuminescence( border, -10 );
+			var primaryHover = changeColorLuminescence( primary, 10 );
+			var secondaryHover = changeColorLuminescence( secondary, 10 );
+			const extraCSS = ':root {' +
```
simison, line 75: Is this meant to be `const`, while rest of the file has ES5 `var`?

It'll break on IE11.

hever/inc/wpcom-colors.php
```diff
@@ -0,0 +1,164 @@
...
+	add_color_rule(
+		'bg',
+		$colors['background'],
+		array(
+
+			// This placeholder is needed to make the color annotations work
+			array( '.wp--preset--color--background', 'background-color' ),
+
+		),
+		__( 'Background Color' )
```
simison, line 16: Is it expected not to have translation-text-domain on this and few other `__()`s in this file?

PR #3026, opened 2021-01-14
Change CSS variable assignment order

### LINKED ISSUE #2701: Spearhead: Custom colors not working

pbking:

First Seedlet was changed so that all custom CSS variables for color
definitions are associated with the 'seedlet-custom-color-overrides'
stylesheed.  (Previously only editor override styles were associated
with that stylesheet.

Next Spearhead was changed to first dequeue that stylesheet, THEN assign
the spearhead-specific style variables THEN re-enqueue that
seedlet-custom-color-overrides stylesheet.

This puts the customizer color variable assignments AFTER the Spearhead
variable assignments so that those are the values that are used.

NOTE: [This Pull Request](https://github.com/Automattic/themes/pull/2776) also seems to be addressing this issue.  However there were quite a few more changes in that PR.  If those additional changes are necessary I don't yet grasp all of the reasons.  That work should be reviewed against this proposed change and perhaps that's a better fix. My 'druthers is that since that large amount of changes appears to address an edge-case that it instead be tagged as a new issue and instead play golf with this tiny change.

#### Testing

This has been tested by using the Customizer to change the colors to something recognizable and observing that those colors are shown correctly in: 

The Customizer
image

The Editor
image

The View
image

It was confirmed that this change doesn't effect the WP-COM behavior which utilizes another color customizer utility.

image

It was confirmed that this change doesn't change the behavior of Spearhead which was already working as expected.

#### Related issue(s):
Closes #2701

### DIFF
--- a/seedlet/classes/class-seedlet-custom-colors.php
+++ b/seedlet/classes/class-seedlet-custom-colors.php
@@ -262,8 +262,9 @@ class Seedlet_Custom_Colors {
 	 * Customizer & frontend custom color variables.
 	 */
 	function seedlet_custom_color_variables() {
+		wp_enqueue_style( 'seedlet-custom-color-overrides', get_template_directory_uri() . '/assets/css/custom-color-overrides.css', array(), wp_get_theme()->get( 'Version' ) );
 		if ( 'default' !== get_theme_mod( 'custom_colors_active' ) ) {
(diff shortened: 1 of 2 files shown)


danieldudzic (approved):
Works fine for me!

scruffian:
LGTM

PR #3027, opened 2021-01-14
Seedlet: Fix Button block alignment editor issue

### LINKED ISSUE #2722: Seedlet: Multiple buttons can't sit on the same line
If you activate Seedlet and then add the buttons block you should see the issue:
image

danieldudzic:
Closes #2722

Before:
before
After:
after


jffng (approved):
Just noting that this is no longer an issue on wpcom or using the latest version of Gutenberg.

Still worth shipping this for those that don't have the plugin installed.

PR #3044, opened 2021-01-15
Spearhead: Fix the contrast issue for mobile menu links.

### LINKED ISSUE #2819: Spearhead: Dark Settings on mobile shows site's text color as dark over dark background
## Steps to replicate

- activate Spearhead theme 
- in your mobile device, enable Dark appearance then visit the site (in mobile)
- then click menu and the texts will be dark over dark background: 

Screen Shot 2020-11-15 at 5 14 24 PM

For comparison, here's how Light appearance in mobile settings display the menu text color: 

Screen Shot 2020-11-15 at 5 20 27 PM

## Result

Selecting dark appearance in mobile device and visiting the site will show the menu text color in dark color over dark background, making the texts hard to read. 

## Expected

Dark appearance in mobile device setting should set the background dark, but the menu text color should be white. 

* User report: 3490288-zd
* Follow-up ticket: 3490288-zd

danieldudzic:
Closes #2819

### DIFF
--- a/spearhead/inc/wpcom-colors-utils.php
+++ b/spearhead/inc/wpcom-colors-utils.php
@@ -25,7 +25,6 @@ function seedlet_define_color_annotations( $colors ) {
 
 			// This placeholder is needed to make the color annotations work
 			array( '.global--color-foreground', 'color' ),
-			array( '.site-header .primary-navigation .primary-menu-container a', 'color', '#2d3139' ),
 
 		),
 		__( 'Foreground Color' )
@@ -132,6 +131,7 @@ function seedlet_custom_colors_extra_css() {
 	$secondary    = $colors_array['colors']['fg1'];
 	$tertiary     = $colors_array['colors']['fg2'];
 
+	$background_low_contrast  = change_color_luminescence( $background, -10 );
 	$foreground_low_contrast  = change_color_luminescence( $foreground, 10 );
 	$foreground_high_contrast = change_color_luminescence( $foreground, -10 );
 	$primary_hover            = change_color_luminescence( $primary, 10 );
@@ -149,6 +149,7 @@ function seedlet_custom_colors_extra_css() {
 		--global--color-secondary: <?php echo $secondary; ?>;
 		--global--color-secondary-hover: <?php echo $secondary_hover; ?>;
 		--global--color-tertiary: <?php echo $tertiary; ?>;
+		--primary-nav--color-background: <?php echo $background_low_contrast; ?>;
 	}
 
 	<?php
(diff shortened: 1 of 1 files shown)


danieldudzic:
I am not 100% sure, this is the correct way to handle color annotations in our new themes. 

Guidance much appreciated 🙏

scruffian:
This is not how I'd do it. In Seedlet based themes color annotations are all done via CSS variables. The color annotation only does one thing - output new variables. So in this case you'd need to look at what variables are being targetted and see why they aren't changing...

danieldudzic:
@scruffian yes, that's what I figured - in that case though, my question is how does Seedlet handle contrast calculation that originally was done via standard color annotations?

scruffian:
For Spearhead the colors are defined here: https://github.com/Automattic/themes/blob/0100e61255bc8175d354426db440006b75e4eb2c/spearhead/inc/wpcom-colors-utils.php#L126

danieldudzic:
cc: @jffng

scruffian:
What do you think about fixing this by only applying the dark mode variables when custom colors are not set?

e.g.

```
@media ( prefers-color-scheme: dark ) {
	:root :not(.custom-colors) {
		...
	}	
}
```

jffng:
> What do you think about fixing this by only applying the dark mode variables when custom colors are not set?

That could work, it's also consistent with how we inform the user about dark mode in the customer. I didn't know that `.custom-colors` was output to the markup. 

@danieldudzic and I paired on this and tried to address this by taking the custom color palette's background color, generating a lower contrast version of it, and applying it to the CSS variable responsible for setting the mobile menu's background color. We felt this preserved Spearhead's design intent a bit better to use a different but related background color for the mobile menu.

We ran into an issue — the [`change_color_luminescence`](https://github.com/Automattic/themes/blob/0100e61255bc8175d354426db440006b75e4eb2c/spearhead/inc/wpcom-colors-utils.php#L115) function kept outputting grayscale values, getting caught on this condition:  

https://github.com/Automattic/themes/blob/0100e61255bc8175d354426db440006b75e4eb2c/spearhead/inc/wpcom-colors-utils.php#L100-L102

In any case, @scruffian's solution seems simpler if possible.

scruffian:
My thought was simply that we disable dark mode if you have a custom colour palette. There are a few issues with it though, but they might be simpler to fix...

danieldudzic:
Unless there's an option to select 2 custom color palettes for the Light and the Dark modes separately, I think disabling the Dark Mode (when custom palette is present) makes sense.

danieldudzic:
`.custom-colors` seems to be missing from the front end.

There's `.customizer-styles-applied`- but that's present even without a custom color palette selected.

scruffian:
I see it on https://scruffiansimple.wordpress.com/

danieldudzic:
It's missing for me, here: https://danieldudzic.wordpress.com/

scruffian:
Interesting. It might be worth investigating what's going on with that...

scruffian:
We discussed this on Slack and decided that the best way forward would be to add a class to the body when a site has selected custom colors. Then we can use that class to remove the dark mode when color palettes are active.

Issue #3043, opened 2021-01-15
Seedlet/Spearhead: IE11 error


scruffian:
Something has gone wrong with our CSS Variables pony fill in IE11 - neither Seedlet or Spearhead load in IE11.

MaggieCabrera:
We've discussed this here: p1610960441157500-slack-C029FM1EH

This is only happening on production and not on our sandboxes.

There's currently 2 types of errors showing up for production. An issue with backticks on the primary navigation file (doesn't seem to be related to this but I'm pushing a PR to remove them anyway) and some CORS issues that seem to be causing the ponyfill to not work with the css on production. This issue seems relevant to the case: https://github.com/jhildenbiddle/css-vars-ponyfill/issues/62

<img width="1661" alt="Screenshot 2021-01-18 at 11 14 25" src="https://user-images.githubusercontent.com/3593343/104905415-98da5c80-5982-11eb-9e17-b0d28d722b86.png">

MaggieCabrera:
Pending solving the CORS issue p3topS-Rh-p2

MaggieCabrera:
After fixing the CORS issue, Seedlet now works on IE, Spearhead needs to have the same change for the backticks apparently.

scruffian:
This was fixed by #3147

PR #3053, opened 2021-01-18
Seedlet: Fix IE11 issues

MaggieCabrera:

This PR addresses two errors that appear on IE11 for Seedlet:
- Fixes the invalid character error by removing the backticks
- Adds crossorigin="anonymous" to all CSS tags in an attempt to fix the CORS issue that currently is only affecting production

#### Related issue(s):

#3043

### DIFF
--- a/seedlet/functions.php
+++ b/seedlet/functions.php
@@ -483,3 +483,11 @@ if ( class_exists( 'WooCommerce' ) ) {
  * Load Jetpack compatibility file.
  */
 require get_template_directory() . '/inc/jetpack.php';
+
+/**
+ * Adds crossorigin anonymous to all css scripts so that the IE11 ponyfill doesn't have CORS issues
+ */
+function seedlet_css_crossorigin( $html, $handle ) {
+    return str_replace( "media='all'", "media='all' crossorigin='anonymous'", $html );
+}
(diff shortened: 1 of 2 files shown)


MaggieCabrera:
D55506-code -> diff

seedlet/functions.php
```diff
@@ -483,3 +483,11 @@ function seedlet_author_bio() {
  * Load Jetpack compatibility file.
  */
 require get_template_directory() . '/inc/jetpack.php';
+
+/**
+ * Adds crossorigin anonymous to all css scripts so that the IE11 ponyfill doesn't have CORS issues
+ */
+function seedlet_css_crossorigin( $html, $handle ) {
+    return str_replace( "media='all'", "media='all' crossorigin='anonymous'", $html );
```
scruffian: Should we only do this in IE?

scruffian (approved):
LGTM. I guess we have to test on production...

seedlet/functions.php
MaggieCabrera: yes, that makes sense

scruffian:
🚢

MaggieCabrera:
I removed the crossorigin function since that doesn't work but left the backtick fix for IE. This was already on prod, merging this so the reverted function gets to .com with the next batch of fixes for seedlet

PR #3055, opened 2021-01-18
Independent Publisher 2: fix outdent for large images

### LINKED ISSUE #3048: Independent Publisher 2: Images are no longer wide since version 2.1.10
Since the latest update to version 2.1.10, the images added on the editor are missing the class `size-big`. This is a breaking change for users who chose this theme for its wider image width.

## Steps to replicate

* Go to Editor, switch to Classic Block and add an image.
* The image should be having class `size-big` that made it a little wider than normal.

The latest commit ac4759d shows that code related to the addition of class `size-big` was removed.

## Result

Images are no longer wider.

## Expected

Screenshot from web.archive.org on how it used to look: 

* User report: #24125801-hc
* Follow-up ticket:  #3654498-zen

sgomes:
- Independent Publisher 2: Fix image sizing bug introduced in #3020

#### Related issue(s):
This should fix #3048. Please test following the instructions on that issue,

### DIFF
--- a/independent-publisher-2/functions.php
+++ b/independent-publisher-2/functions.php
@@ -213,7 +213,7 @@ function independent_publisher_2_scripts() {
 	wp_enqueue_script( 'independent-publisher-2-navigation', get_template_directory_uri() . '/js/navigation.js', array(), '20170317', true );
 
 	if ( ! is_active_sidebar( 'sidebar-1' ) ) {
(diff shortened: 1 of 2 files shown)


pbking (approved):
Fixes the issue.  No concerns.

sgomes:
Thank you for the review, @pbking! 👍

PR #3056, opened 2021-01-18
Seedlet: Remove word-break style in tables

### LINKED ISSUE #2120: Varia: words get cut randomly in table block
This happens to Varia and child themes.

## Steps to replicate

* Add a table block with some content, say 1 row, 3 cells.
* Publish, and see how words get cut.

Note that using the fixed-cell feature prevents this from happening.

## Result

Words get cut randomly at the end of the cell.

Screen Shot 2020-06-12 at 17 13 15

## Expected

Instead of getting cut, words should go to the next line.

It was reported in #3037564-zd and I was able to reproduce it.

### LINKED ISSUE #2375: Shawburn: table blocks aren't wrapping correctly
## Steps to replicate
1. Create a table block with fixed column widths and long, wrapping lines
2. Visit the page in IE11

I believe this is due to inconsistent browser implementation of `word-break: break-word`, which is deprecated:
https://www.w3.org/TR/css-text-3/#valdef-word-break-break-word
https://caniuse.com/#feat=word-break

## Result
Text seems to `break-all`:

image

## Expected
image

As a bonus, if the table block is set to NOT have fixed column widths, it does `break-all` on all browsers. Not sure if that's intended behavior or not.

* User report: 23497642-hc
* Follow-up ticket: 3243763-zen

### LINKED ISSUE #2410: Seedlet: WooCommerce Variable Product Styles: "Clear" button text breaks to next line
## Steps to replicate

* Create a variable product
* Make sure one attribute for variation is at least 13 characters (important—in the below example, "Tanning Color" is 13 characters)
* Check product page

## Result

"Clear" link/button to clear variation selections breaks to next line

## Expected

"Clear" button stays on same line, or line break happens before the button

Screenshot
*Link to image: https://d.pr/i/ifZduu*

jacklenox:
The `word-break: break-all` style creates unwanted and unexpected breaks in text in tables. Let's remove it.

Fixes #2410.

### DIFF
--- a/seedlet/assets/sass/blocks/table/_style.scss
+++ b/seedlet/assets/sass/blocks/table/_style.scss
@@ -12,6 +12,5 @@ table,
 	th {
 		padding: calc( 0.5 * var(--global--spacing-unit) );
 		border: 1px solid;
-		word-break: break-all;
 	}
 }


jffng:
Sounds good.

> Fixes #2410, fixes #2375, fixes #2120.

This PR only makes changes to Seedlet. Do you intend to address the Varia child theme changes in a follow-up PR?

jacklenox:
Good point! I've reduced the "fixes" to just the Seedlet one. I'll do the Varia ones in another PR now.

Issue #2898, opened 2020-12-08
Spearhead: Jetpack Related Posts on Atomic are aligned to the left


AtrumGeost:
## Steps to replicate

1. Activate Spearhead on an AT site
2. Enable related posts by clicking on **My Site > Marketing > Traffic**
3. Access a blog post

## Result

The Related Posts are aligned all the way to the left:

![Screen Shot on 2020-12-08 at 11:58:22](https://user-images.githubusercontent.com/3696121/101515771-b8767200-394c-11eb-9559-c9b24a15a222.png)

## Expected

For the Related Posts to be centered

* User report: 26021739-hc

Provided a temporary CSS fix:

`#jp-relatedposts {
  margin: 0 auto;
}`

danieldudzic:
This is happening in (almost) all of our Gutenberg themes due to the following css coming from Jetpack:

```
#jp-relatedposts {
	display: none;
	padding-top: 1em;
	margin: 1em 0;
	position: relative;
	clear: both;
}
```

TT1 has a fix for this (in a Jetpack compat file: https://github.com/Automattic/jetpack/blob/master/projects/plugins/jetpack/modules/theme-tools/compat/twentytwentyone.css
):
```
/**
 * Related Posts
 */

.entry-content #jp-relatedposts {
	max-width: var(--responsive--aligndefault-width);
	margin-left: auto;
	margin-right: auto;
}
```

I will add the above fix in Varia & Seedlet and also [report the issue in Jetpack](https://github.com/Automattic/jetpack/issues/18440).

danieldudzic:
This issue has been moved to the Jetpack repo: https://github.com/Automattic/jetpack/issues/18440

PR #3063, opened 2021-01-19
Seedlet: Fix the alignment of Related Posts

danieldudzic:
Addresses the #2898 issue in Seedlet

fixed


scruffian:
Do you think this fix should live in the block itself?

danieldudzic:
> Do you think this fix should live in the block itself?

In Jetpack, yes: https://github.com/Automattic/jetpack/issues/18440

scruffian:
Oh nice one, thanks. I suggest we don't do the temporary fixes, and just fix it directly in the block..

PR #3064, opened 2021-01-19
Spearhead: add styles for product page

jacklenox:

On a product page:

- [x] Changed the colour of the sale price
- [x] Changed the alignment of the navigation
- [x] Changed the sizing and appearance of the item quantity and the "Add to cart" button
- [x] Changed the colour of the category and SKU text
- [x] Changed the colour of the active tab text
- [x] Changed the font of the description

#### Related issue(s):

Partially addresses: #2715


jffng (commented):
My assumption is that the design should match [this comp](https://user-images.githubusercontent.com/1202812/97455004-be207700-190d-11eb-861d-05259363ab6f.png) as closely as possible: 

Before | After
------- | -------
Screen Shot 2021-01-25 at 1 49 44 PM | Screen Shot 2021-01-25 at 1 49 51 PM

A few design notes:

- The sale marker should be a circle. 
- The quantity is not centered in the input box.
- Both inputs (quantity and Add to Cart) have more vertical padding in the comp. 
- The price seems really _big_  

cc @kjellr

MaggieCabrera:
Thanks for your review @jffng !

 
>     * The sale marker should be a circle.

Initially it was, but on the product list page we decided to go for a square version: https://github.com/Automattic/themes/pull/2745#issuecomment-719621845

>     * The quantity is not centered in the input box.

it's not because `<input type="number"/>` adds arrows inside the input when you are selecting a quantity. I believe we have the same on Seedlet


>     * Both inputs (quantity and Add to Cart) have more vertical padding in the comp.

You are right, I'll change that

>     * The price seems really _big_

I agree with that. Hopefully, @kjellr can tell us how small we can make it?

scruffian (approved):
LGTM. Do we need to adjust the size of the price?

kjellr:
> Do we need to adjust the size of the price?

Yeah, let's use `var(--global--font-size-lg)` here instead of `var(--global--font-size-xl)`. Thanks!

MaggieCabrera:
This should be ready for another review

scruffian (approved):
LGTM

PR #3065, opened 2021-01-19
Varia: Remove word-break style in tables

### LINKED ISSUE #2120: Varia: words get cut randomly in table block

### LINKED ISSUE #2375: Shawburn: table blocks aren't wrapping correctly

jacklenox:
Related to #3056.

The `word-break: break-all` style creates unwanted and unexpected breaks in text in tables. Let's remove it.

Fixes #2375, fixes #2120.

### DIFF
--- a/alves/style-rtl.css
+++ b/alves/style-rtl.css
@@ -2321,7 +2321,6 @@ table th,
 .wp-block-table th {
 	padding: calc( 0.5 * 16px);
 	border: 1px solid;
-	word-break: break-all;
 }
 
 table.is-style-stripes tbody tr:nth-child(odd),
--- a/alves/style.css
+++ b/alves/style.css
@@ -2321,7 +2321,6 @@ table th,
 .wp-block-table th {
 	padding: calc( 0.5 * 16px);
 	border: 1px solid;
-	word-break: break-all;
 }
 
 table.is-style-stripes tbody tr:nth-child(odd),
--- a/balasana/style-rtl.css
+++ b/balasana/style-rtl.css
@@ -2321,7 +2321,6 @@ table th,
 .wp-block-table th {
 	padding: calc( 0.5 * 16px);
 	border: 1px solid;
-	word-break: break-all;
 }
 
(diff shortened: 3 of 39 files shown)


alves/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "alves",
-  "version": "1.4.14",
+  "version": "1.4.15",
```
MaggieCabrera: I think we don't want to update the version numbers just yet, since there are more changes pending to be deployed to Varia children, we would update them once all are ready for deploy

MaggieCabrera:
Besides the issue with the version numbers, this LGTM

jacklenox:
> Besides the issue with the version numbers, this LGTM

Thanks @MaggieCabrera. Agh, yeah. The script does that automatically. I'm trying to think of a clever way of undoing those version numbers. I think I know how.

jacklenox:
(This was exactly the opposite of what I wanted to do. New commit incoming.)

PR #3066, opened 2021-01-20
Varia: Fix the alignment of Related Posts

danieldudzic:
Addresses the #2898 issue in Varia:

fixed


scruffian:
I suggest we fix this directly in the block.

PR #3071, opened 2021-01-20
Multiple Themes: Clean up top margin when header & homepage title are hidden

kjellr:
Followup from #2961. 

When both the "Hide Header" and "Hide Homepage Title" settings are activated, many of Varia's child themes needed some slight adjustments to ensure: 

- That the first child block of `entry-content` had top margin, and didn't bump up against the top of the screen. 
- That certain full-width blocks _do_ properly bump up against the top of the screen. 

The fix was generally just to:

- Add a rule to Varia to ensure that the top margin is present by default. 
- Recompile CSS for all of the child themes.
- Add special case margins for themes that use non-standard margins (Shawburn, Rockfield, Morden, Redhill).

## Screenshots

Generally, this is what things looked like before, using Barnsbury (on the left) and Stow (on the right) as examples: 

### Before

Full-width Media/Text Block|Paragraph
---|---
Screen Shot 2021-01-20 at 11 44 53 AM|Screen Shot 2021-01-20 at 11 48 02 AM

### After

Full-width Media/Text Block|Paragraph
---|---
Screen Shot 2021-01-20 at 11 39 52 AM|Screen Shot 2021-01-20 at 11 48 25 AM

### DIFF
--- a/alves/style-rtl.css
+++ b/alves/style-rtl.css
@@ -3169,6 +3169,22 @@ body:not(.fse-enabled) .footer-menu a {
 	overflow: scroll;
 }
 
+.home.page.hide-homepage-header.hide-homepage-title .entry-content {
+	margin-top: 0;
+	padding-top: 0;
+}
+
+.home.page.hide-homepage-header.hide-homepage-title .entry-content > *:first-child {
+	margin-top: 32px;
+}
+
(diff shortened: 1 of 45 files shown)


scruffian (approved):
LGTM

PR #3075, opened 2021-01-20
Varia: fix customizer on wpcom

jffng:
This PR fixes the customizer fatals on Varia themes on wpcom.

It removes the duplicate sanitization function definition and pulls it out so the wpcom function can use it also.

Side note — php linting and auto-formatting ran on these files.

### DIFF
--- a/varia/inc/wpcom.php
+++ b/varia/inc/wpcom.php
@@ -16,9 +16,12 @@ function varia_wpcom_setup() {
 	global $themecolors;
 
 	// Disable automatically generated color palettes.
-	add_theme_support( 'wpcom-colors', array(
-	        'only-featured-palettes' => true,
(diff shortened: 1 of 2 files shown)


varia/inc/wpcom.php
```diff
@@ -168,7 +165,7 @@ function varia_wpcom_editor_scripts() {
 /**
  * Enqueue CSS for customizer a11y warning.
  */
-function varia_enqueue_message_scripts() {
+function varia_wpcom_enqueue_message_scripts() {
```
jffng, line 168: Let's namespace this function more clearly if its specific to wpcom.

MaggieCabrera (approved):
LGTM, tested on wpcom and it gives me no problems

PR #3076, opened 2021-01-20
Improved sandbox script to delete appropriately

pbking:
This change collects a list of what files have been deleted in the
repository and removes them from the sandbox via rm.

If those files have since returned (deleted but returned) they will
still be deleted but, since those files will be in your local copy will
be synced to the sandbox.

Also this change is determining which files that have been deleted locally but NOT as a committed change to the repository and removing those files from the sandbox.

### DIFF
--- a/sandbox.sh
+++ b/sandbox.sh
@@ -82,11 +82,11 @@ How do you wish to proceed? [1]"
 
       # These are the files changed in THIS branch since it diverged
       # EVEN IF they changed in trunk we STILL want to include them in our sync
-      comitted_files_to_include=$(git diff ${hash2} ${current_branch} --name-only)
-      comitted_files_to_include=($comitted_files_to_include)
+      committed_files_to_include=$(git diff ${hash2} ${current_branch} --name-only)
+      committed_files_to_include=($committed_files_to_include)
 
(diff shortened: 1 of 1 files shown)


sandbox.sh
```diff
@@ -125,6 +125,13 @@ How do you wish to proceed? [1]"
 
   fi
 
+  # Determine which files have been removed from the repository (as of this head, within the past month) according to git
+  files_to_delete=$(git log --format=format:"" --name-only --since="last month" --diff-filter=D HEAD)
+
+  # Remove all of those files from the destination.  (Note, if a file has since been re-added to the repo it will be uploaded in the next step)
+  rmcmd="ssh $SANDBOX_USER@$SANDBOX_LOCATION 'cd $SANDBOX_PUBLIC_THEMES_FOLDER && rm -f $files_to_delete'"
+  eval $rmcmd;
+
```
MaggieCabrera: I know you are not done but this worked a treat to fix the issues I had yesterday. Thank you

jffng (approved):
I pushed a spelling change but this works in my testing!

What I did to test: 

- Removed a file tracked by git and svn
- I did _not_ commit that deletion
- Ran `./sandbox.sh push` and went with option 2
- Verified the file was removed on my sandbox

PR #3079, opened 2021-01-21
Blank Canvas: Change level of Invitation pattern heading

kjellr:
This PR includes a slight change to the Invitation pattern bundled with Blank Canvas: It updates the title from an H2 to an H1. This is more semantic, and also just looks better anyway. 

Before

Screen Shot 2021-01-21 at 9 08 38 AM

After

Screen Shot 2021-01-21 at 9 08 10 AM


jffng (approved):
🚢

PR #3086, opened 2021-01-22
Varia: Fix gap that appears below the header

MaggieCabrera:

This PR reverts the change done in this commit https://github.com/Automattic/themes/commit/c6ff72af3e6938aadce5aa971c7ff0c24ccb1435 that is causing a gap below the header on multiple varia children. Sometimes it's more noticeable like on Rockfield because it separates the fixed header from the cover block, in other themes it simply adds a larger separation. 

#### Related issue(s):

https://github.com/Automattic/wp-calypso/issues/49190#event-4238742550

### DIFF
--- a/varia/inc/style-wpcom.css
+++ b/varia/inc/style-wpcom.css
@@ -9,6 +9,10 @@
 	display: none;
 }
 
+.home.page.hide-homepage-title .site-main > article > .entry-content {
+	margin-top: 0;
+}
+
 /**
  * Fix Direct Manipulation icons in the Customizer
  */


kjellr:
Looks great in:

Alves
Balasana
Barnsbury
Brompton
Coutoire
Dalston
Exford
Hever
Leven
Maywood
Mayland
Morden
Rivington
Rockfield
Shawburn
Stow

I also tested in Redhill, and noticed an issue there: 

From [the theme screenshot](https://wordpress.com/theme/redhill/kjellreigstad.wordpress.com), it appears it's supposed to have more padding in the top than it does with this PR active: 

Screen Shot 2021-01-22 at 10 35 29 AM

MaggieCabrera:
I checked Redhill on the Wayback Machine and it looks like it does today though: 

Screenshot 2021-01-22 at 16 41 57

kjellr:
Hmm weird. I guess that's ok for now then.

I noticed an issue with Stratford though: it still has the empty space for me. Can you confirm?

Screen Shot 2021-01-22 at 10 48 47 AM

MaggieCabrera:
I can, but weirdly enough the wayback machine shows me the same thing..:
Screenshot 2021-01-22 at 16 56 58

kjellr (approved):
Hmm ok. Well I've tested all the other themes, and this works fine for them! I think this is good to go. 🎉

MaggieCabrera:
Thanks @kjellr I'm deploying this then with D55860-code

PR #3088, opened 2021-01-22
Seedlet: Fix button colors

kjellr:
This fixes a couple button color issues I saw in Seedlet: 

1. Outlined buttons with a custom text color were not appearing correctly in the front-end. 
2. When inside of a block that contained a background color (Cover, Group), all buttons were using the `currentColor` for the text color. This often led to unreadable buttons. 

Before|After
---|---
Screen Shot 2021-01-22 at 3 07 27 PM|Screen Shot 2021-01-22 at 3 06 33 PM

### DIFF
--- a/seedlet/assets/sass/blocks/button/_style.scss
+++ b/seedlet/assets/sass/blocks/button/_style.scss
@@ -39,6 +39,10 @@ input[type="submit"],
 			&.has-focus {
 				color: var(--button--color-background-hover);
 			}
+
+			&.has-text-color {
+				color: inherit;
+			}
 		}
 	}
 
--- a/seedlet/assets/sass/blocks/button/_style.scss
+++ b/seedlet/assets/sass/blocks/button/_style.scss
@@ -25,10 +25,13 @@ input[type="submit"],
 
 		&.wp-block-button__link,
 		.wp-block-button__link {
-			color: var(--button--color-background);
 			background: transparent;
 			border: var(--button--border-width) solid currentcolor;
 			padding: var(--button--padding-vertical) var(--button--padding-horizontal);
+
+			&:not(.has-text-color) {
+				color: var(--button--color-background);
+			}
 
 			&:active {
 				color: var(--button--color-background);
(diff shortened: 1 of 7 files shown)


seedlet/assets/sass/blocks/cover/_editor.scss
```diff
@@ -12,7 +12,7 @@
 	.block-editor-block-list__block {
 		color: currentColor; // uses text color specified with background-color options in /blocks/utilities/_style.scss
 
-		a:not(.has-text-color) {
+		a:not(.has-text-color):not(.wp-block-button__link) {
```
kjellr, line 15: This rule isn't strictly necessary since Gutenberg's inline color rules will override it. But I think it's worth having to keep in sync with the front end (and to prepare for if/when Gutenberg drops that inlining in the editor).

seedlet/assets/sass/blocks/button/_style.scss
```diff
@@ -39,6 +39,10 @@ input[type="submit"],
 			&.has-focus {
 				color: var(--button--color-background-hover);
 			}
+
+			&.has-text-color {
+				color: inherit;
```
jffng: I'm not sure this is the correct fix. 

The problem is that when a preset custom text color is selected, the class `has-foreground-color[class]` is overridden by the button outline styles: 

Screen Shot 2021-01-25 at 12 27 24 PM

If I'm understanding correctly, setting to inherit just sets the link to the wrapping div's default color — which in Seedlet's case also happens to be `var(--global--color-foreground)`.

kjellr: Good catch — I just pushed a fix that only applies the color rule if the block doesn't include a `has-text-color` class. Mind retesting?

jffng: That makes sense and looks good in my testing, thanks!

jffng:
This LGTM, thanks @kjellr !

PR #3097, opened 2021-01-25
Blank Canvas: Improve block patterns HTML tags

kjellr:
This PR just makes a few of the bundled block pattern tags more semantic. Since these are all meant for single pages, they should all use H1s for their header text. This PR makes that change, and adjusts the font size down to what it was before so that there's no change in visual appearance. 

- About Me Pattern: Changes the `<h2>` to a `<h1>`
- Links Pattern: Changes the `<p>` header to a `<h1>`.
- Card Pattern: Changes the `<h2>` to a `<h1>`
- Email Signup Pattern: Changes the `<h2>` to a `<h1>`


scruffian (approved):
LGTM

PR #3098, opened 2021-01-25
Blank Canvas: Update theme description

kjellr:
Updates the theme description to match what we have on [the WordPress.com theme landing page](https://wordpress.com/theme/blank-canvas).

Also updates a textdomain while I was in there. Though that function comes from Seedlet, themes should _technically_ only have one textdomain in use (as per the Theme Check plugin).


danieldudzic (approved):
Looks legit!

Issue #2618, opened 2020-10-16
Seedlet, Varia child themes: Extra padding added to inner blocks in Media & Text block


iamtakashi:
There are extra paddings added to `block-editor-inner-blocks` in Media & Text block that make it misaligned in the editor. This issue makes some patterns don't look right in the preview as well. It displays expectedly on the front of the site.

**Varia themes**
![iamtakashi eu ngrok io_wp-admin_post php_post=106 action=edit](https://user-images.githubusercontent.com/908665/96258802-7a944780-0fb4-11eb-8be4-065959a7ec0f.png)

**Seedlet**
![iamtakashi eu ngrok io_wp-admin_post php_post=106 action=edit (2)](https://user-images.githubusercontent.com/908665/96258780-6f411c00-0fb4-11eb-9f54-bcb9d71f3d95.png)

Twenty Twenty has no issue.
![iamtakashi eu ngrok io_wp-admin_post php_post=106 action=edit (1)](https://user-images.githubusercontent.com/908665/96258845-8bdd5400-0fb4-11eb-8221-5ae159c38861.png)

pbking:
Added two PR's (#3099 for Seedlet and #3100 for Varia and Children)

If the classes removed in those PRs served a higher purpose I wasn't able to discern it.

PR #3099, opened 2021-01-25
Remove unnecessary h-padding styles

pbking:

The editor-only horizontal padding for the block element doesn't appear
to be helpful. Removing these styles allows all elements (including
those blocks that add .block-editor-inner-blocks when in the editor) to
lay out as expected.

Note that neither Spearhead or Blank Canvas suffer from the same.

#### Related issue(s):

This will fix the seedlet portion of #2618. A separate PR was opened for [Varia (and children)](https://github.com/Automattic/themes/pull/3100) for the same.


jffng (approved):
🚢

PR #3105, opened 2021-01-26
Spearhead: Use the background color for sticky posts and menus

scruffian:
Some parts of the Spearhead design rely on a slightly different background color. In the color annotations we also need to override these colors. You can see this on the menu and when you have no sticky posts.

Before:
Screenshot 2021-01-26 at 12 22 43

After:
Screenshot 2021-01-26 at 12 18 54

This is an iteration on https://github.com/Automattic/themes/pull/3044

#### Related issue(s):
#2819

### DIFF
--- a/seedlet/inc/wpcom-customize-preview.js
+++ b/seedlet/inc/wpcom-customize-preview.js
@@ -70,6 +70,7 @@ function changeColorLuminescence( hex, amount ) {
 			var secondaryHover = changeColorLuminescence( secondary, 10 );
 			const extraCSS = ':root {' +
 					'--global--color-background: ' + background + ';' +
+					'--global--color-background-high-contrast: ' + background + ';' +
 					'--global--color-foreground: ' + foreground + ';' +
 					'--global--color-foreground-low-contrast: ' + foregroundLowContrast + ';' +
 					'--global--color-foreground-high-contrast: ' + foregroundHighContrast + ';' +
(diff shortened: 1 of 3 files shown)


jffng:
Here are the use cases I tested on a simple site with Spearhead and the results:

Mode | Default palette | Featured Palette A | Featured Palette B | Featured Palette C
-------- | -------- | --------- | -------- | ---------
Normal / light mode | ✅ | ✅ | ✅ | ❌ Font color is unreadable when sticky post is present
Dark mode  | ✅ | ❌ Mobile menu font color is unreadable (original issue) |  ❌ Mobile menu font color is unreadable (original issue) | ✅ 

From the fonts in your screenshots it looks like you're testing Seedlet? Should I test Seedlet also?

scruffian:
Thanks for testing. I've now addressed all of these issues. I was testing in Spearhead, I just have custom fonts on...

jffng (approved):
This works for me.

I noticed in dark mode — the mobile menu background color + sticky post background color is the same as the main background color for all featured palettes. I think this is an acceptable compromise. For example:  

Screen Shot 2021-01-26 at 10 29 28 AM

_Edit_
Haha my above "observation" was the title of this PR, so I would assume it was intentional 😄 🚢

scruffian:
Yeah, it was the simplest way and a fair compromise I thought :)

PR #2886, opened 2020-12-03
Spearhead: address remaining TRT feedback

jffng:
This PR addresses the outstanding to-dos in #2875:

- [x] `functions.php` L226 Please add: `if ( is_admin() ) return $more;` 
- [x] Links within content must be underlined
- [x] Mobile menu: links in it are unreadable on focus, black on black (awaiting design feedback here)

PR #3107, opened 2021-01-26
Don't style social links

scruffian:
Excludes social links from the anchor styling


pbking:
```
.has-link-color a, a 
...
       &:hover:not(.wp-block-button__link):not(.wp-block-file__button) {
```
transpiles to the very specific
```
...a:hover:not(.wp-block-button__link):not(.wp-block-file__button)
```

Which is going to match all anchors other than those that have wp-block-button__link and wp-block-file__button.  Is that level of specificity necessary for that styling?

Further, the styles in that selection seem to be attempting to style anchors differently than the block below it.  [This change](https://github.com/Automattic/themes/pull/2886) seems to have inverted the underline styling for regular/focus from the original intent so I'm not sure that block is valid any more.

I feel like this could be fixed by being less specific rather than more...

scruffian:
@pbking feel free to create a PR if you have an idea how to simplify things :)

scruffian:
This was fixed in a different way

PR #3109, opened 2021-01-26
Simplify link selectors to avoid bleed

### LINKED ISSUE #3095: Spearhead: Default link color bleeding into hover state of Social Icons block
## Steps to replicate

1. With Spearhead theme active, add Social Icons block to a post or a page.
2. Add one or more social icons.
3. Save and preview. Hover over the icons and note the color change.

The same thing happens with other block styles as well.

This is a preview of the Social Icons block:
xoTBaOaQgv

(also note the slightly jerky animation and how the Edit link below jumps when hovering over the icons, and the fact that all the icons are enlarged when hovering)

You can see the effect live on https://gilbertstarter.wordpress.com/

Here's how Social Icons block should behave for comparison:
VFE53TVKN7

No user ticket found through my own testing.

pbking:

The link selectors were changed to apply a default border style to all
anchors (with a loose selector, just 'a') but leave that border-color
transparent.  All of these elements get a :hover to apply the primary
hover style.

Any anchor that's in wp-block or entry-content will instead have those
border color values inverted (defaults to --color-primary, hover is
transparent)

Buttons, social links (any other <a> that I found to evaluate) had more
specific selection rules to eliminate the border-bottom styles
preventing undesired underlines.

#### Related issue(s):

Closes #3095


spearhead/assets/sass/_text.scss
```diff
@@ -1,19 +1,18 @@
-.has-link-color a, a {
-    border-bottom: 1px solid transparent;
+.has-link-color, a {
```
scruffian: I think this should be `.has-link-color a, a`, otherwise when you add a link color setting it does this:

Screenshot 2021-01-27 at 17 06 35

scruffian (changes requested):
This breaks custom link colors

scruffian:
See also: https://github.com/Automattic/themes/pull/3122

spearhead/assets/sass/_text.scss
pbking: Ah, now I see.  Thank you.

pbking:
I tested with as many custom link configurations and methods as I could find.  The only difference that I saw was that the custom link colors weren't turning to the primary color on hover; that has been addressed.  If there are other custom link color issues please lmk.

scruffian (approved):
LGTM. Great work!

PR #3118, opened 2021-01-27
Prevented color change on social in cover

### LINKED ISSUE #3111: Seedlet: Social Icons color conflict with Cover block (Editor Styles)
When a Social Icons block is placed inside of a cover block that uses a "White" or "Tertiary" preset color background, the icon color is incorrect.

This only happens in the  editor. The front-end is fine. 

**To reproduce** 

1. Insert a Cover Block
2. Assign it a [white or tertiary color background color](https://cloudup.com/cUOr0ES8LR7). 
3. Add a Social Icons block. 
4. Note the icon color is dark. 

**Screenshots**

Screen Shot 2021-01-26 at 3 55 11 PM

Screen Shot 2021-01-26 at 3 55 18 PM

pbking:
The social buttons color setting assigned by Gutenburg default styles
was being clobbered by this color setting.  This change prevents any
wp-social-link elements from being selected.

This was reported against Spearhead but I found that it was caused by something in Seedlet and also is reproduced there.  The fix was tested against both themes.  

image

#### Related issue(s):

This fixes #3111

### DIFF
--- a/seedlet/assets/css/style-editor.css
+++ b/seedlet/assets/css/style-editor.css
@@ -533,33 +533,33 @@ object {
 .wp-block-cover .wp-block-cover__inner-container,
 .wp-block-cover .wp-block-cover-image-text,
 .wp-block-cover .wp-block-cover-text,
-.wp-block-cover .block-editor-block-list__block,
+.wp-block-cover .block-editor-block-list__block:not(.wp-social-link),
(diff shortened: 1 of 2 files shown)


scruffian (approved):
This solves the problem. As I dig more into it I don't think we need this CSS at all, but let's just merge this as it solves the issue.

PR #3119, opened 2021-01-27
Coutoire: Apply editor colors for Coutoire after Varia ones

### LINKED ISSUE #3121: Theme color palette in Coutoire appears to be overridden with bright colours
#### Steps to reproduce

1. Starting at URL: wordpress.com/new
2. Select the "Bowen" design and create a new site
3. Go to change the colours on a block, and look at the primary, secondary, foreground and background colours (a Button block is a good one for testing this)

#### What I expected

I expected the colours to use the theme's colour palette from the Customizer

#### What happened instead

Bright, unexpected colours are used instead

#### Screenshot / Video

Screenshot of a Button block selected in the editor, with unfortunate colours (this is from bowenstarterdesign, but any site created with Bowen design will show this):

image

#### Context / Source

Issue raised in D49646-code, which ensures that the headstart annotation for Bowen isn't applying these colours to any of the blocks in the annotation. However that change doesn't fix the issue, and it sounds like other themes might be affected too (p7DVsv-8MQ-p2#comment-31010)

scruffian:
This adds the action to apply editor colors after the Varia ones so that the correct ones show in the editor.

Before:
Screenshot 2021-01-27 at 15 29 42

After:
Screenshot 2021-01-27 at 15 26 28

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/3121

### DIFF
--- a/coutoire/functions.php
+++ b/coutoire/functions.php
@@ -104,7 +104,7 @@ if ( ! function_exists( 'coutoire_setup' ) ) :
 		);
 	}
 endif;
-add_action( 'after_setup_theme', 'coutoire_setup' );
+add_action( 'after_setup_theme', 'coutoire_setup', 12 );
 
 /**
  * Filter the content_width in pixels, based on the child-theme's design and stylesheet.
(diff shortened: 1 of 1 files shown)


jffng (approved):
LGTM. Does the version need a patch bump or can this be merged?

scruffian:
Let's merge and we can bump the numbers on Monday when we deploy.

Issue #3110, opened 2021-01-26
Seedlet: Fix custom color button hovers


kjellr:
When a user selects a totally custom color for their button, they have no hover state: 

![Kapture 2021-01-26 at 15 44 26](https://user-images.githubusercontent.com/1202812/105903049-6b7a5800-5fed-11eb-9950-dda11dcc49a4.gif)

This is (because of the inlined color rule from Gutenberg, but it is also) not totally great. 

Similarly, when a user chooses button colors that _are_ part of our preset palette, we force a blue hover color. This frequently doesn't work: 

![Kapture 2021-01-26 at 15 41 35](https://user-images.githubusercontent.com/1202812/105903294-b2684d80-5fed-11eb-88b9-684bfa3f451d.gif)
![Kapture 2021-01-26 at 15 42 23](https://user-images.githubusercontent.com/1202812/105903296-b300e400-5fed-11eb-96f2-a20644250563.gif)

Can we try using an opacity change for hovers when there is a custom color active? Something light — 90% opacity or so. That might solve both scenarios.

MaggieCabrera:
Same issue happens in Varia too: https://github.com/Automattic/themes/issues/3006

MaggieCabrera:
@kjellr do you think the same opacity solution would work for the outline buttons too?

kjellr:
I think it's worth a try there too, yes.

MaggieCabrera:
I opened a PR to fix the blue background (#3120), adding a proper hover state to buttons with custom colors should be done by GB, there's an issue open for that https://github.com/WordPress/gutenberg/issues/27075

MaggieCabrera:
I went ahead and merged #3120, let me know if you think we should do anything else on the theme side.

kjellr:
@MaggieCabrera I'm seeing a regression for outlined button hovers after this change. Can you confirm? 

![Kapture 2021-01-28 at 09 26 48](https://user-images.githubusercontent.com/1202812/106152060-03da1f00-614b-11eb-8e75-4ef4d4558f17.gif)

scruffian:
The rest of this will be addressed in Gutenberg, so I suggest we close this one.

PR #3120, opened 2021-01-27
Seedlet: removed hover states for buttons with custom background colors

MaggieCabrera:

This PR removes the hover states for buttons that have a custom background color.

#### Related issue(s):

Partially addresses #3110

### DIFF
--- a/seedlet/assets/css/style-editor.css
+++ b/seedlet/assets/css/style-editor.css
@@ -354,12 +354,12 @@ a {
 	margin-top: -calc(.5em * var(--button--line-height) + -.39);
 }
 
-.wp-block-a8c-blog-posts + .button:active {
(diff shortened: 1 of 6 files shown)


pbking:
This looks fine by me; removes the hover as-advertised.

Only thing to note, the "Secondary" color in the screencap above shows a hover color, however manually setting said secondary color on the button causes there to be no hover state (which was as I expected).

MaggieCabrera:
> Only thing to note, the "Secondary" color in the screencap above shows a hover color, however manually setting said secondary color on the button causes there to be no hover state (which was as I expected).

Yes, the secondary color hover is working fine, this PR only removes the hover for the rest of them because they were all using the same color as the secondary. I'm not removing the secondary hover because that will alter people's sites as they are now and it's working as intended.

PR #3123, opened 2021-01-27
Invert Calypso Global Style Font Assignments for Redhill

pbking:

This theme has had the primery/secondary <> heading/body variable
assignments inverted since it was brought in. This fix inverts the
relationship between primary/secondary <> heading/body variables at a
high level so that they are consumed as expected.

NOTE: This fix (or any fix) will not only change/fix the experience for those changing the fonts using this tool, it will also invert the heading/base font choices for any site using Redhill that has leveraged the Calypso Global Style Editor to manage their font.

#### Related issue(s):
https://github.com/Automattic/wp-calypso/issues/43904


pbking:
A more complete fix might be to evaluate each instance where the primary/secondary font variables are used in Redhill and invert those (it might be that not all of them SHOULD be inverted).  This change just flip-flops everything full stop.  That would be a much more involved fix however.

PR #3130, opened 2021-01-28
Seedlet: Fix for hover state on outline buttons

MaggieCabrera:

This PR removes hover effects on outlined buttons that don't have the secondary color.

#### Related issue(s):

#3110

### DIFF
--- a/seedlet/assets/sass/base/_extends.scss
+++ b/seedlet/assets/sass/base/_extends.scss
@@ -1,4 +1,9 @@
 @mixin button-style {
+	@include button-main-styles;
+	@include button-hover-styles;
+}
+
+@mixin button-main-styles {
 	@include crop-text(var(--button--line-height));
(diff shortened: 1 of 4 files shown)


kjellr:
This is much closer! The only issue I'm seeing is with outlined buttons that have a background color applied. It looks like in that case, the background colors aren't being picked up in the front end: 

Editor|Front End
---|---
Screen Shot 2021-01-28 at 10 48 31 AM|Screen Shot 2021-01-28 at 10 48 15 AM

MaggieCabrera:
> This is much closer! The only issue I'm seeing is with outlined buttons that have a background color applied. It looks like in that case, the background colors aren't being picked up in the front end

Thank you Kjell, you are so thorough :)

I think this should be fixed now.

kjellr (approved):
This is working well in my testing. Thank you!

Issue #2644, opened 2020-10-21
Spearhead: Dark mode text is unreadable by default when a preset color is selected


kjellr:
In light mode, the theme automatically adjusts the foreground text to pass contrast tests when one of the preset background colors is selected. This is broken when dark mode is active: 

Light Mode|Dark Mode
---|---
![Screen Shot 2020-10-21 at 9 33 13 AM](https://user-images.githubusercontent.com/1202812/96727062-a6327b80-1380-11eb-8b58-ada4da149f22.png)|![Screen Shot 2020-10-21 at 9 33 19 AM](https://user-images.githubusercontent.com/1202812/96727068-a894d580-1380-11eb-8699-06200426eb24.png)

1. While using dark mode, add paragraph blocks.
2. Set them to the preset background colors. 
3. Notice that the text is unreadable.

pbking:
Some observations:

When using the "Primary/Secondary" colors as background colors the colors shown in the "editor" vs the "viewer" differ.  This seems to stem from the same issue that incorrectly defines what the foreground color should be for elements with a "primary/secondary" background color while in dark mode.

Editor:
![image](https://user-images.githubusercontent.com/146530/97295855-3e6bad00-1826-11eb-9148-1d5a69fc5bb7.png)

View:
![image](https://user-images.githubusercontent.com/146530/97295890-49bed880-1826-11eb-8ecc-a85d2a1b5835.png)

pbking:
The cause seems to be that... the Customizer? is adding the background color as a style.  Which is necessary for a "non primary/secondary/tertiary" color.  However, the "dark mode" colors of what those "primary/secondary/tertiary" colors means something different than what the color swatches represent and shouldn't be enforced in that situation and rather rely on the .has-foreground-background-color[class] stylings.

pbking:
So further research has determined that the cause is due to the DYNAMIC colors being defined in [/spearhead/functions.php](https://github.com/Automattic/themes/blob/trunk/spearhead/functions.php#L79) that define the 'editor-color-palette' colors.  ONLY the 'light mode' colors are defined, thus even if the user's browser is in "dark mode" the "light mode" defined colors are shown in the editor.  These are the colors that are applied to the element as a style which is causing the issue.

Because "foreground" is "Black" in light mode and "White" in dark mode, but only defined as "White" in the functions.php block the editor is always representing "foreground" as "black".

[...]

PR #3132, opened 2021-01-28
Migrate Palette definitions from PHP to JSON

pbking:

This change moves the palette definitions from the functions.php and into the theme.json file.

The CSS variables that are generated and their availability is not expected to change (either to the block editor or in the view) only how that information is expressed by the theme.

The exception to the above are the Global CSS variables generated and inlined by Gutenberg for both the Block Editor:
image
and in the view:
image

#### Related issue(s):

This MAY be the beginning of a fix for https://github.com/Automattic/themes/issues/2644

With Spearhead generating palette information via theme.json we might be able to use THAT as the "flag" to NOT style elements with inline styles in the block editor if the color definitions come from theme.json (since in that situation it is guaranteed that the classes will exist).  This check could be done as an alternative to an additional theme supports.

### DIFF
--- a/spearhead/experimental-theme.json
+++ b/spearhead/experimental-theme.json
@@ -9,6 +9,25 @@
 		"settings": {
 			"border": {
         		"customRadius": true
+			},
+			"color": {
+				"palette": [
+					{
+						"slug": "primary",
+						"color": "#DB0042",
+						"name": "Primary"
+					},
+					{
+						"slug": "foreground",
+						"color": "#000000",
+						"name": "Foreground"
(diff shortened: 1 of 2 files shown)


MaggieCabrera:
This change would require .org users to have the gutenberg plugin installed. I'm not sure how we work with adding dependencies to a theme or if we should make any kind of fallbacks.

pbking:
Yea, I think that because of this there's no expectation to bring this into trunk until there is.  I'm going to close this PR and we'll make that transition when we can.

PR #3151, opened 2021-02-01
Blank Canvas: Add additional "Link in Bio" block patterns

kjellr:
Partially addresses #3161.

This PR adds a new "Link in Bio" category of block patterns, and populates it with all of the un-blocked patterns from #3161:

- Emoji Links
- Gradient Links
- Musician Links
- Product Links
- Text Links

---

To test: 

1. Open the Block Patterns Panel.
2. Select the "Link in Bio" category.
3. Insert the included patterns. 
4. Check for any errors on the front and backend. 

---

## Screenshots

patterns
Screen Shot 2021-02-02 at 3 40 42 PM
patterns-2


kjellr:
This one should be ready for review!

melchoyce:
Product Links and Text Links look good, but the ones using multiple buttons are having some alignment issues on my end:

image
image
image

I also just noticed Product Links has some extra spacing around the button:

image

I'm testing on 5.6 without Gutenberg installed. All of these issues are resolved when I install the latest release of Gutenberg, where everything looks great 👍 Don't know if there's anything we can do about making the alignment better on default 5.6 installs. If not, I don't think it's a deal-breaker.

kjellr:
Ah yes. WordPress.com has the Gutenberg plugin installed, and also I'm relatively sure that the button widths setting is shipping in core alongside 5.7. 

That being the case, I think it doesn't make much sense to build in a fallback for earlier versions.

scruffian:
So nice to see no CSS in this PR!

scruffian (approved):
LGTM. My only concern is that these patterns aren't translated yet, but I don't want to hold off merging because of that. We do need to find a solution though.

PR #3157, opened 2021-02-02
Varia + children: css variable approach to color annotations

### LINKED ISSUE #3038: Varia: css variable approach to color annotations
Implement the same approach to color annotations used in https://github.com/Automattic/themes/pull/3025 

Move the javascript file that handles color annotations from Hever to Varia as suggested by this comment: https://github.com/Automattic/themes/pull/3025#discussion_r558140479

Make sure that Hever is inheriting correctly from Varia and the color annotations work for both themes so we can start rolling these changes to the rest of the themes missing color annotations.

MaggieCabrera:

This is a WIP

#### Related issue(s):

Closes #3038


hever/style.css
```diff
@@ -181,6 +175,43 @@ body {
  * Redefine Sass map values for child theme output.
  * - See: style-child-theme.scss
  */
+/**
+ * Redefine Sass map values for child theme output.
+ * - See: style-child-theme.scss
+ */
+/**
+ * Global
```
MaggieCabrera: This is really ugly but I need to import Varia's variable file in order to merge it with the child's. We could change the comments to not be multiline (I think those don't carry over to the compiled css) or just leave it like it is.

scruffian: Fixed!

varia/sass/child-theme/_global-variables.scss
```diff
@@ -0,0 +1,25 @@
+@mixin global-variables() {
+
+	/* Colors */
+	--wp--preset--color--primary: blue;
```
pbking, line 4: It seems to me that these values, rather than attempting to copy the function of a "palette" and named such (--wp--preset--color) should instead be CUSTOM values and named according to that...

--wp--custom--color--primary   (etc)

As a way to reflect values that might be changed (variables) as opposed to values that can't (palettes).

MaggieCabrera: I see what you mean, but for these themes, will we ever have a preset palette? what values would it have?

stow/functions.php
```diff
@@ -52,43 +64,6 @@ function stow_setup() {
...
-					'slug'  => 'primary',
-					'color' => '#404040', 
-				),
-				array(
-					'name'  => __( 'Secondary', 'stow' ),
-					'slug'  => 'secondary',
-					'color' => '#f25f70', 
-				),
-				array(
-					'name'  => __( 'Dark Gray', 'stow' ),
```
scruffian, line 71: Not sure that it's ok to remove this

stow/functions.php
```diff
@@ -52,43 +64,6 @@ function stow_setup() {
...
-					'slug'  => 'foreground-dark',
-					'color' => '#111111',
-				),
-				array(
-					'name'  => __( 'Gray', 'stow' ),
-					'slug'  => 'foreground',
-					'color' => '#444444',
-				),
-				array(
-					'name'  => __( 'Light Gray', 'stow' ),
```
scruffian, line 81: Not sure that it's ok to remove this either

stow/functions.php
```diff
@@ -52,47 +63,25 @@ function stow_setup() {
...
+	);
+	$palette[] = array(
+		'name'  => __( 'Light Gray', 'stow' ),
+		'slug'  => 'foreground-light',
+		'color' => '#767676',
+	);
+	return $palette;
+}
+add_filter( 'varia_editor_color_palette', 'stow_editor_color_palette' );
+
```
MaggieCabrera: ~~@Automattic/theam we need to add this to the themes that have Foreground light/dark options on their editor color palette. It's not the case for all of them. Thanks for this Ben!~~

We have discussed this and each theme handles this a bit differently so we are going to try and streamline them all for better UX, so that all themes show the main 4-5 colors. We need to test that if the user has applied some of the other colors their site remains unchanged. After testing this on Stow I have found that my sandbox didn't change when I removed the extra colors from the palette.

stow/functions.php
```diff
@@ -52,47 +63,25 @@ function stow_setup() {
...
-				),
-			)
-		);
 	}
 endif;
 add_action( 'after_setup_theme', 'stow_setup', 12 );
 
+function stow_editor_color_palette( $palette ) {
+	$palette[] = array(
+		'name'  => __( 'Dark Gray', 'stow' ),
```
MaggieCabrera: ~~Shouldn't this be dynamic and change with the palette? Looks like Stow's editor color palette were static, but some themes had them change when the user selects a new palette, like Alves does: https://github.com/Automattic/themes/blob/6862cbc6552a5fac9d9592c681588f461c32c3be/alves/functions.php#L94~~

varia/sass/child-theme/_global-variables.scss
pbking: I'm not sure I agree with past-pbking.

scruffian:
I think we can close this as we have broken it into smaller PRs

PR #3163, opened 2021-02-02
Blank Canvas: Even out footer spacing

kjellr:
Blank canvas adjusts the top padding when certain full-width blocks are at the top of the page. This PR adjusts the bottom spacing for those instances too so it's more even. This is especially useful for some of the new patterns we're adding in #3151.

We only want this to happen if the comments and footer options are disabled, so this adds a couple body classes to hook the styles onto. Since we're adding body classes anyway, it migrates the existing `hide-post-and-page-titles` class from `entry-content` to the body instead.

Before|After
---|---
Screen Shot 2021-02-02 at 2 55 16 PM|Screen Shot 2021-02-02 at 2 55 38 PM


scruffian (approved):
LGTM

PR #3172, opened 2021-02-03
WIP refactor of brompton to use CSS variables

pbking:
WIP branch to refactor Brompton to use the new Varia CSS Variables.

This almost-mostly works but I feel like I'm missing something.

If the CSS variables are included in the theme CSS then they "win" (they appear to be loaded after the wpcom-colors.php has done its thing).  Excluding them means that the styles pick up the variables defined by the wp-com-colors.php plugin in varia (as expected).

However, if there are NO custom colors selected the "default" colors aren't defined.  That seems to be what _global-variables.scss is supposed to be for but I couldn't get it to work that way.  (If _global-variables.scss is included then the values that are there are always the color values used).


pbking:
The [inclusion of the css variables here](https://github.com/Automattic/themes/pull/3172/files#diff-9af0797562dc6caf22f58630478a129731d75dda7e00ff839e4accd56ccc8270R36) is what seems to be throwing me.

This is mirroring the change f[ound here](https://github.com/Automattic/themes/pull/3157/commits/fc1a92535848cb8b848e8dd536a378a20192f8b4#diff-67d96ac48ba30fee69244f69b7057f6d27a5ae8da3b383e66c6c596ef35f0150R11).

brompton/sass/_config-child-theme-deep.scss
```diff
@@ -70,37 +70,37 @@ $config-global: (
 	/* Colors */
 	"color": (
```
jffng: I believe this entire color map should be removed, and after this config is set, you merge the varia config like so: 

https://github.com/Automattic/themes/blob/f264ac103ba2b07896acb6a9887a1e88b4155160/alves/sass/_config-child-theme-deep.scss#L108

brompton/inc/wpcom-editor-colors.php
```diff
@@ -1,224 +1,2 @@
...
-	// Text-color
-	array( '#editor .editor-styles-wrapper .has-secondary-color,
-			#editor .editor-styles-wrapper .wp-block .has-secondary-color', 'color' ),
-
-	// Background-color
-	array( '#editor .editor-styles-wrapper .has-secondary-background-color,
-			#editor .editor-styles-wrapper .wp-block .has-secondary-background-color', 'background-color' ),
-
-), __( 'Secondary Color' ) );
+L:<?php
```
jffng, line 1: Is this a typo?

brompton/functions.php
```diff
@@ -61,56 +73,7 @@ function brompton_setup() {
...
-					'slug'  => 'foreground',
-					'color' => $foreground,
-				),
-				array(
-					'name'  => __( 'Foreground Light', 'barnsbury' ),
-					'slug'  => 'foreground-light',
-					'color' => $foreground_light,
-				),
-				array(
-					'name'  => __( 'Foreground Dark', 'barnsbury' ),
```
scruffian, line 108: You'll need to add these last two back in. If you look at what I've done for Stow you can see how.

PR #3180, opened 2021-02-04
Fix errors for redefined functions

scruffian:
These redeclared functions is causing errors. We need to add a check.


jffng (commented):
Should the following functions also be wrapped in a `function_exists` for safety?

`hex_to_rgb`
`rgb_to_lum`
`rgb_to_hsvl`

These themes appear to redeclare them also.

MaggieCabrera (approved):
LGTM

scruffian:
I added the other checks @jffng suggested, so this will need another review

jffng:
I tested this on wpcom and verified that the custom color HSL calculations are working. LGTM.

PR #3182, opened 2021-02-04
Varia: re-order Gutenberg color utilities rules

### LINKED ISSUE #2548: Mayland: Buttons Block predefined colors issue
## Steps to replicate

1. Activate Mayland theme
2. Insert a Buttons block into a page/post
3. Choose predefined colors for fore- and background

## Result

Editor:
Button Block Color Settings

Front-end:
Front-end

Mayland CSS file:
Mayland CSS

## Expected

I expected the button to show the colors from the picked predefined color settings on the front-end, but it seems the CSS selectors are overwritten or not specific enough in the Mayland CSS file.
The front-end screenshot shows the CSS file line numbers, which are overwriting the chosen color.

jffng:
This PR addresses #2548.

Previously, if a block had both a preset color and background color set, the cascade would cause the color to be overridden by the background color utility class:

Screen Shot 2021-02-04 at 12 18 34 PM

This PR re-orders them so that if a preset color is selected, it will take precedence — without increasing the specificity.

**To test**
- Activate this PR + Mayland
- Add a button with a preset color + preset background
- Verify the front-end and editor match and appear as expected

Note that this should probably be compiled into all the child themes if we like the change.

### DIFF
--- a/mayland/style-rtl.css
+++ b/mayland/style-rtl.css
@@ -2424,39 +2424,6 @@ table.is-style-stripes tbody tr:nth-child(odd),
 	background-attachment: fixed;
 }
 
-.has-primary-color {
-	color: #000000;
-}
-
-.has-secondary-color {
-	color: #1a1a1a;
-}
-
-.has-foreground-color {
-	color: #010101;
-}
-
-.has-foreground-light-color {
-	color: #666666;
-}
-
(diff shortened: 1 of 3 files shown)


scruffian:
Closing in favour of #3185

PR #3185, opened 2021-02-04
Varia + Children: Separate background & color

### LINKED ISSUE #2866: Shawburn: text colors are not reflected on the published version when a default light background color is selected
## Steps to replicate

1. Open a post or a page editor
2. Add some text using paragraph block or heading block
3. Under Color Settings > select one of the default text color
4. Then select one of the default light background color, such as Subtle Gray or Light Gray
5. Save the changes

Screenshot of the editor: 

Screenshot on 2020-11-28 at 23-32-16

Screenshot of the published version: 

Screenshot on 2020-11-28 at 23-33-59

## Result

The text colors are reflected in the editor, but not on the Preview and on the published version. On the Preview and the published version, the text color is black #444.

## Expected

The text colors are the same on the editor, on the Preview, and on the published version. 

## Workaround

Use custom background color instead of the default ones.

* User report: #3513446-zen
* As a workaround, I suggested to change the background color from the Subtle Gray (#fafafa) to #f8f8f8.

pbking:

This change creates a more specific selector for applying the color of
an element with a background color applied.

Previously the class that applied the background color also applied a
reasonable text color.

With this change only matches that also do not also supply a
'has-text-color' class will set a color property for a .has-*-background

#### Related issue(s):

This is a fix for #2866 that has a wider range effect.  

This may also address #2548.

Editor
image

View
image

### DIFF
--- a/alves/style-rtl.css
+++ b/alves/style-rtl.css
@@ -6,7 +6,7 @@ Author: Automattic
 Author URI: https://automattic.com/
 Description: Convincing design for your charity or organization’s online presence. Highlight your actions, causes and projects, Alves is versatile enough to be your personal site too.
 Requires at least: WordPress 4.9.6
-Version: 1.5.1
+Version: 1.5.2
 License: GNU General Public License v2 or later
 License URI: LICENSE
 Template: varia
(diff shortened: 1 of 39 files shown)


MaggieCabrera (approved):
LGTM

danieldudzic (approved):
This looks solid!

Issue #3156, opened 2021-02-02
Mayland: Custom link color interfering with social media icons


alaczek:
## Steps to replicate

1. Activate Mayland and open the Customizer.
2. Go to Colors & Backgrounds tab and set custom colors. My settings here were:
 - Background: #f0f2e8
 - Link & Text: #bf462c
 - Secondary: #1a1a1a
3. Add a new post or a page and insert a social icons block, and add a few icons.
4. Note the square, red background when you hover an icon in the Editor.
5. Preview the site. Note the inside of the social icons is using the custom link color, instead of white.

## Result
Editor view, hover over an icon:
![image](https://user-images.githubusercontent.com/4550351/106544482-b94a0f80-655b-11eb-8700-4e4ddea098ef.png)

Preview:
![image](https://user-images.githubusercontent.com/4550351/106544527-ca931c00-655b-11eb-8410-4b1fb7483f5c.png)

No customer report, found through my own testing.

jffng:
Closed by #3189, but will ship with a bundle of updates next week.

Issue #2895, opened 2020-12-07
Mayland: background colour affecting social links colour


ianstewart:
What happened:

Using a background colour with Mayland changes the colour of the Social Links block.

![image](https://user-images.githubusercontent.com/545779/101407032-e5b91680-389f-11eb-8f71-8db069cbf3bc.png)

What I expected: 

The social links block to be unaffected by the background colour change.

alaczek:
Sounds like I duplicated this issue in #3156, but I also noticed weird square background on hover in the editor.

pbking:
I tried hard but I've been unable to reproduce this (or #3156) using any of the steps provided.  Colors for social icons appear as expected in every scenario I could think to test Mayland in: Customiser, Post Editor, View... all appears as expected with backgrounds (and every other color adjusted) both with pre-defined palettes and custom colors.

![image](https://user-images.githubusercontent.com/146530/106940871-14ad0680-66f0-11eb-9214-68a345c1b466.png)

alaczek:
@pbking you can see this issue live on https://balanstarter.wordpress.com/ and https://knappstarter.wordpress.com/

jffng:
Closed by #3189, but will ship with a bundle of updates next week.

PR #3189, opened 2021-02-04
Mayland: fix social links when custom colors are applied

jffng:
This PR addresses #3156 and #2895.

I'll try to explain the problem, solution, and why it's frustrating.

1. Gutenberg adds [blocks styles to the social links](https://github.com/WordPress/gutenberg/blob/master/packages/block-library/src/social-links/style.scss#L105-L108) so that svg's color and fill attributes are set to `currentColor`. I think this is the right thing to do.
2. Mayland's [color annotations apply a more specific rule](https://github.com/Automattic/themes/blob/trunk/mayland/inc/wpcom-colors.php#L148): `a:not(.has-text-color)`, which end up applying the custom link color to the svg, causing the issues reported. 
3. To fix this, I needed to add a more specific CSS rule, which is this PR. I would have made this change in Varia, but Mayland is the only theme that has `a:not(.has-text-color)` in its color annotations, so it felt like overkill.

Why it's frustrating:
- The offending rule is only present in Mayland color annotations. I do not know why. It has no corresponding sass/css in Varia or the child theme.
- Having to trace what should just be a CSS issue ends up touching Gutenberg, Varia, Mayland, and the color annotations, it feels unmanageable.

### DIFF
--- a/mayland/inc/wpcom-colors.php
+++ b/mayland/inc/wpcom-colors.php
@@ -3,10 +3,14 @@
 
 // Background Color
 // $config-global--color-background-default
-add_color_rule( 'bg', '#ffffff', array(
+add_color_rule(
+	'bg',
+	'#ffffff',
+	array(
 
-	// Background-color
-	array( '.screen-reader-text:focus,
+		// Background-color
+		array(
+			'.screen-reader-text:focus,
 			body,
 			body .widget_eu_cookie_law_widget #eu-cookie-law,
 			body .widget_eu_cookie_law_widget #eu-cookie-law.negative input.accept,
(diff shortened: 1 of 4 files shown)


scruffian:
Could we just remove the rule from the color annotations?

jffng:
I don't think we can remove it entirely because it's that line that applies the custom color to links. 

I made it more generic and it seems to work at the correct level of specificity.

mayland/inc/wpcom-colors.php
```diff
@@ -145,7 +205,7 @@
 			.wp-block-newspack-blocks-homepage-articles article .entry-title a:active,
 			.wp-block-newspack-blocks-homepage-articles article .entry-title a:focus,
 			.wp-block-newspack-blocks-homepage-articles article .entry-title a:hover,
-			a:not(.has-text-color),
+			a,
```
jffng, line 148: This is the only line that changed, the rest of the changes are linter fixes.

pbking: Do the linting fixes have to be in there?
With such a small change it's a shame not to play golf with the DIFF to make the history clear.

jffng: They don't! I do like golf.

jffng:
Thanks for your testing and feedback @pbking @scruffian.

PR #3192, opened 2021-02-05
Varia: Copy style changes to editor styles

scruffian:
I think the changes we made in #3185 should also be made to the editor styles. If this passes review I'll rebuild before I merge.

### DIFF
--- a/varia/sass/blocks/utilities/_editor.scss
+++ b/varia/sass/blocks/utilities/_editor.scss
@@ -69,53 +69,87 @@
 	}
 }
 
-.wp-block .has-primary-background-color,
-.has-primary-background-color {
+.has-primary-background-color,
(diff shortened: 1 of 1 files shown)


pbking:
There won't be an obvious change in the editor due to the ([bug](https://github.com/WordPress/gutenberg/issues/26506)) where the style color/background-color attributes are always set on an element in the editor (though not in the view).

If/when this bug gets addressed THEN these changes will be necessary, however putting them in place now while it's on the brain makes sense.

To keep the logic DRY, would it be appropriate to put the changes into a common .scss file?

scruffian:
> To keep the logic DRY, would it be appropriate to put the changes into a common .scss file?

Yes, that would be preferable. There are a lot of instances of where we need to do this, but I guess we have to start somewhere!

PR #3193, opened 2021-02-05
Varia: Remove overrides for links in elements with background color set

### LINKED ISSUE #2667: Hever: Links lose colour inside List and other blocks if background colour is changed
## Steps to replicate

* Open page/post editor
* Add a list block and add some text inside it.
* Hyperlink a text. 
* Change the 'List' block's background colour. 

https://prnt.sc/v4gr9t

## Result

The Hyperlinks lose their colour and turn to black. 
This can be restored if you use the button "Clear" to clear the background colour of the List block.

https://prnt.sc/v4gtxr

## Expected

I expect the links to hold on to their colour as I've seen with other themes.

EDIT: This is happening across most of the blocks in the Hever theme.

* User report: #24603976-hc
* Follow-up ticket: #3424787-zd

scruffian:
This removes special overrides for links inside elements with background colors. I don't think we need these fallbacks, but is we do Gutenberg should be able to handle them.

I have only rebuilt hever to make this easier to review, but we should really rebuild all. I can do that once the review is done.

Fixes: #2667

### DIFF
--- a/alves/style-editor.css
+++ b/alves/style-editor.css
@@ -1055,7 +1055,6 @@ pre.wp-block-verse {
 	color: #ffffff;
 }
 
-.has-background:not(.has-background-background-color) a:not(.wp-block-button__link),
 .has-background p:not(.has-text-color),
 .has-background h1:not(.has-text-color),
 .has-background h2:not(.has-text-color),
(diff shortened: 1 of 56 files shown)


MaggieCabrera (approved):
This worked for the bug it's solving and I couldn't find a case where it broke on a different circumstance. LGTM after rebuild

Issue #3016, opened 2021-01-13
Spearhead/Blank Canvas: Non-wpcom custom colors implementation has no effect.


kjellr:
When testing locally or on a self-hosted site, these themes inherit Seedlet's custom colors implementation. Unfortunately, the default colors are incorrect, and changing them has no effect on the site. 

To test: 

1. Load Spearhead or Blank Canvas on a self-hosted site.
2. Open Customizer > Colors.
3. Select "Custom"
4. Note that the colors there are from Seedlet.
5. Modify the colors. Note that this has no effect on the preview on the right. 

## Screenshot

![Screen Shot 2021-01-13 at 8 09 15 AM](https://user-images.githubusercontent.com/1202812/104456556-d7cd7400-5576-11eb-8853-376cbdba786c.png)

scruffian:
This might be fixed by #2701

scruffian:
I think this was fixed by #2701. If it's still an issue please reopen.

supernovia:
This still seems to be an issue on my .org instance with a fresh copy of Spearhead today.

pbking:
I tested this and the only portions of the Spearhead that I was unable to customize were the BACKGROUND color of Sticky Posts (and other elements referenced in #3105).  I opened #3199 to address that shortfall (#3105 addressed it for the .com color customizer, #3199 should address it for .org)

I evaluated the fix in both LIGHT and DARK mode.  

All other aspects of Spearhead color customizations were working as expected for me.

scruffian:
I think there's still more work needed here. When I load spearhead color settings they are still the Seedlet colors:

<img width="317" alt="Screenshot 2021-02-09 at 12 20 04" src="https://user-images.githubusercontent.com/275961/107362801-26682280-6ad1-11eb-949c-0ed080561986.png">

PR #3199, opened 2021-02-05
Use custom background color for high-contrast

pbking:

Continuing #3105 but ensuring that the background color is used as the
high contrast background color when set in the .org color customizer.

Prior to the change the 'sticky post' background (and anything else that used the --global--color-background-high-contrast CSS variable) wouldn't pick up changes made via the .org color customizer.   (Due to changes in #3105 those portions were working fine using .com color customizer).

#### Related issue(s):

This may be the issue that caused #3016 to be reopened.


maharzan:
This is not fixing the color options in spearhead .org version.

scruffian (approved):
This is good in that it addresses the issue with sticky posts but I don't think it's a fix for #3016

pbking:
version bumped and deployed here: #3224

PR #3206, opened 2021-02-08
Update version numbers for varia themes

scruffian:
This updates the version numbers for Varia themes and also updates the scripts so that version numbers can be updates separately from rebuilding.


MaggieCabrera (approved):
LGTM! it looks like the rebuild from a previous PR didn't create the editor versions of the files, but this should be fine

PR #3209, opened 2021-02-08
Varia: remove extra specificity of background color classes

### LINKED ISSUE #2994: Coutoire: Unexpected results in Colour Schemes for Cover Block Overlay
## Steps to replicate

1. Visit WP.com/new and select the Bowen design
2. Create a site.
3. Compare the cover block colour styles in the Editor vs the Front-End

## Result

image

## Expected

The Editor style colours to match the front-end.

jffng:
I don't think we should use the extra level of specificity `not:has-text-color` to target classes that are meant to apply a background color. It causes unexpected behavior like #2994. This PR removes them.

If we are comfortable with this, the changes would need to be recompiled for all Varia themes.

### DIFF
--- a/coutoire/style-editor.css
+++ b/coutoire/style-editor.css
@@ -1082,9 +1082,9 @@ pre.wp-block-verse {
 	color: #FFFFFF;
 }
 
-.has-background-dim:not(.has-text-color),
-.has-foreground-background-color:not(.has-text-color),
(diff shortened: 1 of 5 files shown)


pbking:
Looks like the right fix to me.  Fixes the issue in Coutoire and spot checked a couple of other child themes.

PR #3220, opened 2021-02-09
Varia: Add a customiser option to show featured images on pages

### LINKED ISSUE #3212: Varia: Add Content Option for featured images on pages that defaults to off
We recently fixed an issue (#3195) where featured images were not displayed on pages on sites using Varia and its child themes. This has caused existing sites to suddenly show featured images where they weren't before with the only options for hiding being CSS (requires the site be on a supported plan) or removing the image, which removes the ability to control the image shared on social media.

We should add a content option controlling the display of featured images on pages and default it to off to fix this.

## Steps to replicate

0. On a site using Varia or a child theme
1. Add a featured image to a page
2. It is displayed on the front-end, which is a change from previous behavior

cc @kjellr @danieldudzic

scruffian:
Adds a customizer option to turn featured images off by default but allow users to enable them.

#### Related issue(s):
Fixes #3212

### DIFF
--- a/varia/functions.php
+++ b/varia/functions.php
@@ -263,7 +263,8 @@ function varia_scripts() {
 	if ( $is_IE ) {
 		// If IE 11 or below, use a ponyfill to add CSS Variable support
 		wp_register_script( 'css-vars-ponyfill', get_template_directory_uri() . '/js/css-vars-ponyfill2.js' );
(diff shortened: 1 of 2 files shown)


MaggieCabrera (approved):
Works as expected and removes the image from the page when the patch is applied ⛵

PR #3223, opened 2021-02-09
Varia + Hever: Refactor color annotations

MaggieCabrera:

This PR refactors the color annotations so that most of the code lives on Varia while maintaining the ability to leverage either of the two methods by its children.

This PR:

* Moves the IE stylesheet to a separate stylesheet that is not compiled by the children but loaded only when it's needed.
* Defines the editor color palette within Varia. The children using the old method can still redefine the palette the old way and the children using the new method can avoid doing so by defining the colors initially within `varia_default_colors`
* Creates a secondary file `wpcom-editor-colors.php` with the new implementation of the color annotation that should be imported only by the children that are using it and ignored by the rest (including Varia).
* Moves all the JS color utilities to Varia instead of having them on the children that implement de CSS variable method.

To test:

* Check that Varia color annotations work ~~fine~~ **the same way they did before this patch** (and one of the children with the old implementation for good measure)
* Check that the editor color palette is still working ~~as expected~~ **the same way it did before this patch** and changes when the colors are changed.
* Do the same with Hever.

This PR will require a full build after it's approved, but I'm holding on for clarity of the code.


pbking:
A quick comparison of EDITOR vs VIEW shows a number of differences between potential button styles:

image

The buttons that are different: 
Stock Outline
Primary Outline
Background Outline
Primary/Secondary Outline

Removing [this line](https://github.com/Automattic/themes/blob/trunk/varia/sass/blocks/button/_style.scss#L22) in Varia's button styles (untested in other child themes) seems to correct the style of outline buttons on the view eliminating the discrepancies. (with the exception of a "Stock Outline" button which still renders incorrectly in the EDITOR only)  There may be a better way to address that, but that's what my sleuthing uncovered.

pbking:
The variables produced by the "Tertiary Color" appear to be different for the view and what the customizer generates during a live preview.

image

(specifically the *-low-contrast and *-high-contrast variants)

View CSS variables (on a fresh load of the view OR the customizezr):
--wp--preset--color--border: #cbe86b;
--wp--preset--color--border-low-contrast: hsl( 0,0%,100.98039215686%);
--wp--preset--color--border-high-contrast: hsl( 0,0%,80.980392156863%);

CSS variables after making any changes in the customizer to the 'Tertiary Color':
-wp--preset--color--border: #cbe86b;
--wp--preset--color--border-low-contrast: hsl( 73.92000000000002,73.09941520467835%,76.47058823529412%);
--wp--preset--color--border-high-contrast: hsl( 73.92000000000002,73.09941520467835%,56.470588235294116%);

From what I can tell this is due to change_color_luminescence() resolving to a DIFFERENT value if calculated in [JavaScript](https://github.com/Automattic/themes/pull/3223/files?file-filters%5B%5D=.js#diff-fc0ee5d9339d9cadd58dfb259e7988bec719aeb625d291ee73fa2edce627e006R36) vs calculated in [PHP](https://github.com/Automattic/themes/pull/3223/files?file-filters%5B%5D=.js&file-filters%5B%5D=.php#diff-a80b9d88f1219e6cd3d5b48264038a5c0f236de49182acee64ad4abcf9f45661R116).

varia/functions.php
```diff
@@ -268,6 +294,7 @@ function varia_scripts() {
 			array( 'css-vars-ponyfill' ),
 			'1.0'
 		);
+		wp_enqueue_style( 'varia-print-style', get_template_directory_uri() . '/ie.css', array(), wp_get_theme()->get( 'Version' ), 'print' );
```
jffng: Should this style handle be renamed and not refer to `print` styles?

```suggestion
		wp_enqueue_style( 'varia-ie-styles', get_template_directory_uri() . '/ie.css', array(), wp_get_theme()->get( 'Version' ) );
```

hever/inc/wpcom.php
```diff
@@ -6,12 +6,11 @@
...
-
-
+require_once get_template_directory() . '/inc/wpcom-colors-css-variables.php';
 /**
  * Bind JS handlers to instantly live-preview changes.
  */
-function hever_wpcom_customize_preview_js() {
-	wp_enqueue_script( 'hever_wpcom_customize_preview', get_theme_file_uri( '/inc/customize-preview-wpcom.js' ), array( 'customize-preview' ), '1.0', true );
+function hever_wpcom_color_annotations_preview_js() {
+	wp_enqueue_script( 'hever_wpcom_color_annotations_preview', get_template_directory_uri() . '/inc/color-annotations-preview.js', array( 'customize-preview' ), '1.0', true );
```
jffng: This is not entirely specific to this PR, but what is the reason for using a fixed value (`1.0`) for the version string, instead of `wp_get_theme()->get( 'Version' )`?

MaggieCabrera:
> A quick comparison of EDITOR vs VIEW shows a number of differences between potential button styles:

Thanks for this!! I think this bug is out of the scope of this PR though, which only intends a refactor to prepare for other themes to use color annotations. I didn't see regressions from what is right now on production to what happens after this patch is applied, though I agree the discrepancies between editor and frontend are there, they are not introduced by this PR and should be handled on a different issue so we don't block this one and can move on to add color annotations to the themes missing them, do you agree?

MaggieCabrera:
I reported the color discrepancies here: https://github.com/Automattic/themes/issues/3232

MaggieCabrera:
> The variables produced by the "Tertiary Color" appear to be different for the view and what the customizer generates during a live preview.
> From what I can tell this is due to change_color_luminescence() resolving to a DIFFERENT value if calculated in [JavaScript](https://github.com/Automattic/themes/pull/3223/files?file-filters%5B%5D=.js#diff-fc0ee5d9339d9cadd58dfb259e7988bec719aeb625d291ee73fa2edce627e006R36) vs calculated in [PHP](https://github.com/Automattic/themes/pull/3223/files?file-filters%5B%5D=.js&file-filters%5B%5D=.php#diff-a80b9d88f1219e6cd3d5b48264038a5c0f236de49182acee64ad4abcf9f45661R116).

Yes, I believe this has always been the case and the preview hover colors are not exactly the same as the ones that end up being used but I think they are similar enough. It's a good point and maybe we could improve the way we are doing this.

hever/inc/wpcom.php
MaggieCabrera: No reason, I added it, I think it's best that way

varia/style-editor.css
```diff
@@ -1066,51 +1065,85 @@ pre.wp-block-verse {
 	color: currentColor;
 }
 
-.wp-block .has-primary-background-color,
-.has-primary-background-color {
+.has-primary-background-color,
+.has-primary-background-color.has-background-dim {
```
scruffian: I guess this is from another PR where we forgot to rebuild varia...

scruffian: it went away when i rebased...

MaggieCabrera: makes sense that it did, I was suspecting it would :)

varia/inc/color-annotations-preview.js
```diff
@@ -0,0 +1,107 @@
...
+
+			var background = to.bg;
+			var foreground = to.txt;
+			var primary = to.link;
+			var secondary = to.fg1;
+			var border = to.fg2;
+
+			var cssVariables = '';
+
+			if(background !== ''){
```
scruffian: Spacing issue

varia/inc/color-annotations-preview.js
```diff
@@ -0,0 +1,107 @@
...
+				var backgroundLowContrast = changeColorLuminescence( background, -10 );
+				var backgroundHighContrast = changeColorLuminescence( background, 10 );
+				cssVariables += '--wp--preset--color--background: ' + background + ';' +
+				'--wp--preset--color--background-low-contrast: ' + backgroundLowContrast + ';' +
+				'--wp--preset--color--background-high-contrast: ' + backgroundHighContrast + ';';
+			}
+
+			if(foreground !== ''){
```
scruffian: Spacing issue

varia/inc/color-annotations-preview.js
```diff
@@ -0,0 +1,107 @@
...
+				var foregroundLowContrast = changeColorLuminescence( foreground, 10 );
+				var foregroundHighContrast = changeColorLuminescence( foreground, -10 );
+				cssVariables += '--wp--preset--color--foreground: ' + foreground + ';' +
+				'--wp--preset--color--foreground-low-contrast: ' + foregroundLowContrast + ';' +
+				'--wp--preset--color--foreground-high-contrast: ' + foregroundHighContrast + ';';
+			}
+
+			if(primary !== ''){
```
scruffian: White space...

varia/inc/color-annotations-preview.js
```diff
@@ -0,0 +1,107 @@
...
+				var primaryHover = changeColorLuminescence( primary, 10 );
+				cssVariables += '--wp--preset--color--primary: ' + primary + ';' +
+				'--wp--preset--color--primary-hover: ' + primaryHover + ';';
+			}
+
+			if(secondary !== ''){
```
scruffian: White space...

varia/inc/color-annotations-preview.js
```diff
@@ -0,0 +1,107 @@
...
+				var secondaryHover = changeColorLuminescence( secondary, 10 );
+				cssVariables += '--wp--preset--color--secondary: ' + secondary + ';' +
+				'--wp--preset--color--secondary-hover: ' + secondaryHover + ';';
+			}
+
+			if(border !== ''){
```
scruffian: White space...

scruffian (approved):
This looks great. The only thing I noticed were minor linting issues.

🚢

scruffian (approved):

MaggieCabrera:
Deployed with D56820-code

PR #3122, opened 2021-01-27
Spearhead: Fix link color

scruffian:
#### Changes proposed in this Pull Request:
We need to specify the link color in theme.json.

Before:
Screenshot 2021-01-27 at 17 15 58

After:
Screenshot 2021-01-27 at 17 15 47

PR #3225, opened 2021-02-09
Varia: add experimental link color

### LINKED ISSUE #2566: Varia Themes: Add support for custom link color
It'd be great if Varia themes support custom link color like Seedlet.

https://github.com/WordPress/gutenberg/blob/755ef781166c2ef60fe44c0af465baa3acb25b3d/docs/designers-developers/developers/themes/theme-support.md#experimental--link-color-control

jffng:
This PR addresses #2566.

Tested with and without color annotations on the Mayland theme, and appears to work fine. 

My hesitation in adding this is that there is no support yet for setting the link's hover color in Gutenberg, so may result in some strange combinations.

### DIFF
--- a/hever/experimental-theme.json
+++ b/hever/experimental-theme.json
@@ -1,0 +1,9 @@
+{
+	"styles": {
+		"root": {
+			"color": {
+				"link": "var(--wp--preset--color--primary)"
+			}
+		}
+	}
+}
--- a/hever/functions.php
+++ b/hever/functions.php
@@ -126,6 +126,9 @@ if ( ! function_exists( 'hever_setup' ) ) :
 		if ( function_exists( 'varia_mobile_nav_on_side_setup' ) ) {
 			varia_mobile_nav_on_side_setup();
 		}
+
+		// Add support for experimental link color via Gutenberg: https://github.com/WordPress/gutenberg/blob/master/docs/designers-developers/developers/themes/theme-support.md
+		add_theme_support( 'experimental-link-color' );
 	}
 endif;
 add_action( 'after_setup_theme', 'hever_setup', 12 );
--- a/mayland/style-editor.css
+++ b/mayland/style-editor.css
@@ -192,7 +192,7 @@ p {
 }
 
 a {
-	color: #000000;
+	color: var(--wp--style--color--link, #000000);
 }
 
 a:hover {
--- a/mayland/style-editor.css
+++ b/mayland/style-editor.css
@@ -192,7 +192,7 @@ p {
 }
 
 a {
-	color: var(--wp--style--color--link, #000000);
+	color: #000000;
 }
 
 a:hover {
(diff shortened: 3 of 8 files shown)


jffng:
At first I thought I needed to do [this](https://github.com/Automattic/themes/pull/3225/commits/4242a059343dde04489b9dbf0b254876ce936297#diff-2eed8a5a3e8be8ecb0f3a9855056c152e5ce24a6094169ba42be5d08b5d9dee9R24), but it's unnecessary because Gutenberg handles applying the custom link styles cleanly.

pbking:
According to the [doc](https://github.com/WordPress/gutenberg/blob/755ef781166c2ef60fe44c0af465baa3acb25b3d/docs/designers-developers/developers/themes/theme-support.md#experimental--link-color-control) (and mirroring the [work in spearhead](https://github.com/Automattic/themes/pull/3122)) I believe the experimental-theme.json will need to be added for this to work correctly. (Otherwise I believe the link colors will not be colored correctly by default.)

Which I think will mean that this change will require adding experimental-theme.json to ALL of the Varia children.

jffng:
> If a theme opts in, it should define default link colors in experimental-theme.json (**or in its theme styles if no experimental-theme.json is present**)

It should work without the theme.json, did you try it?

scruffian:
I think we need the theme.json to set a default color for links...

scruffian (approved):
LGTM

jffng:
After testing more use cases, I think we should only add this theme support to themes that have implemented CSS variables — which is only Hever to my knowledge.

My reasoning is because Gutenberg makes use of CSS variables (e.g. `--wp--preset--color--primary`) to apply custom link colors when a preset color is chosen in the editor. If we add the flag to Varia, custom link styles would not be applied correctly in any child theme without those preset values — either coming from theme.json or its stylesheet. And I don't think we should endeavor to add those variables via this PR.

Thoughts @scruffian @pbking @MaggieCabrera ?

scruffian:
So what would happen in other themes? Couldn't we just add theme.json to all the child themes?

jffng:
> Couldn't we just add theme.json to all the child themes?

I think this would break link colors for themes using a custom palette / color annotations. 

I'm having trouble getting this to work on dotcom, but it is working locally. It may have to do with the changes to the shape of the theme.json.

pbking:
Works fine locally.
Doesn't work as expected on .com (yet, until Gutenberg 9.9 is available there)

scruffian:
Moving this to blocked while we wait for 9.9 to launch on wpcom

jjchrisdiehl:
> Moving this to blocked while we wait for 9.9 to launch on wpcom

@scruffian  It looks like 9.9 was updated on atomic sites, is this change likely to be merged soon?

scruffian:
This was done in #3327

PR #3237, opened 2021-02-10
Add simple wp-env config file

pbking:
Added a simple wp-env file to import all of the themes into a wp-env
instance.

Added .wp-env.override.json to .gitignore so that those local
customizations (mostly pointing to my local gutenberg instance) will be
safely ignored


scruffian (approved):
LGTM

PR #3241, opened 2021-02-11
Stow: Add color annotations support

MaggieCabrera:

This PR adds support for color annotations on Stow using the CSS variable approach that is already included on Hever.

Screenshot 2021-02-11 at 13 52 12
Screenshot 2021-02-11 at 13 52 42


MaggieCabrera:
The third palette has some issues if we follow the same color mapping that the theme has by default (that's a pullquote block on the screenshot that is using foreground color):

Screenshot 2021-02-11 at 13 49 15

@kjellr what has been done in cases like this?

MaggieCabrera:
> The third palette has some issues if we follow the same color mapping that the theme has by default (that's a pullquote block on the screenshot that is using foreground color):

This looks better with Ben's commit, still very low contrast with the yellow links on the same palette:

Screenshot 2021-02-11 at 14 15 36

kjellr:
Those already don't pass contrast tests in the theme's default palette. I'd suggest moving forward and filing a separate issue to fix that.

scruffian (approved):
LGTM

PR #3244, opened 2021-02-11
Enhance/3213 spearhead toggle dark enabled

### LINKED ISSUE #3213: Spearhead: Add a toggle to disable automatic switching to dark mode
We've run into some issues with our podcasting patterns and Spearhead running in dark mode. Basically, we have two starter designs [Gilbert](https://gilbertstarter.wordpress.com/) and [Riley](https://rileystarter.wordpress.com/), and they are designed with white (or at least light) background in mind. So when you preview them with dark mode on, they look quite broken. In theory, we could design a set of alternate dark patterns, but there's no way to automatically switch to them. 

Would it be possible to have an option to turn off the automatic switching to dark mode in Spearhead?

## Steps to replicate

1. Go to system preferences > general and pick Dark appearance.
2. Preview [Gilbert](https://gilbertstarter.wordpress.com/) and [Riley](https://rileystarter.wordpress.com/) starter sites.

## Result
Gilbert:
image

Found through my own testing.

pbking:

This change moves all of the "dark mode stuff" to a single CSS file which is optionally loaded based on a new color_darkmode_disable theme_mod.

Like the informational panel that the checkbox panel replaces this control is only shown if the user has made NO customizations to the default color palette.  

This control is ONLY offered in the wp.com color customizer.

image

#### Related issue(s):

Fixes #3213


scruffian (approved):
This is brilliant!

There are only a couple of things I might change:

1. Make the setting available on .org installs
2. When a custom color is selected we automatically check the checkbox.

Both of these things can happen in a different PR though.

pbking:
Diff generated: D57024-code

pbking:
Regarding the "check this box automatically" we should disable it too, since dark mode isn't an option at all then.  Or just make it go away completely automatically. (Reload the customizer page in that situation and it should be gone).

PR #3251, opened 2021-02-12
Stratford: Add color annotations

scruffian:
Adds color annotations for Stratford. To test, load this onto your sandbox and try changing palettes.

### DIFF
--- a/stratford/functions.php
+++ b/stratford/functions.php
@@ -9,6 +9,18 @@
  * @since 1.0.0
  */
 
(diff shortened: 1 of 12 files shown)


jffng (approved):
This works in my testing.

I did notice a potential visual bug that occurs when the first element is a different background color than the main background color: 

Screen Shot 2021-02-12 at 10 41 27 AM

I think it can be addressed separately if this actually a bug or just a side effect of setting a different background color on the first element.

PR #3252, opened 2021-02-12
Shawburn: Add color annotations

scruffian:
Adds color annotations for Shawburn. To test, load this onto your sandbox and try changing palettes.

### DIFF
--- a/shawburn/functions.php
+++ b/shawburn/functions.php
@@ -9,6 +9,18 @@
  * @since 1.0.0
  */
 
+if ( ! function_exists( 'varia_default_colors' ) ) {
+	function varia_default_colors() {
+		return array(
+			'background' => '#FFFFFF',
+			'foreground' => '#444444',
+			'primary'    => '#0C80A1',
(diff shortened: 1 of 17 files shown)


pbking:
If I load the view after applying the change, but have not yet visited the customizer or editor my colors appear to be absent.  

image
image
However, once I click "default" in the customizer the site appears as expected.
I imagine that if I were to [Save Changes] then my view would be working fine (though I have abstained to keep my .com instance in a reproducible state). 
image

scruffian:
Thanks for finding that. I've fixed it in 610fa9c.

pbking:
I found that using the 1st palette the customizer shows a menu color scheme of:
image
while the view:
image

pbking:
When the changes in fix/rgb_to_hslv and add/color-annotations-for-shawburn are combined the issue noted above goes away.  Otherwise I see no further errors.  Once the PR is cleaned up the changes (in both branches) seem G2G.

PR #3260, opened 2021-02-12
Color Annotations: Check for the existence of color annotations in a more robust way

### LINKED ISSUE #3231: Mayland: Color palette in the editor looks broken
The theme color palette swatches in the editor are all the same.

## Steps to replicate
1. On a site with Mayland theme active, create a post or a page.
2. Add a block with color options, for example, Buttons, Paragraph, or Group. 
3. Look at the color settings panel in the sidebar. Note all the color swatches are the same (black).

image

Found through my own testing.

scruffian:
On .org sites get_theme_mod returns `false` when there is nothing set, but on Simple and Atomic sites it returns an array. This makes the check for the existence of color annotations more robust, so that it works if no color annotations are set.

#### Testing instructions
- Create a simple site with no color annotations and visit the editor - you won't see any editor palettes.
- Or run wpcomsh locally and see the same thing.

#### Related issue(s):
Fixes #3231

### DIFF
--- a/alves/functions.php
+++ b/alves/functions.php
@@ -59,12 +59,12 @@ if ( ! function_exists( 'alves_setup' ) ) :
 		 * - if the customizer color is empty, use the default
 		 */
 		$colors_array     = get_theme_mod( 'colors_manager' ); // color annotations array()
-		$primary          = ! empty( $colors_array ) ? $colors_array['colors']['link'] : '#3E7D98'; // $config-global--color-primary-default;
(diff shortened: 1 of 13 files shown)


MaggieCabrera (approved):
I tested this on a simple site, on a local installation, on a business site, and on atomic. I found that the color palette was working on the editor and the color annotations were fine for the business and atomic too. I didn't find any fatals or warnings in any case. This LGTM.

PR #3274, opened 2021-02-15
Update/3243 seedlet link color (refactor theme.json)

### LINKED ISSUE #3243: Seedlet: Experimental link color isn't working correctly.
In Seedlet when you select a link color from the palette, it doesn't work:

Screenshot 2021-02-11 at 17 00 57

pbking:
Refactored the experimental-theme.json introduced in #3271 to fix #3243 to fit the structure for the
change as of Gutenberg 9.9

### DIFF
--- a/seedlet/experimental-theme.json
+++ b/seedlet/experimental-theme.json
@@ -1,0 +1,9 @@
+{
+	"global": {
+		"styles": {
+			"color": {
+				"link": "var( --global--color-primary )"
+			}
+		}
+	}
+}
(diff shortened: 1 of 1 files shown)


jffng (approved):
This LGTM. 

Just noting there is a bug in the view when a text color and a link color are both set — the text color takes precedence because of rules of like [this](https://github.com/Automattic/themes/blob/trunk/seedlet/assets/sass/blocks/utilities/_style.scss#L208)

pbking:
Moving to 'blocked' until Gutenberg 9.9 is on WP.com

Issue #3280, opened 2021-02-15
Multiple Themes: Editor font styles not loading in Gutenberg v9.9.3


kjellr:
## Steps to replicate

Many of our older themes target a classname that doesn't appear to exist anymore in the editor, which results in their font styles not being applied. Here's an example from AltoFocus: 

https://github.com/Automattic/themes/blob/3fa9d68c06fb4d174aebe58f58952061839efbb2/altofocus/assets/stylesheets/editor-blocks.css#L22-L28

This targets `editor-block-list__block`, which no longer exists in Gutenberg v9.9.3 as far as I can tell. Changing it to `block-editor-block-list__block` works. (Or more preferably, these styles should be enqueued via `add_editor_style()` and the font styles should be applied to `editor-styles-wrapper`). 

This is also a problem in: 

- Independent Publisher 2
- Penscratch 2
- Libre 2
- ... and likely [many other older themes](https://github.com/Automattic/themes/search?q=editor-block-list__block). 

## Result

The editor loads the browser default for font + font size, or it uses Noto Serif. 

Safari:
![Screen Shot 2021-02-15 at 4 08 18 PM](https://user-images.githubusercontent.com/1202812/107994044-d6460000-6fa9-11eb-8f4e-b0826537fddf.png)

Chrome:
![Screen Shot 2021-02-15 at 4 11 21 PM](https://user-images.githubusercontent.com/1202812/107994045-d6460000-6fa9-11eb-9487-6ce79d0910a7.png)

## Expected

The theme should be using the Libre Baskerville font stack as specified in the stylesheet.

xpurichan:
Experienced this in the Veni theme on a customer's website in 27820699-hc. I wanted to note that this issue is present in the customizer (not only in the block editor).

We can even see some oddities when we go to edit the Veni theme demo site in the customizer (not the different fonts used for the "featured pages" description texts:
![Screen Shot 2021-02-17 at 2 50 59 PM](https://user-images.githubusercontent.com/33190975/108260004-66f61a80-7127-11eb-9519-0e1e48917f0c.jpg)

Versus what we see when we visit the Veni theme demo site at https://venidemo.wordpress.com/:
![Screen Shot 2021-02-17 at 2 51 12 PM](https://user-images.githubusercontent.com/33190975/108259990-62316680-7127-11eb-93d8-91c18b511ae7.jpg)

donalirl:
[...]

PR #3283, opened 2021-02-16
Independent Publisher: Add theme styles to the editor so that fonts match

scruffian:
This adds the theme styles to the editor so that the frontend and the editor match. Ideally we wouldn't need editor styles at all, but the post title requires them, and maybe other blocks do to...

#### Related issue(s):
Part of #3280


jffng (approved):
This fixes the issue of the incorrect fonts loading in the editor.

jffng:
Should the minor version number be bumped?

scruffian:
Changed to local file references and bumped the version number.

scruffian:
I also removed the `.edit-post-visual-editor` selector from editor-blocks.css. When this file is added as an editor style then the `.editor-styles-wrapper` prefix is added, so `.edit-post-visual-editor` won't work. I also renamed `.editor-block-list__block` to `.block-editor-block-list__block` since the class has changed names in Gutenberg.

scruffian:
I updated the approach here to remove as many editor styles as possible. This is ready for another review.

jffng:
This is being affected in the same way that Penscratch was, where some styles are leaking into the editor UI:

Screen Shot 2021-02-19 at 9 12 20 AM

PR #3284, opened 2021-02-16
Alto Focus: Update editor styles

scruffian:
This adds theme support for editor styles and updates the editor styles so that the frontend and the editor match. Ideally we wouldn't need editor styles at all, but there's no consistent way to target editor content at the moment.

#### Related issue(s):
Part of #3280

### DIFF
--- a/altofocus/assets/stylesheets/editor-blocks.css
+++ b/altofocus/assets/stylesheets/editor-blocks.css
@@ -19,7 +19,7 @@ Description: Used to style Gutenberg Blocks in the editor.
 1.0 General Typography
 --------------------------------------------------------------*/
 
-.edit-post-visual-editor .block-editor-block-list__block,
+.block-editor-block-list__block,
 .editor-default-block-appender textarea.editor-default-block-appender__content {
 	font-size: 13.875px;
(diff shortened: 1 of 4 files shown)


jffng (commented):
This works for me locally but I cannot verify on wpcom. The editor styles are not loading there for me, and I do not not know why. I tried changing the selectors to be more generic, and the url to `get_theme_file_uri('/assets/stylesheets/editor-blocks.css')` but no luck.

scruffian:
To get this working on your wpcom sandbox you need to point remote requests from your sandbox. You can use this code to do that: 286f7-pb/#php. You'll need to add your test site url to the array for it to work.

jffng:
🚢

scruffian:
I changed the approach here. By using local file references rather than URLs we can avoid the environmental issues above.

I also decided to try to move closer to having the same styles for front end and the editor. We now add all style to the editor. I have removed duplicated styles from the editor-blocks.css file. Ideally we'd only have one file that works for both, but that would require a lot of reviewing and testing. This is a first step towards that

MaggieCabrera (approved):
LGTM!

PR #3285, opened 2021-02-16
Penscratch: Update editor styles

scruffian:
Updates editor styles so that they work with Gutenberg 9.9.

It would be better to just use the normal theme styles but the classes in Gutenberg don't match the front end :(

#### Related issue(s):
Part of #3280


jffng (commented):
Same as #3284, works locally but cannot verify on wpcom.

scruffian:
I updated this to:
- Reference the files with a local reference not a URL
- Queue all CSS in the editor, not just the editor blocks
- Add a rule to make the body of the editor white
- Remove CSS selectors that aren't needed and update old ones to new ones...

MaggieCabrera:
Looks like the headings are not working on the editor: 

Frontend | Editor
--- | ---
Screenshot 2021-02-18 at 11 36 14 | Screenshot 2021-02-18 at 11 36 10

scruffian:
I took a different approach with this. I have removed almost all editor styles, and just kept one for the body content.

This meant that the reset was leaking into the editor and breaking buttons. I have removed the reset from the main stylesheet and queued it separately in the front end.

penscratch-2/css/editor-blocks.css
```diff
@@ -19,560 +19,6 @@ Description: Used to style Gutenberg Blocks in the editor.
 1.0 General Typography
 --------------------------------------------------------------*/
 
-.edit-post-visual-editor .editor-block-list__block,
-.editor-default-block-appender textarea.editor-default-block-appender__content {
-	font-family: "Roboto Slab", Georgia, Times, serif;
+body {
+	background: #fff;
```
jffng, line 23: It's kind of a shame that this line is needed just to override the default background color, otherwise we could get rid of this file entirely.

jffng:
The fonts are loading correctly but I noticed some regressions and styles leaking into the editor UI:

Before | This PR
--------- | ---------
Screen Shot 2021-02-18 at 11 28 04 AM | Screen Shot 2021-02-18 at 11 29 16 AM

penscratch-2/css/editor-blocks.css
scruffian: We could add a Gutenberg class to the CSS that targets the site content, but I don't want to start mixing editor CSS into the main file.

scruffian:
Thanks for testing. I moved the form CSS into a separate file, which fixes the problems you found.

MaggieCabrera (approved):
I think this is ready to ⛵

PR #3287, opened 2021-02-16
Blank Canvas + Seedlet: refactor so we can reuse parts of Blank Canvas in other child themes

jffng:
This PR moves parts of Blank Canvas to Seedlet. The goal of this PR is to set us up to create / generate child themes of Seedlet, based on Blank Canvas, that are more DRY and hopefully easier to maintain. 

It also consolidates Blank Canvas' customizer functionality which was previously scattered in a few places into a single class.

To test, verify that Blank Canvas still works locally and on dotcom. I'd also love any feedback on the approach and if it makes sense why we would make this change.


blank-canvas/functions.php
```diff
@@ -187,30 +102,6 @@ function blank_canvas_customizer_enqueue() {
...
-	}
-
-	if ( false === get_theme_mod( 'show_comments', false ) ) {
-		$classes[] = 'hide-comments';
-	}
-
-	return $classes;
-}
-add_filter( 'body_class', 'blank_canvas_body_classes' );
+require get_template_directory() . '/inc/blank-canvas.php';
```
jffng, line 107: To create a child theme of Seedlet that works like Blank Canvas, we can just include this file.

scruffian:
What do you think about making Blank Canvas the parent?

jffng:
I like that idea! But it raises the question — why is Blank Canvas a child of Seedlet?

scruffian:
Exactly! If Blank Canvas was the parent then all these themes could inherit from it so they'd be even simpler.

jffng:
Is your suggestion to make Blank Canvas a block theme, or a just a classic parent theme from which we could make child block themes?

MaggieCabrera:
If we'd make Blank Canvas the parent theme (as a classic theme) I think we could improve/simplify it quite a bit to avoid all the issues we've encountered while creating Seedlet children. I'm talking more on the css side of things in this case. I still think that blank canvas/link in bio should be block-based and we could make them work as such, but there's value in having a simpler, cleaner slate for classic themes should we (or others) need to build more classic child themes.

scruffian:
Yeah I'd much prefer to just build them as block themes and not have to worry about this, but if that isn't an option then making blank canvas a parent theme might be an alternative approach.

jffng:
Thank you! Closing this PR, because it seems like the better plan is to make Blank Canvas the parent. Whether or not it can be a block theme is a separate exploration.

PR #3295, opened 2021-02-17
Link In Bio: try as a block theme

jffng:
This PR adds Link in Bio as a block theme child of Seedlet. It's an alternative to https://github.com/Automattic/themes/pull/2363


linkinbio/block-templates/singular.html
```diff
@@ -0,0 +1,9 @@
+<!-- wp:group {"tagName":"main", "align":"full"} -->
+<main class="wp-block-group alignfull">
+	<div class="wp-block-group__inner-container">
+
+	<!-- wp:post-content {"align":"full"} /-->
+
```
jffng, line 6: Maybe worth adding a post comments block here.

PR #3296, opened 2021-02-18
Experiment using Block-Based templates via PHP

pbking:
An experiment leveraging HTML templates by way of PHP.

Uses do_blocks w/ file_get_contents to load the HTML templates with classic PHP templates instead of loading the FSE templates directly.


jffng:
What are the advantages of this approach?

In general, I see `do_blocks` as a means for existing themes with complicated template logic to migrate to block themes.

pbking:
> What are the advantages of this approach?

* It allows for the building of block-based themes (or just MOSTLY block-based themes) without involving the Full Site Editor.  
* PHP can still be used to fill in whatever gaps remain.
* Classic navigation can be integrated until a smoother migration there is possible
* The Customizer can still be used as needed.  (That being a pro or a con depends I suppose...)

jffng:
> It allows for the building of block-based themes (or just MOSTLY block-based themes) without involving the Full Site Editor.

We don't actually need the full site editor to use block templates though.

> PHP can still be used to fill in whatever gaps remain.

What gaps exist with Link in Bio that we need php for? Translation? 

> Classic navigation can be integrated until a smoother migration there is possible

Yes good point, if we can't accept shipping a theme that doesn't support creating menus in the classic way.

> The Customizer can still be used as needed. (That being a pro or a con depends I suppose...)

If the customizer is loaded in an iframe, would updating something like the background color (if it's hooked into the correct CSS variables) still work with a block template?

pbking:
> If the customizer is loaded in an iframe, would updating something like the background color (if it's hooked into the correct CSS variables) still work with a block template?

Yes.  I've got other [experiments](https://github.com/WordPress/theme-experiments/pull/173) that demonstrate that it can work just fine. 

> What gaps exist with Link in Bio that we need php for? Translation?

I'm not sure that gaps exist for Link in Bio it was just a really easy theme to demonstrate the technique for.  
Translation, perhaps but I'm not sure that this technique would actually cover that; Translation may be a gap still...

> We don't NEED the FSE to use block templates.

I only mentioned that because using block templates automatically enables FSE and disables Customizer.  Using do_blocks gets around that restriction. You mentioned an idea to be able to disable the FSE (even when /block-templates are in play) and if that's possible (as a theme_supports/flag-in-theme.json) along with the same for the Customizer then leveraging do_blocks instead of the templates directly would be necessary in LESS circumstances.

PR #3305, opened 2021-02-18
Libre 2: Remove editor styles

scruffian:
Removes editor styles from Libre 2, and just loads regular styles into the editor.

#### Related issue(s):
Part of #3280


libre-2/css/editor-blocks.css
```diff
@@ -1,535 +0,0 @@
...
-.wp-block {
-	max-width: 705px; /* 675px + 30px to account for padding */
-}
-
-/* Link styles */
-
-.edit-post-visual-editor a,
-.editor-block-list__block a,
-.wp-block-freeform.block-library-rich-text__tinymce a {
-	box-shadow: 0 1px 0 0 currentColor;
```
MaggieCabrera, line 130: This has been lost from the editor, but before your change, it wasn't exactly like the frontend either... It's just that the frontend targets `.entry-content a` so that will never work on the editor

MaggieCabrera:
Quote blocks look much better now than they did without this change but they are still loading some extra styles on the editor that are different from the frontend. They come from `gutenberg/build/block-library/theme.css` and the main issue is the border showing on the editor that is not showing up on the frontend:

Editor:

Screenshot 2021-02-19 at 10 38 46

Frontend:

Screenshot 2021-02-19 at 10 38 52

MaggieCabrera:
The latest comments block is miss aligned on the editor:

Screenshot 2021-02-19 at 10 44 57

MaggieCabrera:
The pullquote block is showing similar borders on the editor too:

Screenshot 2021-02-19 at 10 48 04

MaggieCabrera:
The gallery blocks are misaligned and the captions are a bit broken on the editor:
Screenshot 2021-02-19 at 10 49 44

scruffian:
I added some fixes to override the settings coming from theme.css. This is still not ideal and the front and back ends don't match up perfectly, but maybe it's good enough for now...

MaggieCabrera (approved):
I think this looks so much better now than it did before the PR. Great job @scruffian. I say ⛵

PR #3308, opened 2021-02-18
Blank Canvas: Add support for comments on Pages

### LINKED ISSUE #3276: Blank Canvas Theme - comments enabled but not working
## Steps to replicate

- Checked all comment related settings under Settings > Discussion, nothing should be preventing comments on the site
- Screenshot of user's Discussion settings: https://d.pr/i/fB0se6
- Allow Comments is checked on their post (post is published) 
- Site is currently set to Private, but the user made it Public while we were testing and it still didn't allow comments. 

## Result

- Even though settings should have been set to allow comments, they weren't working. 

## Expected

Comments to be enabled and allowed. 

* User report: [shortlink to chat/forum thread/ticket, ie 27762571-hc]
* Follow-up ticket: [shortlink to ticket, ie 3742086-zen]

danieldudzic:
Closes #3276

### DIFF
--- a/blank-canvas/page.php
+++ b/blank-canvas/page.php
@@ -20,6 +20,13 @@ get_header();
 				the_post();
 
 				get_template_part( 'template-parts/content/content-singular' );
+				
+				if ( true === get_theme_mod( 'show_comments', false ) ) :
+					// If comments are open or we have at least one comment, load up the comment template.
+					if ( comments_open() || get_comments_number() ) {
+						comments_template();
+					}
+				endif; 
 
 			endwhile; // End of the loop.
 			?>


jffng:
Is this something we want to allow?

Blank Canvas already explicitly supports opting in to display comments on _Posts_, not Pages. I get that the distinction between posts and pages is confusing to users, but this seems like an intentional choice to not show comments on pages.

cc @kjellr

danieldudzic:
Having comments on Posts and Pages seems to be the standard in (most) themes, but I'd be interested to hear from @kjellr too.

kjellr:
Oh weird, is that the issue? I thought the user was experiencing an issue with comments now showing up on posts (I double checked that and couldn't reproduce though). 

It seems really weird to me to enable comments on pages, but since Seedlet + Varia both do it I suppose we should enable it. We should update this text though: 

Screen Shot 2021-02-19 at 8 17 22 AM

Maybe it should say: "Check to show comments underneath each post and page."

Open to other suggestions though.

jffng:
Thanks for confirming, the language suggestion sounds good.

kjellr (approved):
I'll add the message in #3315. This looks good in the meantime. 👍

Issue #3134, opened 2021-01-28
Seedlet: Moar Extensibility


georgestephanis:
When child theming seedlet, it'd be tremendously useful if there was a filter that the child theme could use to change which Google Fonts are pulled in (basically filtering the args before they're add_query_arg'd in here:

https://github.com/Automattic/themes/blob/06314c27f86c4f30a78f559eb82022e14a7d5f71/seedlet/functions.php#L307-L312

Also, if it would be possible to filter the default colors here:

https://github.com/Automattic/themes/blob/06314c27f86c4f30a78f559eb82022e14a7d5f71/seedlet/functions.php#L156-L162

that would far simplify things to match a site's branding without having to override a bunch of additional code.

Finally -- if possible, if there's a way for a child theme to add extra svg assets that could be loaded in via the Seedlet svg functions, it would provide a much more consistent devex -- but that's more wishlist.

** minor extra other bit: there's a bunch of very minor phpcs issues in Seedlet that the WordPress rulesets flag -- stuff like some conditionals not being 'yoda' style, or indenting style not being consistent or the like that may be worth tidying at some point, but that's not really bothering anyone apart from our build tools whenever we load it in to child theme.

<3

georgestephanis:
Also noting here that for overriding the colors in a child theme, I did so like so:

```php
/**
 * Create a "Colors Manager" class to short out the Seedlet colors customizer stuff.
 */
class Colors_Manager {}

/**
 * Hardcode the values.
 */
add_filter( 'theme_mod_custom_colors_active', '__return_null' );

add_filter( 'theme_mod_seedlet_--global--color-primary',    function() { return '#24446b'; } );
add_filter( 'theme_mod_seedlet_--global--color-secondary',  function() { return '#c3e0ec'; } );
add_filter( 'theme_mod_seedlet_--global--color-tertiary',   function() { return '#f4d03e'; } );
add_filter( 'theme_mod_seedlet_--global--color-foreground', function() { return '#1f1f1f'; } );
add_filter( 'theme_mod_seedlet_--global--color-background', function() { return '#ffffff'; } );
```

PR #3309, opened 2021-02-18
Add filter to change bundled google fonts in assorted themes.

georgestephanis:

Initially intended as a part-fix for #3134 to enable child themes to swap typefaces that get included more easily, this also extends that change across to many themes using the common patterns.

#### It explicitly does not change the following themes:

* spearhead (uses different endpoint and method and is already a child theme)

And the following themes that have multiple functions that generate typeface includes so as to avoid confusion:

* canard
* gazette
* illustratr
* publication

### DIFF
--- a/affinity/functions.php
+++ b/affinity/functions.php
@@ -248,6 +248,13 @@ function affinity_fonts_url() {
 			$font_families[] = 'Lora:400,400italic,700,700italic';
 		}
 
+		/**
+		 * A filter to enable child themes to add/change/omit font families.
+		 * 
+		 * @param array $font_families An array of font families to be imploded for the Google Font API
+		 */
+		$font_families = apply_filters( 'included_google_font_families', $font_families );
+
 		$query_args = array(
 			'family' => urlencode( implode( '|', $font_families ) ),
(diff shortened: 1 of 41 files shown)


georgestephanis:
I left the namespace for the filter the same across themes, but would have no objection to namespacing each filter for the theme it was in specifically.

jffng:
Hey @georgestephanis ! Thanks for the PR.

Does this change need to be propogated to themes other than Seedlet? I ask because many of the themes you've added the filter to are Varia themes — which are not technically child themes, but are not really meant to be parent themes. If it will make it easier to override those themes fonts, then I think it's a pretty harmless change, but just want to make sure it's necessary.

georgestephanis:
As far as necessity goes, I personally only need it to be Seedlet.

The rest of them were figuring it may as well make folks lives easier.  But I'm 100% fine trimming this back down to just Seedlet.

jffng:
Gotcha — let's stick to just Seedlet and this LGTM 🚢

scruffian:
FWIW I think it's fine to add to all themes..

PR #3310, opened 2021-02-18
Dara: Update editor styles

scruffian:
This updates Dara to use the main style sheets as editor styles with exceptions for the title and content background.

#### Related issue(s):
Part of #3280


pbking:
I appears that due to [this recent Gutenberg](https://github.com/WordPress/gutenberg/pull/28514/files#diff-7bf60a59a90bcf0f53fa61af0e1717f2db7e31ff7bca1c0eb2e17389a351e1efL11) change that default Button Block styles now receive a black border by default (though, in the view only, not the editor... which seems EXTRA strange).
image

Issue opened to address that [opened here](https://github.com/WordPress/gutenberg/issues/29149).

pbking:
There appear to be alignment discrepancies between the view and the editor.  I don't believe this is due to this change as it's the view that appears to be incorrect.  Just noting it...

Editor:
image

View:
image

pbking:
The editor appears to have lost the padding in the Cover Block.

Editor Before:
image

Editor After:
image

View:
image


Note that the cover block is pink in the view and black in the editor, but this is something that existed before the change.

PR #3314, opened 2021-02-19
Blank Canvas: Theme Review Updates

kjellr:
Includes a couple updates related to [the WordPress.org theme review](https://themes.trac.wordpress.org/ticket/94482) of Blank Canvas. 

> What's the thinking behind changing the info panel in the site identity section to look like a notice? You shouldn't really be changing the styles in the customizer for existing core items. If you think this should be displayed as a notice instead then open a core ticket and suggest that :)

I've removed the styles for this by default, but moved them over so that they'll still display on WordPress.com. Since we'd seen users miss this message in testing, I think we still want to display it as a notice on WP.com. 

Before|After (Self-hosted only)
---|---
Screen Shot 2021-02-19 at 9 50 48 AM|Screen Shot 2021-02-19 at 9 50 25 AM

---

> You should add a screenshot that better represents what a user would see when they use your theme. At no point when testing the theme did I have a blank white page. There was always some kind of content visible - even if that was just a the default 'hello world' post. I don't think it's realistic for users to have this theme installed with zero content for very long if at all.

I've added an alternate screenshot that we can use for WP.org. I _think_ this should satisfy the requirements, but if for some reason it doesn't, we can just use a screenshot of one of the bundled patterns I guess.

The new screenshot is named `screenshot-alt.png`, and we'd need to change the name for the upload. I think the WP.com theme screenshot should still be empty unless we decided to change it later.

Before|After
---|---
screenshot|screenshot-alt

---

This also bumps the version, which is required for a new upload. I've switched to using a `-wpcom` version suffix here like we do with our other themes.


kjellr:
I added one more small change: I added a clarifying note to the Site Header + Menus setting, to point users to other panels to manage these options. 

Before|After
---|---
Screen Shot 2021-02-19 at 10 12 22 AM|Screen Shot 2021-02-19 at 10 12 02 AM

MaggieCabrera (approved):
This works as expected with the differences to the notice on .org and .com. I personally think that this alternative screenshot works well for this theme 👍

kjellr:
Thank you!

PR #3133, opened 2021-01-28
Update shape of theme.json file

pbking:
#### Changes proposed in this Pull Request:

In accordance with
https://github.com/WordPress/gutenberg/pull/28110

the theme.json file has been updated to reflect the new expected shape.

This was tested locally with Gutenberg 9.8.2 on the /master branch.  This will NOT work on .com until that version of Gutenberg has been deployed there.  However ONCE IT IS then fixes for #3106 and #3122 will no longer work.

#### Related issue(s):

#3106 are related as is #3122 

https://github.com/WordPress/theme-experiments/pull/182 is similar but for TT1-blocks.

PR #3325, opened 2021-02-22
Seedlet & Spearhead Custom Link Color

pbking:

Changes the shape (or adds) theme.json files from the following Pull Requests:

Seedlet:
https://github.com/Automattic/themes/pull/3271
https://github.com/Automattic/themes/pull/3274

Spearhead:
https://github.com/Automattic/themes/pull/3133

#### Related issue(s):
https://github.com/Automattic/themes/issues/3243

#### Diff
D57441-code

### DIFF
--- a/seedlet/assets/css/style-editor.css
+++ b/seedlet/assets/css/style-editor.css
@@ -315,6 +315,10 @@ a:active {
 	border-bottom-color: transparent;
 }
 
+.has-primary-background-color a:not(.has-link-color):not(.has-text-color) {
+	color: var(--global--color-background);
+}
+
 .has-background:not(.has-background-background-color) .has-link-color a,
 .has-background:not(.has-background-background-color).has-link-color a {
 	color: var(--wp--style--color--link, var(--global--color-primary));
--- a/seedlet/assets/css/style-editor.css
+++ b/seedlet/assets/css/style-editor.css
@@ -315,7 +315,7 @@ a:active {
 	border-bottom-color: transparent;
 }
 
-.has-primary-background-color a:not(.has-link-color) {
+.has-primary-background-color a:not(.has-link-color):not(.has-text-color) {
 	color: var(--global--color-background);
 }
 
--- a/seedlet/assets/css/style-editor.css
+++ b/seedlet/assets/css/style-editor.css
@@ -315,13 +315,8 @@ a:active {
 	border-bottom-color: transparent;
 }
 
-.has-primary-background-color a:not(.has-link-color):not(.has-text-color) {
+.has-primary-background-color a:not(.has-link-color) {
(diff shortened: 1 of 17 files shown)


jffng:
Custom link colors are still not working for me in the view:

Editor | View
------- | ------
Screen Shot 2021-02-22 at 3 49 10 PM | Screen Shot 2021-02-22 at 3 49 17 PM

scruffian:
It's not working in the view for me either - I tested in a cover block...

pbking:
I found and fixed an [issue with the cover block specifically](https://github.com/Automattic/themes/pull/3325/files#diff-0518cab2ad99da1c18c91f1dd4a82bc31b1d8ddb55e49325d76c0370c96c4017R17) where the specificity of some anchor styling was preventing the 'link' style from cascading properly.  

I suspect there are other blocks with the same situation, I'll comb through spearhead and see where else it might be occurring. 

@jeffikus I was unable to reproduce your screenshot however (tested both Spearhead and Seedlet).  It only KINDA happened for me when I was testing and had SEEDLET in the editor, switched themes (but didn't reload the editor) and set the color to that green (secondary color) which isn't a color provided by Spearhead so it rolled back to the 'default'.  But that seems like I was trying too hard...

pbking:
Shoot...
The "cover block" fix noted above breaks styles when a custom link is NOT applied.  Need to re-evaluate further...

scruffian:
I pushed a fix to Seedlet colors for this, just needs another review...

MaggieCabrera:
A cover block with custom link color shows differently on the editor vs the frontend:

**Editor:**
Screenshot 2021-03-02 at 10 19 44

**Frontend:**
Screenshot 2021-03-02 at 10 19 40

scruffian:
Yeah sorry I forgot to update the editor styles too. Should be good to go now...

pbking:
In the editor, in a GROUP with background {Primary}
A P with link-color set to {Foreground} renders correctly in the editor
image
but not so in the view:
image
Perhaps .has-link-color should be included in 
`.has-background:not(.has-background-background-color) a:not(.has-text-color):not(.wp-block-button__link)` which is coloring the link to 'currentColor'.

scruffian:
This will be a lot easier to resolve once this is fixed: https://github.com/WordPress/gutenberg/pull/29557

scruffian:
I removed a bunch of code from Seedlet to fix this. Some of this code was added in this PR: https://github.com/Automattic/themes/pull/3088

I think if you run the latest version of Gutenberg it will all work as expected, but there could be more cases I've not considered...

Screenshot 2021-03-04 at 13 09 11

MaggieCabrera (approved):
I could not get this to break, it looks good to me!

PR #3326, opened 2021-02-22
Varia: fix button styles leaking into editor

### LINKED ISSUE #3319: Varia: Primary background color set on classic block toolbar
## Steps to replicate

1. Switch theme to a Varia theme (I tested Shawburn, Stow, Hever)
2. Add Classic block
3. See background color on toolbar

Screen Shot 2021-02-21 at 11 04 57 AM

## Result

The primary background color is on the toolbar buttons. This makes them very difficult to read/not very accessible. 

## Expected

For the default styles of the toolbar to show instead.

* User report: 3760468-zen
* Follow-up ticket: same as above

jffng:
This is a hotfix for https://github.com/Automattic/wp-calypso/issues/50332.

As a follow up I will also test https://github.com/WordPress/gutenberg/pull/29229 to see if it resolves the issue.

D57440-code

**To test**
- Activate Hever or another Varia theme
- Verify the button styles don't leak into the editor UI

### DIFF
--- a/varia/sass/blocks/button/_editor.scss
+++ b/varia/sass/blocks/button/_editor.scss
@@ -1,7 +1,6 @@
 /**
  * Button
  */
-input[type="submit"],
 .wp-block-button__link,
 .wp-block-file__button {
 	// Extend button style
--- a/varia/sass/blocks/button/_editor.scss
+++ b/varia/sass/blocks/button/_editor.scss
@@ -1,14 +1,12 @@
 /**
  * Button
  */
-button:not(.components-button),
-.button:not(.components-button),
- input[type="submit"],
- .wp-block-button__link,
- .wp-block-file__button {
-	 // Extend button style
-	 @extend %button-style;
- }
+input[type="submit"],
+.wp-block-button__link,
+.wp-block-file__button {
+	// Extend button style
+	@extend %button-style;
+}
 
 /* Default Style */
 .wp-block-button__link {
--- a/varia/sass/blocks/button/_editor.scss
+++ b/varia/sass/blocks/button/_editor.scss
@@ -1,8 +1,8 @@
 /**
  * Button
  */
- button,
- .button,
+button:not(.components-button),
+.button:not(.components-button),
  input[type="submit"],
  .wp-block-button__link,
  .wp-block-file__button {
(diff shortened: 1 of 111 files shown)


varia/sass/blocks/button/_editor.scss
```diff
@@ -1,8 +1,8 @@
 /**
  * Button
  */
- button,
```
jffng, line 4: These two lines are the only meaningful change.

varia/sass/blocks/button/_editor.scss
```diff
@@ -1,8 +1,8 @@
...
- .button,
+button:not(.components-button),
```
jffng: These two lines are the only meaningful change.

pbking:
This seems to be effecting the classic block button negatively and doesn't seem to address the issue of their discoloration:

Before Change:
image

After Change:
image

varia/sass/blocks/button/_editor.scss
```diff
@@ -1,14 +1,12 @@
...
  */
- button,
- .button,
- input[type="submit"],
- .wp-block-button__link,
- .wp-block-file__button {
-	 // Extend button style
-	 @extend %button-style;
- }
+input[type="submit"],
```
jffng: What if we just got rid of the generic targeting of buttons in the editor?

jffng:
Thanks for the review @pbking — I'm wondering if we can just remove the `button` and `.button` selector entirely in the editor. It seems too generic. On Mayland:

Screen Shot 2021-02-22 at 4 43 39 PM

varia/sass/blocks/button/_editor.scss
simison: That might be easiest, because `input`, `button` etc are highest on CSS specificity, and it's a bit hard to battle against it with `:not` selectors.

jffng: True, I'm going to remove the `input` selector as well. Looking at [how Seedlet handles button styles in the editor](https://github.com/Automattic/themes/blob/trunk/seedlet/assets/sass/blocks/button/_editor.scss), we don't target these elements at all.

jffng:
I updated the editor styles to remove generic targeting of generic `button` and `input` elements, and rebuilt all the child themes. This could use one more review.

pbking:
I can't break it!  Looks great to me!

A nice tiny golf change (that results in > 100 changed files!)

🚢

PR #3336, opened 2021-02-23
Varia: Fix link and button color settings

### LINKED ISSUE #3250: Varia + children: Links and Outline Buttons same color as primary background
## Steps to replicate

1. Varia or a child theme active
2. In post or page, add a block with background color set to primary color (e.g. cover, group, or media & text block)
2. Inside block add a regular link, or a button with outline (otherwise default colors)

## Result

The link and button are invisible because they are also showing the primary color.

Screenshot with Barnsbury:

barnsbury-button-color

Screenshot with Exford:

exford-button-color

Can be seen on my test site:
https://badaboomtest.wordpress.com/testing-colors/?theme=pub%2Fvaria
https://badaboomtest.wordpress.com/testing-colors/?theme=pub%2Fbarnsbury
https://badaboomtest.wordpress.com/testing-colors/?theme=pub%2Fexford

## Expected

The link and button color should contrast the background so it can be seen.

From what I can see, the themes all adjust the regular text colors for elements with background color using CSS such as:
```
.has-primary-background-color:not(.has-text-color),
.has-primary-background-color.has-background-dim:not(.has-text-color) {
	color: white;
}
```
but there is no corresponding CSS for links and buttons to override the primary color being applied to links and buttons in this situation.

A possible fix may be to just include links in the above CSS:

```
.has-primary-background-color:not(.has-text-color),
.has-primary-background-color:not(.has-text-color) a,
.has-primary-background-color.has-background-dim:not(.has-text-color),
.has-primary-background-color.has-background-dim:not(.has-text-color) a {
	color: white;
}
```

* User report: 27566071-hc

Maybe related to these issues, but not sure:
https://github.com/Automattic/themes/issues/2582
https://github.com/Automattic/themes/issues/3232

danieldudzic:
This hopefully fixes #3250

### DIFF
--- a/varia/sass/blocks/button/_editor.scss
+++ b/varia/sass/blocks/button/_editor.scss
@@ -27,6 +27,27 @@
 		color: #{map-deep-get($config-button, "color", "text-hover")};
 		background-color: #{map-deep-get($config-button, "color", "background-hover")};
 	}
+	
+	.has-background & {
+		&:not(.has-background) {
+			background: #{map-deep-get($config-button, "color", "text")};
+
+		}
+		&:not(.has-text-color) {
+			color: #{map-deep-get($config-button, "color", "background")};
+		}
+	}
+	
+	.has-background &.is-style-outline,	
+	.has-background .is-style-outline & {		
+		&:not(.has-background) {
+			background: transparent;
+		}
+		
+		&:not(.has-text-color) {
+			color: #{map-deep-get($config-button, "color", "text")};
+		}
+	}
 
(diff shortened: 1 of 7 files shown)


scruffian:
We removed this to fix another bug, I'm not sure which one now. I'm going to look into implementing a solution in Gutenberg so we don't have to keep fixing this in each theme...

scruffian:
I went for a different solution in the end: https://github.com/Automattic/themes/pull/3338

Let me know what you think...

PR #3339, opened 2021-02-23
Seedlet: Add underline to comment author/edit links on hover

### LINKED ISSUE #3330: Seedlet: Comment author/edit links need an underline on hover
In Seedlet, the comment action links currently do not have a visible indication that they've been hovered: 

Seedlet-Comments-Hover

These should use the same `border-bottom` border treatment we use for other links throughout the theme: 

Screen Shot 2021-02-22 at 3 03 43 PM

danieldudzic:
Closes #3330

### DIFF
--- a/seedlet/assets/sass/components/comments/_comment-form.scss
+++ b/seedlet/assets/sass/components/comments/_comment-form.scss
@@ -26,6 +26,7 @@
 		
 		&:hover,
 		&:focus {
+			color: var(--global--color-primary-hover);
 			border-bottom: 1px solid var(--global--color-secondary);
 		}
 	}
--- a/seedlet/assets/sass/components/comments/_comment-form.scss
+++ b/seedlet/assets/sass/components/comments/_comment-form.scss
@@ -23,6 +23,11 @@
 	.url {
 		border-bottom: none;
 		color: currentColor;
+		
+		&:hover,
+		&:focus {
+			border-bottom: 1px solid var(--global--color-secondary);
+		}
 	}
 }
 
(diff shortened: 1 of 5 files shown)


seedlet/assets/sass/style.scss
```diff
@@ -7,7 +7,7 @@ Description: Seedlet is a free WordPress theme. A two-column layout and classica
 Requires at least: 5.5
 Tested up to: 5.5
 Requires PHP: 5.6.2
-Version: 1.1.11-wpcom
+Version: 1.1.10-wpcom
```
scruffian: Shouldn't this go up?

seedlet/assets/sass/components/comments/_comment-form.scss
```diff
@@ -23,6 +23,11 @@
 	.url {
 		border-bottom: none;
```
scruffian, line 24: I wonder if we should just remove this line, so the behaviour is the same as on the rest of the theme, i.e. the link is underlined until it's hovered.

scruffian (changes requested):
I think we should make the behaviour consistent with the rest of the theme, and hide the underline on hover, but we should probably get a second opinion from @kjellr

kjellr (commented):
> I think we should make the behaviour consistent with the rest of the theme, and hide the underline on hover, but we should probably get a second opinion from @kjellr

I prefer it this way (underline only appears on hover) for utility links that appear close together like these ones do. We already do this for post meta. 👍 

The only issue I see in this PR is that the author link should be green on hover, just like the other links: 

hover

Once that's fixed, this should be good to go.

kjellr:
I just pushed a quick update sync up with `trunk` and also cover that text color:

hover

Should be all set, if someone can give it one last review.

scruffian (approved):
LGTM

PR #3343, opened 2021-02-23
Seedlet: fix keyboard navigation selector

### LINKED ISSUE #3331: Seedlet: Console + accessibility focus error when using mobile navigation
This was reported [in the theme review for Blank Canvas](https://themes.trac.wordpress.org/ticket/94482#comment:14), but it's about Seedlet's menu:

> When I add a menu to the primary navigation and test the responsive version, 
there is a JS error, but this is from the parent theme:

```
primary-navigation.js?ver=1.2.4:42 Uncaught DOMException: 
Failed to execute 'querySelector' on 'Document': '.${ id }-navigation' is not a valid selector
```

> When I tab through the menu, the focus is not kept inside the modal, instead, when I tab away from 
the last menu item, the focus seems to be on the content below the menu.
> I can't tell where I am on the page because the open menu is covering the content.
> Perhaps this is because of the JS error.

jffng:
This PR addresses #3331.

**To test**
- Activate Seedlet on mobile
- Navigate to the primary menu via keyboard, and verify that the tab navigation is trapped within the modal.
- Verify no console errors appear

Kapture 2021-02-23 at 14 30 52

### DIFF
--- a/seedlet/assets/js/primary-navigation.js
+++ b/seedlet/assets/js/primary-navigation.js
@@ -39,7 +39,7 @@
 			} 
 			var modal, elements, selectors, lastEl, firstEl, activeEl, tabKey, shiftKey, escKey;
 
-			modal = document.querySelector( '.${ id }-navigation' );
+			modal = document.querySelector( '.' + id + '-navigation' );
 			selectors = "input, a, button";
 			elements = modal.querySelectorAll( selectors );
 			elements = Array.prototype.slice.call( elements );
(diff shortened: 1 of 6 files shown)


scruffian (approved):
LGTM

We could also change the `'` to ``` but I think this is better for older browsers.

MaggieCabrera:
> LGTM
> 
> We could also change the `'` to ``` but I think this is better for older browsers.

I believe that's what it was doing before and it caused issues in IE, my change for IE may have been what caused this issue in the first place so this fix would work for both cases. Thanks @jffng!

jffng:
> We could also change the ' to ``` but I think this is better for older browsers.

Yeah I think when I first wrote this I used some ES6 features to be fancy without realizing this would break on IE, so sticking to `'` seems safer for this purpose.

kjellr (approved):
This works in my testing as well. Thank you!

PR #3345, opened 2021-02-23
Blank Canvas: clean up block patterns

### LINKED ISSUE #3332: Blank Canvas: Block patterns code cleanup
The [theme review for Blank Canvas](https://themes.trac.wordpress.org/ticket/94482#comment:14) noted a few things that need tidying up in the block patterns: 

- [ ] The text strings are currently translatable, but they need to be escaped too. `esc__attr()` for the attributes, and probably `esc__html()` for the text.
- [ ] We use `get_stylesheet_directory_uri()` to pull in the images. Those URLs need to be escaped with `esc_url()`.
- [ ] We should swap out `@juliapaxton` sample content for just the name `Julia Paxton` to avoid promoting anyone using that username.

jffng:
This PR addresses #3332.

I think I got them all but could definitely use a check.


scruffian:
I made some changes to change __ to esc_html__. I also removed some names from being translated as I guess names should stay the same. We're also translating prices and brand names, which I found a bit weird but I left it in.

I'm also not really sure why we need to be escaping this content, since it is provided by us so we know its safe!

kjellr (approved):
_[ Whoops, approved by mistake ]_

kjellr (changes requested):
This looks pretty good, thanks @jffng! Though it seems like overkill, I do believe we need to wrap the proper names in escape + translate tags. [According to the Theme Review Guidelines](https://make.wordpress.org/themes/handbook/review/required/#language):

> All theme text strings are to be translatable

(It seems a bit overkill to me too, but that's the rule).

kjellr:
@jffng Actually, a few of these images aren't showing up for me: 

Screen Shot 2021-02-24 at 10 05 52 AM

The Link in Bio ones seem to be working ok. 🤔

kjellr:
And also, we need to make sure the line break tags here are still functioning: 

Screen Shot 2021-02-24 at 10 09 18 AM

Screen Shot 2021-02-24 at 10 10 53 AM

jffng:
I was able to fix the first two, but not sure what's going on with the invitation block pattern.

kjellr:
@jffng They're all working for me now — the split screen one was missing a closing parenthesis. Just need to make the people's names translatable and [fix those `<br />` tags](https://github.com/Automattic/themes/pull/3345#issuecomment-785143675), and then I think we're good to go.

jffng:
@kjellr this still look good? If so feel free to merge.

kjellr (approved):
Yes, this is working for me! Thanks — I'll go ahead and deploy to WP.com. 👍

PR #3346, opened 2021-02-23
Try/mayland blocks globalstylestheme

pbking:
This branch uses the "Global Styles" as a parent theme for Mayland Blocks.

There are a (few) "custom" values added to the theme.json 

Some of the templates have been tweaked for maximum layout-ability (though those changes were unnecessary to get things to work).


pbking:
Note that most of the changes are just the addition of the Global Styles theme.
The same could actually be accomplished by simply enqueuing the CSS file (it doesn't have to be a parent style for these resources to be beneficial).

https://github.com/Automattic/themes/pull/3346/commits/2b66243da86de9eaa608e60bfe3567328ceafa65
That is the only meaningful change, and even then the change to the single.html template wasn't necessary (just made things look a little nicer).

pbking:
Before:
image
image

After:
image
image

PR #3357, opened 2021-02-25
Spearhead: Show the Disable Dark Mode checkbox even if the colors aren't modified

### LINKED ISSUE #3312: Show "Disable Dark Mode" when palette has never been modified.
## Steps to replicate
See: https://github.com/Automattic/themes/issues/3213#issuecomment-781008574

With the recent change in #3213 there is a new checkbox that disables dark mode that is only shown when the "default palette" is used.

## Result

This message is not showing up if the palette has never been modified.  A user needs to change the palette and return it to default in order to see the control.

## Expected

The control should be available in any scenario that the default palette is used (especially if it has never been touched)

scruffian:
This changes the condition to show the checkbox to also show when there are no colors set.

To test (on wpcom):
- Add the theme to a site that has never had custom colors set
- Check that the checkbox still shows

#### Related issue(s):
Fixes #3312

### DIFF
--- a/spearhead/assets/sass/style.scss
+++ b/spearhead/assets/sass/style.scss
@@ -7,7 +7,7 @@ Description: Share your podcast with the world using Spearhead.
 Requires at least: 5.5
 Tested up to: 5.5
 Requires PHP: 5.6.2
-Version: 1.3.1
+Version: 1.3.2
 License: GNU General Public License v2 or later
 License URI: LICENSE
 Template: seedlet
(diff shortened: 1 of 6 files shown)


jffng (commented):
I verified that this indeed shows the checkbox to Disable Dark Mode on a site whose colors have never been modified:

Screen Shot 2021-03-02 at 9 28 22 AM

For reasons I couldn't quickly identify, this PR also had the effect of not showing dark mode  in the customizer (my system settings are set to dark), but I'm not sure why that would be impacted...

PR #3359, opened 2021-02-25
Blank canvas: textdomain fixes + version bump to 1.2.7

kjellr:
Fixes a couple textdomain fixes in Blank Canvas: 

- Missing textdomain in `inc/block-patterns.php`
- Incorrect textdomains in `functions.php`

It also updates the .pot file to include these new strings, and bumps the version so we can re-upload to WP.org.


MaggieCabrera (approved):
⛵

PR #3360, opened 2021-02-25
Seedlet: Fix missing textdomains in block patterns

kjellr:
A few text strings in Seedlet are missing textdomains. The Theme Check plugin threw up a warning about it. 

This PR adds them in to fix that.


MaggieCabrera (approved):
LGTM

PR #3385, opened 2021-03-02
Mayland Blocks: Add a dynamic navigation block option

scruffian:
This makes it possible to have dynamic menus in a navigation block. 

One known issue is that when you update the menu in the customizer the preview doesn't update.

This seems so simple that you wonder if it would make sense to add it to core...

## To test
- Switch to Mayland Blocks
- Notice that your menu doesn't change


scruffian:
The problem with this is that it doesn't work in the site editor. I think we should close this and focus on block themes.

PR #3387, opened 2021-03-02
Blank Canvas: Add `theme name` context to block pattern category string

yuliyan:
Add `theme name` context to the `Blank Canvas` string that is used to register block patterns category in order to get it extracted as a separate original string.

#### Testing instructions:
* Review code changes
* Review the generated `blank-canvas/languages/blank-canvas.pot ` and check if it contains `Blank Canvas` string with context. Confirm other changes in the file make sense.

#### Related issue(s):
p1614696959004800-slack-CEM18K8LT

### DIFF
--- a/blank-canvas/inc/block-patterns.php
+++ b/blank-canvas/inc/block-patterns.php
@@ -18,7 +18,7 @@ if ( ! function_exists( 'blank_canvas_register_block_patterns' ) ) :
 
 			register_block_pattern_category(
 				'blank-canvas',
-				array( 'label' => __( 'Blank Canvas', 'blank-canvas' ) )
+				array( 'label' => _x( 'Blank Canvas', 'theme name', 'blank-canvas' ) )
 			);
 
 			register_block_pattern_category(
--- a/blank-canvas/languages/blank-canvas.pot
+++ b/blank-canvas/languages/blank-canvas.pot
@@ -9,18 +9,17 @@ msgstr ""
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"POT-Creation-Date: 2021-02-25T14:20:31+00:00\n"
+"POT-Creation-Date: 2021-03-02T17:27:23+00:00\n"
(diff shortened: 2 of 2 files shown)


blank-canvas/inc/block-patterns.php
```diff
@@ -18,7 +18,7 @@ function blank_canvas_register_block_patterns() {
 
 			register_block_pattern_category(
 				'blank-canvas',
-				array( 'label' => __( 'Blank Canvas', 'blank-canvas' ) )
+				array( 'label' => _x( 'Blank Canvas', 'theme name', 'blank-canvas' ) )
```
akirk, line 21: Should the context not be `block pattern name`? It does collide with the theme name (which doesn't get a context since it's extracted from style.css) but this here is the name of a block pattern, or is it not? If this is supposed to be the theme name here, then actually no change is necessary.

yuliyan: Considering the code context, I think it's the theme name used a category label that groups various block patterns and therefore the context `theme name` suggests that it shouldn't be translated. Though, I might be wrong so I'll be happy to hear other's thoughts.

Edit: Actually, the string `Blank Canvas` is not being used anywhere else in the theme, so this change seems unnecessary indeed. In which case, resolving the original issue should be just a matter of fixing the translations for https://translate.wordpress.com/projects/wpcom/themes/blank-canvas/-all-translations/632936/

dlind1: Are block pattern category names usually translated? If it shouldn't be translated since it's matching the theme name, wouldn't it be easier to just not wrap it in the gettext call?

yuliyan: @dlind1, I'm not sure if or in what case the theme name may need to be translated, but either way if the category name refers to the theme, we should **not** remove the gettext call they both keep matching.

scruffian:
It looks like nothing needs to happen here, is it OK to close this?

yuliyan:
> It looks like nothing needs to happen here, is it OK to close this?

Yes - the change doesn't seem to be needed after all, so let's close this for now.

PR #3394, opened 2021-03-03
Mayland Blocks: Update buttons colors

scruffian:
This implements button colors using theme.json:

Screenshot 2021-03-03 at 19 08 19

### DIFF
--- a/mayland-blocks/experimental-theme.json
+++ b/mayland-blocks/experimental-theme.json
@@ -182,6 +182,11 @@
 			"typography": {
 				"fontSize": "var(--wp--preset--font-size--small)"
 			}
+		},
+		"core/button": {
+			"color": {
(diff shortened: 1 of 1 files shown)


jffng:
Does this PR depend on a Gutenberg change that has yet to be merged?

I tested this on wpcom and using GB master and 10.1.0 locally without success. I would expect the background color to be set to a CSS custom property here, but it is not:

Screen Shot 2021-03-03 at 3 08 06 PM

scruffian:
Yeah this will only work if you're running the latest Gutenberg

pbking:
Compiling the latest gave me problems, running `npm install` like... 2-3 times?  seemed to get it to build.

Can confirm it works on latest!

MaggieCabrera (approved):
Can confirm it works for me as well!

PR #3400, opened 2021-03-04
First Build of Ponyville

pbking:
This is the first commit of the parent theme "Ponyville"

Ponyville is an unopinionated parent theme that will "Ponyfill" the
gaps between where Gutenberg is today and where it is going.


ponyville/functions.php
```diff
@@ -0,0 +1,61 @@
...
+			$args = new stdClass();
+			$args->menu = null;
+			$args->theme_location = $parsed_block['attrs']['location'];
+			return gutenberg_output_block_nav_menu( $block_content, $args );
+		}
+	}
+	return $block_content;
+}
+// It would be better to use render_block_core/navigation but it's not working
+add_filter( 'render_block', 'render_dynamic_navigation_block', 10, 2);
```
scruffian: This won't work in the site editor, so I think we should remove it.

pbking: I agree. Even if it worked this theme wouldn't be the place for it. An interesting hack but a conversation belonging elsewhere.

ponyville/block-templates/index.html
```diff
@@ -0,0 +1,7 @@
+<!-- wp:query -->
+	<!-- wp:query-loop -->
+		<!-- wp:post-title {"isLink":true} /-->
+	<!-- /wp:query-loop -->
+<!-- /wp:query -->
+
+<!-- wp:post-content /-->
```
MaggieCabrera, line 7: Do we want to show the FSE to the users with this theme? Or should we insert the templates using `do_blocks()` ?

ponyville/sass/blocks/_style.scss
```diff
@@ -0,0 +1,4 @@
+@import "button/style";
+@import "paragraph/style";
+@import "navigation/style";
+@import "group/style";
```
MaggieCabrera, line 4: I wonder if this will grow to Seedlet levels when we want this theme's children to be able to customize all the rest of the blocks. This is the part that worries me the most.

ponyville/style.css
```diff
@@ -0,0 +1,15 @@
+/*
+Theme Name: Ponyville
+Theme URI: 
+Author: Automattic
+Author URI: https://automattic.com/
+Description: Ponyville is a lightly-opinionated base theme for block-based or classic themes using Global Styles
+Requires at least: ?
```
MaggieCabrera, line 7: I wanna say at least 5.7. I think we can have the templates be loaded using `do_blocks()` before FSE is released and then do a major version update that removes the `do_blocks()` and brings back pure html templates so we can support FSE after it launches on core, then we update the requirement to that version of WordPress

ponyville/block-templates/index.html
pbking: I wish do_blocks() was fully featured enough. :(
* There are still a lot of blocks that can't be created using this method, they are limited to the block templates only (if I understand correctly) so there is much we won't be able to do.
* Template Parts don't work with do_blocks(). :(

If both of those things weren't true then we could I think.

However, I would argue that instead Themes should have control as to which "management" options should be available when using it.  Rather than FSE/Customizer being enabled/disabled with the presence of a /block-templates folder I think a better option would be to give themes a theme_supports() option (configurable via a call to theme_supports() or via theme.json) to enable/disable FSE/Customizer.  (Which... I will continue to argue that those things shouldn't be mutually exclusive).

MaggieCabrera: Yes, 100% with you on that. I think we should focus on getting that before we can actually launch any "Ponytheme" then since we don't really want the editor to show until it's really ready for general use.

ponyville/style.css
scruffian: I love this idea

scruffian:
Can we close this one now?

Issue #2361, opened 2020-08-15
Varia: Setting homepage as "Shop" page in WooCommerce breaks it


davipontesblog:
## Steps to replicate

1. On an atomic site, choose a Varia child theme.
2. Set homepage as any page that has a "Cover block" in full width alignment, and also a contact "Form block" in any part of the page as well, for testing purposes.
3. In Design > Customize > Homepage settings, choose to "Hide homepage title"
4. Now go to WP Admin > Plugins, and install and activate WooCommerce.
5. In WP Admin > WooCommerce > Settings > Products, set the homepage as also the "Shop page":

![Screen Shot 2020-08-15 at 14 23 20](https://user-images.githubusercontent.com/31009008/90319080-ec8fd800-df02-11ea-8002-8e0a207bd9c4.png)

6. Save the changes, and visit the homepage on a new browser window (front-end of the site)

## Result

- The page title appears on the homepage, regardless of the option selected in the customizer to hide it.
- No blocks appear in full width, the whole page is "boxed".
- Any contact forms created with the Form block are broken, showing the shortcode instead.

See below:

<img width="1212" alt="Screen Shot 2020-08-15 at 14 25 48" src="https://user-images.githubusercontent.com/31009008/90319115-40022600-df03-11ea-85a1-bef40b362044.png">

- It also breaks the spacing in-between all blocks in general on this page as seen above.

## Expected

I expected none of this to happen. Set the "Shop page" as any other page of the site, in step 5 of "Steps to replicate", and all the issues go away.

* User report: #3227588-zen (I noticed this while in a QS session with a user today, and it puzzled me. I was able to figure it out in testing later).

danieldudzic:
Hi @davipontesblog!

Thanks for the report, and thank you @arcangelini for the PR.

I have spent some time, digging into this issue, and unfortunately, I don't see an easy, correct fix.

The problem is not Varia specific but affects all Gutenberg themes, including Seedlet, Twenty Twenty One and even Storefront - so a fix should be applied on a higher level to avoid patching up all themes individually.

Please note the odd markup of the "Shop" page:

The page content (in this case blocks) ends up inside of:
```
<article id="woocommerce-wrapper">
<header class="woocommerce-products-header">
<div class="page-description">
Blocks end up here
</div>
</header>
</article>
```

The above sort of makes sense, because I believe the "Shop" page content was always meant to hold a page description rather than "full page content".

[...]

PR #3404, opened 2021-03-05
Varia: add wrappers to fix WooCommerce styling

arcangelini:
## Changes proposed in this Pull Request
In order to allow the current styling to apply to the page content that gets placed into the `page-description` I adjusted the wrappers.

I tried testing this on the child themes but the `build:woocommerce` doesn't seem to run through the child themes. I can continue testing if I can get some pointers on that.

Also, I set the page content `max-width` based on the varia setting, I was not sure if there was a global setting for this.

#### Before
[Screenshot](https://d.pr/i/JCLMva)
[Recording](https://d.pr/i/rSl1RO)

#### After
[Screenshot](https://d.pr/i/csqenR)
[Recording](https://d.pr/i/zGcAWp)

#### Related issue(s)

#2361

### DIFF
--- a/varia/inc/style-wpcom.css
+++ b/varia/inc/style-wpcom.css
@@ -5,7 +5,8 @@
  /**
  * Hide page title on the homepage
  */
-.home.page.hide-homepage-title .entry-header {
+.home.page.hide-homepage-title .entry-header,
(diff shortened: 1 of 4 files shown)


danieldudzic:
This should be fixed in the core: https://github.com/woocommerce/woocommerce/issues/29456

Issue #3413, opened 2021-03-08
[Blank Canvas Blocks] Evaluate Blocks


pbking:
Ready the following blocks for Blank Canvas Blocks (A Ponyfill theme)

For each of the following blocks:

Evaluate "Blank Canvas Blocks" against "Blank Canvas"

If something can be achieved with Gutenberg theme.json styling for that block then leverage that first.

If something is expected to be achievable in Gutenberg, adjust the theme.json's "custom" block to express that opinion and add custom SASS to `/sass/blocks` consume those variables expressing that opinion.

 ([bcb demo](https://blank-canvas-blocks.mystagingwebsite.com/))

- [x] audio ([theamdemo](https://theamdemo.wordpress.com/2019/06/26/gutenberg-audio/?theme=pub/blank-canvas)) ([bcb demo](https://blank-canvas-blocks.mystagingwebsite.com/2019/06/26/gutenberg-audio/))
- [x] button ([theamdemo](https://theamdemo.wordpress.com/2019/06/25/gutenberg-button/?theme=pub/blank-canvas))([bcb demo](https://blank-canvas-blocks.mystagingwebsite.com/2019/06/25/gutenberg-button/))
- [ ] calendar
- [x] code ([theamdemo](https://theamdemo.wordpress.com/2020/03/05/gutenberg-misc-formatting/?theme=pub/blank-canvas)) ([bcb demo](https://blank-canvas-blocks.mystagingwebsite.com/2020/03/05/gutenberg-misc-formatting/))
- [x] columns ([theamdemo](https://theamdemo.wordpress.com/2018/10/24/gutenberg-columns/?theme=pub/blank-canvas)) ([bcb demo](https://blank-canvas-blocks.mystagingwebsite.com/2018/10/24/gutenberg-columns/))

- [x] cover ([theamdemo](https://theamdemo.wordpress.com/2019/06/24/gutenberg-cover/?theme=pub/blank-canvas)) ([bcb demo](https://blank-canvas-blocks.mystagingwebsite.com/2019/06/24/gutenberg-cover/))
- [x] file ([theamdemo](https://theamdemo.wordpress.com/2019/06/25/gutenberg-file/?theme=pub/blank-canvas)) ([bcb demo](https://blank-canvas-blocks.mystagingwebsite.com/2019/06/25/gutenberg-file/))
[...]

PR #3421, opened 2021-03-08
Blank Canvas Blocks: patch quote styles

jffng:
Part of #3413. This PR adds styles for the quote block.

Before | This PR
------- | -------
Screen Shot 2021-03-08 at 11 37 21 AM | Screen Shot 2021-03-08 at 11 37 03 AM

### DIFF
--- a/blank-canvas-blocks/sass/blocks/quote/_style.scss
+++ b/blank-canvas-blocks/sass/blocks/quote/_style.scss
@@ -1,0 +1,5 @@
+.wp-block-quote {
+	cite, .wp-block-quote__citation {
+		font-size: var(--wp--preset--font-size--small);
+	}
+}
--- a/blank-canvas-blocks/sass/blocks/quote/_style.scss
+++ b/blank-canvas-blocks/sass/blocks/quote/_style.scss
@@ -1,5 +1,8 @@
 .wp-block-quote {
+	font-family: var(--wp--custom--quote--typography--font-family);
+	border-left: var(--wp--custom--quote--border--width) solid var(--wp--custom--quote--border--color);
 	cite, .wp-block-quote__citation {
-		font-size: var(--wp--preset--font-size--small);
+		font-size: var(--wp--custom--quote--citation--typography--font-size);
+		font-style: var(--wp--custom--quote--citation--typography--font-style);
 	}
 }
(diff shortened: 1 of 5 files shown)


blank-canvas-blocks/sass/blocks/quote/_style.scss
```diff
@@ -0,0 +1,5 @@
+.wp-block-quote {
+	cite, .wp-block-quote__citation {
+		font-size: var(--wp--preset--font-size--small);
```
pbking: Would you expect to be able to configure this?  Could this be a large font in another theme?  If so how would you expect to configure this once Gutenberg would allow for that?

There's no guarantee that a particular preset is going to be available in a child theme (since they can change those preset just like color). Maybe we should be leveraging only variables expressed in "custom".

Perhaps: `font-size: var(--wp--custom--quote--citation--typography--font-size)`?

and that could map in theme.json:

`custom: quote: citation: typography: font-size: var(--wp--preset--font-size--small),`

Maybe that's deeper of an object that's needed...

jffng: > There's no guarantee that a particular preset is going to be available in a child theme (since they can change those preset just like color). Maybe we should be leveraging only variables expressed in "custom".

I see your point.

> Would you expect to be able to configure this? Could this be a large font in another theme? If so how would you expect to configure this once Gutenberg would allow for that?

As a theme developer and user, I would want to be able to change the font-size of the citation in a quote. I also expect the front-end and view to match. So to clarify for our purposes — should that be done with CSS + custom theme.json variables in this theme? 

(Aside: it seems to me the main difference with this approach compared to Seedlet is that I'm expressing the variables in the custom section of theme.json, rather than modular sass files.)

jffng:
Noting this is what the quote looks like in Blank Canvas:
 
Screen Shot 2021-03-08 at 11 12 47 AM

blank-canvas-blocks/sass/blocks/quote/_style.scss
pbking: (Aside the II: That's the main difference yea.  There's some part in seedlet where the question to how to build the CSS was "how do we want it to work?" where the question here is rather "how do I expect to be able to configure that for Gutenberg and then take advantage of that configuration". I imagine the answer is often the same but likely not always.)

I think you're right. I expect that any CSS we add to this would be because it's something we're configuring in theme.json.  Hopefully little would be added that isn't directly linked to a variable.

jffng:
Font families will be taken care of by #3422.

scruffian (approved):
This looks good. I expect some themes would want more customization than this can offer - for example big quote marks (like libre 2). I guess we should add these as they are needed though...

PR #3428, opened 2021-03-09
Blank Canvas Blocks: Add alignment rules

scruffian:
This improves alignment styles to account for different content / viewport widths.

To test look at the audio block with different alignments.
Also check different viewport sizes.


pbking:
Cherry Picked via #3429 and merged from there.

PR #3430, opened 2021-03-09
Add default styles to BCB for Image Block

pbking:
There appear to be consistent default styles added to every theme for images and captions.
These are added in this change.  The Gutenberg PR most likely to correct the need for these additions is noted in the change for future reference.

Validated with [Columns](https://theamdemo.wordpress.com/2018/10/24/gutenberg-columns/) and [Image](https://theamdemo.wordpress.com/2019/06/25/gutenberg-image/) theamdemo content.

Before:
image
image

After:
image
image

### DIFF
--- a/blank-canvas-blocks/sass/blocks/_image.scss
+++ b/blank-canvas-blocks/sass/blocks/_image.scss
@@ -5,11 +5,15 @@
 	default styles.  It's difficult to say how this will land, however 
 	based on discussion found in (many) related issues here: 
 	https://github.com/WordPress/gutenberg/issues/28923
+	*/
+	text-align: center;
+}
 
+/*
 	Further gutenberg APPEARS to be defaulting the margin to the same as 
 	the below, however (for reason's that aren't clear to me now) this
 	isn't coming through.
-	*/
-	text-align: center;
+*/
+.wp-block-column .wp-block-image {
 	margin: 0 0 1em 0;
 }
--- a/blank-canvas-blocks/sass/blocks/_image.scss
+++ b/blank-canvas-blocks/sass/blocks/_image.scss
@@ -1,0 +1,12 @@
+.wp-block-image {
+	/* 
+	From what I can tell the below are styles regularly used by themes
+	to fix the image block.  I believe these should go into the block's
+	default styles.  It's difficult to say how this will land, however 
+	based on discussion found in (many) related issues here: 
+	https://github.com/WordPress/gutenberg/issues/28923
+	*/
+	text-align: center;
+	margin: 0;
+	margin-bottom: 1em;
+}
--- a/blank-canvas-blocks/sass/blocks/_image.scss
+++ b/blank-canvas-blocks/sass/blocks/_image.scss
@@ -5,8 +5,11 @@
 	default styles.  It's difficult to say how this will land, however 
 	based on discussion found in (many) related issues here: 
 	https://github.com/WordPress/gutenberg/issues/28923
+
+	Further gutenberg APPEARS to be defaulting the margin to the same as 
+	the below, however (for reason's that aren't clear to me now) this
+	isn't coming through.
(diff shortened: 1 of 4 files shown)


blank-canvas-blocks/sass/blocks/_image.scss
```diff
@@ -0,0 +1,12 @@
+.wp-block-image {
+	/* 
+	From what I can tell the below are styles regularly used by themes
+	to fix the image block.  I believe these should go into the block's
+	default styles.  It's difficult to say how this will land, however 
+	based on discussion found in (many) related issues here: 
+	https://github.com/WordPress/gutenberg/issues/28923
+	*/
+	text-align: center;
+	margin: 0;
```
jffng: Are these margin styles needed? Asking because I saw a recent Gutenberg PR that reset the margins for the image block: 

https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/image/style.scss#L1-L3

pbking: Oh!  I hope not.  That would be much better.  I'm not sure why that isn't reflecting for me though... I pulled and built gutenberg today and that should have picked up that change.

pbking: I really wanted it to be true, but for reason's I can't explain the margin being supplied by Gutenberg isn't 0-ed out.

I simplified the values to match what's expected, but (at least for the moment) I think we've got to set that here. :/

blank-canvas-blocks/sass/blocks/_image.scss
```diff
@@ -0,0 +1,15 @@
...
+	default styles.  It's difficult to say how this will land, however 
+	based on discussion found in (many) related issues here: 
+	https://github.com/WordPress/gutenberg/issues/28923
+
+	Further gutenberg APPEARS to be defaulting the margin to the same as 
+	the below, however (for reason's that aren't clear to me now) this
+	isn't coming through.
+	*/
+	text-align: center;
+	margin: 0 0 1em 0;
```
scruffian, line 18: Should this use --wp--custom--margin--vertical?

scruffian: Margin bottom is already set like this in Gutenberg. Is the purpose of this line to reset the left and right margins? Why?

blank-canvas-blocks/sass/blocks/_image.scss
```diff
@@ -0,0 +1,15 @@
...
+	to fix the image block.  I believe these should go into the block's
+	default styles.  It's difficult to say how this will land, however 
+	based on discussion found in (many) related issues here: 
+	https://github.com/WordPress/gutenberg/issues/28923
+
+	Further gutenberg APPEARS to be defaulting the margin to the same as 
+	the below, however (for reason's that aren't clear to me now) this
+	isn't coming through.
+	*/
+	text-align: center;
```
scruffian: I think this should be customizable in the theme.json - `--wp-custom-image-alignment`

blank-canvas-blocks/sass/blocks/_image.scss
pbking: Yea, `margin: 0 0 1em 0;` is that the code in Gutenberg looks like it's supposed to be.  But that's not what I see when I run this with a fully up-to-date Gutenberg.  My 'druthers is to just leave out the margin attribute completely, but in the meantime I thought it better to set "the whole thing" to what it's going to be.

blank-canvas-blocks/sass/blocks/_image.scss
pbking: @jffng and I reviewed and because of image caption alignment decided not to offer configurable alignment (yet).  This could possibly be covered in the future if we want to configure that in a theme.

PR #3432, opened 2021-03-09
Add default styles for Separator block

jffng:
Part of #3413.

Before | After
-------- | --------
Screen Shot 2021-03-09 at 11 17 14 AM | Screen Shot 2021-03-09 at 11 16 57 AM


pbking:
Super duper! 🚢 -ing it

PR #3433, opened 2021-03-09
Adjusting font size configuration to match Blank Canvas

pbking:
Adjusting font size configuration to match Blank Canvas.
Values in Blank Canvas were derived from [Seedlet definitions](https://github.com/Automattic/themes/blob/trunk/seedlet/functions.php#L124).

Also removed inappropriate color definitions.

Before:
image

After:
image


jffng (approved):
LGTM 🚢

PR #3434, opened 2021-03-09
Blan Canvas Blocks: Add default styles for Video block

MaggieCabrera:
Part of #3413

I added `add_theme_support( 'responsive-embeds' );` because that adds the editor styles for responsive videos. The markup of the blocks on Theam demo site wasn't the correct one and I had to re create the blocks in order for the support to work on the video blocks.

### DIFF
--- a/blank-canvas-blocks/assets/ponyfill.css
+++ b/blank-canvas-blocks/assets/ponyfill.css
@@ -366,4 +366,9 @@ p.has-text-color a {
 	max-width: var(--wp--custom--separator--width);
 }
 
+.wp-block-video figcaption {
+	margin: var(--wp--custom--video--caption--margin);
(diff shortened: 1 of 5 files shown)


pbking:
I believe the video blocks playback controls should be on for most of those blocks.  
https://theamdemo.wordpress.com/2019/06/25/gutenberg-video/
Noticed as the "Playback Controls Off" option was the only one with controls turned on.  LMK if that doesn't align with your expectations though and I'll revert that content change.

pbking:
I see that the left/right aligned videos in the theamdemo content don't shrink to the same size as they do on Blank Canvas.  I expect that has everything to do with _alignment.scss (and it's differences with seedlet) than styles for video so not holding up this merge for that.

LGTM.  Bringing it in.

PR #3436, opened 2021-03-09
Add paragraph drop cap style configuration

pbking:
Adds configuration of the drop cap style.  This should complete the necessary changes for the "paragraph" block portion of #3413

Before:
image

After:
image

### DIFF
--- a/blank-canvas-blocks/experimental-theme.json
+++ b/blank-canvas-blocks/experimental-theme.json
@@ -175,9 +175,9 @@
 					"dropcap": {
 						"margin": ".1em .1em 0 0",
 						"typography": {
-							"font-family": "var(--wp--preset--font-family--base)",
-							"font-size": "110px",
-							"font-weight": "400"
+							"fontFamily": "var(--wp--preset--font-family--base)",
+							"fontSize": "110px",
+							"fontWeight": "400"
 						}
 					}
 				},
--- a/blank-canvas-blocks/experimental-theme.json
+++ b/blank-canvas-blocks/experimental-theme.json
@@ -173,10 +173,11 @@
 				},
 				"paragraph": {
 					"dropcap": {
+						"margin": ".1em .1em 0 0",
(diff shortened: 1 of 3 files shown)


blank-canvas-blocks/experimental-theme.json
```diff
@@ -171,6 +171,15 @@
 						"box-shadow": "none"
 					}
 				},
+				"paragraph": {
+					"dropcap": {
+						"typography": {
+							"font-family": "var(--wp--preset--font-family--base)",
```
jffng: Nit pick but are we going with camel case or trying matching the CSS attribute?

I saw in [another PR a camel case](https://github.com/Automattic/themes/pull/3437/files#diff-6b32b96741b2b4b9e621f53f3709cb5e5dcbb41c0467f83a2f30c99577f724b8R149), and it made me think we should be consistent with what Gutenberg is doing.

jffng:
Nice one, this is close but I'm wondering if we can get it closer, at least align the top of the letter with the rest of the paragraph. 

BCB this PR | Blank Canvas
------- | -------
Screen Shot 2021-03-09 at 3 23 44 PM | Screen Shot 2021-03-09 at 3 23 39 PM

blank-canvas-blocks/experimental-theme.json
MaggieCabrera: +1 on this, I opened a PR to change to camel case here https://github.com/Automattic/themes/pull/3444

pbking:
Added a (single) "margin" property to apply to the dropcap letter.  That let's us pull it down to align it with the top.
Adjusted the font-size and font-weight to match Blank Canvas.  
Then I further increased the font-size (a little) so that it would fit a LITTLE more snugly.  

Blank Canvas Numbers:
image

Final: 
image
image

jffng:
This looks much better, thank you! 🚢

PR #3439, opened 2021-03-09
Add configurable header styles to Blank Canvas Blocks

pbking:
Part of #3413 (for headings)

Added a custom headings configuration block including font-weight, -height, and -family  and set the values to what they are in Blank Canvas.

Mapped font-family and font-weight to those properties to all h elements (h1-h6) in scss.

Refactored 'line-height' out of 'custom'
* for body just moves it to the basic typography object as it doesn't appear to be used elsewhere
* for header moved it into custom object specifically for heading block configuration and mapped that to the header blocks. (This keeps all the "custom headings configuration" in one place)

Set the header font size values in theme.json to the same size values in Blank Canvas

Before:
image

After:
image

### DIFF
--- a/blank-canvas-blocks/experimental-theme.json
+++ b/blank-canvas-blocks/experimental-theme.json
@@ -66,6 +66,11 @@
 						"fontFamily": "var(--font-base, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif)",
 						"slug": "base",
 						"name": "Base"
+					},
+					{
+						"fontFamily": "var(--font-headings, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif)",
+						"slug": "headings",
+						"name": "Headings"
 					}
 				]
 			},
@@ -87,10 +92,6 @@
 					"foreground": "var(--wp--preset--color--almost-black)",
 					"background": "var(--wp--preset--color--white)",
 					"selection": "var(--wp--preset--color--almost-white)"
-				},
-				"lineHeight": {
-					"body": 1.6,
-					"headings": 1.125
 				},
 				"margin": {
 					"horizontal": "20px",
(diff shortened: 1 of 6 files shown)


jffng (commented):
This LGTM!

Just one comment that is more general — it seems we're leaning towards using the hyphenated CSS attribute name in our custom variables. When it's generated by theme.json / global styles mechanism proper, though, camel case is used. 

Do we want to make these consistent, or is it better to leave them hyphenated so we know which variables we're creating via the "custom" section?

MaggieCabrera:
> Do we want to make these consistent, or is it better to leave them hyphenated so we know which variables we're creating via the "custom" section?

Good point! I think we should use camel case for consistency, we are already adding the `--wp--custom` prefix to all of our custom variables anyway

blank-canvas-blocks/experimental-theme.json
```diff
@@ -175,44 +183,44 @@
...
-				"lineHeight": "var(--wp--custom--line-height--headings)"
+				"fontSize": "48px",
+				"lineHeight": "var(--wp--custom--heading--typography--line-height)"
 			}
 		},
 		"core/heading/h2": {
 			"typography": {
-				"fontSize": "var(--wp--preset--font-size--huge)",
-				"lineHeight": "var(--wp--custom--line-height--headings)"
+				"fontSize": "32px",
```
MaggieCabrera, line 192: why are we not using variables for h1 and h2 like we are doing for the rest?

pbking: Because those sizes exceeded what was configured as font size. I wasn't sure if it was a good idea to add two more sizes, which seem more focused on paragraph styles, just to style the larger header elements.

pbking:
Good point y'all.  Refactored for consistency.
I'll keep an eye on that in the future.

blank-canvas-blocks/experimental-theme.json
jffng: I lean towards not adding more variables for headings, since we have the ability to change it directly in the global styles supported style properties of heading blocks.  

Unless these font sizes would be used again outside the H1 + H2. Now that I think about it, they might for a block like site title... Hmm.

jffng:
Once merge conflicts are resolved this LGTM! Thanks for the follow up on consistency in the key names.

blank-canvas-blocks/experimental-theme.json
pbking: That's more of a decision at the "theme" level then the "polyfill" level and those additional variables certainly could be added by that theme (or this one if we decide to style the block title in that way).
`Don't code it 'till ya need it.`

PR #3444, opened 2021-03-10
Blank Canvas Blocks: Switch custom variables to camel case

MaggieCabrera:
This PR changes our custom variables in theme.json to use camel case for consistency with the way the GB presets work. No CSS needs to change because the camel case gets transformed to the same hyphenated format we were using before once the variables are created (GB is really smart :D)


scruffian (approved):
LGTM

PR #3416, opened 2021-03-08
Blank Canvas: Add the "Subscription Form and Links" pattern

kjellr:
Adds the **Jetpack Subscribe + Links** pattern from https://github.com/Automattic/themes/issues/3161, now that https://github.com/Automattic/jetpack/issues/18085 has been resolved. This PR also properly escapes the wpcom-only block pattern strings + URLs to match up with the changes made in #3345.

Two small caveats: 

- The design is changed slightly from the mockup in the other thread: I've removed the noisy background so that the form field text is more readable.
- It's pretty obvious that the two 50% buttons don't add up to the full width of the container. This is a Gutenberg bug, and there's [a potential fix here](https://github.com/WordPress/gutenberg/pull/29165). I don't think that should hold this PR up though. 

## Screenshots

Editor:

kjellreigstad wordpress com_wp-admin_post php_post=2075 action=edit(1200x900)

Front End:

Desktop|Mobile
---|---
kjellreigstad wordpress com_(1200x900)|kjellreigstad wordpress com_(iPhone X)

PR #3445, opened 2021-03-10
Blank Canvas: Temporarily comment out the "Subscription Form and Links" pattern

kjellr:
Followup from https://github.com/Automattic/themes/pull/3416. 

The "Subscription Form and Links" block pattern revealed a Jetpack bug (https://github.com/Automattic/jetpack/issues/19113) that leaves the post-submit messages unreadable. While that gets sorted out, we'll temporarily disable this pattern.


jeffikus (approved):
LGTM!

Issue #3418, opened 2021-03-08
[Blank Canvas Blocks] Configure Base Typography


pbking:
Configure the HEADER and BODY typefaces for Blank Canvas Blocks including loading them via function.php and configuring them via theme.json

They should match Blank Canvas ([example](https://theamdemo.wordpress.com/2019/06/24/gutenberg-paragraph/?theme=pub/blank-canvas)).

jffng:
We should probably get this in place first so that we don't end up having to configure font-family unnecessarily for many blocks, like I'm doing here: https://github.com/Automattic/themes/pull/3421/files#diff-6b32b96741b2b4b9e621f53f3709cb5e5dcbb41c0467f83a2f30c99577f724b8R184-R186

scruffian:
In the end we didn't need to queue any fonts in functions.php since Blank Canvas doesn't use custom fonts. I feel like we should add that scaffolding in though to make it easier for child themes to override. We should also look at https://github.com/Automattic/themes/pull/3309

pbking:
What do you mean by that scaffolding?  Is that something that can be successfully derived exclusively from theme.json?  Will that be some boilerplate in the function.php to get that kicked off?

----

In the change we also eliminated the header font configuration.  The 'body' font looks to the --font-base variable but not the --font-headings for headers.

If that font selection tool won't be used with Blank Canvas than that's cool and we can probably eliminate the usage of that --font-base variable as well.

Otherwise we might want to include the headings variable, even though it's the same font (by default) so that the headings are (bulk) configurable.

scruffian:
I'm thinking about the boilerplate code we need to load the font files themselves, which is in functions.php

MaggieCabrera:
> I'm thinking about the boilerplate code we need to load the font files themselves, which is in functions.php

Things like this are why BC needs to be a parent theme and not a plugin. It's very hard to have it be **just** CSS.

scruffian:
I made a suggestion for how to achieve this here: https://github.com/Automattic/themes/pull/3447

scruffian:
Closed by #3447

PR #3447, opened 2021-03-10
Blank Canvas Blocks: Add Google fonts

scruffian:
This adds some scaffolding to make it possible to add custom fonts to Blank Canvas, or any child themes. It also adds a filter as per https://github.com/Automattic/themes/pull/3309, which will make it easier for child themes to extend.

I have implemented this in the heading block as an example, but we should remove this change before merging.

@kjellr what do you think about adding Google fonts to Blank Canvas Blocks?

#### Related issue(s):
#3418

### DIFF
--- a/blank-canvas-blocks/experimental-theme.json
+++ b/blank-canvas-blocks/experimental-theme.json
@@ -80,6 +80,10 @@
 				]
 			},
 			"custom": {
+				"fonts": {
+					"primary": "'Playfair Display', Georgia, Times, serif",
+					"secondary": "'Fira Sans', Helvetica, Arial, sans-serif"
+				},
 				"color": {
 					"primary": "var(--wp--preset--color--black)",
 					"secondary": "var(--wp--preset--color--blue)",
--- a/blank-canvas-blocks/experimental-theme.json
+++ b/blank-canvas-blocks/experimental-theme.json
@@ -85,14 +85,6 @@
 				]
 			},
 			"custom": {
-				"fonts": {
-					"primary": "'Playfair Display', Georgia, Times, serif",
-					"secondary": "'Fira Sans', Helvetica, Arial, sans-serif"
-				},
-				"fontsToLoadFromGoogle": [
-					"family=Fira Sans:ital,wght@0,400;0,500;1,400",
-					"family=Playfair Display:ital,wght@0,400;0,700;1,400"
-				],
 				"color": {
 					"primary": "var(--wp--preset--color--black)",
 					"secondary": "var(--wp--preset--color--blue)",
--- a/blank-canvas-blocks/experimental-theme.json
+++ b/blank-canvas-blocks/experimental-theme.json
@@ -66,6 +66,11 @@
 						"fontFamily": "var(--font-base, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif)",
 						"slug": "base",
 						"name": "Base"
+					},
+					{
+						"fontFamily": "var(--font-headings, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif)",
+						"slug": "headings",
+						"name": "Headings"
 					}
 				]
 			},
@@ -91,10 +96,6 @@
 					"foreground": "var(--wp--preset--color--almost-black)",
 					"background": "var(--wp--preset--color--white)",
 					"selection": "var(--wp--preset--color--almost-white)"
-				},
-				"lineHeight": {
-					"body": 1.6,
-					"headings": 1.125
 				},
 				"margin": {
 					"horizontal": "20px",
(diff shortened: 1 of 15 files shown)


blank-canvas-blocks/functions.php
```diff
@@ -28,10 +29,62 @@ function blank_canvas_blocks_support()  {
...
+		 * @param array $font_families An array of font families to be imploded for the Google Font API
+		 */
+		$font_families = apply_filters( 'included_google_font_families', $font_families );
+
+		$query_args = array(
+			'family' => urlencode( implode( '|', $font_families ) ),
+			'subset' => urlencode( 'latin,latin-ext' ),
+		);
+
+		$fonts_url = add_query_arg( $query_args, 'https://fonts.googleapis.com/css' );
```
jffng: I think if we do add support for loading Google Fonts, we should use the [CSS2 api](https://developers.google.com/fonts/docs/css2). 

It's a slightly different format but there's an [example in Spearhead](https://github.com/Automattic/themes/blob/trunk/spearhead/functions.php#L167-L184):

blank-canvas-blocks/functions.php
```diff
@@ -28,10 +29,62 @@ function blank_canvas_blocks_support()  {
...
+		if ( 'off' !== $playfair_display ) {
+			$font_families[] = 'Playfair Display:ital,wght@0,400;0,700;1,400';
+		}
+
+		/**
+		 * A filter to enable child themes to add/change/omit font families.
+		 *
+		 * @param array $font_families An array of font families to be imploded for the Google Font API
+		 */
+		$font_families = apply_filters( 'included_google_font_families', $font_families );
```
pbking: Could we define/pull this from theme.json? `"custom": { "fontsToLoadFromGoogle": ["Playfair Display..."` or some such?

kjellr:
> @kjellr what do you think about adding Google fonts to Blank Canvas Blocks?

I think it makes sense to include the scaffolding in here so that Google fonts could be enqueued for child themes, but if we could still use the default system font stack for Blank Canvas itself, that would be my preference.

blank-canvas-blocks/sass/blocks/_heading.scss
```diff
@@ -1,5 +1,5 @@
 h1, h2, h3, h4, h5, h6 {
 	clear: both;
-	font-family: var(--wp--custom--heading--typography--font-family);
+	font-family: var(--wp--custom--fonts--primary);
```
pbking: I believe this shouldn't change from --heading--typography--font-family so that this can be configured via theme.json specifically.  

If we want to use custom fonts 'primary', 'secondary' or 'pickle' perhaps that should stay in theme.json rather than being consumed directly by any .scss blocks.  Setting the value of --heading--typography--font-family to "var(--wp--custom--fonts-primary)" seems fine though.

blank-canvas-blocks/experimental-theme.json
```diff
@@ -85,6 +85,14 @@
 				]
 			},
 			"custom": {
+				"fonts": {
+					"primary": "'Playfair Display', Georgia, Times, serif",
+					"secondary": "'Fira Sans', Helvetica, Arial, sans-serif"
+				},
```
pbking: What is the advantage of using "custom.fonts.primary" over the font preset variables?

blank-canvas-blocks/functions.php
```diff
@@ -28,10 +29,35 @@ function blank_canvas_blocks_support()  {
...
+ * Add Google webfonts
+ *
+ * @return $fonts_url
+ */
+
+function blank_canvas_blocks_fonts_url() {
+	$fonts_url = '';
+
+	$theme_data = WP_Theme_JSON_Resolver::get_merged_data()->get_settings();
+	$font_families = $theme_data['defaults']['custom']['fontsToLoadFromGoogle'];
```
pbking: This makes me so happy. :)

jffng (approved):
This is very cool! I tested and verified this scaffolding and endpoints works. Once we remove the test fonts this LGTM 🚢 

I looked into how we could toggle on/off fonts that do not support a given language / character set. 

Right now, the way translating theme.json works is by loading a predetermined tree of keys to translate: https://github.com/WordPress/gutenberg/blob/9f4f7d41b0129fc934951d0e5b4d2c45a1bf8b5c/lib/class-wp-theme-json-resolver.php#L81-L201

Which keys to translate come from a file called [experimental-i18n-theme.json](https://github.com/WordPress/gutenberg/blob/trunk/lib/experimental-i18n-theme.json)

I doubt the following suggestion would fly given this would need to land in core, but in theory using this mechanism we could add to experimental-i18n-theme.json:

```json
"settings": {
	"*": {
		"typography": {
			"fontsToLoadFromGoogle": [ { "name": "Google fonts" } ]
```

And then to our theme.json:

```json
"fontsToLoadFromGoogle": [
	{
		"name": "Fira Sans",
		"load": "On",
		"path": "family=Fira Sans:ital,wght@0,400;0,500;1,400"
	}
],
```  

Then translators would set the "load" key to "Off".

blank-canvas-blocks/functions.php
jffng: Me too!!

scruffian:
Could we add our own experimental-theme-i18n.json to the theme?

blank-canvas-blocks/experimental-theme.json
scruffian: It's just for testing purposes.

blank-canvas-blocks/sass/blocks/_heading.scss
scruffian: This change was just for testing purposes

jffng:
> Could we add our own experimental-theme-i18n.json to the theme?

Adding our own experimental-i18n-theme.json doesn't work right now, but maybe in the future?

scruffian:
Ok let's just merge this for now and then circle back to the enabling/disabling question later. I opened an issue for that here: https://github.com/Automattic/themes/issues/3468

PR #3448, opened 2021-03-10
Handle nested blocks and remove Gutenberg theme CSS.

jffng:
I am not sure if I'm heading in the right direction here: 

Before | This PR
------ | -------
screencapture-localhost-4759-2021-03-10-12_47_40 | screencapture-localhost-4759-2021-03-10-12_47_15

Right now Gutenberg adds a bunch of padding via its theme.css. My thought was to remove this so that only the group block's custom padding values supplied via CSS variables are applied.

I also added a patch for nested alignfull and alignwide group blocks, but I'm not sure how we should handle nested default, wide, and full aligned blocks. Should we just mirror exactly what [Blank Canvas](scrugM-group-p2?theme=pub/blank-canvas) / Seedlet does?

### DIFF
--- a/blank-canvas-blocks/assets/ponyfill.css
+++ b/blank-canvas-blocks/assets/ponyfill.css
@@ -363,8 +363,18 @@ p.has-drop-cap:not(:focus):first-letter {
 	font-style: var(--wp--custom--quote--citation--typography--font-style);
 }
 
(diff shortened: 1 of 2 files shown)


jffng:
Closing in favor of #3449

PR #3449, opened 2021-03-10
Adjusted Group Alignment especially for nested groups

pbking:
Adjusted Group alignment, especially for nested groups, according to the content found here: https://theamdemo.wordpress.com/2019/06/25/group/

* The "padding" to groups with a background color was removed and instead applied to the inner container.
* Any blocks, inside of a group, set to alignwide should be as wide as it's container, however wide that is.
* When a group is nested in another group, and no alignment value is assigned, that groups width defaults to the theme default.

#### Related issue(s):
#3413 (Groups)

#### NOTE

This works dandily on the front end.  (At least I think the blocks are all laying out as I would expect them to in the content test.  However the editor is NOT currently playing nice with all of the situations demoed in that content and I'm just not sure how to get it to.

Before:
image
image

After:
image

### DIFF
--- a/blank-canvas-blocks/sass/blocks/_group.scss
+++ b/blank-canvas-blocks/sass/blocks/_group.scss
@@ -1,3 +1,10 @@
+// NOTE: Gutenberg sets a padding value by default for groups with a background.
+// However when this top level element has padding applied it mucks up the alignment
+// calculations, so it's removed and set instead (to a configurable value) on the inner-container instead.
+.wp-block-group.has-background {
+	padding: unset;
+}
+
 .wp-block-group {
 	&.has-background {
 		.wp-block-group__inner-container {
--- a/blank-canvas-blocks/sass/blocks/_group.scss
+++ b/blank-canvas-blocks/sass/blocks/_group.scss
@@ -3,12 +3,7 @@
 // calculations, so it's removed and set instead (to a configurable value) on the inner-container instead.
 .wp-block-group.has-background {
 	padding: unset;
-}
-
-.wp-block-group {
-	&.has-background {
-		.wp-block-group__inner-container {
-			padding: var(--wp--custom--margin--vertical) var(--wp--custom--margin--horizontal);
-		}
+	.wp-block-group__inner-container {
+		padding: var(--wp--custom--margin--vertical) var(--wp--custom--margin--horizontal);
 	}
 }
--- a/blank-canvas-blocks/assets/ponyfill.css
+++ b/blank-canvas-blocks/assets/ponyfill.css
@@ -92,6 +92,7 @@ body {
 	max-width: none;
 }
 
+.wp-block-group .wp-block-group,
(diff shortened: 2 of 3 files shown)


blank-canvas-blocks/sass/blocks/_group.scss
```diff
@@ -1,7 +1,9 @@
-.wp-block-group {
-	&.has-background {
-		.wp-block-group__inner-container {
-			padding: var(--wp--custom--margin--vertical) var(--wp--custom--margin--horizontal);
-		}
+// NOTE: Gutenberg sets a padding value by default for groups with a background.
+// However when this top level element has padding applied it mucks up the alignment
+// calculations, so it's removed and set instead (to a configurable value) on the inner-container instead.
```
jffng, line 3: Just noting I think this container may be going away in https://github.com/WordPress/gutenberg/pull/29335

blank-canvas-blocks/sass/blocks/_group.scss
```diff
@@ -1,7 +1,9 @@
...
+.wp-block-group.has-background {
+	padding: unset;
+	.wp-block-group__inner-container {
+		padding: var(--wp--custom--margin--vertical) var(--wp--custom--margin--horizontal);
```
jffng, line 7: > Any blocks, inside of a group, set to alignwide should be as wide as it's container, however wide that is.

What about nested blocks that do not have an alignment set? Right now these are defaulting to the width of the container block, less the padding. IMO they should be the default width, meaning we may need a rule like this: 

```scss
	.wp-block-group__inner-container {
		*:not(.alignwide):not(.alignfull) {
			max-width: var(--wp--custom--width--default);
			margin-left: auto;
			margin-right: auto;
		}
	}
```

pbking: I kind of went back and forth about this one and wound up thinking otherwise (though I've gone back-and-forth).
[This](https://github.com/Automattic/themes/pull/3449/files#diff-24c1332a2754d3bffa38ace74a2f607b1c1a1b7fb3fc2cebb951ecea826cf811R39) ensures that if you nest GROUPS then that's the behavior so there's an avenue for that behavior.  But I kind of thought that if I made a group and made it full width, then dropped a paragraph in it then I would expect it to follow the width of the group I just added to it.  

I (mostly) don't disagree with you though, and to me that feels like behavior (whatever it is) that should be consistent across Themes and not opinionated.  How does that compare to how the incoming alignment changes are expected to default to?  I suggest standardizing on whatever behavior that is.

scruffian:
Let's put this on ice until we see what happens with https://github.com/WordPress/gutenberg/pull/29335

scruffian:
Actually since the work is done let's go ahead and merge this. If WordPress/gutenberg#29335 doesn't get merged then we can return to the editor styles later.

scruffian (approved):

PR #3454, opened 2021-03-11
Added default WP Prettier config

pbking:
With this change run `npm install` to install the [@wordpress/pretter-config](https://github.com/WordPress/gutenberg/blob/trunk/packages/prettier-config/README.md) which is loaded via package.json

Anybody not using [prettier](https://prettier.io/) shouldn't notice the change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,8 @@ To set this up follow these instructions:
 
 Now when you commit changes to a file PHPCBF will attempt to fix any issues with the file.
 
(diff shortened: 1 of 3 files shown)


scruffian:
I tried testing this but I couldn't get it to work. Today was not my day.

danieldudzic:
Tried getting it to work with Nova and the Prettier extension but unfortunately failed.

The config does seem to be correct code-wise, though.

MaggieCabrera (approved):
It worked for me on VSCode. I followed this (I already had the ESlint module installed): https://medium.com/@pgivens/write-cleaner-code-using-prettier-and-eslint-in-vscode-d04f63805dcd and tested using this example here: https://prettier.io/playground/

I'd say a little mention on the readme file won't do us any harm here. Otherwise: 🚢

pbking:
Documentation added.

Issue #3424, opened 2021-03-09
Mayland Blocks: Add support for spacing overrides


alaczek:
Similar to https://github.com/Automattic/themes/issues/2191, could we bring the spacing override CSS classes available in Varia:
* margin/padding-top/bottom/left/right-none
* margin/padding-top/bottom/left/right-half
* margin/padding-top/bottom/left/right-default

alaczek:
Closing, see PR discussion for details.

PR #3456, opened 2021-03-12
Mayland Blocks: Add support for spacing overrides

alaczek:
**Changes proposed in this Pull Request:**
Mayland and other Varia child themes have a few classes built-in that help to control the spacing. This PR introduces the same spacing override classes to Mayland Blocks.

**Related issue(s):**
Addresses #3424

**Testing:**

1. Checkout this branch
2. Zip the mayland-blocks theme to upload to a jurassic.ninja website.
3. Import & activate the theme into a WordPress jurassic.ninja site.
4. Open the Gutenberg editor and add two blocks. There should be a margin between them.
5. Open the HTML element inspector and modify the classname of either or both blocks.
6. Include variations of spacing overrides introduced in this PR, like `margin-top-none`, `margin-bottom-none`, `margin-top-half`, etc.
7. Verify that the margins on the blocks are modified.

### DIFF
--- a/mayland-blocks/experimental-theme.json
+++ b/mayland-blocks/experimental-theme.json
@@ -100,7 +100,8 @@
 					"headings": 1.125
 				},
 				"margin": {
-					"horizontal": "32px"
+					"horizontal": "32px",
+					"vertical": "32px"
 				},
 				"width": {
 					"default": "750px",
--- a/mayland-blocks/style.css
+++ b/mayland-blocks/style.css
@@ -65,3 +65,124 @@ body {
 h1, h2, h3 {
 	letter-spacing: -0.015em;
 }
+/**
+ * Spacing Overrides
+ */
+/*
+ * Margins
+ */
+.margin-top-none {
+	margin-top: 0 !important;
+}
+
+.margin-top-half {
+	margin-top: calc(0.5 * var(--wp--custom--margin--vertical)) !important;
+}
+
+.margin-top-default {
+	margin-top: var(--wp--custom--margin--vertical) !important;
+}
+
+.margin-right-none {
+	/*rtl:ignore*/
+	margin-right: 0 !important;
+}
+
+.margin-right-half {
+	/*rtl:ignore*/
+	margin-right: calc(0.5 * var(--wp--custom--margin--horizontal)) !important;
+}
+
+.margin-right-default {
+	/*rtl:ignore*/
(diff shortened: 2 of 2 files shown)


alaczek:
One thing I noticed but wasn't sure how to fix was the default top/bottom block margin (without any classes), which is 20px. I would love to bump it up to 32px.

scruffian:
I feel uncomfortable introducing these classes to a block based theme. If we need this kind of control then I think it should made available to users in the UI, but I'm curious what others think.

cc @jasmussen

jasmussen:
Yeah this is a tough one. In my own unreleased experimental theme, I've created a few "helper classes" like these myself, classes that I apply to blocks to have them do my bidding, in absence of the block editor itself supporting them. The good news is, nearly all of them I've been able to retire as the global styles project keeps gaining in features.

And ultimately I would agree with Ben, these all seem like classes that exist to be absorbed by the global styles project, so they are user configurable. The thing is, they will land there, hopefully sooner rather than later, I recall @youknowriad mentioning it just last week. And when they do, people will be confused in the cases where it doesn't work, because the `!important` is overriding it.

Is there a way to solve it that is less reliant on these "utility" helper classes? I know TT1 did some super opinionated stuff with style variations that might be a holdover?

alaczek:
To give some background, we're putting a few header patterns in preparation for full site editing, and manipulating the vertical space is essential with things like the visual grouping of elements (for example, site title and tagline):

rosetta_pattern

To Joen's idea, I saw this kind of thing in [Eksell theme](https://themes.trac.wordpress.org/ticket/96348), where a few blocks offer a "No Vertical Margin" style. Seeing it in action, I'm not sure this is the right way to go. It really feels like something that should be handled via block setting (which brings us back to Ben's point).

image


I agree that having this handled in core would be ideal. If adding these classes would later interfere with settings, then it's a solid reason no to proceed with this PR.

kjellr:
While we could technically build in some theme-specific overrides to address this, those sorts of spacing changes should be the same across all themes. Frankly, the lack of consistent margin + padding controls is one of the most glaring issues when trying to build block-based themes (or block patterns for that matter). 

This Gutenberg issue would address the issue, so I recommend we put our effort towards getting this one closed instead: https://github.com/WordPress/gutenberg/issues/28356

alaczek:
Agreed, closing, and let's focus on the core issue instead.

PR #3458, opened 2021-03-12
Add margin control to cover block

jffng:
This PR adds bottom margin control to the cover block. I only created one custom variable "margin", because the left and right margins are controlled by the overall alignment of the block and I'm not sure the top margin is needed.

### DIFF
--- a/blank-canvas-blocks/sass/blocks/_cover.scss
+++ b/blank-canvas-blocks/sass/blocks/_cover.scss
@@ -1,0 +1,3 @@
+.wp-block-cover.wp-block-cover.wp-block-cover {
+	margin-bottom: var(--custom--cover--margin);
+}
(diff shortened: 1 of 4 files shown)


blank-canvas-blocks/sass/blocks/_cover.scss
```diff
@@ -0,0 +1,3 @@
+.wp-block-cover.wp-block-cover.wp-block-cover {
```
jffng, line 1: Open to suggestions here but the 3x was needed to fight our alignment styles 😕

pbking:
[This change](https://github.com/Automattic/themes/pull/3459) removes the default margins.

I believe that with that change this can probably be closed (if indeed that comes in and[ retains this change](https://github.com/Automattic/themes/pull/3459/files#diff-24c1332a2754d3bffa38ace74a2f607b1c1a1b7fb3fc2cebb951ecea826cf811L14-L19)).

jffng:
Closing as we are removing default vertical margins.

PR #3459, opened 2021-03-12
Refactor/bcb alignmnet

### LINKED ISSUE #3419: [Blank Canvas Blocks] Configure Alignment & evaluate alignment polyfill
Ensure that the alignment ponyfill that exists is appropriate and correct.

Ensure that the "full/wide" values in the theme.json match the opinion of blank-canvas.

pbking:
After looking at https://github.com/WordPress/gutenberg/pull/29335, which I believe will be forthcoming in the next few weeks or months I have made some adjustments so that it plays nicer, and works more closely like, these upcoming changes.

I believe (I've only done minor testing) that ALL of the code in _alignments.scss will be remove-able with no changes to how things render once this lands... probably.  Additionally this change seems to play nice with the incoming allowing them to work in concert should there be overlap between this change in Gutenberg and updates to themes to support those changes.

Ultimately quite a bit has been removed and simplified and the result SEEMS to be the same.  I've checked all known alignment-related pages in the theamdemo content but there might be additional use-cases that I broke with this change that I didn't consider.

A couple of items of note: 

@jffng mentioned the default alignment of content inside of a wide/full block.  The current (unmodified) editor behavior seems to suggest the "default width" method while (unmodified) view suggest "follow parent size", however after the incoming change the behavior in both appears to be "follow the width of the block unless otherwise set".  The change introduces the ability for the user to manipulate that themselves (rather than depending on the theme which is the only option today) so that's a good thing, and I believe that following the behavior expressed in this theme will be the smoothest upgrade path (no changes without the user taking action).  I'm not 100% on this though.

There are no more theme-specific "left/right" alignment styles with this change.  There was an expression that would calculate and then limit the size of content to 1/2 of the "normal" width for left/right aligned blocks.  However (at least with this more minimal alignment base-styling) the left/right alignment of elements works fine out-of-the-box (seemingly with default/wide/full width containers) assuming those blocks are already small enough to align to one side or another.  Forcing that size seemed like overreach to me, especially for a ponyfill (perhaps a child theme could make that decision?  Or it could be expressed outside of the ponyfill?). Or maybe that's expected behavior and it indeed should be returned.  The incoming alignment changes don't appear to express any size constraints on alignment (but it's a bit PR and I may have missed it in review or testing).

The "default vertical margins" (originally copied in from Seedlet) has been removed.  After discussing with @jffng the selector that was using was SO VERY specific it was very difficult to override margins (specifically bottom margins) that supplying a "default margin" on all the content seemed really aggressive.  If we need to do that (and allow it to be configured via themes) then the method that was removed probably isn't what we need.

### DIFF
--- a/blank-canvas-blocks/assets/ponyfill.css
+++ b/blank-canvas-blocks/assets/ponyfill.css
@@ -75,12 +75,6 @@ body {
 	box-sizing: border-box;
 }
 
-.entry-content > *:not(:last-child):not(.wp-block-spacer),
-*[class*="__inner-container"] > *:not(:last-child):not(.wp-block-spacer),
-*[class*="__content"] > *:not(:last-child):not(.wp-block-spacer) {
-	margin-bottom: var(--wp--custom--margin--vertical);
-}
-
 .wp-site-blocks *[class*="__inner-container"] > .wp-block-group,
 .wp-site-blocks > *:not(.wp-block-post-content):not(.wp-block-template-part),
 .wp-site-blocks .wp-block-post-content > * {
--- a/blank-canvas-blocks/assets/ponyfill.css
+++ b/blank-canvas-blocks/assets/ponyfill.css
@@ -81,92 +81,26 @@ body {
 	margin-bottom: var(--wp--custom--margin--vertical);
 }
 
-.wp-site-blocks {
-	padding: 0 var(--wp--custom--margin--horizontal);
-}
-
-.wp-block {
-	max-width: var(--wp--custom--width--default);
-}
-
-.wp-block[data-align=wide] {
-	max-width: var(--wp--custom--width--wide);
-}
-
-.wp-block[data-align=full] {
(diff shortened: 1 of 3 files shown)


danieldudzic:
@pbking I gave it a try, and I like it. It seems simpler and more elegant.

I wasn't able to spot any differences between this and the old method, but also my review didn't include _all_ the blocks.

👍

scruffian:
I don't think you can remove all of the alignleft/right CSS. This is how the audio block should look:

Screenshot 2021-03-15 at 13 29 14

and in this branch:

Screenshot 2021-03-15 at 13 28 02

scruffian:
I think this is ready for another review.

pbking:
Looks fine to me.  Seems to play well enough with the experimental Gutenberg alignment branch, though I don't believe that change covers the left/right alignment sizing.  I expect we'll consider that when we remove this shim.

🚢

PR #3460, opened 2021-03-12
Add Navigation styling to BCB

pbking:
This change just adds configuration to the already-present navigation styles.
You'll have to add the navigation elements in FSE for this to show up.

This change also adds the drop shadow setting to match existing Blank Canvas style.

image

Additionally any `core/navigation` block that is in a `<header>` will receive "mobile" styling.
image
image

To support mobile styling quite a few additional custom styles have been added under `custom.navigation.mobile`.

Additionally javascript is loaded to support that menu by adding the button that opens the mobile menu.  It's as small an amount of JavaScript as I could add and still be functional simply creating two buttons - open and close - and labeling them based on CSS variables which add/remove a `show` class to the menu block.

### DIFF
--- a/blank-canvas-blocks/assets/navigation.js
+++ b/blank-canvas-blocks/assets/navigation.js
@@ -43,7 +43,7 @@
 	}
 
 	window.addEventListener( 'load', () => {
-		mobalizeNavigationBlock(document.querySelector( 'header .wp-block-navigation' ) );
+		mobilizeNavigationBlock( document.querySelector( 'header .wp-block-navigation' ) );
 	});
 
 })();
--- a/blank-canvas-blocks/assets/navigation.js
+++ b/blank-canvas-blocks/assets/navigation.js
@@ -15,6 +15,9 @@
 
 	function addMobileMenuOpenButton( navMenu ) {
 		const menuContainer = navMenu.querySelector( '.wp-block-navigation__container' );
+		if( !menuContainer ) {
+			return;
+		}
 		const openButton = document.createElement( 'button' );
 		const openButtonLabel = getComputedStyle(menuContainer).getPropertyValue( '--wp--custom--navigation--mobile--menu--open-label' );
 		openButton.classList.add( 'wp-block-navigation__mobile-menu-open-button' );
@@ -24,7 +27,10 @@
 			menuContainer.scrollTop = 0;
 			if( 0 === clickEvent.detail ) {
 				// Menu was opened with keyboard, apply focus to close button.
-				menuContainer.querySelector('button:first-of-type')?.focus();
+				const firstButton = menuContainer.querySelector('.wp-block-navigation__mobile-menu-close-button');
+				if( firstButton ) {
+					firstButton.focus();
+				}
 			}
 		});
 		navMenu.appendChild( openButton );
@@ -32,6 +38,9 @@
 
 	function addMobileMenuCloseButton( navMenu ) {
 		const menuContainer = navMenu.querySelector( '.wp-block-navigation__container' );
+		if( !menuContainer ) {
+			return;
+		}
 		const closeButton = document.createElement( 'button' );
 		const closeButtonLabel = getComputedStyle(menuContainer).getPropertyValue( '--wp--custom--navigation--mobile--menu--close-label' );
 		closeButton.classList.add( 'wp-block-navigation__mobile-menu-close-button' );
--- a/blank-canvas-blocks/assets/navigation.js
+++ b/blank-canvas-blocks/assets/navigation.js
@@ -28,7 +28,7 @@
 	}
 
 	window.addEventListener( 'load', () => {
-		Array.from( document.querySelectorAll( '.wp-block-navigation.is___experimental-mobile' ) )
+		Array.from( document.querySelectorAll( 'header .wp-block-navigation' ) )
 			.forEach( addMobileMenuButton );
 	});
 
--- a/blank-canvas-blocks/assets/navigation.js
+++ b/blank-canvas-blocks/assets/navigation.js
@@ -22,7 +22,7 @@
 		openButton.addEventListener( 'click', (clickEvent) => {
 			navMenu.classList.add( 'show' );
 			menuContainer.scrollTop = 0;
-			if( 0 === clickEvent.detail) {
+			if( 0 === clickEvent.detail ) {
 				// Menu was opened with keyboard, apply focus to close button.
 				menuContainer.querySelector('button:first-of-type')?.focus();
 			}
--- a/blank-canvas-blocks/assets/navigation.js
+++ b/blank-canvas-blocks/assets/navigation.js
@@ -1,0 +1,35 @@
+/**
+ * File navigation.js.
+ *
+ * Required to open the mobile navigation.
+ */
+(function () {
+
+	function addMobileMenuButton( navMenu ) {
+
+		const menuContainer = navMenu.querySelector( '.wp-block-navigation__container' );
+		const openButton = document.createElement( 'a' );
+		const openButtonLabel = getComputedStyle(menuContainer).getPropertyValue( '--wp--custom--navigation--mobile--menu--label' );
+		openButton.setAttribute( 'href', '#' );
+		openButton.classList.add( 'wp-block-navigation__mobile-menu-open-button' );
+		openButton.innerText = ( openButtonLabel );
+		openButton.addEventListener( 'click', () => {
+			navMenu.classList.add( 'show' );
+			menuContainer.scrollTop = 0;
+		});
+
+		menuContainer.addEventListener( 'click', (mouseEvent) => {
+			if ( mouseEvent.target === menuContainer ) {
+				navMenu.classList.remove( 'show' );
+			}
+		});
+
+		navMenu.appendChild( openButton );
+	}
+
+	window.addEventListener( 'load', () => {
+		Array.from( document.querySelectorAll( '.wp-block-navigation.is___experimental-mobile' ) )
+			.forEach( addMobileMenuButton );
+	});
+
+})();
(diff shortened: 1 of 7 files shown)


blank-canvas-blocks/sass/blocks/_navigation.scss
```diff
@@ -3,22 +3,15 @@
 	a {
 		border-bottom: none;
```
jffng, line 5: This line caught my attention, even though you didn't add it. I checked the small themes that we're looking to develop and indeed there are is no border bottom to be found, but I'm wondering why we need to do this here.

It looks like we set the default link style (not yet configurable) over here: https://github.com/Automattic/themes/blob/make/blank-canvas-blocks/blank-canvas-blocks/sass/elements/_links.scss#L7

Then we need to override it because the menu should not ever use underlined links? Does that make sense to you?

pbking: The anchor styles originally came pretty much as-is from Seedlet and that informed the navigation styles I think.  Is this the right opportunity to revisit those default anchor styles?  I believe @kjellr confirmed that the preferred default style (and likely default Gutenberg behavior) is to use the underline style as the primary link defining mechanism.

Seems to me like we should refactor that link style to use underline instead of border bottom and if we want to implement that effect (for instance in Seedlet-blocks) then the customization to do so would be to the responsibility of the theme.

blank-canvas-blocks/assets/navigation.js
```diff
@@ -0,0 +1,35 @@
...
+		const openButtonLabel = getComputedStyle(menuContainer).getPropertyValue( '--wp--custom--navigation--mobile--menu--label' );
+		openButton.setAttribute( 'href', '#' );
+		openButton.classList.add( 'wp-block-navigation__mobile-menu-open-button' );
+		openButton.innerText = ( openButtonLabel );
+		openButton.addEventListener( 'click', () => {
+			navMenu.classList.add( 'show' );
+			menuContainer.scrollTop = 0;
+		});
+
+		menuContainer.addEventListener( 'click', (mouseEvent) => {
```
jffng: For accessibility reasons, I think there needs to be an explicit close button, rather than clicking anywhere in the container to close it. This way, it can also be closed via keyboard navigation.

blank-canvas-blocks/assets/navigation.js
```diff
@@ -0,0 +1,35 @@
...
+				navMenu.classList.remove( 'show' );
+			}
+		});
+
+		navMenu.appendChild( openButton );
+	}
+
+	window.addEventListener( 'load', () => {
+		Array.from( document.querySelectorAll( 'header .wp-block-navigation' ) )
+			.forEach( addMobileMenuButton );
```
jffng: By this logic — there could be multiple nav blocks in the header, and each one would receive its own separate mobile menu?

blank-canvas-blocks/assets/navigation.js
```diff
@@ -0,0 +1,35 @@
...
+ * File navigation.js.
+ *
+ * Required to open the mobile navigation.
+ */
+(function () {
+
+	function addMobileMenuButton( navMenu ) {
+
+		const menuContainer = navMenu.querySelector( '.wp-block-navigation__container' );
+		const openButton = document.createElement( 'a' );
```
jffng: Semantically, should this be a button instead of an `a`?

blank-canvas-blocks/functions.php
```diff
@@ -32,6 +32,7 @@ function blank_canvas_blocks_support()  {
  * Enqueue scripts and styles.
  */
 function blank_canvas_blocks_scripts() {
+	wp_enqueue_script( 'blank-canvas-navigation-script', get_template_directory_uri().'/assets/navigation.js', array(), wp_get_theme()->get( 'Version' ), true );
```
jffng:
```suggestion
	wp_enqueue_script( 'blank-canvas-navigation-script', get_template_directory_uri() . '/assets/navigation.js', array(), wp_get_theme()->get( 'Version' ), true );
```

blank-canvas-blocks/assets/navigation.js
pbking: Yes, that's correct.  Initially I wrote that leveraging an "opt in class" rather than using the header.

Either way though that maybe should change to just `.querySelector` so we just execute this behavior on the FIRST menu that is in the header.  Which might not always be the menu that we want but I'm not sure what else would be best...

blank-canvas-blocks/assets/navigation.js
pbking: Yes, absolutely.  Thank you.  I used to be quite the hound about that very thing... :P

blank-canvas-blocks/assets/navigation.js
pbking: Good call.  I had wanted to avoid that to play golf with the amount of JavaScript but you're right there.

pbking: Close **button** added.

blank-canvas-blocks/assets/navigation.js
pbking: Refactor to button complete.

blank-canvas-blocks/assets/navigation.js
pbking: Refactored to use `querySelector` instead.  If/when the multiple menus comes up we'll refactor as needed then.

blank-canvas-blocks/sass/blocks/_navigation.scss
pbking: Links are refactored to use underline here: https://github.com/Automattic/themes/pull/3478

blank-canvas-blocks/assets/navigation.js
```diff
@@ -0,0 +1,50 @@
+/**
+ * File navigation.js.
+ *
+ * Required to open the mobile navigation.
+ */
+(function () {
+
+	function mobalizeNavigationBlock( navMenu ) {
```
jffng: Sorry, this is not meant to be pedantic :)

```suggestion
	function mobilizeNavigationBlock( navMenu ) {
```

blank-canvas-blocks/assets/navigation.js
```diff
@@ -0,0 +1,50 @@
...
+		closeButton.classList.add( 'wp-block-navigation__mobile-menu-close-button' );
+		closeButton.innerText = closeButtonLabel;
+		closeButton.addEventListener( 'click', () => {
+			navMenu.classList.remove( 'show' );
+		});
+		menuContainer.prepend( closeButton );
+	}
+
+	window.addEventListener( 'load', () => {
+		mobalizeNavigationBlock(document.querySelector( 'header .wp-block-navigation' ) );
```
jffng:
```suggestion
		mobilizeNavigationBlock( document.querySelector( 'header .wp-block-navigation' ) );
```

blank-canvas-blocks/assets/navigation.js
```diff
@@ -0,0 +1,50 @@
...
+		const openButtonLabel = getComputedStyle(menuContainer).getPropertyValue( '--wp--custom--navigation--mobile--menu--open-label' );
+		openButton.classList.add( 'wp-block-navigation__mobile-menu-open-button' );
+		openButton.innerText = openButtonLabel;
+		openButton.addEventListener( 'click', (clickEvent) => {
+			console.log('clickEvent', clickEvent);
+			navMenu.classList.add( 'show' );
+			menuContainer.scrollTop = 0;
+			if( 0 === clickEvent.detail) {
+				// Menu was opened with keyboard, apply focus to close button.
+				menuContainer.querySelector('button:first-of-type')?.focus();
```
jffng: To confirm, we can use optional chaining (`?.`) because we are _not_ supporting IE, correct?

blank-canvas-blocks/assets/navigation.js
```diff
@@ -0,0 +1,50 @@
...
+		const menuContainer = navMenu.querySelector( '.wp-block-navigation__container' );
+		const openButton = document.createElement( 'button' );
+		const openButtonLabel = getComputedStyle(menuContainer).getPropertyValue( '--wp--custom--navigation--mobile--menu--open-label' );
+		openButton.classList.add( 'wp-block-navigation__mobile-menu-open-button' );
+		openButton.innerText = openButtonLabel;
+		openButton.addEventListener( 'click', (clickEvent) => {
+			console.log('clickEvent', clickEvent);
+			navMenu.classList.add( 'show' );
+			menuContainer.scrollTop = 0;
+			if( 0 === clickEvent.detail) {
```
jffng:
```suggestion
			if( 0 === clickEvent.detail ) {
```

blank-canvas-blocks/assets/navigation.js
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	function addMobileMenuOpenButton( navMenu ) {
+		const menuContainer = navMenu.querySelector( '.wp-block-navigation__container' );
+		const openButton = document.createElement( 'button' );
+		const openButtonLabel = getComputedStyle(menuContainer).getPropertyValue( '--wp--custom--navigation--mobile--menu--open-label' );
+		openButton.classList.add( 'wp-block-navigation__mobile-menu-open-button' );
+		openButton.innerText = openButtonLabel;
+		openButton.addEventListener( 'click', (clickEvent) => {
+			console.log('clickEvent', clickEvent);
```
jffng: Remove this before merging?

```suggestion
```

jffng (approved):
Left a few minor comments, but in general this LGTM! I can live with the amount of code this adds to give us a temporary solution to navigation. We can iterate as needed when we implement a few child themes.

blank-canvas-blocks/assets/navigation.js
pbking: Oh, man, force of habit.  Previously I ran all JS code through Babel to transpile it for IE.  
None of this would work in IE though since it's so dependent on CSS Variables.  Easy enough to add a null check instead though in case the CSS vars polyfill can do the trick here.

pbking: Refactored out the cool stuff.
Replaced with boring 'ol null checks.

pbking:
This (mobile) shim can be replaced when https://github.com/WordPress/gutenberg/issues/22824 is completed.

PR #3469, opened 2021-03-15
Blank Canvas Blocks: Add default styles for search block

MaggieCabrera:
This PR adds styles for the search block. I went ahead and tweaked it as much as I could so it will look fine for Seedlet. This block required a little extra customization on the padding and line heights for all the variations to look the same height.

Part of https://github.com/Automattic/themes/issues/3419

View: 
image

Editor:
image

(note that the emboldened "Search" is caused by the editor improperly loading the "theme.css" files)

### DIFF
--- a/blank-canvas-blocks/experimental-theme.json
+++ b/blank-canvas-blocks/experimental-theme.json
@@ -128,8 +128,7 @@
 					}
 				},
 				"form": {
-					"padding": "calc( 0.5 * var(--wp--custom--padding--horizontal) )",
-					"fontSize": "var(--wp--preset--font-size--normal)",
+					"padding": "calc( 0.5 * var(--wp--custom--margin--horizontal) )",
 					"border": {
 						"radius": "0",
 						"color": "#EFEFEF",
@@ -169,13 +168,6 @@
 					"color": "#ccc",
 					"thickness": "2px",
 					"width": "150px"
-				},
-				"search": {
-					"button": {
-						"lineHeight": "27px",
-						"padding": "calc( .9 * var(--wp--custom--padding--horizontal) )",
-						"margin": "0 0 0 calc( 1.4 * var(--wp--custom--padding--horizontal))"
-					}
 				},
 				"video": {
 					"caption": {
@@ -279,12 +271,6 @@
 			},
 			"border": {
 				"radius": "var(--wp--custom--button--border--radius)"
-			}
-		},
-		"core/search": {
-			"color": {
-				"text": "var(--wp--custom--color--background)",
-				"background": "var(--wp--custom--color--secondary)"
 			}
 		},
 		"core/heading/h1": {
--- a/blank-canvas-blocks/experimental-theme.json
+++ b/blank-canvas-blocks/experimental-theme.json
@@ -88,7 +88,16 @@
 				"contentSize": "620px",
 				"wideSize": "1000px"
 			},
+			"custom-comments":{
+				"root": "I copied the root typography preset variables to custom until this (or a similar PR) gets merged: https://github.com/WordPress/gutenberg/pull/29714",
+				"search-block": "the core/search variables won't work until https://github.com/WordPress/gutenberg/pull/29830 is merged"
+			},
 			"custom": {
+				"typography": {
+					"fontSize": "var(--wp--preset--font-size--normal)",
+					"lineHeight": "1.6",
+					"fontFamily": "var(--wp--preset--font-family--base)"
+				},
 				"color": {
 					"primary": "var(--wp--preset--color--black)",
 					"secondary": "var(--wp--preset--color--blue)",
@@ -119,10 +128,11 @@
 					}
 				},
 				"form": {
-					"padding": "5px",
+					"padding": "calc( 0.5 * var(--wp--custom--padding--horizontal) )",
+					"fontSize": "var(--wp--preset--font-size--normal)",
 					"border": {
 						"radius": "0",
-						"color": "#cccccc",
+						"color": "#EFEFEF",
 						"width": "2px",
 						"style": "solid"
 					},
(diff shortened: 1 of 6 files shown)


blank-canvas-blocks/experimental-theme.json
```diff
@@ -161,6 +175,15 @@
 					"thickness": "2px",
 					"width": "150px"
 				},
+				"search": {
+					"button": {
+						"lineHeight": "27px",
+						"padding": "calc( .9 * var(--wp--custom--padding--horizontal) )"
+					},
+					"input": {
+						"margin": "0 calc( .66 * var(--wp--custom--padding--horizontal)) 0 0"
```
jffng: In Gutenberg, the search button styles have a `margin-left` applied rather than the input. 

Screen Shot 2021-03-16 at 12 31 45 PM

Do you think it would make sense to add margin to the button (overwriting Gutenberg's styles) instead of adding a new margin control to the input element? Right now, the result is two margins sitting next to each other:

Screen Shot 2021-03-16 at 12 34 34 PM

blank-canvas-blocks/experimental-theme.json
```diff
@@ -161,6 +175,15 @@
 					"thickness": "2px",
 					"width": "150px"
 				},
+				"search": {
+					"button": {
+						"lineHeight": "27px",
```
jffng: I think I understand why this is needed stylistically, but feels opaque to me. Like, I need to set the line height and padding on the button in order to control its overall height...

MaggieCabrera: yeah, and the icon is not going to be the same height as the text, so that's why we need to streamline it. We could do it in the CSS but it's needed for both Seedlet and BCB. so I didn't know of a more elegant way to do it.

blank-canvas-blocks/sass/blocks/_button.scss
```diff
@@ -7,7 +7,8 @@
 	font-family: var(--wp--custom--button--font-family);
 	font-size: var(--wp--custom--button--font-size);
 	line-height: var(--wp--custom--button--line-height);
-	border-radius: var(--wp--custom--button--border-radius);
+	border-radius: var(--wp--custom--button--border--radius);
+	border-width: var(--wp--custom--button--border--width);
```
pbking, line 11: Because of the unique calculations that the button block is doing with the assumption of a 2px border-width (to make the regular and outline-style buttons the exact same height) I'm a little nervous opening that up for configuration.

blank-canvas-blocks/sass/ponyfill.scss
```diff
@@ -9,14 +9,15 @@
 // Blocks
 // - These styles replace key Gutenberg Block styles for fonts, colors, and
 //   spacing with CSS-variables overrides
-@import "blocks/heading";
 @import "blocks/button";
 @import "blocks/gallery";
+@import "blocks/group";
+@import "blocks/heading";
```
scruffian: Thank you!

blank-canvas-blocks/sass/blocks/_search.scss
```diff
@@ -0,0 +1,32 @@
...
+		border-radius: var(--wp--custom--form--border--radius);
+		color: var(--wp--custom--form--color--text);
+		padding: var(--wp--custom--form--padding);
+		background: var(--wp--custom--form--color--background);	
+		font-size: var(--wp--custom--form--font-size);
+		line-height: var(--wp--custom--typography--line-height);
+	}
+
+	.wp-block-search__button {
+		@include button-main-styles;
```
scruffian, line 23: What do you think about including the file that the mixin is in, to make the dependencies explicit?

blank-canvas-blocks/experimental-theme.json
```diff
@@ -161,6 +175,13 @@
...
+						"padding": "calc( .9 * var(--wp--custom--padding--horizontal) )",
```
scruffian: If we're making the padding on this button customizable by theme.json, should we do the same for the button block?

scruffian:
When I ran this I saw a few issues:

The search button should probably have the same color as the default:
Screenshot 2021-03-17 at 13 50 10

There are some spacing issues in the editor:
Screenshot 2021-03-17 at 13 50 19

MaggieCabrera:
> The search button should probably have the same color as the default:

Yes, that's what this is for: https://github.com/WordPress/gutenberg/pull/29830 and I added the core/search properties so that when/if that lands, it will be set to the same color as the rest of the buttons.

MaggieCabrera:
> There are some spacing issues in the editor:

Whose appeared after the spacer PR was merged 🤔

blank-canvas-blocks/sass/ponyfill.scss
pbking: And now those two are gone again!

blank-canvas-blocks/sass/blocks/_search.scss
pbking: Done

blank-canvas-blocks/experimental-theme.json
pbking: Padding customizations are out (in this PR).  Doing so at the button block level makes problems and I believe we should avoid.  The cleanest way to express that padding opinion seems to be line-height.

pbking:
This branch no longer depends on WordPress/gutenberg#29830, though if it lands styles can probably be cleaned out and refactored a bit.

blank-canvas-blocks/experimental-theme.json
pbking: Rather than enforcing a line-height of a configured value to bring the icon into alignment, setting the line-height when an icon is used seems to do the same job without the need for tweaking to match.

jffng:
This is what I see on Gutenberg trunk + this PR — is this expected? Are we waiting? The button padding looks a bit squishy compared to [Blank Canvas](pcrugM-hf-p2).

Editor | View 
------- | --------
Screen Shot 2021-03-19 at 2 31 36 PM | Screen Shot 2021-03-19 at 2 31 06 PM

pbking:
Just pushed another change that incorporates the default padding value to the non-button-block-button styles.  Looks, maybe too blown out?  But I think that's how the buttons would look, unmodified, once they are button blocks instead.

Thoughts?
image

scruffian (approved):
LGTM

PR #3470, opened 2021-03-15
Varia: fix color utils

jffng:
This PR attempts to address a 'Division by Zero: warning'. I'm not sure why the exact equality comparison is failing to catch when `$chroma` is zero, so this is a different approach.


jeffikus (approved):
LGTM. I think the problem here is that with the === operator, that could evaluate to boolean of false as well as zero. I think your solution will solve that. Thank you!

PR #3476, opened 2021-03-16
Blank Canvas Blocks: File block: Add rules for the button

scruffian:
This adds rules for the button inside the file block, so it matches the button block.

#### Related issue(s):
https://github.com/Automattic/themes/issues/3413

### DIFF
--- a/blank-canvas-blocks/sass/blocks/_file.scss
+++ b/blank-canvas-blocks/sass/blocks/_file.scss
@@ -5,13 +5,15 @@
 	a.wp-block-file__button:focus,
 	a.wp-block-file__button:hover,
 	a.wp-block-file__button:visited {
-		color: var(--wp--custom--color--background);
-		background: var(--wp--custom--color--secondary);
+		color: var(--wp--custom--button--color--hover-text);
+		background: var(--wp--custom--button--color--hover-background);
 	}
 
 	.wp-block-file__button {
 		// Extend button style
 		@include button-main-styles;
+		color: var(--wp--custom--button--color--text);
+		background: var(--wp--custom--button--color--background);
 		padding: calc(0.667em + 2px) calc(1.333em + 2px);
 	}
 }
--- a/blank-canvas-blocks/sass/blocks/_file.scss
+++ b/blank-canvas-blocks/sass/blocks/_file.scss
@@ -1,4 +1,4 @@
-// Needed until https://github.com/WordPress/gutenberg/issues/27760 is fixed
+// TODO: Remove when https://github.com/WordPress/gutenberg/issues/27760 is fixed
 .wp-block-file {
 	// Undo Gutenberg hover defaults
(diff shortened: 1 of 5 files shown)


blank-canvas-blocks/sass/blocks/_file.scss
```diff
@@ -0,0 +1,17 @@
+// Needed until https://github.com/WordPress/gutenberg/issues/27760 is fixed
```
jffng: Minor suggestion but could make our lives easier in the future:

```suggestion
// TODO: Remove when https://github.com/WordPress/gutenberg/issues/27760 is fixed
```

blank-canvas-blocks/sass/blocks/_file.scss
```diff
@@ -0,0 +1,17 @@
+// TODO: Remove when https://github.com/WordPress/gutenberg/issues/27760 is fixed
+.wp-block-file {
+	// Undo Gutenberg hover defaults
+	a.wp-block-file__button:active,
+	a.wp-block-file__button:focus,
+	a.wp-block-file__button:hover,
+	a.wp-block-file__button:visited {
+		color: var(--wp--custom--color--background);
+		background: var(--wp--custom--color--secondary);
```
pbking: Would it be unreasonable to return the `custom.button.color.text` and `custom.button.color.background` values to theme.json?  The `core/button` style could be mapped to that variable (the right way) and the things like this (and search block) could use the same color variable.

In this way changing the color value of `custom.button.color.text/.background` changes the colors of all the buttons, and the choice to leverage `custom.color.primary` or `custom.color.secondary` as the button color is at the discretion of the configuration of theme.json instead of a decision to be made in the polyfill.

MaggieCabrera: This makes sense to me too

scruffian:
I updated this to use the same variables in both the file block button and the button block.

blank-canvas-blocks/sass/elements/_links.scss
```diff
@@ -13,7 +13,7 @@ a:not(.wp-block-button__link) {
 }
 
 // Links that are in the content
-.entry-content a:not(.wp-block-button__link) {
+.entry-content a:not(.wp-block-button__link):not(.wp-block-file__button) {
```
pbking, line 16: Regarding the nots see some discussion here: https://github.com/Automattic/themes/pull/3478#discussion_r596028025

Do you think we can overcome these styles in the `.wp-bloc-file__button` class instead of proving a :not() here?   Or perhaps [this change](https://github.com/Automattic/themes/commit/3c05ff6c48a2f4d4ee53ffc84abfeb2814a58ce3) already encompasses that.

pbking:
Potential alternative: #3494

scruffian:
Closing in favour of https://github.com/Automattic/themes/pull/3494

PR #3478, opened 2021-03-16
Refactor links in BCB to use underline instead of border-bottom.

pbking:
This change refactors the link styles to use underline instead of border bottom.

This change continues to differentiate between links "in the content" (which have underlines until they have focus at which point they are removed) and "out of the content" (which do NOT have underlines until they have focus at which point they do).

image

The following CSS can be used (for instance by Seedlet-Blocks) to return the behavior to Seedlet's original expression:

```
a:not(.wp-block-button__link) {
	color: var( --wp--style--color--link);
	text-decoration: none;
	border-bottom: 1px solid transparent;
	transition: border-color 1s ease-in;
}

a:not(.wp-block-button__link):hover,
a:not(.wp-block-button__link):focus {
	border-bottom-color: currentColor;
}

.entry-content a:not(.wp-block-button__link) {
	border-bottom: 1px solid currentColor;
}

.entry-content a:not(.wp-block-button__link):hover,
.entry-content a:not(.wp-block-button__link):focus {
	border-bottom-color: transparent;
}
```


blank-canvas-blocks/sass/elements/_links.scss
```diff
@@ -2,29 +2,23 @@
...
-		border-bottom-color: currentColor;
+		text-decoration: underline;
 	}
 }
 
 // Links that are in the content
 .entry-content a:not(.wp-block-button__link) {
-	border-bottom: 1px solid currentColor;
-	text-decoration: none;
+	text-decoration: underline;
```
jffng, line 16: Since it's top of mind (#3476), this will interfere with the file block: 

Screen Shot 2021-03-16 at 4 17 51 PM

pbking: Hrm.  I **already** didn't like including the :not(.wp-block-button__link), I'm not keen to add more and actually would maybe prefer to refactor that existing :not() out and instead target .wp-block-button__link directly in the button block styles, and the file block's button class to set the text-decoration to none.

Thoughts on the best way to address?

jffng: I like your idea! If I understood it correctly, I tried it in https://github.com/Automattic/themes/pull/3478/commits/3c05ff6c48a2f4d4ee53ffc84abfeb2814a58ce3. It plays nicely with #3476 too since we're making use of the `button-main-styles` mixin.

scruffian: I don't think this is going to work for the editor either, since there's no `.entry-content` element in the editor. I think you can use `.site-blocks` instead.

scruffian: This is what I see in the editor: Screenshot 2021-03-17 at 17 41 30

scruffian: Turns out `wp-site-blocks` is no good as it wraps the whole site. In the end I had to use two classes - one for the post editor and one for the frontend and site editor.

blank-canvas-blocks/sass/elements/_links.scss
```diff
@@ -1,30 +1,23 @@
...
 
-// Links that are in the content
-.entry-content a:not(.wp-block-button__link) {
-	border-bottom: 1px solid currentColor;
-	text-decoration: none;
+// Links that appear in the main content area
+.block-editor-block-list__layout a, // Needed for the post area
+.wp-block-post-content a {
+	text-decoration: underline;
 	transition: border-color 0.1s ease-out;
```
jffng: I don't think this line does anything anymore because we're not using borders? 
```suggestion
```

jffng (approved):
I left one small comment but other than that this LGTM.

blank-canvas-blocks/functions.php
```diff
@@ -44,6 +44,7 @@ function blank_canvas_blocks_scripts() {
 
 	wp_enqueue_script( 'blank-canvas-navigation-script', get_template_directory_uri() . '/assets/navigation.js', array(), wp_get_theme()->get( 'Version' ), true );
 	wp_enqueue_style( 'blank_canvas_blocks-ponyfill', get_template_directory_uri() . '/assets/ponyfill.css', array(), wp_get_theme()->get( 'Version' )  );
+	wp_enqueue_style( 'blank_canvas_blocks', get_stylesheet_uri(), array('blank_canvas_blocks-ponyfill'), wp_get_theme()->get( 'Version' )  );
```
scruffian: We should remove this too.

```suggestion
```

PR #3482, opened 2021-03-17
Blank Canvas Blocks: Fix getting data from theme json

scruffian:
To get data from the theme.json in the functions.php file, we call this Gutenberg function:

`WP_Theme_JSON_Resolver::get_merged_data()->get_settings()`

This sanitizes the theme.json against the currently registered blocks and saves it into memory, so that future calls to the theme.json don't need to run the process again. The problem is that if this happens too early, no block are registered, so settings like "core/button" in the theme.json are removed. By hooking this function call later in the running order (admin_init) we ensure that the blocks we need are registered so that the theme.json loads correctly.

To test, add some headings to a page and a button. Check that the headings have the "Playfair Display" font and the buttons are blue, in both the frontend and the editor.

### DIFF
--- a/blank-canvas-blocks/assets/ponyfill.css
+++ b/blank-canvas-blocks/assets/ponyfill.css
@@ -256,8 +256,8 @@ input[type=checkbox] + label {
 
 h1, h2, h3, h4, h5, h6 {
 	clear: both;
-	font-family: var(--wp--custom--heading--typography--font-family);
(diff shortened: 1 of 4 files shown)


pbking:
Works for me!
🎆 

Should we remove the header font before merging this in?
FWIW I quite like there being a different header, but I understand it's not the original design and maybe shouldn't stay.

scruffian:
Yeah, those changes were just for testing. I'll clean this up and get it merged.

PR #3485, opened 2021-03-17
Add CORE/CODE configurable styles to BCB

pbking:
This change adds configurable border (color/style/width/radius) and color (text/background) to the core/code block via custom theme.json configuration.

The [code block example for Blank Canvas](https://theamdemo.wordpress.com/2020/03/05/gutenberg-misc-formatting/) doesn't seem to be behaving properly (I couldn't say why...) so the styles from the form configuration were used for this block's border configuration value.

Additionally the "padding" attributes applied in _alignment have been r[efactored slightly to be less specific](https://github.com/Automattic/themes/pull/3485/files#diff-24c1332a2754d3bffa38ace74a2f607b1c1a1b7fb3fc2cebb951ecea826cf811L22).  As it was a block was unable to define a padding variable without additional specificity hacking.  This was discovered while attempting to add custom padding configuration to this core/code block per the linked Gutenberg PR.  Additionally that code was (inappropriately) enforcing a top/bottom padding of 0 on all the blocks, even if that wasn't their natural state.

In the [Gutenberg change that adds the ability to configure this block](https://github.com/WordPress/gutenberg/pull/27582) additionally the font-stylings were discussed as something to be added.  However there wasn't a need to do that and bring the styles in alignment with Blank Canvas so that customization option has been deferred until it's needed (and hopefully that will be after [something like this](https://github.com/WordPress/gutenberg/issues/29778) is complete).

Before:
image

After:
image

### DIFF
--- a/blank-canvas-blocks/assets/ponyfill.css
+++ b/blank-canvas-blocks/assets/ponyfill.css
@@ -77,24 +77,27 @@ body {
 
 .wp-site-blocks *[class*="__inner-container"] > .wp-block-group,
 .wp-site-blocks > *:not(.wp-block-post-content):not(.wp-block-template-part),
-.wp-site-blocks .wp-block-post-content > * {
+.wp-block-post-content > * {
 	max-width: var(--wp--custom--width--default);
 	margin-left: auto;
 	margin-right: auto;
-	padding: 0 var(--wp--custom--margin--horizontal);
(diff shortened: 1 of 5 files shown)


pbking:
This is no longer needed since it can be done via styles in [this branch](https://github.com/Automattic/themes/pull/3487) instead.

PR #3487, opened 2021-03-17
Refactor/bcb custom to styles

### LINKED ISSUE #3486: Evaluate what shims can be removed from BCB
There is a [change coming in Gutenberg](https://github.com/WordPress/gutenberg/pull/29941) which will allow for the configuration of any style that has been [registered here](https://github.com/WordPress/gutenberg/blob/trunk/lib/class-wp-theme-json.php#L259).

Those attributes include:
```
style: {
  color: {
    background
    text
    link
  }
  border: {
    radius
    color
    width
    style
  }
  typography: {
    fontFamily
    fontSize
    fontStyle
    fontWeight
    lineHeight
    textDecoration
    textTransform
  }
  spacing: {
    padding: {
      top
      right
      left
      bottom
    }
  }
}
```

Any block that has custom styles that targets any of the above attributes MIGHT be ripe to refactor.  It's understood that NOT ALL STYLE ATTRIBUTES WILL WORK AS EXPECTED FOR ALL BLOCKS but many of them are likely to work out-of-the-box.

pbking:
This change attempts to implement #3486 by moving whatever style attributes are possible from `custom: {` to `style: {`

Some "custom" things had to stay in custom and are referenced from style (such as the button color, so that other CSS bits could still make use of those values).

Others deal with nested elements and can't be taken care of in this way (for example, image captions).

Others still had problems with the CSS that Gutenberg is dealing and couldn't be converted for some reason or another (such as the separator which has issues customizing the style when it's "wide").

Lastly adjustments have been made to the alignment styles, primarily to remove as much as what was there as possible.  "block-styles" have been removed eliminating the "reasonable defaults" allowing us to work with more defaulty defaults.  The post-content block defaults to "inherit layout" in the index template (in lieu of that being the default behavior) but will need [this PR](https://github.com/WordPress/gutenberg/pull/29982) before it works properly.  A padding value is applied to the "post-content" and "negative margin" behavior for .alignwide elements added allowing them to accomplish "full viewport width" for those particular elements.  (Which was agreed should be behavior that a theme should implement rather than be ingrained behavior.)


blank-canvas-blocks/sass/blocks/_button.scss
```diff
@@ -1,15 +1,17 @@
...
-	font-weight: var(--wp--custom--button--font-weight);
-	font-family: var(--wp--custom--button--font-family);
-	font-size: var(--wp--custom--button--font-size);
-	line-height: var(--wp--custom--button--line-height);
-	border-radius: var(--wp--custom--button--border-radius);
+	font-weight: var(--wp--custom--button--typography--font-weight);
+	font-family: var(--wp--custom--button--typography--font-family);
+	font-size: var(--wp--custom--button--typography--font-size);
+	border-radius: var(--wp--custom--button--border--radius);
 	text-decoration: none; // Needed because link styles inside .entry-content add a text decoration
```
jffng, line 12: I think this line now needs to be moved to the `.wp-block-button__link` selector so that buttons don't appear with an underline: 

Screen Shot 2021-03-18 at 5 51 26 PM

blank-canvas-blocks/experimental-theme.json
```diff
@@ -97,20 +101,19 @@
...
 				"button": {
-					"fontWeight": "normal",
-					"fontFamily": "var(--wp--preset--font-family--base)",
-					"fontSize": "var(--wp--preset--font-size--small)",
-					"borderRadius": "4px",
+					"typography": {
+						"fontWeight": "normal",
+						"fontFamily": "var(--wp--preset--font-family--base)",
+						"fontSize": "var(--wp--preset--font-size--small)",
+						"lineHeight": ""
```
jffng: Should this value be empty?

blank-canvas-blocks/sass/blocks/_group.scss
```diff
@@ -1,6 +0,0 @@
-// NOTE: Gutenberg sets a padding value by default for groups with a background.
```
jffng, line 1: Since we no longer `add_editor_support( 'wp-block-styles' )`, removing these styles incur a difference in the editor and frontend for me, I just want to verify we're aware of that: 

Editor | Front-end
------ | ------
Screen Shot 2021-03-18 at 5 55 51 PM | Screen Shot 2021-03-18 at 5 55 46 PM

jffng:
I have tested this with Ben's PR (#29982) checked out and is looking good, especially since it removes ~3x as much code as it adds!

Pending a few minor comments, I think it's in a spot to merge to `make` and iterate on.

blank-canvas-blocks/sass/blocks/_group.scss
pbking: I understand that to be a reported Gutenberg issue where those styles are loaded when they shouldn't be.  I also thought that it was understood to be addressed though so perhaps it needs to be reopened.  I don't believe we should be attempting to overcome that difference here.

blank-canvas-blocks/experimental-theme.json
pbking: Probably not, and likely also needs to be referenced below in the "styles" block.

blank-canvas-blocks/sass/blocks/_group.scss
MaggieCabrera: Yes, this is the editor loading theme.css even though we removed the support for it on functions.php :(

blank-canvas-blocks/sass/blocks/_button.scss
pbking: Thanks.  Added to reg'ler block button stylin' too.

blank-canvas-blocks/experimental-theme.json
pbking: These have been returned.

PR #3494, opened 2021-03-19
Style BCB File Block Buttons

pbking:
This is an alternative to #3476 that is a bit simpler and depends on some of the changes found in #3469.  This branch includes those changes so shouldn't be merged until that has landed.

View:
image

Editor:
image


scruffian:
This was looking off in the editor, so I made some changes to get the editor to match with the front end.

pbking:
Looks good to me.  Bringing this is.

PR #3501, opened 2021-03-22
Blank Canvas Blocks: Add Pullquote Block styles

scruffian:
This adds styles for the Pullquote block in Blank Canvas Blocks

Editor:
Screenshot 2021-03-22 at 13 52 13

Frontend:
Screenshot 2021-03-22 at 13 52 20

#### Related issue(s):
Part of https://github.com/Automattic/themes/issues/3413

### DIFF
--- a/blank-canvas-blocks/experimental-theme.json
+++ b/blank-canvas-blocks/experimental-theme.json
@@ -155,6 +155,12 @@
 						}
 					}
 				},
+				"pullquote": {
+					"typography": {
+						"fontSize": "var(--wp--preset--font-size--large)",
+						"fontStyle": "italic"
+					}
+				},
 				"separator": {
 					"color": "#ccc",
 					"thickness": "2px",
--- a/blank-canvas-blocks/experimental-theme.json
+++ b/blank-canvas-blocks/experimental-theme.json
@@ -153,7 +153,7 @@
 					},
 					"citation": {
 						"typography": {
-							"fontSize": "0.8em",
+							"fontSize": "var(--wp--preset--font-size--tiny)",
 							"fontStyle": "italic"
 						}
 					}
@@ -164,7 +164,7 @@
 					},
 					"citation": {
 						"typography": {
-							"fontSize": "0.5em",
+							"fontSize": "var(--wp--preset--font-size--tiny)",
 							"fontStyle": "italic"
 						}
 					}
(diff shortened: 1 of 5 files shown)


pbking:
I made a few changes to this PR to refactor some styles from "custom" to "styles" that were available.

Additionally the expressed styles in theme.json align more closely with the styles already expressed by "Blank Canvas" here: ([quote](https://theamdemo.wordpress.com/2019/06/24/gutenberg-quote/), [pullquote](https://theamdemo.wordpress.com/2019/06/25/gutenberg-pullquote/)).

During this change "Quote" changed a bit in the same way and should have another look-see as well.

Pullquote View:
image

Pullquote Editor:
image

Quote View:
image

Quote Editor:
image

blank-canvas-blocks/experimental-theme.json
```diff
@@ -148,9 +148,23 @@
 					}
 				},
 				"quote": {
+					"typography": {
+						"textAlign": "left"
+					},
 					"citation": {
 						"typography": {
-							"fontSize": "var(--wp--preset--font-size--small)",
+							"fontSize": "0.8em",
```
scruffian: IMO we should be using a size from the theme rather than a new size.

pbking: I liked that by using `em` units in this way the citation tracks against the size of the quote.  And particularly since the quote citation font size isn't something the user has control over.  Ultimately I see that as a theme design decision though and I'm mostly impartial as to what that "default" decision is, only that it is one that CAN be expressed.

scruffian: Makes sense. My thought was that one important thing that keeps designs from looking messy is having a small set of font sizes, so my concern is that this would add another. Maybe @kjellr can decide?

scruffian:
I noticed that there was some extra space around the citation in the editor vs the frontend, so I added a bit more CSS so they match.

blank-canvas-blocks/experimental-theme.json
kjellr: Yeah, we should use a standard theme size for this.

pbking: 👍 Makes sense to use a standardized size.  I don't know why I've been on such an 'em' kick lately...
Replaced both with `--wp--preset--font-size--tiny` and there was no visible change.

pbking:
With the latest editor changes I see no differences and think this is good to bring in and mark as done.

PR #3503, opened 2021-03-22
Blank Canvas Blocks: Add table styles

scruffian:
This adds CSS to match the frontend to the editor for tables

#### Related issue(s):
https://github.com/Automattic/themes/issues/3413

Editor:
Screenshot 2021-03-22 at 15 39 56

Frontend:
Screenshot 2021-03-22 at 15 40 02

### DIFF
--- a/blank-canvas-blocks/experimental-theme.json
+++ b/blank-canvas-blocks/experimental-theme.json
@@ -224,6 +224,11 @@
 							"hoverBackground": "var(--wp--custom--color--background)"
 						}
 					}
+				},
+				"table": {
+					"typography": {
+						"fontSize": "var(--wp--preset--font-size--tiny)"
+					}
 				}
 			}
 		}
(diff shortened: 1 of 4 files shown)


blank-canvas-blocks/experimental-theme.json
```diff
@@ -224,6 +224,11 @@
 							"hoverBackground": "var(--wp--custom--color--background)"
 						}
 					}
+				},
+				"table": {
+					"typography": {
+						"fontSize": "var(--wp--preset--font-size--tiny)"
```
jffng: Can this declaration live directly under "styles" -> "core/table" now that we've decoupled block supports and theme.json for supported style attributes?

scruffian: No I don't think so because in this case it is being mapped to the caption, not the table itself. Maybe the format of the theme.json should better represent that?

pbking: Yea, images and such keep that in a sub-object.  Makes sense to do that here as well I think.

scruffian: Good call. Done in 6b8b6dd1

jffng (approved):
LGTM

PR #3505, opened 2021-03-22
Blank Canvas Blocks: Use Gutenberg breakpoints.

### LINKED ISSUE #3502: Blank Canvas Blocks: Use Gutenberg's media queries
To make it easier for themes to match the editor we should reuse the media queries from Gutenberg.

scruffian:
This updates Blank Canvas Blocks to use the breakpoints that come from Gutenberg rather than a custom one. This means it will be easier to match the frontend to the editor.

#### Related issue(s):
#3502


blank-canvas-blocks/sass/blocks/_navigation.scss
```diff
@@ -54,7 +55,7 @@
 	}
 }
 
-@include media(mobile-only) {
+@media (max-width: #{ ($break-mobile) }) {
```
MaggieCabrera: Should this be a mixin like on line 48? If Gutenberg doesn't have it, maybe we want to make one using this variable?

scruffian: Gutenberg doesn't have a mixin for this. IMO the important thing it to be using the same breakpoints. Happy to add a mixin if you see value in it...

MaggieCabrera: I'm just thinking in case we need to use it multiple times, just for consistency and to type a bit less. So far we use it just once, but that may change

pbking: Yea, I'm there too.  My 'druthers would be to keep the mixins we already have and adjusting them to leverage Gutenberg's values (though renaming them to less silly names sounds nice while we have the chance).

scruffian: I don't see the value in keeping our own mixing when we can use Gutenberg's ones. If you look at the diff for this change, I'm not convinced that the mixin adds much, but if we do add one we should only do it for the cases that Gutenberg doesn't cover.

pbking:
I didn't like the idea of importing external dependencies anywhere we might want to use a Media Query and instead keep those external dependencies in a single place (_breakpoints.scss).  Gutenberg provides the values that BCB leverages in the media-query mixin.

I too wouldn't want to provide/leverage any breakpoints that weren't derived from Gutenberg's breakpoint values.  

Really though it's only one location, isn't likely to be many more and don't feel very strongly about it; just a 'druthers.

blank-canvas-blocks/sass/blocks/_navigation.scss
pbking: I didn't like the idea of importing external dependencies anywhere we might want to use a Media Query and instead keep those external dependencies in a single place (_breakpoints.scss). Gutenberg provides the values that BCB leverages in the media-query mixin.

I too wouldn't want to provide/leverage any breakpoints that weren't derived from Gutenberg's breakpoint values.

Really though it's only one location, isn't likely to be many more and don't feel very strongly about it; just a 'druthers.

scruffian:
I think i'd rather be explicit with the breakpoints and import them each time, rather than loading them invisibly.

blank-canvas-blocks/sass/blocks/_navigation.scss
pbking: Because this uses the same pixel value as the other anything EXACTLY at 480px is likely to behave strangely.  Can we -1 the 'mobile only' (or +1 the other)

scruffian: Yeah we should do that. That's a good reason for adding our own mixin too.

scruffian:
I added a mobile only breakpoint. This is ready for another review.

jffng (approved):
This LGTM. The only thing that gave me pause was the name of the mixin ("break-mobile") but I don't have a strong opinion, e.g. "media-mobile" + "media-mobile-only".

blank-canvas-blocks/assets/ponyfill.css
```diff
@@ -267,14 +298,14 @@ ol {
 	padding: var(--wp--custom--navigation--submenu--padding);
 }
 
-@media only screen and (min-width: 482px) {
+@media (min-width: 480px) {
```
danieldudzic, line 301: Why was `only screen` removed? I'm assuming that's what core is doing, but I'm curious about the reason.

scruffian: I don't know the reason either, this is just following the way Gutenberg does it :)

scruffian:
> the name of the mixin ("break-mobile")

This is the name in Gutenberg so we're just following that...

blank-canvas-blocks/sass/blocks/_navigation.scss
```diff
@@ -45,7 +43,7 @@
 	}
 }
 
-@include media(mobile) {
+@include break-mobile {
```
pbking, line 46: ~~I don't see a break-mobile mixin, only a break-mobile-only.~~
NM.  :/ That's a mixin provided by Gutenberg.

blank-canvas-blocks/assets/ponyfill.css
pbking: I think it would only matter if we wanted to do something specific for "screen" instead of "print" or perhaps for screen readers.  Probably gone not because it needs to be gone but because it doesn't need to be there. ¯\_(ツ)_/¯

PR #3510, opened 2021-03-22
Experiment to render 'default css variables' for a child theme

### LINKED ISSUE #3420: [Blank Canvas Blocks] Evaluate 'base' CSS variable generation
Blank Canvas Blocks has expectations of the availability of particular CSS variables generated from the "custom" block of theme.json.  If those variables aren't defined and available then the page won't be correct.

Until we have "theme.json castcading" any theme that leverages Blank Canvas Blocks as a parent will either:
* BE REQUIRED to express all variables in the parent theme.json (including any new ones)
* need to have all default css variables defined by default which is also loaded with the polyfill CSS.

This task is to explore the second option.  Hopefully as something that could happen at Blank Canvas Blocks' "build" cycle.

pbking:
An experiment to discern if/how a child theme of BCB can make use of the parent theme's defined "custom" values (without defining them itself).

This change renders the 'custom' css variables exactly how Gutenberg/Global Styles will and enqueues them BEFORE the Gutenberg rendered Global Styles CSS variables are rendered.

In this way all of the "custom" variables are guaranteed to be defined even if not explicitly set in the child theme's theme.json file.  But if they ARE provided then those generated by Gutenberg will take priority.

This utility will no longer be necessary when a CHILD theme is able to extend a PARENT theme's theme.json file.

To evaluate:
Execute the new script (`npm run build:defaultVariables` within the BCB project, or `npm run build` to build it along with the SCSS OR `npm start` do do the same and rebuild on changes).  Note that the CSS variables in /assets/generatedDefaultVariables.css matches what is expressed in the BCB theme.json file's 'custom' block)

Load BCB and note that this CSS file AND the inline CSS files generated by Global Styles matches.

Activate the "Seedlet Blocks" theme and load a view page.

Note that ALL of the variables in "custom" are provided and available for the SCSS to use, but that the Seedlet-blocks theme.json file ONLY has variables for the colors.  Note that the 'secondary' color maps to what Seedlet-blocks expresses (green) and NOT what BCB expresses (blue).

I don't necessarily expect to keep the "Seedlet-blocks" changes in this change, however a child theme was necessary to test this behavior.  Likely it should be removed before we bring this change in (and advanced separately).


scruffian:
My concern with this approach is that it still doesn't help with things defined in the `styles` area. I took a stab at adding support to Gutenberg: https://github.com/WordPress/gutenberg/pull/30147

pbking:
I wholeheartedly agree, and the styles bit was something I was thinking through last night.  I came up with an EVEN HACKIER solution that I like even less.  The gutenberg path is (IMO) the only path forward.

I'm glad to have evaluated this so that I better understand that we shouldn't be doing it this way. :P

pbking:
Closing in lieu of WordPress/gutenberg#30147

PR #3511, opened 2021-03-22
Block themes - update templateParts json

Addison-Stavlo:
json schema for template parts and their area terms was recently updated in https://github.com/WordPress/gutenberg/pull/29828.  Here, we update this for seedlet-blocks as well as add this json for mayland-blocks.  This allows template parts defined by the theme to render under their corresponding block variations by default.

For example, when inspecting the header for mayland-blocks, we can see the block has the expected 'header' variation by presence of the block variations icon throughout the interface:

Screen Shot 2021-03-22 at 5 22 40 PM

### DIFF
--- a/mayland-blocks/experimental-theme.json
+++ b/mayland-blocks/experimental-theme.json
@@ -1,4 +1,14 @@
 {
+	"templateParts": [
+		{
+			"name": "header",
+			"area": "header"
+		},
+		{
+			"name": "footer",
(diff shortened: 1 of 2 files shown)


danieldudzic (approved):
Tested this for both Mayland Blocks and Seedlet Blocks and it works fine 👍 

test

pbking:
c-D59303

PR #3513, opened 2021-03-23
Adjusted theme.json to match expected shape

pbking:
Seedlet Block's theme.json was still expressed with the old structure.  This change brings it up-to-date.

In some recent versions of Gutenberg (doesn't appear to be happening in /trunk) this invalid structure may be causing a WSOD.

### DIFF
--- a/seedlet-blocks/experimental-theme.json
+++ b/seedlet-blocks/experimental-theme.json
@@ -7,8 +7,8 @@
 			"area": "footer"
 		}
 	},
-	"global": {
-		"settings": {
+	"settings": {
+		"defaults": {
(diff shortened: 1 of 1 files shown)


jffng (approved):
This LGTM, although I wasn't able to replicate the WSOD in my quick test.

pbking:
c-D59180 for the SVN DIFF

pbking:
I was only able to replicate on /trunk of Gutenberg yesterday.  The released version seemed fine and the current version seems fine so I think there was something in there that failed in that situation that has since been addressed.  Regardless it was incorrect and a quick fix.

Issue #3415, opened 2021-03-08
[Blank Canvas Blocks] Add block patterns


jffng:
Evaluate each of the block patterns available in Blank Canvas and determine which can be made available in Blank Canvas Blocks without adding any additional custom CSS to the theme.

kjellr:
> ⚠️ EDIT: I left more up-to-date comments in the PR: https://github.com/Automattic/themes/pull/3514#issuecomment-805844014. 

I took a look at the existing Blank Canvas patterns, and they're looking way off at this point: 

A few obvious bugs: 

- Full-width doesn't go full width, there are other content-width errors throughout.
- The page title is present. (Should I be switching to a different template here? Not sure how that is supposed to work)
- The theme colors used in the original theme don't match up with those in BCB. This makes some text unreadable, and changes a bunch of the button colors. (Might make sense to hardcode the colors in the patterns instead?)
- The spacing between blocks is different or wrong. 
- Font weights don't match. 

Here's [the test content](https://gist.github.com/kjellr/9bacbfda03531d2e2e570dc471906ec8) taken right from the original Blank Canvas. 

I'm going to hold off on testing again until this is in a closer state, but please let me know if you have any design questions about that stuff. 

Blank Canvas|Blank Canvas Blocks
---|---
![dotcomthemes test__p=1040](https://user-images.githubusercontent.com/1202812/112319697-b24e9c00-8c84-11eb-8944-adfb2b787e6d.png)|![dotcomthemes test__p=1040](https://user-images.githubusercontent.com/1202812/112319708-b4b0f600-8c84-11eb-94a4-ae0347a88f25.png)

MaggieCabrera:
Closed via #3514

PR #3514, opened 2021-03-23
Blank Canvas Blocks: Added Block Patterns

MaggieCabrera:

This PR adds block patterns to BCB. I have only tested the patterns that load on .org and I had to make adjustments due to the change in color naming conventions.

#### Related issue(s):

Closes https://github.com/Automattic/themes/issues/3415


blank-canvas-blocks/functions.php
```diff
@@ -73,3 +73,8 @@ function blank_canvas_blocks_fonts_url() {
 
 	return esc_url_raw( $fonts_url );
 }
+
+/**
+ * Block Patterns.
+ */
+require get_stylesheet_directory() . '/inc/block-patterns.php';
```
scruffian: Will this break in child themes?

MaggieCabrera: Good call!

scruffian:
There are a couple of patterns that don't work for me:

Screenshot 2021-03-23 at 16 33 48

I do often have problems with cover blocks though, so this might just be me!

MaggieCabrera:
> There are a couple of patterns that don't work for me:
> 
> Screenshot 2021-03-23 at 16 33 48
> 
> I do often have problems with cover blocks though, so this might just be me!

Nope! that happens when the block pattern is calling a color (background, primary) that doesn't exist on the current theme and it's asking you to choose a new one. I think you are using a plugin that I'm not cause I don't see those, I'll look for that.

blank-canvas-blocks/inc/block-patterns.php
```diff
@@ -0,0 +1,503 @@
...
+									<div class="wp-block-cover has-white-background-color has-background-dim" style="min-height:95vh"><div class="wp-block-cover__inner-container"><!-- wp:image {"align":"center","id":624,"width":200,"height":200,"sizeSlug":"large","linkDestination":"none","className":"is-style-rounded"} -->
+									<div class="wp-block-image is-style-rounded"><figure class="aligncenter size-large is-resized"><img src="' . esc_url( get_stylesheet_directory_uri() . '/assets/images/pattern-woman.jpg' ) . '" alt="' . esc_attr__( 'Photo of a woman.', 'blank-canvas-blocks' ) . '" class="wp-image-624" width="200" height="200"/></figure></div>
+									<!-- /wp:image -->
+
+									<!-- wp:heading {"textAlign":"center","level":1,"textColor":"almost-black"} -->
+									<h1 class="has-text-align-center has-almost-black-color has-text-color"><strong>'. esc_html__( 'Jasmine Baker', 'blank-canvas-blocks' ) . '</strong></h1>
```
pbking: Since there is no guarantee that 'almost-black' is going to be available in any child themes (or any theme once switched) I don't believe `has-almost-black-color` (or any of the descriptive color values) should be used here.

scruffian:
>  I think you are using a plugin that I'm not cause I don't see those, I'll look for that.

The only plugin I have is Gutenberg

blank-canvas-blocks/inc/block-patterns.php
MaggieCabrera: I think we could use just the hex color but this is not a universal block pattern, is it? according to our slack conversation, child themes are going to unregister these.

blank-canvas-blocks/inc/block-patterns.php
```diff
@@ -0,0 +1,503 @@
...
+				)
+			);
+
+			register_block_pattern(
+				'blank-canvas-blocks/links-gradient',
+				array(
+					'title'      => __( 'Gradient Links', 'blank-canvas-blocks' ),
+					'categories' => array( 'link-in-bio' ),
+					'content'    => '<!-- wp:cover {"overlayColor":"background","minHeight":1090,"minHeightUnit":"px","align":"full"} -->
+									<div class="wp-block-cover alignfull has-background-background-color has-background-dim" style="min-height:1090px"><div class="wp-block-cover__inner-container"><!-- wp:image {"align":"center","id":130,"width":96,"height":96,"sizeSlug":"large","linkDestination":"none","className":"is-style-rounded"} -->
```
pbking: `has-background-background-color` would be super if we had semantic colors define-able in our themes.  But alas the "background color" is only available as a CSS variable since it's expressed in the "custom" block of theme.json.  This color should probably be expressed another way.

blank-canvas-blocks/inc/block-patterns.php
```diff
@@ -0,0 +1,503 @@
...
+				)
+			);
+
+			register_block_pattern(
+				'blank-canvas-blocks/links-musician',
+				array(
+					'title'      => __( 'Musician Links', 'blank-canvas-blocks' ),
+					'categories' => array( 'link-in-bio' ),
+					'content'    => '<!-- wp:cover {"url":"' . esc_url( get_stylesheet_directory_uri() . '/assets/images/pattern-links-gradient.jpg' ) . '","id":181,"hasParallax":true,"dimRatio":0,"overlayColor":"primary","minHeight":100,"minHeightUnit":"vh","align":"full"} -->
+									<div class="wp-block-cover alignfull has-primary-background-color has-parallax" style="background-image:url(' . esc_url( get_stylesheet_directory_uri() . '/assets/images/pattern-links-gradient.jpg' ) . ');min-height:100vh"><div class="wp-block-cover__inner-container"><!-- wp:spacer {"height":10} -->
```
pbking: There's no `has-primary-background-color` available to be used here.

blank-canvas-blocks/inc/block-patterns.php
pbking: I don't understand how all block patterns aren't "universal".  If on theme A I use a block pattern, then change themes the blocks are still there but they they (potentially) look bad because classes that were expected are no longer there.

MaggieCabrera: yeah, that's the problem if we don't use hex codes or until we get semantic colors that are available to all themes...

scruffian:
I think this is ready for another test

pbking:
Pushed a [minor change](https://github.com/Automattic/themes/pull/3514/commits/fe1a3b762084e467dad1662a59f2bca311d41029) to alignment as it related to containers that aren't groups that this branch illuminated.

image

kjellr:
Whoops — I posted some original notes over in https://github.com/Automattic/themes/issues/3415#issuecomment-805834259 before I realized this PR was here. This definitely improves upon that but I still have a bunch of notes: 

- There are content-width errors throughout.
- The page title is present. (Should I be switching to a different template here? Not sure how that is supposed to work)
- The spacing between blocks is very different than in the original theme. Can we adjust that?
- The heading font weights don't match. 
- The separator blocks don't match the original theme.
- The "Antonio Miller" heading is not readable. 
- The "Patricia Jones" logo image is missing for me. 
- The "Patricia Jones" link underlines are different, but I think that's ok. 

Blank Canvas|Blank Canvas Blocks
---|---
dotcomthemes test__p=1040|dotcomthemes test__p=1043

blank-canvas-blocks/inc/block-patterns.php
kjellr: Yeah, I think for now we should use hex colors in all of the Blank Canvas patterns to combat this.

jffng:
We made a number of changes to bring these patterns more in line with their designs:

- Adjusted button hover + text colors
- Adjusted font weights for headings — removing the "strong" tag / bold treatment from the pattern and using the custom font weight we are declaring for all headings.
- Adjusted spacing variable to use the same value as Seedlet (30px) and applied it to paragraphs and heading elements.
- Fixed width errors occurring in button groups by wrapping them inside a group block and inheriting the default layout.
- Applied `has-black-color has-text-color` classes to Antonio Miller heading
- Fixed Patricia Jones logo image
- Adjusted separators

> The page title is present. (Should I be switching to a different template here? Not sure how that is supposed to work)

My assumption is FSE handles this.

blank-canvas-blocks/sass/base/_normalize.scss
```diff
@@ -16,6 +16,10 @@ img {
 	max-width: 100%;
 }
 
+hr {
+	border: none; // Undo the wp-admin common.css: https://github.com/WordPress/WordPress/blob/fc6e4cbe3ef1dcd3a9977272d7ddcc413acdde0d/wp-admin/css/common.css#L865-L869
```
MaggieCabrera: [This PR ](https://github.com/WordPress/gutenberg/pull/30235)should undo what wp admin defines, but the browser defaults are not `border: none;` so it will be different if/when that PR lands. `add_theme_support( 'wp-block-styles' );` does provide some defaults for the separator, though.

kjellr:
> My assumption is FSE handles this.

In general, if this is a 1:1 reproduction of Blank Canvas, we'll want the default post/page template to have no page title. We should definitely provide an alternate that does though. I see that in general, we need to add more templates, so this can totally be handled separately though. 

I'll put this on my list to re-test today. 👍

blank-canvas-blocks/sass/base/_normalize.scss
MaggieCabrera: That PR just landed, by the way

blank-canvas-blocks/experimental-theme.json
```diff
@@ -413,6 +414,11 @@
 		"core/separator": {
 			"color": {
 				"background": "var(--wp--custom--color--foreground)"
+			},
+			"border": {
+				"color": "currentColor",
+				"style": "solid",
+				"width": "0 0 1px 0"
```
MaggieCabrera: The border definitions work fine, but the background is not doing really anything. It's the color property that's picked up by currentColor on the border. You can test this by changing colors on the editor, or try and set `"background": "var(--wp--custom--color--foreground)" `to red so you can see that it doesn't really work.

kjellr:
These changes look good, and match the existing patterns pretty well. 🙌 Only a few comments:

- We don't necessarily need to adjust the heading weight globally. With these latest changes, `<strong>` [has no effect on the heading block](https://cloudup.com/cHAtvx-XPQZ), which isn't great. I'd recommend setting the Bold headings to be bold in the pattern itself (like the original patterns did), and then just making sure that renders at the same weight it did for the original Blank Canvas. 
- [Button hovers for custom-colored outline buttons are broken](https://cloudup.com/c5Bbj0JQIWY). We should probably adopt @MaggieCabrera's [Seedlet fix](https://github.com/Automattic/themes/pull/3120) for that issue.

pbking:
Because we aren't able to customize what `strong` means without configuration we don't expect from Gutenberg I don't think we should do that in BCB.  So for now that locks us in at (I think?) a font-weight of ~700 for `<strong>`. Which is a tad heavier than the 600 that Blank Canvas achieves but will be our only choice without theme-specific CSS overrides.  (A child-them of this could.)

I suggest we revert the emboldened changes, keeping them a tad bolder than desired but retaining the ability to embolden headers with rich text controls.

Weights look like this now
image

pbking:
image

kjellr:
Do we need a Gutenberg issue for allowing global styles to specify a font weight for bold? Or is that tracked already somewhere?

pbking:
I don't believe so; I'm of the opinion that doing that in the font's CSS is appropriate.  BCB is in a unique place where we're trying to avoid specifying things that aren't going to be controlled by Gutenberg (so subsequent child themes won't have to work to overcome it).

As far as I know an issue doesn't exist for that.  If that seems like a thing that should be user or theme configurable - outside of CSS - I wouldn't argue against it; just seems abstract enough a thing to expect to be in a theme's limited specific CSS.

pbking:
Regarding the button hover state, since the colors are applied with style properties there's no CSS solution.  I don't believe the fix you reference is quite the same situation.   I'm afraid that using that solution to style the buttons results in hover-less states.

I would certainly be keen to address that problem in some way but I'm not sure how.  It should probably [have something to do with the button block directly](https://github.com/WordPress/gutenberg/issues/4543) though.

kjellr:
> Regarding the button hover state, since the colors are applied with style properties there's no CSS solution. I don't believe the fix you reference is quite the same situation. I'm afraid that using that solution to style the buttons results in hover-less states.
> I would certainly be keen to address that problem in some way but I'm not sure how. It should probably have something to do with the button block directly though.

We definitely fixed this for the normal version of Blank Canvas blocks — maybe it was in [a followup PR actually](https://github.com/Automattic/themes/issues/3110#issuecomment-769094256)? Or it could've been somewhere else. 🤔 

In any case, we should make sure it's working the same way for Blank Canvas blocks, since I don't expect the hover states to be sorted out in Gutenberg for quite some time. The hover states were a blocking issue for launching the patterns back when we initially launched them, and I consider it a blocking issue for BCB too. I've opened https://github.com/Automattic/themes/issues/3538 to track it.

kjellr:
I've also opened a Gutenberg issue for the bold font weight setting: https://github.com/WordPress/gutenberg/issues/30297

PR #3522, opened 2021-03-24
Work on updating the default templates to better match Blank Canvas

pbking:
This change adds some default templates that bring the theme better in light with Blank Canvas.

This now leverages THREE template parts: Header, Footer and Navigation.  Since Template Parts can't (yet) include other template parts (and so that changes to the navigation block don't create an entirely new 'header' template part) the navigation resides in its own template part.

This includes minor changes to index.html and the addition of `single.html`, `page.html` (both of which are the same but different from index) `search.html` and `404.html` (both of which have very simple search/error content).

To mirror Blank Canvas, none of the content pages include the Title block.

404:
image

Index:
image

Post (single):
image

Page:
image

Search:
image


blank-canvas-blocks/block-templates/page.html
```diff
@@ -0,0 +1,6 @@
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
```
jffng: My assumption is that the default post (`single`) + `page` templates default do not have a header or footer. Should we remove the header, nav, and footer template parts from these?

pbking: I would think if any would/wouldn't all wouldn't.  I'm impartial as to which the default is.

I think if there isn't a template part defined then it also wouldn't define that 'zone' (has another name) so perhaps there should be an "empty" template part that is the default in those spaces so that the spaces are defined?  
Does that make sense to do? That's my only argument of having SOMEthing there.

MaggieCabrera: I think the default for this theme is not to have them. As the default for Seedlet would be to have them, being the main difference between them (besides font families and color configurations, etc).

The way I was originally thinking about this, what made sense to me is to have 2 different versions of the single/page template, one with the header and footer and one without, as you may want to make use of both on your site depending on the case. And that will also be the case for the children, where you will want to have the option of no header/footer too, but the defaults will be reversed.

blank-canvas-blocks/experimental-theme.json
```diff
@@ -3,6 +3,10 @@
 		{
 			"name": "header",
 			"area": "header"
+		},
+		{
+			"name": "footer",
+			"area": "footer"
```
MaggieCabrera, line 9: do we need one for the navigation or is that not a thing?

pbking: TY.  I believe we do, yes.

PR #3535, opened 2021-03-26
Mayland Blocks: Preserve image ratios

### LINKED ISSUE #3382: Mayland Blocks: Distorted images
Images inserted via Image block and Media & Text are distorted in preview (but appear correctly in the editor).

## Steps to replicate
1. Create a site with Full Site Editing and Mayland Blocks active.
2. Open the site editor and add an Image block into the post content (upload new, select from the media library, or from Pexels).
3. Save and preview the site in a new tab. Note the image is distorted.

https://cloudup.com/c-OFkzuESqb

Found through my own testing.

kjellr:
Fixes https://github.com/Automattic/themes/issues/3382.

Adds some temporary CSS to preserve correct image ratios for wide/full/full-size images. Same as #3450, but for Mayland Blocks. 

@ianstewart noted that this issue is blocking testing on WP.com, so I recommend we add this in temporarily rather than wait for [a permanent solution in Gutenberg](https://href.li/?https://github.com/WordPress/gutenberg/pull/30092).

---

Before|After
---|---
Screen Shot 2021-03-26 at 8 29 21 AM|Screen Shot 2021-03-26 at 8 29 33 AM

### DIFF
--- a/mayland-blocks/style.css
+++ b/mayland-blocks/style.css
@@ -65,3 +65,12 @@ body {
 h1, h2, h3 {
 	letter-spacing: -0.015em;
 }
+
+/* 
+ * Preserve image ratios.
+ * Needed until https://github.com/WordPress/gutenberg/pull/27518/ is merged. 
+ */
+img {
+	height: auto;
+	max-width: 100%;
+}


danieldudzic (approved):
This works as expected 👌

PR #3536, opened 2021-03-26
Blank Canvas Blocks: Add a build tool

scruffian:
This adds a simple build tool which combines the experimental-theme.json file from Blank Canvas Blocks with the `child-theme.json` file in a child theme. This will make it easier to build child themes when the parent theme needs to change.

To test, install this child theme: 
[child.zip](https://github.com/Automattic/themes/files/6212357/child.zip)

Delete the experimental-theme.json file from the child theme.
Then run `npm run build-child child` from the blank-canvas-blocks directory.
Check that the child theme has an experimental-theme.json file.


jffng:
I think it would be nice if the script had a watch option for development, so it recompiled when we make a change to the child's theme.json.

Also maybe we could agree to just name it `experimental-child-theme.json`, the script looks for changes to that, and recompiles automatically.

EDIT
Maybe this is something we can add via npm scripts when we're happy with the build script.

scruffian:
I updated this to use a js script, and watch the file. Also updated the child theme json file name, and added a new zip...

MaggieCabrera:
A couple of things we saw with the build tool:
- It would be helpful for the script to run `npm start` on bcb at the same time that it's working
- If we add a new css variable to bcb's theme.json, they don't get built into the child's

scruffian:
I'll improve this in a follow up PR

PR #3539, opened 2021-03-26
Update/ponyfill to styles

scruffian:
Quick rename for the stylesheet for consistency with other themes.


pbking:
https://github.com/Automattic/themes/blob/f1698a833e810016df1a2b7b208c6bcfb25f899a/blank-canvas-blocks/package.json#L35

That will need some love too I think.

scruffian:
I merged the ponyfill with the existing style.css.

pbking:
It doesn't appear to merge, but replace losing the meta info.
I didn't initially realize that was the .css file you were targeting.  
Because that's the source of meta, and so that we DON'T have to keep that in 2 places (styles.css/styles.scss) (as we do with other themes) which can become out-of-sync I would much prefer to target a separate file unless there is a technical reason not to.  I don't much care what it's called, but I was hoping it wouldn't be /style.css

jffng (commented):
Should this PR also remove the `ponyfill.css` file? I don't see that deletion...

scruffian:
It should if we go with it. We seem divided :)

scruffian:
We decided not to do this.

Issue #3546, opened 2021-03-29
Blockbase: Add default styles for comments block.


MaggieCabrera:
We need to add default styles for the Post comments and Post comments form blocks for BCB and children:

Logged in:

<img width="729" alt="Screenshot 2021-03-29 at 17 23 05" src="https://user-images.githubusercontent.com/3593343/112859851-86338080-90b3-11eb-903c-a267f72a246d.png">

Logged out:

<img width="725" alt="Screenshot 2021-03-29 at 17 24 26" src="https://user-images.githubusercontent.com/3593343/112859959-a8c59980-90b3-11eb-80b7-ba15f4c3a5a8.png">

jffng:
Partially addressed in #3553, also depends on https://github.com/WordPress/gutenberg/pull/30382

scruffian:
This is unblocked now

jffng:
Is there a design target for this now that the above PRs were merged? cc @kjellr 

Current state:
<img width="729" alt="Screen Shot 2021-05-27 at 11 35 30 AM" src="https://user-images.githubusercontent.com/5375500/119855332-a9698900-bec6-11eb-9219-e9031cb45fbe.png">

kjellr:
Not yet. In general, I think the sorts of font size changes we made in Quadrat should be moved back here, but I'll take a closer look.

PR #3553, opened 2021-03-29
BCB: Add post comments styles

jffng:
Starts to address #3546.

Which of these styles do you think should live with Gutenberg? 

Styling in Gutenberg a bit tricky because the post comments block relies on the Core comments template: https://github.com/WordPress/gutenberg/blob/9b75a5c44e934f580282e2d16cc462ff4567cded/packages/block-library/src/post-comments/index.php#L33

Before | After
--------- | ---------
Screen Shot 2021-03-29 at 1 00 16 PM | Screen Shot 2021-03-29 at 12 56 25 PM


danieldudzic:
Here's the Gutenberg issue: https://github.com/WordPress/gutenberg/issues/26864

I have added the button styling to the Reply link:
Comments_–_themes

scruffian:
Moved this to Gutenberg too: https://github.com/WordPress/gutenberg/issues/26864

jffng:
I removed some styles because of the PR to add some styles in Gutenberg.

danieldudzic:
@jffng Shouldn't this be merged with https://github.com/Automattic/themes/tree/make/blank-canvas-blocks rather than trunk?

jffng:
I believe since we merged BCB to trunk, any changes to BCB can be made directly to trunk / synced to svn as soon as they're merged. `make/blank-canvas-blocks` is now a development branch for the new versions of Seedlet Blocks and Mayland Blocks.

PR #3555, opened 2021-03-29
Add/seedlet bcb scss

pbking:
This change styles the Site Title in the same way that Seedlet did.

To do so a SCSS build process has been introduced.

The existing STYLES from 'style.scss' were moved to the '/scss/index.scss' file (though should probably relocate).  The METADATA was moved to /scss/metadata.scss.

My 'druthers is to render this to /assets/theme.css (and leave ./style.css for just metadata) but did it this way for more likely acceptance. 

This also makes the --watch functionality of BCB's child theme building optional and calls it from Seedlet-Blocks.

Before:
image

After: 
image


scruffian:
I'm going to take these build script changes back into the make/blank-canvas-blocks branch.

pbking:
I think all of the important bits have been moved over to the appropriate places.  Closing this up.

Issue #3550, opened 2021-03-29
Mayland Blocks: Remove alignments css


MaggieCabrera:
The alignment.css file is no longer needed, we need to handle this now via the parent theme.

MaggieCabrera:
Closed via #3558

PR #3558, opened 2021-03-30
Mayland Blocks: Layout and alignments

MaggieCabrera:

This PR defines the layout for Mayland blocks and removes the previously existing alignment CSS rules

#### Related issue(s):

Closes https://github.com/Automattic/themes/issues/3550


mayland-blocks/block-templates/page.html
```diff
@@ -1,11 +1,17 @@
...
 <div style="height:32px" aria-hidden="true" class="wp-block-spacer"></div>
 <!-- /wp:spacer -->
 
 <!-- wp:post-title {"level":1} /-->
 
-<!-- wp:post-content /-->
+</div>
+<!-- /wp:group -->
+
+<!-- wp:post-content {"layout":{"inherit":true}} /-->
```
MaggieCabrera, line 15: If I don't do this, full width alignment doesn't work inside what's on post-content. I don't know if I'm missing something here.

jffng: You mean adding layout -> inherit -> true? I think this is the correct way to declare alignment for the post content.

MaggieCabrera: No, I mean I had to add a group before and after the post-content block to have all 3 with the inherited layout and having the post-content also working with the full width alignments. If I only add one group block that contains all of it (post title, post content, comments etc), the blocks inside post-content are not direct children of the group that has the layout rules and they never go full wide.

mayland-blocks/block-templates/single.html
```diff
@@ -12,7 +14,13 @@
...
+</div>
+<!-- /wp:group -->
+
+<!-- wp:post-content {"layout":{"inherit":true}} /-->
+
+<!-- wp:group {"layout":{"inherit":true}} -->
+<div class="wp-block-group">
 
 <!-- wp:spacer {"height":16px} -->
 <div style="height:16px" aria-hidden="true" class="wp-block-spacer"></div>
```
jffng, line 26: Could be addressed separately, but are spacer blocks the right way to go for adding this kind of margin in the template? They are nice because they give users control in a FSE context, but I also saw some discussion about this yesterday in slack that suggested we use margin/padding instead.

jffng:
In general this LGTM. What about `front-page.html`? Should that one be removed or amended to use the new alignments system also?

MaggieCabrera:
> In general this LGTM. What about `front-page.html`? Should that one be removed or amended to use the new alignments system also?

The front page for Mayland is actually all full width so I didn't add it there because of that

mayland-blocks/block-templates/page.html
jffng: Ah I see what you mean. Yeah I still think it's correct even though it requires us to add wrapping containers / group blocks all over the place.

You've probably seen it but there is some related discussion on why this is the case [here](https://github.com/WordPress/gutenberg/issues/29983)

MaggieCabrera: Oh I missed that, thank you! then I guess it's fine if we keep doing this in this way, it's not so terrible after all

mayland-blocks/block-templates/single.html
MaggieCabrera: Yeah, and with the coming margins to the group block that might be even better. I will address it on a separate PR since it's not alignment related, thanks!

PR #3576, opened 2021-03-30
Fix seedlet-bcb site title

pbking:
Extends #3575 

Adds site title block styles from Seedlet.

Before:
image

After:
image


jffng (commented):
In general this LGTM!

I just have a small comment about the file naming. We have:

- theme.css
- style.css for metadata only
- index.scss (compiles to theme.css)

Could we just have theme.scss or index.scss? What is the reason for naming them differently?

pbking:
Maybe it's a purple elephant. ;)

theme.css/theme.scss sounds super to me.

pbking:
Refactored naming per @jffng suggestion [here](https://github.com/Automattic/themes/pull/3575/commits/d8973dc3093763f659383b387d6dfae9e606db1b).

PR #3579, opened 2021-03-30
Refactored build scripts to standardize

pbking:
This change refactors the BCB build scripts to standardize them on the following common commands:

`npm run build`: "Build all the things that need to be built, and just do it once"
`npm start`: "Build the things, whatever they are, then rebuild them when something changes"

Additionally, in order to target a child theme these commands would also be available:

`npm run build:child {CHILDFOLDER}`: "Build the target child's theme.json based on it's child-theme.json and BCB's theme.json and just do it once.
`npm run start:child {CHILDFOLDER}`: "Build the  target child's theme.json, then rebuild when something changes.  ADDITIONALLY also do the same process as `npm start` above (rebuilding BCB css when BCB scss changes).

### DIFF
--- a/blank-canvas-blocks/build.js
+++ b/blank-canvas-blocks/build.js
@@ -20,11 +20,12 @@ const buildChildTheme = () => {
 			console.log( "\x1b[32m" + childThemeName + "/experimental-theme.json created successfully." );
 		} );
 	} catch ( error ) {
-		console.error( "\x1b[41m" + error );
+		console.error( "\x1b[31m" + error );
 	}
 };
 
 if ( ! fs.existsSync( childThemeJsonFileName ) ) {
+	console.log( "\x1b[31m" + childThemeJsonFileName + ' not found :(' );
 	return;
 }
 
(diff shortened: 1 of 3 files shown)


scruffian:
I dont think this will work - the build:child job won't get run...

pbking:
> I dont think this will work - the build:child job won't get run...

I don't believe it should.  I'm proposing that building a child would be [ran from the child theme project](https://github.com/Automattic/themes/pull/3575/files#diff-94e2fffd78e39f192bd53e3354b4a9a37544faadb74b29f59083596edd9d44f8R16).

We'll still need to figure out a solution for "build ALL the children" but I believe that's a different ball of wax.

Since "child theme.json" doesn't have anything to do with building THIS theme I don't believe it belongs in the start/build workflow.

If you are working on BCB then 
* `npm start` will target scss file changes and build css

If you are working on a child that doesn't have a build process (I don't expect any, but ¯\_(ツ)_/¯ )
* `npm run build:child pickles` will build a child theme (useful for child themes that don't have a build process I suppose...)
* `npm run build:child pickles watch` will build the child theme and rebuild it when parent/child .json files change.

If we add a target `npm run start:child pickles` that would do both the job of `npm start` and `npm run build:child pickles watch` would that work for your workflow?

pbking:
Updated with a `start:child` target.  LMK what you think about that.  Will that work for your needs?

scruffian (approved):
LGTM

PR #3580, opened 2021-03-30
Fix/seedlet bcb pullquote

pbking:
Extends #3575 

Changes Seedlet PullQuote block styles to match seedlet.

Minor changes to BCB to allow for configuration of citation font family and quote line heights.

Includes some margin changes for paragraph.

Before:
image

After:
image

### DIFF
--- a/seedlet-blocks/experimental-theme.json
+++ b/seedlet-blocks/experimental-theme.json
@@ -229,7 +229,13 @@
 					"citation": {
 						"typography": {
 							"fontSize": "var(--wp--preset--font-size--tiny)",
+							"fontFamily": "var(--wp--preset--font-family--base)",
 							"fontStyle": "italic"
+						},
+						"spacing": {
+							"margin": {
+								"top": "20px"
+							}
 						}
 					}
 				},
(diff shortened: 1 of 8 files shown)


seedlet-blocks/sass/theme.scss
```diff
@@ -1,6 +1,14 @@
 @import './blocks/site-title.scss';
 @import './blocks/links.scss';
 
+/**
+ * Paragraph Margins
+ */
+
+p {
+	margin-top: 0;
```
jffng: This is pretty generic. Are we sure it's what we want?

For example, if I place a cover block followed by a paragraph: 

Screen Shot 2021-03-31 at 11 26 20 AM

jffng:
The pullquote looks good :)

pbking:
The paragraph as it existed didn't provide the same layout as before, but probably not the pr for it.  I'm still not sure how it should land for SB but I think you've demonstrated this isn't it.  I'll remove the P margins before bringing in.

seedlet-blocks/experimental-theme.json
```diff
@@ -469,19 +475,21 @@
...
-				"fontStyle": "italic"
+				"fontSize": "32px",
+				"fontStyle": "italic",
+				"lineHeight": "1.3",
+				"fontFamily": "var(--wp--preset--font-family--headings)"
 			},
 			"border": {
-				"style": "solid",
+				"style": "none",
 				"width": "1px 0"
```
jffng: Nit but I think this can be removed if style is none:
```suggestion
```

jffng (approved):
LGTM!

PR #3594, opened 2021-03-31
Mayland Blocks: remove theme supports

MaggieCabrera:
This PR removes theme supports that are no longer needed since they are being loaded by the theme's parent.

### DIFF
--- a/mayland-blocks/functions.php
+++ b/mayland-blocks/functions.php
@@ -3,21 +3,6 @@
 if ( ! function_exists( 'mayland_blocks_support' ) ) :
 	function mayland_blocks_support() {
 
-		// Alignwide and alignfull classes in the block editor.
-		add_theme_support( 'align-wide' );
-
-		// Adding support for core block visual styles.
-		add_theme_support( 'wp-block-styles' );
-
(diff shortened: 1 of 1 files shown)


mayland-blocks/functions.php
```diff
@@ -3,21 +3,6 @@
 if ( ! function_exists( 'mayland_blocks_support' ) ) :
 	function mayland_blocks_support() {
 
-		// Alignwide and alignfull classes in the block editor.
-		add_theme_support( 'align-wide' );
-
-		// Adding support for core block visual styles.
-		add_theme_support( 'wp-block-styles' );
```
scruffian, line 10: Are you sure we don't want to keep this one in?

MaggieCabrera: mmmh, Seedlet blocks doesn't have it. Maybe we need it in both themes.

pbking: My take is to be done with wp-block-styles forever.  Anything that support might style seems like it should be in theme.json instead so that it's explicit and understood WHY a block is styled as it is.

PR #3596, opened 2021-03-31
JS Pre commit hook

scruffian:
This adds Prettier to our pre-commit hook for JS files.


pbking:
Worked well for me for .js files.  Would love to see this on .css and .scss files as well.  I think it's a work-able concept.

Tinkered with it a bit to get it to do that using the method you started but to no avail.

PR #3598, opened 2021-03-31
Seedlet Blocks: Update blocks .scss file names to adhere to the SASS …

danieldudzic:
…naming convention for partials


scruffian:
I think we don't need the _ on the import declarations, so I removed it. What do you think?

danieldudzic:
> I think we don't need the _ on the import declarations, so I removed it. What do you think?

That was dumb on my part. You are 100% correct. Thanks for the fix :)

PR #3603, opened 2021-04-01
Blank Canvas Blocks: Remove unnecessary priority parameter from ponyf…

danieldudzic:
…ill css load


pbking:
Not sure where the 11 came from, but removing it doesn't seem to effectively re-order the priority in any way.  LGTM.

PR #3604, opened 2021-04-01
Add/bcb index query

pbking:
> Currently, Blank Canvas Blocks contains no archive page templates at all — the index.html template only shows the post/page content. Are there plans to change that? The original Blank Canvas shows a list of posts in its index template, and only the post/page content on its singular template.

This change:

Adds the site-logo to the header (and centers it).
Adds a singular template (with exclusively content, just like index template was).
Added query loop (including post/page title) to index page template.
Added header/footer template parts to index page.

image

### DIFF
--- a/blank-canvas-blocks/assets/ponyfill.css
+++ b/blank-canvas-blocks/assets/ponyfill.css
@@ -114,6 +114,10 @@ img {
 	overflow: auto;
 }
 
+.aligncenter {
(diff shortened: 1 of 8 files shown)


kjellr (approved):
This looks pretty good and it matches what I was expecting. I adjusted some of the spacing more closely match the original Blank Canvas. 👍

PR #3605, opened 2021-04-01
Remove experimental theme json

scruffian:
We can't use experimental-theme.json files on Classic themes because it opts them into the new alignment styles. This removes these files. As long as you are running the latest Gutenberg experimental link color will work as before...

#### Related issue(s):
https://github.com/WordPress/gutenberg/pull/30452#event-4542045021


pbking (approved):
LGTM
Spearhead appears to be the only theme that requires the Gutenberg change, the rest seem to be handling it OK with their own styles.  Without the Gutenberg fix, though Spearhead links are browser-default-blue if they have been assigned one of the palette colors.

jffng:
Also LGTM across these themes, thanks @scruffian.

scruffian:
We can't merge this until the Gutenberg patch happens on wpcom...

PR #3611, opened 2021-04-02
Adds style linting like we do it in Gutenberg

scruffian:
This adds a stylelint config and also adds stylelint to the pre commit hook so we catch and fix issues in our Sass files.

To test, try committing a change to sass file. You should see warnings about issues in only that file.

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,0 +1,3 @@
+{
+    "extends": [ "@wordpress/stylelint-config/scss" ]
+}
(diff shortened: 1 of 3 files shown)


pbking:
This worked for me and I recommend going forward with bringing it in.

My only suggestion is to remove the `lint-css` and (especially) the `lint-css:fix` scripts.  The fix in-particular, in my opinion, should never ever be ran.  Ensuring everything henceforth is quality sounds great.  Fixing everything in the past means one very large commit of questionable value.

scruffian:
Good call. They were there for testing :)

PR #3620, opened 2021-04-06
Seedlet Blocks: bring up-to-date with GB 10.3

jffng:
This PR cherry picks the work to make Seedlet Blocks a child of BCB, so we can deploy it and iterate separate from Mayland Blocks.

If it looks okay, I'll squash merge and deploy to wpcom.

**Testing instructions**
- Use the horizon link on wordpress.com with gutenboarding + site editor flags to create a new site
- Check out this PR onto your sandbox and sandbox the site you just created
- Activate seedlet blocks using the CLI (`wp theme activate pub/seedlet-blocks`)
- Go to the site editor and verify it looks right


jeffikus (changes requested):
I tested on my sandbox and made a [video of my experience](https://www.dropbox.com/s/t492jo0djzxr51b/Screen%20Recording%202021-04-07%20at%2013.04.24.mov?dl=0).

Takeaways:
- Site Editor is blank.
- Page Editor for homepage doesn’t look right.

The Site Editor has the following console errors: https://cloudup.com/c0Lg4-g52Yc

jffng:
Thanks for testing @jeffikus .

> Site Editor is blank.

On initial load, I also get a blank site editor. But if you navigate to one of the templates using the left hand nav, the templates do load when you select one. So my hunch is this is not a bug with the theme but the site editor. 

Another way around it is if you visit the site editor using a URL with the following query params: `youredgesite.wordpress.com/wp-admin/admin.php?page=gutenberg-edit-site&calypsoify=1&block-editor=1&postId=pub%2Fseedlet-blocks%2F%2Findex&postType=wp_template`

> Page Editor for homepage doesn’t look right.

The theme still appears to be loading parent styles from Seedlet, not BCB. I can't figure out why; it says "Template: blank-canvas-blocks" in the style.css, and I've cleared my cache. But these responsive styles from Seedlet are causing the alignments to appear off center:

Screen Shot 2021-04-07 at 3 00 18 PM

jeffikus (approved):
We tested this again, seems like an issue with .com - happy to ship this and log in issue in Calypso repo.

PR #3621, opened 2021-04-06
Mayland Blocks: fix site editor alignment

jffng:
This PR removes styles that were added before the new alignments system to control the default width of a block.

**To test**
- Check the site editor and see that the blocks are being constrained by these rules
- Apply this PR
- Verify that the default width / alignment is correct in the site and post editor


MaggieCabrera (approved):
LGTM!

PR #3648, opened 2021-04-13
Make: Quadrat

jffng:
This PR is for the active development of the Quadrat theme.

Any changes for the theme's initial development should be merged into this PR.

**To develop**
- `cd quadrat`
- `npm i && npm start`
- Activate this theme on your site, using Gutenberg master if possible (the plugin should be fine)
- Make a change to `child-experimental-theme.json` or a sass file
- The theme should rebuild its json / styles


quadrat/index.php
```diff
@@ -0,0 +1,22 @@
...
+ * @since 1.0.0
+ */
+	// the header
+	echo do_blocks( file_get_contents( 'templates/header.html' ) );
+
+	// the query
+	echo do_blocks( file_get_contents( 'templates/query.html' ) );
+
+	// the footer
+	echo do_blocks( file_get_contents( 'templates/footer.html' ) );
```
MaggieCabrera: This is very smart! it would solve our issues with the FSE being available to users, but so far it doesn't seem to be working on my local install. It seems like the theme is using the index.html template from blank canvas instead of this one. Does that happen to you too?

MaggieCabrera: I went ahead and made a PR to fix that: https://github.com/Automattic/themes/pull/3651

jffng: Thank you for making that fix and opening an issue in GB about the child theme bug!

seedlet-blocks/sass/theme.scss
```diff
@@ -1,6 +1,8 @@
 @import './blocks/site-title.scss';
 @import './blocks/links.scss';
 @import './blocks/latest-posts';
+@import './blocks/post-comments';
```
jffng: I don't know why there are seedlet blocks changes in here, maybe I branched off of something I didn't mean to.

scruffian:
I think we should merge this soon.

PR #3651, opened 2021-04-14
Quadrat: Override BCB's index.html template

MaggieCabrera:
This PR overrides the html template when we are on the home or front page and loads index.php instead. 
When using do_blocks on index.php we need a header and footer templates to be defines or else we won't have css (when using html templates, the missing html from header and closing tags for the footer are provided by gutenberg, this is no longer the case when the template is built with php)

### DIFF
--- a/quadrat/footer.php
+++ b/quadrat/footer.php
@@ -1,0 +1,19 @@
+<?php
+/**
+ * The template for displaying the footer
+ *
+ * Contains the closing of the #content div and all content after.
+ *
+ * @link https://developer.wordpress.org/themes/basics/template-files/#template-partials
+ *
+ * @package WordPress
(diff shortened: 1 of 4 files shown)


MaggieCabrera:
I also changed the query block for the content block since the home page can also be a static page, and that wasn't working with the current template.

danieldudzic (approved):
That template override works fine as a temporary solution. Ideally, this should be core WordPress functionality in the future.

MaggieCabrera:
> That template override works fine as a temporary solution. Ideally, this should be core WordPress functionality in the future.

100% agree, I commented on [this GB issue](https://github.com/WordPress/gutenberg/issues/25612) about that and added a comment on our code, I expect this to not be needed in the future

PR #3654, opened 2021-04-14
Quadrat: use php based header template

jffng:
This PR adds a header template that renders a classic menu. It's un-styled but I'm looking for input on the approach. I tried distilling it down to the minimum boilerplate required to support classic menus.

Screen Shot 2021-04-14 at 11 44 48 AM

The PR also reorganizes the project a bit into `block-template-parts` and `template-parts`. 

I have a feeling this is going to be the ugliest part of the theme, so wanted to tackle it early.


quadrat/template-parts/header.php
```diff
@@ -0,0 +1,63 @@
...
+		'fallback_cb'    => false,
+		'echo'           => false,
+	)
+);
+?>
+
+<header id="masthead" class="site-header" role="banner">
+	<div class="site-branding">
+		<?php if ( has_custom_logo() ) : ?>
+			<div class="site-logo"><?php the_custom_logo(); ?></div>
```
MaggieCabrera: I think we could reduce this file to only use php for mostly just the navigation. We could replace some of the php using blocks like in this case we could simply do `<?php echo do_blocks( '<!-- wp:site-logo {"align":"center","width":128} /-->' ); ?>` I tested locally and that seemed to work fine and was editable via the customizer

jffng:
Note: we prefer #3658 which depends on a Gutenberg change.

quadrat/template-parts/header.php
```diff
@@ -0,0 +1,48 @@
...
+			$location_name = 'primary';
+			$menu_id       = $locations[ $location_name ];
+
+			wp_nav_menu(
+				array(
+					'menu' => $menu_id,
+					'theme_location'  => 'primary',
+					'menu_class'      => 'menu-wrapper',
+					'container_class' => 'primary-menu-container',
+					'items_wrap'      => '<ul id="%1$s" class="%2$s">%3$s</ul>',
```
pbking: Even if we have to use PHP to render this I think it would still be good to use add_theme_support( 'block-nav-menus' ); and render the menu [like this](https://github.com/WordPress/gutenberg/pull/30852/files#diff-90cb5b64c95087ded3c7bb160e38189c8499b0f14f61beb884c887452449fa12R111-R118).  Those nav blocks are already styled in BCB settings.

jffng: Great point, I will refactor this to use block-nav-menus.

jffng:
I updated this so that the theme renders the menu using the nav block, and we pick up some styles from BCB:

Screen Shot 2021-04-20 at 12 25 22 PM

jffng:
Closing! See #3714

PR #3655, opened 2021-04-14
Deploy/quadrat

pbking:
Add staging Deploy Action for Quadrat.

Removes Staging Deploy Action for Seedlet-Blocks and BCB.  (Those deploys won't be stopped by this change until `make/quadrat` hits `/main`)

Any changes to the `make/quadrat` branch will push the contents of `/blank-canvas-blocks` and `/quadrat` to https://quadrat.mystagingwebsite.com/

I had attempted to use a [plugin](https://github.com/afragen/git-updater) that would have polled this repo and pulled updates.  However that plugin expects a single theme per-repository and our project's structure will not (currently) work with that.  (I may submit a patch to that project to better suit our needs but this is what works for today.)


.github/workflows/quadrat_deploy.yml
```diff
@@ -20,15 +20,15 @@ jobs:
...
           ftp-server: sftp://sftp.pressable.com:22/htdocs/wp-content/themes/
-          ftp-username: ${{ secrets.SB_DEMO_USERNAME }}
-          ftp-password: ${{ secrets.SB_DEMO_PASSWORD }}
+          ftp-username: ${{ secrets.QUADRAT_DEMO_USERNAME }}
+          ftp-password: ${{ secrets.QUADRAT_DEMO_PASSWORD }}
           git-ftp-args: --insecure
-          local-dir: ./seedlet-blocks/
+          local-dir: ./quadrat/
       - name: Deploy-Parent
         uses: SamKirkland/FTP-Deploy-Action@3.1.1
```
jffng, line 28: I should have asked this question a long time ago, but does this action delete files that are no longer in the theme? Say if we rename or remove a template?

jffng:
Let's bring this in and my q above can be addressed later if needed, thanks!

.github/workflows/quadrat_deploy.yml
pbking: I poked around the docs and couldn't tell for sure but... I THINK so.
The only way to know for sure is to remove something and see though.  In [v4 there is a "clean target" option](https://github.com/SamKirkland/FTP-Deploy-Action#settings) (which the docs imply shouldn't be necessary) but we're[ using v3 ](https://github.com/SamKirkland/FTP-Deploy-Action/tree/3.1.1)as v4 doesn't work with the host we're uploading to (Pressable).

jffng: I see. Bless Sam Kirkland.

PR #3658, opened 2021-04-14
Remove manual menu data stuff and replace with target menu data slug

pbking:
This is an experiment to see if a "classic menu data source" can be used in the Navigation Block.

[This branch of Gutenberg](https://github.com/pbking/gutenberg/tree/tinker/nav-classic-data-source) (from my fork) is needed to test out that functionality. 

When a navigation block has NO child items it will attempt this alternate rendering.

The change renders a menu using the classic method `wp_nav_menu()` and modifies as much as possible using the provided tools to make the result appear like the navigation block.  The remainder is done with DOM/string manipulation.  

image

#### Thoughts 
Ideally the Navigation block is able to target this data source and use that as the source of menu options instead of IMPORTING that data and storing it on the block.  

The block could be given a control in the block editor so that a classic "source" can be chosen to be rendered dynamically OR child 'nav' blocks could be added. 

Even better would be the ability for the navigation block to allow for editing of this data and persist it (as data instead of as blocks).

EVEN IF the ability to target a menu location is ONLY available as a hand-coded attribute in a block-template this still provides value as a tool themes can use during the transitional time.

ALTERNATIVELY rather than adding this (handy dandy) functionality to the Navigation block another "Classic Navigation" block could be created where that is the only purpose (rendering classic navigation data) which could just be swapped out for a regular Navigation block at the appropriate time.


quadrat/templates/header.html
```diff
@@ -8,20 +8,10 @@
 <!-- /wp:column -->
 
 <!-- wp:column {"verticalAlignment":"center"} -->
-<div class="wp-block-column is-vertically-aligned-center"><!-- wp:navigation {"orientation":"horizontal","itemsJustification":"right"} -->
-<!-- wp:navigation-link {"label":"Home","url":"#"} /-->
+<div class="wp-block-column is-vertically-aligned-center">
+	
+<!-- wp:navigation {"orientation":"horizontal","itemsJustification":"right","source":"primary"} -->
```
jffng, line 13: I love this API of identifying a nav area to render via a block attribute.

MaggieCabrera: +10000

jffng:
I gave this another spin and it worked swimmingly when I had a navigation menu defined and set to display in the "Primary" nav area. 

I tried to de-activate the menu to see what would happen, and got caught in a loop: 

Screen Shot 2021-04-15 at 6 05 09 PM

Not sure what may be causing this.

pbking:
The PR has triggered a lot of conversation but the branch has a lot of extra baggage.  Closing this, though some of the ideas may live on...

Issue #3640, opened 2021-04-12
Quadrat: Implement typography & color 


jffng:
Design here: https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/?node-id=50%3A766

Final color, typography and button styles: https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/Quadrat?node-id=760%3A112

- [x] Headings
- [x] Paragraphs
- [x] Blockquotes
- [x] Lists 
- [x] Tables
- [x] Code
- [x] Buttons

MaggieCabrera:
We may want to wait https://github.com/WordPress/gutenberg/pull/30791 out (I'm not sure if it's going to be ready soon or not) for the tables.

jffng:
I think just the button styles need to be audited and then we can close this?

scruffian:
This is now done.

PR #3663, opened 2021-04-16
Quadrat: Add responsive Paragraph styles and start styling the Code b…

danieldudzic:
Add responsive styles for the Paragraph

Part of #3640

### DIFF
--- a/quadrat/child-experimental-theme.json
+++ b/quadrat/child-experimental-theme.json
@@ -42,12 +42,12 @@
 					},
 					{
 						"name": "Extra Large",
-						"size": "35px",
+						"size": "min(max(28px, 5vw), 35px)",
 						"slug": "extra-large"
 					},
 					{
 						"name": "Huge",
-						"size": "48px",
+						"size": "min(max(36px, 5vw), 48px)",
 						"slug": "huge"
 					}
 				],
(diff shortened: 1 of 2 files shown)


pbking:
So this change makes 'Huge' and 'Extra Large' font sizes dynamic right?  Neat technique, I hadn't seen that before).
Are there specs outlining this dynamic behavior that I should check the values against?  I checked against the [figma](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/Quadrat?node-id=50%3A157) but I wasn't sure what to look for.

Looks like the code block should have a background according to the design.
image

Perhaps that (the code block changes) would be better served in its own PR.

danieldudzic:
@pbking 

@MaggieCabrera suggested using fluid typography :) - it seems to be working pretty well.

>  I checked against the figma but I wasn't sure what to look for.

Paragraph font sizes can be found in the following Figma: https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/Quadrat?node-id=760%3A112

> Perhaps that (the code block changes) would be better served in its own PR.

Good idea. I'll move it to a separate PR tomorrow.

jffng:
Very cool, thanks @danieldudzic !

I revised some of the heading max values because they weren't actually hitting their largest size with the previous minmax rules.

Kapture 2021-04-22 at 14 19 30

jffng (approved):

Issue #3642, opened 2021-04-12
Quadrat: Block Patterns


jffng:
Quadrat should ship with the following block patterns: 

- [x] [Illustrated Promo](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/?node-id=50%3A11) https://github.com/Automattic/themes/pull/3662/
![image](https://user-images.githubusercontent.com/5375500/114431592-82543380-9b74-11eb-8313-fe6f67f0f6d8.png)

- [x] [CTA](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/?node-id=50%3A11) https://github.com/Automattic/themes/pull/3662/
![image](https://user-images.githubusercontent.com/5375500/114431475-5fc21a80-9b74-11eb-9130-3b94a85f0c84.png)

- [x] [Text over rhombus](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/?node-id=50%3A11) https://github.com/Automattic/themes/pull/3676
![image](https://user-images.githubusercontent.com/5375500/114431384-4c16b400-9b74-11eb-8294-84e21d546d76.png)

- [x] [Latest Episodes](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/?node-id=50%3A11) https://github.com/Automattic/themes/pull/3665
![image](https://user-images.githubusercontent.com/5375500/114431632-926c1300-9b74-11eb-8749-10577dc167b7.png)

- [x] [Episode](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/?node-id=50%3A11) https://github.com/Automattic/themes/pull/3695
![image](https://user-images.githubusercontent.com/5375500/114431688-a31c8900-9b74-11eb-90ec-e17523c7ba02.png)

- [x] [Join](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/?node-id=50%3A11)
![image](https://user-images.githubusercontent.com/5375500/114431798-c0515780-9b74-11eb-8356-214b3dd0d7e5.png)

- [x] [Headline and button](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/?node-id=50%3A158) https://github.com/Automattic/themes/pull/3744
![image](https://user-images.githubusercontent.com/5375500/114431756-b4659580-9b74-11eb-8e5e-5c52542ca84f.png)

[...]

PR #3665, opened 2021-04-16
Quadrat: Add Latest episodes block pattern

MaggieCabrera:

This PR adds the "Latest Episodes" block pattern. This pattern uses one new utility class to remove the bottom margin from the image block so that it sits flush with the following block. To do: upload the images to the staging website. ✅ 

Screenshot 2021-04-16 at 12 39 40

#### Related issue(s):

Partially addresses #3642

### DIFF
--- a/quadrat/inc/block-patterns.php
+++ b/quadrat/inc/block-patterns.php
@@ -80,32 +80,32 @@ if ( ! function_exists( 'quadrat_register_block_patterns' ) ) :
 					'content'    => '<!-- wp:columns {"align":"wide"} -->
 					<div class="wp-block-columns alignwide"><!-- wp:column -->
 					<div class="wp-block-column"><!-- wp:image {"id":1438,"sizeSlug":"large","linkDestination":"none","className":"image-no-margin mb-0"} -->
-					<figure class="wp-block-image size-large image-no-margin mb-0"><img src="http://freethemes.local/wp-content/uploads/2021/04/Screen-usage-1024x623.jpg" alt="Screen usage" class="wp-image-1438"/></figure>
+					<figure class="wp-block-image size-large image-no-margin mb-0"><img src="http://freethemes.local/wp-content/uploads/2021/04/Screen-usage-1024x623.jpg" alt="' . esc_html__( 'Screen usage', 'quadrat') . '" class="wp-image-1438"/></figure>
 					<!-- /wp:image -->
 					
 					<!-- wp:cover {"overlayColor":"darker-blue","minHeight":425,"className":"mt-0","style":{"spacing":{"padding":{"top":"40px","right":"40px","bottom":"40px","left":"40px"}}}} -->
 					<div class="wp-block-cover has-darker-blue-background-color has-background-dim mt-0" style="padding-top:40px;padding-right:40px;padding-bottom:40px;padding-left:40px;min-height:425px"><div class="wp-block-cover__inner-container"><!-- wp:heading {"level":3,"textColor":"pink"} -->
-					<h3 class="has-pink-color has-text-color">Episode 1: How Screens Affect Hormones</h3>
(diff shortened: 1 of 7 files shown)


quadrat/inc/block-patterns.php
```diff
@@ -27,6 +27,47 @@ function quadrat_register_block_patterns() {
...
 
+			register_block_pattern(
+				'quadrat-blocks/example',
+				array(
+					'title'      => __( 'Latest Episodes', 'quadrat' ),
+					'categories' => array( 'quadrat' ),
+					'content'    => '<!-- wp:columns {"align":"wide"} -->
+					<div class="wp-block-columns alignwide"><!-- wp:column -->
+					<div class="wp-block-column"><!-- wp:image {"id":1438,"sizeSlug":"large","linkDestination":"none","className":"image-no-margin mb-0"} -->
+					<figure class="wp-block-image size-large image-no-margin mb-0"><img src="http://freethemes.local/wp-content/uploads/2021/04/Screen-usage-1024x623.jpg" alt="' . esc_html__( 'Screen usage', 'quadrat') . '" class="wp-image-1438"/></figure>
```
MaggieCabrera: We probably need a better alt text than this

MaggieCabrera: Also, @beafialho should we keep the lorem ipsum or do we want other kind of text for these?

beafialho: Yes, @MaggieCabrera here's the alt text for the images of the patterns. I'm also writing alt text for all the other abstract illustrations [here](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/Quadrat?node-id=111%3A34).

"Illustration of two women."
Illustration of two women

"Illustration of an introspective woman."
Illustration of an introspective woman

"Illustration of a woman working on a laptop."
Illustration of a woman working on a laptop

"Illustration of a woman climbing steps."
Illustration of a woman climbing steps

> Also, @beafialho should we keep the lorem ipsum or do we want other kind of text for these?

I've replaced the lorem ipsum [here](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/Quadrat?node-id=443%3A5).

MaggieCabrera: Thanks so much!

jffng:
On mobile, should these have some vertical margin?

Screen Shot 2021-04-16 at 12 26 59 PM

If so, I wonder if that's something more global with the columns block CSS, or maybe we can wrap them in a group, I'm not sure.

quadrat/sass/_utility.scss
```diff
@@ -0,0 +1,7 @@
+.image-no-margin {
```
jffng, line 1: Nice candidate to move to BCB eventually?

beafialho:
> On mobile, should these have some vertical margin?
> 
> Screen Shot 2021-04-16 at 12 26 59 PM
> 
> If so, I wonder if that's something more global with the columns block CSS, or maybe we can wrap them in a group, I'm not sure.

I think this is very similar to what I had mocked up, but would like @kjellr's input here.

kjellr:
> If so, I wonder if that's something more global with the columns block CSS, or maybe we can wrap them in a group, I'm not sure.

I do think it makes sense to add a little top/bottom margin. Do they have margin in the editor? We should also ensure it matches in both places.

quadrat/sass/_utility.scss
MaggieCabrera: Yes, I think so, I found out there's a .com plugin that does exactly this but adding a toggle inside the editor!

MaggieCabrera:
I added a group block wrapper with 30px padding bottom. 

The editor (the front is the same):

Screenshot 2021-04-22 at 15 27 45

The frontend on mobile:

Screenshot 2021-04-22 at 15 27 52

kjellr (commented):
This looks great! Just one small comment: the body text in here should have a line height of `1.7`. Once that's all set, this should be good to go.

jffng:
Nice work, thanks everyone!

PR #3667, opened 2021-04-16
Quadrat: override BCB's templates with php ones

MaggieCabrera:

As discussed on zoom, this PR overrides BCB's templates with new ones for Quadrat using the same approach that was used for the home one.

### DIFF
--- a/quadrat/functions.php
+++ b/quadrat/functions.php
@@ -41,21 +41,16 @@ function quadrat_override_index_template( $template ) {
 	switch ($template) {
 		case is_home() || is_front_page() :
 			return locate_template( array( 'index.php' ) );
-			break;
 		case is_404() :
 			return locate_template( array( '404.php' ) );
-			break;
 		case is_search() :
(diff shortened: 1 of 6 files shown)


quadrat/functions.php
```diff
@@ -37,9 +37,25 @@ function quadrat_scripts() {
...
  */
 function quadrat_override_index_template( $template ) {
-	if ( is_home() || is_front_page() ) :
-		$template = locate_template( array( 'index.php' ) );
-	endif;
+
+	switch ($template) {
+		case is_home() || is_front_page() :
+			$template = locate_template( array( 'index.php' ) );
+			break;
```
pbking: Rather than setting a variable and breaking can this just be an early exit?

return locate_template( array( 'index.php' ) );

jffng: Also, can the default case just be `index.php` or do we need to have a case for `is_home() || is_front_page()`?

pbking:
This LGTM!
I think it's the right fix for the day.

PR #3670, opened 2021-04-16
Experimental: Try Maywood FSE version for Dotcom FSE migration

vindl:
**Do not merge.**

Quick experiment for creating a Maywood theme version that works with FSE. This is pretty much a Frankentheme in which I used Maywood and grafted chunks of Blank Canvas Blocks on top of it, just to get a working prototype in the least amount of time. 😄 

For more details and testing instructions check out pbAok1-27H-p2.


pbking:
Have you experimented with [Maywood-blocks](https://github.com/Automattic/themes/tree/trunk/mayland-blocks)? 

That theme is a [child of blank-canvas-blocks]( https://github.com/Automattic/themes/blob/trunk/mayland-blocks/style.css#L11) and is already FSE-ready.

MaggieCabrera:
You may want to use [Mayland Blocks](https://github.com/Automattic/themes/tree/trunk/mayland-blocks) as a reference since it's a Varia theme (like Maywood) that has been converted to be a BCB child. It is quite straightforward to create a BCB child and I believe you have done most of the work here, but doing it the Mayland way will require much less css and boilerplate.

vindl:
> It is quite straightforward to create a BCB child and I believe you have done most of the work here, but doing it the Mayland way will require much less css and boilerplate.

Thanks! I did rely on Mayland's example as a guide for this and made it a child of BCB, but didn't dive into much of the styling details yet.

MaggieCabrera:
Yeah, sorry, I missed that this actually a BCB child on my first look, my bad! I've started a PR that adds instructions on how to create a child theme for BCB here: https://github.com/Automattic/themes/pull/3689 

On a first look, I don't think most (if any) of the CSS of this PR is needed at all. Check how we build the theme.json file on the instructions above and that should cover most of the styles that the theme needs (at least it did for Mayland, that's the bit that's going to need more manual work, checking what wasn't covered by BCB's styles+theme.json combination).

In terms of the actual idea of creating a block-based version of Maywood, I'm thinking maybe that's not what's needed at all. As part of a big effort to refactor Varia children so that they have less complicated headers, we actually moved Maywood's header to Varia: https://github.com/Automattic/themes/tree/trunk/varia/template-parts/header (and we did the same for the footer). As you can see, there's a template part called [fse-header](https://github.com/Automattic/themes/blob/trunk/varia/template-parts/header/fse-header.php) with the code for the deprecated FSE. Wouldn't removing the places where this file is imported remove the problem completely? Sorry if I'm missing a bit of context here.

scruffian:
Is this about getting a block theme version of Maywood that can be used with core's site editor, or a adding support for the dotcom version of FSE which uses the fse-header?

MaggieCabrera:
I'm not sure I understand the original problem that originated this PR, but I was worried that a different solution could be simpler than this (since block based themes are not exactly ready for the public to use yet). If the problem is that we want to discontinue using the dotcom FSE, I thought fixing the original theme would be a tad easier than building the block version of it, but I'm not sure of the implications of doing that.

vindl:
> As you can see, there's a template part called fse-header with the code for the deprecated FSE. Wouldn't removing the places where this file is imported remove the problem completely? Sorry if I'm missing a bit of context here.

I think we discussed this in Slack recently, but adding it here for context too - removing this would cause custom header and footer content to disappear for legacy Dotcom FSE sites.

> Is this about getting a block theme version of Maywood that can be used with core's site editor, or a adding support for the dotcom version of FSE which uses the fse-header?

Mainly the latter, but I think they are the same thing actually.

> I'm not sure I understand the original problem that originated this PR, but I was worried that a different solution could be simpler than this (since block based themes are not exactly ready for the public to use yet). If the problem is that we want to discontinue using the dotcom FSE, I thought fixing the original theme would be a tad easier than building the block version of it, but I'm not sure of the implications of doing that.

I tried to provide some clarification in the post pbAok1-27H-p2. I'm not sure what fixing the original theme would entail, but we can't just upgrade the existing one in place since that would roll out core FSE for all users that have Maywood active (and we are not ready for that yet).

In any case we haven't yet decided whether we are going to pursue this further, and until that point I'm punting the work on it.

scruffian:
Let's close this since the work is being punted. Feel free to reopen if you want to do more on it.

PR #3537, opened 2021-03-26
Updating Seedlet and family from WPCOM

scruffian:
#### Changes proposed in this Pull Request:
This pulls the changes from D59314-code back into GH

PR #3671, opened 2021-04-19
Seedlet: Update require paths

scruffian:
I updated this code from WPCOM in https://github.com/Automattic/themes/pull/3537/files but I got the path wrong 🤦 

This should fix it...


jffng (approved):
These match what are on wpcom, thanks!

PR #3672, opened 2021-04-19
Quadrat: Add navigation via widget mechanism

scruffian:
This attempts to add navigation to Quadrat by using widgets

### DIFF
--- a/quadrat/functions.php
+++ b/quadrat/functions.php
@@ -38,14 +38,14 @@ require get_stylesheet_directory() . '/inc/block-styles.php';
  */
 function quadrat_override_block_templates( $template ) {
 
-	switch ($template) {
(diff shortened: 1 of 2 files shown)


draganescu:
The theme could also by default add a navigation widget to the Navigation sidebar? The navigation menu widget only supports selecting menus, it would be lovely if it supported locations :)

scruffian:
We're gong to try to achieve this using menu locations in Gutenberg.

scruffian:
> The theme could also by default add a navigation widget to the Navigation sidebar? The navigation menu widget only supports selecting menus, it would be lovely if it supported locations :)

This works fine but the problem is there's no way to show this in the site editor (AFAIK)

Issue #3668, opened 2021-04-16
Quadrat: styles are being overridden by Gutenberg


jffng:
Gutenberg's default block styles are overriding BCB's, for example in the button:

<img width="1277" alt="Screen Shot 2021-04-16 at 11 25 52 AM" src="https://user-images.githubusercontent.com/5375500/115052444-cdcb5200-9e92-11eb-8536-1aadd1e73c84.png">

This started after #3667

MaggieCabrera:
This looks like a GB issue:

Block-based template | PHP template
--- | ---
<img width="522" alt="Screenshot 2021-04-16 at 18 13 38" src="https://user-images.githubusercontent.com/3593343/115053592-946d0380-9edf-11eb-91df-c1a24b24a3d7.png"> | <img width="530" alt="Screenshot 2021-04-16 at 18 13 45" src="https://user-images.githubusercontent.com/3593343/115053589-933bd680-9edf-11eb-8eff-15fb3b7b3f62.png">

When the template is a FSE template the default styles for the button are loaded inline before the styles that come from theme.json. When the template is a php one, the default button styles are loaded from a stylesheet (plugins/gutenberg/build/block-library/blocks/button/style.css) and loaded before what's coming from theme.json.

jffng:
Thanks @MaggieCabrera that helps, I dug into this further. The styles loading in the wrong order seems to be specific to the `locate_template` filter we introduced, not that it's a php template nor a child theme of a block theme.

I tried to recreate the bug unsuccessfully by doing the following:

- Activate emptytheme
- Add an `index.php` which loads the `index.html` (renamed to something else like `block-based-index.html` as to not trigger FSE) via `do_blocks`
- Add a style via theme.json for example `{ styles.core/button.color.background: red }`

The styles load in the correct order, even though it's a php template:

<img width="1791" alt="Screen Shot 2021-04-19 at 3 24 37 PM" src="https://user-images.githubusercontent.com/5375500/115292082-628bb500-a10a-11eb-9ec3-86dbbbddcce3.png">

I can fix this issue by adding `remove_theme_support( 'block-templates' )`.  The downside of doing this is that it disables the ability to add new templates introduced in https://github.com/WordPress/gutenberg/pull/30438

Some related Gutenberg code:

- Load separate block assets: https://github.com/WordPress/gutenberg/blob/7c134eb66f1e5ed67db646e19234d9edf6e3090e/lib/compat.php#L11-L26
[...]

PR #3675, opened 2021-04-19
Quadrat: fix styles loading incorrectly

jffng:
Addresses #3668 

I don't think this is a good solution but maybe helps point to why something is wrong.


MaggieCabrera:
I created a PR on GB to try to address the template loading issue. If that lands, we won't need this and can remove the hacky template overriding in functions.php: https://github.com/WordPress/gutenberg/pull/31123

Issue #3615, opened 2021-04-05
Mayland Blocks + Seedlet Blocks + BCB: Change meta icons to use mask 


MaggieCabrera:
We can use [this approach](https://codepen.io/joen/pen/poRwpPj?editors=1100) (thanks @jasmussen! ) to be able to change the color of the icons based on the corresponding css variable

PR #3676, opened 2021-04-19
Quadrat: add cover pattern

jffng:
This PR adds the following block pattern (part of #3642) that leverages a new block style variation on the cover block.

Screen Shot 2021-04-20 at 10 54 25 AM

h/t @MaggieCabrera for the implementation 🙌 . Doing it this way means users can control the color using the cover blocks built in overlay. Unfortunately it doesn't seem to be working with gradients but maybe we can look into that as a next step.

**Demo**
Kapture 2021-04-20 at 14 26 23


MaggieCabrera:
You may want to have a look at https://github.com/Automattic/themes/issues/3615#issue-850265526 for a way to alter the color of the svg when loaded as a background. This method is being used on the responsive navigation PR for gutenberg

jffng:
Very cool! This seems like a perfect candidate for that method — mask-image + background-color

quadrat/sass/blocks/_cover.scss
```diff
@@ -0,0 +1,12 @@
+.is-style-quadrat-cover-diamond {
+	background-color: inherit;
+	-webkit-mask-image: url(/wp-content/themes/themes/quadrat/assets/rotated.svg);
```
jffng: @MaggieCabrera you may need to adjust the path depending on your setup.

jffng:
Some block markup to try it with:

```
<!-- wp:cover {"overlayColor":"darker-blue","align":"full","className":"is-style-quadrat-cover-diamond"} -->
<div class="wp-block-cover alignfull has-darker-blue-background-color has-background-dim is-style-quadrat-cover-diamond"><div class="wp-block-cover__inner-container"><!-- wp:group {"textColor":"pink","layout":{"inherit":true}} -->
<div class="wp-block-group has-pink-color has-text-color"><!-- wp:heading {"textAlign":"center"} -->
<h2 class="has-text-align-center">Episode 3: Cycle of Emotions</h2>
<!-- /wp:heading --></div>
<!-- /wp:group -->
```

MaggieCabrera:
Ok, it's not super pretty, but it works. I had to use the :after pseudoelement to cover the background of the container div because I couldn't hide it or remove the background value **and** still use the background value on the before pseudoelement. If only the selected values where also defined as css variables...

Screenshot 2021-04-20 at 19 38 52

jasmussen:
This is wild, and very cool.

It seems like the primary headache is the need for _two_ custom colors, one for the background, one for the square. Could we simply show the square as black/transparent to achieve mostly the same, or do they need to be able to be set separately?

kjellr:
I think having the square be black/transparent would be a decent tradeoff, provided that we do the same for the squares we use elsewhere in the theme. 

Can we use a blend mode here, either in the SVG or the CSS? It's possible to replicate the color pretty closely if we use mix-blend mode. The downside is that it renders just a little differently depending on your browser. 

https://codepen.io/kjellr/pen/MWJzEKK?editors=1100

Actually... have we considered creating the entire square shape using only CSS? I'm not sure we're gaining much here by using a SVG.

jffng:
I like all of those suggestions and it simplifies things quite a bit. I pushed a change to reflect those ideas. 

Here's what it looks like with the three theme colors presets as the cover's background color:

a | b | c
------ | ------- | -------
Screen Shot 2021-04-21 at 4 34 22 PM | Screen Shot 2021-04-21 at 4 34 10 PM | Screen Shot 2021-04-21 at 4 41 00 PM

> have we considered creating the entire square shape using only CSS?

I am not sure, but I think downside is with an svg, we get the background-image resizing capabilities of CSS, whereas I think we have to set the size of the div explicitly if we want to guarantee it stays a square.

MaggieCabrera:
I prefer the svg over the square too for the same reasons Jeff explained. And I really like the solution with the blend mode.

kjellr:
> I am not sure, but I think downside is with an svg, we get the background-image resizing capabilities of CSS, whereas I think we have to set the size of the div explicitly if we want to guarantee it stays a square.

Yeah, good call. 👍 

Those screenshots look great! Does it work on pure white/black backgrounds?

jffng:
Ok after some more trial and error, I think I have an approach. It even works with gradients!

Screen Shot 2021-04-22 at 2 40 43 PM
 
I ended up targeting the `:after` element. When I tried adding it to the `:before` element (which is already being used by the cover block) I was getting strange results with mix-blend-mode, like the backgrounds being multiplied: 

Screen Shot 2021-04-22 at 2 44 47 PM

> Does it work on pure white/black backgrounds?

~Yes on pure white since the svg fill is black, but no on black.~ Scratch that, since we're using mix-blend mode it won't work with either pure white or black.

kjellr:
> Scratch that, since we're using mix-blend mode it won't work with either pure white or black.

Maybe we can add a little extra style so if the block includes `has-black-background-color` or `has-white-background-color`, we turn off the blend mode and adjust as necessary?

MaggieCabrera:
Would it be ok to add black and white to the palette? That way we make sure those colors classes exist, else I don’t think the generated class is actually has-black-background-color

MaggieCabrera:
So yeah, since the theme doesn't have a black/white color definition, picking them as custom on the editor will generate inline styles but not a white/black class that we can detect for this. Still, in light of [the direction that colors will take in the future](https://github.com/WordPress/gutenberg/issues/29568), it's safe to assume that those colors will exist in the default palette provided by Gutenberg, and adding them to the theme today shouldn't be an issue. They will show up as part of the palette today, but we can remove them once the changes to GB are done and let them be handled by the editor's default palette then.

kjellr:
> So yeah, since the theme doesn't have a black/white color definition, picking them as custom on the editor will generate inline styles but not a white/black class that we can detect for this.

Until they're provided by default by Gutenberg, I think all of our themes should include black and white options in their palettes.

MaggieCabrera:
I rebased and added the black and white colors to the json and created the exceptions. I had to use mask again, since we have to manipulate the color of the svg when we are not using the blend mode.

Screenshot 2021-04-23 at 17 50 23

kjellr:
Since they're not using blend mode, can we make the black and white versions have a lower opacity? Maybe something like 25%?

Also, a few other small notes: 

- This cover should be wide by default. Let's make it have a min-height of 700px too.
- We also need to shrink down the square. It's currently taking up the full height, but it should be a smaller element proportionally: 

Current|Mockup
---|---
Screen Shot 2021-04-23 at 11 57 16 AM|Cover block with heading

MaggieCabrera:
@kjellr lower the opacity of the square or of the background?

kjellr:
The square: 

Before|After
---|---
Frame 1|Frame 2

MaggieCabrera:
I rebased this, lowered the opacity for the b/w versions, made the square to be 70% of the height of the cover and made the cover wide and min 700px height. Should be good to go I think.

Screenshot_2021-04-26 Screenshot

quadrat/sass/theme.scss
```diff
@@ -1,3 +1,3 @@
 @import '../../blank-canvas-blocks/sass/base/breakpoints'; // Get the mobile-only media query and make it available to this theme's styles
-@import 'blocks/media-text';
-@import "utility";
+@import 'blocks/cover';
+@import 'blocks/media-text';@import "utility";
```
scruffian: I guess this shouldn't be on one line?

quadrat/sass/blocks/_cover.scss
```diff
@@ -0,0 +1,36 @@
...
+		background-repeat: no-repeat;
+		background-size: auto 70%;
+		content: "";
+		mix-blend-mode: soft-light;
+		position: absolute;
+		top: 0;
+		left: 0;
+		bottom: 0;
+		right: 0;
+		z-index: 0 !important; //this avoids weird interactions on the editor
```
scruffian: Do you know what line in the editor causes these issues?

quadrat/functions.php
```diff
@@ -19,6 +19,7 @@ function quadrat_support() {
 				'social'  => __( 'Social Links Navigation', 'quadrat' ),
 			)
 		);
+		remove_theme_support( 'block-templates' );
```
scruffian: Why do we need this?

MaggieCabrera: I don't think we do. I suspect this comes from a rebase I didn't do correctly.

quadrat/sass/theme.scss
MaggieCabrera: probably from the same rebase that I botched

scruffian:
The black/white variations don't seem to be working: Screenshot 2021-04-26 at 12 35 46

scruffian:
I added autoprefixing so that `mask` works in webkit browers.

quadrat/sass/blocks/_cover.scss
MaggieCabrera: I think we don't need this anymore

jffng:
@kjellr want to give this another spin, with an eye for the black + white versions? 🙏

quadrat/sass/blocks/_cover.scss
```diff
@@ -21,7 +21,7 @@
 			opacity: 0;
 		}
 
-		&::after {
+		&:after {
```
scruffian: Why only one `:`?

MaggieCabrera: must have slipped honestly

jffng:
This one's had a lot of back and forth. I think the main points have been addressed, it's getting a bit hard to follow so going to merge and we can address any additional feedback in a follow-up. Thanks everyone!

kjellr:
> @kjellr want to give this another spin, with an eye for the black + white versions? 🙏

Sorry, I was afk! I'll give this a test soon and open a new issue if I run into anything.

PR #3679, opened 2021-04-20
Seedlet: Add alignment support to the separator block

kjellr:
This PR adds proper front-end alignment support to the separator block. This is only needed for the "Wide" block style, since the standard and dots separators don't extend wide anyway. 

---

Editor: 

Screen Shot 2021-04-20 at 9 16 41 AM

Frontend Before: 

Screen Shot 2021-04-20 at 9 16 26 AM

Frontend After: 

Screen Shot 2021-04-20 at 9 17 10 AM


pbking (approved):
LGTM

kjellr:
Thanks @pbking!

PR #3682, opened 2021-04-21
Mayland Blocks - fix block validation errors from single.html

Addison-Stavlo:
If you open the single template in the site editor, there are block validation errors regarding the spacer block:

image

image

This seems to be a result of the markup comments for these two spacer blocks using 'px' units.  After removing the unnecessary units these blocks seems to validate correctly and are set to their 16px height attribute without errors as expected.

### DIFF
--- a/mayland-blocks/block-templates/single.html
+++ b/mayland-blocks/block-templates/single.html
@@ -10,7 +10,7 @@
 
 <!-- wp:template-part {"slug":"post-meta"} /-->
 
-<!-- wp:spacer {"height":16px} -->
+<!-- wp:spacer {"height":16} -->
 <div style="height:16px" aria-hidden="true" class="wp-block-spacer"></div>
 <!-- /wp:spacer -->
 
@@ -22,7 +22,7 @@
 <!-- wp:group {"layout":{"inherit":true}} -->
 <div class="wp-block-group">
 
-<!-- wp:spacer {"height":16px} -->
+<!-- wp:spacer {"height":16} -->
 <div style="height:16px" aria-hidden="true" class="wp-block-spacer"></div>
 <!-- /wp:spacer -->
 


jeyip:
## Testing
### Requirements
- [x] Spacer blocks render properly for the "single" block template
- [x] Spacer block validation errors are no longer present in the dev console 

### Browsers
- [x] Safari
- [x] Edge
- [x] Chrome
- [x] Firefox

**Note:**
Here's how I tested this PR:
1. [Spin up a local instance of Gutenberg](https://github.com/WordPress/gutenberg/blob/trunk/docs/getting-started/tutorials/devenv/README.md#wordpress-development-site)
2. Install the Mayland blocks theme on the local instance (if using `wp-env`, this can be done through the `.wp-env.overrides` file)
3. Activate the Mayland blocks theme from wp-admin. This should enable the site editor.
4. Navigate to the site editor
5. In the navigation sidebar, click on the "Single Post" template
6. Verify that the space blocks are loading properly

jeyip (approved):
This looks good to me 👍

Addison-Stavlo:
@jeyip thanks for testing! And thanks for including testing instructions (I probably should have done that in the initial post 😅 ).

Addison-Stavlo:
Also if anyone curious why that creates a validation error, its because the 'height' attribute is [defined as a number](https://github.com/WordPress/gutenberg/blob/6244cf57555665dd55d64caad98674848f73678a/packages/block-library/src/spacer/block.json#L7) (not a string).  But when we add 'px' units at the end it is no longer interpreted as a number and isn't a valid attribute value.

jeyip:
>  thanks for testing! And thanks for including testing instructions (I probably should have done that in the initial post 😅 ).

Of course! Thanks for setting up a fix so quickly 👍

PR #3688, opened 2021-04-22
BCB + children: added border supports to buttons

MaggieCabrera:

After https://github.com/WordPress/gutenberg/pull/30124 was merged, now we have support for border color, style and width for blocks. BCB was only defining custom values for buttons (and forms, but since those are not blocks, they don't apply here) so this PR removes the custom CSS that BCB was applying to add support for them instead.

I rebuilt Mayland + Seedlet with the changes.

### DIFF
--- a/blank-canvas-blocks/assets/ponyfill.css
+++ b/blank-canvas-blocks/assets/ponyfill.css
@@ -452,11 +452,9 @@ p.has-drop-cap:not(:focus):first-letter {
 	font-weight: var(--wp--custom--button--typography--font-weight);
 	font-family: var(--wp--custom--button--typography--font-family);
 	font-size: var(--wp--custom--button--typography--font-size);
-	border-radius: var(--wp--custom--button--border--radius);
 	line-height: var(--wp--custom--button--typography--line-height);
 	text-decoration: none;
 	color: var(--wp--custom--button--color--text);
-	border-color: var(--wp--custom--button--color--text);
 	background-color: var(--wp--custom--button--color--background);
 	padding: calc(.667em + 2px) calc(1.333em + 2px);
 }
@@ -575,11 +573,9 @@ p.has-drop-cap:not(:focus):first-letter {
 	font-weight: var(--wp--custom--button--typography--font-weight);
 	font-family: var(--wp--custom--button--typography--font-family);
 	font-size: var(--wp--custom--button--typography--font-size);
-	border-radius: var(--wp--custom--button--border--radius);
 	line-height: var(--wp--custom--button--typography--line-height);
 	text-decoration: none;
 	color: var(--wp--custom--button--color--text);
-	border-color: var(--wp--custom--button--color--text);
 	background-color: var(--wp--custom--button--color--background);
 	padding: calc(.667em + 2px) calc(1.333em + 2px);
 }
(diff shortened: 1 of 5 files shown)


jffng:
I'm running Gutenberg trunk and this PR, and I don't see the controls: 

Screen Shot 2021-04-22 at 9 18 00 AM

Screen Shot 2021-04-22 at 9 21 22 AM

Maybe it's an environment thing... Any ideas on what to do? 

In general, I know we know the motions of testing these changes and they're sort of implied, but maybe helpful to add some testing instructions and expected behavior of the change.

MaggieCabrera:
> In general, I know we know the motions of testing these changes and they're sort of implied, but maybe helpful to add some testing instructions and expected behavior of the change.

You are so right :S 

I'm not going to lie, I only tested this in the frontend, so I didn't see that the editor doesn't show the options. These are the GB  styles on the frontend:

Screenshot 2021-04-22 at 15 40 14

I'm not sure why it's not showing on the editor though :S

MaggieCabrera:
I'm testing TT1 blocks like the PR shows and still nothing, on either buttons or the group block. So weird.

jffng:
:shakes fist: Weird because in the Gutenberg PR it shows controls in the editor!

pbking:
I believe that has to be enabled via:
```
   "settings": { 
     "defaults": {
        "core/button":{
			"border": {
				"customColor": true,
				"customRadius": true,
				"customStyle": true,
				"customWidth": true
			}
```

Those are set to false for all blocks, though the [radius is enabled for the button block ](https://github.com/WordPress/gutenberg/blob/trunk/lib/experimental-default-theme.json#L182)(by default).

pbking:
I pushed a change that allows what blocks express support to show the appropriate panels.  Now the **group block** offers the controls.  

The button block, however, still doesn't express support for that in its block.json file.  A quick test shows that adding it seems to work, giving the buttons the option too but I only gave it a quick glance.

image

MaggieCabrera:
Thank you for investigating that Jason! I rebuilt the child themes again to add your change

scruffian:
The controls for the button don't show up in the editor because they haven't been enabled for buttons yet. However it works fine in the front end, so IMO we can merge this. What do y'all think?

MaggieCabrera:
> The controls for the button don't show up in the editor because they haven't been enabled for buttons yet. However it works fine in the front end, so IMO we can merge this. What do y'all think?

I agree, to me it's more important that the theme can control it than the user at the moment

pbking:
This appears to be breaking some stuff, I don't believe we should have brought it in. 

I'm seeing other issues though so it's hard to put my finger on what's amiss...

I'm not sure we should have brought this in though.  The PR really just allows a USER to make changes that a block expresses support for.  As far as I can tell it doesn't have anything to do with how/what we can style via theme.json

pbking:
Or maybe something else completely is breaking what I'm looking at...

scruffian:
> As far as I can tell it doesn't have anything to do with how/what we can style via theme.json

It also removes some CSS

pbking:
Sorry, to clarify by 'it' I meant the Gutenberg issue.  I think it just added user controls to blocks that express support, it doesn't change what can be styled.

scruffian:
Why do you think we should back it out? What is it breaking?

MaggieCabrera:
Apparently the corresponding classes or the borders were missing from the editor and have been very recently added: https://github.com/WordPress/gutenberg/pull/31217 can you tell me what issues you are seeing?

pbking:
The only issues I'm seeing involves the 'outline style' button.
Editor:
image
View:
image
Settings:
```
					"border": {
						"radius": "4px",
						"width": "10px",
						"style": "solid",
						"color": "red"
					},
```

I assume that once the button block expresses support for those border styles the above issues will be accounted for.

PR #3689, opened 2021-04-22
BCB: updated readme with instructions to create child themes

MaggieCabrera:
This PR updates the readme file with instructions to create child themes. It may be interesting to create a tool to build a child for BCB. I don't think it's as useful for block-based children as it may be for universal children, so I'm leaving that up for discussion.


pbking (approved):
LGTM 🚢 

We can update again once Quadrat lands.

PR #3693, opened 2021-04-23
Quadrat: Add menu locations

scruffian:
This adds menu locations to Quadrat for testing with https://github.com/WordPress/gutenberg/pull/30852


quadrat/functions.php
```diff
@@ -9,6 +9,17 @@ function quadrat_support() {
...
 			)
 		);
+
+		add_theme_support( 'block-nav-menus' );
+
+		// This theme uses wp_nav_menu() in two locations.
+		register_nav_menus(
+			array(
+				'primary' => __( 'Primary Navigation', 'quadrat' ),
+				'footer'  => __( 'Footer Navigation', 'quadrat' ),
```
jffng: Could we add just the primary location for now or do you see a benefit to having all three locations?

scruffian: I think we need the social one too, to do the Quadrat nav. Not sure about the footer one...

jffng: The [comps](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/?node-id=443%3A4) suggest the footer is just a WordPress credit, so let's remove that and 🚢 ?

Screen Shot 2021-04-23 at 10 12 55 AM

scruffian: Done

PR #3704, opened 2021-04-23
Blank Canvas Blocks: Make Code block's font-family overwritable

danieldudzic:
#### Related issue(s):
https://github.com/WordPress/gutenberg/issues/31135

PR #3706, opened 2021-04-23
Quadrat: Add Code block styles

danieldudzic:
Until this WordPress/gutenberg#31135 gets resolved, I have added a temporary fix in Blank Canvas Blocks: #3704 which is required for this PR to work

#3640

### DIFF
--- a/quadrat/child-experimental-theme.json
+++ b/quadrat/child-experimental-theme.json
@@ -132,11 +132,6 @@
 						"left": "calc( 2 * var(--wp--custom--margin--horizontal) )"
 					}
 				},
-				"code": {
-					"typography": {
-						"fontFamily": "unset"
-					}
-				},
 				"quote": {
 					"citation": {
 						"typography": {
@@ -230,7 +225,6 @@
 		},
 		"core/code": {
 			"typography": {
-				"fontFamily": "var(--wp--preset--font-family--base)",
 				"fontWeight": "400",
 				"fontSize": "20px",
 				"lineHeight": "1.6"
--- a/quadrat/child-experimental-theme.json
+++ b/quadrat/child-experimental-theme.json
@@ -89,7 +89,8 @@
 					"selection": "var(--wp--preset--color--darker-blue)"
 				},
 				"margin": {
-					"horizontal": "20px"
+					"horizontal": "20px",
+					"vertical": "30px"
 				},
 				"line-height": {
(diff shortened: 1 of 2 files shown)


quadrat/child-experimental-theme.json
```diff
@@ -126,6 +127,11 @@
 						"left": "calc( 2 * var(--wp--custom--margin--horizontal) )"
 					}
 				},
+				"code": {
+					"typography": {
+						"fontFamily": "unset"
```
scruffian: Do we need this?

scruffian: Yes

scruffian:
I removed some unneeded declarations from the child-theme.json.

pbking:
I'm not sure I follow the font family part of this change...

The linked gutenberg issue notes that the code block's font family cannot be changed with styles, and while that does seem to be true the Figma Design doesn't seem to express something other than the browser's default monospace (rather than the base font which seems to be applied).

From what I can tell neither the Gutenberg fix or the BCB fix should be necessary to achieve the[ intended design.](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/Quadrat?node-id=50%3A766)

image

danieldudzic:
> I'm not sure I follow the font family part of this change...
> 
> The linked gutenberg issue notes that the code block's font family cannot be changed with styles, and while that does seem to be true the Figma Design doesn't seem to express something other than the browser's default monospace (rather than the base font which seems to be applied).
> 
> From what I can tell neither the Gutenberg fix or the BCB fix should be necessary to achieve the[ intended design.](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/Quadrat?node-id=50%3A766)
> 
> image

What browser are you using? I noticed the issue in Chrome.

pbking:
I'm using chrome for development.  

On this branch I see the below:
image

Which is not correct for the intended design (noted above).

If I remove all of the typeface changes from quadrat's theme.json then it looks more correct:
image

However, the fix to adjust the typeface isn't needed in that situation, since the default is correct (I believe).

pbking:
I cleaned out this branch and cherry-picked only the important things.
I also removed the font-related theme.json settings to get the monospace font back per the design.
I think this is ready for another look-see.

scruffian (approved):
LGTM

PR #3714, opened 2021-04-26
Try/header template part wrapped

pbking:
Tinkered with https://github.com/Automattic/themes/pull/3654 but put a different spin on it.

I'm still hopeful we can use the navigation block, however this might be an important way to do without that method anyway.

This moves the rendering of the menu to a function, which outputs the HTML of a rendered block, wrapped in an html block to be rendered by the header's do_blocks() call.

This method allows the navigation block markup to be rendered WITHIN a do_blocks() call (rather than 'stacked' as the only option).

[Here ](https://github.com/Automattic/themes/compare/make/quadrat...try/header-template-part-wrapped?expand=1#diff-ee8f0fc00c772d51e645ecbeb1dea27e89e9635b3d3bafcd8a138d59c6f9e862R22)is the magic with the spell being cast [here](https://github.com/Automattic/themes/compare/make/quadrat...try/header-template-part-wrapped?expand=1#diff-8bdc5e7db09a9153176740186c5b59d3b085ab4b70def7caf65f8ff4b2b29fbdR19)..

image


scruffian (commented):
This looks great. My only hesitation is that I think we need a robust plan to move users to the Site Editor seamlessly before we can commit to this approach, otherwise we'll still end up in a situation where we have users stuck on Classic themes.

pbking:
I absolutely wholeheartedly 100% agree.  That was my hesitation in even WORKING on this solution, and even more so pushing it for review and discussion.

It allows things to work in this narrow targeted situation, and I'm not keen to call the Quadrat a 'universal' theme today since it can ONLY work in the classic environment.  IMO it more fits the definition of a 'hybrid' theme (using pieces of classic and block-based) and in that case this solution could be fine.

If the goal is truly 'universal' this won't work.  If we're OK with 'hybrid' and it's just a classic theme with a few new toys it's a viable option.  I only present it as a theme-specific temporary solution to this narrow use-case, not the type of solution that address the broader issues of universality.

MaggieCabrera:
I feel like there's value in exploring solutions that are "not universal". We may not find the perfect universal implementation right away and we may require interim solutions while we get there. Quadrat may need refactors after launch to take it there. Thanks for looking into all possible avenues to get there.

pbking:
The only issues I'm seeing involves the 'outline style' button.
Editor:
image
View:
image
Settings:
```
					"border": {
						"radius": "4px",
						"width": "10px",
						"style": "solid",
						"color": "red"
					},
```

I assume that once the button block expresses support for those border styles the above issues will be accounted for.

scruffian (approved):
It seems like we will have to solve the navigation problem in Gutenberg somehow, so I say lets go with this and wait for a solution to land there...

quadrat/inc/render-navigation-block.php
```diff
@@ -0,0 +1,35 @@
...
+				'echo'           => false,
+			)
+		);
+
+		if ( ! $has_primary_nav_items ) {
+			return;
+		}
+
+		$menu_markup  = '<!-- wp:html -->';
+		$menu_markup .= wp_nav_menu(
```
scruffian: I wonder if we can save this in a variable for use in `$has_primary_nav_items` and then just output it here, to save calling `wp_nav_menu` twice.

scruffian (approved):
LGTM. Just one minor suggestion...

quadrat/inc/render-navigation-block.php
pbking: Oh man good call.  Deleted much codes with that change. :)

PR #3715, opened 2021-04-27
BCB: Fix PHP warnings

### LINKED ISSUE #3698: Mayland Blocks: Pre-activation Live Preview causes Fatal Error
This may be related to my testing environment but live preview pre-activation with Mayland Blocks causes a fatal error. This does not happen with Classic Themes in the same environment.

image

scruffian:
Fixes a couple of warnings from assuming that things are defined which might not be.

To test, try running BCB without the Gutenberg plugin.

#### Related issue(s):
Fixes #3698

### DIFF
--- a/blank-canvas-blocks/functions.php
+++ b/blank-canvas-blocks/functions.php
@@ -58,7 +58,6 @@ add_action( 'wp_enqueue_scripts', 'blank_canvas_blocks_scripts' );
  */
 
 function blank_canvas_blocks_fonts_url() {
-	$fonts_url = '';
 	if ( ! class_exists( 'WP_Theme_JSON_Resolver' ) ) {
 		return '';
 	}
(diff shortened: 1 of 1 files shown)


pbking:
Totally works and fixes the issue in BCB and kin.

Only suggestion is to continue the 'early exit' strategy and also get rid of the last if() block:

```
	if ( ! class_exists( 'WP_Theme_JSON_Resolver' ) ) {
		return '';
	}

	$theme_data = WP_Theme_JSON_Resolver::get_merged_data()->get_settings();
	if ( empty( $theme_data ) || empty( $theme_data['defaults'] ) || empty( $theme_data['defaults']['custom'] ) ) {
		return '';
	}

	if ( ! array_key_exists( 'fontsToLoadFromGoogle', $theme_data['defaults']['custom'] ) ) {
		return '';
	}

	$font_families = $theme_data['defaults']['custom']['fontsToLoadFromGoogle'];
	$font_families[] = 'display=swap';

	// Make a single request for the theme fonts.
	return esc_url_raw( 'https://fonts.googleapis.com/css2?' . implode( '&', $font_families ) );
```

scruffian:
Good call. ✅

PR #3716, opened 2021-04-27
BCB: fix custom fonts loading conditional

jffng:
This PR changes a conditional in BCB so that the custom font loading function does not exit early / incorrectly.

To test, load a child theme of BCB with a custom font defined (e.g. Seedlet Blocks). 

Verify the theme loads the fonts properly.


scruffian:
Yeah I thought I must have done that. Whoops!

PR #3717, opened 2021-04-27
Quadrat: Style lists

scruffian:
This adds styles for lists in Quadrat:

Screenshot 2021-04-27 at 15 34 37

#### Related issue(s):
Part of #3640

### DIFF
--- a/quadrat/child-experimental-theme.json
+++ b/quadrat/child-experimental-theme.json
@@ -129,8 +129,7 @@
 				"list": {
 					"padding": {
 						"left": "calc( 2 * var(--wp--custom--margin--horizontal) )"
-					},
-					"sub-list-style-type": "disc"
+					}
 				},
 				"quote": {
 					"citation": {
--- a/quadrat/child-experimental-theme.json
+++ b/quadrat/child-experimental-theme.json
@@ -129,7 +129,8 @@
 				"list": {
 					"padding": {
 						"left": "calc( 2 * var(--wp--custom--margin--horizontal) )"
-					}
+					},
+					"sub-list-style-type": "disc"
 				},
 				"quote": {
 					"citation": {
--- a/quadrat/sass/blocks/_list.scss
+++ b/quadrat/sass/blocks/_list.scss
@@ -1,4 +1,4 @@
-ul ul,
-ol ul {
-	list-style-type: var(--wp--custom--list--sub-list-style-type);
+ul li,
+ol li {
+	list-style-type: disc;
 }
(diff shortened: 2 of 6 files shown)


quadrat/sass/blocks/_list.scss
```diff
@@ -0,0 +1,4 @@
+ul ul,
+ol ul {
```
pbking: Based on conversation and comps I believe this should be:
```
ul li,
ol li {
```

quadrat/child-experimental-theme.json
```diff
@@ -129,7 +129,8 @@
 				"list": {
 					"padding": {
 						"left": "calc( 2 * var(--wp--custom--margin--horizontal) )"
-					}
+					},
+					"sub-list-style-type": "disc"
```
pbking: I'm not sure I'm keen to mix a variable not used in blank-canvas-blocks into the mix here.  I think it would be confusing and suggest we limit the 'custom' variables to only extend what BCB is doing to ponyfill gutenberg.

My 'druthers would be to just set the li types to 'disc' rather than use a variable at all.

OR to integrate this into Blank Canvas Blocks.  (Though I prefer just ... setting it in quadrat's scss).

quadrat/sass/blocks/_list.scss
scruffian: `list-style-type` is a property of `ol` and `ul` not `li`.

pbking: I [beg to differ](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type).

```
Only a few elements (<li> and <summary>) have a default value of display: list-item. 
However, the list-style-type property may be applied to any element whose display 
value is set to list-item. Moreover, because this property is inherited, it can be set on 
a parent element (commonly <ol> or <ul>) to make it apply to all list items.
```

pbking: However, I see my suggestion was wrong, I think only this is necessary (so that we don't style ol's too)
```
ul li {
  sub-list-style-type: disc;
}
```
Otherwise I continue to see the default stylings for anything other than immediate children of the ul.

pbking: Also I clearly don't know what I'm doing because `ul ul {` also work because of course you know. :P

However, the branch wasn't working correctly, and it must have been caused by something else.

pbking:
Updated:
image

scruffian:
Thanks for the fixes!

Issue #3643, opened 2021-04-12
Quadrat: templates


jffng:
The following templates should be available in Quadrat:

- [x] post
<img width="334" alt="Screen Shot 2021-04-12 at 12 39 20 PM" src="https://user-images.githubusercontent.com/5375500/114430428-44a2db00-9b73-11eb-8b56-6e20fc38a58c.png">

- [x] page
<img width="506" alt="Screen Shot 2021-04-12 at 12 42 27 PM" src="https://user-images.githubusercontent.com/5375500/114430561-68feb780-9b73-11eb-99d1-b8e5f902d06b.png">

- ~search~

- [x] index 
<img width="344" alt="Screen Shot 2021-04-12 at 12 39 03 PM" src="https://user-images.githubusercontent.com/5375500/114430469-508e9d00-9b73-11eb-972b-3499258fbe95.png">

- [x] 404
<img width="339" alt="Screen Shot 2021-04-12 at 12 39 12 PM" src="https://user-images.githubusercontent.com/5375500/114430449-49678f00-9b73-11eb-9db3-326bc56f8005.png">

Figma here: https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/?node-id=50%3A158

kjellr:
Just a heads up that while it's fine to implement these if you'd like, they are not all the required templates for a MVP launch. [Please refer to the "Small Theme" section of the file](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/Quadrat?node-id=443%3A4), which includes the only required page template: 

![Page Layout](https://user-images.githubusercontent.com/1202812/114705005-3ed50300-9cf5-11eb-9fde-e2fef36bbac5.png)

Other pages can follow this template, and posts can follow the default Seedlet post and archive templates, shared across all "small" themes.

jffng:
Thanks @kjellr!

> follow the default Seedlet post and archive templates

From a design perspective that makes sense. But since this won't be a Seedlet child theme I was thinking that we still need to add them — then they can be duplicated across our "small themes". 

I think we can nix the search template. So that leaves the following:

- index
- page
- post (supports featured image + pagination + comments) 
- archive
- 404

Thoughts?

kjellr:
That sounds about right to me. Once you build them once, you likely won't need to make many adjustments to them — they should be more or less the same across all of the small themes.

jffng:
[...]

PR #3721, opened 2021-04-28
Quadrat: Add page templates

scruffian:
This moves templates to the block-templates directory. This should mean that the theme works in the Site Editor if you have it enabled, but if you don't then we fall back to the PHP templates. You can test by disabling the Site Editor with this PR: https://github.com/WordPress/gutenberg/pull/31339

#### Related issue(s):
https://github.com/Automattic/themes/issues/3643


quadrat/block-template-parts/404.html
```diff
@@ -2,7 +2,7 @@
 <div class="wp-block-group">
 
 <!-- wp:heading {"textAlign":"center","level":1,"fontSize":"large"} -->
-<h1 class="has-text-align-center has-large-font-size">Oops! That page can’t be found.</h1>
+<h1 class="has-text-align-center has-large-font-size">Quadrat Oops! That page can’t be found.</h1>
```
MaggieCabrera: I think this change comes from testing template loading, right? cause I've done the same :D

scruffian: Yeah :)

jffng:
- Need separate page + single templates
- Change index.php to use query block so title of post can appear

quadrat/functions.php
```diff
@@ -19,6 +19,8 @@ function quadrat_support() {
 				'social'  => __( 'Social Links Navigation', 'quadrat' ),
 			)
 		);
+
+		remove_theme_support( 'block-templates' );
```
jffng, line 23: Why do we want to remove this?

jffng:
I made some revisions to this PR: 

- Index template now uses query block
- Separated singular into page + single templates because they contain different blocks
- Revised single template to include post tags, post date, and post featured image
- Rebased `make/quadrat`
- Removed `index.html` for now while we consider the theme support to disable the site editor 

It's a significant improvement over what we have now, so going to merge and we can iterate.

PR #3725, opened 2021-04-29
Change colors used in theme.json from descreptive to semantic

pbking:
Does what it says on the tin.


jffng (approved):
Buttons look the same 🚢

PR #3728, opened 2021-04-30
Mayland Blocks: Remove colors from the navigation block

### LINKED ISSUE #3701: Mayland Blocks: Theme palette produces strange effect with Navigation Block
## Steps to replicate

1. Insert nav block into header with Site Editor.
2. Change background colour of nav block to Black using Theme palette.
3. Change text colour to White using Theme palette.

## Result

image

## Expected

Something more like this effect produced with custom colours.

image

scruffian:
This removes some of the color rules that were set for navigation blocks in BCB so that background colors cascade.

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/3701

### DIFF
--- a/blank-canvas-blocks/assets/ponyfill.css
+++ b/blank-canvas-blocks/assets/ponyfill.css
@@ -264,14 +264,7 @@ ol {
 	border-style: var(--wp--custom--navigation--submenu--border--style);
 	border-width: var(--wp--custom--navigation--submenu--border--width);
 	border-color: var(--wp--custom--navigation--submenu--border--color);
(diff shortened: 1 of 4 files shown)


pbking:
This isn't quite dealing with the sub-menu item styling:
image

scruffian:
I added a commit to remove the submenu colors, but it still looks like this:
Screenshot 2021-04-30 at 15 53 06

That's because the colors are coming from Gutenberg. There is work going on to make them more customizable so for now I suggest we leave it like this...

MaggieCabrera (approved):
🚢

PR #3734, opened 2021-04-30
Quadrat: Headlines and Buttons Block styles and Pattern 

MaggieCabrera:

This is a WIP. This PR adds a block style for both the Query block and the Latest Posts block that follows the design of [Headlines and Buttons](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/Quadrat?node-id=443%3A4) on the figma file.

#### Related issue(s):

Partially addresses https://github.com/Automattic/themes/issues/3642

### DIFF
--- a/blank-canvas-blocks/functions.php
+++ b/blank-canvas-blocks/functions.php
@@ -12,6 +12,10 @@ if ( ! function_exists( 'blank_canvas_blocks_support' ) ) :
 		// https://github.com/WordPress/gutenberg/issues/26901
 		add_theme_support( 'responsive-embeds' );
 
+		// Add support for Post Thumbnails on posts and pages.
+		// https://developer.wordpress.org/themes/functionality/featured-images-post-thumbnails/
+		add_theme_support( 'post-thumbnails' );
+
 		// Add support for editor styles.
 		add_theme_support( 'editor-styles' );
 
--- a/quadrat/sass/blocks/_query.scss
+++ b/quadrat/sass/blocks/_query.scss
@@ -1,22 +1,11 @@
 .is-style-quadrat-staggered-posts {
 	li {
+		margin-bottom: 140px;
 		display: grid;
-		grid-column-gap: 20px;
-		grid-template-columns: 1fr 1fr;
-		grid-template-areas: 
-			"post-image post-title"
-			"post-image post-excerpt"
-			"post-image post-author"
-			"post-image post-date"
-			"post-separator post-separator";
-		&:nth-child(even){
-			grid-template-areas: 
-				"post-title post-image"
-				"post-excerpt post-image"
-				"post-author post-image"
-				"post-date post-image"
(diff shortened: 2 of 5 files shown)


blank-canvas-blocks/functions.php
```diff
@@ -12,6 +12,10 @@ function blank_canvas_blocks_support() {
 		// https://github.com/WordPress/gutenberg/issues/26901
 		add_theme_support( 'responsive-embeds' );
 
+		// Add support for Post Thumbnails on posts and pages.
+		// https://developer.wordpress.org/themes/functionality/featured-images-post-thumbnails/
+		add_theme_support( 'post-thumbnails' );
```
MaggieCabrera, line 17: without this the user can't add thumbnails to posts/pages, and those are kind of the point of this block pattern :D

quadrat/sass/blocks/_query.scss
```diff
@@ -0,0 +1,64 @@
...
+	.wp-block-post-excerpt,
+	.wp-block-post-featured-image,
+	.wp-block-separator,
+	.wp-block-post-date,
+	.wp-block-latest-posts__post-author,
+	.wp-block-latest-posts__post-date,
+	.wp-block-latest-posts__featured-image,
+	.wp-block-latest-posts__post-excerpt {
+		//undos the above fix
+		display: revert;
```
MaggieCabrera, line 24: It was either this or using :not on the rule that adds display: contents; I might revisit this and do it some other way, there's 3 levels of extra divs that the editor is adding...

MaggieCabrera:
Even though this is a very interesting experiment, I'm not sure now that the query block/latest posts block are the corrects ones for this pattern in particular. Looking at the design, these blocks are not really posts, but the "Read more" and the category where misleading me to thinking they were. Maybe instead they should be just static blocks that the user adds content to? @beafialho @kjellr I would like your ideas

Right now this is how it looks like for the blocks (it's still a WIP)

Query block:
Screenshot 2021-05-03 at 16 01 21

Latest posts block:
Screenshot 2021-05-03 at 16 01 07

kjellr:
> Maybe instead they should be just static blocks that the user adds content to? @beafialho @kjellr I would like your ideas

If it's a struggle to get this working as a latest posts/query pattern, I think it would be totally fine to make it into a static pattern of a few media + text blocks.

MaggieCabrera:
It’s not that it’s a struggle but that the content designed doesn’t look much like posts to me to be using these blocks.

kjellr:
Yeah, I see what you mean. In any case, a more straightforward media & text blocks version might be better since it would be an opportunity for us to integrate the bundled illustrations. So I think I have a slight preference for that.

MaggieCabrera:
Closed in favor of https://github.com/Automattic/themes/pull/3744 but this is a good experiment for these kinds of layouts using grid.

PR #3744, opened 2021-05-04
Quadrat: Headlines and Buttons Block Pattern

MaggieCabrera:

This PR adds a block pattern that follows the design of [Headlines and Buttons](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/Quadrat?node-id=443%3A4) on the figma file.

Screenshot_2021-05-04 Screenshot

#### Related issue(s):

Partially addresses #3642

### DIFF
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -96,6 +96,13 @@ ul ul {
 	border-bottom-width: 1px;
 }
 
+@media (max-width: 479px) {
+	.headlines-pattern {
+		padding-top: 50px !important;
+		padding-bottom: 50px !important;
+	}
+}
+
 .site-header .wp-block-group {
 	display: flex;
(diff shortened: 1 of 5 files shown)


jffng:
This LGTM, the spacing is a bit much on mobile but I'm not sure we want to write CSS in this case to target it:

Screen Shot 2021-05-04 at 10 23 34 AM

What do you think?

MaggieCabrera:
> This LGTM, the spacing is a bit much on mobile but I'm not sure we want to write CSS in this case to target it:
> 
> What do you think?

Looking at the design on mobile it goes from 140px to around 110. The group block allows for ems, rems and vw units as well, maybe we can use them, but I doubt we can be as precise as what the design determines

jffng (approved):

PR #3745, opened 2021-05-04
Quadrat: Add Join pattern

scruffian:
Adds a "Join" pattern.

To test you need Jetpack installed and running on a custom domain, to make the subscription block work. It might be easier to test on a simple site.

Screenshot 2021-05-04 at 14 09 25

Note: The button / input on the subscription form doesn't match the theme. This is because of this issue:https://github.com/Automattic/jetpack/issues/17590

It would also be really good to get this done so we don't need the custom CSS: https://github.com/WordPress/gutenberg/issues/31459

#### Related issue(s):
https://github.com/Automattic/themes/issues/3642

### DIFF
--- a/quadrat/inc/block-patterns.php
+++ b/quadrat/inc/block-patterns.php
@@ -41,4 +41,4 @@ if ( ! function_exists( 'quadrat_register_block_patterns' ) ) :
 	}
 endif;
 
-add_action( 'init', 'quadrat_register_block_patterns', 1 );
+add_action( 'init', 'quadrat_register_block_patterns' );
--- a/quadrat/inc/block-patterns.php
+++ b/quadrat/inc/block-patterns.php
@@ -27,7 +27,7 @@ if ( ! function_exists( 'quadrat_register_block_patterns' ) ) :
 				'listen-to-the-podcast',
 			);
 
-			if ( \WP_Block_Type_Registry::get_instance()->is_registered( 'jetpack/subscriptions' ) ) {
+			if ( class_exists( 'WP_Block_Type_Registry' ) && \WP_Block_Type_Registry::get_instance()->is_registered( 'jetpack/subscriptions' ) ) {
 				$block_patterns[] = 'join';
 			}
 
@@ -41,4 +41,4 @@ if ( ! function_exists( 'quadrat_register_block_patterns' ) ) :
 	}
 endif;
 
-add_action( 'init', 'quadrat_register_block_patterns' );
+add_action( 'init', 'quadrat_register_block_patterns', 1 );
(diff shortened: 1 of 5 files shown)


quadrat/inc/block-patterns.php
```diff
@@ -151,8 +151,34 @@ function quadrat_register_block_patterns() {
...
+							<!-- /wp:columns -->',
+					)
+				);
+			}
 		}
 	}
 endif;
 
-add_action( 'after_setup_theme', 'quadrat_register_block_patterns', 12 );
+add_action( 'init', 'quadrat_register_block_patterns' );
```
scruffian, line 44: This change is needed so that the blocks are registered when we register the patterns. The problem is it moves Quadrat to the bottom of the patterns list drop down.

quadrat/inc/block-patterns.php
```diff
@@ -151,8 +151,34 @@ function quadrat_register_block_patterns() {
 				)
 			);
 
+			if ( \WP_Block_Type_Registry::get_instance()->is_registered( 'jetpack/subscriptions' ) ) {
```
jffng: Should we add a check if the `WP_Block_Type_Registry` class exists?

quadrat/inc/block-patterns.php
jffng: Hmm I see. I wonder if we can re-order the list somehow?

jffng (commented):
This makes sense and looks good considering the upstream changes that should happen. 

My only comment is maybe the pattern should be alignwide by default?

scruffian:
We can get it higher in the list:
Screenshot 2021-05-06 at 11 57 28

@kjellr do you have a preference for where this should go?

quadrat/sass/theme.scss
```diff
@@ -10,3 +10,7 @@
 .wp-block-post-featured-image {
 	margin-top: 0;
 }
+
+.subscription-column {
+	text-align: right;
+}
```
MaggieCabrera: should this be inside [this](https://github.com/Automattic/themes/tree/make/quadrat/quadrat/sass/block-patterns) folder?

MaggieCabrera (approved):
> This makes sense and looks good considering the upstream changes that should happen.

+1

kjellr:
> @kjellr do you have a preference for where this should go?

Theme patterns generally show up at the top of the list, so if we can do that I think that's the best route. (I haven't tested this PR yet, so it's possible you're already doing that now).

quadrat/inc/block-patterns.php
```diff
@@ -28,15 +27,18 @@ function quadrat_register_block_patterns() {
 				'listen-to-the-podcast',
 			);
 
+			if ( class_exists( 'WP_Block_Type_Registry' ) && \WP_Block_Type_Registry::get_instance()->is_registered( 'jetpack/subscriptions' ) ) {
```
kjellr, line 30: We should update the wpcom-only patterns in Blank Canvas to use this method, rather than loading them in their own separate `wpcom.php` file.

PR #3746, opened 2021-05-04
Quadrat: Add listen pattern

scruffian:
This adds a listen pattern:

Screenshot 2021-05-04 at 15 15 54

It's hard to style the audio player without custom JS. The alternative would be to use the Jetpack podcast player:

Screenshot 2021-05-04 at 15 18 19

#### Related issue(s):
https://github.com/Automattic/themes/issues/3642

### DIFF
--- a/quadrat/inc/block-patterns.php
+++ b/quadrat/inc/block-patterns.php
@@ -124,10 +124,10 @@ if ( ! function_exists( 'quadrat_register_block_patterns' ) ) :
 					'categories' => array( 'quadrat' ),
 
 					'content'    => '<!-- wp:media-text {"mediaId":1497,"mediaLink":"https://quadrat.mystagingwebsite.com/wp-content/uploads/2021/04/Introspection-scaled.jpg","mediaType":"image","mediaWidth":40,"mediaSizeSlug":"large","imageFill":false,"className":"has-pink-color has-darker-blue-background-color has-text-color has-background is-style-default"} -->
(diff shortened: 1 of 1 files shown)


beafialho:
I'd love to see this player come to life as none of these are ideal, @kjellr what do you think?

Captura de ecrã 2021-05-05, às 09 36 55

The first player has a simpler less constrained design, but if it's not possible to customise its shape and colors, I'd go for the second one. Is this one an option? Can it be customised?

Captura de ecrã 2021-05-05, às 09 35 05

MaggieCabrera:
Can we make a block style for the audio block? And hide the default player with css and with a little javascript make the play/pause functions? The design only shows the play button, the native player also has volume controls, by the way.

kjellr:
The custom player seems like block territory. For now, I think it makes sense to just use the default players, and if something doesn't work nicely with those players, we omit for launch. 

We can always come back later and add a more complicated, custom-styled player, but I don't think it's a main priority for launch.

scruffian:
I'm going to merge this as it. It would be great explore a nicer player in the future...

jeffikus:
> We can always come back later and add a more complicated, custom-styled player, but I don't think it's a main priority for launch. 

We should probably talk with the Jetpack folks about if they have a themable player planned perhaps?

scruffian:
There is already a themeable player in Jetpack (see the second screenshot).

jeffikus:
@scruffian cool, that is interesting - last time I checked that was for a playlist of files, so you can do individual files now too? Then that makes sense to add support for skinning that one.

Perhaps we could look at a core contribution down the road to skin the default audio block.

PR #3752, opened 2021-05-04
Quadrat: remove background color from buttons 

jffng:
This PR removes the background color attributes from the buttons inside this pattern.

This way, the hover styles will work by default.


pbking:
Good call. 👍

PR #3763, opened 2021-05-05
Use build process in child theme.

scruffian:
This PR updates Mayland Blocks to use the same build process as Seedlet Blocks.


scruffian:
I'm approving this as @jffng wrote it.

PR #3768, opened 2021-05-06
Refactor quadrat theme.json to match v1 shape

pbking:
Refactor quadrat theme.json to match v1 shape

Also, changed BCB's element>link configuration to uh... work.


jffng:
Closing since #3773 was merged

PR #3770, opened 2021-05-06
Refactor/theme heading styles json

pbking:
The change: 
* moves the header styles from "blocks" to "elements"
* moves common header styles (family/height/weight) from h?-specific blocks to a common 'core/header' location
* Refactors BCB, Mayland-blocks and Seedlet-Blocks
* Fixes Mayland's theme build script
* Fixes block styles of post-title


jffng (approved):
LGTM, thank you!

PR #3775, opened 2021-05-06
BCB: Only show patterns on the parent theme

scruffian:
We shouldn't show Blank Canvas patterns on the child themes.

To test, make sure that the Blank Canvas patterns don't show on Seedlet Blocks, but they do when you are on BCB.


danieldudzic (approved):
Simple solution - confirmed that it works just fine :)

PR #3761, opened 2021-05-05
Quadrat: add custom logo support

jffng:
This PR adds support for a custom logo. It depends on this PR: https://github.com/WordPress/gutenberg/pull/31513 

Screen Shot 2021-05-05 at 12 06 11 PM

This is a part of #3661

PR #3776, opened 2021-05-06
Quadrat: add custom logo support (rebase)

jffng:
Clean version of #3761.

Desktop | Mobile
------- | --------
Screen Shot 2021-05-06 at 2 47 45 PM|Screen Shot 2021-05-06 at 2 47 39 PM

### DIFF
--- a/quadrat/sass/_header.scss
+++ b/quadrat/sass/_header.scss
@@ -1,16 +1,14 @@
 .site-header {
-	margin-bottom: var(--wp--custom--margin--vertical);
-
 	> .wp-block-group {
 		align-items: center;
 		display: flex;
 		flex-wrap: wrap;
 		justify-content: flex-start;
 		overflow: inherit;
-		padding: 60px 20px;
+		padding: 20px 20px 60px; // TODO: Maybe replace with a responsive custom variable?
 
 		@include break-mobile() {
-			padding: 35px 35px;
+			padding: var(--wp--custom--margin--vertical);
 		}
 
 		.wp-block-site-logo {
@@ -18,7 +16,8 @@
 
 			@include break-mobile-only(){
 				flex-basis: 100%;
-				margin-bottom: 20px;
+				margin: 0 0 10px;
+				text-align: center;
 			}
 
 			a > img {
--- a/quadrat/sass/_header.scss
+++ b/quadrat/sass/_header.scss
@@ -1,14 +1,28 @@
 .site-header {
+	margin-bottom: var(--wp--custom--margin--vertical);
+
 	> .wp-block-group {
+		align-items: center;
 		display: flex;
-		justify-content: space-between;
+		justify-content: flex-start;
 		overflow: inherit;
+		flex-wrap: wrap;
+
 		.wp-block-site-logo {
-			margin-bottom: var(--wp--custom--margin--vertical);
+			margin-right: var(--wp--custom--margin--horizontal);
+			@include break-mobile-only(){
+				margin: 0 0 var(--wp--custom--margin--vertical);
+				flex-basis: 100%;
+			}
 		}
 		
 		.wp-block-site-title {
-			margin-top: 0;
+			margin: 0;
+		}
+
+		.wp-block-navigation {
+			margin-left: auto;
+			padding-right: 0;
 		}
 	}
 
--- a/quadrat/sass/_header.scss
+++ b/quadrat/sass/_header.scss
@@ -22,7 +22,7 @@
 			}
 
 			a > img {
-				height: 35px;
+				height: 64px;
 				width: auto;
 			}
 		}
--- a/quadrat/sass/_header.scss
+++ b/quadrat/sass/_header.scss
@@ -5,7 +5,7 @@
 		flex-wrap: wrap;
 		justify-content: flex-start;
 		overflow: inherit;
-		padding: 20px 20px 60px; // TODO: Maybe replace with a responsive custom variable?
+		padding: 10px 20px 60px; // TODO: Maybe replace with a responsive custom variable?
 
 		@include break-mobile() {
 			padding: var(--wp--custom--margin--vertical);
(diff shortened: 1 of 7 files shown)


quadrat/template-parts/header.php
```diff
@@ -14,7 +14,8 @@
 <header class="site-header" role="banner">
 	<?php
 	echo do_blocks(
-		'<!-- wp:group {"style":{"spacing":{"padding":{"right":"35px","left":"35px"}}}} --><div class="wp-block-group" style="padding-right:35px;padding-left:35px">
+		'<!-- wp:group {"style":{"spacing":{"padding":{"top":"35px","right":"35px","left":"35px"}}}} --><div class="wp-block-group" style="padding-top:35px;padding-right:35px;padding-left:35px">
```
jffng: What do you think about using CSS to create the padding for the header instead?

If we add the padding via a block style attribute (like is being done here), it can in theory be controlled by the site editor. The drawback is this value isn't responsive. And since this is already a work around for the nav, this template part wouldn't actually be consumed by the site editor when the theme switches to using it.

scruffian: > And since this is already a work around for the nav, this template part wouldn't actually be consumed by the site editor when the theme switches to using it.

I think this will not be the case when the Site Editor launches - we should be able to use the location attribute and switch to a block template.

scruffian: > The drawback is this value isn't responsive. 

Could we think about this differently by follow the intrinsic design idea?

jffng:
I took a pass at this to match the [mock-ups](https://github.com/Automattic/themes/pull/3761#issuecomment-833399092):

Desktop | Mobile
------- | --------
Screen Shot 2021-05-10 at 4 53 00 PM | Screen Shot 2021-05-10 at 4 52 46 PM

It is ready for another review cc @Automattic/theam @kjellr @beafialho

quadrat/sass/_header.scss
```diff
@@ -1,8 +1,39 @@
...
+
+	> .wp-block-group {
 		align-items: center;
+		display: flex;
+		flex-wrap: wrap;
+		justify-content: flex-start;
 		overflow: inherit;
+		padding: 60px 20px;
+
+		@include break-mobile() {
```
jffng, line 10: I'm not sure about another way to handle the responsiveness besides media queries, but I've opened an issue here about spacing to discuss more generally: https://github.com/Automattic/themes/issues/3818

quadrat/template-parts/header.php
```diff
@@ -14,7 +14,8 @@
 <header class="site-header" role="banner">
 	<?php
 	echo do_blocks(
-		'<!-- wp:group {"style":{"spacing":{"padding":{"right":"35px","left":"35px"}}}} --><div class="wp-block-group" style="padding-right:35px;padding-left:35px">
```
jffng, line 17: Once this is a template part, users will be able to adjust the spacing freely, so I don't see value to applying the styles in line at this stage.

kjellr:
This looks good, but since folks won't be able to adjust the size until FSE is available, I wonder if we should increase it? The logo is pretty small today: 

Now (35px):
Screen Shot 2021-05-11 at 1 10 22 PM
Screen Shot 2021-05-11 at 1 10 29 PM


Suggested (64px): 
Screen Shot 2021-05-11 at 1 11 14 PM
Screen Shot 2021-05-11 at 1 12 36 PM

@beafialho, what do you think?

jffng:
Adjusted height to 64px, let know if this is good to go or needs any further revisions.

kjellr:
@jffng: would it be possible to center the logo on mobile like this: 

Screen Shot 2021-05-18 at 8 47 08 AM

The only scenario where that would look weird is if there's no site title, but it's currently not possible to remove that anyway, right?

(I'm also seeing a lot of extra space up top there, but I'm not sure if that's something we'd like to address in this PR?)

jffng:
@kjellr 

> would it be possible to center the logo on mobile like this:

👍 

> The only scenario where that would look weird is if there's no site title, but it's currently not possible to remove that anyway, right?

Right, we currently are not adding any customizer options to hide the site title.

> I'm also seeing a lot of extra space up top there, but I'm not sure if that's something we'd like to address in this PR?)

Ah, I was following the design [here](https://github.com/Automattic/themes/pull/3761#issuecomment-833399092) but let me know if it should be reduced.

kjellr:
> Ah, I was following the design here but let me know if it should be reduced.

Oh, I see the confusion — the measurements there are for the logo width (which we decided to ignore for now), not the height above the logo. 

Let's go with this: 

Screen Shot 2021-05-18 at 10 19 19 AM


That's `20px 20px 60px 20px` padding around the header on mobile:

Screen Shot 2021-05-18 at 10 20 11 AM

And only `10px` margin below the logo on mobile: 

Screen Shot 2021-05-18 at 10 20 18 AM

jffng:
Okay, I took another pass to update the header spacing on mobile: 

Screen Shot 2021-05-18 at 12 26 08 PM

Screen Shot 2021-05-18 at 12 26 16 PM

Should be ready for one more review @kjellr

kjellr:
This is close, but now that I'm seeing it again I think we should even the space out just a little more. Apologies for the back and forth on this, that's on me. 😅 I promise this is my final recommendation here!

No Logo|With Logo
---|---
Screen Shot 2021-05-18 at 1 30 24 PM|Screen Shot 2021-05-18 at 1 29 53 PM

To get there: 

Use `10px 20px 60px` padding around the header itself on mobile: 

Screen Shot 2021-05-18 at 1 32 56 PM

... and then `20px` of top and bottom margin around the logo on mobile: 

Screen Shot 2021-05-18 at 1 35 33 PM

jffng:
No worries! All in the details 😄 

No logo | With logo
------- | --------
Screen Shot 2021-05-18 at 2 01 33 PM | Screen Shot 2021-05-18 at 2 01 04 PM

kjellr (approved):
Looks good! Thanks for bearing with me.

Issue #3743, opened 2021-05-04
Quadrat: Adjust pullquote text size and line height


beafialho:
The pullquote text currently doesn't feel balanced. Citation text size is ok as it is but here's what I think can be adjusted:

1. Change font to Medium instead of Regular Italic for the quote.
2. Change text size to 50pt and line height to 1,3.
3. Change Regular Italic to Regular in citation.
4. Remove the divider line at the bottom, leave only the one at the top.

How it looks:
<img width="734" alt="Captura de ecrã 2021-05-04, às 09 54 47" src="https://user-images.githubusercontent.com/65220155/116999278-4e81af00-acd7-11eb-8cda-9ff5a482f52c.png">

How it should look:
<img width="842" alt="Captura de ecrã 2021-05-04, às 12 53 28" src="https://user-images.githubusercontent.com/65220155/116999618-c2bc5280-acd7-11eb-8c5c-0fea70332372.png">

jffng:
Closed by #3780

PR #3780, opened 2021-05-07
Quadrat: Update Pullquote styles

scruffian:
Updates pullquote styles to match https://github.com/Automattic/themes/issues/3743

Before:
Screenshot 2021-05-07 at 11 18 26

After:
Screenshot 2021-05-07 at 11 18 11

### DIFF
--- a/blank-canvas-blocks/sass/blocks/_pullquote.scss
+++ b/blank-canvas-blocks/sass/blocks/_pullquote.scss
@@ -2,8 +2,6 @@
 .wp-block-pullquote {
 	text-align: var(--wp--custom--pullquote--typography--text-align);
 	blockquote {
-		padding: 0;
-		margin: 0;
 		p {
 			font-size: 1em;
 			padding: 0;
@@ -25,5 +23,12 @@
 	&.is-style-solid-color {
 		background-color: var(--wp--custom--color--foreground);
 		color: var(--wp--custom--color--background);
+	}
+}
+
+.wp-block-pullquote:not(.is-style-solid-color) {
+	blockquote {
+		padding: 0;
+		margin: 0;
 	}
 }
--- a/blank-canvas-blocks/sass/blocks/_pullquote.scss
+++ b/blank-canvas-blocks/sass/blocks/_pullquote.scss
@@ -2,6 +2,8 @@
 .wp-block-pullquote {
 	text-align: var(--wp--custom--pullquote--typography--text-align);
 	blockquote {
+		padding: 0;
+		margin: 0;
 		p {
 			font-size: 1em;
 			padding: 0;
@@ -26,9 +28,3 @@
 	}
 }
 
-.wp-block-pullquote:not(.is-style-solid-color) {
-	blockquote {
-		padding: 0;
-		margin: 0;
-	}
-}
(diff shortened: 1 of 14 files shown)


blank-canvas-blocks/sass/blocks/_pullquote.scss
```diff
@@ -17,6 +17,7 @@
 			font-size: var(--wp--custom--pullquote--citation--typography--font-size);
 			font-family: var(--wp--custom--pullquote--citation--typography--font-family);
 			font-style: var(--wp--custom--pullquote--citation--typography--font-style);
+			font-weight: var(--wp--custom--pullquote--citation--typography--font-weight);
```
MaggieCabrera, line 20: Does this break for BCB and the other children that don't have this defined on their theme.json files? Maybe we should default it on BCB's theme.json to whatever the editor/browser has by default

pbking: It appears to fail gracefully (the missing variable just seems to roll the property back to the browser-defined value of 400.

However it does seem dirty that BCB's CSS would use a variable that it doesn't express.  I suggest that we default that to 400 in BCB.

Additionally this scenario is EXACTLY what I'm excited for by styling elements of blocks in theme.json.  Once that expands beyond anchors and paragraphs we'll be able to remove that ponyfill from BCB entirely.

pbking:
The lack of left/right padding makes the colored-background-variant seem crowded to me:

image

Should it get a css-driven exception?

scruffian:
I pushed a change to BCB to only reset the padding/margin for the blockquote element when you don't choose a solid style.

blank-canvas-blocks/sass/blocks/_pullquote.scss
```diff
@@ -26,3 +25,10 @@
 		color: var(--wp--custom--color--background);
 	}
 }
+
+.wp-block-pullquote:not(.is-style-solid-color) {
```
pbking: Is that how we would expect Gutenberg to deal with the scenario?  It feels theme-specific to me and thus no business of BCB's.  Another theme might have a preference for the blockquote paddings to be the same for both and there would be no way to do that.
Or I guess they could with CSS.  But as a polyfill to work like we expect Gutenberg to work what is expected?

danieldudzic: @pbking I agree with this being too opinionated for BCB. How about this approach? #3790

scruffian: These rules are already in BCB, this change just makes them less powerful. I'm happy to remove them from BCB but we need to check the impact on the other child themes if we do.

pbking: I actually meant something more like [this change](https://github.com/Automattic/themes/pull/3780/commits/9e64988e7071053924415f91e2545cc4dfaedd30), where BCB is relatively ignorant of any deviations in style from regular to filled-background.  The [exception is handled](https://github.com/Automattic/themes/pull/3780/files#diff-cb4e625b42338af0b0a25a9effda1fc71e03ada4a095e25c4886f4de88e86726R1) in the child theme with a simple style declaration:

pbking:
Oops!  Looks like you fixed up one of my rebase remessups.  Thanks! 

This looks ready to 🚢  to me.

jffng:
I made some additional refinements to the font size and style:

- Uses a variable font size (same as the "Huge" font-size preset) so that it scales down on mobile
- Line height to 1.4 — 1.3 felt a bit too tight

What do you think @beafialho ?

Desktop | Mobile
------- | ------
Screen Shot 2021-05-11 at 12 49 51 PM | Screen Shot 2021-05-11 at 12 49 37 PM

beafialho:
It looks good @jffng, thanks.

Issue #3694, opened 2021-04-23
Quadrat: code quality


jffng:
This issue is to capture a few code quality to-dos:

- [x] Break block patterns into separate PHP files
- [x] Refactor to use `gutenberg_block_template_part`
- [x] Refactor theme.json files to have variables on alphabetical order
- [ ] Identify areas that can be migrated to BCB eventually and add comments

MaggieCabrera:
The utility classes can be moved to BCB when we refactor the block patterns.

MaggieCabrera:
Refactor theme.json files to have variables on alphabetical order

MaggieCabrera:
I'm going to remove this one from the v1 milestone. I already moved the utility classes to BCB but I don't know if there's anything else we potentially want to move to BCB so that's why I'm not closing this.

MaggieCabrera:
I'm going to close this as the last point should be an ongoing thing

PR #3781, opened 2021-05-07
BCB + Children: reorder theme json

MaggieCabrera:

Reordered theme.json files to alphabetical order. Rebuilt all 3 child themes.

#### Related issue(s):

Partially addresses #3694

### DIFF
--- a/blank-canvas-blocks/experimental-theme.json
+++ b/blank-canvas-blocks/experimental-theme.json
@@ -300,15 +300,6 @@
 		}
 	},
 	"styles": {
-		"color": {
-			"background": "var(--wp--custom--color--background)",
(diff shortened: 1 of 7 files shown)


pbking:
Is this something that can/should be scripted in BCB?  Part of the build process?  A pre-commit hook?

MaggieCabrera:
mmh, I did this manually, I'm not sure if we want to automate this cause we may want to ignore the alphabetical order in some cases (I did ignore it for left, right, top, bottom rules for example). Should be enough to just try to keep it up.

pbking (approved):
I used a machine to ensure the contents of the before/after JSON files is identical.  👍 
I ensured each theme loaded as expected.
🚢

Issue #3788, opened 2021-05-07
Quadrat: Incorrect spacing below posts. 


kjellr:
To match the comps, we should have 150px of spacing below the post content and the post pagination, and then again before the page footer. 

Current Dev|Recommended
---|---
<img width="1265" alt="Screen Shot 2021-05-07 at 1 42 46 PM" src="https://user-images.githubusercontent.com/1202812/117488481-77ee4500-af3a-11eb-9027-a6da87ced008.png">|<img width="1265" alt="Screen Shot 2021-05-07 at 1 44 15 PM" src="https://user-images.githubusercontent.com/1202812/117488486-791f7200-af3a-11eb-898e-469817ed3aa2.png">

Issue #3784, opened 2021-05-07
Quadrat: Post header details are not in sync with the mockups


kjellr:
We need to take another pass at this.

- The vertical spacing is pretty far off.
- The title should be wide-width.
- The post categories are not shown alongside the date. 
- The post title font size isn't in sync with that of the comps.
- The font weight of the post date should be normal (400), not 300. 

Mockup|Development
---|---
![one](https://user-images.githubusercontent.com/1202812/117485512-702ca180-af36-11eb-9ead-bda5d1f704f9.jpg)|![two](https://user-images.githubusercontent.com/1202812/117485513-70c53800-af36-11eb-8115-ab183094109a.jpg)

PR #3801, opened 2021-05-10
Quadrat: added polygons to header and footer

MaggieCabrera:

This PR adds the squares from the design to the header and the footer. I tried using the group block as a control for the user to enable/disable/customize the color of the squares but it's not possible, since the background color is in the group block and we don't have access to a variable (in case the user uses a custom color). We could on the cover block because the color was actually on the before pseudoelement.

We could consider providing two different header/footer options for the user to use with or without the squares. I don't believe we can make a block style of the header/footer though.

I used clip-path instead of the square svgs because it was simpler to handle the overflow when the element was bigger than the viewport. The shapes are done "by eye" so they may need adjusting, but they are responsive and maintain aspect ratio no matter the screen size.

On big screens or posts with very little content, the top square and the bottom square are very close to one another. I'd love input from @beafialho and @kjellr about that. We could tweak with media queries so the proportions of the squares are different on some viewports (like forcing a max size of the squares for the bigger screens) or adapt the shapes so this issue is a little less prominent.

I've been headbutting a bit against this PR, so any new eyes that can point out other ways of achieving this or improving it are more than welcome.

Design:

Screenshot 2021-05-10 at 14 53 56

Header | Footer
--- | ---
ezgif-6-dc2d4b2b5081 | ezgif-6-35f211ea3934

### DIFF
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -431,9 +431,22 @@ a:active, a:focus {
 	z-index: -1;
 }
 
+@media (max-width: 479px) {
+	.site-header:before {
+		-webkit-clip-path: polygon(0 0, 100vw 0, 100vw 50vw, 52vw 83vw, 0 20vw);
+		        clip-path: polygon(0 0, 100vw 0, 100vw 50vw, 52vw 83vw, 0 20vw);
+	}
+}
+
 .post-meta {
 	align-items: center;
 	justify-content: center;
+}
+
+@media (max-width: 479px) {
+	.post-meta {
+		padding-top: 120px !important;
+	}
 }
 
 .post-meta .wp-block-post-date::before {
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -445,8 +445,8 @@ a:active, a:focus {
 
 @media (min-width: 960px) {
 	.site-header:before {
-		-webkit-clip-path: polygon(13vw 0, 100vw 0, 100vw 28vw, 52vw 63vw, 9vw 3vw);
-		        clip-path: polygon(13vw 0, 100vw 0, 100vw 28vw, 52vw 63vw, 9vw 3vw);
+		-webkit-clip-path: polygon(20vw 0, 96vw 0, 100vw 7vw, 100vw 16vw, 50vw 46vw);
+		        clip-path: polygon(20vw 0, 96vw 0, 100vw 7vw, 100vw 16vw, 50vw 46vw);
 	}
 }
 
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -445,8 +445,8 @@ a:active, a:focus {
 
 @media (min-width: 960px) {
 	.site-header:before {
-		-webkit-clip-path: polygon(13vw 0, 95vw 0, 100vw 7vw, 100vw 16vw, 50vw 51vw);
-		        clip-path: polygon(13vw 0, 95vw 0, 100vw 7vw, 100vw 16vw, 50vw 51vw);
+		-webkit-clip-path: polygon(13vw 0, 100vw 0, 100vw 16vw, 50vw 51vw);
+		        clip-path: polygon(13vw 0, 100vw 0, 100vw 16vw, 50vw 51vw);
 	}
 }
 
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -367,7 +367,13 @@ a:active, a:focus {
 	background: var(--wp--custom--color--secondary);
 }
 
-.site-header > .wp-block-group {
+.site-header {
+	position: relative;
+}
+
+.site-header .wp-block-group {
+	display: flex;
+	justify-content: space-between;
 	align-items: center;
 	display: flex;
 	flex-wrap: wrap;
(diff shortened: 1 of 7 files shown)


kjellr:
Thanks for working through this! We'll give it a test. 

> On big screens or posts with very little content, the top square and the bottom square are very close to one another. I'd love input from @beafialho and @kjellr about that. We could tweak with media queries so the proportions of the squares are different on some viewports (like forcing a max size of the squares for the bigger screens) or adapt the shapes so this issue is a little less prominent.

I think we'll need to hold off until https://github.com/Automattic/themes/issues/3784 and https://github.com/Automattic/themes/issues/3788 are resolved before we can really assess the proportions.

kjellr:
This is looking good! I pushed a quick change to sync up with the header and footer changes. I'm _pretty_ sure I fixed the conflicts correctly, but feel free to revert if I missed something. 

Here's how this looks: 

Mobile|Tablet|Desktop
---|---|---
Screen Shot 2021-05-14 at 11 30 33|Screen Shot 2021-05-14 at 11 30 40|Screen Shot 2021-05-14 at 11 30 13

It's pretty tough to get it looking right on every single breakpoint. I think in general, the goal would be that on tablet and desktop breakpoints, the top rectangle would just slightly overlap the intro text of an article. On mobile, having it take up about half of the viewport seems right. Do you think we can adjust to try aiming for that?

> We could tweak with media queries so the proportions of the squares are different on some viewports (like forcing a max size of the squares for the bigger screens) or adapt the shapes so this issue is a little less prominent.

Yeah, this is a tough problem, but maybe the answer is simple... maybe entry-content has a min-height of 500px or something? That might end up looking really weird, but it's worth a try?

One additional note is that I think it would be cool for non-site-editor users to turn this on or off in the customizer. But we can follow up with that separately if need be.

MaggieCabrera:
> This is looking good! I pushed a quick change to sync up with the header and footer changes. I'm _pretty_ sure I fixed the conflicts correctly, but feel free to revert if I missed something.

Yeah it looks fine to me, thank you.

> On mobile, having it take up about half of the viewport seems right. Do you think we can adjust to try aiming for that?

mmh, I can do that but it will probably mean chaging the clip-path a bit to move it more to the left border of the viewport. Could you do a small mock up so I can copy it the best I can?

> Yeah, this is a tough problem, but maybe the answer is simple... maybe entry-content has a min-height of 500px or something? That might end up looking really weird, but it's worth a try?

It's definitely simpler to change that than play with the clip-path :D

> One additional note is that I think it would be cool for non-site-editor users to turn this on or off in the customizer. But we can follow up with that separately if need be.

The problem with that is how do we carry that customization to the FSE when that's available, I don't think we have tools to do that at the moment.

kjellr:
> > I think in general, the goal would be that on tablet and desktop breakpoints, the top rectangle would just slightly overlap the intro text of an article. On mobile, having it take up about half of the viewport seems right.
> 
> Could you do a small mock up so I can copy it the best I can?

Sure, here's what I'm thinking: 

Desktop|Tablet|Mobile
---|---|---
Polygon Placement (Desktop)|Polygon Placement (Tablet)|Polygon Placement (Mobile)

I'm less concerned about the footer placement — we can take another look at those after we adjust the headers. 

@beafialho do these mockups look alright to you?

MaggieCabrera:
This is what it looks like now. I only adapted the distance between the date/tags and the header for the mobile version:

Desktop | Tablet | Mobile
--- | --- | ---
Screenshot 2021-05-14 at 15 11 09 | Screenshot 2021-05-14 at 15 10 53 | Screenshot 2021-05-14 at 15 10 10

kjellr:
Cool! The mobile and tablet ones are looking good so far. 🌟

MaggieCabrera:
The desktop version adapted to the new design:

Screenshot 2021-05-17 at 09 54 20

beafialho:
> @beafialho do these mockups look alright to you?

These are looking good in general! 

I was wondering if there's a specific reason this square isn't reaching to the corner of the viewport on Desktop?

Captura de ecrã 2021-05-17, às 09 53 32

If it's possible, might be better to have it reach the corner, as shown here:

Captura de ecrã 2021-05-17, às 09 57 23

MaggieCabrera:
> These are looking good in general!
> 
> I was wondering if there's a specific reason this square isn't reaching to the corner of the viewport on Desktop?
> 

Yeah, I just copied Kjell's new design from this comment, that gap is in there: https://github.com/Automattic/themes/pull/3801#issuecomment-841193933

kjellr:
> I was wondering if there's a specific reason this square isn't reaching to the corner of the viewport on Desktop?

Yeah, I liked that little detail, since it clarifies the full shape of the polygon. But I can see how it might look like a mistake. Happy to remove that detail if you'd prefer, @beafialho.

beafialho:
> But I can see how it might look like a mistake. Happy to remove that detail if you'd prefer, @beafialho.

That's what it looked like to me. I see your point, but I think it should look similar to Mobile and Tablet.

kjellr:
Cool, let's expand the shape so that it doesn't leave that little detail then. 👍

MaggieCabrera:
@beafialho @kjellr I rebased this and changed the corner of the polygon on desktop. This is what it looks like with the latest changes to the header and footer spacing:

Screenshot_2021-05-18 Screenshot

kjellr (approved):
This feels good! Let's merge it in.

PR #3803, opened 2021-05-10
Balasana: Update screenshot

kjellr:
We received a report that the original image was triggering for folks aware of the [Bikram Yoga history](https://en.wikipedia.org/wiki/Bikram_Yoga#Sexual_abuse_by_Choudhury). Replacing with an alternate.


ianstewart (approved):
LGTM.

Issue #3747, opened 2021-05-04
Quadrat: Column Block Margins


beafialho:
When I set an Image block to full width, it's ok that it touches the edges of the screen:

<img width="1536" alt="Captura de ecrã 2021-05-04, às 15 20 36" src="https://user-images.githubusercontent.com/65220155/117018662-9eb73c00-acec-11eb-9045-e086ad1e1a52.png">

However, when it's a Column Block, it has no margins, which is visually unpolished and gets misaligned with the navigation on top. This doesn't happen on the Cover Block (white text), where there is padding by default.

<img width="1531" alt="Captura de ecrã 2021-05-04, às 15 32 18" src="https://user-images.githubusercontent.com/65220155/117020435-5731af80-acee-11eb-9f1e-b46db3c44b4b.png">

Is it possible to align the columns with the site title and menu?

I'm not sure I described it well, but I'm thinking of users who wouldn't contemplate alignment details.

PR #3816, opened 2021-05-10
Quadrat: Add alignments for image cover and video

scruffian:
In Quadrat we want a gap around the site, but we want image, cover and video blocks to extend into it:
Screenshot 2021-05-10 at 20 53 12

All other blocks should have space around them.

#### Related issue(s):
https://github.com/Automattic/themes/issues/3747

### DIFF
--- a/blank-canvas-blocks/experimental-theme.json
+++ b/blank-canvas-blocks/experimental-theme.json
@@ -129,6 +129,10 @@
 			},
 			"margin": {
 				"baseline": "10px",
+				"bodyHorizontal": "0px",
+				"bodyVertical": "0px",
+				"bodyHorizontalDesktop": "0px",
+				"bodyVerticalDesktop": "0px",
 				"horizontal": "30px",
 				"vertical": "30px"
 			},
--- a/blank-canvas-blocks/sass/base/_alignment.scss
+++ b/blank-canvas-blocks/sass/base/_alignment.scss
@@ -5,7 +5,7 @@
 }
 
 @include break-mobile {
-	// limit size of any element that is aligned left/right 
+	// limit size of any element that is aligned left/right
 	.wp-block[data-align="left"], // This is for the editor
 	.wp-block[data-align="right"], // This is for the editor
 	.wp-site-blocks .alignleft,
@@ -15,9 +15,9 @@
 }
 
 // When content is aligned left/right (particularly inside of a container) it is floated left/right
-// and needs something to ensure that the content follows the block rather than nestling up beside the floated element.  
+// and needs something to ensure that the content follows the block rather than nestling up beside the floated element.
 // The issue should be resolved upstream: https://github.com/WordPress/gutenberg/issues/10299
-.wp-block-group:not(.site-header) { 
+.wp-block-group:not(.site-header) {
 	overflow: auto;
 }
 
@@ -26,4 +26,19 @@
 // class which would do this for us.  I'm not sure why but this centers things appropriately.
 .aligncenter {
 	text-align: center;
+}
+
+// These blocks go right to the edge of the viewport
+.wp-block-image,
+.wp-block-cover,
+.wp-block-video {
+	&.alignfull {
+		margin: 0 calc(-1 * var(--wp--custom--margin--body-horizontal));
+		width: calc(100% + var(--wp--custom--margin--body-horizontal)*2);
+
+		@include break-mobile {
+			margin: 0 calc(-1 * var(--wp--custom--margin--body-horizontal-desktop));
+			width: calc(100% + var(--wp--custom--margin--body-horizontal-desktop)*2);
+		}
+	}
 }
--- a/blank-canvas-blocks/sass/elements/_body.scss
+++ b/blank-canvas-blocks/sass/elements/_body.scss
@@ -1,0 +1,9 @@
+@import '../base/breakpoints';
+
+body {
+	padding: var(--wp--custom--margin--body-vertical) var(--wp--custom--margin--body-horizontal);
+
+	@include break-mobile {
+		padding: var(--wp--custom--margin--body-vertical-desktop) var(--wp--custom--margin--body-horizontal-desktop);
+	}
+}
--- a/quadrat/experimental-theme.json
+++ b/quadrat/experimental-theme.json
@@ -129,6 +129,10 @@
 			},
 			"margin": {
 				"baseline": "10px",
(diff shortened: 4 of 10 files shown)


quadrat/experimental-theme.json
```diff
@@ -193,8 +197,8 @@
 			},
 			"post-content": {
 				"padding": {
-					"left": "var(--wp--custom--margin--horizontal)",
-					"right": "var(--wp--custom--margin--horizontal)"
+					"left": "0px",
```
scruffian, line 200: Need to look at the impact of this change

jffng:
> All other blocks should have space around them.

I'm not sure I agree. For example if I have a group or columns block with a different background color, should that extend to the full width of the page?

Screen Shot 2021-05-10 at 5 45 28 PM

blank-canvas-blocks/experimental-theme.json
```diff
@@ -129,6 +129,10 @@
 			},
 			"margin": {
 				"baseline": "10px",
+				"bodyHorizontal": "0px",
+				"bodyVertical": "0px",
+				"bodyHorizontalDesktop": "0px",
```
jffng, line 134: I'm concerned that by adding more custom variables within the theme that are media query specific, we are going down a similar path to alignments that won't be compatible with FSE.

scruffian:
> I'm not sure I agree. For example if I have a group or columns block with a different background color, should that extend to the full width of the page?

@beafialho what do you think?

blank-canvas-blocks/experimental-theme.json
scruffian: Do you think that these variables should be set by the Site Editor?

pbking: I rather think they make sense as variables in a theme for re-use, but (IMO) doesn't HAVE to be surfaced to the user (and is unlikely to be a standard way).  So they make sense in a child theme, just not BCB.  I'm not sure that I'm keen for most of this to be in BCB; it all feels very theme-specific and something unlikely to be handled by gutenberg (thus not something to be handled by BCB).

beafialho:
I'd say if a column block has a background color, the best solution is to extend it to the full width of the page. However, if it doesn't have a background color and the columns have text, the problem remains, so I think it should have the margins in this case.

Captura de ecrã 2021-05-12, às 09 35 27

@kjellr what do you think?

kjellr:
That's right. The general rule is: No text should bump up against the edges of the screen.

pbking:
To account for that scenario does it make sense to just add a padding to the columns (and group) if it's wide width (background color or not?)

```
.wp-block-columns.alignfull {
	padding: var(--wp--custom--margin--vertical) var(--wp--custom--margin--horizontal);
}
```

kjellr:
It gets more complicated than that, because if for example there's a full-width image inside of a columns or group block, we _do_ want it to extend to the edges. It's only text that we want to add padding to.

scruffian:
> It gets more complicated than that, because if for example there's a full-width image inside of a columns or group block, we do want it to extend to the edges. It's only text that we want to add padding to.

Are you sure that's a good idea? @beafialho said it was better not to do that with the gallery block because of the spaces between the images:
Screenshot 2021-05-20 at 17 50 42

kjellr:
It's only text and the gallery block that we want to add padding to then. 😄 

The priority is text though: when that bumps up against the edges it looks like a bug. When the gallery does, it's more of a design choice.

blank-canvas-blocks/experimental-theme.json
pbking: Since we have a standard method of describing padding perhaps it would be better to define this just like any other element.  And to take a note from the button hover state refactor, perhaps could also express the mobile 'state' in the same way.  i.e.
```
		"custom": {
			"alignment": {
				"alignedMaxWidth": "50%"
			},
			"body": {
				"spacing": {
					"padding": {
						"top": "0px",
						"bottom": "0px",
						"left": "0px",
						"right": "0px"
					}
				},
				"mobile": {
					"spacing": {
						"padding": {
							"top": "0px",
							"bottom": "0px",
							"left": "0px",
							"right": "0px"
						}
					}
				}
			},
```

It's more verbose but follows the convention of Gutenberg.

blank-canvas-blocks/sass/elements/_body.scss
```diff
@@ -0,0 +1,9 @@
+@import '../base/breakpoints';
+
+body {
+	padding: var(--wp--custom--margin--body-vertical) var(--wp--custom--margin--body-horizontal);
+
+	@include break-mobile {
+		padding: var(--wp--custom--margin--body-vertical-desktop) var(--wp--custom--margin--body-horizontal-desktop);
+	}
```
pbking, line 8: Leveraging and then redefining the variables here would allow for other places that use the body padding variables to do so only once (the value of the variable would change based on the breakpoint)

```
body {
	padding-top: var(--wp--custom--body--padding--top);
	padding-bottom: var(--wp--custom--body--padding--bottom);
	padding-left: var(--wp--custom--body--padding--left);
	padding-right: var(--wp--custom--body--padding--right);

	@include break-mobile-only {
		--wp--custom--body--padding--top: var(--wp--custom--body--mobile--padding--top);		
		--wp--custom--body--padding--bottom: var(--wp--custom--body--mobile--padding--bottom);		
		--wp--custom--body--padding--left: var(--wp--custom--body--mobile--padding--left);		
		--wp--custom--body--padding--right: var(--wp--custom--body--mobile--padding--right);		
	}
}
```

blank-canvas-blocks/sass/base/_alignment.scss
```diff
@@ -26,4 +26,19 @@
...
+.wp-block-cover,
+.wp-block-video {
+	&.alignfull {
+		margin: 0 calc(-1 * var(--wp--custom--margin--body-horizontal));
+		width: calc(100% + var(--wp--custom--margin--body-horizontal)*2);
+
+		@include break-mobile {
+			margin: 0 calc(-1 * var(--wp--custom--margin--body-horizontal-desktop));
+			width: calc(100% + var(--wp--custom--margin--body-horizontal-desktop)*2);
+		}
```
pbking, line 42: Redefining variables (as noted below in _body.scss) would eliminate the need for additional breakpoints like this.  There would just be one calculation, the value of the variable changing depending on the viewport width.

scruffian:
Closing in favour of #3953

Issue #3661, opened 2021-04-15
Quadrat: customize the theme colors


jffng:
We should identify what options are available to users of Quadrat via the Customizer, bearing in mind these will likely be carried through to other small themes.

I'm not suggesting we add support for all (or any!) of these, but I want to at least have the discussion. 

- Colors 
- Excerpt settings
- Hide the page title on the home page

Do you have thoughts about these options? Can you think of any more?

cc @kjellr @beafialho

kjellr:
In addition to that list, we should also include the option to hide the header and footer on the homepage, like we do in all the Varia + Seedlet themes.

beafialho:
I agree with the customisation options you both mentioned.

I was wondering if it's possible to give users the option to choose which meta they wish to display. Like this:
<img width="287" alt="Captura de ecrã 2021-04-15, às 17 59 00" src="https://user-images.githubusercontent.com/65220155/114908742-4808c180-9e14-11eb-87ee-e69b4d3c565a.png">

One other thing is to have an option of choosing the blog's layout between 'Grid' or 'Stack', like this:
<img width="289" alt="Captura de ecrã 2021-04-15, às 17 56 54" src="https://user-images.githubusercontent.com/65220155/114908498-11cb4200-9e14-11eb-9a2d-9416a1d795a1.png">

scruffian:
My concern about adding these options is that they won't carry over into the site editor. So any customizations I add here will be lost once the version of WordPress is updated

kjellr:
I agree with that concern. The same goes for the other way around: any adjustments you make in the site editor won't carry through if you were to (somehow) revert back to the customizer. 

Even so, I do think that the color control at least is very necessary — especially since the theme has such an opinionated purple + pink palette by default. Maybe we can start with that?

scruffian:
I think going in the reverse direction is less of a concern - that will only happen through user intervention, whereas with auto updates, most users will end up in the Site Editor without doing anything.

I think colors is a good idea. @pbking do you think we can use some of your prior art to hook into theme.json?

beafialho:
Good points. I agree, color control is necessary here.

pbking:
I have SO MUCH prior art for that!
Yes, I can build a thing...

jffng:
I've updated this issue to reflect the primary concern for customization — being able to change theme colors via the customizer or FSE.

PR #3817, opened 2021-05-10
Adds Custom Color Controls to Customizer for Quadrat

pbking:
Accomplishes the color customization of #3661 

This uses the 'overwrite CSS variables' method to customize values.

These customizations are ONLY available in the Customizer.

The things to be customized are defined in the theme.json's 'custom_meta' block.  

The 'Customizer Bridge' presents controls to the user based on that JSON and renders the stored values as Global Styles CSS variables.

The user DOES NOT have access to the selected colors via the block editor panel.  

image

image

### DIFF
--- a/quadrat/child-experimental-theme.json
+++ b/quadrat/child-experimental-theme.json
@@ -51,7 +51,7 @@
 				"secondary": "var(--wp--preset--color--darker-blue)",
 				"foreground": "var(--wp--preset--color--pink)",
 				"background": "var(--wp--preset--color--blue)",
-				"selection": "var(--wp--preset--color--darker-blue)"
+				"selection": "var(--wp--custom--color--secondary)"
 			},
 			"fontsToLoadFromGoogle": [
 				"family=Poppins:ital,wght@0,300;0,400;0,500;0,600;1,400"
@@ -105,6 +105,40 @@
 				}
 			}
 		},
+		"custom_meta": [
+			{
+				"name": "Colors",
+				"type": "section",
+				"slug": "colors",
+				"description": "Color Customization for Quadrat",
+				"controls": [
+					{
+						"name": "Foreground Color",
+						"type": "color",
+						"default": "#FFD1D1",
+						"slug": "color--foreground"
+					},
+					{
+						"name": "Background Color",
+						"type": "color",
+						"default": "#292C6D",
+						"slug": "color--background"
+					},
+					{
+						"name": "Primary Color",
+						"type": "color",
+						"default": "#FFD1D1",
+						"slug": "color--primary"
+					},
+					{
+						"name": "Secondary Color",
+						"type": "color",
+						"default": "#151853",
+						"slug": "color--secondary"
+					}
+				]
+			}	
+		],
 		"layout": {
 			"contentSize": "664px",
 			"wideSize": "1128px"
--- a/quadrat/functions.php
+++ b/quadrat/functions.php
@@ -49,4 +49,4 @@ require get_stylesheet_directory() . '/inc/block-styles.php';
 /**
  * Customizer Bridge
  */
-require get_template_directory() . '/inc/customization.php';
+require get_stylesheet_directory() . '/inc/customization.php';
--- a/quadrat/functions.php
+++ b/quadrat/functions.php
@@ -31,8 +31,6 @@ endif;
 function quadrat_scripts() {
 	// Enqueue front-end styles.
 	wp_enqueue_style( 'quadrat-styles', get_stylesheet_directory_uri() . '/assets/theme.css', array( 'blank_canvas_blocks-ponyfill' ), wp_get_theme()->get( 'Version' ) );
-	// Enqueue customizer bridge styles
-	wp_enqueue_style( 'customizer-style', get_template_directory_uri() . '/assets/css/noop.css', array( 'quadrat-styles' ), wp_get_theme()->get( 'Version' ) );
 }
 add_action( 'wp_enqueue_scripts', 'quadrat_scripts' );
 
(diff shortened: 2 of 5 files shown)


quadrat/inc/customization.php
```diff
@@ -0,0 +1,102 @@
...
+	function customizer_bridge_output_variables() {
+		wp_add_inline_style( 'customizer-style', $this->customizer_bridge_generate_custom_color_variables() );
+	}
+
+	/**
+	 * Render the customized variables.
+	 */
+	function customizer_bridge_generate_custom_color_variables( $context = null ) {
+
+		$css_variables = 'body {';
```
MaggieCabrera: targetting just the body is not enough, this doesn't work on the post editor

MaggieCabrera:
This works fine in the customizer and the frontend but not on the editor, and we are also missing the new colors from the block options (they are showing the old palette)

quadrat/child-experimental-theme.json
```diff
@@ -105,6 +105,40 @@
 				}
 			}
 		},
+		"custom_meta": [
```
MaggieCabrera, line 108: this is kind of a nitpick but does it make sense to have something like 

```
"custom": {
    "meta":{ ...
```

having custom and custom_meta is kind of strange

quadrat/functions.php
```diff
@@ -31,6 +31,8 @@ function quadrat_support() {
 function quadrat_scripts() {
 	// Enqueue front-end styles.
 	wp_enqueue_style( 'quadrat-styles', get_stylesheet_directory_uri() . '/assets/theme.css', array( 'blank_canvas_blocks-ponyfill' ), wp_get_theme()->get( 'Version' ) );
+	// Enqueue customizer bridge styles
+	wp_enqueue_style( 'customizer-style', get_template_directory_uri() . '/assets/css/noop.css', array( 'quadrat-styles' ), wp_get_theme()->get( 'Version' ) );
```
MaggieCabrera: I'm curious, why do we need this?

quadrat/inc/customization.php
```diff
@@ -0,0 +1,102 @@
+<?php
+class CustomizerBridge {
+
+	private $theme_customizations;
+
+	function __construct() {
+		add_action( 'customize_register', array( $this, 'customizer_bridge_register' ) );
+		add_action( 'wp_enqueue_scripts', array( $this, 'customizer_bridge_output_variables' ) );
```
MaggieCabrera, line 10: this only works for the frontend, for the editor I think we need `enqueue_block_editor_assets`

quadrat/inc/customization.php
```diff
@@ -0,0 +1,102 @@
...
+
+	/**
+	 * Render the customized variables.
+	 */
+	function customizer_bridge_generate_custom_color_variables( $context = null ) {
+
+		$css_variables = 'body {';
+		foreach ( $this->theme_customizations as $custom_section ) {
+			foreach ( $custom_section->controls as $custom_option ) {
+				$css_variables = $css_variables . '--wp--custom--' . $custom_option->slug . ':' . get_theme_mod( 'customizer-bridge-' . $custom_option->slug ) . ';';
```
MaggieCabrera: if we don't select all 4 custom colors this can return something like `--wp--custom--color--secondary:;`

quadrat/child-experimental-theme.json
pbking: It does but then all of the 'meta' information renders as css variables and that didn't seem appropriate either.  Putting stuff outside of the 'custom' block prevented that.  I'm good with renaming it, but I hesitate to move it into "custom".

quadrat/functions.php
pbking: The customizer puts inline CSS into the HTML file.  But (as far as I've been able to figure out) it only does that [targeting an existing stylesheet](https://github.com/Automattic/themes/pull/3817/files#diff-aa202940a1e08444050cfbe4d440df8b20a1540b6f377d6d6caee50ef43dc36fR22).  This is that stylesheet and is the line that determines what dependencies the CSS has and where in the page its rendered.

quadrat/child-experimental-theme.json
MaggieCabrera: oh yeah I totally forgot! makes total sense then

quadrat/inc/customization.php
pbking: Hrm...  We may need to target the body AND root:...

[Originally](https://github.com/WordPress/theme-experiments/pull/209/files#diff-df9b569afcaa8f63490652d68e97e19a4b73eab5e1f08f2e8a44e434768aaa1dR30) this targeted only `root:` and that worked ok.  But between the original experiment and today the majority of Global Styles (on the view) are applied to the BODY not ROOT.  Thus applying these variables to ROOT they are simply overwritten by the global styles applied to the body (no matter the load order).

quadrat/inc/customization.php
pbking: Hrm..  So you're saying we need something defensive there to ensure we aren't rendering variables the user hasn't assigned.  Yea that makes sense...

quadrat/inc/customization.php
MaggieCabrera: yeah, old themes target both for the color annotations to work (Seedlet as an example) but I tried and it didn't work, even when I enqueued the css to show on the editor, I never saw the styles showing up at all.

quadrat/inc/customization.php
pbking: Added: https://github.com/Automattic/themes/pull/3817/commits/6e1e690836230db933d74104f58a59dec54993a1

quadrat/inc/customization.php
pbking: Good call.  Thanks for helping me fix this. 👍

quadrat/inc/customization.php
pbking: We got it to work targeting #editor as well as body.

pbking:
Still not convinced this is the right approach but now I believe that all of the bugs and notes noted have been resolved except for what is noted in the description.

jffng:
Nice exploration @pbking — could we close or remove this from the milestone in favor of #3856?

pbking:
Yes, absolutely.  It was a worthwhile experiment. :)

PR #3821, opened 2021-05-10
Block themes: Update templates to include a Query block wrapper

danieldudzic:
WordPress/gutenberg#30804

I was able to observe the Alignment control for the Query block on the BCB Search template but not on the Index - not sure, what that is about.


MaggieCabrera:
Sadly this breaks the alignments for the query pagination block on Quadrat, I'm trying to confirm if it's a bug or not, will come back to this when I find out.

MaggieCabrera:
So this PR is fine but needs https://github.com/WordPress/gutenberg/pull/31833 in order to land since it would break Quadrat's query pagination alignment otherwise. Once that's merged, it's good to go.

MaggieCabrera (approved):
This is ok to merge now!

PR #3824, opened 2021-05-11
Quadrat: revise post header and meta

### LINKED ISSUE #3784: Quadrat: Post header details are not in sync with the mockups

### LINKED ISSUE #3786: Quadrat: Default font weight is incorrect.
Most of the site's text is currently set to a 300 weight, but it should be Normal/400 to match the Figma file. 

- Navigation
- Body text
- Post pagination
- Page footer
- etc.

Current Dev|Comps
---|---
Screen Shot 2021-05-07 at 1 31 07 PM|Screen Shot 2021-05-07 at 1 32 27 PM

### LINKED ISSUE #3787: Quadrat: Post editor title does not match the front end
Currently, the text is bold, left-aligned, and the same width as the text column itself. To match the comps, it should be a 400 weight, centered, wide-width, and have a 1.2 line height. 

We should tackle https://github.com/Automattic/themes/issues/3784 first though, and then try to sync up with that. 

Editor|Front End
---|---
Screen Shot 2021-05-07 at 1 37 14 PM|Screen Shot 2021-05-07 at 1 37 27 PM

jffng:
This PR addresses #3784.

Before | After
-------- | -------
Screen Shot 2021-05-10 at 8 32 56 PM | Screen Shot 2021-05-10 at 8 31 59 PM

One thing I wasn't able to address is changing the divider between categories. This is baked into the block at the moment.

### DIFF
--- a/quadrat/block-template-parts/single.html
+++ b/quadrat/block-template-parts/single.html
@@ -9,7 +9,7 @@
 </div>
 <!-- /wp:group -->
 
-<!-- wp:group {"layout":{"inherit":true},"style":{"spacing":{"padding":{"top":"30px","right":"20px","bottom":"30px","left":"20px"}}}} -->
+<!-- wp:group {"layout":{"inherit":true},"style":{"spacing":{"padding":{"top":"30px","right":"20px","bottom":"60px","left":"20px"}}}} -->
 <div class="wp-block-group" style="padding-top:30px;padding-right:20px;padding-bottom:30px;padding-left:20px">
 	<!-- wp:post-title {"textAlign":"center","level":1,"align":"wide"} /-->
 </div>
--- a/quadrat/block-template-parts/single.html
+++ b/quadrat/block-template-parts/single.html
@@ -18,7 +18,7 @@
 <!-- wp:post-featured-image {"align":"full"} /-->
 
 <!-- wp:spacer {"height":60} -->
-<div style="height:100px" aria-hidden="true" class="wp-block-spacer"></div>
+<div style="height:60px" aria-hidden="true" class="wp-block-spacer"></div>
 <!-- /wp:spacer -->
 
 <!-- wp:post-content {"layout":{"inherit":true}} /-->
--- a/quadrat/block-template-parts/single.html
+++ b/quadrat/block-template-parts/single.html
@@ -1,14 +1,26 @@
-<!-- wp:post-date {"textAlign":"center","fontSize":"tiny"} /-->
-<!-- wp:post-tags {"textAlign":"center","fontSize":"tiny"} /-->
+<!-- wp:spacer -->
+<div style="height:170px" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- /wp:spacer -->
+
+<!-- wp:group {"className":"post-meta"} -->
+<div class="wp-block-group post-meta">
+	<!-- wp:post-date {"textAlign":"center","fontSize":"tiny"} /-->
+	· 
+	<!-- wp:post-terms {"term":"category","textAlign":"center","fontSize":"tiny"} /-->
+</div>
+<!-- /wp:group -->
 
 <!-- wp:group {"layout":{"inherit":true},"style":{"spacing":{"padding":{"top":"30px","right":"20px","bottom":"30px","left":"20px"}}}} -->
 <div class="wp-block-group" style="padding-top:30px;padding-right:20px;padding-bottom:30px;padding-left:20px">
-
-	<!-- wp:post-title {"textAlign":"center","level":1} /-->
+	<!-- wp:post-title {"textAlign":"center","level":1,"align":"wide"} /-->
 </div>
 <!-- /wp:group -->
 
 <!-- wp:post-featured-image {"align":"full"} /-->
+
+<!-- wp:spacer -->
+<div style="height:100px" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- /wp:spacer -->
 
 <!-- wp:post-content {"layout":{"inherit":true}} /-->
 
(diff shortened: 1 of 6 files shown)


quadrat/block-template-parts/single.html
```diff
@@ -1,15 +1,27 @@
-<!-- wp:post-date {"textAlign":"center","fontSize":"tiny"} /-->
-<!-- wp:post-tags {"textAlign":"center","fontSize":"tiny"} /-->
+<!-- wp:spacer -->
+<div style="height:170px" aria-hidden="true" class="wp-block-spacer"></div>
```
MaggieCabrera, line 2: is it better to use padding on the group block? even [margin](https://github.com/WordPress/gutenberg/pull/30608) now that we can do that? I'm writing this but I think I prefer spacers when the distance is so prominent so feel free to ignore

quadrat/child-experimental-theme.json
```diff
@@ -199,9 +199,9 @@
 			},
 			"core/post-title": {
 				"typography": {
-					"fontSize": "min(max(36px, 5vw), 65px)",
+					"fontSize": "min(max(48px, 7vw), 80px)",
```
MaggieCabrera: this is the same we are using for h1, should we make a variable?

MaggieCabrera (approved):
This looks good to me in terms of code

kjellr:
> One thing I wasn't able to address is changing the divider between categories. This is baked into the block at the moment.

That's weird! Can we adjust using CSS in the meantime?

jffng:
@kjellr 

> Can we adjust using CSS in the meantime?

It looks like the divider is output in raw HTML (no wrapping or pseudo element), so I can't target / change its contents with CSS:

Screen Shot 2021-05-11 at 10 01 13 AM

I will look at submitting an improvement in Gutenberg (EDIT: opened an issue here https://github.com/WordPress/gutenberg/issues/31710). In the meantime, is the rest of this PR okay to you?

I also just realized the way I'm adding the dot in between the date isn't good because it will always display, even if the post has no categories. So I will change that. 

----

quadrat/block-template-parts/single.html
jffng: Yeah, for some reason using spacers _feels_ a bit nicer when the spacing is large and opinionated like this. It makes the design more explicit via blocks.

quadrat/child-experimental-theme.json
jffng: Ohh, perhaps. I will look at adding one.

kjellr:
While that issue gets sorted out, can you do something like this pseudocode to hide the pipes?

```
.wp-block-post-terms {
 color: var(--background);
}

.wp-block-post-terms a {
 color: var(--foreground);
}
```

jffng:
Okay, here's what the post header is like now: 

Desktop | Mobile
----- | ------
Screen Shot 2021-05-11 at 12 08 13 PM | Screen Shot 2021-05-11 at 12 09 33 PM

(Header + mobile menu styling will be addressed separately)

kjellr:
This looks good, but since folks won't be able to adjust the size until FSE is available, I wonder if we should increase it? The logo is pretty small today: 

Now (35px):
Screen Shot 2021-05-11 at 1 10 22 PM
Screen Shot 2021-05-11 at 1 10 29 PM


Suggested (64px): 
Screen Shot 2021-05-11 at 1 11 14 PM
Screen Shot 2021-05-11 at 1 12 36 PM

@beafialho, what do you think?

quadrat/block-template-parts/single.html
```diff
@@ -1,15 +1,26 @@
...
+<!-- /wp:spacer -->
+
+<!-- wp:group {"className":"post-meta"} -->
+<div class="wp-block-group post-meta">
+	<!-- wp:post-date {"textAlign":"center","fontSize":"tiny"} /-->
+	<!-- wp:post-terms {"term":"category","textAlign":"center","fontSize":"tiny"} /-->
+</div>
+<!-- /wp:group -->
 
 <!-- wp:group {"layout":{"inherit":true},"style":{"spacing":{"padding":{"top":"30px","right":"20px","bottom":"30px","left":"20px"}}}} -->
```
kjellr:
```suggestion
<!-- wp:group {"layout":{"inherit":true},"style":{"spacing":{"padding":{"top":"30px","right":"20px","bottom":"60px","left":"20px"}}}} -->
```

quadrat/block-template-parts/single.html
```diff
@@ -1,15 +1,26 @@
...
 <div class="wp-block-group" style="padding-top:30px;padding-right:20px;padding-bottom:30px;padding-left:20px">
-
-	<!-- wp:post-title {"textAlign":"center","level":1} /-->
+	<!-- wp:post-title {"textAlign":"center","level":1,"align":"wide"} /-->
 </div>
 <!-- /wp:group -->
 
 <!-- wp:post-featured-image {"align":"full"} /-->
 
+<!-- wp:spacer -->
```
kjellr:
```suggestion
<!-- wp:spacer {"height":60} -->
```

quadrat/block-template-parts/single.html
```diff
@@ -1,15 +1,26 @@
...
+<div style="height:100px" aria-hidden="true" class="wp-block-spacer"></div>
```
kjellr:
```suggestion
<div style="height:60px" aria-hidden="true" class="wp-block-spacer"></div>
```

kjellr (approved):
Added some small adjustments to even things out. Once those are in place, I think this is good to merge. We may want to make some adjustments on mobile somehow, since some of the spacing seems extra-large there. But we can reassess in a separate PR. 

Screen Shot 2021-05-11 at 1 25 13 PM

PR #3826, opened 2021-05-11
Quadrat Body font size fix

### LINKED ISSUE #3785: Quadrat: Body text size is incorrect.
As per the figma file, the default body size should be 18px, with a line height of 1.6. Instead, the current font size is 20px, with a line height of 1.7. 

Comps|Development
---|---
Frame 2|current

While we're making this change, I think a line height of 1.65 or 1.7 might actually be a bit more comfortable. @beafialho do you have a preference there? Either way, the font size definitely should be updated.

MaggieCabrera:

This PR sets the body font size to 18px for Quadrat (the line height is already 1.7).

#### Related issue(s):

Fixes #3785

### DIFF
--- a/quadrat/child-experimental-theme.json
+++ b/quadrat/child-experimental-theme.json
@@ -279,7 +279,7 @@
 			}
 		},
 		"typography": {
-			"fontSize": "20px",
+			"fontSize": "18px",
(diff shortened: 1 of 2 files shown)


scruffian (approved):
LGTM

PR #3828, opened 2021-05-11
Quadrat: cover block pattern hover bug

### LINKED ISSUE #3738: Quadrat: "Cover Block with Heading" pattern bugs
Followup to #3676. Two small observations on the Cover Block with Heading pattern: 

1. To match the comps, the default background color should be "Blue" instead of "Dark Blue".
2. I'm seeing some weirdness on select in the editor: The square gets darker, and the text color is different. See the GIF below: 

test

MaggieCabrera:

This PR changes the default color for the cover pattern to blue from dark blue and fixes the weird bug that was changing the color of the text when the cover is selected in the editor.

#### Related issue(s):

Fixes #3738

### DIFF
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -70,6 +70,7 @@
 	left: 0;
 	bottom: 0;
 	right: 0;
+	z-index: 0 !important;
 }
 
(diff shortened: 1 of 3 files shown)


scruffian:
LGTM. Shame about the !important but I can't see a way round that.

PR #3830, opened 2021-05-11
Dalston: Separate script and style enqueues for better performance

sixhours:

* Use separate functions with separate hooks for enqueuing styles and scripts in Dalston to avoid unnecessary enqueues on the front end.
* Props to @westonruter for the bug report and original fix in #1652 

This doesn't appear to break anything in my testing, but I could use a Theam review to confirm things are still working as expected. I'm not sure what the block customizations are supposed to add/change based on the code. :)

#### Testing instructions

* Check out this PR theme to your local install, or sync it to your sandbox, and activate Dalston.
* Check the source and ensure the scripts and styles located in functions.php are being enqueued in the appropriate locations:
* Everything in `dalston_block_extends_styles()` should be enqueued on both the front end and the block editor
* Everything in `dalston_block_extends()` should only be enqueued in the block editor

### DIFF
--- a/dalston/functions.php
+++ b/dalston/functions.php
@@ -216,7 +216,7 @@ function dalston_editor_styles() {
 add_action( 'enqueue_block_editor_assets', 'dalston_editor_styles' );
 
 /**
- * Enqueue Custom Cover Block Styles and Scripts
+ * Enqueue Custom Cover Block Scripts
  */
 function dalston_block_extends() {
 
@@ -232,11 +232,6 @@ function dalston_block_extends() {
 		array( 'wp-blocks' )
 	);
 
-	wp_enqueue_style(
-		'dalston-extend-cover-block-style',
-		get_stylesheet_directory_uri() . '/block-extends/extend-cover-block.css'
-	);
-
 	// Columns Block Tweaks
 	wp_enqueue_script(
 		'dalston-extend-columns-block',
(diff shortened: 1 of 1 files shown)


dalston/functions.php
```diff
@@ -232,36 +232,48 @@ function dalston_block_extends() {
...
+/**
+ * Enqueue Custom Cover Block Styles
+ */
+function dalston_block_extends_styles() {
+
+	// Bail out early while in AMP endpoint.
+	if ( dalston_is_amp() ) {
+		return;
+	}
+
```
westonruter: This shouldn't be needed as it is only relevant when enqueuing scripts, since custom scripts enqueued in this way are not AMP-compatible. Stylesheets on the other hand _are_ compatible, so they should be retained on AMP to ensure proper styling.

```suggestion
```

westonruter (changes requested):

dalston/functions.php
sixhours: Thank you!! Updated :)

kwight:
@sixhours How are we testing these?..

sixhours:
> @sixhours&nbsp;How are we testing these?..

Oops! Added some testing instructions. :)

kwight:
Great! I synced this branch to my sandbox and switched to Dalston, and found the stylesheets only on the front-end, and both the stylesheets and the scripts loaded in the editor.

kwight (approved):
:shipit:

sixhours:
Synced in r59024-wpcom-themes

PR #3835, opened 2021-05-11
Styled calendar block per design

### LINKED ISSUE #3809: Quadrat: Calendar block is unreadable by default
Screen Shot 2021-05-10 at 12 52 33 PM

We should change the grays text colors here to pink, and change the gray background and borders to match the transparent black color we used when building the cover block variation in #3676.

pbking:
image

This leverages primary/secondary colors instead of black and blend modes.  I think the effect should be the same but toss it back if it's not good.

### DIFF
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -59,6 +59,20 @@
 	padding: calc(0.667em + 2px) calc(1.333em + 2px);
 }
 
+.wp-block-calendar table caption {
+	color: var(--wp--custom--color--primary);
+}
+
+.wp-block-calendar table th {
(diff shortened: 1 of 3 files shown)


jffng (commented):
The compiled CSS looks fine to me, but is the sass file not checked into this PR?

pbking:
Oops!  Well that was silly!  Added.  Should be ready to go.

jffng:
LGTM. As an aside, the functionality of this block feels dated / limited. I have a hard time imagining a situation where I would want to use it. Unless I could control / filter what kinds of post it links to — e.g. if it could inherit from the query block and only show posts of a certain type, eg a CPT of events...

PR #3856, opened 2021-05-14
Quadrat: Add color customization

### LINKED ISSUE #3661: Quadrat: customize the theme colors

scruffian:
This is a different approach to https://github.com/Automattic/themes/pull/3817.

The general approach is to replicate the functionality of Global Styles in the Customizer. This means that we modify the custom post type, so that the data can be shared between both environments.

To simplify this I have also removed the custom background and custom foreground colors and implemented new variables for these, which can be referenced in the CSS.

I have also added a preview to the Customizer which just injects some CSS to target the relevant selectors and also modifies the variables.

The code is written to be extendable so it should be trivial to add support for custom link colors too.

### DIFF
--- a/blank-canvas-blocks/functions.php
+++ b/blank-canvas-blocks/functions.php
@@ -88,3 +88,8 @@ function blank_canvas_blocks_fonts_url() {
 if ( ! is_child_theme() ) {
 	require get_template_directory() . '/inc/block-patterns.php';
 }
+
+/**
+ * Add Global Styles Variables
+ */
+require get_template_directory() . '/inc/global-styles-variables.php';
--- a/quadrat/inc/customization.php
+++ b/quadrat/inc/customization.php
@@ -5,7 +5,13 @@ class CustomizerBridge {
 
 	function __construct() {
 		add_action( 'customize_register', array( $this, 'customizer_bridge_register' ) );
+
+		//this is for the view
 		add_action( 'wp_enqueue_scripts', array( $this, 'customizer_bridge_output_variables' ) );
+
+		//this is for the editor
+		add_action( 'enqueue_block_editor_assets', array( $this, 'customizer_bridge_output_variables' ) );
+
 		$this->theme_customizations = $this->get_theme_customizations();
 	}
 
@@ -19,15 +25,20 @@ class CustomizerBridge {
 	 * Enqueue color variables for customizer & frontend.
 	 */
 	function customizer_bridge_output_variables() {
+		wp_register_style( 'customizer-style', false, array(), true, true );
+		wp_enqueue_style( 'customizer-style' );
 		wp_add_inline_style( 'customizer-style', $this->customizer_bridge_generate_custom_color_variables() );
 	}
 
 	/**
 	 * Render the customized variables.
+	 *
+	 * NOTE: Ideally this wouldn't happen at all and instead the theme.json would be filtered
+	 * and modified to reflect the users's choice.
 	 */
 	function customizer_bridge_generate_custom_color_variables( $context = null ) {
 
-		$css_variables = 'body {';
+		$css_variables = 'body, #editor {';
 		foreach ( $this->theme_customizations as $custom_section ) {
 			foreach ( $custom_section->controls as $custom_option ) {
 				$css_variables = $css_variables . '--wp--custom--' . $custom_option->slug . ':' . get_theme_mod( 'customizer-bridge-' . $custom_option->slug ) . ';';
--- a/quadrat/inc/customization.php
+++ b/quadrat/inc/customization.php
@@ -1,144 +1,84 @@
 <?php
-class CustomizerBridge {
 
-	private $theme_customizations;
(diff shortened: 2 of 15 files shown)


pbking:
> I'm not convinced that we should save the background/foreground in custom - I think they should probably write to the same place as they do with Global Styles.

I wholeheartedly agree.  That makes those two properties "universal" so it's a solid step.

pbking:
Oh, man this PR has been a LOT of fun to tinker with.  Thank you so much!  A couple of things I now realize:

This CPT IS where the global styles store user data.  I see that now and it would be trivial (ish) to pull/push those values from `styles.color.text` and `styles.color.background` such that the SAME property is viewable/editable in both the customizer and the FSE.  Fantastic.

After tinkering with this I also don't believe that we should be storing any of these customization in theme_mod but instead rely completely on the CPT for storing the user's preference.

The "merging" conversation regarding `custom` I'm still wrestling with but I'm (currently, at the time of writing) agreeing with you that it seems the theme and user 'custom' blocks should be merged rather than replaced.

pbking:
Currently it seems that custom variables aren't coming through (at least that's my guess)

The site title size isn't correct and all of the button styles have gone away.  I suspect there's more and that it's all related to the same something.

image

I thought it might have been from a previous incarnation of user customized theme.json values so to be sure I recreated the environment and have the same result.

From what I can tell it's the call to `get_merged_data` that's causing the theme custom attributes to barf:
```
		$all           = WP_Theme_JSON_Resolver::get_merged_data( $settings, 'theme' );
```

jffng:
I can confirm seeing the same as @pbking.

scruffian:
Thanks for testing. I've now fixed that issue.

pbking:
I've banged around a lot on this.  Love the real-time preview.  I haven't been able to get it to do anything other than what I expected so in that I think it's pretty solid.

I'm super-tickled to have a mechanism by which to manipulate the user-controlled theme.json!  Really appreciate you making this work for real!

To finalize this we should still account for more colors beyond just foreground/background, and that those colors should also be stored in the CPT in the "custom" space.  I realize that won't be editable in the FSE but as color management tooling it just not complete for what Quadrat needs.  

I've seen some usage of color picking tooling that included custom palette colors.  Would it be possible to add that in  (and to expand Quadrat's descriptive palette to include some of the alternate colors in the figma)?

blank-canvas-blocks/functions.php
```diff
@@ -88,3 +88,8 @@ function blank_canvas_blocks_fonts_url() {
 if ( ! is_child_theme() ) {
 	require get_template_directory() . '/inc/block-patterns.php';
 }
+
+/**
+ * Add Global Styles Variables
+ */
+require get_template_directory() . '/inc/global-styles-variables.php';
```
pbking, line 95: I would be keen for this (and related) changes (moving foreground/background variable consumption from custom to generated variables) into trunk sooner than later (even without the color changing bits).  Definitely a good change.

scruffian: Yeah I agree. I'll make a PR for this.

jffng:
I poked around a bit and if you remove [this condition](https://github.com/WordPress/gutenberg/blob/298f50847b315e7dca622c8d2cacd0628ee7cc67/lib/global-styles.php#L104) from global styles, you get your color customizations in the post editor too!

scruffian:
> I poked around a bit and if you remove this condition from global styles, you get your color customizations in the post editor too!

Is that still necessary if we add a file in block-templates/index.html?

jffng:
No but if we add a `block-templates/index.html`, we get the site editor instead of the customizer, until there is theme support / theme.json setting for disabling it.

quadrat/inc/customization.php
```diff
@@ -0,0 +1,127 @@
...
+		$this->custom_colors = array(
+			'name'        => __( 'Colors' ),
+			'type'        => 'section',
+			'slug'        => 'customize-global-styles',
+			'description' => __( 'Color Customization for Quadrat' ),
+			'controls'    => array(
+				array(
+					'name'     => __( 'Foreground Color' ),
+					'type'     => 'color',
+					'default'  => $this->get_color( $theme_json['styles']['color']['text'], $color_palette ),
```
jffng: I just ran into an issue:

1. I change the foreground and background colors on this PR
2. Switched to a different PR, the color customizations persisted
3. Switched back to this PR to change the colors back to default
4. Clicking default in the color picker didn't work as expected — instead it just returned them to whatever custom color they were last set to.

Do you know why that would happen?

scruffian:
Closing in favour of https://github.com/Automattic/themes/pull/4090

PR #3857, opened 2021-05-14
Quadrat: fix site title and post title styles

jffng:
This PR fixes some styling on the site title block in the header and post title blocks that appear in a query loop.

Before | After
------- | -------
Screen Shot 2021-05-14 at 2 08 32 PM | Screen Shot 2021-05-14 at 2 07 19 PM


kjellr (approved):
🚢

PR #3866, opened 2021-05-17
BCB + children: removed color and border customizations from selects

### LINKED ISSUE #3810: Quadrat: Form dropdowns are unreadable by default
To demonstrate, here's the archives block, with the "Show as dropdown" option active: 

Screen Shot 2021-05-10 at 12 53 11 PM

@beafialho do you have a design for how these should look?

MaggieCabrera:

This PR removes border and color styles from the selects to improve legibility, since some browsers add a background that clashes with certain color schemes. 

#### Related issue(s):

Fixes https://github.com/Automattic/themes/issues/3810

### DIFF
--- a/blank-canvas-blocks/assets/ponyfill.css
+++ b/blank-canvas-blocks/assets/ponyfill.css
@@ -173,8 +173,6 @@ textarea:focus {
 }
 
 select {
-	color: var(--wp--custom--colors--foreground);
-	border: var(--wp--custom--form--border--width) solid var(--wp--custom--form--border--color);
 	font-family: inherit;
 	font-size: 100%;
 }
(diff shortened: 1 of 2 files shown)


jffng (approved):
This LGTM. To confirm, this is what a default dropdown looks like now in Quadrat:

Screen Shot 2021-05-17 at 11 18 00 AM

MaggieCabrera:
> This LGTM. To confirm, this is what a default dropdown looks like now in Quadrat:
> 
> Screen Shot 2021-05-17 at 11 18 00 AM

It will depend on the browser :D which is the point of the PR, since making a properly styled version that is uniform for all browser is kind of complicated (there's discussion about this on the related issue)

PR #3869, opened 2021-05-17
Replace Poppins font with DM Sans

### LINKED ISSUE #3867: Quadrat: Try changing the font from Poppins to DM Sans
I know we're a little late in the process for this, but @beafialho and I have been chatting about updating the default font here from Poppins to DM Sans. It has a similar feel, but is more refined overall. I'll open a PR to give this a try before we make a final decision. 

Poppins|DM Sans
---|---
Poppins|DM Sans

kjellr:
Closes https://github.com/Automattic/themes/issues/3867. 

Replaces Poppins with DM Sans, and adjusts the font weights to match. 

Before|After
---|---
Screen Shot 2021-05-18 at 10 47 49|Screen Shot 2021-05-18 at 10 48 14

Before|After
---|---
Screen Shot 2021-05-18 at 10 45 21|Screen Shot 2021-05-18 at 10 45 01


kjellr:
What is that `child-experimental-theme` file? Does that need to be updated as well? I noticed it has mention of Poppins in there.

MaggieCabrera:
> What is that `child-experimental-theme` file? Does that need to be updated as well? I noticed it has mention of Poppins in there.

The way block based themes work they don't inherit their parent's json file. To actually have that inheritance we made a script that actually merges the json file from BCB with the one from the child theme (`child-experimental-theme`) and generates the experimental-theme.json file that is actually being loaded by the FSE. 

Short version: we edit the "child" json, then run `npm run build` to generate the combination of that with the parent's theme.

kjellr:
Ok cool, I can update this PR to do that then. 👍

kjellr:
Ok, this should be ready for review again. Please note that I also shrunk down the text size for the header so it matches out other typical text sizes. 

Before|After
---|---
Screen Shot 2021-05-18 at 11 24 22|Screen Shot 2021-05-18 at 11 22 15

Before|After
---|---
Screen Shot 2021-05-18 at 11 22 25|Screen Shot 2021-05-18 at 11 24 38

jffng (approved):
LGTM, digging the replacement!

beafialho:
It's looking good!

PR #3870, opened 2021-05-17
BCB + children: use nav block responsive control

jffng:
* Removes BCB's mobile navigation hacky bits.
* Tells the rendered navigation block to render responsively.  

[This change](https://github.com/WordPress/gutenberg/pull/31911) is needed so that the navigation block can be configured to support responsive mobile stylings.
A fix for [this issue](https://github.com/WordPress/gutenberg/issues/31909) is necessary to get mobile navigation stylings to work as expected. ([fix here](https://github.com/WordPress/gutenberg/pull/31917)) (fixed)

With BOTH of those fixes in place these changes work to style the navigation block's mobile styling.  **This change shouldn't come in until fix #31917 has landed.**  Quadrat will not work as expected until [#31917](https://github.com/WordPress/gutenberg/pull/31917) lands.

NOTE: [This bug](https://github.com/WordPress/gutenberg/issues/31957) was found and logged while working on this.  It's irritating but not a blocker to bring this change in.

In addition to the above this work this change:
* Positions the mobile menu button in a more expected location
* Styles the mobile menu to match foreground/background semantic colors
* Styles the submenus (as core/navigation-link block styling)

#### Screenshots

BCB:
image
image

Seedlet-Blocks:
image
image

Mayland-Blocks:
image
image

Quadrat:
image
image
(known issue)


pbking:
This is ready for evaluation (though you'll have to merge two branches into a local Gutenberg instance to evaluate currently) but isn't ready to bring in until those (or at least one of the items) lands.  I would appreciate some eyes on the change though.

blank-canvas-blocks/block-template-parts/navigation.html
```diff
@@ -1,4 +1,4 @@
-<!-- wp:navigation {"orientation":"horizontal","itemsJustification":"center"} -->
+<!-- wp:navigation {"orientation":"horizontal","itemsJustification":"center","isResponsive":true} -->
```
jffng, line 1: Should we update Seedlet + Mayland's header template parts to include the responsive attribute?

Also, those template parts come with some embedded nav links which I think we should remove so the nav block appears in its empty state by default.

jffng:
I get some weirdness when I try to close the menu, but it might just be trying to test this in a browser rather than actually on mobile:

Kapture 2021-05-21 at 10 25 48

In any case, if it is an issue it should be reported / worked on in the nav block. 

Other than that and my other comment about the template parts, this PR LGTM. We can address Quadrat mobile styling in a separate PR.

blank-canvas-blocks/block-template-parts/navigation.html
pbking: Yes to both, I'll push a change.

pbking:
> I get some weirdness when I try to close the menu 

[Already reported ](https://github.com/WordPress/gutenberg/issues/31957)and noted in the description.

pbking:
Moved the nav stuff from BCB into Quadrat as backup.

This is ready for another look-see.

Suggest any further theme-specific mobile-only changes happen in followups.

jffng:
Tested and verified that Quadrat's mobile menu is working as it was before this PR. 

Sadly I can't give the official green check since I opened the PR. But here it is in spirit ✅

pbking:
Hah, well I can so I think with our powers combined we squash and merge this thing!

pbking (approved):
🚢

PR #3871, opened 2021-05-17
Align the featured image to center, not full.

### LINKED ISSUE #3805: Quadrat: Featured Image doesn't extend full-width if the image is small
When you insert a full-width image into a post, it extends to the edge of the screen all the time, even if the image needs to be scaled up to do so. 

This isn't the case with our full-width featured images though. They max-out at their intrinsic width, and appear left-aligned: 

Screen Shot 2021-05-10 at 12 28 38 PM

To reproduce: 

- Add a small-ish featured image.
- View the post in a large window. 

(This one may end up being a Gutenberg bug)

jffng:
Fixes #3805.

Before | After
------- | ------
Screen Shot 2021-05-17 at 12 24 29 PM | Screen Shot 2021-05-17 at 12 24 10 PM

### DIFF
--- a/quadrat/block-template-parts/single.html
+++ b/quadrat/block-template-parts/single.html
@@ -15,7 +15,7 @@
 </div>
 <!-- /wp:group -->
 
-<!-- wp:post-featured-image {"align":"full"} /-->
+<!-- wp:post-featured-image {"align":"center"} /-->
 
 <!-- wp:spacer {"height":60} -->
 <div style="height:60px" aria-hidden="true" class="wp-block-spacer"></div>


kjellr:
I don't think this fixes it — we'd like our featured images to be full-width, but we also want them to actually extend to the edges of the screen. Right now when they're `alignfull`, they don't go all the way to the edges like an `alignfull` image would.

jffng:
@kjellr 

This PR does allow the featured image to extend to edges of the screen, so long as its intrinsic width is greater than or equal to the window size:

Screen Shot 2021-05-17 at 4 56 48 PM

Are you suggesting that the featured image should always stretch to fullscreen, even if its width is smaller than the browser viewport?

kjellr:
> Are you suggesting that the featured image should always stretch to fullscreen, even if its width is smaller than the browser viewport?

Yeah, that would match the "alignfull" behavior of all other blocks.

jffng:
Closing this, as it should be addressed upstream.

PR #3873, opened 2021-05-18
Seedlet Blocks: Add Post Comments block styles

### LINKED ISSUE #3582: Seedlet Blocks: Style Comment Block
I don't believe this was ever done for BCB.

Current:
image

Target:
image

danieldudzic:
A tentative attempt at updating Seedlet Blocks Post Comments block styles

Before:
Before-2

After:
After

#### Related issue(s):
#3582

### DIFF
--- a/seedlet-blocks/assets/theme.css
+++ b/seedlet-blocks/assets/theme.css
@@ -164,28 +164,12 @@
 
 .wp-block-post-comments .reply a {
 	border-radius: var(--wp--custom--button--border--radius);
-	font-size: var(--wp--preset--font-size--tiny);
-	line-height: 1;
+	font-size: var(--wp--preset--font-size--small);
+	line-height: 1.5;
 	padding: 23px 25px;
 }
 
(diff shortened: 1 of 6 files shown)


MaggieCabrera (commented):
Overall this looks pretty solid! thank you so much for having a go at this. A few comments:

- I had to manually add `<!-- wp:post-comments /-->` to the template in order to see the comments. I think this would be a good PR to add that to the theme.
- Looks like there's a couple of differences between Seedlet and Seedlet blocks that stand a little to me. One is that the links are green instead of black (but I think that's a general bug for the theme, maybe out of the scope of this PR) and the other would be the order and style of the form elements. I think we can use [what Ben did for Quadrat](https://github.com/Automattic/themes/blob/ef4b0ad34bbdca11124ecb537a305173e0307a74/quadrat/sass/blocks/_post-comments.scss#L11) and move it into BCB, since I think both Seedlet blocks and Mayland blocks will benefit from that.

Seedlet | Seedlet Blocks
--- | ---
Screenshot 2021-05-18 at 12 35 02 | Screenshot 2021-05-18 at 12 35 17

danieldudzic:
| Seedlet      | Seedlet Blocks |
| ----------- | ----------- |
| Seedlet      | seedlet_blocks      |

MaggieCabrera:
The styling looks great! I think we are only missing the block on the template so that the comments show up on the frontend automatically and this would be good to go

MaggieCabrera (approved):
LGTM!

Issue #3874, opened 2021-05-18
Seedlet: Improve copy/text shown when user is 'Showing Latest Posts' on front page but has no public posts to show.


jordesign:
As part of the effort in https://github.com/Automattic/wp-calypso/issues/52806 to improve the default state for WPCOM sites when user has 'Showing Latest Posts' on front page but has no public posts.

### Steps to reproduce the behavior

1. Make sure your site has no public posts (no posts, or all posts set to be drafts)
2. Set Customizer > Home Page Settings to 'Show Latest Posts'

### What I expected to happen

As a user I would expect to see any posts I've written (even if drafts) or some kind of content.

### What actually happened

While in the Customizer I see some example content - I presume in order to help preview any changes. But the user has no idea where this content has come from.

<img width="1713" alt="118056654-aa31f380-b3cd-11eb-8612-5cd56ac51ad0" src="https://user-images.githubusercontent.com/1842363/118593537-38054880-b7eb-11eb-9a8c-be2fb067d7dd.png">

When previewing the front end of the site - text is shown which is a little confusing - stating 'Nothing Found'. This can be very confusing for users who either:

* Have draft posts created (they think they have lost them)
* Have created a home page - but don't have it set to show

<img width="1526" alt="Screen Shot 2021-05-13 at 9 38 08 am" src="https://user-images.githubusercontent.com/1842363/118593601-52d7bd00-b7eb-11eb-9bf4-097a425d8978.png">
<img width="1352" alt="Screen Shot 2021-05-13 at 9 29 37 am" src="https://user-images.githubusercontent.com/1842363/118593675-6e42c800-b7eb-11eb-9080-0e74937c5ab0.png">

### What would be better?

<img width="1084" alt="Screen Shot 2021-05-13 at 9 44 15 am" src="https://user-images.githubusercontent.com/1842363/118593591-4fdccc80-b7eb-11eb-876d-37ae28b8cc26.png">

jordesign:
Closing this as it is resolved in https://github.com/Automattic/themes/pull/3875

PR #3875, opened 2021-05-18
Seedlet: Update text and link when there are no published posts

jordesign:
Addressing #3874

This PR changes the content shown when a user has 'Latest Posts' selected for their front page but no public posts. This affects only the logged-in view.

### Updated appearance

Screen Shot 2021-05-18 at 3 14 09 pm

## Testing Instructions

1. Ensure site has no public posts (set them to be draft or remove them).
2. Use the Customizer to set the 'Home Page Settings' to show latest posts
3. Visit the home page of the site.

### DIFF
--- a/seedlet/template-parts/content/content-none.php
+++ b/seedlet/template-parts/content/content-none.php
@@ -12,24 +12,29 @@
 
 <section class="no-results not-found">
 	<header class="page-header default-max-width">
-		<h1 class="page-title"><?php _e( 'Nothing Found', 'seedlet' ); ?></h1>
+		<h1 class="page-title"><?php _e( 'No posts published yet!', 'seedlet' ); ?></h1>
 	</header><!-- .page-header -->
 
 	<div class="page-content default-max-width">
 		<?php
-		if ( is_home() && current_user_can( 'publish_posts' ) ) :
+		if ( is_home() && current_user_can( 'publish_posts' ) ) : ?>
 
-			printf(
+			<p><?php _e( 'Your site is set to show the the most recent posts on your homepage - but you don\'t have any Posts published.', 'seedlet' ); ?></p>
+
+			<?php printf(
 				'<p>' . wp_kses(
-					/* translators: 1: link to WP admin new post page. */
-					__( 'Ready to publish your first post? <a href="%1$s">Get started here</a>.', 'seedlet' ),
+					/* translators: 1: link to WP admin new post page. 2: class for anchor */
+					__( '<a href="%1$s" class="%2$s">Add or publish Posts</a>', 'seedlet' ),
 					array(
 						'a' => array(
 							'href' => array(),
+							'class' => 'button',
 						),
 					)
 				) . '</p>',
-				esc_url( admin_url( 'post-new.php' ) )
+				esc_url( admin_url( 'edit.php' ) ),
+				'button',
+
 			);
 
 		elseif ( is_search() ) :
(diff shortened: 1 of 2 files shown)


kwight:
Great @jordesign – can you include testing instructions please?

jordesign:
Okedoke - have updated with testing instructions - sorry about that.

mreishus:
The translated text used to be in the form `Lorem ipsum <a>dolor sit amet</a>` (tag mixed in with text), but now it's in the form `<a>Lorem ipsum dolor sit amet</a>` (tag strictly surrounding text). Given that, is it more appropriate to take the `<a>` tags out of the translatable text and to move them w/ the other surrounding tag (the `<p>`)?

seedlet/template-parts/content/content-none.php
```diff
@@ -12,24 +12,29 @@
...
 
-			printf(
+			<p><?php _e( 'Your site is set to show the the most recent posts on your homepage - but you don\'t have any Posts published.', 'seedlet' ); ?></p>
+
+			<?php printf(
 				'<p>' . wp_kses(
-					/* translators: 1: link to WP admin new post page. */
-					__( 'Ready to publish your first post? <a href="%1$s">Get started here</a>.', 'seedlet' ),
+					/* translators: 1: link to WP admin new post page. 2: class for anchor */
+					__( '<a href="%1$s" class="%2$s">Add or publish Posts</a>', 'seedlet' ),
```
scruffian: If all the text is in a link then I think we can make this much simpler by moving the translation function inside the link.

jordesign:
Thanks @mreishus and @scruffian - I've just pushed a fix which removes the anchor from the translatable text and (I think) improves the overall formatting of that section.

Still more than open to feedback on how it could be cleaner and tidier.

seedlet/template-parts/content/content-none.php
```diff
@@ -12,27 +12,35 @@
...
 		<?php
-		if ( is_home() && current_user_can( 'publish_posts' ) ) :
+		if ( is_home() && current_user_can( 'publish_posts' ) ) : ?>
 
-			printf(
-				'<p>' . wp_kses(
-					/* translators: 1: link to WP admin new post page. */
-					__( 'Ready to publish your first post? <a href="%1$s">Get started here</a>.', 'seedlet' ),
+			<p>
+			<?php _e( 'Your site is set to show the the most recent posts on your homepage - but you don\'t have any Posts published.', 'seedlet' ); ?></p>
```
scruffian: Very minor, but you could wrap this string with double quotes and then you wouldn't need the `\` on the single quote.

seedlet/template-parts/content/content-none.php
```diff
@@ -12,27 +12,35 @@
...
+
+			<?php printf(
```
scruffian: I think you can remove wp_kses and printf here altogether and just write the button as markup.

jordesign: Oh awesome - thank for these tips @scruffian - I'm finding my way around our themes repo for the first time an more than happy for suggestions of the neatest way to do things but still have them fit our style/methods.

jordesign:
After a little snafu with work on the wrong branch (oops) this is updated and ready for review again. Thanks folks!

scruffian:
This is looking really good. The only other thing that would be good to do is to adhere to the WordPress coding standards. We have a pre-commit hook which can take care of that for you - more details here: https://github.com/Automattic/themes/blob/trunk/README.md#coding-standards

jordesign:
Thanks @scruffian  - Have run it through the pre-commit hook now - and all looks good. Most recent commit should be correct.

scruffian:
Great, thank you, I'll get this merged.

PR #3879, opened 2021-05-18
Theme experiments submodule added to the repo

MaggieCabrera:

If we want to move some of our themes to the theme-experiments repo having said repo as a [submodule](http://git-scm.com/book/en/v2/Git-Tools-Submodules) inside this one would be a good way of avoiding code duplication

This is what it looks like if I make a change in a file in spearhead and in a file in twenty-twenty-blocks:

Screenshot 2021-05-18 at 17 25 28
Screenshot 2021-05-18 at 17 25 40

When pushing using `sandbox.sh push`, only the Spearhead change gets pushed to the sandbox. This means we don't get the whole theme-experiments repo inside the sandbox (which is good) but if we move a theme from this repo to the experiments one, we will need to update our sandboxes manually for that theme.

To test this, if it's the first time you are using a submodule use:

`git submodule update --init --recursive`

if you want to update the existing one:

`git submodule update --recursive --remote`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "theme-experiments"]
+	path = theme-experiments
+	url = git@github.com:WordPress/theme-experiments.git


pbking:
Would it be better to create a repository exclusively for whatever theme we submodule?

I'm not opposed to leveraging a theme outside of our repo as a parent, but I'm not sure I'm keen on it living in a repository labeled "-experiments".  If it's something like a commonly used parent theme perhaps it's own repository is valid.  It might then be easier to refactor the sandbox tooling to include (all of) the changes from said repository.

pbking:
Until such a change also includes an established release cycle and version dependency mechanism for the dependent themes i'm very hesitant to make a change like this regardless of the name of the other repo.

I think this is an interesting experiment but suggest we close it and keep everything in a single repo for the time being.

MaggieCabrera:
yeah I agree, we can revisit this at some point if we ever need it.

PR #3881, opened 2021-05-18
Refactor button hover styles into its own custom object tree

### LINKED ISSUE #3591: Mayland Blocks: Adjust custom theme.json variables
We need to review the custom variables on the theme.json and extend when needed from BCB's defaults. Some are off. An example of this are the spacing values:

Mayland | Mayland Blocks
--- | ---
Screenshot 2021-03-31 at 11 07 45 | Screenshot 2021-03-31 at 11 07 53

### LINKED ISSUE #3811: Quadrat: File block button doesn't match other default button styles
The File block's button appears as rounded: 

Screen Shot 2021-05-10 at 12 53 45 PM

This should be squared off to match the theme's default button styles.

pbking:
A recent change for Quadrat caused non-quadrat button styles to break. Padding values for hover styles were lost.  The below is the hover state of a button in Blank-Canvas-Blocks.

image

While looking into this issue I became very disillusioned by the mechanism by which BCB was handling the styles for the button hover state and took another stab at that.

This is the result.

* Hover styles for the button are a mirror of all button styles, but housed in a "hover" object.  (I tried to make it ':hover' but a css variable like `--wp--button--:hover--color--text` didn't work out very well.) This definition was inspired by the recent "elements" addition to block styling.
* The default calculated padding value was set as the default button padding value and referenced by the button block styles.
* The button hover state styles reference the exact same styles as the button block if the --hover equivalent is absent.

Blank-Canvas-Blocks (of course), Seedlet-Blocks, Mayland-Blocks and Quadrat have been evaluated with this change.

### DIFF
--- a/blank-canvas-blocks/experimental-theme.json
+++ b/blank-canvas-blocks/experimental-theme.json
@@ -57,17 +57,17 @@
 			},
 			"button": {
 				"border": {
-					"color": "var(--wp--custom--button--color--text)",
+					"color": "var(--wp--custom--color--foreground)",
 					"radius": "4px",
 					"style": "solid",
-					"width": "0"
+					"width": "2px"
 				},
 				"spacing": {
 					"padding": {
-						"top": "calc(.667em + 2px)",
-						"bottom": "calc(.667em + 2px)",
-						"left": "calc(1.333em + 2px)",
-						"right": "calc(1.333em + 2px)"
+						"top": "0.667em",
+						"bottom": "0.667em",
+						"left": "1.333em",
+						"right": "1.333em"
 					}
 				},
 				"color": {
@@ -81,9 +81,6 @@
 					"lineHeight": 2
 				},
 				"hover": {
-					"border": {
-						"color": "transparent"
-					},
 					"color": {
 						"background": "#006ba1"
 					}
@@ -317,19 +314,8 @@
 	"styles": {
 		"blocks": {
 			"core/button": {
-				"spacing": {
-					"padding": {
-						"top": "var(--wp--custom--button--spacing--padding--top)",
-						"bottom": "var(--wp--custom--button--spacing--padding--bottom)",
-						"left": "var(--wp--custom--button--spacing--padding--left)",
-						"right": "var(--wp--custom--button--spacing--padding--right)"
-					}
-				},
 				"border": {
-					"color": "var(--wp--custom--button--border--color)",
-					"radius": "var(--wp--custom--button--border--radius)",
-					"style": "var(--wp--custom--button--border--style)",
-					"width": "var(--wp--custom--button--border--width)"
+					"radius": "var(--wp--custom--button--border--radius)"
 				},
 				"color": {
 					"background": "var(--wp--custom--button--color--background)",
--- a/blank-canvas-blocks/experimental-theme.json
+++ b/blank-canvas-blocks/experimental-theme.json
@@ -62,6 +62,15 @@
 					"style": "solid",
 					"width": "2px"
 				},
+				"color": {
+					"background": "var(--wp--custom--color--secondary)",
+					"text": "var(--wp--custom--color--background)"
+				},
+				"hover": {
+					"color": {
+						"background": "#006ba1"
+					}
+				},
 				"spacing": {
 					"padding": {
 						"top": "0.667em",
@@ -70,20 +79,11 @@
 						"right": "1.333em"
 					}
 				},
-				"color": {
-					"background": "var(--wp--custom--color--secondary)",
-					"text": "var(--wp--custom--color--background)"
-				},
 				"typography": {
 					"fontFamily": "var(--wp--preset--font-family--base)",
 					"fontSize": "var(--wp--preset--font-size--normal)",
 					"fontWeight": "normal",
 					"lineHeight": 2
-				},
-				"hover": {
-					"color": {
-						"background": "#006ba1"
-					}
 				}
 			},
 			"code": {
--- a/blank-canvas-blocks/experimental-theme.json
+++ b/blank-canvas-blocks/experimental-theme.json
@@ -58,17 +58,20 @@
 			"button": {
 				"border": {
 					"color": "var(--wp--custom--button--color--text)",
-					"hoverColor": "transparent",
 					"radius": "4px",
 					"style": "solid",
-					"hoverStyle": "solid",
-					"width": "0",
-					"hoverWidth": "0"
+					"width": "0"
+				},
+				"spacing": {
+					"padding": {
+						"top": "calc(.667em + 2px)",
+						"bottom": "calc(.667em + 2px)",
+						"left": "calc(1.333em + 2px)",
+						"right": "calc(1.333em + 2px)"
+					}
 				},
 				"color": {
 					"background": "var(--wp--custom--color--secondary)",
-					"hoverText": "var(--wp--custom--color--background)",
-					"hoverBackground": "#006ba1",
 					"text": "var(--wp--custom--color--background)"
 				},
 				"typography": {
@@ -76,6 +79,14 @@
 					"fontSize": "var(--wp--preset--font-size--normal)",
 					"fontWeight": "normal",
 					"lineHeight": 2
+				},
+				"hover": {
+					"border": {
+						"color": "transparent"
+					},
+					"color": {
+						"background": "#006ba1"
+					}
 				}
 			},
 			"code": {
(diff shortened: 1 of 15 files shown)


seedlet-blocks/sass/blocks/_links.scss
```diff
@@ -15,7 +15,7 @@
  */
 .is-root-container, // Editor class hack.
 .wp-block-post-content {
-	a {
+	a:not(.wp-block-button__link) {
```
pbking: This change was only tangentially related. The anchor style was causing the button (anchor) styles to be black instead of the configured white.

MaggieCabrera:
Screenshot 2021-05-19 at 12 55 07

On the frontend both buttons look the same height, but not on the editor

MaggieCabrera:
I changed the value of `--wp--custom--button--spacing--padding--top` to 100px and I saw this on the editor:

Screen Capture on 2021-05-19 at 12-59-53

and on the frontend:

Screen Capture on 2021-05-19 at 12-59-14

danieldudzic:
This makes a lot of sense.

I'm just wondering whether it would make sense to expand support to cover multiple states:

```
"custom": {
	"button": {
		"state": {
			"hover": {
				"border": {
					"color": "transparent"
				},
				"color": {
					"background": "#006ba1"
				}
			},
			"focus": {
			},
			"active": {
			}
		}
	}
}

```

pbking:
@danieldudzic I think that's a good idea (once we find a use-case for that).  Currently just hover is all we're customizingly styling so I'm keen to stick with that for now.  

My goal is to mirror how I think/hope Gutenberg will allow for state styling and I think that your suggestion continues that idea so yea I like it.

pbking:
This is ready for another review.  The issues @MaggieCabrera raised should be addressed.

Quadrat's custom aspects have been refactored and are a bit more involved but hopefully remain clear.

blank-canvas-blocks/sass/blocks/_button.scss
```diff
@@ -4,36 +4,72 @@
...
+ 	font-weight: var(--wp--custom--button--typography--font-weight);
+	font-family: var(--wp--custom--button--typography--font-family);
+	font-size: var(--wp--custom--button--typography--font-size);
+	line-height: var(--wp--custom--button--typography--line-height);
+	text-decoration: none; // Needed because link styles inside .entry-content add a text decoration
+}
+
+@mixin button-border-styles {
+	border-style: var(--wp--custom--button--border--style);
+	border-color: currentColor;
```
jffng, line 44: Should this be `currentColor` or `var(--wp--custom--button--border--color)`?

blank-canvas-blocks/sass/blocks/_button.scss
```diff
@@ -4,36 +4,72 @@
...
-			border-color: var(--wp--custom--button--border--hover-color);
-			border-style: var(--wp--custom--button--border--hover-style);
-			border-width: var(--wp--custom--button--border--hover-width);
-			padding: calc(.667em + 2px - var(--wp--custom--button--border--hover-width)) calc(1.333em + 2px - var(--wp--custom--button--border--hover-width)); //The padding found on an unmodified Button Block
-			svg {
-				fill: var(--wp--custom--button--color--hover-text);
-			}
+			--wp--custom--button--color--text: var(--wp--custom--button--hover--color--text);
+			--wp--custom--button--color--background: var(--wp--custom--button--hover--color--background);
+			border-color: var(--wp--custom--button--border--color);
```
jffng: Would it make sense for this variable to be set in the `:hover` settings? e.g. 

```suggestion
			border-color: var(--wp--custom--button--hover--color--border);
```

blank-canvas-blocks/sass/blocks/_button.scss
```diff
@@ -43,7 +79,17 @@
...
 	.wp-block-button__link {
 		@include button-hover-styles;
-		text-decoration: none; // Needed because link styles inside .entry-content add a text decoration
+		@include button-main-styles;
+	}
+	&.is-style-outline {
+		&.wp-block-button__link,
+		.wp-block-button__link {
+			--wp--custom--button--color--text: var(--wp--custom--button--border--color);
+			--wp--custom--button--color--background : transparent;
```
jffng: nit 

```suggestion
			--wp--custom--button--color--background: transparent;
```

jffng (commented):
I tested many use cases, including:

- standard + outline buttons in BCB + children
- changing button padding variables via theme.json
- setting custom background colors and text colors on outline and standard buttons
- Fixes #3811

All of these use cases are working as expected! 👏 

I left a couple comments related to how the border color for standard + outline buttons are being set. I don't know if they're right, but I was having trouble following how the variable naming and where / how they end up being applied in CSS.

blank-canvas-blocks/experimental-theme.json
```diff
@@ -57,25 +57,33 @@
...
-					"hoverBackground": "#006ba1",
 					"text": "var(--wp--custom--color--background)"
 				},
 				"typography": {
 					"fontFamily": "var(--wp--preset--font-family--base)",
 					"fontSize": "var(--wp--preset--font-size--normal)",
 					"fontWeight": "normal",
 					"lineHeight": 2
+				},
+				"hover": {
```
scruffian: Let's try to keep properties alphabetised.

pbking:
Cool.  Thanks y'all, I'll take another pass.
It's currently "passing unit tests" (I'm itching to actually write some... another project that's been in my mind for a while) and I can iterate on it to make the variable usage more clear.

blank-canvas-blocks/sass/blocks/_button.scss
pbking: currentColor causes the TEXT and OUTLINE to match when in border styles. i.e. changing the "text color" of a button also changed the outline color.
I'm not sure what the standard or expected behavior was there; trunk wasn't behaving at all to compare against, but that was the behavior I was expecting (relatively naively).

MaggieCabrera: I think it should be currentColor, since that is what the editor is doing too

blank-canvas-blocks/experimental-theme.json
pbking: Good call, thanks.  Reorganized.

blank-canvas-blocks/sass/blocks/_button.scss
pbking: the nit has been picked

blank-canvas-blocks/sass/blocks/_button.scss
pbking: Thank you.  I moved the setting of this to button-color-styles mixin (to set it) and changed it to the --hover value here (to follow the convention of the rest of the hover settings).  This matters if you set a hover.border.color for the 'outline button' instance.

blank-canvas-blocks/sass/blocks/_button.scss
pbking: Setting it to 'currentColor' allows the border to be the same color as the label, which gives the user the ability to change both by changing the 'color' property.  I tried to make that a theme decision (setting the value of --wp--custom--button--border--color to `currentColor`) but then there is no mechanism by which to actually choose a color it should default to.  Doing it this way locks in the opinion that changing the color changes the border color too.  I'm open to suggestions refactoring it, but I think I've tried all the ways I could think of and there were blockers of some kind or another all the other ways I tried to make it behave.

scruffian:
When I hover buttons in the Post Comments block they get bigger:
Screenshot 2021-05-20 at 16 46 37

scruffian:
The file button has opacity:
Screenshot 2021-05-20 at 16 49 40

pbking:
Thank you so much for being thourough!  Those have been addressed and it's ready for another look-see.

jffng (approved):
I pushed a small change (https://github.com/Automattic/themes/pull/3881/commits/10f867db8afaa02c85da56a1f30945afd43f5340) to account for the reply button in the comments of Quadrat being a bit smaller:

Screen Shot 2021-05-20 at 4 21 51 PM

Overall I think this is ready to land.

quadrat/sass/blocks/_post-comments.scss
```diff
@@ -1,7 +1,9 @@
+// @import '../../../blank-canvas-blocks/sass/blocks/button';
```
pbking: commented import.

MaggieCabrera:
This PR is going to affect [this other one](https://github.com/Automattic/themes/pull/3905) about the comments block so I'm holding that one until this one lands

pbking:
Also addresses https://github.com/Automattic/themes/issues/3591

MaggieCabrera (approved):

PR #3883, opened 2021-05-18
Add/bcb all children build script

pbking:
Adds a script to build all BCB children.

Useful when making changes to BCB (especially changes to theme.json)

The change also includes changes to child theme.json based on running the tool.

### DIFF
--- a/blank-canvas-blocks/package.json
+++ b/blank-canvas-blocks/package.json
@@ -35,8 +35,9 @@
   "scripts": {
     "start": "chokidar \"sass/**/*.scss\" -c \"npm run build\" --initial",
     "start:child": "run-p start \"build:child -- {@} watch\" --",
+    "start:all": "chokidar \"sass/**/*.scss\" \"experimental-theme.json\" -c \"npm run build:all\" --initial",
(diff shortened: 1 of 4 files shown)


scruffian:
Thanks for doing this. What do you think about renaming the command to "all",  and then also rebuilding BCB?

pbking:
Yea, that's a good idea.

pbking:
Added `build:all` and `start:all`

Also removed the `npm install` from each build run (it made it less useful in practice since I had so much longer to wait for changes in the child themes to get built)

scruffian:
That start:all job isn't working for me. I tried to fix it but I ended up down a big rabbit hole!

pbking:
Are you getting an error, is it not building correctly or is it not triggering on changes you expect?  It's working fine for me but happy to help fix it if I can get the issue reproduced.

I'm finding `start:all` to be the most helpful command. :)

scruffian:
It doesn't trigger when I changes files in the child theme. Have you tried that?

pbking:
Oh, no the purpose was only if things change in BCB, then it builds stuff in the child themes.

I was using it in tandem with the child theme's build (running both).  Changes to BCB changes all children.  Changes to the child only rebuild the child.

Not saying your attempt isn't a valid thing, just that wasn't the tools intent when I threw it together.

scruffian:
Ah ok, makes sense. Let's merge this and iterate :)

Issue #3878, opened 2021-05-18
Quadrat: margins on mobile


MaggieCabrera:
It looks like on mobile all the content goes straight to the edge of the viewport:

<img width="389" alt="Screenshot 2021-05-18 at 17 12 38" src="https://user-images.githubusercontent.com/3593343/118677202-5b84bf00-b7fc-11eb-9fae-301ab9dfe8c9.png">

scruffian:
This was fixed by #3961

PR #3886, opened 2021-05-19
Quadrat: fix mobile padding on query block

jffng:
This PR is an attempt at #3878.

Kapture 2021-05-19 at 13 54 40

This is a prototype to demonstrate what I think the ideal behavior is, but I'm not sure about the approach using a media query. Do we think this is a use case Gutenberg alignment system should account for? If so, what kind of controls should be provided?

### DIFF
--- a/quadrat/sass/blocks/_query.scss
+++ b/quadrat/sass/blocks/_query.scss
@@ -3,3 +3,10 @@
 		font-size: var(--wp--custom--heading--h3--font-size);
 	}
 }
+
+// If a query block is a direct child of a container, it some padding on mobile
+div[class^="wp-container-"] > .wp-block-query-loop {
+	@media screen and ( max-width: 704px ){ // This is the default content width + padding
+		padding: 0 var(--wp--custom--margin--horizontal);
+	}
+}
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -263,6 +263,12 @@ ul ul {
 	font-size: var(--wp--custom--heading--h3--font-size);
 }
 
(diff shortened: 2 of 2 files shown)


quadrat/sass/blocks/_query.scss
```diff
@@ -2,4 +2,11 @@
 	.wp-block-post-title {
 		font-size: var(--wp--custom--heading--h3--font-size);
 	}
+}
+
+// If a query block is a direct child of a container, it some padding on mobile
+div[class^="wp-container-"] > .wp-block-query-loop {
+	@media screen and ( max-width: 704px ){ // This is the default content width + padding
```
MaggieCabrera, line 9: This value worries me since in the future the padding might change and this will not be the case any more

MaggieCabrera:
This is a hard one. I want to think that the padding should always be present and the exception should be done for the blocks that we want to extend over to the viewport borders instead of the other way around.

quadrat/sass/blocks/_query.scss
MaggieCabrera: Another issue that comes to mind is that this won't work for pages like 404, search etc where there's no loop

jffng:
> I want to think that the padding should always be present and the exception should be done for the blocks that we want to extend over to the viewport borders instead of the other way around.

I'm not sure of another way to do it, if we want some padding on the site overall, as well as the ability for blocks to actually be full width.

I'm going to close this PR because I think https://github.com/Automattic/themes/pull/3890 gets closer approach, but the concerns here are still valid.

PR #3889, opened 2021-05-20
Quadrat: adapt single post header to fit design for mobile

### LINKED ISSUE #3832: Quadrat: Refine header spacing on mobile
Followup from #3824. On mobile, the new spacing is pretty tall. We should figure out a way to decrease that on smaller screens, OR (even better) to have the header area [take up a minimum of 100% of the viewport on mobile screens](https://cloudup.com/c_5mhzCPzoK) as in @beafialho's [original comps](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/Quadrat?node-id=50%3A157), rather than rely on static pixel values. 

Current state: 

117858924-6370bc00-b25c-11eb-9944-5d7d9848bb65

MaggieCabrera:

To fit better the design for the header on single posts and pages I have added an extra group block to wrap the header portion of the post/page and added a class to manipulate the styles accordingly. This is the best way I could think going about this, but it makes it impossible for the user to change what the mobile version of the single post looks like. I don't think we can do much with the editor when the mobile version and the desktop one are this different.

Screenshot 2021-05-20 at 10 22 32

#### Related issue(s):

Fixes #3832


scruffian:
> it makes it impossible for the user to change what the mobile version of the single post looks like

I think that's a fair compromise to have this design. LGTM

kjellr:
@MaggieCabrera I think this approach interferes with the header when the site logo is active. Note the post meta overlapping the site title + nav here: 

Screen Shot 2021-05-20 at 11 02 27 AM

MaggieCabrera:
ok that only happens when the text has many lines (the spacers are doing weirds stuff). I'll make a new PR to address it, should be easy enough

PR #3890, opened 2021-05-20
Quadrat: Add padding to full align column and group blocks

### LINKED ISSUE #3747: Quadrat: Column Block Margins

### LINKED ISSUE #3878: Quadrat: margins on mobile

scruffian:
This is a different approach to https://github.com/Automattic/themes/pull/3816.

The approach is much simpler - we just add padding to the column and group blocks when they are fully aligned. This also requires us to remove the padding from the post content block.

#### Related issue(s):
https://github.com/Automattic/themes/issues/3747

### DIFF
--- a/quadrat/experimental-theme.json
+++ b/quadrat/experimental-theme.json
@@ -202,8 +202,8 @@
 			},
 			"post-content": {
 				"padding": {
-					"left": "var(--wp--custom--margin--horizontal)",
-					"right": "var(--wp--custom--margin--horizontal)"
+					"left": "0px",
+					"right": "0px"
 				}
 			},
 			"pullquote": {
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -110,6 +110,14 @@
 	background: #fff;
 }
 
+.wp-block-columns.alignfull {
+	padding: var(--wp--custom--margin--vertical) var(--wp--custom--margin--horizontal);
+}
+
+.wp-block-group.alignfull {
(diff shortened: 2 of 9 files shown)


quadrat/experimental-theme.json
```diff
@@ -202,8 +202,8 @@
 			},
 			"post-content": {
 				"padding": {
-					"left": "var(--wp--custom--margin--horizontal)",
-					"right": "var(--wp--custom--margin--horizontal)"
+					"left": "0px",
```
jffng, line 205: I understand why you want to remove this, so that content can extend the full width. But it causes all of the blocks (including text) to start at the edge of the screen:

Screen Shot 2021-05-20 at 11 52 38 AM

jffng (commented):
I like this approach a lot better. This will also solve #3878.

jffng:
What should happen here? How does this PR relate to https://github.com/Automattic/themes/pull/3816?

scruffian:
Closing in favour of #3953

PR #3891, opened 2021-05-20
Quadrat: changed the hover effect of text links

### LINKED ISSUE #3782: Quadrat: Reconsider the text hover effects
I think the background color hover effect we're using ends up being a little too prominent for my taste when it's applied to large headings: 

Screen Shot 2021-05-07 at 12 51 29 PM

Also, I think it's a little weird that we have a different hover state for one portion of the header vs. another: 

hovers

So I recommend a few changes: 

1. Use the background color hovers _only_ for links within body text. 
2. Use underlines for all other hovers. 

@beafialho what do you think of that approach? I think another idea would be to just use underlines everywhere, and remove the background-color treatment entirely. 

(If we do keep the background color hovers, I think we also need to be sure to remove the underline whenever the background color hover is active.)

MaggieCabrera:

Following the design guidelines of #3782, this PR changes how links work on hover, having only links inside paragraph tags have the background change (without underline on them). I have a few questions about this:

- What happens to the hover effects of the rest of the links? right now they have none besides the cursor effect.
- Is it ok that we only target links inside paragraphs? or should we target post-content instead? Other small links like pagination blocks etc that are not paragraphs and are not part of the post content will not have the background hover effect either.

Screen Capture on 2021-05-20 at 14-42-52
Screen Capture on 2021-05-20 at 14-44-33

#### Related issue(s):

Closes #3782

### DIFF
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -358,12 +358,13 @@ a {
 	        text-decoration-line: underline;
 }
 
-a:hover {
+p a:hover {
+	text-decoration: none;
 	background: var(--wp--custom--color--primary);
 	color: var(--wp--custom--color--background);
 }
 
-a:active, a:focus {
+p a:active, p a:focus {
 	background: var(--wp--custom--color--secondary);
 }
 
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -231,6 +231,16 @@ ul ul {
 	grid-area: form-submit;
 }
 
+.wp-block-post-comments .comment-body > p > a,
+.wp-block-post-comments .comment-edit-link {
+	text-decoration: underline;
+}
+
+.wp-block-post-comments .comment-body > p > a:hover,
+.wp-block-post-comments .comment-edit-link:hover {
+	text-decoration: none;
+}
+
 .wp-block-post-navigation-link {
 	border-top: 1px solid;
 	display: flex;
@@ -313,6 +323,10 @@ ul ul {
 	justify-self: center;
 }
 
+.wp-block-query-pagination .wp-block-query-pagination-numbers .current {
+	text-decoration: underline;
+}
+
 @media (max-width: 479px) {
 	.wp-block-query-pagination .wp-block-query-pagination-numbers {
 		display: none;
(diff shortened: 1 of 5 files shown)


beafialho:
> What happens to the hover effects of the rest of the links? right now they have none besides the cursor effect.

In the rest of the links, when a link is underlined, the underline disappears on hover. So, this wouldn't happen:

Captura de ecrã 2021-05-20, às 14 53 04

The reverse would happen instead:

Captura de ecrã 2021-05-20, às 14 55 43

Would love your double check here @kjellr

MaggieCabrera:
When does the first case (default not underline, hover underlined) happen? I can only think of the site title on the header

kjellr:
> When does the first case (default not underline, hover underlined) happen? I can only think of the site title on the header

Yeah, I think it's just the page header. @beafialho do we want post headers to have underlines by default or not?

beafialho:
> When does the first case (default not underline, hover underlined) happen? I can only think of the site title on the header

That would happen in a link on post content and in the footer text.

> @beafialho do we want post headers to have underlines by default or not?

You mean post titles? I imagined these wouldn't be underlined by default. Here's what I mean:

Captura de ecrã 2021-05-20, às 15 41 41

MaggieCabrera:
> > When does the first case (default not underline, hover underlined) happen? I can only think of the site title on the header
> 
> That would happen in a link on post content and in the footer text.

I thought links on post contents were going to have the background color? as @kjellr suggested on #3782 :

> Use the background color hovers only for links within body text.

kjellr:
> I thought links on post contents were going to have the background color? as @kjellr suggested

Oh no, sorry for the confusion. We'd like to go with @beafialho's later suggestion in https://github.com/Automattic/themes/issues/3782#issuecomment-836514900.

jffng:
I paired with @beafialho @kjellr to make the following changes:

- Links outside the post content are un-styled by default, and underlined on hover. The same for linked post titles that appear in an index/archive template.
- Links inside the post content are the opposite.
- The active page in the post query pagination now has an underline.
- If a link appears inside the body of a comment, it has an underline.

I think that's it. I decided to address the focus states in a separate PR because it should tackle buttons too.

kjellr (approved):
Looks good, thanks!

PR #3899, opened 2021-05-20
Quadrat: revise focus states

### LINKED ISSUE #3868: Quadrat: Change Links Focus States
Might be related to #3782, currently the focus state for links looks like this:

Captura de ecrã 2021-05-17, às 13 10 46

Here's how it should look with a dotted border:

Captura de ecrã 2021-05-17, às 15 21 27

Captura de ecrã 2021-05-17, às 15 22 54

jffng:
Closes #3868. It adds an `outline` to all links that are not wp-admin related, and form fields. 

Before | After
------- | -------
Screen Shot 2021-05-20 at 3 39 37 PM | Screen Shot 2021-05-20 at 3 36 39 PM

cc @beafialho — I am using `outline` here instead of `border` because a border will grow the size of the element on focus, and I don't think that's what we want. The downside is we have a little less control over the style of the outline.

### DIFF
--- a/quadrat/sass/elements/_forms.scss
+++ b/quadrat/sass/elements/_forms.scss
@@ -1,0 +1,21 @@
+input[type="text"],
+input[type="email"],
+input[type="url"],
+input[type="password"],
+input[type="search"],
+input[type="number"],
+input[type="tel"],
+input[type="range"],
+input[type="date"],
+input[type="month"],
+input[type="week"],
+input[type="time"],
+input[type="submit"],
+input[type="datetime"],
+input[type="datetime-local"],
+input[type="color"],
+textarea {
+	&:focus {
+		outline: 1px dotted currentColor;
+	}
+}
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -376,6 +376,31 @@ ul ul {
 	text-decoration: none;
 }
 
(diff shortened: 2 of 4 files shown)


quadrat/sass/elements/_forms.scss
```diff
@@ -0,0 +1,21 @@
...
+input[type="month"],
+input[type="week"],
+input[type="time"],
+input[type="submit"],
+input[type="datetime"],
+input[type="datetime-local"],
+input[type="color"],
+textarea {
+	&:focus {
+		outline: 1px dotted currentColor;
```
jffng, line 19: This felt very theme specific Maybe it should move to BCB, but I feel like it will be awhile (if ever) before Gutenberg / theme.json provides this kind of control to the themers let alone users.

pbking: Yea the "make it easy to do again" part of my brain agrees it should be in BCB.

The "BCB is for ponyfilling what Gutenberg will do" part of my brain is winning though and I agree with you; this belongs here.

pbking (approved):
Looks nice and does what it says on the tin.
I couldn't find anywhere it didn't work as expected .

quadrat/sass/elements/_forms.scss
scruffian: I think if we end up doing this in a few themes we should move it to BCB.

PR #3896, opened 2021-05-20
Quadrat: fix single post title when a custom logo is present

MaggieCabrera:
#### Changes proposed in this Pull Request:

When a custom logo is active and the post title is long the post meta was moving on top of the menu.

Screenshot 2021-05-20 at 17 43 32
Screenshot 2021-05-20 at 17 43 47

Issue #3804, opened 2021-05-10
Quadrat: Post list cleanup


kjellr:
I have a few suggestions for more closely aligning the list of posts in Quadrat with the full post view: 

- Move the date up to the top, center-align the text.
- Move the title below that, center-align the text.
- The spacing below the post title is currently 30px. Let's increase that to 45px. 
- Increase the spacer block underneath each post from 30px to 60px to allow for more whitespace in between each one. 

Before|After
---|---
<img width="752" alt="Screen Shot 2021-05-10 at 11 59 22 AM" src="https://user-images.githubusercontent.com/1202812/117691131-8ed6a680-b189-11eb-9bcf-099fc7e5cb12.png">|<img width="797" alt="Screen Shot 2021-05-10 at 12 15 41 PM" src="https://user-images.githubusercontent.com/1202812/117691137-9007d380-b189-11eb-9371-3ffe44de811d.png">

cc @beafialho

PR #3904, opened 2021-05-21
Quadrat: mobile post header with featured image

### LINKED ISSUE #3897: Quadrat: Update Meta/Title spacing on small screens
Really minor, but can we trim `20px` from the vertical spacing between the post meta and the title on mobile views? They feel a little far apart to me there. It can stay as is on larger screens. 

Before|After
---|---
Screen Shot 2021-05-20 at 1 41 08 PM|Screen Shot 2021-05-20 at 1 41 44 PM

MaggieCabrera:

The mobile version of the post header was causing issues with a featured image, I went to the figma and saw that this has been designed already and I tried my hand at reproducing the same design as close as possible. I'm using `object-fit: cover;` here so I'm manipulating a bit how the image looks like compared to desktop. I'd love design eyes on this if possible @beafialho @kjellr 

I went ahead and did the change for the spacing between the post meta and the post title in the same PR.

Figma | Site
--- | ---
Screenshot 2021-05-21 at 09 43 41 | Screen Capture on 2021-05-21 at 09-45-31(1)

#### Related issue(s):

Closes #3896
Closes #3897

### DIFF
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -642,6 +642,7 @@ textarea:focus {
 @media (max-width: 479px) {
 	.post-meta {
 		padding-top: 0 !important;
+		margin-bottom: -20px;
 	}
 }
 
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -716,6 +716,9 @@ textarea:focus {
 	.single.wp-custom-logo .post-header {
 		padding-top: 150px !important;
 	}
+	.single.wp-custom-logo .wp-block-post-featured-image {
+		margin-top: -150px !important;
+	}
 	.single .site-header {
 		position: absolute;
 		top: 0;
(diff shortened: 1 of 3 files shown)


beafialho:
Thanks @MaggieCabrera! The spacing looks good to me, the only comment is that there should be a square behind the post title and meta, overlapping the featured image.

Captura de ecrã 2021-05-21, às 12 37 56
Captura de ecrã 2021-05-21, às 12 38 11

MaggieCabrera:
Mmmh, I'm looking at the desktop version of the post with a featured image and that rises more questions:

- I checked @kjellr [opened this issue](https://github.com/Automattic/themes/issues/3804) to change how posts with featured images should look like and that deviates from this design (and also doesn't include the square)
- I'm a bit hesitant about including the square using CSS because that will mean the user can't remove it if they don't want it. I could probably make it a block style but that won't be available for the users to change until FSE is available to them (I'm still unsure if dotcom is including the FSE for users or not when 5.8 lands).

beafialho:
Oh I see, I missed that issue. In that case, I think what you have here looks good.

kjellr:
Hey sorry I missed this PR. I think we should revert, because the site title and menu can very easily become unreadable depending on the featured image used: 

quadrat

PR #3906, opened 2021-05-21
Quadrat: fix markup for header template part

MaggieCabrera:

The header template part that's being used when FSE is active didn't have the same markup as the php version and they didn't look the same.

To test this, check the site with and without an index.html template inside block-templates folder. It should show the same header on both accounts.


quadrat/block-template-parts/header.html
```diff
@@ -1,11 +1,15 @@
...
-			<!-- wp:navigation-link {"label":"Contact","url":"#"} /-->
-		<!-- /wp:navigation -->
-	</div>
-<!-- /wp:group -->
+<!-- wp:group {"tagName":"header","style":{"spacing":{"padding":{"right":"35px","left":"35px"}}},"className":"site-header"} -->
+<header class="wp-block-group site-header" style="padding-right:35px;padding-left:35px"><!-- wp:group -->
+<div class="wp-block-group"><!-- wp:site-title /-->
+
+<!-- wp:navigation {"orientation":"horizontal","textColor":"foreground-light","itemsJustification":"right","fontSize":"small"} -->
+<!-- wp:navigation-link {"label":"Home","url":"#"} /-->
```
pbking: I don't think we should ship navigation blocks with anything in them.  Empty blocks have better prompts for the user in the Block Editor (though the downside is the user won't see ANY navigation in the view until they add something themselves.

I'm relatively impartial but saw discussion about the same RE: TT1-Blocks

jffng:
I made a few changes: 

- Removed nav links by default in the FSE header template part
- Added a site logo block
- Refactored to account for `.site-header` [styles now present in BCB](https://github.com/Automattic/themes/blob/trunk/blank-canvas-blocks/sass/base/_header.scss)
- Simplified block markup to remove an unneeded group block wrapper

pbking:
I tried to look at this but couldn't get index.html to work at all.  (I just copied single.html to have some content).  The page just renders:
```
Template part has been deleted or is unavailable: header Template part has been deleted or is unavailable: single Template part has been deleted or is unavailable: footer
```

Pages NOT using index.html as the template (a post page for instance) are also giving the same error.

I'm not sure what I'm doing wrong...

pbking:
Added a couple of tweaks.  
* I noticed that there were two `<header>` elements so I removed the one int he template part.  
* Padding was being applied in two places resulting in extra-padded header on mobile view.
* Navigation didn't default to 'responsive' in the template.
* There seemed to be a lot of unnecessary attributes in the navigation block.  Removed everything that was set to the default value or was already being managed more appropriately in theme.json configuration.

jffng (approved):
Less code, markup matches, and it appears the same in both FSE and not. 🚢

PR #3909, opened 2021-05-21
Provide reasonable color defaults when palette background color is selected

### LINKED ISSUE #3814: Quadrat: Theme has no auto-text color adjustments for the background color presets
In our traditional themes, we automatically ensure that the font color is readable if someone were to choose one of the theme presets for a background color. This isn't the case in this theme: 

Screen Shot 2021-05-10 at 3 10 19 PM

I'm not sure if this is something we're going to implement, but if we did, I suggest we use our "Blue" color for default text against the white and pink backgrounds. 

Screen Shot 2021-05-10 at 3 14 00 PM

pbking:
image

Sets the color for any element that expresses a background color who's foreground color doesn't work as the default.

#### Related issue(s):

Fixes #3814

### DIFF
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -406,6 +406,11 @@ ul ul {
 	padding: var(--wp--custom--margin--horizontal);
 }
 
+.has-pink-background-color,
+.has-white-background-color {
+	color: var(--wp--preset--color--blue);
+}
+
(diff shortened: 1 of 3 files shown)


jffng (approved):
LGTM. I wonder if we will end up putting anything else in the `_colors` partial.

PR #3917, opened 2021-05-24
Quadrat: removed polygon on the footer outside of the home

### LINKED ISSUE #3903: Quadrat: Comments form on top of footer polygon
Right now the header and footer polygons are visible on every page, this causes some overlap on single pages when the comment form is visible and I think it makes the textarea not very accesible. We could hide the footer polygon on single pages or make changes to the form so that it stands out more. @beafialho @kjellr I'd love your input on this issue, thank you!

Screenshot 2021-05-21 at 09 15 33

MaggieCabrera:

Home | Single
--- | ---
Screenshot_2021-05-24 Screenshot | Screenshot_2021-05-24 Screenshot(1)

#### Related issue(s):

Closes https://github.com/Automattic/themes/issues/3903

### DIFF
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -727,12 +727,12 @@ textarea:focus {
 	}
 }
 
-.site-footer.wp-block-group {
+.home .site-footer.wp-block-group {
 	position: relative;
 	overflow: visible;
 }
 
-.site-footer.wp-block-group:before {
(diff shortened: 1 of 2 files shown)


MaggieCabrera:
The home screenshot is broken cause of my screenshot extension, but it gets the point across :D

pbking (approved):
👍

PR #3920, opened 2021-05-24
Mayland Blocks: fix header template part

jffng:
This PR makes a few changes to Mayland Blocks:

- Fixes a bug where the wrong stylesheet was being enqueued 
- Simplifies the header template to match BCB + Quadrat


pbking (approved):
Fixes what ails.  Let's bring it in.

🚢

PR #3921, opened 2021-05-24
Style quadrat mobile nav

### LINKED ISSUE #3789: Quadrat: Mobile menu is unpolished
These should be pretty obvious, but I didn't see them filed anywhere so I'm recording them here. 

The mobile menu has a lot of pretty obvious bugs. For example: 

- Menu icon isn't properly right-aligned
- Close icon isn't in the same position as the menu icon
- Menu items are not properly centered

For the icons, please use the ones from Seedlet. We should also update the font size to our Extra Large (35px) size. 

Closed|Open
---|---
Screen Shot 2021-05-07 at 1 46 59 PM|Screen Shot 2021-05-07 at 1 47 18 PM

pbking:
This change leverages the Gutenberg mobile nav and styles it according to the [figma design](https://www.figma.com/file/EDoXTvsZt4lMHItzOczFQw/Quadrat?node-id=125%3A2).

This change also eliminates some of the mobile menu "open" button positioning done in Blank Canvas Blocks.  The result is that the menu button shows up where the menu block is, not a set location of the page.  (This approach was working in BCB and Seedlet-blocks but not Mayland blocks or Quadrat; it seemed obvious that this opinionated decision didn't belong in BCB).

This fixes #3789.  However, departing from the suggestions there this change instead styles according to the figma design.

NOTE: According to #3789  "For the icons, please use the ones from Seedlet." But unfortunately I haven't figured out how any other icon other than what Gutenberg provides is available.

Before:
image

After:
image

Design Target:
image

Also Note the menu open button location changes in this change for all BCB themes:

Quadrat:
image

BCB:
image

Seedlet:
image

Mayland:
image

I suggest that MAYLAND and SEEDLET have further changes to address the position of that OPEN button in a separate change.  (I suggest that BCB remain as-is since that's how it operates shipped from Gutenberg and that is unlikely to change; any repositioning would be theme-specific.)


pbking:
The menu open button location changes noted above regarding Mayland Blocks pairs nicely with [this change](https://github.com/Automattic/themes/pull/3920) which addresses that.

jffng (commented):
This is very exciting. I love how much code we can rip out to get a functioning responsive nav. 

My only comment is about the background cover of the responsive modal. Based on the design target, I think it should be `var(--wp--custom--color--secondary)` — the darker blue rather than the background color.

cc @kjellr @beafialho for a review here.

beafialho:
> My only comment is about the background cover of the responsive modal. Based on the design target, I think it should be var(--wp--custom--color--secondary) — the darker blue rather than the background color.

That is correct, the background color for the mobile menu should be the darker blue.

pbking:
Rebased against trunk, re-tweaked blockbase styles (lost in conversion) and corrected the background color.

image

Hopefully G2G now!

kjellr (approved):
LGTM. I noticed you can still scroll underneath the menu overlay, but I think that's likely a Gutenberg bug. I'll file an issue for it. 

scrolling

PR #3947, opened 2021-05-25
Quadrat: Another fix for alignments

scruffian:
This is another attempt at https://github.com/Automattic/themes/issues/3747

The approach is to add left and right padding to paragraph blocks when they are inside a full width container, or when they are inside a container which inherits layout and the screen is smaller than the container.

To test:
- Add a paragraph block inside the post and check that when the page is narrow the content doesn't touch the sides
- Add a paragraph block inside a wide block and check that when the page is narrower than the wide width, the content doesn't touch the sides
- Add a paragraph block inside a full block and check that the content doesn't touch the sides

#### Related issue(s):
#3747

### DIFF
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -659,6 +659,18 @@ textarea:focus {
 	outline: 1px dotted currentColor;
 }
 
+.alignfull p {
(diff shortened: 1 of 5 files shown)


scruffian:
Closing in favour of #3953

PR #3948, opened 2021-05-25
Blockbase: Update header font size

kjellr:
The header font size should be the same size as the menu text: 

Before:
Screen Shot 2021-05-25 at 11 52 12 AM

After:
Screen Shot 2021-05-25 at 11 51 34 AM


pbking:
That large font size was assigned in theme.json so[ I suggest changing it there](https://github.com/Automattic/themes/pull/3948/commits/fcc4791869de9a2baab97186e019894b8b9af83c) instead of the template.

kjellr:
Makes sense, thanks!

PR #3949, opened 2021-05-25
Blockbase: add readme.txt and update style.css

jffng:
This PR adds a `readme.txt` needed for themes on dotorg, and removes some image credits that are no longer packaged in the theme.


blockbase/readme.txt
```diff
@@ -0,0 +1,33 @@
...
+Tested up to: 5.7.2
+Requires PHP: 5.6.2
+License: GPLv2 or later
+License URI: http://www.gnu.org/licenses/gpl-2.0.html
+
+A starting point for building your block-based site.
+
+== Description ==
+
+Blockbase is a simple starting point for themes that work with Full Site Editing. It comes with a set of minimal templates and a system of design settings expressed through theme.json. Use it to build something beautiful.
```
kjellr: I think we should adjust this to be less developer-focused. For example: 

> Blockbase is a simple theme that supports full-site editing. It comes with a set of minimal templates and design settings that can be manipulated through Global Styles. Use it to build something beautiful.

pbking:
Why two files?  if the .txt is required by .org can we just eliminate the .md?

jffng:
> if the .txt is required by .org can we just eliminate the .md?

That works, .txt just isn't formatted / viewed as nicely in GitHub.  We'd probably want to migrate all the developer / child-theme focused info to the txt too so it's not lost.

pbking:
> That works, .txt just isn't formatted / viewed as nicely in GitHub. 
Indeed, but it's [not too bad](https://github.com/Automattic/themes/tree/add/blockbase-readmetxt-test/blockbase).  

I'm already sad that meta-information is in two places instead of one (theme.css + readme).  It hurts my face to think that it would be in three. 

> We'd probably want to migrate all the developer / child-theme focused info to the txt too so it's not lost.

Yuppers

jffng:
Moved the markdown info to a new section of the txt. Any objections to bringing this in?

blockbase/readme.txt
```diff
@@ -12,6 +12,28 @@ A starting point for building your block-based site.
...
+
+- Add a [package.json](https://github.com/Automattic/themes/blob/make/quadrat/seedlet-blocks/package.json), [style.css](https://github.com/Automattic/themes/blob/make/quadrat/seedlet-blocks/style.css) and empty index.php files to your theme
+- Your child theme's theme.json file will be built by combining Blockbase's original theme.json file and your child's `child-experimental-theme.json`. That is the file where your configuration values live (you only want the values for variables that are different from the parent there). To generate the combined file, run `npm run build` from inside your theme's folder or run `npm run build:child child-theme-slug` from inside Blockbase's folder.
+- Extend the parent's templates that you want to customize or add new ones where needed following the [template hierarchy](https://themeshaper.com/2020/12/18/getting-started-with-block-themes-templates/).
+- Using `npm run start` will both watch changes inside the `sass` and `assets` folder and changes to the `child-experimental-theme.json` file.
+- Additionally, your child theme may want to unregister the block patterns and block styles defined by Blockbase.
```
kjellr: Does blockbase do this anymore?

scruffian: No, we can remove that

PR #3954, opened 2021-05-26
Blockbase: added supported tags

MaggieCabrera:

Following this [list of tags](https://make.wordpress.org/themes/handbook/review/required/theme-tags/) I've added the ones that we are supporting right now. I18n is out of our control and a11y needs to pass the required test first.

RTL support:

Screenshot 2021-05-26 at 16 48 06

### DIFF
--- a/blockbase/style.css
+++ b/blockbase/style.css
@@ -11,7 +11,7 @@ Version: 1.0.0
 License: GNU General Public License v2 or later
 License URI: LICENSE
 Text Domain: blockbase
-Tags: one-column, editor-style, featured-images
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
Some of these I've decided to add because [TT1](https://github.com/WordPress/theme-experiments/blob/master/tt1-blocks/style.css) has them, even though on the handbook it's talking about the customizer, these should apply anyway.

scruffian:
I think this is fine. 

`custom-logo` and `custom-menu` might be a problem since we don't use those functions, but let's see what happens when we submit it.

PR #3956, opened 2021-05-26
Remove experimental- from theme.json & include in .org build

pbking:
Updates blockbase and all child themes (Including the defunct Blank-Canvas-Blocks) to remove 'experimental-' from theme.json and child-theme.json files.

Also removed build scripts from blank-canvas-blocks theme.

Includes theme.json file in .org packaging script


seedlet-blocks/theme.json
```diff
@@ -434,7 +434,7 @@
 			},
 			"core/site-title": {
 				"typography": {
-					"fontSize": "60px",
+					"fontSize": "var(--wp--preset--font-size--normal)",
```
jffng, line 437: I believe this causes a visual regression in the header: 

Screen Shot 2021-05-27 at 10 30 40 AM

We can address it by setting the font size of the site title in the header template.

jffng: Done here https://github.com/Automattic/themes/pull/3956/commits/c0e853e9201cc34920f3c9a63afafd3a5276a699

PR #3958, opened 2021-05-27
Blockbase: Move block styles to theme.json

scruffian:
We can set list typography in the theme.json now, we don't need to do it in custom, so this is just a simplification.

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -334,7 +334,6 @@ input[type=checkbox] + label {
 
 ul,
 ol {
-	font-family: var(--wp--custom--list--typography--font-family);
 	padding-left: var(--wp--custom--list--spacing--padding--left);
 }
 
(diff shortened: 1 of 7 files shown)


MaggieCabrera:
I'm running the latest from GB trunk and blockbase is not showing fontfamily for ol/ul:

Screenshot 2021-06-17 at 14 32 19

scruffian:
I changed the approach here. We don't need to declare the font at all, since themes can just use theme.json when they want to change it now.

There are also some new templates that come from rebuilding the child themes which should have happened in a different PR.

pbking:
Looks like the related PR was closed in favor of this one.
This might still need some work to support that particular change, but there are conflicts so I'm moving out of "review" until those are resolved.

scruffian:
This is rebased and good to go.

pbking (approved):
Good to Go Indeed.
🚢

PR #3961, opened 2021-05-27
Quadrat: Fix mobile spacing

scruffian:
This introduces space around the page content on all templates in Quadrat:

Screenshot 2021-05-27 at 15 01 42

I have changed from a px value to a vw value, so that it is intrinsically responsive. It was also necessary to update the header so the two values match.

#### Related issue(s):
#3878

### DIFF
--- a/quadrat/theme.json
+++ b/quadrat/theme.json
@@ -167,8 +167,8 @@
 			},
 			"post-content": {
 				"padding": {
-					"left": "var(--wp--custom--margin--horizontal)",
-					"right": "var(--wp--custom--margin--horizontal)"
+					"left": "min( var(--wp--custom--margin--horizontal), 5vw)",
+					"right": "min( var(--wp--custom--margin--horizontal), 5vw)"
 				}
 			},
 			"pullquote": {
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -662,12 +662,12 @@ textarea:focus {
 .site-header {
 	position: relative;
 	overflow: inherit;
-	padding: 10px 20px 60px;
+	padding: 10px var(--wp--custom--post-content--padding--left) 60px;
 }
 
 @media (min-width: 480px) {
 	.site-header {
-		padding: var(--wp--custom--margin--vertical);
+		padding: var(--wp--custom--post-content--padding--left);
 	}
 }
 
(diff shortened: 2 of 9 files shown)


MaggieCabrera:
There's a couple issues that I see. On one hand, the fullwidth alignments don't seem to be working, I have missed this when reviewing the alignments PR but it's still relevant to this one because of the padding values that this one introduces:

Screenshot 2021-05-27 at 16 11 45

On the other hand on the index page the pagination seems to be going full width but I think it should be have the same as the single post navigation does:

Screenshot 2021-05-27 at 16 09 05

quadrat/experimental-theme.json
```diff
@@ -167,8 +167,8 @@
 			},
 			"post-content": {
 				"padding": {
-					"left": "var(--wp--custom--margin--horizontal)",
-					"right": "var(--wp--custom--margin--horizontal)"
+					"left": "5vw",
+					"right": "5vw"
```
MaggieCabrera: this seems very big on desktop, why do we need it to be a responsive value if we only actually see the effect when we are on mobile?

scruffian: > why do we need it to be a responsive value if we only actually see the effect when we are on mobile?

because this is the same spacing we use in the header

scruffian:
I made a few changes to this:

- The alignments were broken because the branch needed a rebase
- I used a `min` in the definition of the padding, so it doesn't get too big on desktop

quadrat/theme.json
```diff
@@ -167,8 +167,8 @@
 			},
 			"post-content": {
 				"padding": {
-					"left": "var(--wp--custom--margin--horizontal)",
-					"right": "var(--wp--custom--margin--horizontal)"
+					"left": "min( var(--wp--custom--margin--horizontal), 5vw)",
```
jffng, line 170: I like the approach to use min here with a px + relative unit comparison. I wonder if the `5vw` should actually be a smaller value, so the padding on mobile is tighter. Right now it seems to only shrink ever so slightly at viewports < 400px.

jffng (approved):
LGTM, could use a design review to verify the paddings are correct.

quadrat/theme.json
kjellr: In my testing, `5vw` seems fine to me. This seems good to go. 👍

PR #3962, opened 2021-05-27
Remove BCB

scruffian:
Now BCB has been migrated to Blockbase we should be able to remove this code


pbking (approved):
By By BCB.
It's been fun.

PR #3968, opened 2021-05-27
Blockbase: revise padding

### LINKED ISSUE #3963: Blockbase: Left and right mobile padding is incorrect
We should have `20px` of padding on the left and right of the header, post titles, featured image, and post content. 

Current|Proposed
---|---
Screen Shot 2021-05-27 at 11 34 54 AM|Screen Shot 2021-05-27 at 11 35 44 AM

jffng:
This PR uses CSS classes to add padding around the header, post titles + featured image, and post content.

I thought this was a better approach than adding the styles inline via block attributes, because it matches how we are adding the padding around the post content.

**To test**
- Add some posts that have a featured image
- Check the index / query and singular views 
- Ensure the padding is correct (20px unless aligned full)

Before | After
------- | --------
Screen Shot 2021-05-27 at 12 43 09 PM | Screen Shot 2021-05-27 at 12 43 36 PM

Closes #3963.

### DIFF
--- a/blockbase/theme.json
+++ b/blockbase/theme.json
@@ -159,8 +159,8 @@
 			},
 			"post-content": {
 				"padding": {
-					"left": "var(--wp--custom--margin--horizontal)",
-					"right": "var(--wp--custom--margin--horizontal)"
+					"left": "20px",
+					"right": "20px"
 				}
 			},
 			"pullquote": {
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -54,12 +54,15 @@ img {
 /**
  * These are default block editor widths in case the theme doesn't provide them.
  */
-.wp-block-post-content > .alignfull {
+.wp-block-post-content > .alignfull,
+.post-header > .alignfull {
 	margin-left: calc(-1 * var(--wp--custom--post-content--padding--left));
 	margin-right: calc(-1 * var(--wp--custom--post-content--padding--right));
 	width: calc( 100% + var(--wp--custom--post-content--padding--left) + var(--wp--custom--post-content--padding--right));
 }
 
+.site-header,
+.post-header,
 .alignfull p {
 	padding-left: var(--wp--custom--post-content--padding--left);
 	padding-right: var(--wp--custom--post-content--padding--right);
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -62,7 +62,7 @@ img {
 }
 
 .site-header,
-.post-header,
+.page-content,
 .alignfull p {
 	padding-left: var(--wp--custom--post-content--padding--left);
 	padding-right: var(--wp--custom--post-content--padding--right);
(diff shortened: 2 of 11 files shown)


jffng:
Just realized that the search + 404 templates also need some wrappers to add padding around the content.

jffng:
I think this PR should be updated to use the approach in #3961 where there is a single "page content" class applied to the top level group of each template.

jffng:
I updated all the templates to reflect these padding values. This should be ready for a review.

pbking:
The only thing I was confused about was the full-width featured image on the singular page.  I'm probably missing context, but if that's expected then everything else is as I expected it to be.

LGTM

scruffian:
The code looks good, but I don't understand why we're changing the padding for post content to be 20px. Happy for that change to go ahead as it separates the generic site margins from the post content padding, but it is a change...

scruffian (approved):

blockbase/theme.json
```diff
@@ -160,8 +160,8 @@
 			},
 			"post-content": {
 				"padding": {
-					"left": "var(--wp--custom--margin--horizontal)",
-					"right": "var(--wp--custom--margin--horizontal)"
+					"left": "20px",
+					"right": "20px"
```
pbking, line 163: After a 3rd look I'm unsure why this value isn't `var(--wp--custom--post-content--padding--left)` and `var(--wp--custom--post-content--padding--right)`.  If the post-content value needs to change from 30px to 20px then we should change it in the custom block so that the other places that leverage that value are picking up what's expected.

jffng: I think you mean these custom padding variables should be set to `var(--wp--custom--margin--horizontal)` and we should change the value of that custom variable to `20px`. Or am I misunderstanding?

pbking: I'm saying that the "post-content" padding style should be set to the custom variable `--wp--custom--post-content--padding--left/right`.  I don't believe it was ever correct to be set to `--wp--custom--margin--horizontal`, but because the custom value was also set to the same variable it was never perceived as an issue.

That variable (`--wp--custom--post-content--padding--left/right`) is currently being set to `--wp--custom--margin--horizontal` but if we want to reduce the padding around post-content we should be adjust that custom value (which other things are using) instead of changing the value here.

I'm not sure why that value should be 20px instead of the 30px that --wp--custom--margin--horizontal is (that's a design decision and I'm happy not to make those), but I'm saying that it's the custom value that should change (so that other things in CSS know about the change).

TL;DR
```
custom: {
  post-content: {
    padding: {
      left: "20px"
      right: "20px"
    }
  }
}
...
styles: {
    post-content: {
      padding: {
        left: var(--wp--custom--post-content--padding--left)
        right: var(--wp--custom--post-content--padding--left)
```
so that elsewhere...
```
    margin-left: calc(-1 * --wp--custom--post-content--padding--left);
```

pbking: Let be known that I'm completely confused and it's right and I'm not sure why I'm arguing.... :P

PR #3970, opened 2021-05-27
Remove unused navigation header template part.

pbking:
This change removes the unused navigation template part and the
reference to it from theme.json.  It also updates the navigation block
in the header template part to include responsive navigation by default
and remove unnecessary block attributes (values were either default or
expressed in theme.json).

### DIFF
--- a/blockbase/block-template-parts/header.html
+++ b/blockbase/block-template-parts/header.html
@@ -1,6 +1,6 @@
 <!-- wp:group {"align":"full","tagName":"header","style":{"spacing":{"padding":{"right":"35px","left":"35px"}}},"className":"site-header"} -->
-<header class="wp-block-group alignfull site-header" style="padding-right:35px;padding-left:35px">
+<div class="wp-block-group alignfull site-header" style="padding-right:35px;padding-left:35px">
 	<!-- wp:site-title /-->
 	<!-- wp:navigation {"isResponsive":true} /-->
-</header>
+</div>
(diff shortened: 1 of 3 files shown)


jffng (approved):
Makes sense to simplify here and remove unnecessary pieces. 

I do think it's worth citing what you mentioned in slack. 

> The ONLY benefit (I’m aware of) to keeping it in its own template part is that you can change the navigation template part and still get updates to the header template part (since you wouldn’t have a saved version of that one).

But as you said, that's potentially confusing to manage right now and not worth keeping it around for, as far as I'm aware.

blockbase/block-template-parts/header.html
```diff
@@ -1,6 +1,6 @@
 <!-- wp:group {"align":"full","tagName":"header","style":{"spacing":{"padding":{"right":"35px","left":"35px"}}},"className":"site-header"} -->
 <header class="wp-block-group alignfull site-header" style="padding-right:35px;padding-left:35px">
```
jffng: While we are here, I noticed that we have currently have header inception because we are rendering a template part with a header tag and then a group inside of it with a header tag. We should consolidate?

Screen Shot 2021-05-27 at 4 04 35 PM

pbking: I think I fixed that same thing in Quadrat too.  Yes we should.  I don't believe the element in the template part should be a `<header>` but just a `<div>`.

PR #3972, opened 2021-05-27
Fix BB & Co. Button Border Color Configuration

### LINKED ISSUE #3966: Blockbase: Default outline button color is black instead of blue
When you switch the button style from default to outlined, the button appears black instead of blue. I expected it to remain blue, but have that color apply to the outline and text by default: 

Current|Expected
---|---
Screen Shot 2021-05-27 at 12 13 29 PM|Screen Shot 2021-05-27 at 12 15 05 PM

pbking:
Changed CSS and variable handling so that the color of the border and text of a button in 'outline mode' matches the background color of the button when not.

Ensured the hover state of these buttons (when no color customizations have been applied to the button by the user) work as expected (matching the hover state of non-outline buttons).

Also ensured that when changing an "outline button"'s text color that the outline changes to match that color.

Tested in editor and view for BB and all child themes.

Quadrat (no visible changes)
image

Mayland-Blocks (no visible changes)
image

Seedlet-Blocks (changed from black to green)
image

Blockbase (changed from back to blue)
image

#### Related issue(s):

Fixes #3966

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -225,7 +225,7 @@ input[type=checkbox] + label {
 	opacity: 1;
 	color: var(--wp--custom--button--color--text);
 	background-color: var(--wp--custom--button--color--background);
-	border-color: var(--wp--custom--button--border--color);
+	border-color: currentColor;
 	border-radius: var(--wp--custom--button--border--radius);
 }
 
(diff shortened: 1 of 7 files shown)


jffng (approved):
Tested in BB & Co, and all looking good. Code changes check out as well.

PR #3974, opened 2021-05-27
Removed top padding from quote blocks's paragraph element

### LINKED ISSUE #3969: Blockbase: Quote has too much margin on top of the paragraph
The first paragraph inside of the quote block should have no top margin. This is only an issue on the frontend, but it effects both the default and "Large" variants of the quote block. 

Current|Recommended
---|---
Screen Shot 2021-05-27 at 12 45 37 PM|Screen Shot 2021-05-27 at 12 46 45 PM

pbking:
This effects all child themes as well.

BB Before:
image

BB After:
image

Seedlet-Blocks before:
image

Seedlet-Blocks after:
image

Mayland-Blocks before:
image

Mayland-Blocks after:
image

Quadrat before:
image

Quadrat after:
image

Fixes #3969

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -635,6 +635,7 @@ p.has-drop-cap:not(:focus):first-letter {
 .wp-block-quote.is-style-large p,
 .wp-block-quote p {
 	font-style: unset;
+	margin-top: 0px;
 }
 
 .wp-block-quote.is-style-large .wp-block-quote__citation,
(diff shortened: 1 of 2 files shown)


pbking:
I tried to fix this with the "elements" block like so but could not get it to effect anything (since it was a "p" element I was trying to effect I thought it would work...)
```
		"spacing": {
			"customPadding": true,
			"customMargins": true,
	...
			"core/quote": {
                                 ...
				"elements": {
					"p": {
						"spacing": {
							"margin": {
								"top": "0px"
							}
						}
					}
				}
			}
```

jffng (approved):
⛳

PR #3978, opened 2021-05-28
BB - Fix margin alignments in view and editor

pbking:
* Add selectors so that alignment utilities work in the editor.  
* Added `!important`s to utilities to overcome Gutenberg margin utilities.

This change adds a few selectors to the alignment classes to ensure that they work as expected in the editor.

It also adds !important to the alignwide alignment rules to ensure that they work despite Gutenberg's own addition of !important to the margins.

Before (view):
image

Before (editor):
image

After (view):
image

After (editor):
image

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -4,6 +4,7 @@
  */
 body {
 	margin: 0;
+	padding: 0;
 }
 
 body {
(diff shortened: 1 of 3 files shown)


pbking:
This has been tested with (and pairs nicely with) #3968

jffng (approved):
Screen Shot 2021-05-28 at 2 41 43 PM

PR #3985, opened 2021-05-31
Blockbase: separated mixins from rules for the button block

MaggieCabrera:

We had both mixins and CSS rules inside the button block scss partial, and this file was being imported by other blocks that use button mixins too, duplicating code every time they were imported. This PR separates the mixins to another file that's being imported where needed.


MaggieCabrera:
This change removes 645 lines of duplicated code without changing any functionality

pbking:
The refactor changes are great and remove a lot of redundant code!  Thank you.

In addition to the refactor I also removed unnecessary hover styles and defined a hover text color for buttons (recompiling that change into the children theme.json files as well).  This was needed because of how (seemingly) Chrome was handling missing CSS variables in that scenario and aren't missed because no themes were actually using it.  (It was more interesting than helpful.)

seedlet-blocks/theme.json
```diff
@@ -205,8 +206,8 @@
 			},
 			"post-content": {
 				"padding": {
-					"left": "var(--wp--custom--margin--horizontal)",
-					"right": "var(--wp--custom--margin--horizontal)"
+					"left": "20px",
```
jffng, line 209: Just noting that some compiled changes to theme.json are sneaking in here and that's okay.

jffng (approved):
Buttons + hover styles looking good in BB & co 👍 

Love the amount of duplicate code we're removing here.

seedlet-blocks/theme.json
pbking: If I just rebuild all children on the trunk I get those changes as well.  IMO they should come in sooner than later, should have already been in with previous changes.  All of the changes look correct, except in that they are unrelated to the change here and should have been made already.

PR #3987, opened 2021-06-01
Blockbase: add main tag (skip links)

### LINKED ISSUE #3981: Blockbase: Add a main element to wrap the template content
As a part of the [requirements for .org submission](https://themes.trac.wordpress.org/ticket/100060#comment:2), Blockbase needs to have a skip link. Luckily this [is done automatically](https://github.com/WordPress/gutenberg/pull/30336) for block themes, we just need a `main` element in the markup for the link to point to that is wrapping the content.

jffng:
This PR adds main tags (from which Gutenberg infers to add skip links).

Addresses #3981.


pbking (approved):
I see that skip links and targets are properly defined in the view now.  All templates are covered.  LGTM!

Issue #2449, opened 2020-09-10
Add support for Padding Control in Gutenberg


ianstewart:
Padding control for Group is on the way!

https://github.com/WordPress/gutenberg/pull/24966

It looks like it requires theme support …

> Enable the custom spacing support for your theme add_theme_support( 'experimental-custom-spacing' )

We usually run the latest version of Gutenberg on WP.com much earlier than core WordPress. Usually around 1 to 4 weeks after release.

Stubbing out an issue here to make sure we have this support in our Automattic Themes. As this may take a while I've made a list of what might be a priority order.

- All new themes.
- All Varia themes.
- All popular themes.
- All premium themes.
- All themes.

Generally, I think it makes sense to have all new themes support all new features on a rolling basis as a first priority for Gutenberg support. Though saying that I don't mean to suggest that older themes shouldn't get support. Just that we should always have at least one theme on the edge or one iron in the fire for testing.

Currently, only Seedlet and Sidekick have this support.

kjellr:
Just noting that this is in Seedlet already: 

https://github.com/Automattic/themes/blob/2b9e918c66e43be68f54c8a161727b17df2f56da/seedlet/functions.php#L258-L259

... but that it doesn't appear to be working as of my testing today.

scruffian:
This has been done in Seedlet and Varia. I'll keep it open so we keep it in mind for other themes.

danieldudzic:
This doesn't seem to belong in the Maintenance queue. I'm removing it for now.

jeffikus:
@ianstewart is this still an issue? If not let's close this out.

PR #3988, opened 2021-06-01
IP2: add padding control + line height

jffng:
This PR adds support for padding control and line height in IP2. 

Are there any other classic themes we want to do this for? I just picked IP2 because it's still quite actively used. cc @ianstewart 

Part of #2449.

### DIFF
--- a/independent-publisher-2/functions.php
+++ b/independent-publisher-2/functions.php
@@ -103,6 +103,12 @@ if ( ! function_exists( 'independent_publisher_2_setup' ) ) :
 		// Add support for full and wide align images.
 		add_theme_support( 'align-wide' );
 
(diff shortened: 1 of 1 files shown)


jeffikus:
We can check the most popular list. @ianstewart we could plan in a short spring for this and do a list of themes?

@jffng any test criteria for this to pass? Just test if the controls are present?

jffng:
@jeffikus yes — I tested with cover and group blocks.

jeffikus (approved):
LGTM! Ship it!

ianstewart:
Awesome. The list of free themes in this Calypso issue would be good:

https://github.com/Automattic/wp-calypso/issues/52757

Issue #2043, opened 2020-05-19
Varia & Varia Child Themes: Accessibility related issues in Comments


itowhid06:
Screen readers can’t read `<b>` tag content. So, when some important text is wrapped in `<b>` tags, it will come to no benefit for users using screen readers. We should use `<strong>` tags instead.

## Steps to replicate

1. Go to any posts with comments.
2. Inspect any comment author's name in Devtools.
3. See the author's name is wrapped in a `<b>` tag.

## Result

![image](https://user-images.githubusercontent.com/12055657/82280553-eac08200-99b0-11ea-86f4-3d02b5aaad0d.png)

As evident in the above screenshot the author's name is wrapped in a `<b>` tag. This is hapenning because, though there is a Comment Walker class named as `Varia_Walker_Comment` it is not used in `wp_list_comments` function. As a result the aforementioned function is using `Walker_Comment` Walker class from WordPress Core, which is producing those `<b>` tags. As mentioned in [https://core.trac.wordpress.org/ticket/50202](https://core.trac.wordpress.org/ticket/50202), `<b>` tags should be avoided in favor of `<strong>` tags for Accessibility reasons. Note that, there is also an extra badge printing after the comment author's name when `Varia_Walker_Comment` walker is used.

## Expected

 The author's name in a comment should be wrapped in a `<strong>` tag as evident in many core themes like **Twenty Nineteen** & **Twenty Twenty** themes.  Also Varia theme should use `Varia_Walker_Comment` walker.

jffng:
Thanks @itowhid06! Based on the discussion in the core ticket you linked above, it's not clear to me that `<strong>` needs to be used here for screen readers. 

So, I don't think introducing the comment walker here is worth the effort to refactor the styling and test across all Varia + child themes.

That said, we should remove the comment walker class from Varia as it is not used.

PR #3989, opened 2021-06-01
Varia: remove unused comment walker

jffng:
This PR removes a comment walker that does not appear to be in use anywhere. Discovered while looking into https://github.com/Automattic/themes/issues/2043.


pbking:
I see that it's [loaded from here](https://github.com/Automattic/themes/blob/trunk/varia/functions.php#L528).  I'm not sure if/how that means it's used but it doesn't appear to be a dangling file.  I tried to tinker with it but couldn't derive the purpose (which just means I don't know how it works, not that it isn't used. :P )  If it is indeed without purpose then we should remove the reference from function.php as well.

jffng:
True, I think we would want to remove this in two commits to be safe:

1. Remove the reference in functions.php
2. Delete the file.

pbking (approved):
👍

PR #3994, opened 2021-06-02
Fix group block padding when alignfull in varia

### LINKED ISSUE #3733: Varia + children: Text inside wide columns blocks extends to the edge of the screen on mobile
## Description

When there's text inside of a wide or full column block, the test bumps up against the edges of the screen on mobile. There are a couple things happening here: 

- Seedlet (and its children) and Twenty Twenty-One, do not push wide-width blocks to the screen edges on small screens. Varia does. 
- Varia also does not provide a fallback for text inside of columns, so that they don't bump up against the edges of the screen. 

You can replicate this by testing the pattern from https://github.com/Automattic/blockpatterns/issues/169. 

## Screenshots

Seedlet|Varia themes (Mayland)
---|---
Screen Shot 2021-04-30 at 11 18 54 AM|Screen Shot 2021-04-30 at 11 18 34 AM

pbking:
In the same way that the COLUMNS block is already adding padding when
configured as alignwide the GROUP block is now doing the same.

Fixes #3733 

Before:

After:

All varia children have been compiled and verified with these changes (screenshots not included).

### DIFF
--- a/alves/style-rtl.css
+++ b/alves/style-rtl.css
@@ -1707,6 +1707,11 @@ button[data-load-more-btn], .button {
 	}
 }
 
+.wp-block-group.alignfull {
+	padding-right: 16px;
+	padding-left: 16px;
+}
+
 h1, .h1,
 h2, .h2,
 h3, .h3,
--- a/alves/style.css
+++ b/alves/style.css
@@ -1707,6 +1707,11 @@ button[data-load-more-btn], .button {
 	}
 }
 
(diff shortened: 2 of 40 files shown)


jffng (approved):
Works in my testing, also tested with group blocks that are align full and have a background color.

Also in the issue title and description, you mean alignfull, not alignwide, right?

pbking:
Yup.  That's what I meant.  Updated.  Thanks. :)

PR #4001, opened 2021-06-03
First pass at a blockbase child theme generator

### LINKED ISSUE #3999: Blockbase "Getting started" suggestions
Hi folks! I spent a little bit of time this evening testing out Blockbase as a child theme, and I struggled a little bit during setup. Here are some issues that I ran into while trying it out: 

- A command line tool for generating a child theme would be wonderful. Today, our instructions require you to do some copy/paste from existing themes.
- Currently, we say the theme requires "package.json, style.css, and empty index.php files" for setup. But in practice, I needed a block template in block-templates before WordPress would recognize this as a block theme. In addition, the build scripts didn't work unless I'd already added a child-theme.json file to the child theme.
- Also — while it's probably obvious to some — the build steps require requires the child theme and `blockbase` to both to be in the same directory. They also require you to first run `npm install` in the both the `blockbase` and child theme folders. We should make that a bit more clear.

pbking:
Instructions are in the readme.txt

TL;DR
`npm run create:child pickles`

### DIFF
--- a/blockbase/create-child.js
+++ b/blockbase/create-child.js
@@ -1,0 +1,151 @@
+const child_theme_json = `{
+	"settings": {
+		"custom": {}
+	},
+	"styles": {
+		"blocks": {}
+	}	
+}`;
+
+const child_package_json = `{
+  "name": "{newtheme}",
+  "version": "0.0.1",
+  "description": "",
+  "main": "style.css",
+  "dependencies": {},
+  "devDependencies": {
+    "@wordpress/base-styles": "^3.4.0",
+    "chokidar-cli": "^2.1.0",
+    "node-sass": "^5.0.0",
+    "node-sass-package-importer": "^5.3.2"
+  },
+  "scripts": {
+    "start": "chokidar \\"**/*.scss\\" child-theme.json ../blockbase/theme.json -c \\"npm run build\\" --initial",
+    "build": "npm run build:theme && npm run build:scss",
+    "build:theme": "node ../blockbase/build.js {newtheme}",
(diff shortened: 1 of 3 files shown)


blockbase/create-child.js
```diff
@@ -0,0 +1,151 @@
...
+ * Enqueue scripts and styles.
+ */
+function newtheme_scripts() {
+	wp_enqueue_style( 'newtheme-styles', get_stylesheet_directory_uri() . '/assets/theme.css', array('blockbase-ponyfill'), wp_get_theme()->get( 'Version' ) );
+}
+add_action( 'wp_enqueue_scripts', 'newtheme_scripts' );
+
+`;
+
+const block_templates_index_html = `
```
scruffian, line 81: Can we copy these from Blockbase so that if they change we don't have to update here?

pbking: ...maybe.
The files for a child theme aren't the same files blockbase uses (or a derivative theme would use).
If we wanted to copy files instead of have them in this script then we would still need to copy files that aren't the ones blockbase uses, and would still have to change them independently... so it's just moving the contents from one place to another.  Which may be better! But I don't think it addresses "if they change we don't have to update the generation".

scruffian: I just meant the block templates, not the other files. Are the block template different?

pbking: Yea, they are more descriptive than functional.
`<span>Put yer templates here.  Here's an example of where to put 'em and how to use template parts</span>`

scruffian (approved):
In that case I vote merge

PR #4013, opened 2021-06-06
Seedlet-blocks: Update text domain

RavanH:
In inc/block-styles.php the old text domain 'seedlet' is still used. Update it to the theme's own 'seedlet-blocks' ...


pbking (approved):
Thank you.

Issue #756, opened 2019-04-30
Twenty Nineteen (And Child Business Themes): Comment Input Issues


JoshuaGoode:
## Steps to replicate

1. Set up a site to allow non-WPCOM users to comment on posts.
2. Utilize the Twenty Nineteen or [Child Name] Business themes.
3. Visit a post on the site in an incognito window or while logged out of WPCOM.
4. Attempt to input a comment.

## Result

The inputs will appear improperly.

![image](https://user-images.githubusercontent.com/4887830/56989144-c56f9480-6b5f-11e9-9477-16395bcc45ed.png)

## Expected

Proper form input.

![image](https://user-images.githubusercontent.com/4887830/56989211-efc15200-6b5f-11e9-9296-4e9f04df0a7c.png)

* User report: 12462117-hc
* Follow-up ticket: 1998908-zen
* Discussion: p1556651138235400-slack-themes

Reproduced on the following:
* Twenty Nineteen
* Professional Business
* Sophisticated Business
* Elegant Business
* Friendly Business
* Calm Business
* Modern Business

Opening issue for further investigation and to see if it's a conflict with Jetpack.

fresatomica:
I've got another user report: 2142223-zen

scosgro:
Another user report: 14164166-hc

tijosh:
Another report of this in #2363222-zen

User noticed on iPads

It looks like this does not affect mobile at all—only medium and desktop screens

reginabally:
Another report of this in #15232750-hc for Twenty Nineteen theme. User reported this on a desktop screen and I'm able to replicate this on my test site.

tbradsha:
Reported in 3345341-zen (free WordPress.com site) and confirmed on my test site.

tbradsha:
As a workaround, this CSS should override the errant float:

```css
@media only screen and (min-width: 768px) {
  .comment-form .comment-form-author, .comment-form .comment-form-email {
    float: unset;
  }
}
```

serabi:
Reported in 27274967-hc - user would also like to be contacted when issue is fixed.

pbking:
Fix can be found here: c-D62215

Note that in the above change there appear to be additional changes in style-rtl.css which (from what I can tell) are simply because changes to the .scss were not compiled into that file before.

Before:
![image](https://user-images.githubusercontent.com/146530/120537171-fb247e80-c3b2-11eb-94c8-67ebd4e7552b.png)

After:
[...]

PR #4020, opened 2021-06-07
Business themes: Fix comments on wpcom

scruffian:
These rules break the layout of the comments form on wpcom. By adding the not we fix the issue although it's not ideal to have wpcom specific code in our theme. You'll need this diff to test: D62425-code

#### Related issue(s):
https://github.com/Automattic/themes/issues/756

### DIFF
--- a/calm-business/package-lock.json
+++ b/calm-business/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "calm-business",
-  "version": "1.6.1",
+  "version": "1.6.2",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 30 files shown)


danieldudzic:
I'm struggling to test this, as I get a blank white page when I enable any of the Business themes. Since this CSS change is quite non-invasive, I don't see a problem with adding it.

pbking:
The CSS changes here check out.

However the diff noted has some unneeded language changes in the diff that should be removed before merging.

pbking:
Note that the changes are actually in this change: D62215-code

PR #4023, opened 2021-06-08
Add a blank customTemplate to blockbase

### LINKED ISSUE #3895: Blank Canvas Blocks: Turn into a more generic "starting point" theme
👋 I took a look at Blank Canvas Blocks with the idea of turning it into a more generic starting point for a block-based/universal theme. Here are the steps I think we should take to get there: 

- [x] **Header/Navigation Template Parts:** Replace with the slim ones used in Quadrat. 
- [x] **Index template:** Left-align all post titles, add a featured image block below the title. 
- [x] **Current Singular Template:** Make the current "post content only" singular template into a custom page template instead, so folks can optionally choose to use start blank for certain pages. (I'm not sure we need to do this given the template stuff coming in 5.8?)
- [x] **New Singular Template:** Add a new singular template with the header template part, post title, wide-width featured image, post content, and post date. Also include comments + the page footer template part. 
- [x] **Block Patterns:** Remove all bundled block patterns.
- [x] **New Theme Screenshot:** Provide a new theme screenshot to reflect these updates.

We'll likely need to go back in and adjust some margin/padding after these updates, but we can handle that when we get there. 

Rough screenshots of the new template updates: 

Index|Singular
---|---
Screen Shot 2021-05-21 at 10 42 27|Screen Shot 2021-05-21 at 10 48 11

kjellr:
Will close #3895. 

This PR adds a blank [`customTemplate`](https://developer.wordpress.org/block-editor/how-to-guides/themes/theme-json/#customtemplates) to blockbase. 

Once added, you'll see this as an option in the post/page editor like so: 

templates-editor

... and also in the site editor here: 

templates

I can't figure out why the title is showing as lowercase — changing the `name` field doesn't seem to fix it, so maybe that's a Gutenberg bug?

### DIFF
--- a/blockbase/theme.json
+++ b/blockbase/theme.json
@@ -10,6 +10,16 @@
 			"area": "footer"
 		}
 	],
+	"customTemplates": [
+		{
+			"name": "blank",
+			"title": "Blank",
+			"postTypes": [
+				"page",
+				"post"
+			]
+		}
+	],
 	"settings": {
 		"border": {
 			"customColor": true,


pbking:
Works as explained.  Found the workflow a bit confusing since selecting the blank template for a pate/post resulted in an error message (The template is blank).  I'm not a very good judge of that though since I barely understand the workflows in question so will defer to others if the behavior is more helpful than confusing.

It does do all the things the pictures show it doing though. :)

Issue #3520, opened 2021-03-24
Varia children, Blank Canvas, Seedlet, Spearhead - Mobile menu same page jump links don't work


mikeicode:
Certain themes have a mobile menu that overlays the content of the page. Page jumps to anchor links on the same page don’t work within the mobile menu. The URL changes in the browser bar but the menu doesn’t close and the page doesn’t scroll down.

## Steps to replicate

* Activate Hever on a test site
* Add a heading lower on a page with an anchor link. 
* Create a custom link in the menu to this anchor. `https://example.blog/#anchor`
* Test the link on the desktop and mobile menu (from the same page the anchor is on)

## Result

On mobile the URL changes but the menu doesn’t close and the page doesn’t scroll down.

Desktop:

![https://d.pr/i/GGPrAc](https://d.pr/i/GGPrAc+)

Mobile:

![https://d.pr/i/oITxS7](https://d.pr/i/oITxS7+)

## Expected

Menu should close and jump down the page like desktop.

## Notes

Tested Firefox and Chrome on Mac, Safari on iOS.

I’ve found this issue on the following themes:

`Hever, Barnsbury, Dalston, Mayland, Rivington, Alves, Morden, Blank Canvas, Seedlet, Spearhead`

The main Varia parent theme as well as these Varia children themes have different styles for the mobile menu and the scroll does work correctly. 

`Maywood, Balasana, Shawburn, Exford, Rockfield, Stratford, Coutoire, Stow, Leven, Brompton, Redhill`

In case it helps `Twenty Twenty-One` has a similar overlay style for the mobile menu but it does function correctly for the anchor links.

This only happens for jump links on the same page. They work correctly when you try the same link from another page.

* User report: 3840299-zd-woothemes

As a temporary workaround I found that if you create a short link https://bitly.com of the URL `https://example.blog/#anchor` and use that as your custom link it will work.

goblinartificer:
3897030-zen as well, using Barnsbury

rachelwinspear:
Another report #3901556-zen - Hever theme

lizswafford:
Another report #3934935-zen, using Barnsbury

Robertght:
Another one in #21701497-hc using Hever

isocialtish:
Another user report using Hever 29678721-hc

sajmes:
Another report here for Hever 29698414-hc (same user as [this comment above](https://github.com/Automattic/themes/issues/3520#issuecomment-844301943)). 

[...]

PR #4024, opened 2021-06-08
Themes: Fix anchor link behavior on Varia/Seedlet child themes

mreishus:
## Changes

As described in https://github.com/Automattic/themes/issues/3520,
certain child themes of Varia and Seedlet were not handling menu links
correctly when they were navigating to anchor IDs on the page the user
was currently on. An existing bit of logic was already implemented in
the twentytwentyone theme to address this use case, so this diff ports
it into the asset scripts for Varia and Seedlet as well.

## Ported

Ported from D62296-code

## Testing Instructions

- Check out this diff to your sandbox.
- On a sandboxed test site, set your theme to Seedlet (or Spearhead, which is a Seedlet child theme and uses the same navigation script).
- Create a page on your test site with a large amount of content. Somewhere down the page (far enough down that you would have to scroll to it), add a header block with an anchor tag (i.e. `<h3 id="my-anchor-tag">Header Content</h3>`)
In the Customizer, add a custom menu item to the main site navigation which links to the anchor hash.
- Resize your browser to trigger mobile styling on your test site (or make use of Chrome dev tools mobile view.
- Navigate to your test page, then open the menu and click the menu item linking to the anchor hash.
- Verify that the menu closes and the page scrolls to the expected place.
- Repeat the above with any or all of the following themes, which use the equivalent change made to the Varia parent theme: Hever, Barnsbury, Dalston, Mayland, Rivington, Alves, Morde

### DIFF
--- a/seedlet/assets/js/primary-navigation.js
+++ b/seedlet/assets/js/primary-navigation.js
@@ -28,6 +28,19 @@
 			};
 		}
 
+		document.addEventListener( 'click', function( event ) {
+			// If target onclick is <a> with # within the href attribute
+			if ( event.target.hash && event.target.hash.includes( '#' ) ) {
+				wrapper.classList.remove( id + '-navigation-open' );
+				wrapper.classList.remove( 'lock-scrolling' );
+				// Wait 550 and scroll to the anchor.
+				setTimeout(function () {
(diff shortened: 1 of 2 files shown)


kwight (approved):

scruffian:
Thanks for doing this :)

PR #4031, opened 2021-06-09
Add language files from dotcom

scruffian:
This copies the changes to language files from dotcom to GH so that its easier to keep the repos in sync.


scruffian:
Thanks for the additions to this @pbking. I rebased once #4033 merged so the .gitignore changes disappeared. I think we can merge this.

PR #4034, opened 2021-06-10
Add images from SVN

scruffian:
When we sync images from GitHub to SVN we always seem to get them listed in the diff even when they haven't changed. This pulls the images back from SVN to GitHub in the hope that they will now match!


pbking:
So far mine have matched.  This seems to be helping.  🤞

PR #4035, opened 2021-06-10
Blockbase: Add post content to the blank template

kjellr:
Update to #4023. 

Prompted by @pbking's [comment](https://github.com/Automattic/themes/pull/4023#issuecomment-858031985), I realized that this template shouldn't actually be totally blank: it should display the post content. 😅 

This PR adds that in.


pbking (approved):
This helps make the workflow make more sense to me. :)

Not BLANK-blank.  Just... the right AMOUNT of blankness.

PR #4039, opened 2021-06-10
Updates to Mayland Blocks and Seedlet Blocks so that they are compatible with the correct version of WordPress

scruffian:
Needed for theme review.


pbking (approved):
LGTM

PR #4056, opened 2021-06-14
Quadrat: Add wp.com Font Picker support [Refactor Blockbase Font Family handling to support]

pbking:
### Details

This change allows the wp.com Font Picker to work with the Quadrat theme.  

Additionally it refactors Blockbase and children in order to simplify the handling of Font Faces in the configurations.  Some font-related housekeeping was also done.

#### Known Issues

Editor does NOT SHOW updated font choices (either upon load or in real time). In order for variables to flow to enable this functionality one of two things will need to happen:

* The Font Picker will need to target body instead of .editor-styles-wrapper in the editor
OR
* The Global Styles variables will need to target .editor-styles-wrapper instead of body

This is because at the level at which the font family variable is first consumed (body) the value of --font-face isn’t defined so the  browser moves on to the alternate value provided. Even though the variable IS defined further down the stack, it’s not defined THERE so it’s ignored in that context henceforth.  Even though the value of --font-face IS defined in the context in which it’s used (any text element in the .editor-styles-wrapper element) it’s still ignored.

I’m not sure which of these to further explore or recommend.   I don’t believe any themes leveraging Global Styles to define fonts will be able to use the .com Font Picker until that’s resolved in some way.  There may be other situations where other plugins or tools are trying to make use of Global Styles variables so this situation seems likely to happen in other scenarios.  

#### Refactor Details

Prior to this change there was a bit of a mix between "semantic" and "descriptive" font faces in the Blockbase & Co. themes.  This change standardizes on BOTH concepts with "descriptive" font families being supplied in the --preset (and thus provided to the user with meaningful names) and "semantic" font families being defined in the "custom" block and distributed to the blocks (or CSS).

In this way a theme (such as Quadrat) can tap into those two locations to introduce the `var(--font-base)` and `var(--font-headings)` variable usage for the Font Picker (which it does in this change).

A number of other font-related house cleaning was done as well eliminating any duplicate/redundant settings.

Note that prior to this change Seedlet-blocks DID have SOME (-ish) support for the Font Picker in that it would consume the font values assigned (though with no way to change it).  This change REMOVES the usage of the Font Picker `var(--font-base/--font-headings)` from Seedlet (though adding FULL support for the Font Picker for any of the Blockbase children would be a simple thing... though [at least for now] I suggest ONLY supporting that in "Universal" themes such as Quadrat.

### Testing

Either test these changes on a sandbox or install the Editing Toolkit plugin.  
Activate Quadrat and edit any post or page.
Note that the Font Picker icon is active.  Select it.
Change the BASE and HEADER font families and publish the changes.
(Note the bug noted above; currently you will NOT see those changes in the Editor).
View the site.  Note that the headings and body font families match the selections made in the editor.
Activate Blockbase, Seedlet-Blocks and Mayland-blocks
Ensure that the default font-faces show as expected and are not influenced by the typeface selection in the Font Picker.

### Screenshots

Editor:
image

View:
image

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -176,7 +176,7 @@ textarea {
 	border-radius: var(--wp--custom--form--border--radius);
 	box-shadow: var(--wp--custom--form--color--box-shadow);
 	color: var(--wp--custom--form--color--text);
-	font-family: var(--wp--preset--font-family--base);
+	font-family: var(--wp--custom--body--typography--font-family);
 	padding: var(--wp--custom--form--padding);
 }
 
@@ -334,8 +334,8 @@ input[type=checkbox] + label {
 
 ul,
 ol {
-	font-family: var(--wp--custom--list--font-family);
-	padding-left: var(--wp--custom--list--padding--left);
+	font-family: var(--wp--custom--list--typography--font-family);
+	padding-left: var(--wp--custom--list--spacing--padding--left);
 }
 
 .wp-block-navigation.is-responsive .wp-block-navigation__responsive-container.is-menu-open {
--- a/blockbase/sass/blocks/_list.scss
+++ b/blockbase/sass/blocks/_list.scss
@@ -1,5 +1,5 @@
 ul,
 ol {
-	font-family: var(--wp--custom--list--font-family);
-	padding-left: var(--wp--custom--list--padding--left);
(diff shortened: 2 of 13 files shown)


pbking:
Closing this with preference to https://github.com/Automattic/themes/pull/4056 which does most of this change without actually enabling the Font Picker.  (But allowing for that actual change to be very small, or any other change to be cleaner whatever that change might be.)

Issue #4051, opened 2021-06-11
Quadrat: Font customizations


scruffian:
We should look for a way to allow users to modify fonts in Global Styles.

See: pNEWy-dYW-p2#comment-53274

PR #4061, opened 2021-06-15
Blockbase: Add all fonts

scruffian:
This is an alternative approach to https://github.com/Automattic/themes/issues/4051

By loading a large selection of fonts into Blockbase we make all these available to the child themes too.

One problem with this approach is that we load all fonts regardless of which ones you use. The request is pretty small so maybe this is ok. In some ways this could be better as if we always load the same URL for fonts on all themes, it will be cached so it will be faster across sites than having different font packages per theme/site and loading different ones each time.

We should make it possible for child themes to add additional fonts.

To test, load the site editor and change fonts.

#### Related issue(s):
#4051

### DIFF
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -65,16 +65,17 @@ function blockbase_fonts_url() {
 	}
 
 	$theme_data = WP_Theme_JSON_Resolver_Gutenberg::get_merged_data()->get_settings();
-	if ( empty( $theme_data ) || empty( $theme_data['custom'] ) ) {
+	if ( empty( $theme_data ) || empty( $theme_data['typography'] ) || empty( $theme_data['typography']['fontFamilies'] ) || empty( $theme_data['typography']['fontFamilies']['theme'] ) ) {
 		return '';
 	}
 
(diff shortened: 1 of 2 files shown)


blockbase/functions.php
```diff
@@ -65,16 +65,17 @@ function blockbase_fonts_url() {
...
 	}
 
-	$custom_data = $theme_data['custom'];
-	if ( ! array_key_exists( 'fontsToLoadFromGoogle', $custom_data ) ) {
-		return '';
+	$font_list = $theme_data['typography']['fontFamilies']['theme'];
+	$font_families = [];
+	foreach( $font_list as $font ) {
+		if ( 'system-font' !== $font['slug'] ) {
+			$font_families[] = 'family=' . str_replace( ' ', '+', $font['name'] ) . ':ital,wght@0,400;0,500;0,700;1,400;1,500;1,700';
```
MaggieCabrera, line 76: Do these weights and styles apply to every single font family?

scruffian:
This generated some interesting discussion, but I'll close for now while we think more about our options here.

PR #4070, opened 2021-06-17
Quadrat: Simplify templates

scruffian:
This makes a few changes to our template files to simplify things.

Now each HTML template file is just a one liner that includes the corresponding template part. This means our PHP templates can be much simpler. It also means that we can modify the top level templates in the Site Editor and then switch back to classic mode and our edits will be kept.

It also implements the new `__unstable__location` attribute from Gutenberg so that we only need one header template part.

To test:
- Check that the site loads correctly still
- Try adding menu to the primary location and check that it shows up
- Add an index.html template file with this content: `<!-- wp:template-part {"slug":"index"} /-->` and check that everything works as expected in the site editor.

### DIFF
--- a/quadrat/404.php
+++ b/quadrat/404.php
@@ -14,12 +14,6 @@
  */
 get_header();
 
-	// the header
-	get_template_part( 'template-parts/header' );
-
 	echo gutenberg_block_template_part( '404' );
-
-	// the footer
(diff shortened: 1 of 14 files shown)


MaggieCabrera:
I love this. We just need a flag and it would be perfect!

MaggieCabrera (approved):

PR #4072, opened 2021-06-18
added tagline to header templates

### LINKED ISSUE #4059: Quadrat: Make tagline visible
When I write a tagline in the Customizer it isn't visible neither on the editor or in front end.

Captura de ecrã 2021-06-14, às 12 36 28

It should look like this in 15px size:
Captura de ecrã 2021-06-14, às 12 47 41
Captura de ecrã 2021-06-14, às 12 47 54

MaggieCabrera:

Added tagline to the header of Quadrat.

With logo:

Desktop | Mobile
--- | ---
Screenshot 2021-06-18 at 10 45 59 | Screenshot 2021-06-18 at 10 45 49

Without logo:

Desktop | Mobile
--- | ---
Screenshot 2021-06-18 at 10 46 17 | Screenshot 2021-06-18 at 10 46 26

#### Related issue(s):

Closes #4059

### DIFF
--- a/quadrat/sass/templates/_header.scss
+++ b/quadrat/sass/templates/_header.scss
@@ -31,6 +31,15 @@
 		padding-right: 0;
 	}
 
+	.wp-block-site-tagline {
+		margin: 0;
+		@include break-mobile-only() {
+			padding-left: 0 !important;
+			flex-basis: 100%;
+			order: 10;
+		}
+	}
+
 	&:before {
 		content: "";
 		background-color: var(--wp--custom--color--secondary);
--- a/quadrat/sass/templates/_header.scss
+++ b/quadrat/sass/templates/_header.scss
@@ -3,14 +3,14 @@
 	overflow: inherit;
 	padding: 10px var(--wp--custom--post-content--padding--left) 60px; // TODO: Maybe replace with a responsive custom variable?
 
-	@include break-mobile() {
+	@include break-small() {
 		padding: var(--wp--custom--post-content--padding--left);
 	}
 
 	.wp-block-site-logo {
 		margin-right: var(--wp--custom--margin--horizontal);
 
-		@include break-mobile-only(){
+		@include break-small-only(){
 			flex-basis: 100%;
 			margin: 20px 0;
 			text-align: center;
@@ -33,7 +33,7 @@
 
 	.wp-block-site-tagline {
 		margin: 0;
-		@include break-mobile-only() {
+		@include break-small-only() {
 			padding-left: 0 !important;
 			flex-basis: 100%;
 			order: 10;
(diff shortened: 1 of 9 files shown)


beafialho:
This looks good to me :+1:

MaggieCabrera:
I'm having issues with this after rebasing, looking into it.

quadrat/sass/templates/_header.scss
```diff
@@ -31,6 +31,15 @@
 		padding-right: 0;
 	}
 
+	.wp-block-site-tagline {
+		margin: 0;
+		@include break-mobile-only() {
```
scruffian: Could we do this with intrinsic CSS, so that it wraps when it needs to rather than at a specific point?

MaggieCabrera:
This needs more work. Since the header is using flexbox, the layout breaks if no menu is defined, for example.

quadrat/sass/templates/_header.scss
pbking: I tinkered with that, and agree that would be desirable, but I wasn't able to figure out how.  It's not just a matter of wrapping, but the design is a bit different in the two scenarios.  The Navigation block is already switching to mobile design (using the same Gutenberg breakpoint) so it doesn't feel wrong.

pbking: Actually... upon further evaluation I see that the navigation block changes styles at SMALL rather than MOBILE breakpoint... Refactoring slightly to break at the same point for the rest of the quadrat header.

pbking:

Made a small tweak so that this should look OK even if no menu is present.  (Note: [This Gutenberg fix](https://github.com/WordPress/gutenberg/pull/33175) is necessary if you don't have a menu assigned.  At least it was for me.)

In my testing I found that the navigation block uses the 'small' breakpoint instead of 'mobile' (which is, frustratingly and confusingly just a few pixels larger).  Using a different breakpoint created strange behavior for about 50px's worth of size between the two.  This change includes all of the 'mobile' breakpoint changed to 'small' where appropriate.  (I think)

MaggieCabrera:
This LGTM, thank you for finishing this Jason. I'd like to have extra 👀 on this before merging since I started work on this PR originally

mikachan:
This is looking good to me. I've browser tested both with and without a logo and a menu.

The only potential issue I can see is if someone were to upload or crop a logo that isn't square - or is wider than it is tall. If the screen resolution is between 600-668px width and there is no menu, it will display like this:

image

I think we could fix this by forcing the dimensions in the custom logo function to be square, as they're currently set to a height of 35 and a width of 150 in `functions.php`. This means the default cropping area is in these dimensions, rather than a square. It seems to default to a square if I change these dimensions in the `custom-logo` support function to something like this:

```
add_theme_support(
	'custom-logo',
	array(
		'height'      => 150,
		'width'       => 150,
		'flex-width'  => true,
		'flex-height' => true,
	)
);
```

However, this only seems to be affecting the 600-668px screen size, not sure how important this is!

pbking:
Thanks for the solid evaluation @mikachan. I tested around RE: the above behavior and was able to reproduce with a square logo and a long site title and/or tag line.  Anything that makes all 3 elements exceed the page width causes it.

Considering that there are other ways to achieve that situation I don't believe that we should target only one of them.  And IMO the wrapping that it does is about the most appropriate way to handle the situation that I can think of so I believe it's acceptable.

PR #4075, opened 2021-06-18
Blockbase: Add template with just a header and footer

kjellr:
@ianstewart Recommended having a set of three post/page templates bundled for block themes: 

Option A: Header and Footer Only|B: Blank|C: Default 
---|---|---
Option-A|Option-B|Option-C

Blockbase already has B and C, so this PR adds in that first one. I called it "Header and Footer Only" for now, but I'm definitely open to better names. 🤔 

## Screenshots

Screen Shot 2021-06-18 at 9 28 53 AM
Screen Shot 2021-06-18 at 9 29 10 AM
Screen Shot 2021-06-18 at 9 30 48 AM


scruffian (approved):
LGTM

PR #4076, opened 2021-06-21
Blockbase: Add a link to WordPress.com to the blockbase theme footer credits.

ramonjd:

We append WPCOM footer credits to block-based site frontends for free accounts.

See D59022-code for where we introduced this.

If the site already **has a link** to WordPress.com or WordPress.org however we won't append the footer.

This PR adds a link to the default theme footer template so new sites created using this theme won't show the additional footer.

#### Before
Screen Shot 2021-06-21 at 10 20 04 am

#### After
Screen Shot 2021-06-21 at 10 21 34 am

### Related issue(s):

See pbAok1-2d9-p2#comment-4500 for context.

### DIFF
--- a/blockbase/block-template-parts/footer.html
+++ b/blockbase/block-template-parts/footer.html
@@ -1,7 +1,7 @@
 <!-- wp:group -->
 <div class="wp-block-group">
 	<!-- wp:paragraph {"align":"center","fontSize":"tiny"} -->
-	<p class="has-text-align-center has-tiny-font-size">Proudly Powered by <a href="https://wordpress.com" rel="nofollow">WordPress</a></p>
+	<p class="has-text-align-center has-tiny-font-size">Proudly Powered by <a href="https://wordpress.org" rel="nofollow">WordPress</a></p>
 	<!-- /wp:paragraph -->
 </div>
 <!-- /wp:group -->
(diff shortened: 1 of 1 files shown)


blockbase/block-template-parts/footer.html
```diff
@@ -1,7 +1,7 @@
 <!-- wp:group -->
 <div class="wp-block-group">
 	<!-- wp:paragraph {"align":"center","fontSize":"tiny"} -->
-	<p class="has-text-align-center has-tiny-font-size">Proudly Powered by WordPress</p>
+	<p class="has-text-align-center has-tiny-font-size">Proudly Powered by <a href="https://wordpress.com" rel="nofollow">WordPress</a></p>
```
scruffian: From looking at the diff, I believe this can also be a link to wordpress.org, which I would prefer since it means we can also submit the same code to the dotorg repo.

ramonjd: Ah good call. Yes, the regex supports .org as well. I'll change. Thanks!

scruffian:
LGTM

scruffian (approved):

scruffian:
Did you sync this to dotcom?

ramonjd:
> Did you sync this to dotcom?

Sorry, got pulled sideways today. Thanks for the ping.

Here's the patch D63278-code

PR #4077, opened 2021-06-21
Refactor Blockbase font definitions

pbking:
This PR takes the important bits of #4056 as it relates to cleaning up Blockbase and child fonts to standardize Font definitions as such:

* Semantic Font definitions are defined in the "custom" block for distribution through the theme.  These are: `custom.heading.typography.fontFamily` and `custom.body.typography.fontFamily`.  Blocks use these values throughout.

* Descriptive Font definitions are defined in the "settings" block.  These Font Families are provided to the user with names that match the actual font and are assigned to the two semantic font values described above for consumption by the themes.

Not all of the themes followed the same naming structure (especially Seedlet-blocks) and this change brings all of the themes into alignment around that.

This change EXCLUDES the changes from #4056 that allow the wp.com Font Picker to be used with Quadrat.  (The exact future of font selection mechanism is still unknown.)

Additionally there are a couple of attempts to normalize some of the variable definitions using a standardized naming structure (changes to the list block variables for example) and simplifications (reducing the number of variables used when their use seemed to unnecessarily complicate things such as the line height values for headers.)

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -176,7 +176,7 @@ textarea {
 	border-radius: var(--wp--custom--form--border--radius);
 	box-shadow: var(--wp--custom--form--color--box-shadow);
 	color: var(--wp--custom--form--color--text);
-	font-family: var(--wp--preset--font-family--base);
+	font-family: var(--wp--custom--body--typography--font-family);
 	padding: var(--wp--custom--form--padding);
 }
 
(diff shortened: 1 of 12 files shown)


scruffian:
I'm not sure about this approach - I don't know if the abstraction is helpful here since users won't be able to modify custom in Global Styles - I'm not sure what benefit it provides.

pbking:
The issue present (as I see it) is that there are a couple of different flavors of font family abstractions done a couple of different ways across the themes.  The abstraction of "base" and "header" fonts is already there and while I'm fine if we move away from that I would rather do that with all the themes in lock-step as to how those things are accomplished.  Perhaps there's a better, different destination where we'll end up; This is just an attempt to make sure they all work the same first.

I agree this isn't a final solution.  Perhaps it might be better to completely eliminate ANY abstraction with this change...  Regardless a consistent usage of the "style.fontFamily" space should be done across BB & Co. themes.

scruffian:
Ok thank makes sense, lets just merge and iterate :)

PR #4088, opened 2021-06-23
Refactored blockbase color configuration from 'descriptive' to 'semantic

pbking:
NOTE: Child themes have NOT YET been refactored to support this change.  Currently it's blockbase only.

UPDATE: Quadrat has been updated.  (And has a continued experiment with editing these semantic colors and storing them in the CPT [here](https://github.com/Automattic/themes/pull/4090).)

Before:
image

After:
image

### DIFF
--- a/blockbase/theme.json
+++ b/blockbase/theme.json
@@ -43,12 +43,6 @@
 					"name": "Secondary"
 				},
 				{
-					"slug": "tertiary",
-					"color": "#FFFFFF",
-					"color-value-i-would-like-to-use": "var(--wp--custom--color--white)",
-					"name": "Tertiary"
-				},
-				{
 					"slug": "foreground",
 					"color": "#333333",
 					"color-value-i-would-like-to-use": "var(--wp--custom--color--almost-black)",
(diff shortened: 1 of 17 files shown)


pbking:
One thing to point out about this method is that for the two semantic colors we already had - foreground and background - we now have TWO places to change the SAME THING in two DIFFERENT WAYS.

Note how I could change the "Text Color" which would change MOST instances of text, (but not all, many places use the variable `--wp--preset--color--foreground` and those wouldn't change).

I could also change the color value of "Foreground" would WOULD change the color used in all aspects (since everything, the "Text Color" included, are assigned the variable being changed).

blockbase/theme.json
```diff
@@ -31,29 +31,40 @@
 			"gradients": [],
 			"palette": [
 				{
-					"slug": "black",
+					"slug": "primary",
 					"color": "#000000",
-					"name": "Black"
+					"color-value-i-would-like-to-use": "var(--wp--custom--color--black)",
```
pbking, line 36: Because the color picker assumes a color value rather than a css variable this doesn't work as desired.  I THINK I saw some shorthand in other places to use values like these in a way that might make this work. 

Regardless the "custom" colors are where the "descriptive" colors live, though they aren't very useful right now.

scruffian:
Closing this as we don't need it anymore

PR #4090, opened 2021-06-23
Try/quadrat semantic color customizations

pbking:
This is a continuation of #4088 combined with #3856.

PLEASE SQUASH WHEN MERGING!

The result is that color customization of Blockbase and all child themes (most importantly Quadrat) are now available in the Customizer (complete with real-time preview).

The values changed there are the same values change in the Global Styles panel so the changes can be made from either (or both) locations.

Additionally the themes have been refactored slightly with the following:

All --preset colors are the SEMANTIC colors that are presented to the user to modify.  There could be any number of color values here: for instance Blockbase has 5 while Quadrat has 3.

These colors are mapped to --custom color values.  Here there are 5 colors that are used throughout the blockbase theme (primarily in theme.json but also some of the ponyfill.css reference --wp--custom--color values).  This allows the three colors expressed by Quadrat to cleanly map to the five colors expected by Blockbase.

(Note that it would be possible/reasonable for another theme to provide MORE than 5 colors to the user for customization. The child theme would simply need to handle the mapping of that value to whatever blocks or CSS properties are appropriate.)

The DEFAULT meaning of "primary" and "secondary" has been reshaped (somewhat) in blockbase thus allowing the configuration of "hover" colors for buttons.  These values are:

Foreground: All text that isn't otherwise noted
Background: The uh... the background of the page
Primary: Button Color, link color, border colors, etc
Secondary: Button Hover Color
Selection: The color text is when you select it

There's no reason why the values noted above need to be the same in all of the themes (Quadrat being a good example of diverging in a meaningful way).

image


quadrat/inc/customization.php
```diff
@@ -0,0 +1,135 @@
...
+
+		add_action( 'customize_register', array( $this, 'register_section' ) );
+
+		/* Customizer Preview JS */
+		add_action( 'customize_preview_init', array( $this, 'customize_preview_js' ) );
+		add_action( 'wp_enqueue_scripts', array( $this, 'create_customization_style_element' ) );
+	}
+
+	function create_customization_style_element() {
+		//NOTE: The original implementation wasn't quite this convelouted (there was no noop.css file, etc) but I couldn't get it to work otherwise.
```
MaggieCabrera:
```suggestion
		//NOTE: The original implementation wasn't quite this convoluted (there was no noop.css file, etc) but I couldn't get it to work otherwise.
```
Just a typo

scruffian (approved):
I made a few minor changes. This LGTM

blockbase/inc/customization.php
```diff
@@ -0,0 +1,127 @@
...
+	function get_theme_default_color_value( $slug, $palette ) {
+		foreach ( $palette as $palette_item ) {
+			if ( $palette_item['slug'] === $slug ) {
+				return $palette_item['color'];
+			}
+		}
+		return null;
+	}
+
+	function register_color_controls( $wp_customize, $palette ) {
```
MaggieCabrera, line 60: Nitpick, should this be named `register_color_section` ?

quadrat/theme.json
```diff
@@ -10,6 +10,24 @@
...
+		},
+		{
+			"name": "header-footer-only",
+			"title": "Header and Footer Only",
+			"postTypes": [
+				"page",
+				"post"
+			]
+		}
+	],
```
MaggieCabrera, line 30: I know these come from a previous PR but do we want these on the children themes?

MaggieCabrera (approved):
⛵

quadrat/theme.json
scruffian: I wondered that too. What do you think @kjellr?

PR #4098, opened 2021-06-25
Blockbase: Add support for custom palettes customization

MaggieCabrera:

This is a WIP. The goal is to give the user optional palettes provided by the theme's designers. Right now I'm using the code from https://github.com/HardeepAsrani/o2 as a starting point but this is a list of what I hope this will be able to do:

- [x] Get the default palette from theme.json and have it as the default option and available for users to go back to (instead of resetting each color one by one, they can reset the whole palette)
- [x] Grab the rest of the palettes from theme.json too. They should probably be under custom and we can provide labels for them there. Later, when Gutenberg supports mulitple plaettes, we can reuse those custom variables and assign them where appropiate for FSE goodness.
- [x] This PR doesn't handle the functionality that allows for the actual color updates yet.

Screenshot 2021-06-25 at 13 16 08


MaggieCabrera:
@kjellr if you have any input on this UI, I would appreciate your 👀

kjellr:
Thanks! We talked about this on slack, but noting it here for posterity: Let's try to adjust the visual treatment of the palettes to be similar to the Featured palettes treatment on WordPress.com: 

Screen Shot 2021-06-25 at 7 58 43 AM

MaggieCabrera:
Updated the styles of the palettes. I added a border color to mark the selected palette that the color annotations plugin does not have, since we are using radio buttons, I think it's important not to lose the checked value visual cue:

Screenshot 2021-06-25 at 14 29 40

scruffian:
I think this looks good to merge, but it would be good to have some more eyes on it.

blockbase/inc/wp-customize-color-palettes.php
```diff
@@ -0,0 +1,70 @@
...
+		foreach( $default_palette as $default_color ) {
+			$default_palette_setting[ $default_color['slug'] ] = $default_color['color'];
+		}
+
+		$this->palettes['default-palette'] = array(
+			'label' => 'Default',
+			'colors' => $default_palette_setting,
+		);
+
+		$custom_palettes = $theme_json['settings']['custom']['colorPalettes'];
```
pbking, line 32: I'm getting an error here w/ Quadrat due to there being no palette's currently defined in theme.json
Likely the Quadrat palette options weren't included in the change (yet).
But seems reasonable that a BB child might not ship with any customizations so probably a relevant use-case.

```
Notice: Undefined index: colorPalettes in /var/www/html/wp-content/themes/themes/blockbase/inc/wp-customize-color-palettes.php on line 33 Warning: Invalid argument supplied for foreach() in /var/www/html/wp-content/themes/themes/blockbase/inc/wp-customize-color-palettes.php on line 34
```

MaggieCabrera: yeah, that fixed disappeared on a conflict, let me get it back

pbking: almost...
```
Notice: Undefined index: colorPalettes in /var/www/html/wp-content/themes/themes/blockbase/inc/wp-customize-color-palettes.php on line 33
```

MaggieCabrera: should be fine now

pbking: Pushed [this small change](https://github.com/Automattic/themes/pull/4098/commits/1d411e3238b6b519335ca04ec9229deb17077ea5) to only render these controls if there are alternatives to default to show.  Also addressed that error that seemed to stick around despite the recent change.

blockbase/inc/wp-customize-colors.php
```diff
@@ -61,13 +61,15 @@ function register_color_controls( $wp_customize, $palette ) {
 
 		$section_key = 'customize-global-styles';
 
+		$theme = wp_get_theme();
```
pbking, line 64: I couldn't find where `$theme` was used.  Can this be removed?

blockbase/inc/wp-customize-color-palettes.php
```diff
@@ -16,56 +15,68 @@ function customize_preview_js() {
...
+					'label'       => __( 'Color Scheme', 'blockbase' ),
+					'description' => __( 'Choose a color scheme for your website.', 'blockbase' ),
+					'section'     => 'customize-global-styles',
+					'choices'     => $this->palettes,
+					'settings'    => 'color_palette',
 				)
 			)
 		);
+
+		$this->build_palettes( $theme_json );
```
scruffian: do we not need to call before we add the control?

pbking:
I'm of the opinion that the alternative palettes should be BUTTONS instead of RADIO BUTTONS. The action feels more like "apply these colors to my color pickers" rather than "select these colors to use" and BUTTON rather than RADIO BUTTON seems to reflect that better.

Otherwise we will invite the situation where an option might be chosen and then customized.  In this situation the RADIO BUTTON would need to be unselected (no option selected) otherwise that option couldn't be returned to without selecting something else and returning.  If we do keep it as a RADIO BUTTON then I suggest another stylistic mechanism to signify that that option is selected (It's difficult to see.)

That's just an opinionated nit-pick.  Otherwise this looks FANTASTIC and I would be happy to ship as-is and iterate.

blockbase/theme.json
```diff
@@ -116,6 +116,38 @@
 				"secondary": "var(--wp--preset--color--secondary)",
 				"selection": "var(--wp--preset--color--selection)"
 			},
+			"colorPalettes": {
+				"palette-1": {
+					"label": "Featured",
```
jffng, line 121: I think there is a mechanism now for labels for presets to get translated, but is there a way yet to mark that a custom variable needs to be translated?

jffng (commented):
Tested in all four themes and it is a thing of beauty 👏

I did run into an issue with Quadrat. Because of the way our merge / build step works for child theme, the `colorPalettes` from blockbase are added, which results in the following error after I build Quadrat's theme.json:
 
Screen Shot 2021-06-25 at 1 47 54 PM

If I provide a new palette with only the three colors per palette using the same palette slug (e.g. `featured-1`) and run build, I still get the above error because the merge appends the values to the colors. In this instance, I think what needs to happen is that the entire key `colorPalettes` object is replaced with the `child-theme.json` value, rather than the recursive merge.

scruffian:
> I did run into an issue with Quadrat. Because of the way our merge / build step works for child theme, the colorPalettes from blockbase are added, which results in the following error after I build Quadrat's theme.json:

I wonder if this means we should switch to using an array...

blockbase/theme.json
MaggieCabrera: I think we do it like [this](https://github.com/WordPress/gutenberg/pull/29828). Since we haven't done this for any of the strings that we have (such as the template parts) I think this merits its own separate PR

MaggieCabrera:
> I wonder if this means we should switch to using an array...

I think we kind of have to, it's what's consistent with what we've done for other variables like this

blockbase/inc/wp-customize-palette-control.php
```diff
@@ -0,0 +1,44 @@
+<?php
+/**
+ * Color Palette Customizer Control
+ *
+ * Props to https://github.com/HardeepAsrani/o2 go here
```
MaggieCabrera: What's the best way to do this?

scruffian: I'm not sure we need to do we, the code is pretty much a total change

MaggieCabrera (commented):
I tested Blockbase and selected one of the palettes, then I changed to Quadrat and the palette that was selected for Blockbase was showing up for the customizer too (while its colors showed up something else)

Screenshot 2021-06-28 at 11 49 34

scruffian:
> I tested Blockbase and selected one of the palettes, then I changed to Quadrat and the palette that was selected for Blockbase was showing up for the customizer too (while its colors showed up something else)

I think this is because the Global Styles settings weren't populating back into the customizer. I had to put back the settings file I removed because I couldn't find a way to set a dirty value.

scruffian (approved):

PR #4100, opened 2021-06-25
Remove Quadrat PHP Menu Renderer

pbking:
Now that the navigation block has the __unstableLocation attribute the
method of rendering the block is no longer needed and can be removed (as
well as the php header file that was calling it) since it all comes from
blocks now.

No visible changes neither of these files were used any more.


scruffian (approved):
LGTM

PR #4101, opened 2021-06-25
Add __unstableLocation to all BB&Co Themes

pbking:
As BB & Co. themes ALL become "Universal Capable" this addition allows
the menus to be modified (should these themes be used in "Classic Mode")

After (Blockbase):
image


scruffian (approved):
LGTM

PR #4104, opened 2021-06-25
Quadrat: Try adding header patterns

kjellr:
⚠️ Not sure we should merge this one, since the patterns here will duplicate [ones being explored in core](https://href.li/?https://github.com/WordPress/gutenberg/pull/31610). But I'm putting this here as a code demo. 

---

This PR adds four Header block patterns, to allow for convenient customizing of the site's header:

site-headers

### DIFF
--- a/quadrat/block-templates/index.html
+++ b/quadrat/block-templates/index.html
@@ -1,5 +1,5 @@
 <!-- wp:spacer {"height":10} -->
-<div style="height:10px" aria-hidden="true" class="wp-block-spacer"></div>
+<div style="height:10px" aria-hidden="true" class="wp-block-spacer site-header"></div>
 <!-- /wp:spacer -->
 
-<!-- wp:template-part {"area":"header","className":"site-header"} /-->
+<!-- wp:template-part {"area":"header"} /-->
--- a/quadrat/block-templates/index.html
+++ b/quadrat/block-templates/index.html
@@ -1,0 +1,5 @@
+<!-- wp:spacer {"height":10} -->
+<div style="height:10px" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- /wp:spacer -->
+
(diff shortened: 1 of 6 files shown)


quadrat/block-templates/index.html
```diff
@@ -0,0 +1,5 @@
+<!-- wp:spacer {"height":10} -->
+<div style="height:10px" aria-hidden="true" class="wp-block-spacer site-header"></div>
+<!-- /wp:spacer -->
+
+<!-- wp:template-part {"area":"header"} /-->
```
kjellr, line 5: This file is temporary, and only relevant for this demo — it sets up a template for testing the patterns.

jffng (commented):
Very cool to see the patterns and site editor working together like this! Thank you!

One small thing: when I actually add a menu, the links appear without padding. Not sure this is specific to this PR or the nav block itself, just noting for anyone else and in case there's something that can be done in the pattern to fix. 

Screen Shot 2021-06-25 at 3 22 46 PM

Screen Shot 2021-06-25 at 3 23 51 PM

As you mentioned to me on Slack, I think the site logo block encountering an error is a Gutenberg issue.

kjellr:
> One small thing: when I actually add a menu, the links appear without padding.

I've seen that sort of thing happen with Blockbase, so I think it might be a problem upstream. 

_Edit: Narrowed the problem down and filed an issue in the Gutenberg repo:_ https://github.com/WordPress/gutenberg/issues/33002

MaggieCabrera:
I think we should look into adding some of these on Blockbase to offer more options to users so they don't need to build them themselves (pNEWy-eA5-p2)

kjellr:
Definitely. Many of these have made their way into Twenty Twenty-Two, so I wonder if we can take a bunch of those header (and footer?) patterns and put them in Blockbase?

kjellr:
Closing due to inactivity. 🕐

PR #4108, opened 2021-06-28
Quadrat: fix featured images

### LINKED ISSUE #4062: Quadrat: Featured Image not showing up in pages
When I add a featured image, it doesn't show up on the front end, neither does the page name.

Editor | Frontend
------------ | -------------
Captura de ecrã 2021-06-15, às 15 41 02 | Captura de ecrã 2021-06-15, às 15 41 11

### LINKED ISSUE #4063: Quadrat: Make posts featured images content width
Currently, we have featured images on posts looking like this:

Frame 10

The problem is that squared images like the one above get so big that we have to scroll a lot to get to content, which is not cool. 

I had checked with @kjellr that we'd established having images be full width, but the solution didn't seem ideal either when I tried it, at least for posts:

DM Sans-1

So I suggest we keep featured images the same width as text, as exemplified below. They're still there but don't get in the way of content.

DM Sans

jffng:

This PR makes featured images appear the default width of the content in both posts and pages by default.

This means that if a user wants a blank page template, it will need to be created.

#### To Test

- Add / edit a page and post
- Add a featured image
- Verify that it appears the default content width, along with the page title

#### Related issue(s):

Addresses #4063 and #4062


quadrat/block-template-parts/page.html
```diff
@@ -1,5 +1,19 @@
 <!-- wp:template-part {"slug":"header"} /-->
 
+<!-- wp:group {"style":{"spacing":{"padding":{"top":"30px","right":"20px","bottom":"60px","left":"20px"}}},"layout":{"inherit":true}} -->
+<div class="wp-block-group" style="padding-top:30px;padding-right:20px;padding-bottom:60px;padding-left:20px">
+	<!-- wp:post-title {"textAlign":"center","level":1,"align":"wide"} /-->
```
jffng, line 5: @beafialho @kjellr what should the treatment for page titles be? For now I copied what we have for posts — center the text + wide alignment:

Screen Shot 2021-06-28 at 2 11 19 PM

kjellr: Keeping it the same as posts makes sense.

MaggieCabrera:
The desktop looks fine, but the mobile is not working great. I think it needs some design eyes @beafialho and @kjellr:

Screenshot 2021-06-29 at 12 39 34

kjellr:
> I think it needs some design eyes @beafialho and @kjellr:

This should be updated to match the display of featured images for single posts.

MaggieCabrera:
> This should be updated to match the display of featured images for single posts.

Like this?

Screenshot 2021-06-29 at 14 01 29

I'm asking because we are not only enabling the featured image on posts but we also changed the appearance and that's why it broke on mobile for both posts and pages.

kjellr:
Sorry for the confusion — in my testing, it's looking like this on mobile:

Archive Pages|Single Posts/Pages
---|---
Screen Shot 2021-06-29 at 10 24 44|Screen Shot 2021-06-29 at 10 24 48

We should just make those match (the archive page treatment is the correct one. 🙂) And then use the same treatment for featured images on page templates too. 

---

Also, this is a little different than the initial guidance in #4063, but on desktop I think we should make the single pages a little more distinct from archive pages. I suggest we keep the featured image width to the content width on archive pages, but we set it to `alignwide` on single posts and pages. @beafialho does that make sense to you too?

Archive Pages|Single Posts/Pages
---|---
Screen Shot 2021-06-29 at 10 30 17|Screen Shot 2021-06-29 at 10 30 05

beafialho:
@kjell yes, I agree. Wide is ok for Single Posts/Pages, as long as it's not almost full-width as described in #4063

jffng:
Ok, I made the following changes and this should be ready for another review:

- Removed the cool post header CSS 😞 
- Matched the treatment for featured images posts/pages and wide aligned them
- Tightened up the spacing a bit to match the archive treatment

Post Mobile | Post Desktop
-------- | --------
Screen Shot 2021-06-29 at 3 12 06 PM | Screen Shot 2021-06-29 at 3 20 47 PM

Page Mobile | Page Desktop
------- | -------
Screen Shot 2021-06-29 at 3 21 49 PM | Screen Shot 2021-06-29 at 3 23 03 PM

beafialho:
Thanks, looks good to me!

MaggieCabrera (approved):
LGTM

PR #4109, opened 2021-06-28
Fix Comments Styles for BB & Co.

### LINKED ISSUE #3546: Blockbase: Add default styles for comments block.

### LINKED ISSUE #4046: Quadrat: Update comment styles on WPCOM
## Steps to replicate

Comment styles on WPCOM need some love
image-1

From pNEWy-dYW-p2#comment-53274

pbking:
Fixes styles for the following situation for all BB&Co themes.

A comment block was added to Blockbase's 'singular.html' template.

Quadrat and Seedlet-blocks are using a CSS GRID for the comment form that was breaking the "highlander" commenting plugin on .com.  This was addressed by adding (.com-specific) styles reverting the display attribute from "grid".

Font size and lineheight configuration for post-comment block has been added to BB's 'custom' configuration block bringing the default size styling of the comment text to the same size as 'normal' text.

The `<small>Cancel reply</small>` element has been styled to the best of my ability.

Before:
image

The following situations have been reviewed for all 4 themes:
* .org Comment Form
image

* .com Comment Form
image

* .org Reply Form
image

* .com Reply Form
image

* .org Unauth Comment
image
* .com Unauth Comment
image

* .org Unauth Reply
image

* .com Unauth Reply
(I couldn't get this use-case setup...)

Fixes: #4046
Fixes: #3546


quadrat/sass/blocks/_post-comments.scss
```diff
@@ -142,3 +135,13 @@
...
+
+// NOTE: This is a wp.com-specific fix so that the comment form presented there (highlander)
+// is NOT displayed as a GRID.
+body.highlander-enabled {
+	.wp-block-post-comments {
+		form {
+			display: revert;
+		}
+	}
+}
```
MaggieCabrera: I think we should keep the wpcom only css to a separate stylesheet like we have [here](https://github.com/Automattic/themes/blob/trunk/seedlet/inc/wpcom-style.css) on Seedlet. If I'm not mistaken, that doesn't get uploaded to dotorg when we bundle the theme on its zip and we keep that separation

MaggieCabrera (dismissed):
Screenshot 2021-06-29 at 12 45 44

Looks like the label of the second checkbox has an extra margin. It's not the first time I see this and if it's not theme-related maybe the fix should be done on the highlander comment form.

quadrat/sass/blocks/_post-comments.scss
pbking: Beautiful.  I was wondering about that and why I commented as such.  I'll get those broken out.  Thanks @MaggieCabrera

pbking:
> Looks like the label of the second checkbox has an extra margin.

I saw that in all of the themes that I tested with (including Seedlet/Blank Canvas and Varia) so I don't think the themes should do anything about it and should happen on the highlander indeed.

quadrat/sass/blocks/_post-comments.scss
pbking: All moved.  And I learned how .com-specific additions work in the process.

quadrat/inc/wpcom.php
```diff
@@ -0,0 +1,19 @@
...
+
+
+/**
+ * Enqueue our WP.com styles for front-end.
+ * Loads after theme styles so we can add overrides.
+ */
+function quadrat_wpcom_scripts() {
+	wp_enqueue_style( 'quadrat-wpcom-style', get_stylesheet_directory_uri() . '/inc/wpcom-style.css', array( 'quadrat-styles' ) );
+}
+add_action( 'wp_enqueue_scripts', 'quadrat_wpcom_scripts' );
```
scruffian: Although this isn't used for Blockbase, I wonder if it would be better to add this to Blockbase so that we don't need to add it for all child themes...

scruffian:
I put at fix for that checkbox here: D63502-code

quadrat/inc/wpcom.php
scruffian: I did this.

jffng (approved):
Definitely better with this PR + @scruffian's patch applied: 

_Quadrat_
logged-in | logged-out
------- | --------
Screen Shot 2021-07-01 at 11 59 56 AM | Screen Shot 2021-07-01 at 12 04 24 PM

_Seedlet blocks_
logged-in | logged-out
------- | --------
Screen Shot 2021-07-01 at 12 01 57 PM | Screen Shot 2021-07-01 at 12 03 38 PM

PR #4116, opened 2021-06-29
Blockbase: Add font customization

scruffian:
This adds font customizations to Blockbase and therefore all child themes, for example Quadrat:
Screenshot 2021-06-29 at 16 03 03

The one thing that would be nice to add would be to style the fonts in the drop down to match their families, but styling selects is very hard so I'm not sure its worth the effort...

This also includes some refactoring to color customizations to keep the approaches similar and cleaner. There's probably more optimizations that can be done.

#### Related issue(s):
#4051


pbking:
I get the following error on the Customizer view when I make a change:
```
Notice: WP_Scripts::localize was called incorrectly. The $l10n parameter must be an array. To pass arbitrary data to scripts, use the wp_add_inline_script() function instead. Please see Debugging in WordPress for more information. (This message was added in version 5.7.0.) in /var/www/html/wp-includes/functions.php on line 5313
```

It works as expected previewing otherwise, just has the error.

pbking:
Gave it a solid shake and from a user perspective, using the Customizer only everything worked as expected (except for the error noted).

This technique excludes the ability to make the same modifications in the Global Styles panel which makes it a different beast than Color Customizations.  It may even make the situation in the Global Styles panel more complicated in that changing the 'base' font family in the Customizer changes a wider range of font-family instances than the existing "global" (single) font-family control available in the Global Panel.  (And changes them in different ways.)

I would classify this as "not universal" (for now).

This technique heralds back to something mentioned before; the ability for a theme to offer controls for values in the 'custom' block to users in the Global Styles panel.  If something like that should become reality and themes are able to customize values with the same granularity explored here in the Customizer then this is a method that could evolve in the future.  

Without that, however, the fonts users choose in the Customizer are locked in; the only way to change them is in the Customizer.

mikachan:
I've tested this using the Customizer and everything seemed to work as expected! Although I can confirm that I saw the same error as @pbking above in the logs when making changes.

PR #4117, opened 2021-06-29
Styled color to HTML block in editor (overriding Gutenberg's styles)

### LINKED ISSUE #4093: Quadrat: Styling of Custom HTML block has low contrast with background
## Steps to replicate

Insert a Custom HTML block and add content to it.

## Result

Text is Black on Dark Blue - very low contrast

## Expected

The text to be a light color to allow appropriate contrast

pbking:

This change uses the 'form color' for the HTML block when NOT selected (in the same way as it had already applied that color when UNSELECTED).

Before:
image
image

After:
image
image

Note that this change effects the editor only.  The view remains unchanged.

#### Related issue(s):

Fixes #4093

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -321,6 +321,11 @@ input[type=checkbox] + label {
 	font-size: var(--wp--custom--gallery--caption--font-size);
 }
 
+.block-library-html__edit .block-editor-plain-text {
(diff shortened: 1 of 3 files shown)


jffng (approved):
Tested in Quadrat and verified the issue is fixed, as well as BB to ensure no funny business. LGTM.

PR #4107, opened 2021-06-28
Quadrat: fix query block markup

jffng:
While looking into #4060, I discovered that the query block was only showing one post per page. 

This PR fixes that by adding the `<!-- wp:post-template -->` wrapper to the query block in the index.html template and setting a `perPage` attribute to the query block.

Maybe fixes #4060.

PR #4122, opened 2021-06-29
Blockbase: add additional vertical padding

### LINKED ISSUE #3964: Blockbase: Add additional vertical padding between header and post titles
I recommend `40px` of additional padding. Please ensure that the post title position is identical when navigating between the topmost post on an archive page and the single post page. 

Before|After
---|---
Screen Shot 2021-05-27 at 11 46 49 AMScreen Shot 2021-05-27 at 11 46 52 AM|Screen Shot 2021-05-27 at 11 47 05 AMScreen Shot 2021-05-27 at 11 48 38 AM

mikachan:
This change adds an additional `40px` of padding in between the header and main content on Blockbase. It also adds the same additional amount of padding in between each post in a query block.

I've made sure that the post title position is identical when navigating between the topmost post on an archive page and the single post page.

#### Related issue(s):
Closes #3964


blockbase/sass/blocks/_post-template.scss
```diff
@@ -0,0 +1,9 @@
+.wp-block-post-template {
+    li {
+        padding-bottom: var(--wp--custom--padding--vertical);
```
scruffian: Will this also add padding bottom to list items inside posts that are loaded in the post template block?

scruffian:
Do you think adding a separator block to the template would work? The advantage of that approach is that users can customize it...

blockbase/theme.json
```diff
@@ -203,6 +203,9 @@
 				"horizontal": "30px",
 				"vertical": "30px"
 			},
+			"padding": {
```
pbking: I hesitate to add a custom value for this to blockbase for this. Would it be possible to do this just with block styles rather than a custom variable + CSS?  Perhaps just adding a top padding of `40px` (or whatever is appropriate) to the ["core/post-title" block](https://github.com/Automattic/themes/blob/trunk/blockbase/theme.json#L386) would have the intended effect?

blockbase/sass/blocks/_post-template.scss
mikachan: Yes it probably would, I'll make it more specific!

blockbase/theme.json
mikachan: Yeah, applying a top padding to the 'core/post-title' block should have the same effect. I nearly went a similar route (although still with CSS..), but I added a reuseable variable as the same additional padding needed to be added to the posts in a query block. We could just do the same for these as well though - add the padding to the block in the `theme.json` (I'm guessing so anyway, without looking into it yet!)

mikachan:
> Do you think adding a separator block to the template would work? The advantage of that approach is that users can customize it...

Yeah, that definitely sounds like a good option. That would make it easier to apply the same spacing in other places, without needing to change CSS or the `theme.json` each time (I think?) I'll look at this today.

mikachan:
I've changed how this works based on your feedback and some help from @MaggieCabrera 🙌

Instead of using custom variables and additional CSS, I've tried to use the block editor as much as possible. This means fewer changes to the code and users can now customise both of these padding changes.

For the padding between the header and the titles, I've added `40px` of padding to the bottom of the header `wp:group` block.

For the padding underneath each post in a query list, I've added a `wp:spacer` block with a height of `40px`.

blockbase/block-templates/index.html
```diff
@@ -6,6 +6,7 @@
 	<!-- wp:post-title {"isLink":true} /-->
 	<!-- wp:post-featured-image {"isLink":true} /-->
 	<!-- wp:post-excerpt /-->
+	<!-- wp:spacer {"height":40} -->
```
pbking, line 9: The spacer block needs to have the HTML stuff inside of it and get closed off to work properly.  I usually just instantiate the block in the block editor and copy/paste 'cause I usually forget how it goes. :P

```
	<!-- wp:spacer {"height":40} -->
	<!-- /wp:spacer -->

File: blockbase/block-templates/index.html [hunk 4]
```diff
@@ -6,6 +6,7 @@
 	<!-- wp:post-title {"isLink":true} /-->
```
pbking, line 6: KIND of unrelated to the change, but in order to get multiple items to show up to evaluate the spacer below I had to change this `<!-- wp:query-loop -->` to `<!-- wp:post-template -->` to get multiple items per-page.  This was a recent change to Gutenberg and would fit nicely into this change I think.

See [this change](https://github.com/Automattic/themes/pull/4107) for the same change in quadrat's templates.

pbking:
Dig the padding and spacer changes better for sure.  A couple of things to fix up before we bring it in.

blockbase/block-templates/index.html
mikachan: Ahh I did this see when I was trying to figure out the query loop changes. I wasn't sure which one was correct so I went with the current one. This makes sense, thanks for explaining!

mikachan:
This is now ready for another review. Thanks!

scruffian (approved):
This looks great, thanks.

A couple of things I also considered...

Should we add the padding to the template part, rather than the block. This seems easier for a user to edit as selecting the group block inside the template part is tricky. However we want this padding on every page, so I believe the group block is the right place for it, as you have done.

I also considered putting the spacer above the post, then we wouldn't need to add it to the header at all, but this would then also mean adding the same spacer on the singular template,  and if a user want to change it they'd need to update both places, so again, this solution is better!

PR #4123, opened 2021-06-29
Controlled alignfull padding differently to correct paragraph padding in certain situation.

### LINKED ISSUE #4099: Blockbase: p tags inside of a full-width group block don't always need extra padding.
The extra margins for `[data-align="full"] p` that were added in #3978 do not need to apply when a custom layout is active for the group block. The rule will cause paragraphs to be misaligned with other elements on the page. This happens on the frontend too because of the `.alignfull p` rule:

https://github.com/Automattic/themes/blob/c0e9a574ec9cdfcd5733beaae3148567a339f6fb/blockbase/sass/base/_alignment.scss#L18-L19

Since the post excerpt block uses `p` tags, the issue is visible for that block as well. 

## Screenshots:

Example using "Inherit default layout" for a Group Block: 

Screen Shot 2021-06-25 at 8 44 20 AM

Example involving a Query block and Post Excerpt:

Screen Shot 2021-06-25 at 8 56 45 AM

## Example markup: 

```
<!-- wp:paragraph -->
<p>This is a paragraph outside of a Group block. It should align on the left and right edges with the paragraph below.</p>
<!-- /wp:paragraph -->

<!-- wp:group {"align":"full","layout":{"inherit":true}} -->
<div class="wp-block-group alignfull"><!-- wp:paragraph -->
<p>This is a paragraph inside of a Group block. It should align on the left and right edges with the paragraph above.</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group -->

<!-- wp:spacer -->
<div style="height:100px" aria-hidden="true" class="wp-block-spacer"></div>
<!-- /wp:spacer -->

<!-- wp:query {"queryId":15,"query":{"perPage":1,"pages":0,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":false},"align":"full"} -->
<div class="wp-block-query alignfull"><!-- wp:post-template -->
<div class="wp-block-group"><!-- wp:post-title {"isLink":true} /-->

<!-- wp:post-featured-image {"isLink":true} /-->

<!-- wp:post-excerpt /-->

<!-- wp:post-date /--></div>
<!-- /wp:group -->
<!-- /wp:post-template --></div>
<!-- /wp:query -->
```

pbking:
This changes the padding rules around groups (particularly full width groups) so that paragraph padding rules behave as expected.

The result is that ALL group blocks that are configured to `alignfull` introduce a standard-width padding.  This changes the behavior of some blocks when inside of a group block (such as an image) which used to expand fully (only if the original image was wide enough).

Before:

After:

#### Related issue(s):
Fixes #4099

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -60,20 +60,19 @@ img {
 	padding-right: var(--wp--custom--post-content--padding--right);
 }
 
-.block-editor-block-list__layout.is-root-container > .wp-block[data-align=full],
-.wp-block-post-content > .alignfull {
+.wp-block-group.alignfull,
+*[class*="wp-container-"] {
+	padding-left: var(--wp--custom--post-content--padding--left);
+	padding-right: var(--wp--custom--post-content--padding--right);
+}
+
+.wp-block-group.alignfull *[class*="wp-container-"],
+.wp-block-group.alignfull > .alignfull,
+*[class*="wp-container-"] *[class*="wp-container-"],
+*[class*="wp-container-"] > .alignfull {
 	margin-left: calc(-1 * var(--wp--custom--post-content--padding--left)) !important;
 	margin-right: calc(-1 * var(--wp--custom--post-content--padding--right)) !important;
 	width: calc( 100% + var(--wp--custom--post-content--padding--left) + var(--wp--custom--post-content--padding--right)) !important;
-}
-
-.site-header,
-.post-header,
-.page-content,
-[data-align="full"] .wp-block-group,
-.wp-block-group.alignfull {
-	padding-left: var(--wp--custom--post-content--padding--left);
-	padding-right: var(--wp--custom--post-content--padding--right);
 }
 
 @media (min-width: 480px) {
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -70,8 +70,8 @@ img {
 .site-header,
 .post-header,
 .page-content,
-[data-align="full"] p,
-.alignfull p {
+[data-align="full"] .wp-block-group,
+.wp-block-group.alignfull {
 	padding-left: var(--wp--custom--post-content--padding--left);
 	padding-right: var(--wp--custom--post-content--padding--right);
 }
--- a/blockbase/block-templates/404.html
+++ b/blockbase/block-templates/404.html
@@ -1,7 +1,7 @@
-<!-- wp:template-part {"slug":"header","tagName":"header","layout":{"inherit":true}} /-->
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
-<!-- wp:group {"layout":{"inherit":true},"className":"page-content","tagName":"main"} -->
-<main class="wp-block-group page-content">
+<!-- wp:group {"layout":{"inherit":true},"tagName":"main"} -->
+<main class="wp-block-group">
 
 <!-- wp:heading {"level":1,"fontSize":"large"} -->
 <h1 class="has-large-font-size">Oops! That page can’t be found.</h1>
--- a/blockbase/block-templates/header-footer-only.html
+++ b/blockbase/block-templates/header-footer-only.html
@@ -1,7 +1,7 @@
-<!-- wp:template-part {"slug":"header","tagName":"header","layout":{"inherit":true}} /-->
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
-<!-- wp:group {"tagName":"main"} -->
-<main class="wp-block-group post-content">
+<!-- wp:group {"tagName":"main","layout":{"inherit":true}} -->
+<main class="wp-block-group">
 
 <!-- wp:post-content {"layout":{"inherit":true}} /-->
 </main>
--- a/blockbase/block-templates/index.html
+++ b/blockbase/block-templates/index.html
@@ -6,7 +6,7 @@
 <!-- wp:group {"layout":{"inherit":true}} -->
 <div class="wp-block-group">
 	<!-- wp:post-title {"isLink":true} /-->
-	<!-- wp:post-featured-image {"isLink":true, "align":"full"} /-->
+	<!-- wp:post-featured-image {"isLink":true} /-->
 	<!-- wp:post-excerpt /-->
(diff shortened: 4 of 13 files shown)


jffng (commented):
I tried out the markup @kjellr supplied and ran into a case that this change doesn't quite cover:

```
<!-- wp:query {"queryId":15,"query":{"perPage":1,"pages":0,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":false},"align":"full"} -->
<div class="wp-block-query alignfull"><!-- wp:post-template -->
<div class="wp-block-group"><!-- wp:post-title {"isLink":true} /-->

<!-- wp:post-featured-image {"isLink":true} /-->

<!-- wp:post-excerpt /-->

<!-- wp:post-date /--></div>
<!-- /wp:group -->
<!-- /wp:post-template --></div>
<!-- /wp:query -->
```

In the above example, you would need to add `alignfull` to the group to get the correct padding on mobile. This feels a bit counterintuitive, but I'm not sure of a better approach because we don't want to add the padding to _all_ group blocks — this would interfere with the layout control. 

Considering the plethora of ideas we tried in our 🍐  session, I think this is the simplest approach that integrates well with the existing layout controls. 

Curious if others have ideas though; it's a tricky problem.

pbking:
I think that in the instance you raised that the `<!-- wp:group -->` block isn't actually needed at all to achieve the design target, and eliminating it makes the situation simpler. Remove `alignFull` from the `wp:query` block and the content get's the regular margins...

Maybe this borks up something else though like featured images that I haven't considered in my limited testing...

```
<!-- wp:query {"queryId":15,"query":{"perPage":5,"pages":0,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":false}} -->
<div class="wp-block-query"><!-- wp:post-template -->
<!-- wp:post-title {"isLink":true} /-->

<!-- wp:post-featured-image {"isLink":true} /-->

<!-- wp:post-excerpt /-->

<!-- wp:post-date /-->
<!-- /wp:post-template --></div>
<!-- /wp:query -->
```

kjellr:
@pbking I think this PR is missing one additional rule — when layout _is_ active, `alignfull` children should extend all the way to the edges. This is a quick fix, you can just modify [the rules above your edits](https://github.com/Automattic/themes/blob/f0422058c29978ab507a51adbc060bd3e5f0ab30/blockbase/sass/base/_alignment.scss#L8-L13) to add a couple new selectors: 

```
.block-editor-block-list__layout.is-root-container>.wp-block[data-align=full],
.wp-block-post-content > .alignfull,
[data-align="full"] .wp-block-group > [data-align="full"],
.wp-block-group.alignfull > .alignfull {
	margin-left: calc(-1 * var(--wp--custom--post-content--padding--left)) !important;
	margin-right: calc(-1 * var(--wp--custom--post-content--padding--right)) !important;
	width: calc( 100% + var(--wp--custom--post-content--padding--left) + var(--wp--custom--post-content--padding--right) ) !important;
}
```

Before|After
---|---
Screen Shot 2021-07-01 at 10 31 19 AM|Screen Shot 2021-07-01 at 10 30 12 AM

kjellr:
> Maybe this borks up something else though like featured images that I haven't considered in my limited testing...

Yeah, wide and full alignments within the block won't work with this setup. This featured image is supposed to be wide-width for example: 

Screen Shot 2021-07-01 at 10 32 59 AM

jffng:
I gave this another spin and verified a couple things:

- @kjellr's suggestion plays nicely with the changes already in this PR. 
- It still requires that the group block inside the query has "alignfull" for the padding to be applied correctly on mobile. 

Something else worth mentioning — we can't solve this problem with a media query because we don't know the viewport width to add the rule, given how the layout system works.

pbking:
Shooooooo!

@jffng and I gave this ANOTHER another spin around the block and then on the freeway and then around the block again and... did some more refactoring.

The result was the following:

Eliminated ['page-content'](https://github.com/Automattic/themes/pull/4123/commits/3fc199666a81d67f1d818e7fd96517239fe75478#diff-48d2e95a15120f192001e9c23b01a6e0d8609a31615b24ff2f30e2723a8d2389L17) [class](https://github.com/Automattic/themes/pull/4123/commits/3fc199666a81d67f1d818e7fd96517239fe75478#diff-48d2e95a15120f192001e9c23b01a6e0d8609a31615b24ff2f30e2723a8d2389L17) as a means to apply default 'edge'
padding to content.

Introduced ['container-' selection](https://github.com/Automattic/themes/pull/4123/commits/3fc199666a81d67f1d818e7fd96517239fe75478#diff-48d2e95a15120f192001e9c23b01a6e0d8609a31615b24ff2f30e2723a8d2389R9) as the means to apply default 'edge'
padding to content.

Removed 'edge' padding from[ post-content block ](https://github.com/Automattic/themes/pull/4123/commits/3fc199666a81d67f1d818e7fd96517239fe75478#diff-8d19c6d8b05bdcca09ecf489813da09999af47dd7cfafc71b7e263f900b97410L378)(it is all handled by
'container-' elements now).

Refactored the ["bust out of edge padding"](https://github.com/Automattic/themes/pull/4123/commits/3fc199666a81d67f1d818e7fd96517239fe75478#diff-48d2e95a15120f192001e9c23b01a6e0d8609a31615b24ff2f30e2723a8d2389R15-R22) for alignfull elements and
nested 'container-'s.

Lastly we [demonstrated](https://github.com/Automattic/themes/pull/4123/commits/25ca5699b5ee459e0ac11d782c09406ba194b5a6) one of the more complex scenarios this allow for (permitting an element to 'alignfull' inside of a query block loop).  This change is not aligned with the expected design, but the attempt and resolution of that behavior seemed important to demonstrate.  That should be reverted once evaluation is complete.

There's a TON of behavior that we did NOT test while working on this that should be evaluated for correctness.  There aren't many lines changed but all-in-all it's a pretty heft alignments change that should be given heavy scrutiny.

scruffian (approved):
This seems to work in my tests. I'd still love for us to use grid in core...

PR #4124, opened 2021-06-29
Quadrat: Make Site Title uppercase by default

### LINKED ISSUE #4064: Quadrat: Make Site Title all caps by default
Site title should be all caps by default.

**Currently**
Captura de ecrã 2021-06-15, às 16 26 08

**All Caps**
Captura de ecrã 2021-06-15, às 16 27 01

danieldudzic:
Before:
before-3

After:
themes

Closes: #4064

### DIFF
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -533,6 +533,7 @@ textarea:focus {
 
 .site-header .wp-block-site-title {
 	margin: 0;
+	text-transform: uppercase;
 }
 
 .site-header .wp-block-navigation {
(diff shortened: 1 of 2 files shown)


pbking (approved):
No so much "by default" as it is "always" but not sure that it can be a user-controlled thing.  Works as intended.
🚢

danieldudzic:
> No so much "by default" as it is "always" but not sure that it can be a user-controlled thing. Works as intended.
> 🚢

Right. Perhaps Site Title style customization is something that might get added in the future?

jffng:
`settings.typography.customTextTransformations` are a thing in theme.json: 

https://developer.wordpress.org/block-editor/how-to-guides/themes/theme-json/#settings

I wonder if it would make sense / work to set this to "true" for the site title block?

scruffian:
I think we should do it this way: https://github.com/Automattic/themes/pull/4132

pbking:
Closing in favor of #4132

PR #4125, opened 2021-06-29
Blockbase: Fix comment styles

danieldudzic:
Adds WP.com specific styles in Blockbase #4109


danieldudzic:
Yup this isn't required in BB

Issue #4111, opened 2021-06-29
Quadrat: Colors customization does not carried to the editor


retnonindya:
## Steps to replicate

Make sure Quadrat is active, then go to Customizer > Colors to change the color. I was able to customize the color as I prefer, so this is a 👍  -- Also, I able to confirm the color stays on the live site (the changes are saved properly on Customizer and reflected properly on the live site.)

![Screenshot 2021-06-29 at 4 35 33 PM](https://user-images.githubusercontent.com/44726232/123765199-1a51f580-d8f8-11eb-8562-a4de88140c8a.png)

Then, I clicked a page to edit it. I found the color customization is not reflected on the editor (the editor still using the default color palette) -- while I have made sure I set "Use theme styles" to ACTIVE on Preferences (on the editor.)

![Screenshot 2021-06-29 at 4 34 03 PM](https://user-images.githubusercontent.com/44726232/123765599-761c7e80-d8f8-11eb-88f4-a9821cbb98e0.png)

## Result

The editor is still using the default color palette 

## Expected

I expect the editor to use the customized color palette that has been set on Customizer.

* User report: Personal test/Call for Testing/I'm curious
* Details: Chrome Version 91.0.4472.114 (Official Build) (x86_64). macOS Big Sur ver. 11.4
* WordPress.com Simple Site

MaggieCabrera:
I could not reproduce this locally but it does happen on dotcom on a simple site

retnonindya:
Thanks for the check, @MaggieCabrera! I updated the report above with additional information that the website is a WordPress.com Simple site 🙇

scruffian:
I believe that now Quadrat has an index.html file this no longer happens.

PR #4136, opened 2021-06-30
Quadrat: Add an index template

scruffian:
Global Styles only load in the editor when a theme is a block theme. This adds an index.html file to Quadrat so that the colors and fonts load correctly:

Before:
Screenshot 2021-06-30 at 17 47 33

After:
Screenshot 2021-06-30 at 17 47 14

### DIFF
--- a/quadrat/block-templates/index.html
+++ b/quadrat/block-templates/index.html
@@ -1,0 +1,1 @@
+<!-- wp:template-part {"slug":"index"} /-->


jffng:
Does this fix #4111?

quadrat/block-templates/index.html
```diff
@@ -0,0 +1 @@
+<!-- wp:template-part {"slug":"index"} /-->
```
jffng, line 1: We have an `index.html` in the block-template-parts folder that the index.php loads — should we use that instead?

scruffian:
Closing in favour of https://github.com/Automattic/themes/pull/4141

PR #4141, opened 2021-06-30
Blockbase: Add theme support for block templates

### LINKED ISSUE #4111: Quadrat: Colors customization does not carried to the editor

scruffian:
Classic themes need to declare support for block templates to ensure that theme.json loads in the block editor (see https://github.com/WordPress/gutenberg/blob/f1aacbbe451d39a4aee40bff329100b77d2fe2fb/lib/global-styles.php#L107)

This adds a theme_support for block templates to Blockbase so that it gets passed on to all of the child themes.

#### Related issue(s):
Fixes #4111

### DIFF
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -24,6 +24,9 @@ if ( ! function_exists( 'blockbase_support' ) ) :
 		// Experimental support for adding blocks inside nav menus
 		add_theme_support( 'block-nav-menus' );
 
+		// Add support for block templates, so that user Global Styles settings load in the editor.
+		add_theme_support( 'block-templates' );
+
 		// Enqueue editor styles.
 		add_editor_style(
 			array(


MaggieCabrera:
I'm getting weird things with this PR. It does fix the original issue but now the frontend doesn't get the correct colors (the customizer preview and the post editor are showing the new colors but not the frontend) and the layout looks broken:

Screenshot 2021-07-01 at 11 18 39

This is happening for me both on dotcom and on my local install.

scruffian:
This is no longer needed as Quadrat has an index.html

PR #4142, opened 2021-06-30
Blockbase + children: Add footer credits for wordpress.com 

### LINKED ISSUE #4047: Quadrat: Add support for Footer Credits
## Steps to replicate

Reference site for all the screenshot below: https://iantesting609008735.wordpress.com/ 

Footer credits not showing as expected on WordPress.com

## Result

footer-credits

## Expected

See any of the Varia themes for working version.

scruffian:
WordPress.com looks for footer credits that contain a link to wordpress.org (or .com). This adds consistent footer credits to the three child themes (this was done for Blockbase in https://github.com/Automattic/themes/pull/4076) so that they don't get extra credits when on WordPress.com

To test:
- Put this code on wpcom (npm run sandbox:push)
- Switch to a child of blockbase
- Check that the footer credit only appears once

#### Related issue(s):
Fixes #4047

### DIFF
--- a/mayland-blocks/block-template-parts/footer.html
+++ b/mayland-blocks/block-template-parts/footer.html
@@ -19,3 +19,7 @@
 <!-- /wp:navigation --></div>
 <!-- /wp:column --></div>
 <!-- /wp:columns -->
+
+<!-- wp:paragraph {"align":"center"} -->
+<p class="has-text-align-center">Proudly Powered by <a href="https://wordpress.org" rel="nofollow">WordPress</a></p>
+<!-- /wp:paragraph -->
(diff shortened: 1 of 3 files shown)


MaggieCabrera:
I'm not sure what I should be seeing but the original issue mentions that we should compare the credits with what varia children are showing. 

So on Quadrat with this PR I see (link goes to https://wordpress.org/): 

Screenshot 2021-07-01 at 11 11 19

and on Hever (and all Varia children) and the link goes to https://wordpress.com/?ref=footer_custom_powered: 

Screenshot 2021-07-01 at 11 11 29

scruffian:
If you look at https://github.com/Automattic/themes/pull/4076 you can see there is a patch for WPCOM. It seems like footer credits on block themes work differently to classic themes. The main issue we were seeing was that they were being doubled on block themes, although I couldn't replicate this. The Regex is looking for a link, so I thought the best fix for now would be to just match the other themes to Blockbase, since that's what was changed...

MaggieCabrera (approved):
I couldn't replicate the duplication but I did see a space below the footer that disappeared after applying the PR. Thanks for clarifying about the other patch. I'd say go ahead with this

PR #4152, opened 2021-07-02
Seedlet Blocks & Mayland Blocks: Add FSE tag for dotorg.

### LINKED ISSUE #4146: Seedlet + Mayland Blocks: Add FSE tag for dotorg
I was navigating the [theme directory for FSE themes](https://wordpress.org/themes/tags/full-site-editing/) and only saw Blockbase. Seedlet Blocks and Mayland Blocks should have the tag too.

danieldudzic:
Closes #4146

### DIFF
--- a/mayland-blocks/style.css
+++ b/mayland-blocks/style.css
@@ -12,7 +12,7 @@ License: GNU General Public License v2 or later
 License URI: https://raw.githubusercontent.com/Automattic/themes/trunk/LICENSE
 Template: blockbase
 Text Domain: mayland-blocks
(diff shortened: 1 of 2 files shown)


MaggieCabrera:
This is needed for these themes to show up as FSE ready on dotorg, but right now this same tag is being used by themes using the legacy FSE on dotcom (such as Maywood for example). Will this break things for any dotcom integrations? should we add this only for dotorg submissions? @mattwiebe

scruffian:
Should we also add it for Quadrat?

MaggieCabrera:
That's the question of the hour. If you can't use the site editor, then I'd say no.

MaggieCabrera (approved):
After talking to @Copons about this, we shouldn't have problems merging this.

PR #4153, opened 2021-07-02
Blockbase + children: increased minor version 

MaggieCabrera:
After adding the color customizations functionality we should increase the minor version of these themes. I updated the changelog accordingly.

### DIFF
--- a/blockbase/package.json
+++ b/blockbase/package.json
@@ -1,6 +1,6 @@
 {
   "name": "blockbase",
-  "version": "1.1.2",
+  "version": "1.1.0",
   "description": "Blank Parent Theme",
   "bugs": {
     "url": "https://github.com/Automattic/themes/issues"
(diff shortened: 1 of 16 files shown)


blockbase/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "blockbase",
-  "version": "1.0.1",
+  "version": "1.1.2",
```
scruffian: Why 1.1.2, not 1.1.0?

MaggieCabrera: because the package.json wasn't up to date, the stylesheet was 1.0.2

scruffian: right but i think we should go back to 0 in the third position when the second position increases...

MaggieCabrera: oh yeah, I don't know what I was thinking! Let me change it

scruffian (approved):
LGTM

PR #4165, opened 2021-07-05
Added footer triangle to FSE

### LINKED ISSUE #4155: Quadrat: Footer shape not showing in Site Editor
## Steps to replicate

The footer shape doesn't appear in the Site Editor.
Frontend:
Screenshot 2021-07-02 at 13 52 33

Site Editor:
Screenshot 2021-07-02 at 13 52 13

## Expected
I would expect it to look the same.

pbking:
This change adds the footer triangle to the FSE.

Unfortunately this adds this to EVERY page whereas the view only shows it for the HOME page.  I'm not sure how to differentiate the home from others in the FSE so I'm not sure that this is a good fix.  Perhaps it is better to show this on NONE of the pages than on SOME of the pages.

#### Related issue(s):

Fixes: #4155

### DIFF
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -512,11 +512,13 @@ textarea:focus {
 	outline: 1px dotted currentColor;
 }
 
+footer,
 .home .site-footer.wp-block-group {
 	position: relative;
 	overflow: visible;
 }
 
+footer:before,
 .home .site-footer.wp-block-group:before {
 	content: "";
 	background-color: var(--wp--custom--color--secondary);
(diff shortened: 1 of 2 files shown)


MaggieCabrera:
Yeah, that's why I didn't add it originally. I'm still not too happy that the user can't turn these on/off on the FSE at all. I got to think about this to see if there's a better way to handle this.

jffng:
> I'm not sure how to differentiate the home from others in the FSE

These issues / feature requests are related: https://github.com/WordPress/gutenberg/issues/26494 https://github.com/WordPress/gutenberg/issues/17854

I'm not sure if template-specific classes are a good solution, or if there's a different approach we can take.

scruffian:
It seems to me this is a Gutenberg issue. It would be good to try to fix it there....

scruffian:
I'm going to close this as I think it should be fixed in Gutenberg.

PR #4171, opened 2021-07-06
Quadrat: block styles for query block

### LINKED ISSUE #3644: Quadrat: Hover effect
The theme has a neat effect when hovering over post links. A slightly rotated block appears behind it:

Screen Shot 2021-04-12 at 12 43 50 PM

We should determine a good approach for when and how this effect is applied.

MaggieCabrera:

This PR includes a new block style for the Query block and the associated block pattern. The design has a "Load more posts" button but I've decided to maintain the current query navigation design since we don't have the option for a one button solution within the Query block right now.

There are a few known issues that need to be solved in Gutenberg but I don't think they block this PR from merging in its current state:

- The excerpt block is showing content in the editor when the post contains no paragraph blocks but it does contain some other text blocks while in the frontend its showing no content. (https://github.com/WordPress/gutenberg/issues/33245)
- The Featured Image block doesn't have any controls for the image size so those have been provided by this PR. Ideally when the Featured Image block gains such controls we will remove the corresponding CSS. (https://github.com/WordPress/gutenberg/issues/27620)
- Wide align inside the editor for the Query block only affects the selected post content element. This only happens in the editor and it makes the design look quite bad in the editor but the frontend is looking fine. (https://github.com/WordPress/gutenberg/issues/33248)

ezgif-7-1001994f4c3b

#### Related issue(s):

Fixes https://github.com/Automattic/themes/issues/3644


MaggieCabrera:
@kjellr and @beafialho this is ready for your review as well

quadrat/inc/block-patterns.php
```diff
@@ -18,13 +18,14 @@ function quadrat_register_block_patterns() {
...
-				'latest-episodes',
-				'episode',
 				'cover-with-heading',
+				'episode',
+				'headline-button',
 				'headlines-and-buttons',
+				'query-diamond',
+				'latest-episodes',
 				'listen-to-the-podcast',
+				'media-text-button',
```
MaggieCabrera, line 28: @scruffian I hope you are proud of me

quadrat/sass/block-styles/_query.scss
```diff
@@ -0,0 +1,70 @@
...
+		&:last-child {
+			max-width: 400px;
+			flex-shrink: 0;
+		}
+	}
+	.wp-block-post-featured-image {
+		width: 400px;
+		@media (max-width: #{ ($break-small - 1) }) {
+			width: 100%;
+		}
```
MaggieCabrera, line 30: this aligns it to the Columns block media query that collapses the two columns into one. I tried to find a solution without a media query but I couldn't :(

quadrat/sass/block-styles/_query.scss
```diff
@@ -0,0 +1,70 @@
...
+			}
+			&:nth-child(2n+1):hover:before {
+				transform: rotate(-8deg) translate(-110px, 0);
+			}
+			&:nth-child(2n+2):before {
+				transform: rotate(24deg) translate(-130px, 20px);
+			}
+			&:nth-child(2n+2):hover:before {
+				transform: rotate(16deg) translate(-110px, 40px);
+			}
```
MaggieCabrera, line 72: @kjellr we went with a very simple animation that we think looks decent, but feel free to tweak the numbers in any way shape or form

kjellr:
This looks great, @MaggieCabrera! I pushed a couple commits with some minor adjustments: 

- I adjusted the responsive styles so that the featured image is 100% width on small screens. 
- I made the max size of the featured image slightly smaller, and adjusted the text width of the excerpt so it doesn't get too long. 
- I changed the font size of the "read more" text (there's currently no way to modify that independently of the rest of the body text, so this required an `!important` 😞)
- I renamed the pattern to avoid use of the word "hover". Folks on mobile won't generally know the hover exists anyway, and I think it's fine to just call this a list of posts.  
- I added a `blockTypes` prop so this pattern will show up in the Query Loop's pattern carousel. 

The only issue I'm seeing is [the issue you referenced above](https://github.com/WordPress/gutenberg/issues/33248), which unfortunately impacts the preview. But since that's an upstream bug, I think this is ready to go!

Desktop|Mobile
---|---
Screen Shot 2021-07-07 at 9 28 52 AM|Screen Shot 2021-07-07 at 9 29 34 AM

kjellr (approved):

MaggieCabrera:
Since this is an opt-in feature I'm bringing this in and we can iterate if we find it needs more work

PR #4174, opened 2021-07-06
Quadrat: Image alignments in Layout Grid block

### LINKED ISSUE #4158: Quadrat: Image alignments in Layout Grid block
## Steps to replicate

- Add a layout grid block
- Add an image to the block
- Align the image to the right

## Result

- The image appears on the left
Screenshot 2021-07-02 at 14 58 45

## Expected

- The image should appear on the right

Reported in pNEWy-dYW-p2#comment-53296

mikachan:
Currently, when adding an image in a Layout Grid Block and then setting the image alignment to the right, the change is not reflected in the editor. This PR fixes this by applying `float: right` to the image (similar to the styles for alignment on the front end.)

image

#### Testing steps:
- Add a layout grid block
- Add an image to the block
- Align the image to the right
- The image should align to the right of its column in the editor

#### Related issue(s):
Closes #4158

### DIFF
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -649,4 +649,8 @@ textarea:focus {
 	margin-top: calc( var(--wp--custom--margin--vertical) / 2);
 }
 
+.wp-block[data-align=right] > .wp-block-image {
+	float: right;
+}
+
 /*# sourceMappingURL=theme.css.map */
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -149,6 +149,10 @@
 	background: #fff;
 }
 
+.wp-block[data-align=right] > .wp-block-image {
+	float: right;
+}
+
 ul ul {
 	list-style-type: disc;
 }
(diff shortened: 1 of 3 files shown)


mikachan:
I've used CSS where I could have possibly added something to the `child-theme.json`, but I couldn't figure out how to target the same selector (`.wp-block[data-align=right] > .wp-block-image`).

While doing this I've also noticed that if you resize an image inside a layout grid in the editor (by dragging the edges of the image), this is only applied within the editor and not on the front end. I've got a fix for this in `child-theme.json` but as it's a separate problem I didn't know whether to create a new issue for it.

jffng (commented):
👋 

Adding this CSS does fix the issue.

What do you think about these fixes going in the Layout grid block's CSS instead? 

Since the behavior does not match between front-end and editor out of the box, and we do not appear to be adding any conflicting CSS from the theme, it seems like a bug with the block, not the theme.

mikachan:
Thanks for taking a look @jffng! Adding the changes to the Layout Grid block's CSS sounds sensible, it definitely sounds more like a bug with the block than this theme. I'll take a look at the Layout Grid block CSS next and report back here.

scruffian:
I agree, this should go in the block...

mikachan:
I've opened a [PR in block-experiments](https://github.com/Automattic/block-experiments/pull/226).

mikachan:
Closing this as the [temporary fix](https://github.com/Automattic/block-experiments/pull/226) in block-experiments has been merged and I've opened an [issue in Gutenberg](https://github.com/WordPress/gutenberg/issues/33332).

PR #4176, opened 2021-07-06
Fix FSE padding for BB themes.

pbking:
With [this recent change](https://github.com/Automattic/themes/pull/4123) the way page padding works changed a bit.

That change, however, never took into account the FSE.  This change addresses that to ensure that what is reflected in the FSE editor matches the view.

Force the 'body' (which, in the editor is applied to a different
element) margin and padding to obey the values set by the theme.

Remove the padding applied to the "post content" as the padding is now
applied a different way.

Before:

After:

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -3,8 +3,8 @@
  * - Reset the browser
  */
 body {
-	margin: 0;
-	padding: 0;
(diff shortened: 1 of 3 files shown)


mikachan:
This looks good to me, I can see the padding changes correctly in all Blockbase themes 👍

PR #4180, opened 2021-07-07
Added configurable padding for paragraphs and headings when they have background colors.

### LINKED ISSUE #4157: Quadrat: Header block padding
## Steps to replicate

Add background color to a header block:
Screenshot 2021-07-02 at 14 12 07

I think this is probably too much padding. The padding comes from Gutenberg so I expect the fix should be there too.

Originally reported here: pNEWy-dYW-p2#comment-53296

pbking:

Adds configurable padding values for HEADING and PARAGRAPH blocks that are applied with either have .has-background-color.

Note that the original issues suggested changing the default values in Gutenberg.  While I agree that the default values are too large, it's opinionated and I couldn't argue what the 'best/correct' value should be.

I experimented with moving it from style.scss to theme.scss and while that worked it seemed unnecessary as the styles to apply theme-specific padding were simple and weren't clobbered by the block defaults.  I suggest that we leave the block as-is.

Also the original issue didn't note the same padding needs for paragraph but this change makes both values configurable and sets smaller default values.

Design approval for the configured values (all themes are currently the same) would be appreciated.

#### Related issue(s):

Fixes #4157 

Before:

After:

### DIFF
--- a/blockbase/sass/base/_header.scss
+++ b/blockbase/sass/base/_header.scss
@@ -13,9 +13,9 @@
 
 h1, h2, h3, h4, h5, h6 {
 	&.has-background {
-		padding-top: var(--wp--custom--heading--has-background--spacing--padding--top);
-		padding-bottom: var(--wp--custom--heading--has-background--spacing--padding--bottom);
-		padding-left: var(--wp--custom--heading--has-background--spacing--padding--left);
-		padding-right: var(--wp--custom--heading--has-background--spacing--padding--right);
+		padding-top: var(--wp--custom--margin--vertical);
+		padding-bottom: var(--wp--custom--margin--vertical);
+		padding-left: var(--wp--custom--margin--horizontal);
+		padding-right: var(--wp--custom--margin--horizontal);
 	}
 }
--- a/blockbase/sass/base/_header.scss
+++ b/blockbase/sass/base/_header.scss
@@ -13,6 +13,6 @@
 
 h1, h2, h3, h4, h5, h6 {
 	&.has-background {
-		padding: var(--wp--custom--padding--vertical) var(--wp--custom--padding--vertical);
+		padding: var(--wp--custom--padding--vertical) var(--wp--custom--padding--horizontal);
 	}
 }
--- a/blockbase/sass/base/_header.scss
+++ b/blockbase/sass/base/_header.scss
@@ -10,3 +10,12 @@
 		text-decoration: none;
 	}
 }
+
+h1, h2, h3, h4, h5, h6 {
+	&.has-background {
+		padding-top: var(--wp--custom--heading--has-background--spacing--padding--top);
+		padding-bottom: var(--wp--custom--heading--has-background--spacing--padding--bottom);
+		padding-left: var(--wp--custom--heading--has-background--spacing--padding--left);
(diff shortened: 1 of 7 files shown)


mikachan (commented):
I really like this approach to fixing this issue - I'm still getting used to the balance between the styles Gutenberg provides and the styles from the theme, but I agree that it's difficult to argue what the 'correct' values should be here. I think it's great to include the paragraph blocks too.

I've tested this in Blockbase and all child themes, all looking good.

And finally, in my non-designer opinion, these default padding values look decent!

scruffian:
I wonder if we should use the custom spacing variable so it matches our other margins..

kjellr:
> I wonder if we should use the custom spacing variable so it matches our other margins..

Yeah, I was thinking the same thing — we should keep this consistent with our other spacing values.

pbking:
I thought I would really prefer the relative value ems provided and was going to comment as such, with a picture as to why ems were better, etc etc.



But now that I implemented it I don't agree with myself from 5 minutes ago.  Pushing the change. Good call y'all.

blockbase/sass/base/_header.scss
```diff
@@ -10,3 +10,12 @@
 		text-decoration: none;
 	}
 }
+
+h1, h2, h3, h4, h5, h6 {
+	&.has-background {
+		padding-top: var(--wp--custom--margin--vertical);
```
scruffian: Is there a reason you didn't use the shorthand?
`padding: var(--wp--custom--margin--vertical) var(--wp--custom--margin--horizontal)`

pbking: Only because I refactored from using four distinct variables and the thought just didn't cross my mind.

blockbase/sass/base/_header.scss
```diff
@@ -10,3 +10,9 @@
 		text-decoration: none;
 	}
 }
+
+h1, h2, h3, h4, h5, h6 {
+	&.has-background {
+		padding: var(--wp--custom--padding--vertical) var(--wp--custom--padding--vertical);
```
scruffian:
```suggestion
		padding: var(--wp--custom--padding--vertical) var(--wp--custom--padding--horizontal);
```

blockbase/sass/blocks/_paragraph.scss
```diff
@@ -12,4 +12,8 @@ p {
 		font-weight: var(--wp--custom--paragraph--dropcap--typography--font-weight);
 		margin: var(--wp--custom--paragraph--dropcap--margin);
 	}
+
+	&.has-background {
+		padding: var(--wp--custom--padding--vertical) var(--wp--custom--padding--vertical);
```
scruffian:
```suggestion
		padding: var(--wp--custom--padding--vertical) var(--wp--custom--padding--horizontal);
```

blockbase/sass/blocks/_paragraph.scss
```diff
@@ -14,6 +14,6 @@ p {
 	}
 
 	&.has-background {
-		padding: var(--wp--custom--padding--vertical) var(--wp--custom--padding--vertical);
+		padding: var(--wp--custom--padding--vertical) var(--wp--custom--padding--horizontal);
```
pbking, line 17: 🤦

Issue #4173, opened 2021-07-06
Blockbase: Color customizations caching


scruffian:
## Steps to replicate
In Blockbase, change colors in the customizer, and reload the front end. Notice that it takes some time for the new colors to show. I think this is a caching issue.

pbking:
The same is true for Global Styles panel in the FSE.  It starts with the default colors for a few seconds before switching to the customized colors. This was before I saw the content render. This was especially noticeable on my sandbox but also noticed locally.

scruffian:
i can no longer replicate this either. Closing for now.

PR #4181, opened 2021-07-07
Blockbase: Clear cache when changing global styles from the Customizer

scruffian:
Global Styles are cached in a transient, here:
https://github.com/WordPress/gutenberg/blob/f9eb4065cbbebef444ce5f53e7ad708146a2501e/lib/global-styles.php#L40

This deletes the transient when changes are made to Global Styles so that the newest version is loaded when you view the frontend.

To test:
- Go to the customizer and change fonts/colors
- Load the frontend
- Confirm you see your changes immediately

#### Related issue(s):
#4173


pbking:
Frustratingly, when I attempted to test this I was unable to reproduce.  I can confirm that nothing BREAKS but as I have been unable to reproduce the issue this afternoon I can't confirm it fixes.

jffng:
When I change the body font and heading to different fonts, it only saves the last selection, and I get this notice:

```
Notice: Undefined index: in /var/www/html/wp-content/themes/themes/blockbase/inc/wp-customize-fonts.php on line 257 Notice: Trying to access array offset on value of type null in /var/www/html/wp-content/themes/themes/blockbase/inc/wp-customize-fonts.php on line 260 Notice: Trying to access array offset on value of type null in /var/www/html/wp-content/plugins/gutenberg/lib/class-wp-theme-json-gutenberg.php on line 602
```

Here is a video of the bug: https://cloudup.com/cEeA5VwIc8s

I am not sure if it's related to the cache or the fonts customization code. It occurs on both trunk and this branch. I reverted https://github.com/Automattic/themes/pull/4183 locally to see if it was related, and was still getting the bug.

scruffian:
This notice was fixed in https://github.com/Automattic/themes/pull/4194 so maybe you can test it now :)

jffng:
Like @pbking I can't recreate the issue on trunk. Checking out this PR doesn't break anything.

scruffian:
I think some changes in Gutenberg make this unnecessary.

scruffian:
Merging this as a fix for https://github.com/Automattic/themes/issues/4197

PR #4183, opened 2021-07-08
Blockbase: Update font name

scruffian:
To match Quadat and the other themes, we should use the same name for the System font

To test:
- open the customizer fonts panel and confirm that System Font is selected


jffng (approved):
Screen Shot 2021-07-08 at 10 33 19 AM

PR #4185, opened 2021-07-08
Quadrat: Add a new podcast template

MaggieCabrera:

This PR includes a custom template called "Podcast" that will use the new Query loop block style.

Screenshot 2021-07-08 at 12 53 09

While on "classic" mode the user can edit this template using the new Custom template editor, there they can change the Query block to only show a specific category of posts (such as Podcast).

While doing this PR I found out two issues:

- Quadrat inherits Blockbase custom templates even if we are overriding the `customTemplates` array on theme.json with this new change.
- Our php templates are showing up on the Custom template editor dropdown, showing as duplicates, that's confusing but I'm not sure how we can avoid this.

Screenshot 2021-07-08 at 12 22 48

#### Testing instructions:

- Create a new page
- Select the podcast template (the lower case one is the HTML one, the other one is the php one but they are effectively the same thing).
Screenshot 2021-07-08 at 12 43 18
- Test that you can edit the template within the Custom template editor and change the Query block to show a specific category

### DIFF
--- a/quadrat/404.php
+++ b/quadrat/404.php
@@ -1,13 +1,8 @@
 <?php
 /**
- * 404 template file
+ * The template for displaying 404 pages (not found)
  *
- * This is the most generic template file in a WordPress theme
- * and one of the two required files for a theme (the other being style.css).
- * It is used to display a page when nothing more specific matches a query.
- * E.g., it puts together the home page when no home.php file exists.
- *
- * @link https://developer.wordpress.org/themes/basics/template-hierarchy/
+ * @link https://codex.wordpress.org/Creating_an_Error_404_Page
  *
  * @package Quadrat
  * @since 1.0.0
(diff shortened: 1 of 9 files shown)


kjellr:
What's the benefit of having a separate template here? 

I'm wondering if it would it make more sense to just include a page in the demo/headstart content that uses a normal page template, and has this pattern inserted into the entry content instead. That way, users don't have to know to click into the template editor to change this.

MaggieCabrera:
> What's the benefit of having a separate template here?
> 
> I'm wondering if it would it make more sense to just include a page in the demo/headstart content that uses a normal page template, and has this pattern inserted into the entry content instead. That way, users don't have to know to click into the template editor to change this.

Yeah, we could do that. I was also thinking about dotorg users, since this was a theme geared towards podcasting, even though you can easily paste the block pattern yourself. But I agree that I'm not super keen on duplicating the code of the pattern on multiple places

kjellr:
Yeah, Dotorg users could just add the pattern and they'd be all set. I think that's the simplest way of making this work.

MaggieCabrera:
I'm fine closing this then! I'll add the page to the demo so we can add it to the headstart

PR #4186, opened 2021-07-08
Blockbase: Remove alignment fix

scruffian:
This removes an alignment fix we added to Blockbase because it causes other issues:

Before:
Screenshot 2021-07-08 at 11 50 54

After:
Screenshot 2021-07-08 at 11 52 34

This will cause issues with left/right floated content inside a container, but since it causes other issues, I think we'd be better to remove it and let those issues be fixed upstream.

#### Related issue(s):
https://github.com/WordPress/gutenberg/issues/10299

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -79,10 +79,6 @@ img {
 	}
 }
 
-.wp-block-group:not(.site-header) {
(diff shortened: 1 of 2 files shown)


pbking (approved):
Works for me.
Hopefully upstream happens sooner than later.

PR #4187, opened 2021-07-08
Quadrat: Remove PHP templates

scruffian:
The PHP templates in Quadrat are no longer used, so we can remove them and greatly simplify our templates.

To test, check that Quadrat continues to load the same.


mikachan (commented):
Can confirm Quadrat still loads the same for me with this change!

pbking (approved):
LGTM!
I love this PR!

PR #4190, opened 2021-07-08
Quadrat: Remove bottom space

scruffian:
Remove the bottom spacing on the front page template.

When the screen is wider than it is tall, and there's not much content on the page we get a space appear at the bottom:
Screenshot 2021-07-08 at 13 28 23

This changes the way that the polygon is positioned, by stretching the div across the whole site:
Screenshot 2021-07-08 at 13 28 08

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -101,24 +101,25 @@ h1.has-background, h2.has-background, h3.has-background, h4.has-background, h5.h
 
 /**
  * This file pushes the footer to the bottom of the page.
- * The 32px is for the admin bar, if it exsits.
  */
-html {
-	height: calc( 100% - 32px);
-}
-
-body {
-	height: calc( 100% + 32px);
-}
-
-body.admin-bar {
-	height: calc( 100%);
+:root {
+	--wpadmin-bar--height: 46px;
 }
 
 .wp-site-blocks {
+	min-height: 100vh;
 	display: grid;
-	height: 100%;
 	grid-template-rows: auto 1fr;
+}
+
+@media (min-width: 600px) {
+	body.admin-bar {
+		--wpadmin-bar--height: 32px;
+	}
+}
+
+body.admin-bar .wp-site-blocks {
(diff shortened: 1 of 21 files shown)


pbking:
It didn't work out for me...



I also notice that there is a very large vertical scrollbar in this situation and have no idea why...



My index template to force little content:

```

<!-- wp:group {"layout":{"inherit":true}, "className":"page-content"} -->
<div class="wp-block-group page-content">
<p>Just some content.  But not much...</p>
</div>
<!-- /wp:group -->

<!-- wp:template-part {"slug":"footer"} /-->
```

scruffian:
@pbking are you sure you had the latest version of the CSS? It works for me.... :)

mikachan (commented):
This works for me in all browsers 👍

image

pbking:
Just checked again.  There maybe were changes I didn't have last time? 
Still didn't work though. :(

image

scruffian:
I added a commit to force the footer to the bottom in Blockbase and children.

MaggieCabrera:
This one is ready for another review. Be aware that this affects:
- Blockbase and all its children
- Should be checked logged in and out, both on mobile and desktop

mikachan:
The footer looks like it's in the right place for me but when I hardly have any content, the content appears quite far down the page. It's fine with more content.

image

I've cleared my cache etc. I think it's related to the `display: grid` but I haven't used grid enough to know what's up. I tried `align-items: start` but that didn't do the trick.

MaggieCabrera:
I changed this to use flex, it was the simpler way but I had to add the extra class to the footer container. I tested all 4 themes, but please go ahead and see if you can break this.

mikachan:
Awesome, this is now working on the same page I had trouble with earlier!

image

I tested in the other browsers and the other themes, and all looked good to me.

pbking (approved):
w00t!
What a doosie.
This is working great for me for all the themes.
I tested with little content and with a lot of content.
Looked great on all accounts.  I could not break.
Well done.

PR #4194, opened 2021-07-08
Fixed null ref in font customizations

### LINKED ISSUE #4193: Blockbase Font Customization Error
## Steps to replicate

* Activate Quadrat
* Make sure you're starting with an "un-customized" theme environment (Using Global Styles Panel in FSE 'Rest to Defaults')
* Open Customizer and Customize the Fonts
* Set 'Body Font' to 'Domine' and Publish
* Refresh the Customizer.
```
Notice: Undefined index: in /var/www/html/wp-content/themes/blockbase/inc/wp-customize-fonts.php on line 258 Notice: Trying to access array offset on value of type null in /var/www/html/wp-content/themes/blockbase/inc/wp-customize-fonts.php on line 261 Notice: Trying to access array offset on value of type null in /var/www/html/wp-content/plugins/gutenberg/lib/class-wp-theme-json-gutenberg.php on line 602
```
Note the brokenness.

If that doesn't break it then also:
* Activate Blockbase
* Activate Quadrat
* Load Customizer
* Note Brokenness

NOTE: Changing both fonts back to DM Sans and publishing makes the problem go away. 

ANOTHER NOTE: Setting BOTH fonts seems to make everything OK.  Setting ONLY ONE makes the one not set break.  That includes setting something else (like colors) as well.  If neither of the fonts change then both break.

## Result

Arrays aren't as they should be.

## Expected

Errors shouldn't be thrown.

pbking:
Fixes #4193 

When no fonts have been configured the utility was saving an improper
font value.  This change pulls the 'default' and stores that instead.

The PHP code formatter also got ahold of this file though and I wasn't able to play golf because of that.

[This](https://github.com/Automattic/themes/compare/fix/bb-font-customization-defaults?expand=1#diff-e60fbd817cb00ff02ed3cc541e2d1577147299a706f58b281a2721cd534c84f1R319-R326) is the only relevant change.

Through debugging this I have discovered that even if no changes are made to colors or fonts both are saved to the user JSON.  Even if the user returns to default values the default values are still stored in the user space (rather than being removed and allowing the theme value to shine through).  Perhaps we should consider fixing that in our customization scripts.

### DIFF
--- a/blockbase/inc/wp-customize-fonts.php
+++ b/blockbase/inc/wp-customize-fonts.php
@@ -4,188 +4,188 @@ class GlobalStylesFontsCustomizer {
 
 	private $section_key = 'customize-global-styles-fonts';
 
-	private $fontSettings;
+	private $font_settings;
 
 	private $fonts = array(
-		"system-font" => array(
-			"fontFamily" => "-apple-system,BlinkMacSystemFont,\"Segoe UI\",Roboto,Oxygen-Sans,Ubuntu,Cantarell,\"Helvetica Neue\",sans-serif",
-			"slug" => "system-font",
-			"name" => "System Font",
+		'system-font'       => array(
+			'fontFamily' => '-apple-system,BlinkMacSystemFont,"Segoe UI",Roboto,Oxygen-Sans,Ubuntu,Cantarell,"Helvetica Neue",sans-serif',
+			'slug'       => 'system-font',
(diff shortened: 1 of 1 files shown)


scruffian (approved):
LGTM. 

Did the PHP fixes get caught by PHPCS? I'm not sure why mine didn't pick it up...

PR #4195, opened 2021-07-09
Quadrat: Sync post title position on single and archive pages

### LINKED ISSUE #4178: Quadrat: Sync up first post title position between archive and single pages
On the default index template, the first post's meta should line up with where the meta would appear on a single post page. I recall this being the case at some point, but it's no longer in sync. 

Fixing this will require:

- **Index template:** Adding additional spacing before the first post. Leaving mobile as it is today. 
- **Single Template:** Adding additional spacing before the first post on mobile. Leaving desktop as it is today.

## **Desktop:**

Current Index|Current Single
---|---
Screen Shot 2021-07-07 at 10 59 15 AM|Screen Shot 2021-07-07 at 10 59 19 AM

Recommended Index|Current Single (No Change)
---|---
Screen Shot 2021-07-07 at 11 00 14 AM|Screen Shot 2021-07-07 at 10 59 19 AM

## **Mobile:**

Current Index|Current Single
---|---
Screen Shot 2021-07-07 at 11 00 54 AM|Screen Shot 2021-07-07 at 11 01 10 AM

Current Index (No Change)|Recommended Single
---|---
Screen Shot 2021-07-07 at 11 00 54 AM|Screen Shot 2021-07-07 at 11 01 50 AM

mikachan:
This change ensures the post meta is in the same position vertically on both the archive and single page templates.

#### Related issue(s):
Fixes #4178

### DIFF
--- a/quadrat/block-template-parts/header.html
+++ b/quadrat/block-template-parts/header.html
@@ -1,5 +1,5 @@
-<!-- wp:group {"tagName":"header","className":"site-header"} -->
-<header class="wp-block-group site-header">
+<!-- wp:group {"tagName":"header","style":{"spacing":{"padding":{"top":"0px","right":"0px","bottom":"170px","left":"0px"}}},"className":"site-header"} -->
+<header class="wp-block-group site-header" style="padding-top:0px;padding-right:0px;padding-bottom:170px;padding-left:0px">
 	<!-- wp:site-logo /-->
 	<!-- wp:site-title /-->
 	<!-- wp:site-tagline {"style":{"spacing":{"padding":{"top":"0px","right":"0px","bottom":"0px","left":"20px"},"margin":{"right":"auto"}}},"fontSize":"tiny"} /-->
--- a/quadrat/block-template-parts/header.html
+++ b/quadrat/block-template-parts/header.html
@@ -1,5 +1,5 @@
-<!-- wp:group {"tagName":"header","style":{"spacing":{"padding":{"top":"0px","right":"0px","bottom":"170px","left":"0px"}}},"className":"site-header"} -->
-<header class="wp-block-group site-header" style="padding-top:0px;padding-right:0px;padding-bottom:170px;padding-left:0px">
+<!-- wp:group {"tagName":"header","style":{"spacing":{"padding":{"bottom":"170px"}}},"className":"site-header"} -->
+<header class="wp-block-group site-header" style="padding-bottom:170px">
 	<!-- wp:site-logo /-->
 	<!-- wp:site-title /-->
 	<!-- wp:site-tagline {"style":{"spacing":{"padding":{"top":"0px","right":"0px","bottom":"0px","left":"20px"},"margin":{"right":"auto"}}},"fontSize":"tiny"} /-->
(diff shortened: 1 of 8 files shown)


quadrat/sass/blocks/_post-template.scss
```diff
@@ -0,0 +1,3 @@
+.wp-block-post-template {
+    margin: 18px 0;
```
mikachan: I've added vertical margin to this `ul` as before we were reliant on the browser defaults. This means the spacing should be consistent between browsers and matches the other spacing changes for keeping the post meta in the same position between templates.

quadrat/sass/templates/_meta.scss
```diff
@@ -2,7 +2,7 @@
 	align-items: center;
 	justify-content: center;
 	@include break-small-only() {
-		padding-top: 0 !important;
+		padding-top: 18px !important;
```
mikachan: I've changed this to `18px` to match the `margin` on the post template container.

quadrat/block-templates/index.html
```diff
@@ -1,5 +1,9 @@
 <!-- wp:template-part {"slug":"header"} /-->
 
+<!-- wp:spacer {"height":152,"className":"index-spacer"} -->
+<div style="height:152px" aria-hidden="true" class="wp-block-spacer index-spacer"></div>
+<!-- /wp:spacer -->
+
```
mikachan: I used a spacer on the index template because the spacing on the single template is added via the editor using padding on the group block. This way they can both be edited via the editor.

quadrat/block-template-parts/header.html
```diff
@@ -1,8 +1,8 @@
-<!-- wp:group {"tagName":"header","className":"site-header"} -->
-<header class="wp-block-group site-header">
+<!-- wp:group {"tagName":"header","style":{"spacing":{"padding":{"top":"0px","right":"0px","bottom":"170px","left":"0px"}}},"className":"site-header"} -->
```
jffng: The header in this PR appears for me without correct padding:

Screen Shot 2021-07-09 at 11 10 13 AM

I think the top, right, and left padding values should be removed, and it works as expected.

quadrat/sass/blocks/_post-template.scss
```diff
@@ -0,0 +1,4 @@
+.wp-block-post-template {
```
jffng, line 1: I think I understand why this is necessary, because Gutenberg outputs this block as a `ul` and the browser styles add a margin to these by default...

jffng (commented):
I left one main comment, other than that LGTM! Thanks @mikachan!

quadrat/sass/blocks/_post-template.scss
mikachan: Yeah that's right, I was just overriding the browser defaults and getting rid of the extra space it was causing.

quadrat/block-template-parts/header.html
mikachan: Ah I didn't spot this but I can see it too, will sort it!

mikachan: Fixed this now and pushed it, thanks @jffng!

jffng:
LGTM, let's bring it in.

PR #4206, opened 2021-07-12
Blockbase: Disable the Site Editor on WPCOM

scruffian:
This disables the Site Editor on WPCOM for Blockbase and children, unless the site is opted in to the Site Editor.

Requires D63947-code to test.

#### Testing instructions:
- Switch to Blockbase on a WordPress.com simple site
- Check that you don't have a link to the Site Editor on the wp-admin dashboard: https://[slug].wordpress.com/wp-admin/
- Check that you do have a link to the Customizer on the wp-admin dashboard: https://[slug].wordpress.com/wp-admin/
- Check that you don't have a link to the Site Editor on the wpcom dashboard: https://wordpress.com/home/[slug].wordpress.com
- Check that you do have a link to the Customizer on the wpcom dashboard: https://wordpress.com/home/[slug].wordpress.com

Unanswered questions:
- what happens on Atomic?
- what happens on Jetpack?

### DIFF
--- a/blockbase/inc/disable-site-editor.php
+++ b/blockbase/inc/disable-site-editor.php
@@ -3,11 +3,25 @@
 /**
  * Adds a setting to the Gutenberg experiments page to disable the Site Editor.
  */
-function blockbase_add_disable_site_editor_setting() {
+function blockbase_disable_site_editor() {
 	if ( ! is_readable( get_stylesheet_directory() . '/block-templates/index.html' ) ) {
 		return;
 	}
 
+	if ( is_admin() ) {
+		add_action( 'admin_init', 'blockbase_add_settings_field' );
+	}
+
+	if ( get_option( 'gutenberg-experiments' ) ) {
+		if ( array_key_exists( 'universal-theme-disable-site-editor', get_option( 'gutenberg-experiments' ) ) ) {
+			add_action( 'admin_init', 'blockbase_readd_legacy_admin_links' );
+			add_action( 'admin_init', 'blockbase_remove_site_editor_admin_link' );
+			add_action( 'admin_bar_menu', 'blockbase_remove_site_editor_link', 50 );
+		}
+	}
+}
+
+function blockbase_add_settings_field() {
 	add_settings_field(
 		'universal-theme-disable-site-editor',
 		__( 'Site Editor', 'gutenberg' ),
@@ -86,6 +100,10 @@ function is_site_editor_menu_item( $menu_item ) {
 	}
 }
 
+function site_editor_enabled() {
+	return get_option( 'gutenberg-experiments' ) && array_key_exists( 'universal-theme-disable-site-editor', get_option( 'gutenberg-experiments' ) );
+}
+
 /**
  * Removes the Site Editor link from the admin.
  */
@@ -102,4 +120,13 @@ function blockbase_remove_site_editor_admin_link() {
 	unset( $menu[ $site_editor_index ] );
 }
 
-add_action( 'admin_init', 'blockbase_add_disable_site_editor_setting' );
+/**
+ * Removes Site Editor adminbar item.
+ *
+ * @param WP_Admin_Bar $wp_admin_bar The admin-bar instance.
+ */
+function blockbase_remove_site_editor_link( $wp_admin_bar ) {
+	$wp_admin_bar->remove_node( 'site-editor' );
+}
+
+add_action( 'init', 'blockbase_disable_site_editor' );
--- a/blockbase/inc/disable-site-editor.php
+++ b/blockbase/inc/disable-site-editor.php
@@ -3,7 +3,7 @@
 /**
  * Adds a setting to the Gutenberg experiments page to disable the Site Editor.
  */
-function add_disable_site_editor_setting() {
+function blockbase_add_disable_site_editor_setting() {
 	if ( ! is_readable( get_stylesheet_directory() . '/block-templates/index.html' ) ) {
 		return;
 	}
@@ -20,10 +20,10 @@ function add_disable_site_editor_setting() {
 		)
 	);
 
-	readd_legacy_admin_links();
+	blockbase_readd_legacy_admin_links();
 
 	if ( ! site_editor_enabled() ) {
-		remove_site_editor_admin_link();
+		blockbase_remove_site_editor_admin_link();
 	}
 }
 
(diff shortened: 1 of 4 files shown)


blockbase/inc/disable-site-editor.php
```diff
@@ -76,16 +80,17 @@ function_exists( 'gutenberg_use_widgets_block_editor') &&
...
  * Removes the Site Editor link from the admin.
  */
-function remove_site_editor_admin_link() {
+function blockbase_remove_site_editor_admin_link() {
 	global $menu;
+
 	// Remove Site Editor.
 	foreach ( $menu as $index => $menu_item ) {
-		if ( ! empty( $menu_item[5] ) && false !== strpos( $menu_item[5], 'toplevel_page_gutenberg-edit-site' ) ) {
+		if ( ! empty( $menu_item[ 2 ] ) && in_array( $menu_item[ 2 ], array( 'gutenberg-edit-site', 'site-editor' ) ) ) {
```
scruffian: On dotorg it's called `gutenberg-edit-site` on dotcom it's called `site-editor`.

jffng: How do we know to check the index `2` for the presence of the Site Editor?

jffng (commented):
My testing results:

Trunk | This PR
------ | -------
Screen Shot 2021-07-12 at 11 45 46 AM | Screen Shot 2021-07-12 at 11 55 38 AM

It correctly adds the Customizer links to the Themes menu. However it does not remove the Site Editor link from wp-admin nor the admin bar in the front-end.

scruffian:
Index 2 is the where the slug for the menu item is stored

scruffian:
> However it does not remove the Site Editor link from wp-admin nor the admin bar in the front-end.

Sorry about that. I've updated this now so it should work.

You can also test what happens when you add the sticker using this with `wpsh`:

`add_blog_sticker( 'core-site-editor-enabled', null, null, 'SITE_ID' );`

jffng:
Nice, I can confirm that this PR hides the site editor unless the `core-site-editor-enabled` sticker is present. 

There is one case this misses for me though, in the admin header on the front-end:

Screen Shot 2021-07-13 at 11 09 31 AM

scruffian:
Good catch @jffng. I added a commit to address that.

blockbase/inc/disable-site-editor.php
```diff
@@ -3,11 +3,25 @@
...
 
+	if ( is_admin() ) {
+		add_action( 'admin_init', 'blockbase_add_settings_field' );
+	}
+
+	if ( get_option( 'gutenberg-experiments' ) ) {
+		if ( array_key_exists( 'universal-theme-disable-site-editor', get_option( 'gutenberg-experiments' ) ) ) {
+			add_action( 'admin_init', 'blockbase_readd_legacy_admin_links' );
+			add_action( 'admin_init', 'blockbase_remove_site_editor_admin_link' );
+			add_action( 'admin_bar_menu', 'blockbase_remove_site_editor_link', 50 );
```
jffng: This is the right hook, but it's not firing because the conditional is not being met. 

When I move this action outside the top `if ( get_option( 'gutenberg-experiments') )` conditional, only then the site editor link from the admin bar is correctly removed.

scruffian:
@jffng I couldn't reproduce the problem.

I have rebased this from trunk so now the default behaviour is to disable the site editor in all environments.

blockbase/inc/disable-site-editor.php
```diff
@@ -19,27 +29,21 @@ function add_disable_site_editor_setting() {
 			'id'    => 'universal-theme-disable-site-editor',
```
Copons: I know this is unrelated to the PR (it happened in #4219), but it's breaking my brain 😅 

```
'label' => __( 'Enable Site Editor', 'gutenberg' ),
'id' => 'universal-theme-disable-site-editor',
```

Could we update the option id to `universal-theme-enable-site-editor` instead of `disable`?

scruffian: yeah good catch that's me being sloppy....

scruffian: Updated this, and renamed the file for clarity too.

Copons: Thanks! ✨

Copons:
I haven't tested the PR, but I think this makes a lot of sense.

We'll need to enable the option on sites that opted-in to the FSE Beta, but it definitely should stay off by default for now.

Jetpack sites?
Let's not touch them at all. If a self-hosted user has the Gutenberg plugin and activates an experimental block-based theme, they are probably expert enough to expect to gain access to the Site Editor.

Atomic sites?
Carry the option over during the transfer (so FSE beta Simple sites will keep getting the Site Editor on Atomic).
What about also letting them toggle the option at will? This would either require displaying the Gutenberg experiments (which AFAICS is hidden on Atomic), or maybe add the same toggle somewhere else. 🤔

jffng:
> Carry the option over during the transfer (so FSE beta Simple sites will keep getting the Site Editor on Atomic).

On atomic, how do sites get marked as FSE beta?

Copons:
> > Carry the option over during the transfer (so FSE beta Simple sites will keep getting the Site Editor on Atomic).
> 
> On atomic, how do sites get marked as FSE beta?

We haven't decided yet (probably a sticker though).

Though, I wasn't clear. What I mean is that we should just transfer the option from Simple to Atomic.
So if a Simple site has the Site Editor option enabled, it should still have it enabled after becoming Atomic.

Speaking of the FSE Beta: our plan right now is to enable that option at site creation for sites opting-in to the beta.

jffng:
Currently on this PR I am getting the following results, verified by @pbking: 

With `core-site-editor-enabled` sticker, Appearance > Customize is still present, but the Site Editor link does not appear in wp-admin or the admin bar.

Without `core-site-editor-enabled` sticker, the behavior seems correct. Appearance > Customize is present, and no Site Editor links anywhere.

scruffian:
Sorry I think this must have got broken in the rebase. I have updated it again and changed the approach to keep all the logic in one place.

MaggieCabrera (approved):
This works for me as intended!

obenland:
@scruffian The action that was added in D63947-code shouldn't be necessary if menu items are added/removed in the `admin_menu` action rather than on `admin_init`. Did that not work in the context of this PR?

obenland:
@scruffian Oh it looks like this was reverted eventually. Does that mean D63947-code can be, too?

scruffian:
@obenland I think so. Here's a patch D70458-code

PR #4210, opened 2021-07-12
Blockbase + children: Updated tags

MaggieCabrera:
Some of the existing tags were not accurate for the themes (I think Quadrat's were the same as Seedlet's). I updated them to the best of my knowledge following what's on the [dotorg docs](https://make.wordpress.org/themes/handbook/review/required/theme-tags/).


scruffian (approved):
LGTM

PR #4212, opened 2021-07-12
Refactor/bb customize

### LINKED ISSUE #4199: Blockbase: Customizer refactor
The code in [this file](https://github.com/Automattic/themes/blob/486aa30fbea19f695faae18460747b6b5041ace9/blockbase/inc/wp-customize-fonts.php) has way too much repetition (check the multiple `property_exsists` calls) - we should write a function which allows you to create an object if it exists or modify it if it does not.

pbking:
Refactor customization logic to simplify (... and complicate?)

There shouldn't be any VISIBLE changes to the customizer (with the exception of the "reset to defaults" option in the Fonts panel).

This change creates a [utility](https://github.com/Automattic/themes/pull/4212/files#diff-ee629dcc4aa54eece89426a3a11fb0bcf5ce12d71e16ebbabb5f6f2a49555b11R3-R15) to handle setting values in the Global Styles object improving indentation-heavy code like [this](https://github.com/Automattic/themes/pull/4212/files#diff-1e8f5c8da22610cdd824ed551bb21abc65759f8961c8ea9d777a213a483c8284L383-L399).

Additionally the font control creation was refactored to supply the DEFAULT value to the control.  This is used when the JSON is saved to determine if the values should be persisted or not.  If it's found that the values are set to the default values then no font-specific values are persisted to the USER JSON (allowing the theme.json values to be used).

The color saving logic follows the same form, determining when the VALUE is set to the DEFAULT VALUE and not saving any palette information to the USER JSON when that is the case.

Lastly the FONT customization panel now has a "Rest to Default" option.


MaggieCabrera:
Does it make sense to group the fonts/colors customizations on their own folders? I worry that "inc" is turning a little messy

blockbase/inc/customizer/wp-customize-fonts.php
```diff
@@ -273,18 +287,40 @@ function initialize( $wp_customize ) {
 			)
 		);
 
-		$this->add_setting_and_control( $wp_customize, 'body', __( 'Body font', 'blockbase' ), $body_font_slug );
-		$this->add_setting_and_control( $wp_customize, 'heading', __( 'Heading font', 'blockbase' ), $heading_font_slug );
+		// Add a reset button
+		$this->font_control_default_body    = $body_font_default['slug'];
+		$this->font_control_default_heading = $heading_font_default['slug'];
+		$wp_customize->add_control(
```
scruffian, line 293: Can we make this only show when a non-default value is selected?

pbking: yup

pbking: Well MOSTLY.
It determines if they are set to default on startup.
If so, hide the button.

Any time the values changes, show the button.

Any time the values are reset to default, hide the button.

Only use case not covered is that if the values are set to the default values manually the button is still shown.  I didn't find an easy way to work around that and hope we can live with that limitation...

scruffian:
I get some warnings and notices with this:

```[14-Jul-2021 09:47:51 UTC] PHP Deprecated:  array_key_exists(): Using array_key_exists() on objects is deprecated. Use isset() or property_exists() instead in /var/www/html/wp-content/themes/themes/blockbase/inc/customizer/wp-customize-utils.php on line 7
[14-Jul-2021 09:47:57 UTC] PHP Deprecated:  array_key_exists(): Using array_key_exists() on objects is deprecated. Use isset() or property_exists() instead in /var/www/html/wp-content/themes/themes/blockbase/inc/customizer/wp-customize-utils.php on line 7
[14-Jul-2021 09:47:57 UTC] PHP Notice:  Undefined index: google in /var/www/html/wp-content/themes/themes/blockbase/inc/customizer/wp-customize-fonts.php on line 383
[14-Jul-2021 09:47:57 UTC] PHP Notice:  Undefined index: google in /var/www/html/wp-content/themes/themes/blockbase/inc/customizer/wp-customize-fonts.php on line 386

scruffian:
Pushed a commit to fix some notices. This LGTM

mikachan:
(With the caveat of my limited understanding..)

This looks like a great refactor. [This file](https://github.com/Automattic/themes/pull/4212/files#diff-1e8f5c8da22610cdd824ed551bb21abc65759f8961c8ea9d777a213a483c8284) is certainly easier to read and understand. I've tested out changing the fonts from custom to default too and it worked fine for me.

PR #4214, opened 2021-07-13
Quadrat: correct hover styles for titles and squares on block pattern

MaggieCabrera:

Following @beafialho [input](https://github.com/Automattic/themes/issues/3644#issuecomment-878930745) on the Query block pattern issue I have followed up with this PR that addresses those fixes:

This PR:

- Makes the post titles underline appear on hover instead of the other way around.
- Removes the rotation from the square and leaves just the opacity fade-in.

### DIFF
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -85,11 +85,13 @@
 	justify-content: flex-start;
 }
 
+.is-style-quadrat-diamond-posts .wp-block-post-title a,
 .is-style-quadrat-diamond-posts .wp-block-query-pagination a,
 .is-style-quadrat-diamond-posts .post-meta a {
 	text-decoration: none;
 }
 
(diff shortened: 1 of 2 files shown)


kjellr (approved):
I kind of like the rotation animation, but I'm happy to have Bea lead on that decision. 

The revised post title treatment makes sense, and matches up with the usual post title display. 👍

beafialho:
Looks good to me!

beafialho (approved):

scruffian:
I like it too, could we have two different patterns?

kjellr:
> I like it too, could we have two different patterns?

I don't think they'd be different enough to include two patterns — especially because the hover state won't appear at all on touch devices.

beafialho:
> I like it too, could we have two different patterns?

Sorry to come back to this. We can keep the rotation animation then. I checked with @melchoyce too and it seems like most folks like it better 👍

PR #4219, opened 2021-07-13
Blockbase: Switch logic for enabling Site Editor

scruffian:
This reverses the logic for hiding the Site Editor - we hide it by default and the checkbox will enable it:

Screenshot 2021-07-13 at 15 47 13
Screenshot 2021-07-13 at 15 46 58

### DIFF
--- a/blockbase/inc/disable-site-editor.php
+++ b/blockbase/inc/disable-site-editor.php
@@ -20,8 +20,9 @@ function add_disable_site_editor_setting() {
 		)
 	);
 
+	readd_legacy_admin_links();
+
 	if ( ! site_editor_enabled() ) {
-		readd_legacy_admin_links();
 		remove_site_editor_admin_link();
 	}
 }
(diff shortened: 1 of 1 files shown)


pbking:
Definitely agree with the inverse control done here.  Makes more sense to upgrade than to downgrade.
Pushed a small change to ALWAYS keep the legacy stuff active.

pbking:
One note: (unrelated to inverting the logic)
image

No matter what the setting is the page always shows "edit site" and never an option to use the customizer.

I see it's [similarly reported for .com](https://github.com/Automattic/themes/pull/4206#issuecomment-879176089) and seems that [this](https://github.com/Automattic/themes/pull/4206/commits/bf46f79aca59047ae42f8bdd1ab285eeed41cf24) also addresses that...  This comment can probably be ignored when both of these PRs join forces.

scruffian (commented):
LGTM

PR #4222, opened 2021-07-14
Stratford/Varia: Fixed a color mapping which was pointing to the wrong variable

DanielSmithMichigan:
[Original Issue](https://github.com/Automattic/wp-calypso/issues/52562)

### NOTE:
The site title is looking better, but it looks like something is wrong with the menu as well, I'm still looking into that.

---

As best as I can tell, it looks like the site-title was mistakenly set to use the primary-hover color all the time. We can't use the hover color for the site-title because it's not directly configurable from the [custom-colors customizer](https://user-images.githubusercontent.com/10274366/125542660-11c6f5c6-6edf-49e9-9280-e20a0e603413.png).

Another option we could pursue with this change would be to change the site-title to use the primary color. Then the site title would be black by default. I have chosen the secondary color for now so that I don't change the appearance of stratford, just fix the customizer for now.

https://user-images.githubusercontent.com/10274366/125543996-69010108-c5fa-4351-b732-6759c6c801ac.mov

### Testing Steps

1. SCP the stratford and varia folders from this pull request into your sandbox
1. Make sure your sandbox is enabled in the hosts file and calypso is using your sandbox in general
1. Create a new site
1. Either choose "edison" when you get to http://calypso.localhost:3000/new/design, OR choose whatever and switch to "stratford" theme later.
1. Go to the customizer
1. Wait a minute or so while the customizer loads (This step is questionable, I know. I'm still looking into it)
1. Go to "Colors & Backgrounds"
1. Change the secondary color
1. Note that the site title changes.

### DIFF
--- a/stratford/package.json
+++ b/stratford/package.json
@@ -1,6 +1,6 @@
 {
   "name": "stratford",
-  "version": "1.4.10",
+  "version": "1.4.11",
   "description": "Stratford",
   "bugs": {
     "url": "https://github.com/Automattic/themes/issues"
--- a/stratford/sass/_config-child-theme-deep.scss
+++ b/stratford/sass/_config-child-theme-deep.scss
@@ -292,8 +292,8 @@ $config-header: (
 		// Colors
 		"color": (
 			"text": map-deep-get($config-global, "color", "foreground", "light"),
-			"link": map-deep-get($config-global, "color", "primary", "hover"),
-			"link-hover": map-deep-get($config-global, "color", "primary", "default"),
+			"link": map-deep-get($config-global, "color", "secondary", "default"),
+			"link-hover": map-deep-get($config-global, "color", "secondary", "hover"),
 		),
 		// Fonts & Typography
 		"title": (
--- a/stratford/sass/_extra-child-theme.scss
+++ b/stratford/sass/_extra-child-theme.scss
@@ -125,7 +125,7 @@ a {
 							border-color: $color_secondary;
 							border-style: solid;
 							border-width: 0 0 .125em;
-							color: $color_primary_hover;
+							color: $color_secondary;
 						}
 					}
 				}
@@ -137,12 +137,12 @@ a {
 							border-color: $color_secondary;
 							border-style: solid;
 							border-width: 0 0 .125em;
-							color: $color_primary_hover;
+							color: $color_secondary;
 						}
 					}
 					a {
 						&::after {
-							color: $color_primary_hover;
+							color: $color_secondary;
 						}
 					}
 					ul {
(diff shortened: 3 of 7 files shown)


andres-blanco:
@DanielSmithMichigan This is working fine for me. The username color corresponds with the menu underline after applying this fix. As you said, there's still an issue with the selected menu item color, that according with the original issue: https://github.com/Automattic/wp-calypso/issues/52562 it also corresponds with the secondary color.

btw, you can use `./sandbox.sh clean` and `./sandbox.sh push` to make the testing changes easier https://github.com/Automattic/themes/#sandbox-tools

sixhours:
How will this affect existing users using colors on these themes? Will their site's colors suddenly change? It might be good to spot-check these changes with a few sandboxed user sites that use custom colors. I'm not sure how you'd find folks like that, though...Theme team might know. 🤔

DanielSmithMichigan:
> How will this affect existing users using colors on these themes? Will their site's colors suddenly change? It might be good to spot-check these changes with a few sandboxed user sites that use custom colors. I'm not sure how you'd find folks like that, though...Theme team might know. 🤔

I think existing users WILL see their colors change.

Would they be prompted to update their theme first?

The current pull request is basically two parts.

Part 1: Fix priority of CSS rules so that customizer colors override page colors.
Part 2: Fix page title color so that it's directly editable via the palette.

Shown below is what the fix looks like without part 2.


https://user-images.githubusercontent.com/10274366/125673145-e9a68c58-ba1b-4d95-afd3-ea4948590dac.mov

sixhours:
> Would they be prompted to update their theme first?

Not for Simple sites, and I don't think Atomic sites would, either. I know in the past we avoided making large visible changes to custom colors in theme annotations after launch for that reason. :-/ If we do end up doing that, probably a good idea to give Happiness a heads-up in case they get complaints. They can "fix" the color change with custom CSS if the user has an upgraded plan.

But it'a also possible we're talking about a tiny handful of people and this is a non-issue, I just asked because I know from past experience color annotation changes are finicky. :D

DanielSmithMichigan:
> > Would they be prompted to update their theme first?
> 
> Not for Simple sites, and I don't think Atomic sites would, either. I know in the past we avoided making large visible changes to custom colors in theme annotations after launch for that reason. :-/ If we do end up doing that, probably a good idea to give Happiness a heads-up in case they get complaints. They can "fix" the color change with custom CSS if the user has an upgraded plan.
> 
> But it'a also possible we're talking about a tiny handful of people and this is a non-issue, I just asked because I know from past experience color annotation changes are finicky. :D

It's a good point about existing user sites, as discussed in the p2 post, we will leave the default colors alone for now: pd2qbF-c9-p2

I am going to close this PR and open a new PR with a smaller subset of changes

DanielSmithMichigan:
New pull request: https://github.com/Automattic/themes/pull/4230

PR #4223, opened 2021-07-14
Quadrat: bring back rotation animation

MaggieCabrera:
Following up on [Bea's comment](https://github.com/Automattic/themes/pull/4214#issuecomment-879928385), I'm bringing back the rotation animation that was removed from that PR (the rest of the changes to the hover titles are not being reverted here)


scruffian (approved):
LGTM

PR #4225, opened 2021-07-14
Fix/quadrat main element

### LINKED ISSUE #4218: Quadrat: Add a main element to wrap the template content
Before we submit to dotorg we need to make sure that we have [a main element on each of the templates](https://github.com/Automattic/themes/issues/3981) in the same way [we did for Blockbase](https://github.com/Automattic/themes/pull/3987)

pbking:
Added (or converted existing div) `<main>` elements to all Quadrat page templates.

Removed the (now defunct) .page-content class.  (The padding this was providing is now being supplied by [other means](https://github.com/Automattic/themes/blob/trunk/blockbase/sass/base/_alignment.scss#L2).)

Fixes #4218

### DIFF
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -721,11 +721,6 @@ textarea:focus {
 	margin-top: 0;
 }
 
-.page-content {
-	padding-left: var(--wp--custom--post-content--padding--left);
-	padding-right: var(--wp--custom--post-content--padding--right);
-}
-
 .post-meta {
 	align-items: center;
 	justify-content: center;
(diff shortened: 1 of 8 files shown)


jffng (approved):
In the single template, the `main` was not wrapping the post content, I'm not sure why. So I [removed an extra closing div](https://github.com/Automattic/themes/pull/4225/commits/503134af3d6b992bfdc32124b530b5fe459028bd), and I think this is ready to bring in.

pbking:
Ah thank you @jffng.  When I changed the top `<div>` to `<main>` I couldn't find the matching `</div>` so just added the `</main>`.  I guess it just wasn't where I was expecting it.  Thanks for fixing it up.  That fix LGTM.

PR #4227, opened 2021-07-14
Blockbase: Update font weights

### LINKED ISSUE #4198: Blockbase: Fonts
The font weights/styles in [this file](https://github.com/Automattic/themes/blob/486aa30fbea19f695faae18460747b6b5041ace9/blockbase/inc/wp-customize-fonts.php) are probably not the best options. We should replace them with a normal, bold and italic option.

cc @kjellr

scruffian:
This updates the font weights we allow when customizing fonts in Blockbase. The Site Editor allows users to select _any_  font weight, but not all fonts are available in all weights. This makes as many weights available as possible.

To test, check what font weights are available here:
https://fonts.google.com/specimen/Work+Sans

and confirm that these are the same as in the file. Also check that the file can be loaded:
https://fonts.googleapis.com/css2?family=Work+Sans:ital,wght@0,100..900;1,100..900

The syntax allows for us to add a range of weights using `100..900`. This seems to work for some fonts and not others, which I don't understand, but for those that I doesn't work for, the long hand syntax works.

#### Related issue(s):
Fixes #4198


pbking:
From what I can tell 800 and 900 weight fonts aren't available to use by the user and aren't leveraged anywhere in any of the theme configurations.  I believe all of those can just load up to 700 weight to lighten the payload a bit.

pbking:
I stepped through all of the fonts and compared what rendered above with the expectations from the /specimin url as noted above.  Everything I expected was in alignment.  (Interestingly a few weights were showing italics in a manner I wasn't expecting, but that was a good thing.)

Apart from weights 800 and 900 not being necessary (from what I can tell, noted above) my only suggestion is that we eliminate 'Inter' as there are absolutely no italics available:

scruffian:
800 and 900 are possible using the Site Editor:
Screenshot 2021-07-14 at 20 07 42

pbking:
> 800 and 900 are possible using the Site Editor:

Today I Learned that control can be scrolled.  I really though "BOLD" was the last option.

🤦

pbking (approved):
This looks good to me.  The changes seem correct.

Sorry for being silly about the 800/900 font weights. :P

scruffian:
I made the same mistake at first :)

PR #4228, opened 2021-07-14
Fix/quadrat localization domain

pbking:
Minor changes to 

* add localization domain to localization function calls
* change localization function used (per theme-check warnings)

The code formatter also go ahold of the starter-content.php file and spaced some things around.

Not strictly necessary to pass the theme-check plugin but cleaning out the warnings seemed a good idea.


jffng (approved):
This all makes sense to me, thanks for handling it @pbking !

PR #4230, opened 2021-07-14
Stratford/Varia: Some color choices are ignored in customizer

DanielSmithMichigan:
| BEFORE (Watch the blue color of the text in the center).        | AFTER.           |
| ezgif-2-b3eb740e088e | ezgif-2-addba23d7eb6 |

### Testing Steps

1. Push this to your sandbox using:
```
cd themes; ./sandbox.sh clean;
```
Then
```
./sandbox.sh push
```
1. Make sure your sandbox is enabled in the hosts file and calypso is using your sandbox in general
1. Create a new site with stratford as the theme
1. Go to the customizer
1. Give the customizer a minute to load.
1. Go to "Colors & Backgrounds"
1. Change colors and notice that they are applied as you change them.

#### NOTE: When you're on the default colors, the very first color change will produce some expected results. But after that change, the color changes will be what you would expect. See this p2 post for more information: pd2qbF-c9-p2

### DIFF
--- a/alves/package-lock.json
+++ b/alves/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "alves",
-  "version": "1.5.9",
+  "version": "1.5.10",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/alves/package.json
+++ b/alves/package.json
@@ -1,6 +1,6 @@
 {
   "name": "alves",
-  "version": "1.5.9",
+  "version": "1.5.10",
   "description": "Alves",
   "bugs": {
     "url": "https://github.com/Automattic/themes/issues"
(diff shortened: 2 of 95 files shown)


andres-blanco:
This works on stratford and Varia! 
The good news is that this also seems to fix the problem on other Varia child themes, I tested on 
 * Alves
 * Hever

andres-blanco:
This does fix Hever on Atomic also! 🚀 
| Before | After |
|------|-----|
|  | |

andres-blanco (approved):
LGTM! This seems to fix a really serious issue (when you click on colors it does nothing) on Varia and (some of) it's children.

DanielSmithMichigan:
D64265-code

PR #4234, opened 2021-07-15
Blockbase + Children: fix Marketing bar overlap

### LINKED ISSUE #4226: Blockbase + Children: Marketing bar
## Steps to replicate

On your sandbox, comment line 22 of `wp-content/blog-plugins/marketing-bar.php` to show the marketing bar on any site. Check Blockbase and its children.

## Result

The marketing bar covers the menu.

Screenshot 2021-07-14 at 17 47 44

## Expected

The marketing bar includes a class on the body (`.has-marketing-bar`) that adds a margin top to move the content down but Blockbase overrides that using `!important` [here](https://github.com/Automattic/themes/blob/7c0b389fbf2060ab3239d418e6fca95982146cbf/blockbase/sass/base/_normalize.scss#L4). If we didn't have that the margin top would move the content down and the menu wouldn't be covered. The important comes from [this PR](https://github.com/Automattic/themes/pull/4176).

MaggieCabrera:

Removed !important from the normalize file so the marketing bar can push the body down when it's present. It looks like the original purpose of these important rules are no longer the case:

FSE:

Untitled

Frontend with marketing bar:

Screenshot 2021-07-15 at 09 42 49

#### Related issue(s):

Fixes https://github.com/Automattic/themes/issues/4226

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -3,8 +3,8 @@
  * - Reset the browser
  */
 body {
-	margin: 0 !important;
-	padding: 0 !important;
+	margin: 0;
+	padding: 0;
 }
 
 body {
--- a/blockbase/sass/base/_normalize.scss
+++ b/blockbase/sass/base/_normalize.scss
@@ -1,8 +1,8 @@
 
 // Remove the margin in all browsers.
 body {
-  margin: 0 !important;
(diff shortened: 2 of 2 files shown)


mikachan (commented):
This looks good to me, can't see any issues with !important not being used anymore.

PR #4237, opened 2021-07-15
Quadrat: replace screenshot

jffng:

Replaces the theme screenshot so the menu matches the demo site. Here is the replaced version: https://github.com/Automattic/themes/blob/2a6805a4a929b718957b730e6c1028529a8fb137/quadrat/screenshot.png


beafialho (approved):

kjellr:
@beafialho: We should like us to re-adjust this so that "matter" isn't all alone on its own line, and also I think the screenshot needs to be 1200x900, like our other theme screenshots.

beafialho:
You're right @kjellr, thanks for pointing that out. Here's the screenshot with those dimensions.

quadrat_screenshot_1200x900

PR #4238, opened 2021-07-15
Quadrat: Change the name of the front page template

scruffian:
This renames the front page template.

Before this change, the front page of the site would always display a single post. In order to be able to display the latest posts on front page, we need a query block to load on the front page. By changing the name of this template to `page-without-title`, we load the index on the front page, and make it possible to have a page without a title as a front page by setting a static front page.

To test:

- Open the customizer
- Set the front page to be latest posts and confirm that you see the latest posts on the front page
- Set the front page to be a static page and confirm that you see that page
- Set the template of your static page to be `page-without-title` and confirm that you no longer see the title on the front page.


MaggieCabrera (changes requested):
I think we now need to add this new template to theme.json under customTemplates, otherwise, this looks ok to me and works as expected.

scruffian:
I updated this to add it to the theme.json. I also had to update the build script to merge this array rather than overwrite it. Hopefully that will be useful if there are other array we want to merge.

MaggieCabrera:
> I updated this to add it to the theme.json. I also had to update the build script to merge this array rather than overwrite it. Hopefully that will be useful if there are other array we want to merge.

Makes sense, since the child themes are inheriting the rest from blockbase anyway.

MaggieCabrera (approved):

PR #4240, opened 2021-07-16
Blockbase: fixed alignments on single and custom templates

### LINKED ISSUE #4239: Blockbase: Full width alignments not working on the frontend
## Steps to replicate

Create a new page and make a full width cover block. Check the frontend.

## Result

Editor:

Screenshot 2021-07-16 at 11 40 24

Frontend:

Screenshot 2021-07-16 at 11 40 20

## Expected

I expected the frontend to look like the editor

MaggieCabrera:

This PR fixes alignments that were broken after we added the extra main wrapper required for the skip links to work.

To test this, create a page with blocks set to different alignments. Test the singular template and the header-and-footer-only one. The frontend and the editor should match.

Right now the nested group block inside the cover doesn't work with any alignments. I don't know if it's a theme issue or a GB issue.

Editor:

Screenshot 2021-07-16 at 11 54 36

Frontend:

Screenshot 2021-07-16 at 11 54 46

#### Related issue(s):

Fixes https://github.com/Automattic/themes/issues/4239

### DIFF
--- a/blockbase/block-templates/singular.html
+++ b/blockbase/block-templates/singular.html
@@ -8,9 +8,9 @@
 
 <!-- wp:group {"tagName":"main","layout":{"inherit":true}} -->
 <main class="wp-block-group post-content">
-<!-- wp:post-featured-image {"align":"full"} /-->
+<!-- wp:post-featured-image {"align":"wide"} /-->
 
-<!-- wp:post-content {"layout":{"inherit":true}, "align":"full"} /-->
+<!-- wp:post-content {"layout":{"inherit":true}} /-->
 </main>
 <!-- /wp:group -->
 
--- a/blockbase/block-templates/singular.html
+++ b/blockbase/block-templates/singular.html
@@ -10,7 +10,7 @@
 <main class="wp-block-group post-content">
 <!-- wp:post-featured-image {"align":"full"} /-->
 
-<!-- wp:post-content /-->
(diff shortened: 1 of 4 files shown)


MaggieCabrera:
Looks like the nested issue is a GB problem so ignore it from this PR, this is what the same page looks like on emptytheme:

Screenshot 2021-07-16 at 11 59 45

MaggieCabrera:
This PR requires of this GB fix: WordPress/gutenberg#33501

blockbase/block-templates/singular.html
```diff
@@ -8,9 +8,9 @@
 
 <!-- wp:group {"tagName":"main","layout":{"inherit":true}} -->
 <main class="wp-block-group post-content">
-<!-- wp:post-featured-image {"align":"full"} /-->
+<!-- wp:post-featured-image {"align":"wide"} /-->
```
scruffian, line 11: We probably want to change this back, but it's useful for demonstrating the problem

jffng:
I am trying to think of another way to solve it without requiring the GB change to exclude child containers from the alignment rules.

> after we added the extra main wrapper required for the skip links to work

I kind of wish the `post-content` block just outputted the blocks themselves, not an additional wrapper. What is its purpose?If it's necessary, what if we were were able to pass a tag attribute (like we can for the group block) to the `post-content` block?

MaggieCabrera:
Closing this in favor of #4253

PR #4244, opened 2021-07-16
Quadrat: reference bundled assets inside patterns

### LINKED ISSUE #4243: Quadrat: fix images for dotorg
We got feedback from the [dotorg submission](https://themes.trac.wordpress.org/ticket/102464):

>      Are all the illustrations are self-made?
>      The image that is used on patterns should be bundled in the theme. No external links are allowed.

jffng:

Theme review to dotorg flagged our referencing of non-local assets in patterns. This PR updates them.

I don't like including an 8mb wav for the Listen pattern, but not sure what the alternative is there. We can compress it to mp3 which would reduce the file size a lot, but it would still probably be around 1mb.

#### Related issue(s):

Addresses #4243


scruffian:
I replaced the WAV with an MP3. I think we can merge this...

scruffian (approved):

scruffian:
Since the theme isn't launch yet I don't think we need to bump the version number yet but not a big deal..

PR #4245, opened 2021-07-16
Blank Canvas: update screenshot to make it looks like something there

enejb:
Update the screenshot for blank canvas theme to make the theme easier to select and not look so broken in the mobile apps See pb3aDo-13J-p2.


pbking:
@kjellr would like your thoughts on universally using the screenshot with verbiage.

kjellr (approved):
On the P2 post it sounded like the dotcom design folks are on board with this screenshot, so seems fine to me to replace it.

Issue #4246, opened 2021-07-16
Quadrat: make illustrations easier to access


jffng:
We can't bundle all of Quadrat's illustrations via headstart for systems (pbxlJb-gM-p2#comment-310) + UX ( p1626445676107400/1626442192.089100-slack-C029FM1EH) reasons. 

Instead, we should make them available in the following ways:

- [x] as a downloadable zip, linked in the theme showcase entry
- [x] as individual patterns

cc @kjellr

jffng:
Closed by #4247

PR #4247, opened 2021-07-16
Quadrat: add illustrations as patterns

jffng:
This PR adds all of Quadrat's bundled illustrations as individual block patterns.
 
#### Related issue(s):

Partially address #4246


quadrat/inc/block-patterns.php
```diff
@@ -14,6 +14,11 @@ function quadrat_register_block_patterns() {
 				'quadrat',
 				array( 'label' => __( 'Quadrat', 'quadrat' ) )
 			);
+
+			register_block_pattern_category(
+				'theme-images',
+				array( 'label' => __( 'Theme Images', 'quadrat' ) )
```
scruffian: Maybe this should be "Quadrat images"?

scruffian:
I added a commit to do this programatically. Not sure which approach I prefer...

quadrat/inc/block-patterns.php
```diff
@@ -43,3 +43,59 @@ function quadrat_register_block_patterns() {
...
+				'combination-of-purple-triangles' => __( 'Combination of purple triangles' , 'quadrat' ),
+				'combination-of-white-orange-and-pink-triangles' => __( 'Combination of white orange and pink triangles' , 'quadrat' ),
+				'complex-shape-combination-and-two-pink-squares-on-a-dark-purple-background' => __( 'Complex shape combination and two pink squares on a dark purple background' , 'quadrat' ),
+				'group-of-squares-displayed-diagonally' => __( 'Group of squares displayed diagonally' , 'quadrat' ),
+				'group-of-white-pink-and-orange-circles' => __( 'Group of white pink and orange circles' , 'quadrat' ),
+				'many-pink-squares-overlapped' => __( 'Many pink squares overlapped' , 'quadrat' ),
+				'orange-and-pink-shapes-displayed-in-different-ways' => __( 'Orange and pink shapes displayed in different ways' , 'quadrat' ),
+				'orange-triangle-in-purple-background' => __( 'Orange triangle in purple background' , 'quadrat' ),
+				'pink-square-and-dark-purple-triangle' => __( 'Pink square and dark purple triangle' , 'quadrat' ),
+				'pink-square-with-dark-shadow-on-orange-background' => __( 'Pink square with dark shadow on orange background' , 'quadrat' ),
```
MaggieCabrera, line 69: The file for this one starts with an uppercase P, we may want to fix that

MaggieCabrera (approved):
Besides my other comment, I think this works fine like this.

quadrat/inc/block-patterns.php
MaggieCabrera: +1

quadrat/inc/block-patterns.php
scruffian: fixed

scruffian (approved):

jffng:
Thanks for the assist! The programatic approach works for me, fewer files that way...

kjellr:
I think the most obvious use case for these illustrations is as featured images, but this doesn't really make that possible. 😕 But I think it's fine for now. I was about to suggest we call the category "Quadrat Illustrations" instead of "Quadrat Images", but I'm not sure it's worth opening up another PR for that small change.

jffng:
> I think the most obvious use case for these illustrations is as featured images, but this doesn't really make that possible. 😕

I agree it's not ideal. On dotorg we have the starter content approach, and on dotcom there was pushback from both systems and dotcom to automatically add these to a user's media library on theme switch. If you have other ideas, maybe we can discuss here? https://github.com/Automattic/themes/issues/4246

kjellr:
I think the zip option there is fine for now, so paired with this PR I think we're good.

Issue #4149, opened 2021-07-01
All themes: supporting block templates on 5.8


MaggieCabrera:
Coming 5.8 a new featured is going to be introduced: [block templates](https://make.wordpress.org/core/2021/06/16/introducing-the-template-editor-in-wordpress-5-8/). With this feature, classic themes can provide the user with the opportunity to create their own block-based template for any new pages that use them.

Initially just adding `add_theme_support( 'block-templates' );` should be enough to support this feature but that may not be enough, on my testing, Seedlet wasn't behaving on the frontend and it may need some CSS love for the alignments:

Editor | Frontend
--- | ---
<img width="1254" alt="Screenshot 2021-07-01 at 14 27 24" src="https://user-images.githubusercontent.com/3593343/124125721-3c738100-da7a-11eb-915d-e068d2495063.png"> | <img width="1601" alt="Screenshot 2021-07-01 at 14 27 19" src="https://user-images.githubusercontent.com/3593343/124125715-3b425400-da7a-11eb-8303-85006e6d6f7b.png">

I suggest that we add this to Blockbase, Seedlet and Varia + children first and then consider adding it to the rest of our most used themes.

scruffian:
We already have a PR that does this!
https://github.com/Automattic/themes/pull/4141/files

scruffian:
> I suggest that we add this to Blockbase, Seedlet and Varia + children first and then consider adding it to the rest of our most used themes.

Do we need to add this to block themes? It seems to already be there in Blockbase:
<img width="353" alt="Screenshot 2021-07-08 at 07 58 53" src="https://user-images.githubusercontent.com/275961/124876818-59e8a380-dfc2-11eb-92bc-fb9e9518a1ee.png">

MaggieCabrera:
Happiness have been asking if this feature will be rolled out to all of our themes, I said we are targeting our most popular themes first but once we do, we should P2 about it and tag them so they are on the loop.

inaikem:
I'd like to ask if adding block support to classic themes is going to extend to Newspack and its child themes, too.

MaggieCabrera:
Blockbase doesn't need it because block themes have it on by default.

PR #4251, opened 2021-07-19
Added block template support to seedlet & spearhead.

pbking:

Added Block Template support to varia and all varia children.
Defined a default template consisting of just the post content.

To test:
* Activate Seedlet or Spearhead theme.
* Edit a page/post
* Note the new 'Template' panel.
* Create a new template.
* Save the template, save the page.
* Observe the view for that new page; it should follow the design of the template.

#### Related issue(s):

Partially addresses #4149

### DIFF
--- a/seedlet/functions.php
+++ b/seedlet/functions.php
@@ -269,6 +269,16 @@ if ( ! function_exists( 'seedlet_setup' ) ) :
 				'enable_theme_default' => true,
 			)
 		);
+
+		// Add support for block templates (default template is content-only)
+		add_theme_support( 'block-templates' );
+		add_filter(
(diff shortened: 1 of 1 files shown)


pbking:
Related to #4251

jffng (commented):
Running GB trunk + this PR, I'm still getting the post title and tagline and a separator as what's loaded in the template by default:

Screen Shot 2021-07-19 at 5 02 30 PM

Do I need to be on 5.8? I would expect my WP version shouldn't matter if I'm running GB trunk.

scruffian (approved):
I see the same as @jffng, but when I upgrade to 5.8 it works.

PR #4252, opened 2021-07-19
Removed unused post-content class

pbking:

While poking around alignment classes I found an instance of "post-content" class which Blockbase USED to use for top-level alignment purposes but has sense been replaced by other methods.  I believe this class was just missed when the rest were removed.


jffng (commented):
I noticed the same, thanks for the PR!

What about the `post-header` class a few lines up, can we remove that too?

pbking:
:) I was just coming here to do that!

jffng (approved):
🚢

PR #4253, opened 2021-07-19
Unnested blockbase post-content blocks from parent layouts

### LINKED ISSUE #4239: Blockbase: Full width alignments not working on the frontend

pbking:

This change ensures that all instances of post-content blocks have `{"layout":{"inherit":true}}` and are the TOPMOST container that has `{"layout":{"inherit":true}}` applied.

The post editor assumes the above situation (and is USUALLY correct).  This change adjusts all templates in the theme so that it can be correct in the assumption 100% of the time.

#### Related issue(s):

Addresses: #4239
An alternative to: #4240
Relevant discussion in this issue: https://github.com/WordPress/gutenberg/pull/33501

### DIFF
--- a/blockbase/block-templates/search.html
+++ b/blockbase/block-templates/search.html
@@ -1,4 +1,4 @@
-<!-- wp:template-part {"slug":"header","tagName":"header","layout":{"inherit":true}} /-->
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
 <!-- wp:group {"layout":{"inherit":true},"tagName":"main"} -->
 <main class="wp-block-group">
--- a/blockbase/block-templates/singular.html
+++ b/blockbase/block-templates/singular.html
@@ -6,11 +6,11 @@
 </div>
 <!-- /wp:group -->
 
-<!-- wp:group {"tagName":"main","layout":{"inherit":true}} -->
+<!-- wp:group {"tagName":"main"} -->
 <main class="wp-block-group">
 <!-- wp:post-featured-image {"align":"full"} /-->
 
-<!-- wp:post-content /-->
+<!-- wp:post-content {"layout":{"inherit":true}} /-->
 </main>
 <!-- /wp:group -->
 
--- a/blockbase/block-templates/header-footer-only.html
+++ b/blockbase/block-templates/header-footer-only.html
@@ -1,6 +1,6 @@
 <!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
-<!-- wp:group {"tagName":"main","layout":{"inherit":true}} -->
+<!-- wp:group {"tagName":"main"} -->
 <main class="wp-block-group">
 
 <!-- wp:post-content {"layout":{"inherit":true}} /-->


blockbase/block-templates/search.html
```diff
@@ -1,4 +1,4 @@
-<!-- wp:template-part {"slug":"header","tagName":"header","layout":{"inherit":true}} /-->
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
```
pbking, line 1: This I think was an unrelated change, but was discovered while scrutinizing the layout properties in the templates.  This particular change just makes the markup of all usages of the header template part the same.

blockbase/block-templates/singular.html
```diff
@@ -6,11 +6,11 @@
 </div>
 <!-- /wp:group -->
 
-<!-- wp:group {"tagName":"main","layout":{"inherit":true}} -->
+<!-- wp:group {"tagName":"main"} -->
 <main class="wp-block-group post-content">
 <!-- wp:post-featured-image {"align":"full"} /-->
```
jffng, line 11: I think we can get rid of the align full here now too?
```suggestion
<!-- wp:post-featured-image /-->
```

jffng (approved):
I prefer this to #4240 because it doesn't require making an exception for how alignment of nested containers is handled.

blockbase/block-templates/singular.html
pbking: I don't think so.  The intention is for that image to be alight full right?  
I'm not sure that "don't do alignment" (the block with no alignment expressed, and not inside a group that expresses alignment) is the same as "align full" (even though this align full is outside of a container that expresses alignment configuration).

jffng: I see, so the idea is that even though it shows up full width without it, it's better to declare that the intention is for the image to be full width.

What's a bit unintuitive to me is that you can also set the alignment to "wide" or "default" and it won't do anything because it's not inside a container.

pbking: Not quite what I'm saying...

I believe the "natural" state of being is that the featured image will be whatever width the image is, no constraints. 
 
If it's inside of a container expressing alignment opinion (such as a group with inheritLayout:true) then it follows the rules of that container.  Without align full it will be "content" width and with "align full" it will be the full width of said container.

In this scenario the featured image isn't in a container expressing opinion; it's not going to stretch to be "content width" it's just going to ... be there.  And be however wide the image makes the block be.  Therefore by adding the "full width" option to the featured image block it's going to stretch to fill the entire width of its container (which, in this case is the main group which is a group that is the full width of the page) which is the design target of the theme.

> it won't do anything because it's not inside a container.

But it IS in a container, just not in a container that's "inheriting layout".

jffng: Ahhh I understand why the align full is needed now, thanks for clarifying. 

> But it IS in a container, just not in a container that's "inheriting layout".

👍 

I guess my point is that in this scenario, doesn't setting `align: wide` on the featured image have the same result as setting `align: full`, or am I missing something? The featured image stretches to 100% of its container at wide or full width, which I still think is unintuitive but unrelated to this PR :)

pbking: > The featured image stretches to 100% of its container at wide or full width, which I still think is unintuitive but unrelated to this PR :)

Yes, I agree with that!
IMO there's no such thing as "wide width" for that block in this given scenario (since that's a concept introduced by a block inheriting layout) whereas a "full width" seems like a reasonable option for ANY block in any container.  IMO (in this scenario) a "wide width" class on that featured image should do the same thing as a "pickle sandwich" class (having no effect).  But yea, that's a different issue. :)

PR #4254, opened 2021-07-20
Quadrat: Add podcast template

scruffian:
To make it easier to use the diamond query loop we could create a template for it like this.

I was trying to set up the demo to use this template and I thought it would have been much easier to do if we had a template already. Thoughts?


jffng:
What if we made this the index or archive template? 

I think it is unintuitive to have a template named "podcast" when the template just displays a list of posts in a nice way.

scruffian:
We could make this the index but I thought it was useful to have a choice of index templates...

jffng:
Yeah it is useful, and I'm open to it because I don't see a ton of harm in shipping an additional template. Another idea is to add this (or the current index layout) as a pattern?

scruffian:
We don't need this, the pattern is enough

PR #4263, opened 2021-07-21
Seedlet: fixes spacing on query block children

### LINKED ISSUE #4207: Seedlet: Blocks within the Query Loop block do not inherit correct front-end margins
When using the Query Loop block, the child block margins don't match between the front-end and the editor. The editor appears more like I'd expect. 

This happens with Seedlet and its child themes. 

## Steps to replicate

1. Add a Query Loop block with an assortment of children.
2. Save your post or page, view it on the front end.  

## Screenshots

Editor|Front-end
---|---
Screen Shot 2021-07-12 at 10 49 31 AM|Screen Shot 2021-07-12 at 10 49 39 AM

MaggieCabrera:

Seedlet has a specific set of rules for spacing blocks that wasn't being applied to the query block. This PR adds support for them. @kjellr this applies the same spacing rules that the rest of the blocks have throughout the theme on the frontend, but it's hard to gauge compared to the editor with all the placeholders. Let me know what you think about this solution.

Editor:

Screenshot 2021-07-21 at 16 48 21

Before | After
--- | ---
Screenshot 2021-07-21 at 16 50 10 | Screenshot 2021-07-21 at 16 48 11

#### Related issue(s):

Fixes https://github.com/Automattic/themes/issues/4207


kjellr:
The main way to test if this is working is to compare the front-end to the editor and see if it looks about the same. This looks pretty good. The one omission I'm noticing is that in the editor, there's a little bottom margin between each individual post, but that's not present in the front end. Note the spacing between the date and the title below:

Editor|Front end
---|---
Screen Shot 2021-07-21 at 1 17 22 PM|Screen Shot 2021-07-21 at 1 17 01 PM

To correct that, maybe these styles should apply to `.wp-block-post-template > *` in addition to `.wp-block-post`?

MaggieCabrera:
> To correct that, maybe these styles should apply to `.wp-block-post-template > *` in addition to `.wp-block-post`?

You are right, I just added that

pbking (approved):
Reviewed based on all the feedback above.
I didn't see any spacing issues once I got the block situated.
In addition to the above I also tested with another of additional blocks in the loop including featured image.  Seemed to space out as expected.

I believe this is ready to come in.

PR #4264, opened 2021-07-21
Added block template support to seedlet.

pbking:
This is the same PR as #4251 which was previously merged to trunk and then reverted (because the feature was not yet ready on .com).

Partially address #4149 

This shouldn't be brought in until 5.8 has landed more firmly on .com and this feature has been re-evaluated.

### DIFF
--- a/seedlet/functions.php
+++ b/seedlet/functions.php
@@ -269,6 +269,16 @@ if ( ! function_exists( 'seedlet_setup' ) ) :
 				'enable_theme_default' => true,
 			)
 		);
+
+		// Add support for block templates (default template is content-only)
+		add_theme_support( 'block-templates' );
+		add_filter(
+			'block_editor_settings_all',
+			function( $settings ) {
+				$settings['defaultBlockTemplate'] = '<!-- wp:post-content /-->';
+				return $settings;
+			}
+		);
 	}
 endif;
 add_action( 'after_setup_theme', 'seedlet_setup' );
@@ -307,7 +317,7 @@ function seedlet_fonts_url() {
 
 		/**
 		 * A filter to enable child themes to add/change/omit font families.
-		 * 
+		 *
 		 * @param array $font_families An array of font families to be imploded for the Google Font API
 		 */
 		$font_families = apply_filters( 'included_google_font_families', $font_families );
--- a/blank-canvas/style.css
+++ b/blank-canvas/style.css
@@ -12,7 +12,7 @@ License: GNU General Public License v2 or later
 License URI: http://www.gnu.org/licenses/gpl-2.0.html
 Template: seedlet
 Text Domain: blank-canvas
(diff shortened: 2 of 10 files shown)


MaggieCabrera:
I'm getting this error on dotcom when I try to create a new template. Local works fine. Is this the error that caused the previous revert?
Screenshot 2021-08-10 at 11 56 40

kjellr:
As part of this PR, I think we should also add [the new `block-templates` tag](https://core.trac.wordpress.org/ticket/53556) to Seedlet and its child themes. When those themes get uploaded to WP.org, they should have that in place so that they [show up here](https://wordpress.org/themes/tags/template-editing/). 

(As of this comment, there are no themes in the directory that support this feature).

pbking:
Error is no longer happening. (Progress)
~~However, the custom template being supplied in the `block_editor_settings_all` filter is being ignored.  It is working OK locally, but not on wpcom.~~
And template seems to be behaving as expected.

A 2nd verifying this is working as expected would be appreciated though.

seedlet/functions.php
```diff
@@ -269,6 +269,16 @@ function seedlet_setup() {
 				'enable_theme_default' => true,
 			)
 		);
+
+		// Add support for block templates (default template is content-only)
+		add_theme_support( 'block-templates' );
+		add_filter(
+			'block_editor_settings_all',
+			function( $settings ) {
+				$settings['defaultBlockTemplate'] = '<!-- wp:post-content /-->';
```
jffng, line 278: As discussed on Slack p1628868491002800/1628867157.498300-slack-C029FM1EH, it may help to match alignments between template editor and view to wrap this in a group block:

```suggestion
				$settings['defaultBlockTemplate'] = '<!-- wp:group -->
				<div class="wp-block-group"><!-- wp:post-content /--></div>
				<!-- /wp:group -->';
```

jffng (commented):
I gave this a spin and it is working well except for on wpcom: 

Screen Shot 2021-08-13 at 11 55 42 AM
 
Not sure what is going on there, the console error is opaque: 

`Stylesheet seedlet-wpcom-editor-style-css was not properly added.`

pbking:
RE: `publish failed: invalid paremeters` testing with `edge` Gutenberg seems to eliminate the issue.  (I think?) Perhaps we just need to wait a bit longer...

pbking:
This continues to fail (with the same error noted above: `invalid parameters: template`) with Gutenberg 11.4.0 and 11.4.1 on wpcom.

MaggieCabrera:
If this is still broken we should probably close this

pbking:
Still broken. :(
image

The reason isn't clear, and even less so trying to look at it through the fog of time.

I don't think that our non-block themes are going to get block template support on wpcom (without some effort somewhere that I can't contribute to).

PR #4250, opened 2021-07-19
Added block template support to Varia.

pbking:
#### Changes proposed in this Pull Request:

Added Block Template support to varia and all varia children.
Defined a default template consisting of just the post content.

To test:
* Activate any varia theme.
* Edit a page/post
* Note the new 'Template' panel.
* Create a new template.
* Save the template, save the page.
* Observe the view for that new page; it should follow the design of the template.

#### Related issue(s):

Partially addresses #4149

PR #4265, opened 2021-07-21
Added block template support to Varia.

pbking:
This is the same PR as #4250 which was previously merged to trunk and then reverted (because the feature was not yet ready on .com).

Partially address #4149 

This shouldn't be brought in until 5.8 has landed more firmly on .com and this feature has been re-evaluated.

### DIFF
--- a/varia/functions.php
+++ b/varia/functions.php
@@ -223,6 +223,16 @@ if ( ! function_exists( 'varia_setup' ) ) :
 			)
 		);
 
+		// Add support for block templates (default template is content-only)
+		add_theme_support( 'block-templates' );
+		add_filter(
+			'block_editor_settings_all',
+			function( $settings ) {
+				$settings['defaultBlockTemplate'] = '<!-- wp:post-content /-->';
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
When we merge this the Showcase of these themes needs to be updated to include the `Template Editor` tag (the showcase taxonomy, not the style.css tag btw). The description of the tag would be "The theme supports Custom Colours, Sticky Posts, and the Template Editor." (convo: p1628267836046200/1628267129.037800-slack-C01DF571R5Y)

MaggieCabrera:
We need to add `block-templates` tag to style.css for this one too

pbking:
This continues to fail (with the same error noted above: `invalid parameters: template`) with Gutenberg 11.4.0 and 11.4.1 on wpcom.

MaggieCabrera:
Should we close this one?

pbking:
Still broken on wpcom.  I don't suppose varia and kin will have access to block templates.

Closing this one.

Issue #3653, opened 2021-04-14
Quadrat: translation


jffng:
There are a few strings in the theme's block templates that we should figure out what to do with:

- [x] 404.html
- [x] footer.html
- [x] query.html (query navigation labels)
- [x] search.html

jffng:
After discussing with the team, I don't think this within the realm of the theme to solve, especially not for v1.

Gutenberg needs to eventually expose a mechanism to translate strings inside block templates, strings + labels that are rendered by FSE blocks, etc. ~This is a known and tracked stream of work.~

_Edit_
I'm not sure where i18n of FSE is on the roadmap or but here are a couple related issues:

https://github.com/WordPress/gutenberg/issues/27402
https://github.com/WordPress/gutenberg/issues/21932

scruffian:
Removing this from TODO as it blocked.

scruffian:
Another attempt: https://github.com/WordPress/gutenberg/pull/33217

jeffikus:
Tested this with @MaggieCabrera - marking as closed.

PR #4269, opened 2021-07-22
Quadrat: Remove labels from Pagination links

scruffian:
We added these labels so that the default &raquo; and &laquo; arrows didn't show. Now that https://github.com/WordPress/gutenberg/pull/33626 has merged, we don't need to do this, so this PR just removes the labels. This means these buttons will now be translated.

#### Related issue(s):
Should help with: https://github.com/Automattic/themes/issues/3653


jffng (approved):
Looks good, there is no visible change to the pagination.

Issue #4268, opened 2021-07-22
Quadrat: Hide query pagination divider if there's no pagination


kjellr:
If there's no pagination available, we should not display the horizontal divider line for the Query Pagination: 

<img width="1247" alt="Screen Shot 2021-07-22 at 7 52 50 AM" src="https://user-images.githubusercontent.com/1202812/126635129-ab40d3f9-7bb4-4c83-89e9-79430b2765ff.png">

<img width="1247" alt="Screen Shot 2021-07-22 at 7 52 47 AM" src="https://user-images.githubusercontent.com/1202812/126635138-f551a79a-cfea-4de7-b96f-c5b78cbc1a1f.png">

jffng:
I wonder if this should ultimately be addressed in the block.

That is, if there is no pagination, an empty div should not be rendered. I will take a look at fixing it there.

kjellr:
@jffng I'm not 100% sure, but I believe this happens with many other blocks too — for example, if there's no post title, an empty div for post title still shows up on the front end.

jeffikus:
This needs retesting.

scruffian:
This is waiting for GB on dotcom to be updated.

PR #4272, opened 2021-07-22
Quadrat: hide query pagination if there's no pagination

jffng:

This change relies on ~https://github.com/WordPress/gutenberg/pull/33635~ https://github.com/WordPress/gutenberg/pull/35092

To test, activate this PR on a completely empty site or one with less than five posts ( so no pagination appears ).

#### Related issue(s):

Solves https://github.com/Automattic/themes/issues/4268

### DIFF
--- a/quadrat/block-templates/index.html
+++ b/quadrat/block-templates/index.html
@@ -11,7 +11,7 @@
 		<div style="height:60px" aria-hidden="true" class="wp-block-spacer"></div>
 		<!-- /wp:spacer -->
 	<!-- /wp:post-template -->
-	<!-- wp:query-pagination {"align":"wide"} -->
+	<!-- wp:query-pagination {"align":"wide","paginationArrow":"arrow"} -->
 	<!-- wp:query-pagination-previous /-->
 	<!-- wp:query-pagination-numbers /-->
 	<!-- wp:query-pagination-next /-->
--- a/quadrat/block-templates/index.html
+++ b/quadrat/block-templates/index.html
@@ -11,10 +11,17 @@
 		<div style="height:60px" aria-hidden="true" class="wp-block-spacer"></div>
 		<!-- /wp:spacer -->
 	<!-- /wp:post-template -->
+<<<<<<< HEAD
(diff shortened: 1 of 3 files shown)


jffng:
I believe this is ready for a review. 

To test, check out this PR and make sure pagination is appearing correctly. Then, adjust your site so that fewer than five posts exists. No pagination container or dividing line should be rendered.

quadrat/block-templates/index.html
```diff
@@ -1,6 +1,6 @@
 <!-- wp:template-part {"slug":"header"} /-->
 
-<!-- wp:query {"tagName":"main","layout":{"inherit":true},"queryId":1,"query":{"perPage":5,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":"","inherit":true}} -->
+<!-- wp:query {"tagName":"main","layout":{"inherit":true},"queryId":1,"query":{"perPage":5,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":"","inherit":false}} -->
```
jffng, line 3: I had to adjust inherit to false for pagination to work at all... This matches skatepark's index query too: https://github.com/Automattic/themes/blob/trunk/skatepark/block-templates/index.html#L6

MaggieCabrera:
Should we do this on search.html and query-diamond.php too?

MaggieCabrera:
This looks fine to me locally but wpcom still doesn't have the GB change

MaggieCabrera:
Screenshot 2021-10-01 at 10 04 46

I see this on dotcom, don't know if dotcom has the fix or not

jffng:
> Should we do this on search.html and query-diamond.php too?

Oh yeah, I will update there too. 

> I see this on dotcom, don't know if dotcom has the fix or not

Ah I see, shall we hold off on merging until the fix is there?

MaggieCabrera:
Yeah, we talked about holding this one until It lands since the bug is very minor

scruffian:
We should do this for all themes: https://github.com/Automattic/themes/pull/4805

PR #4274, opened 2021-07-23
Improve anchor link scroll position

### LINKED ISSUE #3145: Page Jumps jumping too far down
When clicking a link that is linked to an anchor tag (Page jumps), the jump takes you to below the Heading/text where the anchor tag is located. 
#### Steps to reproduce

ezgif-3-da7b90eb5d31

What I expected:

I expected the Page Jump to take you to the respective section and still show the heading, i.e. the page jump shouldn't go further down the text that has the Anchor tag. 

This was also brought up in 721-gh-en.support-docs-content that made reference to a CSS request p1594910897307800-slack-C025Q5RK2 but it seems we should consider improving the Page jump in anchor tags in general.

mikachan:
This change addresses #3145, which highlights that when clicking a page anchor link the page scrolls too far down, i.e. below the heading or text where the anchor is located. 

It looks like this is because of the WP admin bar height because if I log out, the browser seems to scroll to the correct position (the very top of the anchor element).

I guess this is affecting all themes, so I've started with a fix in Blockbase, and then we can add the same fix where possible/necessary for all other themes. (Or would this go somewhere else if it's related to the admin bar?)

I've gone with a CSS fix to start with as I liked how simple it is (`scroll-margin-top` - [browser support](https://caniuse.com/?search=scroll-margin-top)). But we could also use something like padding and a negative margin, or JS to manipulate the scroll position. I've also targeted all elements with an ID in `.entry-content`, as I thought this could be a good way to target all possible anchor elements, but not sure if this will have any other side effects or will miss any use-cases.

Logged in:

| Before      | After |
| ----------- | ----------- |
| before      | after       |

Logged out (same before and after):

logged-out

#### Related issue(s):
Fixes #3145

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -113,6 +113,10 @@ h1.has-background, h2.has-background, h3.has-background, h4.has-background, h5.h
 	margin-top: auto;
 }
 
+.scroll-to-element, .entry-content > *[id] {
+	scroll-margin-top: var(--wp--custom--scroll--margin-top);
+}
+
 @media (min-width: 600px) {
 	body.admin-bar {
 		--wpadmin-bar--height: 32px;
@@ -121,6 +125,10 @@ h1.has-background, h2.has-background, h3.has-background, h4.has-background, h5.h
 
 body.admin-bar .wp-site-blocks {
 	min-height: calc( 100vh - var(--wpadmin-bar--height));
+}
+
+body.admin-bar .scroll-to-element, body.admin-bar .entry-content > *[id] {
+	scroll-margin-top: calc(var(--wp--custom--scroll--margin-top) + ( var(--wpadmin-bar--height) * 0.9 ));
 }
 
 ::selection {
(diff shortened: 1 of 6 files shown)


MaggieCabrera:
I love this! As we talked on Slack, I think this is better fit for a core patch, since that will affect all themes. While testing this I found a dotcom only issue related to this. When the marketing bar is present (regardless of if the admin bar is present or not, we have the same problem: 

Screenshot 2021-07-23 at 13 07 53

To enable this on your sandbox, edit the file on wp-content/blog-plugins/marketing-bar.php and edit the return on line 22:

```
if ( ! $this->should_show( $current_blog ) ) {
	//return;
}
```

That will enable the bar on every sandboxed site you have.

mikachan:
Just found out that there is already an existing issue being tracked in core: https://github.com/WordPress/wordpress-develop/pull/782 & https://core.trac.wordpress.org/ticket/46371

Looking at the marketing bar fix next.

scruffian:
Shall we close this?

PR #4302, opened 2021-07-23
Add: Theme creation checklist script

scruffian:
This is a script which generates the issues we need to complete when we start a new theme. The script can be run with

`node create-a-theme-checklist.js [themename]`

You will need a GitHub access token to run the script.


create-a-theme-checklist.js
```diff
@@ -0,0 +1,97 @@
...
+		'Templates: Page templates - Index',
+		'Templates: Page templates - Search',
+		'Templates: Page templates - 404',
+		'Templates: Page templates - Archive',
+		'Templates: Header template part (including mobile)',
+		'Templates: Footer template part (including mobile)',
+		'Block patterns (if needed)',
+		'Block styles (if needed)',
+		'Comment form styles (dotcom and dotorg) (if needed)',
+		'Navigation (if needed)',
```
MaggieCabrera, line 55: Not sure if we need an issue specific to navigation if that's included in the header.

MaggieCabrera:
There are a few items in the pre-launch section of the checklist that I think could be considered to be added as issues from the start:

- readme.txt
- screenshot.png
- Theme tags to style.css

Possibly even creating an issue for the headstart. What do you all think?

jffng:
I think readme, screenshot, and headstart are all worth adding either as individual issues or maybe sub-issues of one overview pre-launch issue.

MaggieCabrera:
I've added the pre-launch issues as separate issues but as part of the pre-launch section (editing the content of an issue is a bit more involved). Also added content layout to the spacing issue since it's kind of related and it's not anywhere else (I don't think it merits its own issue but still think it should be mentioned somewhere).

MaggieCabrera:
I'm merging this and we can iterate on it if we see fit

PR #4325, opened 2021-07-26
Skatepark: added a starting boilerplate 

### LINKED ISSUE #4307: Skatepark: Create base theme from boilerplate

MaggieCabrera:

This PR provides a starting point for Skatepark development. I have copied index.html from blockbase so that GB detects the theme as a block theme and it doesn't completely break. I also added skatepark to the list of BB children for the build script.

#### Related issue(s):

Closes https://github.com/Automattic/themes/issues/4307


skatepark/block-templates/index.html
```diff
@@ -0,0 +1,32 @@
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
```
jffng, line 1: Should we bring the header / footer parts in the same PR? Just asking because it errors currently with "template part not found"

skatepark/block-templates/index.html
```diff
@@ -0,0 +1,32 @@
...
+	<!-- /wp:group -->
+	<!-- /wp:post-template -->
+	<!-- wp:group {"layout":{"inherit":true}} -->
+		<div class="wp-block-group">
+		<!-- wp:query-pagination -->
+		<div class="wp-block-query-pagination"><!-- wp:query-pagination-previous {"label":"Previous Page"} /-->
+	
+			<!-- wp:query-pagination-numbers /-->
+	
+			<!-- wp:query-pagination-next {"label":"Next Page"} /--></div>
```
jffng: I wonder if we still need the label attributes here since these labels are the default text, and Ben's Gutenberg PR which removed the arrows: https://github.com/WordPress/gutenberg/pull/33626

jffng (commented):
Left two comments but I think it's fine to bring in to get started :)

skatepark/block-templates/index.html
MaggieCabrera: honestly, I was going to just have the post-template block and leave it like that, the index needs to be re-done completely and it has its own separate issue, I just didn't want to have it be empty at all.

skatepark/block-templates/index.html
MaggieCabrera: We definitely should do that for Blockbase

pbking: For blockbase: https://github.com/Automattic/themes/pull/4331

jffng (approved):
Here we go! 🚀

PR #4332, opened 2021-07-26
Blockbase: make 404 into a php template so it can be translated

jffng:

This makes it so Blockbase and its children's 404 template is php-based so it can be translated.  I

I could not just add a 404.php template to Quadrat. Blockbase's 404.html template would win in the template resolution logic. So I had to replace Blockbase's template with a php-based one to get this approach to work.

#### Related issue(s):

#3653


blockbase/theme.json
```diff
@@ -198,6 +198,9 @@
 					}
 				}
 			},
+			"layout": {
+				"contentSize": "620px"
```
jffng, line 202: We should probably also look into making it so the layout key generates a CSS variable, because this is redundant.

quadrat/block-templates/search.html
```diff
@@ -3,10 +3,6 @@
 <!-- wp:group {"layout":{"inherit":true}, "tagName":"main"} -->
 <main class="wp-block-group">
 
-	<!-- wp:heading -->
-	<h2>Results:</h2>
```
pbking, line 7: Should we do the same for blockbase?  That would make all 4 blockbase themes i11n-ready.

blockbase/theme.json
pbking: Couldn't find anything so opened this: https://github.com/WordPress/gutenberg/issues/33705

pbking (approved):
Other than removing `Results:` from blockbase as well I think this is good-to-go.  I'm glad it's only one (and a simple) page.

MaggieCabrera (approved):
This LGTM! should we create a Glotpress project for Blockbase? I think we only did so for Quadrat

mikachan:
> should we create a Glotpress project for Blockbase? I think we only did so for Quadrat

Yeah we've only created one for Quadrat so far, guessing we should do one for Blockbase too.

PR #4335, opened 2021-07-27
removed fse tag from varia children

MaggieCabrera:

This PR removes the FSE tag from themes that are not using the actual Site Editor but had it because of the legacy dotcom fse (now deprecated)

#### Related issue(s):

https://github.com/Automattic/wp-calypso/issues/54922

### DIFF
--- a/alves/sass/style-child-theme.scss
+++ b/alves/sass/style-child-theme.scss
@@ -10,7 +10,7 @@ License: GNU General Public License v2 or later
 License URI: LICENSE
 Template: varia
 Text Domain: alves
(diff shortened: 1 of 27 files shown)


Addison-Stavlo:
> legacy dotcom fse (now deprecated)

Note: while discontinued, it is still supported.  So one question here is if legacy FSE experience continues to work as expected after these tags are removed on dotcom.  Will removing the tag from the stylesheets prevent [this check](https://github.com/Automattic/wp-calypso/blob/442e5924c9de20f41fa726464811e674d14814ce/apps/editing-toolkit/editing-toolkit-plugin/dotcom-fse/helpers.php#L129-L137) from working as expected?  Or is that independent from the stylesheet at this point?

MaggieCabrera:
Solved via https://github.com/Automattic/themes/pull/4337

PR #4336, opened 2021-07-27
Fix/blockbase customizer meta info

### LINKED ISSUE #4197: Blockbase (+ Quadrat) color changes made via Customizer do not reflect on wpcom
## Steps to replicate

1. Go to the site editor on wpcom
2. Change some colors in the palette
3. Verify they show up in the front-end
4. Go to the customizer (/wp-admin/customize.php) and change some colors or select a palette other than the default. 
5. Save and refresh, or view the front-end

## Result

The colors return to the default palette

## Expected

The color changes from the Customizer to be reflected everywhere.

pbking:

It some situations the Theme User JSON doesn't ALREADY have the meta-information `version` and `isGlobalStylesUserThemeJSON`.  

In that situation when the Theme User JSON is parsed (without the meta-information) the resulting theme data returned from the call to `WP_Theme_JSON_Resolver_Gutenberg::get_merged_data()` results in the Theme User JSON being ignored.

This change ensures that the meta-information is always provided.

#### Testing

I'm unsure how to cause a site to enter into the original situation.  However the result of entering into the situation is that Theme User JSON is set either to `{}` or contains the custom color or font data (but not the meta data).  Using the FSE Global Styles panel once (even if just to "reset global styles to defaults") adds the meta-information to the JSON and stops the issue from happening on that site.

#### Related issue(s):

Fixes #4197

### DIFF
--- a/blockbase/inc/customizer/wp-customize-colors.php
+++ b/blockbase/inc/customizer/wp-customize-colors.php
@@ -133,6 +133,11 @@ class GlobalStylesColorCustomizer {
 		$user_theme_json_post         = get_post( $user_custom_post_type_id );
 		$user_theme_json_post_content = json_decode( $user_theme_json_post->post_content );
 
+		// Set meta settings.
+		$user_theme_json_post_content->version                     = 1;
+		$user_theme_json_post_content->isGlobalStylesUserThemeJSON = true;
+
+		// Set palette settings.
 		$user_theme_json_post_content = set_settings_array(
 			$user_theme_json_post_content,
 			array( 'settings', 'color', 'palette' ),
--- a/blockbase/inc/customizer/wp-customize-fonts.php
+++ b/blockbase/inc/customizer/wp-customize-fonts.php
@@ -415,6 +415,10 @@ class GlobalStylesFontsCustomizer {
 		$user_theme_json_post         = get_post( $user_custom_post_type_id );
 		$user_theme_json_post_content = json_decode( $user_theme_json_post->post_content );
 
+		// Set meta settings.
+		$user_theme_json_post_content->version                     = 1;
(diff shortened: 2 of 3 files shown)


jffng:
When I set the colors, it works like a charm!

When I change the fonts, all of the user customizations disappear and return to the defaults. Looking into this...

jffng:
I believe wpcom does not like it when we supply `fontFamilies` to `settings->typography`.  I verified this by commenting [these lines out](https://github.com/Automattic/themes/blob/trunk/blockbase/inc/customizer/wp-customize-fonts.php#L396-L400), and I'm able to preserve the changes to the user theme json.

I wonder if it's related to theme.json landing in 5.8, and the sanitization / allowed keys.

I was logging using: `error_log( print_f( $user_theme_json_post_content, true) )` and that prints to `/tmp/php-errors`.

jffng:
I did some more debugging and had a thought for a workaround. What if we stored `fontFamilies` in the custom section of the settings? What is the benefit of providing `fontFamilies` via settings.typography if they aren't used by global styles (yet)?

mreishus:
:+1: After applying this to my sandbox and sandboxing the IP for eskothemedemo, I was able to make color palette changes in the customizer that persisted upon F5. Before, that would not work.

mreishus:
Edit 3: Third edit, sorry! I'm not able to get font changes to persist on the starter site with this patch applied to my sandbox and the IP sandboxed.


https://user-images.githubusercontent.com/937354/127696715-c6897e1d-f008-41fd-aa54-fd0355d40b33.mp4

pbking:
@mreishus thanks for verifying.
The issue with fonts is tracked here: https://github.com/Automattic/wp-calypso/issues/55030.

scruffian:
This fixes one known issue with customizations. I think we should merge it to make it easier to test the other issues.

PR #4154, opened 2021-07-02
Universal themes: Add a setting to disable the Site Editor

scruffian:
#### Changes proposed in this Pull Request:
This adds a setting to the Gutenberg experiments page which allows users to disable the Site Editor for universal themes.

We talked about this being a separate plugin, but I think in some ways shipping it in the theme makes more sense.

Screenshot 2021-07-02 at 11 43 25

Screenshot 2021-07-02 at 11 43 14

PR #4340, opened 2021-07-28
Revert Blockbase's site editor toggle

mattwiebe:
Being reverted in favour of a plugin-based solution

See https://github.com/Automattic/wp-calypso/pull/54620

Reverts #4154


jffng:
I'm not against removing it entirely, but I think this toggle may still have value to non-wpcom users of Blockbase and its children.

What if we conditionally loaded the site editor toggle script for non-wpcom environments?

mattwiebe:
Hmm, interesting. @scruffian had sounded on-board for a full removal, but I won't speak for him. 

Let me reason through this a bit.

The tricky thing here is that WP 5.8 will [throw a fatal error](https://github.com/WordPress/WordPress/blob/327852393a5c65610b800080de77fcb33af34c40/wp-includes/theme.php#L911-L923) when activating a theme with the `full-site-editing` tag, telling them to install the Gutenberg plugin. And so the most likely course of action at this point from a user perspective is "gross, this theme is broken" and they move on. 

Or, being curious, they install the Gutenberg plugin to unlock this new Full Site Editing functionality that sounds interesting. They activate the theme but: nothing's different, because we've disabled the Site Editor by default in this theme, confounding the expectations that core provided.

Or, this is even someone already running the Gutenberg plugin, explicitly looking for `full-site-editing` themes, but they activate this one without errors, but also without seeing the Site Editor. Expectations foiled again.

Unless I'm missing a scenario, this means that the only type of user who wants this type of functionality in this way is the sysadmin of a WP Multisite network, who wants to be able to roll out a theme like this to ease the transition from classic WP themes to FSE themes. They would be very unlikely to want this code living in a specific theme, preferring it in a plugin that they can enable or disable at will, like will be the case once Automattic/wp-calypso#54620 lands. WP.com is just such a sysadmin-type of user, which is why this type of functionality is desirable for us. 

So, I have not seen a good use-case for keeping this functionality in the theme, and only good reasons not to. But maybe I'm missing something?

pbking:
I'm still on the fence, but one important aspect that the code being removed also does is RE-adds the links for the customizer.  There are still some things (like font family customization and palette collections) that can only be done in the Customizer and continued access to that is important.  At a minimum I believe that code should stay until a better solution is available.

Addison-Stavlo:
I think @mattwiebe makes some really good points above https://github.com/Automattic/themes/pull/4340#issuecomment-888634442.

Regarding @pbking's concern:
> one important aspect that the code being removed also does is RE-adds the links for the customizer.  There are still some things (like font family customization and palette collections) that can only be done in the Customizer and continued access to that is important. 

I assume we should be able to do this from the plugin approach as well?

pbking:
> I assume we should be able to do this from the plugin approach as well?

What I mean is that out of the box Blockbase (and children) should have the customizer enabled, no further action should be needed (in the .org context).  So that portion of the above removed code should remain (until things are better).  I'm not arguing that the functionality should be in the replacement plugin, but that it never leave the theme.

mattwiebe:
I did not know that Blockbase was still requiring the Customizer for fonts and colors, thanks for that additional context! It's very unfortunate from a UX perspective, but I get that the Site Editor does not yet have parity with all of our desired customization capabilities. 

I'll modify this shortly to simply re-add the Customizer link after Gutenberg removes it.

mattwiebe:
Ok folks, this now restores the Customizer link (or, more properly, prevents Gutenberg from unhooking it) while leaving the Site Editor alone.

blockbase/functions.php
```diff
@@ -95,12 +95,19 @@ function blockbase_fonts_url() {
...
 
+/**
+ * Restores the Customizer since we still rely on it.
+ */
+function blockbase_restore_customizer() {
+	remove_action( 'admin_menu', 'gutenberg_remove_legacy_pages' );
+}
+add_action( 'init', 'blockbase_restore_customizer' );
+// For WPcom REST API requests to work properly.
+add_action( 'restapi_theme_init', 'blockbase_restore_customizer', 11 );
```
jffng: Is it okay if this goes in `inc/wpcom.php` to consolidate the wpcom concerns?

jffng (commented):
Thanks for the reasoning @mattwiebe, I didn't realize that about 5.8 throwing an error for sites with the `full-site-editing` tag and no Gutenberg. I guess that ship has sailed. 

I tested and verified the customizer is still accessible. Other than my one comment, I'm on board with this PR.

pbking:
I moved the wpcom concerns to wpcom.php and testing locally and on sandbox.  
I think this is good-to-go now.

pbking (approved):
LGTM
🚢

mattwiebe:
Thanks for pushing this forward while I was AFK for a long weekend, @pbking 🙌

PR #4341, opened 2021-07-28
Quadrat: add footer + credits in php

jffng:

This PR is an attempt to unblock how footer credits are handled in Quadrat, by solving two issues:

1. Allows the string to be translated on dotorg
2. Allows the wpcom mechanism to add footer credits in that environment, background here: pcjTuq-ff-p2.

It also requires a small change for the wpcom environment: D64851-code 

#### Related issue(s):

This slack discussion and feedback: p1627461480480400/1627413124.455900-slack-C029FM1EH.

**To Test**
- Check out this PR locally
- Verify the footer and credits appear as expected
- Check out this PR onto your sandbox along with [the patch](D64851-code)
- Verify the wpcom-specific credits appear

It does not look great on wpcom, so we could add some styles that position it closer to the dotorg version cc @kjellr @beafialho 

Screen Shot 2021-07-28 at 4 14 45 PM

### DIFF
--- a/quadrat/functions.php
+++ b/quadrat/functions.php
@@ -51,6 +51,31 @@ function quadrat_scripts() {
 add_action( 'wp_enqueue_scripts', 'quadrat_scripts' );
 
 /**
+ * Add a filter to the render callback of the footer, 
+ * so it can be translated or customized for WordPress.com.
+ */
+function quadrat_add_markup_to_footer_template( $block_content, $block ) {
+	if ( $block['blockName'] === 'core/template-part' && $block['attrs']['slug'] === 'footer' ) {
+		// If we are on WPCOM, do not render a credit so that the WP.com footer credit plugin can handle it
+		if ( class_exists( 'WPCOM_Block_Theme_Footer_Credits') ){
+			return '<!-- wp:group {"className":"site-footer","style":{"spacing":{"padding":{"top":"150px","bottom":"150px"}}}} -->
+			<div class="wp-block-group site-footer" style="padding-top:150px;padding-bottom: 150px">
+			</div><!-- /wp:group -->';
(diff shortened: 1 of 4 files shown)


quadrat/functions.php
```diff
@@ -50,6 +50,27 @@ function quadrat_scripts() {
...
 
+/**
+ * Add a filter to the render callback of the footer, 
+ * so it can be translated or customized for WordPress.com.
+ */
+function quadrat_add_markup_to_footer_template( $block_content, $block ) {
+	if ( $block['blockName'] === 'core/template-part' && $block['attrs']['slug'] === 'footer' ) {
+		// If we are on WPCOM, do not render a credit so that the WP.com footer credit plugin can handle it
+		$wp_credit = '';
+		if ( ! class_exists( 'WPCOM_Block_Theme_Footer_Credits') ){
```
pbking: Since this is environment-specific does it make sense to put it into the wpcom-specific code that isn't included in the build?

pbking:
What do you think of the change into environment-specific files?
I wasn't able to get the wpcom version to do anything special with the markup (before or after the suggested refactor change).

jffng:
> What do you think of the change into environment-specific files?

It makes sense to me, thanks for pushing it!

pbking:
I suggest we close this PR and instead make the necessary changes for this functionality to be fully supported in the block theme credit plugin.

jffng:
I agree, this kind of workaround to support translation of a single string does seem worth the added overhead — filtering every single block render, checking if it's the footer, and changing the contents if so. If wpcom has different needs for displaying / augmenting the footer credit than what already exists, it can be addressed in that environment / plugin.

PR #4343, opened 2021-07-29
Skatepark: footer and pre footer block pattern

### LINKED ISSUE #4312: Skatepark: Templates: Footer template part (including mobile)
Desktop:

Screenshot 2021-07-23 at 16 19 50

Mobile:

Screenshot 2021-07-26 at 09 59 56

MaggieCabrera:

This PR includes the footer template part that only has the WP credits and a block pattern with the pre footer. Since the block pattern contains the jetpack subscription block it won't show on a local install. I tested this on my [atomic test site](https://atomictest365300111.wpcomstaging.com/). I didn't want to go much into styling the subscription form mainly because the form has a few settings to control its appearance and I was hesitant to override those using much CSS and have users not be able to use the settings on the form, @melchoyce I would like your input on this.

Screenshot 2021-07-29 at 15 09 46

#### Related issue(s):

Closes https://github.com/Automattic/themes/issues/4312

### DIFF
--- a/skatepark/inc/patterns/pre-footer.php
+++ b/skatepark/inc/patterns/pre-footer.php
@@ -17,7 +17,7 @@ return array(
 		<p class="has-small-font-size">' . esc_html__( 'Skate ipsum dolor sit amet, poseur nollie casper pop shove-it. Kickturn noseblunt tailslide.', 'skatepark' ) . '</p>
 		<!-- /wp:paragraph -->
 		
-		<!-- wp:social-links {"iconColor":"primary","iconColorValue":"#000000","className":"is-style-logos-only"} -->
+		<!-- wp:social-links {"iconColor":"primary","iconColorValue":"var(--wp--custom--color--foreground)","className":"is-style-logos-only"} -->
 		<ul class="wp-block-social-links has-icon-color is-style-logos-only"><!-- wp:social-link {"url":"http://twitter.com","service":"twitter"} /-->
 		
 		<!-- wp:social-link {"url":"http://facebook.com","service":"facebook"} /-->
--- a/skatepark/inc/patterns/pre-footer.php
+++ b/skatepark/inc/patterns/pre-footer.php
@@ -31,14 +31,12 @@ return array(
 		<h3 class="has-tiny-font-size"><strong>' . esc_html__( 'More info', 'skatepark' ) . '</strong></h3>
 		<!-- /wp:heading -->
 		
-		<!-- wp:navigation {"orientation":"vertical","__unstableLocation":"secondary","style":{"typography":{"textDecoration":"underline"}},"fontSize":"small"} /--></div>
+		<!-- wp:navigation {"orientation":"vertical","__unstableLocation":"primary","style":{"typography":{"textDecoration":"underline"}},"fontSize":"small"} /--></div>
 		<!-- /wp:column -->
 
 		<!-- wp:column -->
 		<div class="wp-block-column"><!-- wp:heading {"level":3,"fontSize":"tiny"} -->
-		<h3 class="has-tiny-font-size"><strong>' . esc_html__( 'Search', 'skatepark' ) . '</strong></h3><!-- /wp:heading --><!-- wp:heading {"level":3,"fontSize":"tiny"} -->
-		<h3 class="has-tiny-font-size"><strong>Search</strong></h3>
-		<!-- /wp:heading --></div>
+		<h3 class="has-tiny-font-size"><strong>' . esc_html__( 'Search', 'skatepark' ) . '</strong></h3><!-- /wp:heading --><!-- wp:search {"label":"Search","showLabel":false,"buttonText":"Search","buttonPosition":"button-inside"} /--></div>
 		<!-- /wp:column --></div>
 		<!-- /wp:columns --></div>
 		<!-- /wp:group -->',
--- a/skatepark/inc/patterns/pre-footer.php
+++ b/skatepark/inc/patterns/pre-footer.php
@@ -1,0 +1,45 @@
+<?php
+/**
+ * Pre Footer.
+ *
+ * @package Skatepark
+ */
+
+return array(
+	'title'      => __( 'Footer', 'skatepark' ),
+	'categories' => array( 'skatepark' ),
+	'content'    => '<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"60px","bottom":"60px"}}},"className":"pre-footer"} -->
+	<div class="wp-block-group alignwide pre-footer" style="padding-top:60px;padding-bottom:60px"><!-- wp:columns -->
+		<div class="wp-block-columns"><!-- wp:column -->
+		<div class="wp-block-column"><!-- wp:site-title {"style":{"typography":{"textTransform":"uppercase"}},"fontSize":"large"} /-->
+		
+		<!-- wp:paragraph {"fontSize":"small"} -->
+		<p class="has-small-font-size">' . esc_html__( 'Skate ipsum dolor sit amet, poseur nollie casper pop shove-it. Kickturn noseblunt tailslide.', 'skatepark' ) . '</p>
+		<!-- /wp:paragraph -->
+		
+		<!-- wp:social-links {"iconColor":"primary","iconColorValue":"#000000","className":"is-style-logos-only"} -->
+		<ul class="wp-block-social-links has-icon-color is-style-logos-only"><!-- wp:social-link {"url":"http://twitter.com","service":"twitter"} /-->
+		
+		<!-- wp:social-link {"url":"http://facebook.com","service":"facebook"} /-->
+		
+		<!-- wp:social-link {"url":"http://instagram.com","service":"instagram"} /--></ul>
+		<!-- /wp:social-links --></div>
+		<!-- /wp:column -->
+		
+		<!-- wp:column -->
+		<div class="wp-block-column"><!-- wp:heading {"level":3,"fontSize":"tiny"} -->
+		<h3 class="has-tiny-font-size"><strong>' . esc_html__( 'More info', 'skatepark' ) . '</strong></h3>
+		<!-- /wp:heading -->
+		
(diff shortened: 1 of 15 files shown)


skatepark/inc/patterns/pre-footer.php
```diff
@@ -0,0 +1,45 @@
...
+	'content'    => '<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"60px","bottom":"60px"}}},"className":"pre-footer"} -->
+	<div class="wp-block-group alignwide pre-footer" style="padding-top:60px;padding-bottom:60px"><!-- wp:columns -->
+		<div class="wp-block-columns"><!-- wp:column -->
+		<div class="wp-block-column"><!-- wp:site-title {"style":{"typography":{"textTransform":"uppercase"}},"fontSize":"large"} /-->
+		
+		<!-- wp:paragraph {"fontSize":"small"} -->
+		<p class="has-small-font-size">' . esc_html__( 'Skate ipsum dolor sit amet, poseur nollie casper pop shove-it. Kickturn noseblunt tailslide.', 'skatepark' ) . '</p>
+		<!-- /wp:paragraph -->
+		
+		<!-- wp:social-links {"iconColor":"primary","iconColorValue":"#000000","className":"is-style-logos-only"} -->
```
jffng: Can we use the semantic color variable here in patterns? `var(--wp--custom--color--foreground)`

jffng (commented):
Some small design notes, wasn't sure if it makes sense to address in this PR:

- The site title link is underlined, and in the comps it is not
- It looks like the nav link treatment has a thicker bottom border and more space between the line and type. Can we achieve that with text-decoration, or do we need to use borders if we want finer control there?
- The social links are not visually aligned to text above them because of the padding. Not sure if we want to do anything about it or say it's fine.

skatepark/inc/patterns/pre-footer.php
MaggieCabrera: I thought this would break validation but it works! can't be achieved through the editor but it works from the pattern and it makes a lot of sense, thanks for pointing it out!

MaggieCabrera:
> Some small design notes, wasn't sure if it makes sense to address in this PR:
> 
>     * The site title link is underlined, and in the comps it is not
> 
>     * It looks like the nav link treatment has a thicker bottom border and more space between the line and type. Can we achieve that with text-decoration, or do we need to use borders if we want finer control there?
> 
>     * The social links are not visually aligned to text above them because of the padding. Not sure if we want to do anything about it or say it's fine.

Thank you for this:

About the underlines, I think we should go over them for all the site, not just this block, so I opened an issue about it [here](https://github.com/Automattic/themes/issues/4346). I added a negative margin when the block is using the "only icons" block style and I think that's mostly fine to use for this pattern.

MaggieCabrera:
Nevermind the last comment, apparently [text-underline-offset](https://developer.mozilla.org/en-US/docs/Web/CSS/text-underline-offset) exists! So we can use that instead of the border, which I think is a much better option.

melchoyce:
Let's do a search box instead of a subscribe one!

Will review this PR today 👍

MaggieCabrera:
I updated this to use the search block instead of the subscribe one and it can be tested on a local site but the search block looks terrible (we haven't done any button styling yet and that's part of it too):

Screenshot 2021-08-03 at 18 23 05

I don't think there's design for this block in figma, or is there?

melchoyce:
There's just this:

image

Would it help if I designed all the variations in Figma?

MaggieCabrera:
> Would it help if I designed all the variations in Figma?

Yeah, I think that would help, since that design differs a bit from the default

I have another question regarding this block pattern, should the links on more info target somewhere or be empty links? 
- If they are part of a menu they will be empty for users until they add items to the secondary menu via customizer or editing the block (this is the current behavior, we define a secondary navigation area for this specific block pattern). This means we should create content for those pages on the demo site. If by chance their sites already had a secondary menu, that will show up on this block but that's not a given.
- If they are empty links the user can edit them freely to use their own pages but they already show some content from the start and we don't need to add any content to the demo site for these pages.

melchoyce:
> I have another question regarding this block pattern, should the links on more info target somewhere or be empty links?

Just talked this over with @kjellr, and we agreed it would work best to reuse the primary menu here instead of a secondary menu or empty links. This way we don't have to create additional demo pages, and it'll be easier for folks to set up than using blank links.

melchoyce:
Put together some ideas for search block styling in Figma: https://www.figma.com/file/CA8Gb16mt9yYNWOiVLAWak/Skatepark?node-id=1950%3A2101

Would y'all prefer I make a new issue for that?

MaggieCabrera:
@melchoyce I'm just going to do the search block changes in this PR, it's easier for me that way.

So I found out that Blockbase didn't have the best defaults for the search block styles and I improved it a bit so that we can better style Skatepark.

Before | After
--- | ---
Screenshot 2021-08-05 at 10 17 32 | Screenshot 2021-08-05 at 10 15 44

The content for testing:
```

<!-- wp:search {"label":"Search","placeholder":"Placeholder text...","buttonText":"Search"} /-->

<!-- wp:search {"label":"Search","showLabel":false,"placeholder":"Placeholder text...","buttonText":"Search"} /-->

<!-- wp:search {"label":"Search","placeholder":"Placeholder text...","buttonText":"Search","buttonPosition":"button-inside"} /-->

<!-- wp:search {"label":"Search","showLabel":false,"placeholder":"Placeholder text...","buttonText":"Search","buttonPosition":"button-inside"} /-->

<!-- wp:search {"label":"Search","placeholder":"Placeholder text...","buttonText":"Search","buttonUseIcon":true} /-->

<!-- wp:search {"label":"Search","showLabel":false,"placeholder":"Placeholder text...","buttonText":"Search","buttonUseIcon":true} /-->

<!-- wp:search {"label":"Search","placeholder":"Placeholder text...","buttonText":"Search","buttonPosition":"button-inside","buttonUseIcon":true} /-->

<!-- wp:search {"label":"Search","showLabel":false,"placeholder":"Placeholder text...","buttonText":"Search","buttonPosition":"button-inside","buttonUseIcon":true} /-->
```

Now I'm going to add this to the footer pattern and do the rest of the changes

MaggieCabrera:
This is ready for another review:

Screenshot 2021-08-05 at 11 41 28

melchoyce:
With the search boxes, the only thing I noticed is that the no-button version is shorter than the others, probably because there's no extra padding from the button:

image

Not the end of the world; if it's an easy fix, great, if it's a pain, it's probably fine as-is.

Speaking of height... can the footer search box be a little shorter? I like the idea of the footer being a little more condensed, and the full-height search box feels chunky in there. My mockups have it about 10px shorter than search boxes within the content area, and the label is 14px instead of 16px.

MaggieCabrera:
This is ready for another review with the change to the height of the search form on the footer and the fix for the block when it doesn't have a button.

Screenshot 2021-08-09 at 15 53 26

scruffian:
Do you think that the pre-footer should be included in the footer now that it doesn't have a subscription block in it?

The pre-footer goes full width by default, so we should probably add some alignment rules.

MaggieCabrera:
> Do you think that the pre-footer should be included in the footer now that it doesn't have a subscription block in it?

well, that's a design decision, in figma it shows as a block pattern and that's why I didn't add it. The way I see it, it makes total sense but it would duplicate code inside the theme. If we had a block pattern block it would be beautiful though.

skatepark/sass/blocks/_buttons.scss
```diff
@@ -21,20 +21,23 @@
...
 .wp-block-file .wp-block-file__button {
 	text-transform: uppercase;
+	letter-spacing: 0.1em;
 	@include skatepark_button_hover_styles;
 }	
 
 .wp-block-search {
 	&.wp-block-search__button-inside .wp-block-search__inside-wrapper .wp-block-search__button,
-	.wp-block-search__button {
+	.wp-block-search__button.wp-block-search__button {
```
scruffian, line 30: Why do we need the double selector?

skatepark/sass/elements/_links.scss
```diff
@@ -42,6 +42,7 @@ h6 a,
 }
 
 // Select the focus states of all non-wpadmin and screen reader links
+.wp-block-search__button,
```
scruffian, line 45: Should we add this to the corresponding file in blockbase? I wonder if we could make this a mixin to reduce the duplicate code.

skatepark/sass/blocks/_buttons.scss
MaggieCabrera: [Because of GB](https://github.com/WordPress/gutenberg/blob/d011a7874d82cfa740cb208b42582d824445ced0/packages/block-library/src/search/editor.scss)! I think the block shouldn't be using the components classes, I'll make a PR for it

kjellr:
Let's leave it out of the footer for now — otherwise I worry that it'll be tough for non-FSE users to have a minimal page footer. We can always reconsider later. 

One last thing I realized is that we'll need to swap out all the skatepark-ipsum text in our mockups for something more real. For this pattern, let's just replace this part: 

> Skate ipsum dolor sit amet, poseur nollie casper pop shove-it. Kickturn noseblunt tailslide.

.... with this: 

> Skatepark's coaches will work with you to develop and improve your skating abilities. 

Other than that, this looks good to go!

scruffian (approved):

skatepark/sass/elements/_links.scss
MaggieCabrera: from what I see, blockbase doesn't handle outlines. I think that's one of those things that the editor still doesn't do but will in the future (handle hover/focused states) so it might be interesting to add. Do you think we should include it on BB theme.json or just the styles?

scruffian: Ah no, you're right, Blockbase has code to target links in the content. I think this is fine for now. If we want to do this in another theme we can move it to Blockbase.

MaggieCabrera:
Other reason why we shouldn't add this in the footer template part: i18n. There are quite a few strings that need translation here, so I agree that we better get this as a block pattern for now. I'll make the changes and bring this in, thank you all for your feedback!

PR #4350, opened 2021-08-02
Blockbase: check that alternative colors are defined in the default palette

MaggieCabrera:
Blockbase provides 3 5-color alternative palettes that are "inherited" (via the json build tool) by child themes that don't define their own alternative palettes. If the child theme doesn't define 5 colors in their son file, this causes the customizer to break (we get a JS error and the colors are not applied correctly). This PR only allows the customizer script to add the colors that are already defined in the default palette.


mikachan (commented):
This looks fab to me! Can confirm it's definitely fixed the issue I was seeing earlier with switching colours via the Customizer in Quadrat and Skatepark.

pbking (approved):
Looks good to me too.
🚢

PR #4353, opened 2021-08-02
Skatepark: Add alternative color palettes

### LINKED ISSUE #4315: Skatepark: theme.json: Alternative color palettes
Add the alternative color palettes to theme.json for Skatepark:

Screenshot 2021-07-27 at 13 03 01

mikachan:

This change adds the alternative color palettes from Skatepark's design.

I've added them to `child-theme.json` and each has a primary color and background color. 

Here's what they look like in the Customizer:

image

#### Related issue(s):

Closes #4315.

### DIFF
--- a/blockbase/inc/customizer/wp-customize-color-palette-control.css
+++ b/blockbase/inc/customizer/wp-customize-color-palette-control.css
@@ -37,9 +37,8 @@
 	bottom: 5px;
 	right: 0;
 	left: 0;
-	text-indent: 5px;
-	line-height: 1.5;
-	padding-bottom: 1px;
+	line-height: 1.2;
+	padding: 2px 0 3px 5px;
 }
 
 .color-stripe {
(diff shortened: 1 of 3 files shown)


melchoyce:
Ooh fun, will play around.

In the meantime, maybe we can drop "duotone" from the color palette names, since it doesn't look like the customizer handles having two lines of text well.

MaggieCabrera:
> In the meantime, maybe we can drop "duotone" from the color palette names, since it doesn't look like the customizer handles having two lines of text well.

Can definitely do that but this particular section has its own css on blockbase and can be tweaked by us easily (it's not using a dotcom plugin). We can try and improve it or change the name as you suggested

MaggieCabrera:
I renamed the palettes and improved the customizer CSS in case someone wants to use a longer name for the palettes at any  point (in fact, translations may cause the line break anyway) and I'm bringing this in

MaggieCabrera (approved):

PR #4354, opened 2021-08-02
Skatepark: add the header

### LINKED ISSUE #4313: Skatepark: Templates:Header template part (including mobile)
**Desktop:**
Screenshot 2021-07-23 at 16 19 41 

**Mobile:**
Screenshot 2021-07-23 at 16 20 25

### LINKED ISSUE #4319: Skatepark: Navigation (if needed)

jffng:

This PR adds a header template part and styles per the design for Skatepark. It applies some ideas from the every-layout.dev book. It was pretty interesting to explore and grok the concept of truly flexible layouts not tied to breakpoints. 

Looking at Skatepark's header design and the components involved, there are three states to consider:

A. Not wrapped | B. Wrapped | C. Wrapped w/ responsive navigation 
------ | ------ | ------
Screen Shot 2021-08-03 at 12 32 10 PM | Screen Shot 2021-08-03 at 12 32 03 PM | Screen Shot 2021-08-03 at 12 31 54 PM

B is the interesting scenario. Before, I would have tried to write a media query that tells the header to break onto two lines at a specific pixel value. But this is not a great approach because we don't know for certain the content width / number of menu items, thus making it impossible to write a media query that solves for all the situations. 

I could not get around using a media query for the navigation + social links, which I think is acceptable because the navigation block's responsive behavior is tied to the same media query.

The spacing needs some tightening to match the comps, but I wanted to get your feedback in the meantime. 

#### Related issue(s):

#4313

### DIFF
--- a/skatepark/block-template-parts/header.html
+++ b/skatepark/block-template-parts/header.html
@@ -1,10 +1,9 @@
 <!-- wp:group -->
 <div class="wp-block-group"><!-- wp:site-title {"style":{"typography":{"textTransform":"uppercase","fontStyle":"normal","fontWeight":"900"}},"fontSize":"large"} /-->
-
 <!-- wp:site-tagline {"fontSize":"small"} /--></div>
 <!-- /wp:group -->
 
-<!-- wp:group -->
+<!-- wp:group {"className":"nav-links"} -->
 <div class="wp-block-group nav-links"><!-- wp:navigation {"orientation":"horizontal","itemsJustification":"right","isResponsive":true,"style":{"typography":{"fontStyle":"normal","fontWeight":"900","textTransform":"uppercase"}},"fontSize":"small"} -->
 <!-- wp:navigation-link {"label":"Home","type":"custom","url":"http://skateparkdev.local/","kind":"custom","isTopLevelLink":true} /-->
 
@@ -15,8 +14,8 @@
 <!-- wp:navigation-link {"label":"Contact","type":"page","id":14,"url":"http://skateparkdev.local/contact/","kind":"post-type","isTopLevelLink":true} /-->
 <!-- /wp:navigation -->
 
-<!-- wp:social-links {"iconColor":"primary","iconColorValue":"#000000","iconBackgroundColor":"background","iconBackgroundColorValue":"#BFF5A5","className":"items-justified-right"} -->
-<ul class="wp-block-social-links has-icon-color has-icon-background-color items-justified-right"><!-- wp:social-link {"url":"twitter.com","service":"twitter"} /-->
+<!-- wp:social-links {"iconColor":"primary","iconColorValue":"#000000","className":"items-justified-right is-style-logos-only"} -->
+<ul class="wp-block-social-links has-icon-color items-justified-right is-style-logos-only"><!-- wp:social-link {"url":"twitter.com","service":"twitter"} /-->
 
 <!-- wp:social-link {"url":"facebook.com","service":"facebook"} /-->
 
--- a/skatepark/block-template-parts/header.html
+++ b/skatepark/block-template-parts/header.html
@@ -4,7 +4,7 @@
 <div class="wp-block-group">
 <!-- wp:site-logo /-->
 <!-- wp:site-title {"style":{"typography":{"textTransform":"uppercase","fontStyle":"normal","fontWeight":"900"}},"fontSize":"large"} /-->
-<!-- wp:site-tagline {"fontSize":"small"} /--></div>
+<!-- wp:site-tagline {"fontSize":"small","style":{"typography":{"fontWeight":"500"}}} /--></div>
 <!-- /wp:group -->
 
 <!-- wp:group {"className":"nav-links"} -->
--- a/skatepark/block-template-parts/header.html
+++ b/skatepark/block-template-parts/header.html
@@ -1,13 +1,11 @@
-<!-- wp:group {"align":"wide","layout":{"inherit":true}} -->
-<div class="wp-block-group alignwide"><!-- wp:columns {"align":"wide"} -->
-	<div class="wp-block-columns alignwide"><!-- wp:column -->
-	<div class="wp-block-column"><!-- wp:site-title {"style":{"typography":{"textTransform":"uppercase","fontStyle":"normal","fontWeight":"900"}},"fontSize":"large"} /-->
+<!-- wp:group -->
+	<div class="wp-block-group"><!-- wp:site-title {"style":{"typography":{"textTransform":"uppercase","fontStyle":"normal","fontWeight":"900"}},"fontSize":"large"} /-->
 	
 	<!-- wp:site-tagline {"fontSize":"small"} /--></div>
-	<!-- /wp:column -->
+	<!-- /wp:group -->
 	
-	<!-- wp:column -->
-	<div class="wp-block-column"><!-- wp:navigation {"orientation":"horizontal","itemsJustification":"right","isResponsive":true,"style":{"typography":{"fontStyle":"normal","fontWeight":"900","textTransform":"uppercase"}},"fontSize":"small"} -->
+	<!-- wp:group -->
+	<div class="wp-block-group nav-links"><!-- wp:navigation {"orientation":"horizontal","itemsJustification":"right","isResponsive":true,"style":{"typography":{"fontStyle":"normal","fontWeight":"900","textTransform":"uppercase"}},"fontSize":"small"} -->
 	<!-- wp:navigation-link {"label":"Home","type":"custom","url":"http://skateparkdev.local/","kind":"custom","isTopLevelLink":true} /-->
 	
 	<!-- wp:navigation-link {"label":"About","type":"page","id":10,"url":"http://skateparkdev.local/about/","kind":"post-type","isTopLevelLink":true} /-->
@@ -24,6 +22,4 @@
 	
 	<!-- wp:social-link {"url":"instagram.com","service":"instagram"} /--></ul>
 	<!-- /wp:social-links --></div>
-	<!-- /wp:column --></div>
-	<!-- /wp:columns --></div>
 	<!-- /wp:group -->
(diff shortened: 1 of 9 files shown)


skatepark/block-template-parts/header.html
```diff
@@ -0,0 +1,25 @@
+<!-- wp:group -->
+<div class="wp-block-group"><!-- wp:site-title {"style":{"typography":{"textTransform":"uppercase","fontStyle":"normal","fontWeight":"900"}},"fontSize":"large"} /-->
+
+<!-- wp:site-tagline {"fontSize":"small"} /--></div>
+<!-- /wp:group -->
+
+<!-- wp:group -->
+<div class="wp-block-group nav-links"><!-- wp:navigation {"orientation":"horizontal","itemsJustification":"right","isResponsive":true,"style":{"typography":{"fontStyle":"normal","fontWeight":"900","textTransform":"uppercase"}},"fontSize":"small"} -->
+<!-- wp:navigation-link {"label":"Home","type":"custom","url":"http://skateparkdev.local/","kind":"custom","isTopLevelLink":true} /-->
```
jffng: Not suggesting we ship the theme with these links baked in, but I included them so it's maybe easier to test.

skatepark/sass/templates/_header.scss
```diff
@@ -0,0 +1,33 @@
+header {
+	justify-content: space-between; // Apply a cluster (flex?) layout
+	flex-wrap: wrap-reverse;
```
MaggieCabrera, line 4: TIL!

skatepark/sass/templates/_header.scss
```diff
@@ -0,0 +1,33 @@
...
+		display: flex;
+		justify-content: space-between;
+		width: 100%;
+		border-bottom: 2px solid var(--wp--custom--color--primary);
+		.wp-block-navigation__responsive-container-open { 
+			display: flex;
+			align-items: center;
+
+			&::after {
+				content: "Menu";
```
MaggieCabrera: We can't translate this :(

MaggieCabrera:
I love how this looks! And it's very simple, great job!

This is what I see:

Screenshot 2021-08-04 at 10 09 47

The social icons could have the same treatment as the footer block so that they align with the menu. In this particular case, I'd add a class to the block, since the users may want to re build the header or use another theme's and we don't want the css to break that.

This is what the footer is doing:
```
	.wp-block-social-links.is-style-logos-only {
		margin-left: calc( -1 * ( 8px + 0.25em ) );
	}
```

If either the navigation or the social links are not present, this happens. The scenario with no social links is worth considering.

Screenshot 2021-08-04 at 10 13 22

Screenshot 2021-08-04 at 10 15 13


And right now Skatepark doesn't support a site logo, should we add support for it, @melchoyce If so, should we copy Quadrat's behavior for it?

skatepark/sass/templates/_header.scss
```diff
@@ -0,0 +1,33 @@
+header {
+	justify-content: space-between; // Apply a cluster (flex?) layout
+	flex-wrap: wrap-reverse;
+	> * { 
+		flex-grow: 1;
+		> * { // Apply a stack layout (page 69 of the every-layout.dev PDF) 
+			margin-top: 20px;
+			margin-bottom: 20px;
+		}
```
mikachan, line 10: This is cool, I love the idea of styling the context rather than the element itself.

skatepark/sass/templates/_header.scss
mikachan: Could we use the lang attribute on the `html` tag to translate this? Something like:

```
html[lang="en-US"] {
	.nav-links {
		.wp-block-navigation__responsive-container-open {
			&::after {
				content: "Menu";
			}
		}
	}
}
```

I guess this would need to be added via a PHP template in an inline style.. Or is this horrible and hacky?

mikachan (commented):
This looks ace! Agree with @MaggieCabrera's comments. Looking good in all browsers and resolutions. I agree with needing to use the media query for the navigation and social links, I can't think of another way we could do this.

skatepark/sass/templates/_header.scss
MaggieCabrera: we could add this line of CSS as inline inside functions.php and wrap the string around the PHP translation function directly, no need to target the lang attribute. It's not much better, but it's another option

mikachan: Ah nice ok! Well, we have options 😄

melchoyce:
Whoo 🎉 

Minor notes:
- The last icon in the social block has a right margin. Can we remove that margin for the last item in the menu?
- The tagline's font-weight should be medium
- Navigation items are missing some letter spacing (`0.1em`)

> And right now Skatepark doesn't support a site logo, should we add support for it, @melchoyce If so, should we copy Quadrat's behavior for it?

Yes and yes, though I think I'll want to test once we have Quadrat's behavior in Skatepark.

skatepark/sass/templates/_header.scss
jffng: Ideally there is an option in the nav block itself to provide this label next to the responsive hamburger. I wonder if it's planned

skatepark/functions.php
```diff
@@ -40,5 +40,13 @@ function skatepark_support() {
 function skatepark_scripts() {
 	// Enqueue front-end styles.
 	wp_enqueue_style( 'skatepark-styles', get_stylesheet_directory_uri() . '/assets/theme.css', array( 'blockbase-ponyfill' ), wp_get_theme()->get( 'Version' ) );
+
+	// Allow the responsive menu label to be translated.
+	$menu_label_css = '
```
jffng: This feels very messy, which makes me think it should go in the block.

In the meantime, maybe it should live in blockbase.

skatepark/theme.json
```diff
@@ -381,7 +381,8 @@
 			},
 			"core/navigation": {
 				"typography": {
-					"fontSize": "var(--wp--preset--font-size--normal)"
+					"fontSize": "var(--wp--preset--font-size--normal)",
+					"letterSpacing": "0.05em"
```
jffng, line 384: @melchoyce I added the letter spacing for all navigation blocks — is that okay, or do you think it should be targeted to the navigation block in the header only?

jffng:
Thanks for the reviews and comments!

I made the following changes:

- Removed right margin from social links
- Changed the flex alignment to handle the case when either no menu or social links are present
- Made it so the menu label can be translated
- Added a placeholder for the site logo in the template. 

I think this is ready for another review, be sure to test out adding a logo.

skatepark/theme.json
jffng: Also 0.1em for letter spacing looked like a lot compared to the comps, so I used 0.05 but let me know if you think it should be 0.1em.

skatepark/block-template-parts/header.html
```diff
@@ -0,0 +1,26 @@
...
+
+<!-- wp:navigation-link {"label":"About","type":"page","id":10,"url":"http://skateparkdev.local/about/","kind":"post-type","isTopLevelLink":true} /-->
+
+<!-- wp:navigation-link {"label":"FAQs","type":"page","id":12,"url":"http://skateparkdev.local/faqs/","kind":"post-type","isTopLevelLink":true} /-->
+
+<!-- wp:navigation-link {"label":"Contact","type":"page","id":14,"url":"http://skateparkdev.local/contact/","kind":"post-type","isTopLevelLink":true} /-->
+<!-- /wp:navigation -->
+
+<!-- wp:social-links {"iconColor":"primary","iconColorValue":"#000000","className":"items-justified-right is-style-logos-only"} -->
```
MaggieCabrera: we could use the primary value here so that it changes when we change palettes

jffng: Oooo good idea.

skatepark/functions.php
MaggieCabrera: but blockbase doesn't have the Menu text. I think this is probably [planned for the navigation block](https://github.com/WordPress/gutenberg/issues/22824#issuecomment-789584886)

MaggieCabrera:
The logo has a few differences comparing to Quadrat:

Screenshot 2021-08-05 at 18 14 16
Screenshot 2021-08-05 at 18 14 07
Screenshot 2021-08-05 at 18 13 07
Screenshot 2021-08-05 at 18 13 00

melchoyce:
Can we swap out the menu icon for a slightly thicker, three-bar version? 

image

And if possible, open the menu underneath instead of overlaying the header:

image

If that's not possible, we can keep as-is — just feels especially annoying to open the menu on the left side and then close it on the right. Also happy with us playing with this later in a new PR.

jffng:
> The logo has a few differences comparing to Quadrat:

@melchoyce any guidance on the position of the logo across viewports? The header on Quadrat is different so I wasn't sure how to handle. 

> And if possible, open the menu underneath instead of overlaying the header:

I think this will be tricky / I'm not sure if it's possible with the nav block currently. We can at least move the position of the close button within the modal so it's the same location as the open button.

melchoyce:
Would it work to display side-by-side on desktop, and stack on mobile?

image

MaggieCabrera:
Just adding that this also closes #4319 since I don't think we need a separate issue for that

skatepark/functions.php
jffng: Ok, I'm going to remove it from the theme for now and open an issue for it in Gutenberg.

jffng:
> Would it work to display side-by-side on desktop, and stack on mobile?

I gave this a try. Unless there is a purely flex based approach that I am missing, I don't think it's worth it because it would require another group block and more media-query based CSS, which is very fragile when we open things to the site editor.
 
> The tagline's font-weight should be medium

I opened a small change in Gutenberg to allow this so we don't have to add CSS for it and the user can customize it: https://github.com/WordPress/gutenberg/pull/33983

I also removed the responsive menu label in favor of adding the ability to the nav block itself: https://github.com/WordPress/gutenberg/issues/33985

Thoughts on bringing this in and iterating?

mikachan:
This is looking good for me in all browsers and resolutions. I've tested with and without both a logo and a menu.

Bringing this in and iterating sounds good to me.

MaggieCabrera (approved):
+1 on bringing this in and iterating. And I also agree that the menu label is best handled on the editor.

PR #4360, opened 2021-08-03
Seedlet blocks: fix broken footer

MaggieCabrera:
The footer for Seedlet blocks was broken. This solution leverages [flex layout on the group block](https://github.com/WordPress/gutenberg/issues/33687).

Before:

Screenshot 2021-08-03 at 17 12 07

After:

Screenshot 2021-08-03 at 17 09 38

### DIFF
--- a/seedlet-blocks/sass/theme.scss
+++ b/seedlet-blocks/sass/theme.scss
@@ -31,9 +31,14 @@
 }
 
 .footer-credit {
-	display: flex;
 	font-family: var(--wp--preset--font-family--headings);
 	white-space: pre-wrap;
+	&[class*="wp-container-"] {
+		margin-left: auto !important;
+		margin-right: auto !important;
+		padding: 0;
+		column-gap: 0;
+	}
 }
 
 .footer-credit > * {
--- a/seedlet-blocks/assets/theme.css
+++ b/seedlet-blocks/assets/theme.css
@@ -368,9 +368,15 @@ is passed all of the block attributes on the block definition in the template. *
 }
 
 .footer-credit {
-	display: flex;
 	font-family: var(--wp--preset--font-family--headings);
 	white-space: pre-wrap;
+}
+
+.footer-credit[class*="wp-container-"] {
+	margin-left: auto !important;
+	margin-right: auto !important;
+	padding: 0;
+	column-gap: 0;
 }
 
 .footer-credit > * {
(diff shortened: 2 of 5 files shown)


seedlet-blocks/sass/theme.scss
```diff
@@ -31,9 +31,14 @@
 }
 
 .footer-credit {
-	display: flex;
 	font-family: var(--wp--preset--font-family--headings);
 	white-space: pre-wrap;
+	&[class*="wp-container-"] {
+		margin-left: auto !important;
```
jffng, line 37: It's unfortunate that we are fighting our own alignment and padding rules here.

seedlet-blocks/sass/theme.scss
```diff
@@ -31,9 +31,14 @@
...
+		margin-right: auto !important;
+		padding: 0;
+		column-gap: 0;
```
jffng, line 40: I would be nice to be able to specify this gap value from theme.json, because it's a bit arbitrary right now (0.5em for any flex layouts).

jffng (commented):
This fixes the issue in my testing, thanks! I have one comment but I am not sure it's worth holding up the fix because it may require a deeper rethinking of alignments.

seedlet-blocks/sass/theme.scss
MaggieCabrera: yeah, GB should provide a setting for this

MaggieCabrera: Apparently it's planned: https://github.com/WordPress/gutenberg/pull/33359

seedlet-blocks/sass/theme.scss
MaggieCabrera: yeah, I thought the same, and I don't think it's the first time I've butted heads with this particular code. I'll have to think about it.

MaggieCabrera:
I'm holding on to this one because the gap control looks like it may make it soon and I still want to check if we can improve the Blockbase rules that we are overriding here

pbking:
I think this is the change introducing gap that you are talking about: https://github.com/WordPress/gutenberg/pull/33812

Linking so I don't have to go hunting for it again. :P

MaggieCabrera:
I kind of thought that it was going to make it sooner, we could deploy this as is and then iterate, right now this PR adds the gap via CSS, we can always come back to remove that and add the new markup

PR #4362, opened 2021-08-03
Independent Publisher 2: Make color annotations button rule more specific

### LINKED ISSUE #4330: Atomic: Custom Color option for buttons interfering with button color for Editor
When a theme has Custom Color option for buttons, the color set interferes with the color for all buttons in Editor.

### Steps to reproduce the behavior

1. Make sure to use Atomic site
2. Activate a theme that has Custom Color option for buttons (i.e. **Independent Publisher 2**)
3. Go to "Appearance > Customize > Colors & Backgrounds"
4. Change button color to black for example and save the change

Screen Shot on 2021-07-25 at 09:38:21

5. Open a page editor

#### What actually happened

All the buttons in editor shows up with black.

Screen Shot on 2021-07-25 at 09:36:58

#### Is this specific to the applied theme? Which one?

I reproduced this on **Independent Publisher 2** so far. I didn't have time to find other themes that have Custom Color option for buttons. But I believe this is reproducible even on other themes that have the button color option. 

#### Does this happen on simple or atomic sites or both?

Only Atomic.

sixhours:
`button:not('.menu-toggle')` was meant to avoid applying color annotation styles to the Menu toggle on mobile devices, but it has the unpleasant side effect of targeting all buttons that are not `.menu-toggle`, even if you're in the editor.

Removing the `:not('.menu-toggle')` specificity seems to work fine in the color annotations without changing things for existing users, and should fix the editor.

**Before**

Screen Shot 2021-08-03 at 3 29 27 PM

**After**

Screen Shot 2021-08-03 at 3 41 22 PM

Testing Instructions

* Apply this PR, then find the Independent Publisher 2 theme folder locally
* Rename the theme directory and change the name in `style.css` to `Independent Publisher Test` and ZIP it.
* Switch to an Atomic site and upload the zipped theme, and activate it
* Go to the Customizer and set the color scheme on your Atomic site such that the third color circle is black (or any color you'll notice)
* Save your changes
* Go to the editor by opening a page or post
* The editor buttons should all appear normal, without a black background.

Fixes #4330

This is the same PR as D65100-code -- they should be merged at the same time.


mreishus:
This looks better, but I am still seeing some issues when hovering:

https://user-images.githubusercontent.com/937354/128241120-debe109d-86d2-4198-b9e8-2fcc9b25cc0d.mp4

sixhours:
Eek, good catch, I see that, too.

The `button` selector is still causing problems for hover states, probably because the editor's styles for hover effects don't specify a background color. `button:not('.components-button'):hover` seems to work in the annotation.

It feels weird to put it in the theme as a workaround rather than being more specific in the editor, but it's a problem that's unique to custom colors, so I'm going to say it's fine. 🙃

mreishus:
Ok, it's definitely not taking over the editor controls now! :+1: My only remaining question (maybe I'm asking too many :laughing:) is where should I see this third color? I tried adding a button, a link, and a blockquote to a post and they all use blue elements:



The only place I saw it applied (after an admittedly brief look) was the 404 page:

sixhours:
> Ok, it's definitely not taking over the editor controls now! 👍 My only remaining question (maybe I'm asking too many 😆) is where should I see this third color? I tried adding a button, a link, and a blockquote to a post and they all use blue elements:

Great question! So the elements in the editor won't inherit the Custom Colors as far as I can tell, because users can choose to set those colors. This is because the theme wasn't annotated with blocks in mind. But buttons that are part of the theme files (like post navigation elements) will change colors. I found one example in some starter content, which shows a standard button at the bottom of the form block:

Screen Shot 2021-08-04 at 3 56 03 PM

But I just noticed another bug that I need to fix before that will show up properly. :D

mreishus (approved):
LGTM :+1:

Issue #4329, opened 2021-07-26
Skatepark: Duotone theme-level configuration


MaggieCabrera:
Skatepark includes in its design the application of duotone filter on the images it provides. The theme should apply this filter to all relevant blocks by default.

This will require https://github.com/WordPress/gutenberg/issues/33642 to be addressed, which will allow us to set a filter as default (and we should consider changing the default filter when the user changes the theme color as well).

MaggieCabrera:
Duotone apparently doesn't support css variables, so I'm looking into it to see if I can submit a patch for it

MaggieCabrera:
Apparently it's not trivial to make the duotone filters dynamic, so I opened an [issue](https://github.com/WordPress/gutenberg/issues/33905) about it on GB and contacted the devs on top of duotone so that this is on their radar as an enhancement.

jeffikus:
@MaggieCabrera and I spoke about this, and we're going to pause on this, as it's non blocking. For v1, the customer can make changes in the editor, but changing the global colors won't apply to all instances of the block. We can overcome this initially with head start and patterns, and this can be improved in the block in future.

pbking:
Added this to track [allowing a user to control the application of duotone](https://github.com/Automattic/themes/issues/4434) to all blocks.

PR #4367, opened 2021-08-05
Skatepark: Add comment styles

### LINKED ISSUE #4317: Skatepark: Comment form styles (dotcom and dotorg) (if needed)
**Dotorg comments:**

Screenshot 2021-07-23 at 19 10 57

mikachan:
This PR adds styles for the comment form and comments list on Skatepark.

Comments List:

| Logged In      | Logged Out |
| ----------- | ----------- |
| logged-in-comments      |  logged-out-comments    |

Comments Form:

| Logged In      | Logged Out |
| ----------- | ----------- |
|  logged-in-form  |  logged-out-form |

#### Text underneath 'Leave a Reply' / 'Your email address will not be published. Required fields are marked *':
I've left this in as I thought the information was important/needed, but it's not in the design. Let me know if this should be removed!

#### Avatars:
**Size**
In the designs, the avatars are set to 72px x 72px but the default size is set to 32px x 32px. I'm guessing this needs to be changed but I'm not sure how best to do it in a block-based theme. I couldn't find any similar examples from recent themes. Would appreciate any pointers!

**Duotone**
~~I haven't applied the duotone filter to the avatars as there's [a separate issue](https://github.com/Automattic/themes/issues/4329) for adding duotone support to Skatepark. Is this OK or should I include maybe a default setting in this PR?~~

#### Related issue(s):
Closes #4317.


MaggieCabrera:
It's fine to leave the duotone for the moment, I was working on it and I found a bug

skatepark/sass/blocks/_post-comments.scss
```diff
@@ -0,0 +1,169 @@
...
+
+.wp-block-post-comments {
+
+	line-height: var(--wp--custom--body--typography--line-height);
+
+	.reply {
+		margin-bottom: var(--wp--custom--margin--vertical);
+
+		a {
+			--wp--custom--button--typography--font-size: var(--wp--preset--font-size--normal);
```
mikachan, line 6: This is currently the wrong font size on this branch, and `-wp--preset--font-size--small` is correct. But I noticed these are about to be renamed, so I think 'normal' will be correct after that change.

melchoyce:
Hey, this is look great! Some small notes:

## Comment form
- Can we add .1em letter-spacing to the labels?
- On the logged-out comment form, can we have 45px of space between the "save my info" checkbox and the submit button instead of 30px?

## Comment list
- Can the date/edit links be medium (500) instead of regular weight?
- Let me know if it's not possible to use a bigger avatar, and I'll maybe tweak some spacing.
- Noticed the "awaiting moderation" view is a little messed up:
Screen Shot 2021-08-05 at 4 21 58 PM

^ We might want to test that in Blockbase & Quadrat to see if that's a problem for those themes as well

skatepark/sass/blocks/_post-comments.scss
MaggieCabrera: Yep, normal is the correct one, I just merged that branch so you can rebase to get that change

skatepark/sass/blocks/_post-comments.scss
```diff
@@ -0,0 +1,169 @@
...
+			"notes notes"
+			"comment comment"
+			"author email"
+			"url url"
+			"cookies-consent cookies-consent"
+			"form-submit form-submit";
+
+		input:not([type=submit]):not([type=checkbox]),
+		textarea {
+			border: 3px solid var(--wp--custom--color--primary);
```
MaggieCabrera: I think these should be picking up from theme.json (`custom.form.border`) so that all form-related stuff is consistent. In fact, that should be on blockbase too.

skatepark/sass/blocks/_post-comments.scss
```diff
@@ -0,0 +1,169 @@
...
+				}
+				& + ::before {
+					content: "";
+					position: absolute;
+					top: -0.4em;
+					left: 0;
+					width: 30px;
+					height: 30px;
+					border: 3px solid var(--wp--custom--color--primary);
+				}
```
MaggieCabrera: I see this is also on Quadrat, I'd say this should be on Blockbase too, but I'd appreciate other team member's opinions. Ideally we should add a checkbox inside theme.json's `custom.form` section. I feel strongly that we need consistent form styles defined for the theme from theme.json and reused whenever needed. This will be helpful when we style things like WooCommerce's forms, for example. A mixin for checkboxes like we have for buttons could probably be useful if I'm thinking ahead.

skatepark/sass/blocks/_post-comments.scss
```diff
@@ -0,0 +1,169 @@
...
+
+	.comment-reply-title {
+		margin-bottom: 30px;
+	}
+
+	form {
+		display: grid;
+		column-gap: 1.2em;
+		grid-template-rows: auto;
+		grid-template-areas:
```
MaggieCabrera, line 32: <3 `grid-template-areas`

skatepark/sass/blocks/_post-comments.scss
```diff
@@ -0,0 +1,169 @@
+@import "../../../blockbase/sass/blocks/button-mixins";
+
+.wp-block-post-comments {
+
+	line-height: var(--wp--custom--body--typography--line-height);
```
MaggieCabrera: if this is not the case by default, this is a good thing to do from Blockbase too

skatepark/sass/blocks/_post-comments.scss
```diff
@@ -0,0 +1,169 @@
...
+			display: none;
+		}
+	}
+
+	.commentmetadata {
+		margin-bottom: 36px;
+	}
+
+	.commentlist {
+		margin-bottom: 60px;
```
MaggieCabrera: this could be something like `calc( 3 * var(--wp--custom--margin--baseline) )`

skatepark/sass/blocks/_post-comments.scss
mikachan: Looks like this is the case by default actually, I've just tried removing it and it made no difference. It looks like it's applied to the body in the inline styles at the top of the page (the 'global-styles-inline-css' block).

skatepark/sass/blocks/_post-comments.scss
scruffian: I agree, this should be moved to Blockbase.

mikachan:
Thanks so much everyone for all the comments! I think I've addressed everything apart from moving the checkbox styles into Blockbase. I've started this but I felt like I was making this PR too big by including it here, so I'm thinking of making a separate issue.

@melchoyce: Sorry, I feel like a lot of your comments were me missing things from the design, gah. I've added each thing you've mentioned, other than increasing the avatar size*. 

> * Noticed the "awaiting moderation" view is a little messed up

Great spot! You're right, this also seems to be an issue in Blockbase and Quadrat as well, so I've added a fix for this in Blockbase.

Example in Quadrat before fix:
moderation-problem-quadrat

Skatepark with fix:
comment-awaiting-moderation

*To everyone: is there a standard way to increase the size of the avatars in the comments block? I know there's a way to do it within a PHP template (e.g. comments.php), but I'm not sure where to edit this as a block and I can't find any similar examples.

MaggieCabrera:
> *To everyone: is there a standard way to increase the size of the avatars in the comments block? I know there's a way to do it within a PHP template (e.g. comments.php), but I'm not sure where to edit this as a block and I can't find any similar examples.

The comments block right now is very basic and needs a lot of work, at first glance I think the avatar size is hardcoded, but it would make sense that in the future the block provided means to alter the size of it.

> I've started this but I felt like I was making this PR too big by including it here, so I'm thinking of making a separate issue.

I don't think you need to make a new one, it's easier to work on just one PR instead of two that depend on each other. The code won't get bigger, it will just move from skatepark to Blockbase (and deleted from Quadrat, sure). If you think it will help you, go ahead, but I don't see an issue in reviewing everything together.

mikachan:
> The comments block right now is very basic and needs a lot of work, at first glance I think the avatar size is hardcoded, but it would make sense that in the future the block provided means to alter the size of it.

Ah I see, thanks for confirming. Maybe for now it's best for @melchoyce to tweak some spacing to work with the hardcoded size.

> I don't think you need to make a new one, it's easier to work on just one PR instead of two that depend on each other.

Great OK, I'll carry on in here then. I wasn't sure what the best thing to do was! Thanks!

blockbase/sass/blocks/_post-comments.scss
```diff
@@ -12,14 +12,52 @@
...
+			border: var(--wp--custom--form--border--width) var(--wp--custom--form--border--style) var(--wp--custom--form--border--color);
+		}
+
+		.comment-form-cookies-consent {
+			input[type="checkbox"] {
+				margin-left: 0;
+				
+				& + ::before,
+				& + ::after {
+					box-sizing: border-box;
```
MaggieCabrera: These are working for Quadrat but not for Blockbase because of the missing `position:relative;`. I think they need to be `& + label::before` and we need to add `position:relative;` to `& + label`

blockbase/sass/blocks/_post-comments.scss
```diff
@@ -12,14 +12,52 @@
...
 			width: 100%;
 		}
+
+		input:not([type=submit]):not([type=checkbox]),
+		textarea {
+			border: var(--wp--custom--form--border--width) var(--wp--custom--form--border--style) var(--wp--custom--form--border--color);
+		}
+
+		.comment-form-cookies-consent {
+			input[type="checkbox"] {
```
MaggieCabrera, line 75: This needs the `@include hide-accesibly;` mixin to hide the original checkbox too

quadrat/theme.json
```diff
@@ -217,6 +217,31 @@
 				},
 				"typography": {
 					"fontSize": "var(--wp--preset--font-size--normal)"
+				},
+				"checkbox": {
+					"checked": {
+						"content": "\"\\2715\"",
```
MaggieCabrera, line 210: Smart!

skatepark/sass/blocks/_post-comments.scss
```diff
@@ -0,0 +1,142 @@
...
+			"notes notes"
+			"comment comment"
+			"author email"
+			"url url"
+			"cookies-consent cookies-consent"
+			"form-submit form-submit";
+
+		input:not([type=submit]):not([type=checkbox]),
+		textarea {
+			background: var(--wp--custom--color--background);
```
MaggieCabrera: I think the background is already defined on Blockbase

skatepark/sass/blocks/_post-comments.scss
```diff
@@ -0,0 +1,142 @@
...
+		display: grid;
+		column-gap: 1.2em;
+		grid-template-rows: auto;
+		grid-template-areas:
+			"notes notes"
+			"comment comment"
+			"author email"
+			"url url"
+			"cookies-consent cookies-consent"
+			"form-submit form-submit";
```
MaggieCabrera, line 38: I just noticed that this is repeated on Quadrat. I think it's safe to handle this in Blockbase too and repeat less code over the child themes, since they are mostly going to have the same layout, what do you think @kjellr?

skatepark/sass/blocks/_post-comments.scss
```diff
@@ -0,0 +1,142 @@
...
+			grid-area: cookies-consent;
+			margin-bottom: var(--wp--custom--margin--vertical);
+
+			input[type="checkbox"] {
+				@include hide-accesibly;
+				& + label {
+					margin-left: 0;
+					padding-left: 3em;
+					position: relative;
+				}
```
MaggieCabrera: yeah, this looks like it has to be done on the parent theme too, this is what's missing on Blockbase and why it's not working on it

quadrat/theme.json
```diff
@@ -217,6 +217,31 @@
 				},
 				"typography": {
 					"fontSize": "var(--wp--preset--font-size--normal)"
+				},
+				"checkbox": {
```
MaggieCabrera, line 208: Are we defining these only on the child themes? these should come from Blockbase and then rebuilt into the children

MaggieCabrera: And adapted when needed ofc

mikachan: The reason I skipped Blockbase (and didn't move the other checkbox styles) is because it looks like Blockbase is using a native checkbox:

native checkbox

This is what it looks like when I move all these custom checkbox related styles into Blockbase:

custom checkbox

If we ideally want all Blockbase and children using a custom checkbox, should I make the border of the custom Blockbase one any darker, similar to the native one? My screenshots are from Firefox. Mainly asking as the box is so tiny and I'm not sure the light grey is great for accessibility. It's using the same border colour as the other form inputs too, so then I don't know if they should all change or just the checkbox border. What do you think?

MaggieCabrera: Yeah, that makes sense. Blockbase styles are based off of Blank Canvas', which looks the same on Firefox (Chrome shows a very small checkbox with a tiny border). In turn Blank Canvas is a child of Seedlet:

Seedlet | Blank Canvas
--- | ---
Screenshot 2021-08-11 at 18 34 11 | Screenshot 2021-08-11 at 18 33 46

I think this is a minor change and I'd rather remove as much of the CSS from the children as possible. We could make the custom checkbox on Blockbase look the same color as it does on Firefox without being the native one but using our theme.json settings, or maybe we could change all the inputs borders on Blockbase a more contrasting color. @kjellr what do you think about this?

skatepark/sass/blocks/_post-comments.scss
kjellr: Yeah, definitely. I expect this comment form layout will be mostly identical across our themes.

quadrat/theme.json
kjellr: I'm not sure I follow. 😅 What are the two options here?

MaggieCabrera: so right now Blockbase (and Blank Canvas and Seedlet) have native checkboxes for the comments form. The idea is that since Quadrat and Skatepark used a styled one, have that code be in Blockbase too, and make it custom styled instead of native, but that brings the a11y issues @mikachan pointed out. I don't have a preference as to how to prevent that, besides wanting to avoid the code duplication of having the child themes all provide the custom css for the checkboxes instead of providing it from Blockbase.

kjellr: Thanks for clarification. I think it's fine to move the styles to Blockbase, but we should make sure that all form fields are made darker so as to avoid the a11y issue. (It looks like the checkbox border is just as light as the other form field borders, which are presumably inaccessible as well). 

Since that's more of a global change, I think it would be fine to explore it in a followup PR.

mikachan: Thanks @kjellr, that makes sense, I'll do that! Thanks @MaggieCabrera for clarifying.

MaggieCabrera:
I'm bringing this in, thank you @mikachan for your hard work on this, it's looking awesome!

PR #4368, opened 2021-08-05
Blockbase: Clear gutenberg transient

scruffian:
Since the transient name is being updated in Gutenberg (https://github.com/WordPress/gutenberg/pull/33844) we should also clear this in Blockbase.


jffng (approved):
Makes sense, LGTM.

PR #4369, opened 2021-08-05
Try adding "Stewart", a light blogging theme with a sidebar

kjellr:
This is a light Blockbase spinoff for blogging that includes a left sidebar. Still needs a little light polish (and a bunch of testing?), but it seems to work generally well as a block-based theme given the small amount of effort I've put into it. It's named "Stewart", after @ianstewart (who asked us to make a blogging theme with a sidebar). 

## Screenshot

new-theme


kjellr:
I kept the left sidebar content really simple for now, so that this could theoretically work as a universal theme pretty easily. It just contains a navigation block (which we already have working for Quadrat for example), and the social links block. (Do we have a way to make a social nav work in both classic themes and block themes?). 

That said, to make this a _truly_ customizable universal sidebar theme we'd probably need to figure out a way to let non-FSE folks customize the sidebar with widgets. 😬

ianstewart:
Let me know when you need a beta tester. I'll switch up my blog on WordPress.com. I _have_ to use this theme _now_. :)

scruffian:
I wonder if we can use a "widget area" block to achieve this.

kjellr:
Spent some time today making this a bit more of a "theme" and less of a blank starting point. I changed the colors and font, added a screenshot, and tidied up the templates. 

Before|After
---|---
stewart-old|stewart-new

kjellr:
Just a heads up that I've gone in and pushed a few commits to start separating this out from Blockbase. It should work just fine as a standalone block theme now, though there are some spacing inconsistencies and likely a few unnecessary theme.json styles. 

cc @jeffikus @scruffian

kjellr:
Known issues: 

- [x] Needs a font loading implementation.
- [x] Needs `theme.json` to be pruned of blockbase-only keys.
- [x] Needs consistent [top spacing](https://cloudup.com/c8jM4UYKgFI) for all templates. 
- [x] Needs to show all pages by default in the sidebar nav area. 
- [x] Figure out why the background color doesn't appear on the frontend.

stewart/functions.php
```diff
@@ -0,0 +1,71 @@
...
+	* supported by Work Sans, translate this to 'off'. Do not translate
+	* into your own language.
+	*/
+	$work_sans = esc_html_x( 'on', 'Work Sans: on or off', 'stewart' );
+
+	if ( 'off' !== $work_sans ) {
+		$font_families = 'Work+Sans:ital,wght@0,400;0,700;1,400;1,700';
+
+		$query_args = array(
+			'family' => $font_families,
```
kjellr, line 63: [Usually](http://themeshaper.com/2014/08/13/how-to-add-google-fonts-to-wordpress-themes/) this `$font_families` variable gets wrapped in `urlencode()`, but for some reason that wasn't working for me? Could use a second look.

kjellr:
This should be ready for review! I've tidied up all the templates and cleaned out the `theme.json`. 

Homepage|Single Post
---|---
Screen Shot 2021-12-08 at 15 05 08|Screen Shot 2021-12-08 at 15 05 18

stewart/theme.json
```diff
@@ -0,0 +1,361 @@
...
+					"slug": "background-and-foreground",
+					"name": "Background and Foreground"
+				},
+				{
+					"colors": [ "#C14420", "#F1F2F2" ],
+					"slug": "foreground-and-background",
+					"name": "Foreground and background"
+				}
+			],
+			"palette": [
```
pbking, line 66: Looks like the palette colors `primary` `secondary` and `selection` aren't used at all, perhaps they should be removed to avoid confusion.

stewart/functions.php
pbking: We don't appear to urlencode the value in blockbase.  This seems right to me.

scruffian:
At "tablet" sizes the columns go to 50/50:
Screenshot 2021-12-09 at 11 52 05

This makes the right column quite small. I'm not sure if this is acceptable, but if its not I think we need to fix it in Gutenberg. I've encountered the same limitation in Remote too, so I think we should have a way to turn it off regardless. I added an issue here: https://github.com/WordPress/gutenberg/issues/37252

stewart/theme.json
kjellr: Primary is actually used in a couple places (like hovers). In any case, I just went in and rearranged these so they the most often used colors are up front. I also replaced "selection" with "tertiary" so that these colors match up with the ones we're using in our Blockbase themes. That way if the user has existing patterns that use these color names, they'll translate over just fine.

kjellr:
> This makes the right column quite small. I'm not sure if this is acceptable, but if its not I think we need to fix it in Gutenberg. I've encountered the same limitation in Remote too, so I think we should have a way to turn it off regardless. I added an issue here: WordPress/gutenberg#37252

Yeah, this is not ideal. I just pushed a quick fix that uses a custom classname and makes those columns 100% wide on small screens. Should be fine until that issue is addressed.

scruffian:
> Yeah, this is not ideal. I just pushed a quick fix that uses a custom classname and makes those columns 100% wide on small screens. Should be fine until that issue is addressed.

I'm concerned about doing this in CSS as it means that this block behaves differently to other column blocks. The feedback we've had in the past about using CSS is "CSS should not be used for structural changes".

scruffian:
We should explore adding our Google font stack to this...

kjellr:
> I'm concerned about doing this in CSS as it means that this block behaves differently to other column blocks. The feedback we've had in the past about using CSS is "CSS should not be used for structural changes".

Yeah, I agree. But I think this is mostly harmless? I don't feel strongly, so I'm happy to strip it out and let the editor do its wrong thing for now. If the columns are 50/50 on tablets for now, I don't consider it a blocker. (EDIT: Reverted for now in 3fa0b28)

> We should explore adding our Google font stack to this...

Yeah, probably through a plugin and/or some dotcom functionality though, so it can be re-used. Doesn't make sense to add that into every single theme.

pbking:
This theme is completely crashing for me right now.  Can't figure out why.  No errors just a completely empty html payload.

Just this theme.  Not sure what's up.

skorasaurus:
> Yeah, probably through a plugin and/or some dotcom functionality though, so it can be re-used. Doesn't make sense to add that into every single theme.

I ran into it; requested at https://github.com/WordPress/gutenberg/issues/34752 with some short-term workarounds for it.

kjellr:
> This theme is completely crashing for me right now. Can't figure out why. No errors just a completely empty html payload.
> 
> Just this theme. Not sure what's up.

That's odd — @pbking what WP + Gutenberg version are you running? It's working fine for me w/Gutenberg 12.1 & WP 5.8.2.

pbking:
I was using Gutenberg 12.0.1 and WP 5.8.2
Testing again with Gutenber /trunk and the theme is working again.  So something to do with that slightly older version of Gutenberg I suppose...

Regardless it's working for me now.

kjellr:
Do you folks think this is in a decent-enough place where we can merge it in and get it up on dotcom for testing?

stewart/parts/header.html
```diff
@@ -0,0 +1,6 @@
+<!-- wp:group {"align":"full","layout":{"type":"flex","justifyContent":"space-between"},"style":{"spacing":{"padding":{"top":"var(--wp--custom--margin--horizontal, 30px)","bottom":"40px"}}},"className":"site-header"} -->
```
scruffian: I think we should remove this class

stewart/templates/header-footer-only.html
```diff
@@ -0,0 +1,10 @@
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
+
+<!-- wp:group {"tagName":"main"} -->
+<main class="wp-block-group">
+
+<!-- wp:post-content {"layout":{"inherit":true}} /-->
+</main>
+<!-- /wp:group -->
+
+<!-- wp:template-part {"slug":"footer","tagName":"footer","layout":{"inherit":true},"className":"site-footer-container"} /-->
```
scruffian: I think we should remove this class name too.

stewart/parts/header.html
MaggieCabrera: then we are missing the responsive styles: https://github.com/Automattic/themes/blob/trunk/blockbase/sass/base/_header.scss

stewart/templates/header-footer-only.html
MaggieCabrera: then we'll be missing the "sticky footer" styles https://github.com/Automattic/themes/blob/75542ea9cbaade68bc827b353bd8e23c8c69e8f8/blockbase/sass/base/_layout.scss#L12

stewart/theme.json
```diff
@@ -0,0 +1,361 @@
...
+				},
+				"spacing": {
+					"margin": {
+						"top": "0",
+						"bottom": "var(--wp--custom--margin--horizontal)"
+					}
+				}
+			},
+			"core/post-date": {
+				"color": {
```
scruffian: Setting these colors means that when a user updates the colors in the site editor, these ones stay the same.

stewart/style.css
```diff
@@ -0,0 +1,125 @@
...
+
+a {
+	text-decoration-thickness: 0.075ex;
+	text-underline-offset: 0.125em;
+}
+
+.wp-block-site-title a,
+.wp-block-post-title a,
+.wp-block-post-date a {
+	color: var(--wp--preset--color--foreground);
```
scruffian: Do we need these lines?

stewart/templates/header-footer-only.html
scruffian: This isn't a Blockbase child

stewart/parts/header.html
scruffian: This isn't a Blockbase theme

scruffian:
The main problem I see is that when I customize colors they don't all get changed:
Screenshot 2021-12-14 at 11 10 01

stewart/parts/header.html
MaggieCabrera: sorry I didn't check!

kjellr:
@scruffian I've adjusted what I can, but some of this I think is just an issue with Global Styles. You used to be able to edit the "Primary" color, but that's not possible anymore (unless I'm missing something?).

scruffian:
Yeah that's what I was noticing too. Seems like we might need to open a Gutenberg issue for that as it's going to be a bit of a blocker...

kjellr:
@scruffian It's still there actually, I think we both actually just overlooked this control: 
colors

scruffian (approved):
In that case I say LGTM

kjellr:
Awesome. I'll leave this up to you folks to merge and deploy. 👍

PR #4376, opened 2021-08-10
Blockbase: reduces specificity for search block selectors

MaggieCabrera:

This is a follow-up to https://github.com/WordPress/gutenberg/pull/33961 After that's merged we can actually remove the extra specificity that was needed for styles to persist on the editor. To test this check that the search block looks the same with and without this PR.

Test this with the mentioned PR active on your site, I'm waiting on tests to pass to merge it.


scruffian (approved):
LGTM

PR #4377, opened 2021-08-10
Blockbase: Load editor styles for blockbase before those of the child theme

scruffian:
In the front end we load the blockbase styles before the child theme, so we should do the same in the editor:

Before:
Screenshot 2021-08-10 at 10 33 37

After:
Screenshot 2021-08-10 at 10 34 01

### DIFF
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -1,7 +1,6 @@
 <?php
 if ( ! function_exists( 'blockbase_support' ) ) :
 	function blockbase_support() {
-		var_dump( 'blockbase support' );
 		// Alignwide and alignfull classes in the block editor.
 		add_theme_support( 'align-wide' );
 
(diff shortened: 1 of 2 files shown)


blockbase/functions.php
```diff
@@ -1,7 +1,7 @@
 <?php
 if ( ! function_exists( 'blockbase_support' ) ) :
 	function blockbase_support() {
-
+		var_dump( 'blockbase support' );
```
MaggieCabrera: I think you missed this one

skatepark/functions.php
```diff
@@ -3,7 +3,7 @@
 
 if ( ! function_exists( 'skatepark_support' ) ) :
 	function skatepark_support() {
-
+		var_dump('skatepark_support');
```
MaggieCabrera: and here

blockbase/functions.php
scruffian: oops!

MaggieCabrera (approved):
LGTM! Thanks for handling this

PR #4379, opened 2021-08-10
Blockbase: remove unused code

scruffian:
This removes some code in theme.json which is never used.


jffng (approved):
LGTM, thanks for rebuilding.

Issue #4308, opened 2021-07-23
Skatepark: Block patterns 


scruffian:
Skatepark should ship with the following block patterns:

- [x] Full-width image with aside caption https://github.com/Automattic/themes/pull/4396
<img width="598" alt="Screenshot 2021-07-26 at 09 42 11" src="https://user-images.githubusercontent.com/3593343/126951747-ce4bca0b-7ff1-4112-83c9-03573079cb36.png">

- [x] Text list with button https://github.com/Automattic/themes/pull/4390
<img width="578" alt="Screenshot 2021-07-26 at 09 43 01" src="https://user-images.githubusercontent.com/3593343/126951785-eef22d7f-7f68-4188-8701-13c2b737f122.png">

- [x] Columns in container https://github.com/Automattic/themes/pull/4449
<img width="517" alt="Screenshot 2021-07-26 at 09 44 05" src="https://user-images.githubusercontent.com/3593343/126951880-3550edc0-96d8-403d-8b36-bf2300fdaa3c.png">

- [x] Mixed media in container #4403
<img width="569" alt="Screenshot 2021-07-26 at 09 45 07" src="https://user-images.githubusercontent.com/3593343/126951987-f56ab4c9-b47a-4c01-b172-a906aa7b80bb.png">

- [x] Testimonial https://github.com/Automattic/themes/pull/4380
<img width="575" alt="Screenshot 2021-07-26 at 09 46 34" src="https://user-images.githubusercontent.com/3593343/126952391-776743c9-f1aa-421b-907d-db2bae48cd50.png">

- [x] Paragraph with quote https://github.com/Automattic/themes/pull/4414
<img width="579" alt="Screenshot 2021-07-26 at 09 46 38" src="https://user-images.githubusercontent.com/3593343/126952376-1af27891-b527-428c-be56-f077e520d72e.png">

- [x] Two columns of text https://github.com/Automattic/themes/pull/4397
<img width="582" alt="Screenshot 2021-07-26 at 09 46 41" src="https://user-images.githubusercontent.com/3593343/126952352-cbbea399-50d1-4d33-a52a-64910a502b27.png">

- [x] Footer https://github.com/Automattic/themes/pull/4343
<img width="581" alt="Screenshot 2021-07-26 at 09 46 51" src="https://user-images.githubusercontent.com/3593343/126952335-a2ac28b9-000d-42bd-98ab-ba650c4cb338.png">

MaggieCabrera:
[...]

PR #4380, opened 2021-08-10
Skatepark: Add Testimonial block pattern

mikachan:
This PR adds a Testimonial block pattern for Skatepark.

I couldn't see a design for the smaller resolutions, so I've had a go at stacking the elements and making the border full width (similar to some of the other elements). I also tried changing the blockquote from `flex` to `display: block` at mobile, as this makes the border the same width as the author content which looked nice too. @melchoyce, would appreciate any input from you!

| Larger Resolutions      | < 599px |
| ----------- | ----------- |
| desktop      | mobile    |

#### Related issue(s):
Part of #4308.


MaggieCabrera:
I understand the decision behind using the quote block but that means that we need to write a lot of CSS for it. Semantically it makes a lot of sense but it comes with a few problems:
- None of this CSS is customizable in the editor by the user
- As soon as the `testimonial` class is gone (for example: when we change themes) the style of the block changes dramatically

Maybe this is a good candidate to be a [Block Style](https://github.com/Automattic/themes/blob/0ce22c2eebf0ffb34289102dd37bda27a2ec7fe9/quadrat/inc/block-styles.php) if we stayed with the Quote Block approach. @kjellr @scruffian @jffng @pbking what do you all think? Should we do this kind of pattern with columns instead?

jffng:
Hmm. While the copy in the design suggests the pattern is for a testimonial / quote, this kind of layout could be used for other kinds of text too. So I'm not 100% sold on using the quote for semantic reasons.

I think my first intuition would have been to make this pattern with columns: the first column with a separator + normal paragraph, and the second being a paragraph with a preset font size. Then adding a `wrap-reverse` to get the stacking order as @mikachan has shown above for < 600px.

The pattern might also be able to be achieved without columns, and purely using a group block + sidebar flex layout (page 124 of every-layout.dev).

kjellr:
I think using a quote block style makes the most sense here. Some reasoning behind that: 

- We need some CSS in here so that it'll stack in the reverse order on mobile. 
- A block style is the official user-selectable way to apply a CSS class via a UI. 
- If the user switches themes, that block style disappears, but it'll still appear as a quote, which will maintain the role of each of the text fields here.

MaggieCabrera:
Using a block style makes a lot of sense to me. Ideally in the near future the quote block will let you control better the style of the blockquote and the cite so we wouldn't need as much CSS for something like this

mikachan:
Thanks so much all for the input. I'll try out a block style for the quote block next then. 

I used a blockquote for semantic reasons but I also completely understand the user may choose to use this for something that isn't a quote as well. 

> The pattern might also be able to be achieved without columns, and purely using a group block + sidebar flex layout (page 124 of every-layout.dev).

I really like the sound of this! I'll start with the quote block style, but this is great to have as another potential layout option.

MaggieCabrera:
I think we still need the block pattern, we just need the block pattern to be using the block style :D Otherwise, this looks great to me

mikachan:
> I think we still need the block pattern, we just need the block pattern to be using the block style :D Otherwise, this looks great to me

Ahh got it, will sort this!

melchoyce:
Looking good 🙌 

Any chance we can artificially add some more space above the citation so it lines up a little better with the text? 

image

(@kjellr, do you think this is a bad idea because it'll be uneven if someone changes fonts?)

kjellr:
> (@kjellr, do you think this is a bad idea because it'll be uneven if someone changes fonts?)

I think it's fine. 👍 We should maybe use an `em` or `ex` unit there.

mikachan:
Thanks both! I've just added `margin-top: 1.5ex` to the citation, mainly because I really liked the idea of using `ex` here.

MaggieCabrera:
> Thanks both! I've just added `margin-top: 1.5ex` to the citation, mainly because I really liked the idea of using `ex` here.

+1 it's a unit that makes sense to me when you really want to align things to the font, rather than just keep things generally proportional, LGTM!

MaggieCabrera (approved):

PR #4388, opened 2021-08-11
Skatepark: created duotone presets

MaggieCabrera:

After talking with @jeffikus we thought that the best course of action for the moment is to provide the presets for all the alternative palettes on the filter so that the user can change their images manually until a solution for https://github.com/WordPress/gutenberg/issues/33905 is applied and we can have dynamic filters. This PR adds filters for all the palettes in the theme:

Screenshot 2021-08-11 at 12 07 15

Screenshot 2021-08-11 at 12-03-50 Test duotone – free

#### Related issue(s):

#4329

### DIFF
--- a/skatepark/child-theme.json
+++ b/skatepark/child-theme.json
@@ -2,6 +2,38 @@
 	"settings": {
 		"color": {
 			"gradients": [],
+			"duotone": [
+                {
+                    "colors": [ "#000", "#BFF5A5" ],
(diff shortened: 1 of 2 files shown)


mikachan (commented):
Just from a technical perspective, these look and work great on my local install and sandbox 👍

kjellr (approved):
This makes sense and looks good! Thanks

PR #4389, opened 2021-08-11
Blockbase: Update font customization approach

scruffian:
This changes our approach to font customization in Blockbase to remove our reliance on CSS variables, and instead declare the fonts directly on the elements/blocks in Global Styles.

This has meant reworking the changes we make the the theme user json CPT when we save the customizer changes, as well as the preview. We also load fonts from Google directly using the declaration in the typography section which is a nice simplification.

This should allow us to work around the issues described in pNEWy-e9M-p2, but that needs testing in tandem with https://github.com/WordPress/gutenberg/pull/33919#issuecomment-896229913.

#### Related issue(s):
https://github.com/Automattic/wp-calypso/issues/55030

### DIFF
--- a/blockbase/inc/customizer/wp-customize-fonts.php
+++ b/blockbase/inc/customizer/wp-customize-fonts.php
@@ -232,9 +232,11 @@ class GlobalStylesFontsCustomizer {
 
 	function create_customization_style_element( $wp_customize ) {
 		wp_enqueue_style( 'global-styles-fonts-customizations', ' ', array( 'global-styles' ) ); // This needs to load after global_styles, hence the dependency
-		$css  = ':root,body{';
-		$css .= '--wp--custom--body--typography--font-family:' . $this->font_settings['body'] . ';';
-		$css .= '--wp--custom--heading--typography--font-family: ' . $this->font_settings['heading'] . '}';
+		$css  = 'body{';
+		$css .= 'font-family:' . $this->font_settings['body'] . ';';
+		$css .= '}';
+		$css .= 'h1,h2,h3,h4,h5,h6,.wp-block-post-title{';
+		$css .= 'font-family:' . $this->font_settings['heading'] . ';';
 		$css .= '}';
 		wp_add_inline_style( 'global-styles-fonts-customizations', $css );
 	}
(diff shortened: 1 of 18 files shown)


kjellr:
> We could update all Blockbase themes so that Pullquote is always a heading font (this seems like the best approach to me)

This sounds reasonable to me. 👍

pbking:
I gave this a whirl but I'm not able to get my head around where the google fonts values are handled.

```
Notice: Undefined index: google in /var/www/html/wp-content/themes/blockbase/functions.php on line 88
```

scruffian:
> I gave this a whirl but I'm not able to get my head around where the google fonts values are handled.

Oh good catch thanks, I've added this to the defaults in theme.json

scruffian:
> We could update all Blockbase themes so that Pullquote is always a heading font (this seems like the best approach to me)

I've done this.

I've also updated our approach to forms so that we just use `inherit`.

blockbase/inc/customizer/wp-customize-fonts.php
```diff
@@ -313,8 +314,8 @@ function initialize( $wp_customize ) {
 		$this->add_setting_and_control( $wp_customize, 'heading', __( 'Heading font', 'blockbase' ), $heading_font_default['slug'], $heading_font_selected['slug'] );
 	}
 
-	function get_font_family( $location, $configuration ) {
-		$variable = $configuration['settings']['custom'][ $location ]['typography']['fontFamily'];
+	function get_font_family( $array, $configuration ) {
+		$variable = get_settings_array( $array, $configuration );
```
jffng, line 318: Would it be better to name this something specific e.g. `$font-name-variable`

jffng (approved):
It works 🎉 I tested both Blockbase and Quadrat on wpcom:

Screen Shot 2021-08-11 at 12 39 52 PM

> This should allow us to work around the issues described in pNEWy-e9M-p2, but that needs testing in tandem with WordPress/gutenberg#33919 (comment).

I also needed to put https://github.com/WordPress/gutenberg/pull/33844/ on wpcom for it to work there. I have no major comments about the code.

PR #4396, opened 2021-08-12
Skatepark: full width image pattern

MaggieCabrera:

This PR creates a block pattern and a block style for a full-width image with an aside caption. Positioning the caption correctly was incredibly convoluted and brittle without an extra wrapper on the text inside the caption so I wrapped the text in a `strong` tag to make it simpler. I'm not sure how much I hate it. Any ideas to improve this are welcome.

Desktop | Mobile
--- | ---
Screenshot 2021-08-12 at 10-53-19 aaaa – free | Screenshot 2021-08-12 at 10-53-04 aaaa – free

(the wide group block on the screenshot is not part of the pattern, it's there for reference)

#### Related issue(s):

Part of #4308

### DIFF
--- a/skatepark/child-theme.json
+++ b/skatepark/child-theme.json
@@ -151,7 +151,8 @@
 				}
 			},
 			"layout": {
-				"contentSize": "684px"
+				"contentSize": "684px",
+				"wideSize": "1194px"
 			},
 			"line-height": {
 				"body": 1.6
--- a/skatepark/child-theme.json
+++ b/skatepark/child-theme.json
@@ -231,6 +231,11 @@
 					"fontFamily": "var(--wp--preset--font-family--red-hat-display)"
 				}
 			},
+			"core/separator": {
+				"border": {
+					"width": "0 0 3px 0"
+				}
+			},
 			"core/site-title": {
 				"typography": {
 					"fontSize": "var(--wp--preset--font-size--medium)",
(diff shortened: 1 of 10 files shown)


skatepark/child-theme.json
```diff
@@ -228,6 +229,11 @@
 					"letterSpacing": "0.1em"
 				}
 			},
+			"core/separator": {
+				"border": {
+					"width": "0 0 3px 0"
+				}
+			},
```
MaggieCabrera, line 239: on a first pass, I was using columns and a separator but then changed to use the caption of the image. I'm leaving this here even though it's unrelated, because it may be needed elsewhere

MaggieCabrera:
@melchoyce we need an alt text for this image

scruffian:
> I wrapped the text in a strong tag to make it simpler.

The problem with this is it makes it hard for users to implement via the block style

scruffian:
I added a commit to update the approach for the aside caption, but I probably missed some cases...

MaggieCabrera:
I didn't think about the Block Styles flow for the user, so @scruffian is totally right, using strong is a bad idea. Though his commit doesn't make the caption align with the wide block below. I'm not sure we can actually manage to make it look like the design without the extra wrapper, do you think having it full width aligned is a good compromise @melchoyce?

Screenshot 2021-08-12 at 16 00 19

melchoyce:
I wonder if this should just be an image + column block instead of using a caption 😕 

For alt text:

```
A skateboarder does a grab trick in a bowl-shaped skate park. In the background is a watching crowd, palm trees, and the ocean.
```

PR #4397, opened 2021-08-12
Skatepark: two columns of text block pattern 

MaggieCabrera:

This PR implements the two columns of text block pattern and a block style for the paragraph block with text-indent that is reused on other block patterns too in the design.

Screenshot 2021-08-12 at 16 36 31

@melchoyce I think we need real text for this one.

#### Related issue(s):

Part of #4308


melchoyce:
Yeah, let me work on writing up some copy we can use here.

MaggieCabrera:
I'm bringing this in with the new text. If there's any design issues about it, we can iterate.

PR #4402, opened 2021-08-12
Blockbase: Fix 404 template

### LINKED ISSUE #4323: Skatepark: Templates: Page templates: 404
There's no 404 page designed so we will default to inherit Blockbase's. Leaving the issue open so that designers can review if they want to make any tweaks to it for Skatepark. 
/cc @melchoyce

scruffian:
The blockbase 404 was missing some of the classes that are needed to do things like force the footer to the bottom of the page. It was also missing html and body tags. This adds that stuff to header and footer templates so it works like a classic theme for the 404 template. This will also address some of the concerns in https://github.com/Automattic/themes/issues/4323.

Also adding a 404 template for Skatepark so that the title matches the post title.

Fixes #4323

### DIFF
--- a/blockbase/footer.php
+++ b/blockbase/footer.php
@@ -1,9 +1,9 @@
-<div class="wp-block-template-part site-footer-container">
+<footer class="wp-block-template-part site-footer-container">
 		<?php
 			echo gutenberg_block_template_part( 'footer' );
 		?>
 	</div>
-</div>
+</footer>
 
 <?php wp_footer(); ?>
 
--- a/blockbase/footer.php
+++ b/blockbase/footer.php
@@ -1,0 +1,11 @@
+<div class="wp-block-template-part site-footer-container">
+		<?php
+			echo gutenberg_block_template_part( 'footer' );
+		?>
+	</div>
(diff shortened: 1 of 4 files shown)


blockbase/footer.php
```diff
@@ -0,0 +1,11 @@
+<div class="wp-block-template-part site-footer-container">
```
MaggieCabrera: I think this needs to be `footer` instead of `div`

MaggieCabrera (approved):
Besides the tag issue, this LGTM! 🚢

skatepark/404.php
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * The template for displaying 404 pages (not found)
+ *
+ * @package Skatepark
+ * @since 1.1.1
+ */
+get_header();
+?>
+	<main class="container-404">
+		<h2 class="wp-block-post-title"><?php _e( 'Oops! That page can&rsquo;t be found.', 'blockbase' ); ?></h1>
```
MaggieCabrera: we should change the domain here

PR #4403, opened 2021-08-12
Skatepark: add mixed media in container pattern

jffng:

Adds the Mixed Media in Container block pattern:

Screen Shot 2021-08-16 at 4 16 42 PM

#### Related issue(s):

Part of #4308


skatepark/inc/patterns/mixed-media-in-container.php
```diff
@@ -0,0 +1,90 @@
...
+	<div class="wp-block-column" style="flex-basis:66.66%"><!-- wp:paragraph {"className":"is-style-indented-paragraph"} -->
+	<p class="is-style-indented-paragraph">' . esc_html__( 'Skateboarding can teach people discipline, patience, and resiliency. It’s a great sport for kids with a lot of energy! We’ll help build up some structure and support in your life, all while you meet new people and have a ton of fun learning how to skateboard. Just think of all the cool tricks you’ll learn!', 'skatepark' ) . '</p>
+	<!-- /wp:paragraph -->
+	
+	<!-- wp:spacer {"height":40} -->
+	<div style="height:40px" aria-hidden="true" class="wp-block-spacer"></div>
+	<!-- /wp:spacer -->
+	
+	<!-- wp:image {"sizeSlug":"large","style":{"color":{"duotone":["#000","#BFF5A5"]}}} -->
+	<figure class="wp-block-image size-large"><img src="' . get_stylesheet_directory_uri() . '/assets/images/skateboard-sideways.jpg" " alt="' . esc_attr__( 'A skateboard laying on its side on top of concrete.', 'skatepark' ) . '"/></figure>
```
jffng, line 33: We should probably compress this.

skatepark/sass/block-patterns/_mixed-media-in-container.scss
```diff
@@ -0,0 +1,6 @@
+.skatepark-mixed-media-in-container {
+	hr, p, h1, h2, h3, h4, h5, h6 { // Shrink default vertical margins for all the elements in this pattern
```
jffng, line 2: I'm not crazy about this but not sure of a better way to reduce and standardize the margins between headings, paragraphs, and separators for this pattern.

mikachan: I can't think of another way to target these elements... I think this is fine, especially as it's specific to this pattern.

mikachan (commented):
Given this a test and it's looking really good to me 👍

MaggieCabrera:
This is not specific to this PR but I guess we could fix it here. The text-indent block style for the paragraph is not looking nicely on mobile:

Design | Theme
--- | ---
Screenshot 2021-08-18 at 10 47 17 | Screenshot 2021-08-18 at 10 48 36

melchoyce:
Should I hold off on reviewing this one until https://github.com/Automattic/themes/pull/4436 is merged, or is that already in this PR?

skatepark/sass/block-patterns/_mixed-media-in-container.scss
MaggieCabrera: makes sense to me that we do it this way, since we are adding the margin globally from the theme css, it's not really the editor that's failing us here. In the end this is just an exception, right?

jffng:
Sorry for the delay replying here. I don't think this should be held up by #4436, I think it can be reviewed from a design perspective.

> The text-indent block style for the paragraph is not looking nicely on mobile:

It sounds like the way we handle the text indent is [changing](https://github.com/Automattic/themes/issues/4445), so maybe let's leave the mobile styling for the PR that addresses it to avoid conflicts.

skatepark/sass/block-patterns/_mixed-media-in-container.scss
jffng: Yeah they are exceptions. But if we [reduce the vertical margins](https://github.com/Automattic/themes/pull/4444#issuecomment-902833612), I don't think these exceptions will be needed.

melchoyce:
The percents do scale nicely on mobile, though I don't know that I have a strong opinion on specifically using percents — my main concern is making sure it looks consistent across other patterns and blocks.

Could we increase the spacing a tiny bit? Thinking:

```
padding-top: 3%;
padding-right: 4%;
padding-bottom: 4%;
padding-left: 4%;
```

melchoyce (approved):

jffng:
Thanks for the ✅ ! I compressed the image, so I'll bring this in. If it turns out we can remove the CSS that alters the vertical spacing, we'll do it in a follow up PR.

PR #4406, opened 2021-08-13
Skatepark: Add default duotone configuration to theme.json

### LINKED ISSUE #4329: Skatepark: Duotone theme-level configuration

scruffian:
This adds a default duotone filter to all image blocks.

Needs https://github.com/WordPress/gutenberg/pull/34073/files

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/4329

### DIFF
--- a/skatepark/child-theme.json
+++ b/skatepark/child-theme.json
@@ -313,12 +313,13 @@
 			},
 			"core/image": {
 				"color": {
-					"duotone": "var(--wp--preset--duotone--red-filter)"
+					"duotone": "url(#wp-duotone-red-filter)"
+
 				}
 			},
 			"core/cover": {
 				"color": {
-					"duotone": "var(--wp--preset--duotone--blue-filter)"
+					"duotone": "url(#wp-duotone-blue-filter)"
 				}
 			}
 		},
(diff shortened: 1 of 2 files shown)


skatepark/child-theme.json
```diff
@@ -310,6 +310,16 @@
 				"typography": {
 					"letterSpacing": "0.05em"
 				}
+			},
+			"core/image": {
+				"color": {
+					"duotone": "var(--wp--preset--duotone--red-filter)"
+				}
+			},
+			"core/cover": {
```
scruffian: This one doesn't work yet.

skatepark/theme.json
```diff
@@ -537,6 +537,11 @@
 					"fontStyle": "normal"
 				}
 			},
+			"core/post-author": {
+				"color": {
+					"duotone": "url(#wp-duotone-red-filter)"
+				}
+			},
```
MaggieCabrera: I added this too

MaggieCabrera: It works fine!
Screenshot 2021-09-02 at 15 58 23

MaggieCabrera:
I think we can close this one

PR #4410, opened 2021-08-16
Skatepark: Add search template

### LINKED ISSUE #4322: Skatepark: Templates:Page templates: search
There's no Search page designed so we will default to inherit Blockbase's. Leaving the issue open so that designers can review if they want to make any tweaks to it for Skatepark.
/cc @melchoyce

mikachan:
This PR adds a search block template for Skatepark.

Here's how it's looking so far:

search

@melchoyce, I know you mentioned adding a page title in the original issue. If we add this to the block template, I think that currently means we can't translate the title text. But then I believe we can get around this by creating a PHP template instead. Should we go with a PHP template so we can have a page title?

I've also added the search block at the top and made the results titles all H3.

#### Related issue(s):
Closes #4322

### DIFF
--- a/skatepark/block-templates/search.html
+++ b/skatepark/block-templates/search.html
@@ -1,0 +1,20 @@
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
+
+<!-- wp:group {"layout":{"inherit":true}, "tagName":"main"} -->
+<main class="wp-block-group">
+
+	<!-- wp:search {"label":"","buttonText":"Search"} /-->
+
+	<!-- wp:query {"queryId":1,"query":{"perPage":10,"pages":0,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true}} -->
+	<!-- wp:query-loop -->
+
+		<!-- wp:post-title {"level":3,"isLink":true} /-->
+		<!-- wp:post-excerpt /-->
+
+	<!-- /wp:query-loop -->
+	<!-- /wp:query -->
+
+</main>
+<!-- /wp:group -->
+
+<!-- wp:template-part {"slug":"footer","tagName":"footer","layout":{"inherit":true},"className":"site-footer-container"} /-->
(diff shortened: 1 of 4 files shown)


melchoyce:
> If we add this to the block template, I think that currently means we can't translate the title text. But then I believe we can get around this by creating a PHP template instead. Should we go with a PHP template so we can have a page title?

Nah, in that case, let's go without.

I'll test this today!

melchoyce:
Any idea why I'm getting this error?

image

melchoyce:
What if we added separators between posts on this page? I think it might help scannability.

image

mikachan:
> Any idea why I'm getting this error?

Hmm.. I've not seen that before, not sure why you're seeing it, sorry! I'm guessing it's something to do with a Gutenberg update maybe? Are you seeing it on any other themes?

> What if we added separators between posts on this page? I think it might help scannability.

Yeah, I think this looks good, will add this today.

melchoyce:
> Hmm.. I've not seen that before, not sure why you're seeing it, sorry! I'm guessing it's something to do with a Gutenberg update maybe? Are you seeing it on any other themes?

Nope, only this PR 😕 I don't see it on trunk or other themes.

skatepark/block-templates/search.html
```diff
@@ -0,0 +1,27 @@
...
+<main class="wp-block-group">
+
+	<!-- wp:search {"label":""} /-->
+
+	<!-- wp:spacer {"height":45} -->
+		<div style="height:45px" aria-hidden="true" class="wp-block-spacer"></div>
+	<!-- /wp:spacer -->
+
+	<!-- wp:query {"queryId":1,"query":{"perPage":10,"pages":0,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true}} -->
+	<!-- wp:query-loop -->
```
jffng: I think if you change this to `post-template` instead of `query-loop` it will fix the warning, e.g. https://github.com/Automattic/themes/blob/trunk/blockbase/block-templates/search.html#L8-L13

mikachan: Thanks @jffng! I've just pushed this, hopefully it fixes the warning 🤞

MaggieCabrera (approved):
LGTM!

Issue #4417, opened 2021-08-16
Skatepark: Page/post headings without featured images should have a divider


melchoyce:
## Steps to replicate

Go to a page or post that **doesn't** include a featured image

## Result

There isn't anything between the page/post title and the content:

![image](https://user-images.githubusercontent.com/2846578/129599943-942d5a6c-0414-4de0-9530-999a38bbcc3c.png)

## Expected

A dividing line between the title and content:

![image](https://user-images.githubusercontent.com/2846578/129599974-36fcf055-267a-4eed-b691-7d12dc79c1bf.png)

Follow-up to #4416

PR #4418, opened 2021-08-16
Skatepark: Update green color

kjellr:
While comparing [screenshots in another issue](https://github.com/Automattic/themes/issues/4417), I noticed that the green in use right now isn't quite as bright and saturated as the green from the comps. This is because browser color profiles differ somewhat from the profile in Figma. 

This PR replaces that color with a different value that looks more similar to the original comps. 

Before|After
---|---
Screen Shot 2021-08-16 at 3 00 25 PM|Screen Shot 2021-08-16 at 3 00 00 PM

### DIFF
--- a/skatepark/child-theme.json
+++ b/skatepark/child-theme.json
@@ -4,7 +4,7 @@
 			"gradients": [],
 			"duotone": [
                 {
-                    "colors": [ "#000", "#BFF5A5" ],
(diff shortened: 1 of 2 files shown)


kjellr:
(also, I double checked, and this new color still passes all contrast checks — even AAA!)

melchoyce (approved):
LGTM 👍 Thanks for your keen eye!

PR #4436, opened 2021-08-17
Blockbase: remove margin-top from first descendants of columns and groups

### LINKED ISSUE #4412: Skatepark: Group with border missing spacing on front-end
## Steps to replicate

Paste this into your editor:

```
<!-- wp:group {"align":"wide","style":{"border":{"style":"solid"}},"borderColor":"primary"} -->
<div class="wp-block-group alignwide has-border-color has-primary-border-color" style="border-style:solid"><!-- wp:columns -->
<div class="wp-block-columns"><!-- wp:column -->
<div class="wp-block-column"><!-- wp:heading {"level":4} -->
<h4>Heading</h4>
<!-- /wp:heading -->

<!-- wp:paragraph -->
<p>Skate ipsum dolor sit amet, poseur nollie casper pop shove-it. Kickturn noseblunt slide tailslide lien air Bonite. Rocket air poseur snake shoveit Chico Brenes. Crooked grind locals casper slide bank.</p>
<!-- /wp:paragraph --></div>
<!-- /wp:column -->

<!-- wp:column -->
<div class="wp-block-column"><!-- wp:heading {"level":4} -->
<h4>Heading</h4>
<!-- /wp:heading -->

<!-- wp:paragraph -->
<p>Vert quarter pipe tailslide grab. Nose slide Bam Margera rocket air axle set disaster. Fast plant crail grab Japan air chicken wing Bullet 66. Hurricane coping masonite crail grab Brooklyn Banks.</p>
<!-- /wp:paragraph --></div>
<!-- /wp:column --></div>
<!-- /wp:columns -->

<!-- wp:image {"align":"wide","id":9,"sizeSlug":"full","linkDestination":"none"} -->
<figure class="wp-block-image alignwide size-full"><img src="https://cldup.com/BrvgOAqDT3.jpg" alt="" class="wp-image-9"/></figure>
<!-- /wp:image -->

<!-- wp:paragraph -->
<p><meta charset="utf-8">Skate ipsum dolor sit amet, poseur nollie casper pop shove-it. Kickturn noseblunt slide tailslide lien air Bonite. Rocket air poseur snake shoveit Chico Brenes. Crooked grind locals casper slide bank. Disaster frigid air bail aerial. Hang ten smith grind tailslide sick.</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group -->
```

## Result

Back-end:

image

Front-end:

image

## Expected

Spacing should be consistent between the back- and front-end of the site, and additionally, should be larger — 60px would probably work best.

jffng:

This PR removes the top margin of first children of columns and groups. It started as an exploration to add a default padding to all groups with borders.

#### Related issue(s):

Helps with #4412

### DIFF
--- a/skatepark/sass/base/_text.scss
+++ b/skatepark/sass/base/_text.scss
@@ -2,8 +2,3 @@
 ::selection {
 	color: var(--wp--preset--color--background);
 }
-
-p {
-	margin-top: calc( 0.5 * var(--wp--custom--margin--vertical) );
-	margin-bottom: calc( 0.5 * var(--wp--custom--margin--vertical) );
-}
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -159,6 +159,10 @@ p {
 	margin-bottom: var(--wp--custom--margin--vertical);
 }
 
+.wp-block-column :first-child, .wp-block-group :first-child {
+	margin-top: 0;
+}
+
 .image-no-margin {
 	margin: 0;
 }
--- a/blockbase/sass/base/_text.scss
+++ b/blockbase/sass/base/_text.scss
@@ -11,3 +11,9 @@ p {
 	margin-top: var(--wp--custom--margin--vertical);
 	margin-bottom: var(--wp--custom--margin--vertical);
 }
+
+.wp-block-column, .wp-block-group {
+	:first-child {
+		margin-top: 0;
+	}
+}
--- a/skatepark/assets/theme.css
+++ b/skatepark/assets/theme.css
@@ -147,10 +147,6 @@ p {
 	background-color: var(--wp--custom--button--color--text);
 }
 
-.wp-block-group.has-border-color {
-	padding: var(--wp--custom--margin--vertical);
-}
-
 .wp-block-post-comments .reply a {
 	--wp--custom--button--typography--font-size: var(--wp--preset--font-size--normal);
 	border-style: var(--wp--custom--button--border--style);
--- a/skatepark/assets/theme.css
+++ b/skatepark/assets/theme.css
@@ -43,9 +43,14 @@
 	color: var(--wp--preset--color--background);
 }
 
+<<<<<<< HEAD
 p {
 	margin-top: calc( 0.5 * var(--wp--custom--margin--vertical));
 	margin-bottom: calc( 0.5 * var(--wp--custom--margin--vertical));
+=======
+.wp-block-column > p:first-child, .wp-block-column > h1:first-child, .wp-block-column > h2:first-child, .wp-block-column > h3:first-child, .wp-block-column > h4:first-child, .wp-block-column > h5:first-child, .wp-block-column > h6:first-child, .wp-block-group > p:first-child, .wp-block-group > h1:first-child, .wp-block-group > h2:first-child, .wp-block-group > h3:first-child, .wp-block-group > h4:first-child, .wp-block-group > h5:first-child, .wp-block-group > h6:first-child {
+	margin-top: 0;
+>>>>>>> Refactor text styles and add padding to groups w border.
 }
 
 .wp-block-post-author__content {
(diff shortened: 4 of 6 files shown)


skatepark/sass/base/_text.scss
```diff
@@ -1,4 +1,12 @@
 // Text selection text color
 ::selection {
 	color: var(--wp--preset--color--background);
+}
+
+.wp-block-column, .wp-block-group {
```
jffng: These styles might better belong if blockbase, but I'm not sure about this idea in general. 

My reasoning for adding it is — I don't think that the first piece of text should have a top margin when contained in a column or group, because it ends up stacking if there is a padding declared. But maybe there is a better way of solving this.

MaggieCabrera: I'm not sure we would do this for every theme, it feels kind of opinionated, it makes sense for Skatepark because of the border but that's it. When I first saw the pattern using the border I was thinking about it having a block style even but this approach makes sense too, but just in the scope of this specific theme.

kjellr: > ... it makes sense for Skatepark because of the border but that's it.

Since Blockbase [opts into border controls](https://github.com/Automattic/themes/blob/85249e67b98fc7b8d2b40521c17bee49b7ae799f/blockbase/theme.json#L32-L37), there's a UI for users to create this bordered effect on any Group block though — using Blockbase or any of its children. 

I don't love these rules (and I hope Gutenberg adds greater per-block margin controls so folks can manually fix this sort of thing in the editor), but I think if we add these, they should be in Blockbase.

kjellr: Also, putting them in Blockbase would mean more universal expectations around this behavior for our themes, and also would make it easier to strip this out if it's not needed in the future. (Or if we come up with a better solution)

MaggieCabrera: yeah, I see what you mean, I don't like it but I agree, and having it on Blockbase is easier to control and go back to it when it eventually can be done better. I just hope we don't have to do this too much, Blockbase is growing instead of shrinking.

jffng:
I tinkered with this some more and am coming to a different conclusion — I don't think we should add these rules at all. My reasoning:

- [This PR](https://github.com/Automattic/themes/pull/4448) should solve the editor and front-end mismatch reported in the [original issue](https://github.com/Automattic/themes/issues/4412).
- The `.has-border-color` class is only applied when a border color is specified. The default border is applied via inline css directly to the element, so we do not have a reliable way to add padding to _all_ groups with borders because not all of them have the `has-border-color` class.

That leaves us with two options:

1. Apply a default padding to _all_ group blocks via theme.json. This approach seems like it could be okay, because the user can overwrite this default padding it in the block inspector.
2. Apply the padding in our patterns via the group's block style attributes, as we're doing here: https://github.com/Automattic/themes/blob/5626fd438a726045dc4a99d7e9cda859635f8ac8/skatepark/inc/patterns/mixed-media-in-container.php#L15 The downside here is that if a user creates a group block and adds a border on their own, it will have no padding and they will have to add it themselves.

scruffian:
I agree that these rules should live in blockbase. I have removed the border rules from skatepark altogether, I agree that they should live in the pattern.

jffng:
Hi @melchoyce — would you mind retesting with the markup you provided in #4412? 

To simplify my [long winded comment](https://github.com/Automattic/themes/pull/4436#issuecomment-902893402) above, we do not have a reliable way to add padding to groups with borders. Despite this, I think this PR is still an improvement.

melchoyce:
Just checking, the end goal is it should look like this by default?

image

jffng:
That's right.

melchoyce:
Cool. Let's merge this, and talk about this option more when @kjellr is back:

> Apply a default padding to all group blocks via theme.json. This approach seems like it could be okay, because the user can overwrite this default padding it in the block inspector.

We should also consider bumping this issue upstream, if that's possible (I might be understanding the problem)

scruffian:
> We should also consider bumping this issue upstream, if that's possible (I might be understanding the problem)

Added an issue for this https://github.com/WordPress/gutenberg/issues/34315

PR #4438, opened 2021-08-18
Skatepark: single+post templates

### LINKED ISSUE #4387: Skatepark: Page template - singular
Screenshot 2021-08-11 at 10-27-07 Gutenberg Group – free

This is what it looks like right now on trunk. @melchoyce do you think we should change something about this template for Skatepark? Right now it's inheriting Blockbase's singular template.

### LINKED ISSUE #4428: Skatepark: Singular template always shows a comments message
## Steps to replicate

1. Add a new page, let it use the singular template.
2. In Page > Discussion settings, ensure that "Allow comments" is unchecked (this is the default for pages). 
3. Publish & visit the front-end.
4. Note that a "Comments are closed." message appears just above the footer. 

## Expected

If comments were never on for the post, there should be no message that appears. (This may be an upstream Gutenberg bug?). It looks like there's currently no way to turn this off other than changing the page template. 

## Screenshot

Screen Shot 2021-08-17 at 9 31 43 AM

MaggieCabrera:

This PR removes the singular template and creates a single and page templates according to the discussion on https://github.com/Automattic/themes/issues/4387.

This PR uses https://github.com/WordPress/gutenberg/pull/34113 and https://github.com/WordPress/gutenberg/pull/34142 for duotone support on the featured image and font-weight support on the post terms block.

**Single:**

Screenshot 2021-08-18 at 17-21-49 5 Simple Skateboarding Tricks – free

**Page:**

Screenshot 2021-08-18 at 17-21-39 5 Simple Skateboarding Tricks – free

The post author block and the post navigation blocks don't allow for the features that are designed for the theme (the "posts published by" text and the previous/next post tags). I didn't add the Prev/Next labels as a paragraph because that won't be translatable if it's on the block template.

This PR fixes a bug on Blockbase and Mayland for the Post term blocks icon, this block has changed its class. 

#### Related issue(s):

Closes https://github.com/Automattic/themes/issues/4387


kjellr:
Not sure if this is new from this PR, but I'm seeing just a little weirdness in the category icon here: 

Screen Shot 2021-08-18 at 1 44 14 PM

kjellr:
Also, can we do just a little refinement around the placement of the author avatar? I'd like to move it down just a few pixels so that it aligns with the top of the author name: 

Current|After
---|---
Screen-Shot-2021-08-18-at-1 51 24-PM|after

Also, if it's possible to do something like this when there's no author bio, that would be great: 

Screen-Shot-2021-08-18-at-1 51 24-PM-alt3

blockbase/sass/base/_mixins.scss
```diff
@@ -12,3 +12,19 @@
...
+	margin-right: calc(2 * var(--wp--custom--margin--baseline) );
+	&:before {
+		align-self: center;
+		content: '';
+		display: inline-block;
+		margin-right: calc(0.5 * var(--wp--custom--margin--baseline) );
+		height: 16px;
+		width: 16px;
+		mask-size: contain;
+		mask-repeat: no-repeat;
```
MaggieCabrera, line 28: @kjellr The issue you are seeing with the icons, I fixed it with this line, I'm not sure why you are not seeing it, is this line being applied on your test site?

kjellr: It's not — looks like [safari requires a `-webkit` prefix](https://caniuse.com/mdn-css_properties_mask-size).

MaggieCabrera: I thought the sass compiler took care of that! I'll add it, thank you for checking!

melchoyce:
Pages are looking good — only notes I have there are increasing spacing above and below the page title, and between the page content and the footer:

image

Do y'all want to tackle https://github.com/Automattic/themes/issues/4417 in here as well, or wait until this PR is merged?

Will review posts next 👍

MaggieCabrera:
Oh, about #4417 I'm not sure how we can know if the featured image is present or not

kjellr:
Yeah, we may just need to make that a custom page template?

melchoyce:
Ah, is the featured image class not output in an FSE theme?

melchoyce:
re: posts, aside from @kjellr's comments, the only other issues I'm seeing are also tighter spacing. If we could do 120px between the post meta and comments list, and then 120px between the comments list and comment form, that would be great.

MaggieCabrera:
I've done all the changes except the separator/border on the page title, there's no class being added to control this, should we create an alternative template for this? It would be two templates, one for the page and one for the post without an image.

melchoyce:
This PR looks good, so let's go ahead and merge it and maybe talk about the template question in another issue — I feel like I need to think on it 🤔

Issue #4435, opened 2021-08-17
Blockbase: Ponyfill css overrides block-level global styles for Button block


jeyip:
## Description

I'm seeing an issue with block-level global styling being overridden in the dotcom site editor and frontend (atomic + simple sites), and it looks like it has to do with the [ponyfill.css](https://github.com/Automattic/themes/blob/trunk/blockbase/assets/ponyfill.css) file in `blockbase`. When I remove `ponyfill.css`, per block global styles work as expected.

**List of affected blocks**
1. Button (text color and background color)
2. ~Cover (text color and background color)~ [Not caused by blockbase](https://github.com/Automattic/themes/issues/4435#issuecomment-901524216)
3. ~Site Title (text color and background color)~ [Not caused by blockbase](https://github.com/Automattic/themes/issues/4435#issuecomment-902577643)
4. Cover and Group blocks (right/left padding)

## Steps to replicate

1. On a simple or atomic site, activate a block theme that depends on `blockbase`
2. Activate the site editor
3. Navigate to the site editor
4. Insert a button block onto the page and add text content
5. Open the global styles sidebar, and select the "By Block Type" menu tab
6. Unfold the Button block-level global styles panel and select different color options
7. Confirm that the color options do not change according to user customizations

### Button Block
![2021-08-16 13 58 39](https://user-images.githubusercontent.com/5414230/129790854-598db288-59b6-4a7b-abba-1c24df7e48d7.gif)

## Result
User customized global styles (like background color) weren't applied to the block

## Expected
The block styles would change according to user customizations

## Notes
* Slack Discussion: p1629226103080900-slack-C029FM1EH

pbking:
The button portion of the related ponyfill is [here](https://github.com/Automattic/themes/blob/trunk/blockbase/sass/blocks/_button.scss).  That relates to the video shown but it's not the only place where this is happening.

This is happening because many of Blockbase's selectors have a higher specificity than the styles generated by Global Styles.  This is especially obvious with the button which has a rich ponyfill system including hover states (which are going to be particularly problematic to address in this situation).

jffng:
I believe this affects the editor and the view slightly differently. 

In the case of the button (ignoring the hover states issue), we could theoretically solve the issue by remove the wrapping selector `.wp-block-button` and the global styles customizations win because they load later in the cascade. @mikachan and I tested this and it works in the editor. 

[...]

PR #4439, opened 2021-08-18
Blockbase: Refactor Button Color styles

jffng:

This PR is an attempt to solve one part of #4435 by lowering the CSS specificity of the button styles. It removes the hover styles for now while we discuss a way forward there.

We discussed a few ideas p1629312308145500/1629295598.131100-slack-C029FM1EH, I'm not sure this is the right approach but wanted to get a feel for the problem and spark a discussion in any case.

**To test**
- ~Check out https://github.com/WordPress/gutenberg/pull/34147~
- Activate blockbase
- Add a button somewhere to a template via the Site Editor
- Change the button's colors (text and background) and save
- Verify those customizations show up correctly in the editor and view
- Go to the post editor
- Add a button, verify you can still use the "Outline" block style variation

#### Related issue(s):

#4435

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -254,16 +254,7 @@ input[type=checkbox] + label {
 	padding-bottom: calc( var(--wp--custom--button--spacing--padding--bottom) + var(--wp--custom--button--border--width));
 	padding-left: calc( var(--wp--custom--button--spacing--padding--left) + var(--wp--custom--button--border--width));
 	padding-right: calc( var(--wp--custom--button--spacing--padding--right) + var(--wp--custom--button--border--width));
-	font-weight: var(--wp--custom--button--typography--font-weight);
-	font-family: inherit;
-	font-size: var(--wp--custom--button--typography--font-size);
-	line-height: var(--wp--custom--button--typography--line-height);
 	text-decoration: none;
-	opacity: 1;
-	color: var(--wp--custom--button--color--text);
-	background-color: var(--wp--custom--button--color--background);
-	border-color: currentColor;
-	border-radius: var(--wp--custom--button--border--radius);
 }
 
 .wp-block-button.wp-block-button__link:not(.has-background):not(.has-text-color):hover, .wp-block-button.wp-block-button__link:not(.has-background):not(.has-text-color):focus, .wp-block-button.wp-block-button__link:not(.has-background):not(.has-text-color).has-focus,
@@ -284,11 +275,6 @@ input[type=checkbox] + label {
 .wp-block-button .wp-block-button__link:not(.has-background):not(.has-text-color):hover svg,
 .wp-block-button .wp-block-button__link:not(.has-background):not(.has-text-color):focus svg,
 .wp-block-button .wp-block-button__link:not(.has-background):not(.has-text-color).has-focus svg {
-	fill: var(--wp--custom--button--color--text);
-}
-
-.wp-block-button.wp-block-button__link svg,
-.wp-block-button .wp-block-button__link svg {
 	fill: var(--wp--custom--button--color--text);
 }
 
--- a/blockbase/sass/blocks/_button.scss
+++ b/blockbase/sass/blocks/_button.scss
@@ -10,7 +10,8 @@
 	&.wp-block-button__link,
 	.wp-block-button__link {
 		@include button-hover-styles;
-		@include button-main-styles;
+		@include button-padding-styles;
+		text-decoration: none;
 	}
 	&.is-style-outline {
 		&.wp-block-button__link,


MaggieCabrera:
Thanks for looking into this! I've noticed a couple of things:

First of all this PR breaks the hover on the frontend like this



And also I think we are overriding too many of the styles we define on theme.json with our CSS:

Screenshot 2021-08-19 at 12 01 26

Skatepark theme.json defines:


```
"styles": {
  "blocks": {
	  "core/button": {
		  "border": {
			  "radius": "var(--wp--custom--button--border--radius)"
		  },
		  "color": {
			  "background": "var(--wp--custom--button--color--background)",
			  "text": "var(--wp--custom--button--color--text)"
		  },
		  "typography": {
			  "fontFamily": "var(--wp--preset--font-family--red-hat-display)",
			  "fontSize": "var(--wp--custom--button--typography--font-size)",
			  "fontWeight": "var(--wp--custom--button--typography--font-weight)",
			  "lineHeight": "var(--wp--custom--button--typography--line-height)",
			  "letterSpacing": "0.1em",
			  "textTransform": "uppercase"
		  }
	  },
  }
}
```

And we are redefining these values with our `button-main-styles` mixin. This mixin makes sense for other buttons on the theme that don't have all the controls that the core button has. The search button or the comment form buttons are good examples of why we need this for consistency between all the button son the site, but the core button block doesn't need all that CSS, we should let Global Styles handled that with the theme.json presets. The hover on the other hand, that's not being handled by the editor at all, and we do need the CSS there at the moment.

jffng:
> And also I think we are overriding too many of the styles we define on theme.json with our CSS:

Did you check out this PR https://github.com/WordPress/gutenberg/pull/34147 when testing this? It's needed to make sure that blockbase's CSS doesn't override global styles inline CSS.

> The hover on the other hand, that's not being handled by the editor at all, and we do need the CSS there at the moment.

If a user changes the button block colors globally, we do not have a way handling the hover state so it will look broken.

jffng:
I think this is ready for another review. It removes the button block CSS that was overriding what was already being set by theme.json. It does not require the Gutenberg ordering change above to solve the issue. 

**Testing instructions**

- Activate Blockbase
- Create a post / page and add a button with no styling
- Go to the site editor -> global styles panel and reset to defaults 
- Under Root > Color > Color Palette, change the primary color and verify the button's background color changes accordingly in the editor and view.
- Under By Block Type > Button > Color, change the Text and Background color. Verify the changes reflect in the editor and view

pbking:
Looks like we lost outline styles:
image

Was this prior:
image

pbking:
image

We seem to have lost control of the ability to style the search button.

jeyip:
## Testing
### Requirements
- [x] Per block global styles for the button block respect user customized styles
- [ ] Buttons within other blocks are styled as expected

### Browsers
- [x] Edge
- [x] Firefox
- [x] Safari
- [x] Chrome

### Notes
- I can confirm issues with search button block styling as noted by @pbking

pbking:
I reworked how this works a little bit and would appreciate another review @jeyip if you are so inclined.

This change doesn't effect the specificity at all, however it does remove from the button block styles anything controlled by Global Styles (leaving only hover state styles applied).

Additionally all "padding" stuff has been removed from the styles and theme.json.  I believe this was in place in order to match styles with non-block buttons, however that doesn't appear to be a useful thing after all.

The change seems to cover all of the points raised above with the benefit of simplifying the code in question.

pbking:
Seeing some problems with padding that I created with the change... working through that.

pbking:
Ok, dropped this down to the SMALLEST possible change to achieve the desired result.
Color and Typography styles are removed from the Theme's styles for the Button Block.
This allows COLOR and TYPOGRAPHY styles configured in Global Styles to be the only source of color and typography configuration for button BLOCKS.
Non-block buttons continue to be COLORED and .. uh .. TYPOGRAPHICALIZED from custom variables.

GLOBAL PADDING CONFIGURATION OF BUTTONS continues to NOT WORK.  Global configurations of padding will be overwritten by the theme due to how OUTLINE STYLES and other theme-specific stylistic concerns for outlines and hover states.

All Blockbase themes have been tested to work.
The `file` and `search` blocks have been evaluated to ensure that they continue to work as expected (meaning that they follow semantic colors, not button color configuration).

jffng (commented):
@pbking I think the golf change here is the right way to go while other issues in Gutenberg are discussed / worked out. 

I was unable to test in Skatepark or Quadrat because of Site Editor issues, but Blockbase is looking great.

I noticed a small issue with the search block's button missing a `cursor: pointer;`, but I don't think that should hold this one up.

🚢

jffng:
@jeyip 

>  Buttons within other blocks are styled as expected

Just want to make sure we're on the same page — at this time, changing the color of the button **block** in the global styles panel will not change the color of buttons _everywhere_ (e.g. buttons in forms, the search block, file block). This is not because of anything the theme is doing or can be expected to do. I do agree this is a gap in the experience and should be considered. That discussion is documented a bit here: pNEWy-ecm-p2 and here https://github.com/WordPress/gutenberg/issues/34206  

In other words, this PR shouldn't break or create inconsistencies in the default styling of buttons, but it also can't make button colors match when a user makes changes to the button block in the global styles panel.

What do you think? Is there anything else you imagine from this PR before bringing it in?

jffng:
I gave this one more spin, still working as expected. Going to bring this in as it's a pretty minimal change that solves the issue of global button color styles not being respected. If we need to look at it again we can, thanks everyone.

jeyip:
> I reworked how this works a little bit and would appreciate another review @jeyip if you are so inclined.

Super strange 🤔 Somehow I didn't see the pings. My mistake for missing them, and big thanks to @pbking + @jffng for taking care of this.

I'll be sure to catch up on the code changes and the P2 this afternoon.

jeyip:
> the global styles panel will not change the color of buttons everywhere

> What do you think? Is there anything else you imagine from this PR before bringing it in?

I reviewed the work, and I think what y'all have done makes a lot of sense. I'll be sure to communicate the limitations to #cylon.

> I do agree this is a gap in the experience and should be considered.

💯 Totally agree with this sentiment. The button block per-block global styling now behaves as expected, but the overall global styling experience itself could use improvement. Thanks for making the corresponding Github issue ⭐

PR #4443, opened 2021-08-19
Ignore Paxton

MaggieCabrera:

Added generated theme paxton to .gitignore so we don't pull it from the sandbox


mikachan (approved):
I saw you mention this on Slack, so that's my entire knowledge of this, but looks right to me!

PR #4444, opened 2021-08-19
Skatepark: query loop block pattern

MaggieCabrera:

As discussed on slack, this PR brings in the index.html query loop as a pattern since we didn't have any pattern for the block:

Screenshot 2021-08-19 at 17 39 33

### DIFF
--- a/skatepark/block-templates/index.html
+++ b/skatepark/block-templates/index.html
@@ -10,7 +10,7 @@
 	<hr class="wp-block-separator is-style-wide"/>
 	<!-- /wp:separator -->
 	
-	<!-- wp:post-featured-image {"style":{"color":{"duotone":["#000","#B9FB9C"]}}} /-->
+	<!-- wp:post-featured-image {"isLink":true,"style":{"color":{"duotone":["#000","#B9FB9C"]}}} /-->
 	
 	<!-- wp:post-title {"isLink":true,"fontSize":"normal"} /-->
 	
--- a/skatepark/block-templates/index.html
+++ b/skatepark/block-templates/index.html
@@ -5,8 +5,8 @@
 	<!-- wp:query-title {"type":"archive","align":"wide"} /-->
 	<!-- wp:query {"query":{"perPage":6,"pages":0,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"exclude","inherit":false},"displayLayout":{"type":"flex","columns":3},"align":"wide"} -->
 	<div class="wp-block-query alignwide"><!-- wp:post-template -->
-	<!-- wp:group {"style":{"spacing":{"padding":{"top":"30px","right":"30px","bottom":"30px","left":"30px"}}},"layout":{"inherit":false}} -->
-	<div class="wp-block-group" style="padding-top:30px;padding-right:30px;padding-bottom:30px;padding-left:30px"><!-- wp:separator {"className":"is-style-wide"} -->
+	<!-- wp:group {"style":{"spacing":{"padding":{"top":"30px","right":"30px","left":"30px"}}},"layout":{"inherit":false}} -->
+	<div class="wp-block-group" style="padding-top:30px;padding-right:30px;padding-left:30px"><!-- wp:separator {"className":"is-style-wide"} -->
 	<hr class="wp-block-separator is-style-wide"/>
 	<!-- /wp:separator -->
 	
--- a/skatepark/inc/patterns/blog-posts.php
+++ b/skatepark/inc/patterns/blog-posts.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * Blog posts.
+ *
+ * @package Skatepark
+ */
+
+return array(
+	'title'      => __( 'Blog posts', 'skatepark' ),
+	'categories' => array( 'skatepark' ),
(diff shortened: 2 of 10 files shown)


skatepark/inc/patterns/blog-posts.php
```diff
@@ -0,0 +1,36 @@
...
+	'title'      => __( 'Blog posts', 'skatepark' ),
+	'categories' => array( 'skatepark' ),
+	'content'    => '<!-- wp:query {"query":{"perPage":6,"pages":0,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"exclude","inherit":false},"displayLayout":{"type":"flex","columns":3},"align":"wide"} -->
+	<div class="wp-block-query alignwide"><!-- wp:post-template -->
+	<!-- wp:group {"style":{"spacing":{"padding":{"top":"30px","right":"30px","bottom":"30px","left":"30px"}}},"layout":{"inherit":false}} -->
+	<div class="wp-block-group" style="padding-top:30px;padding-right:30px;padding-bottom:30px;padding-left:30px"><!-- wp:separator {"className":"is-style-wide"} -->
+	<hr class="wp-block-separator is-style-wide"/>
+	<!-- /wp:separator -->
+	
+	<!-- wp:post-featured-image {"style":{"color":{"duotone":["#000","#B9FB9C"]}}} /-->
```
kjellr: We should make sure `isLink` is true for the featured images here (same for the actual template itself)

melchoyce:
What's up with the link placeholder alignment here? 👀 

image

MaggieCabrera:
Added image link to the pattern and the template and fixed the read more text (the text-indent from the except was being applied to the paragraph that contains the read more link)

mikachan (commented):
This is looking good on my side! The only things I can spot compared to the design is an underline on the dates, or maybe these should be links?

And should the 'read more' link have a heavier font-weight, maybe `500`?

Here's the design I'm looking at:

Screenshot 2021-08-20 at 12 24 14

kjellr:
Yeah, I'll second the notes from @mikachan. Also, the vertical spacing between all the post meta (featured image, title, excerpt, etc.) is all much greater than it is in the comps. Can we make that a little closer to the mockups? 

Comps|This PR
---|---
Screen Shot 2021-08-20 at 8 54 40 AM|Screen Shot 2021-08-20 at 8 54 54 AM

jffng:
> Also, the vertical spacing between all the post meta (featured image, title, excerpt, etc.) is all much greater than it is in the comps. Can we make that a little closer to the mockups?

I've noticed this too during most of my attempts to make a pattern in Skatepark, and I think it has to do with this: https://github.com/Automattic/themes/blob/trunk/skatepark/child-theme.json#L203

Should we reduce the value of this variable to something closer to the vertical spacing between the text in the comps?

MaggieCabrera:
I reduced the vertical spacing from 45 to 40px, I think it's more fitting to the design. The biggest problem about the spacing is that we were applying the same distance between the headings and the rest of the content and between every paragraph and the rest. According to figma, the distance between paragraphs should be 20px instead so I override that for Skatepark on the CSS and that makes the theme look like the design. 

I've created a [separate issue to support RTL](https://github.com/Automattic/themes/issues/4453) with the arrow so that we don't block this PR with that issue.

skatepark/inc/patterns/blog-posts.php
```diff
@@ -0,0 +1,36 @@
...
+	<!-- wp:group {"style":{"spacing":{"padding":{"top":"30px","right":"30px","left":"30px"}}},"layout":{"inherit":false}} -->
+	<div class="wp-block-group" style="padding-top:30px;padding-right:30px;padding-left:30px"><!-- wp:separator {"className":"is-style-wide"} -->
+	<hr class="wp-block-separator is-style-wide"/>
+	<!-- /wp:separator -->
+	
+	<!-- wp:post-featured-image {"isLink":true,"style":{"color":{"duotone":["#000","#B9FB9C"]}}} /-->
+	
+	<!-- wp:post-title {"isLink":true,"fontSize":"normal"} /-->
+	
+	<!-- wp:post-excerpt {"moreText":"Continue reading","className":"is-style-indented-post-excerpt"} /-->
```
scruffian: This text should be translated.

MaggieCabrera: Great catch thank you!

skatepark/block-templates/index.html
```diff
@@ -5,12 +5,12 @@
...
-	<div class="wp-block-group" style="padding-top:30px;padding-right:30px;padding-bottom:30px;padding-left:30px"><!-- wp:separator {"className":"is-style-wide"} -->
+	<!-- wp:group {"style":{"spacing":{"padding":{"top":"30px","right":"30px","left":"30px"}}},"layout":{"inherit":false}} -->
+	<div class="wp-block-group" style="padding-top:30px;padding-right:30px;padding-left:30px"><!-- wp:separator {"className":"is-style-wide"} -->
 	<hr class="wp-block-separator is-style-wide"/>
 	<!-- /wp:separator -->
 	
-	<!-- wp:post-featured-image {"style":{"color":{"duotone":["#000","#B9FB9C"]}}} /-->
+	<!-- wp:post-featured-image {"isLink":true,"style":{"color":{"duotone":["#000","#B9FB9C"]}}} /-->
 	
 	<!-- wp:post-title {"isLink":true,"fontSize":"normal"} /-->
```
scruffian, line 16: Should we add the same more text to this template?

MaggieCabrera: it's not translatable there...

scruffian: I know, but without it there's no link at all. I guess that's still better than a link that's not translated.

scruffian (approved):
LGTM

PR #4446, opened 2021-08-20
Seedlet Blocks: Fix block-level global styling for site title block

jeyip:
Site title block level global styling **text color changes** are being overridden because of a specificity conflict. To elaborate:

- Block level global styles are being applied to the site title block through `.wp-block-site-title a`
- The Seedlet blocks theme has a competing declaration via `.is-root-container a:not(.wp-block-button__link):not([download])`

I'm curious how folks feel about this PR. To keep the surface area of changes small, I tried piggybacking on the `:not` pseudo-selector convention. Rather than tinkering with increasing or decreasing specificities, I include the site title block as another element _not_ to select for the seedlet blocks selector. In doing so, the user customized global style always wins out.

I didn't see any obvious visual regressions.

## Related issue(s):
https://github.com/Automattic/themes/issues/4447

### DIFF
--- a/seedlet-blocks/assets/theme.css
+++ b/seedlet-blocks/assets/theme.css
@@ -291,7 +291,6 @@ is passed all of the block attributes on the block definition in the template. *
 
 .wp-block-site-title a:hover, .wp-block-site-title a:focus {
 	text-decoration: none;
-	color: var(--wp--custom--color--secondary);
 	background-size: 8px 0px;
 }
 
--- a/seedlet-blocks/assets/theme.css
+++ b/seedlet-blocks/assets/theme.css
@@ -140,26 +140,21 @@
 /**
  * Links in navigation
  */
+.wp-block-navigation-link a {
+	text-underline-offset: 0.4rem;
+}
+
 .wp-block-navigation-link a :hover {
 	text-decoration: underline;
-	text-underline-offset: 0.4rem;
 }
 
 /**
  * Links in Content
  */
-.is-root-container a:not(.wp-block-button__link):not([download]),
-.wp-block-post-content a:not(.wp-block-button__link):not([download]) {
-	color: var(--wp--custom--color--primary);
-	text-decoration-color: var(--wp--style--color--link);
+.is-root-container a,
+.wp-block-post-content a {
 	text-decoration-thickness: 1px;
 	text-underline-offset: 0.2em;
-}
-
-.is-root-container a:not(.wp-block-button__link):not([download]):focus, .is-root-container a:not(.wp-block-button__link):not([download]):hover,
-.wp-block-post-content a:not(.wp-block-button__link):not([download]):focus,
-.wp-block-post-content a:not(.wp-block-button__link):not([download]):hover {
-	color: var(--wp--style--color--link);
 }
 
 .wp-block-navigation__responsive-container.has-modal-open .wp-block-navigation-link {
@@ -291,9 +286,7 @@ is passed all of the block attributes on the block definition in the template. *
 .wp-block-site-title a {
 	text-decoration: underline;
 	text-decoration-thickness: 2px;
-	text-decoration-color: var(--wp--custom--color--secondary);
 	text-underline-offset: 0.13em;
-	transition: background-size 0.1s ease-out;
 }
 
 .wp-block-site-title a:hover, .wp-block-site-title a:focus {
--- a/seedlet-blocks/sass/blocks/_links.scss
+++ b/seedlet-blocks/sass/blocks/_links.scss
@@ -3,9 +3,9 @@
  */
 .wp-block-navigation-link {
 	a {
+		text-underline-offset: 0.4rem;
 		:hover{
 			text-decoration: underline;
-			text-underline-offset: 0.4rem;
 		}
 	}
 } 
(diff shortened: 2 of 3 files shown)


seedlet-blocks/assets/theme.css
```diff
@@ -148,7 +148,7 @@
 /**
  * Links in Content
  */
-.is-root-container a:not(.wp-block-button__link):not([download]),
+.is-root-container a:not(.wp-block-button__link):not([download]):not([aria-label="Site title text"]),
```
jeyip: **Note:**

Using an aria-label attribute as an identifier seems no bueno, but I couldn't think of a better way to handle the specificity conflict. I didn't see any other obvious ways of identifying the site title `a` tag.

pbking:
@jeyip changes made to Blockbase themes need to happen in the .scss file (there's not much CSS that doesn't go through the scss compiler).

I tinkered with this PR a big and would appreciate your eyes on it.

I also found that the `:not()`'s that were already there don't seem to be needed if we just take out the color attribute.  As far as I can tell that shouldn't have been there in the first place so I have removed it as well as cleaned out a few other related bits to clean that up.

I also found that the site-title in the content area was being handled differently than what's in a header template part.  Pay careful attention to that when taking a look at this change please.  I think it's good now.

I tested changing the colors of the site-title block (globally) as well as link color (globally) to make sure that it was doing what I expected it to do.

image

jeyip:
> changes made to Blockbase themes need to happen in the .scss file (there's not much CSS that doesn't go through the scss compiler).

Thank you! That makes total sense.

> I tinkered with this PR a big and would appreciate your eyes on it.

> I also found that the site-title in the content area was being handled differently than what's in a header template part. Pay careful attention to that when taking a look at this change please. I think it's good now.

> I tested changing the colors of the site-title block (globally) as well as link color (globally) to make sure that it was doing what I expected it to do.

Taking a look now. Thanks again for the guidance here 🙏

jeyip:
I saw one potential styling issue where the hover states of links are different from the site title. It's made more apparent if we set a global link color.



[We remove the color css declaration ](https://github.com/Automattic/themes/pull/4446/files#diff-ed0761ef8bd65beb4d0eb15cbe1ebb391e7bd92b9e5fd3e3d43574012a07f355L24-L27) in `_link.scss` on hover and focus. [Should we do the same](https://github.com/Automattic/themes/pull/4446/files#diff-04045fb947d5ddf2e26b19aeaf0e2b28e87a303cd958c46014ecc09188aafca8L10-L16) for `_site-title.scss`? Happy to make the changes for us if you agree. Besides that, your updates generally look good to me.

jeyip:
While testing, I also saw an issue unrelated to your changes. I noticed that the site title in the header has a hard-coded inline styled font-size, which is preventing us from modifying the block-level global styling for font size.

https://github.com/Automattic/themes/blob/1ca042362f230f0169732d36fec2115179420acb/seedlet-blocks/block-template-parts/header.html#L6-L8

**Edit:**
Nevermind! The unrelated issue I thought I saw here is actually working as expected. The hard-coded inline font size creates a block setting font size selection for site title. When a block setting competes with block-level global styles, it should _always_ override global styles.

jffng (commented):
I gave this a spin and here's what I found. Caveat that I am not sure what the expected behavior should be.

- Changing the global text color of the Site Title block does **not** change the color
- Changing the global link color of the Site Title block does change the color (as shown [here](https://user-images.githubusercontent.com/5414230/130218348-d1937964-8942-45ab-a2a0-4b2763c418c5.gif))
- The hover state for the site title does not respect the global color customizations. Maybe [this line](https://github.com/Automattic/themes/blob/trunk/seedlet-blocks/sass/blocks/_site-title.scss#L13) should be removed?

jeyip:
> - Changing the global text color of the Site Title block does not change the color
> - Changing the global link color of the Site Title block does change the color

@jffng From what I've seen, this brings the site title block to parity with other blocks. For example, the Post Comments Link Block and the Query Pagination Block behave similarly to what you're describing.

I do agree that the behavior doesn't seem ideal though. How do you feel about moving forward with this PR to get the site title block working and I can make an issue in core Gutenberg to discuss the global styles user experience further?

> - The hover state for the site title does not respect the global color customizations. Maybe this line should be removed?

Great minds think alike 🧠 

I saw the same issue and mentioned the same potential solution in [this comment](https://github.com/Automattic/themes/pull/4446#issuecomment-902883532). I'll go ahead and update the PR in a bit so that the behavior is more consistent.

jeyip:
> The hover state for the site title does not respect the global color customizations. Maybe this line should be removed?

Made the changes in 257494b

Lemme know what y'all think 👍

jffng (approved):
@jeyip I appreciate you working on this! I am a bit torn about the change. This PR fixes global color customizations for _all_ links, including the site title. So from that perspective the changes make sense.

However I don't like that we have to degrade the hover states a bit for all links. I don't think there's a way around this, since we do not yet have a way via global styles to handle hover states for buttons or links. So weighing those two things, I'm going to give a ✅  and hopefully there is a way to handle this sooner than later.

jeyip:
> I gave this a spin and here's what I found. Caveat that I am not sure what the expected behavior should be.

> Changing the global text color of the Site Title block does not change the color
> Changing the global link color of the Site Title block does change the color (as shown here)

Quick follow-up to this strange behavior @jffng. I created a PR that should address this. There was a bug where, even though blocks were explicitly disabling color and background color supports, they were still being displayed in the block-level global styles UI. https://github.com/WordPress/gutenberg/pull/34293

In the case of the Site Title block, text colors _should_ be disabled. https://github.com/WordPress/gutenberg/blob/599b60d2cc8b26b9bc2141305cfe5e07b23dab88/packages/block-library/src/site-title/block.json#L20-L23

When the aforementioned bug fix is merged, text color options should no longer be displayed in the UI.

PR #4449, opened 2021-08-20
Skatepark: Add columns in container block pattern

mikachan:
This PR adds the columns in container block pattern for Skatepark.

Screenshot 2021-08-23 at 15 54 17

(Mainly copied from the demo site, thank you!)

I've had a couple of issues with getting the alignment right, especially for the bottom paragraph. initially, with `"layout":{"inherit":true}`, I was seeing the whole container aligned to the left, but the bottom paragraph was correctly centered. With `"layout":{"inherit":false}`, the container would be centered but the paragraph would fill the width of the whole container, like this:

Screenshot 2021-08-20 at 16 29 58

For what it's worth, it looks perfect in combination with this PR: https://github.com/Automattic/themes/pull/4448

I'll go ahead and request a review, as it sounds like maybe this container could help debug the current alignment issues, or alternatively, they'll be magically fixed for this container from the other alignment work we have going on.

#### Related issue(s):
Part of #4308.


skatepark/inc/patterns/columns-in-container.php
```diff
@@ -0,0 +1,50 @@
...
+	<!-- /wp:heading -->
+
+	<!-- wp:paragraph {"className":"is-style-indented-paragraph"} -->
+	<p class="is-style-indented-paragraph">' . esc_html__( 'Skateboarding is a social sport! Come hang out and meet some new friends to cheer you on while you skate. We have a strict no-bullying policy and maintain a supportive environment.', 'skatepark' ) . '</p>
+	<!-- /wp:paragraph --></div>
+	<!-- /wp:column --></div>
+	<!-- /wp:columns -->
+
+	<!-- wp:image {"align":"wide","id":26,"sizeSlug":"large","linkDestination":"none","style":{"color":{"duotone":["#000","#B9FB9C"]}}} -->
+	<figure class="wp-block-image alignwide size-large"><img src="https://skateparkdemo.files.wordpress.com/2021/08/693372530767c766f2db45bbfb132770-2048x1365-1.jpeg?w=1024" alt="Close-up of a person riding a skateboard, focusing on their feet and the board. One foot is on the board, while the other foot is up, in motion. A skatepark is blurred in the background." class="wp-image-26"/></figure>
```
scruffian: This text needs i18n

melchoyce:
For some reason I'm seeing this off-center:

image

Can we update the spacing to match https://github.com/Automattic/themes/pull/4403? Though the headings are also adding some extra spacing... 😕  I mostly just want the two patterns to be even.

MaggieCabrera:
> For some reason I'm seeing this off-center:
> 

Yep, it's a bug, we are handling that over at https://github.com/Automattic/themes/issues/4426

mikachan:
> Can we update the spacing to match #4403? Though the headings are also adding some extra spacing... 😕 I mostly just want the two patterns to be even.

Yep sure, I think the main difference is the vertical spacing from the headings like you say. I've used the same styles to shrink the heading margins for this pattern. I think this looks better but let me know what you think.

And yeah.. Maggie's already mentioned but the off-centered issue will hopefully be fixed alongside the other alignment work. If you remove the `wp-container-*` class from the container for this pattern it will be centered, but the last paragraph then incorrectly fills the width of the container (which is the part that should be fixed in the other issue).

melchoyce:
Cool, thanks for the info 👍 

I think if we make that top padding 3.5%, we're good with the container.

I noticed this is pretty tight:

image

Maybe a 60px spacer block would help there?

This could also use a little bit more space, to make it about ~45px between the image and paragraph:

image

Dunno if that one would be easier to do with a spacer or css, your call 🤷

skatepark/inc/patterns/columns-in-container.php
```diff
@@ -0,0 +1,46 @@
...
+
+return array(
+	'title'      => __( 'Columns in container', 'skatepark' ),
+	'categories' => array( 'skatepark' ),
+	'content'    => '<!-- wp:spacer {"height":40} -->
+	<div style="height:40px" aria-hidden="true" class="wp-block-spacer"></div>
+	<!-- /wp:spacer -->
+ 
+	<!-- wp:group {"align":"wide","style":{"border":{"width":"3px","style":"solid"},"spacing":{"padding":{"top":"3%","right":"4%","bottom":"4%","left":"4%"}}},"borderColor":"primary","className":"columns-in-container","layout":{"inherit":true}} -->
+	<div class="wp-block-group alignwide columns-in-container has-border-color has-primary-border-color" style="border-style:solid;border-width:3px;padding-top:3%;padding-right:4%;padding-bottom:4%;padding-left:4%"><!-- wp:columns {"align":"wide"} -->
```
scruffian: I think when https://github.com/Automattic/themes/pull/4436 merges we can make the vertical and horizontal paddings the same.

skatepark/sass/block-patterns/_mixed-media-in-container.scss
```diff
@@ -1,5 +1,6 @@
-.skatepark-mixed-media-in-container {
-	hr, p, h1, h2, h3, h4, h5, h6 { // Shrink default vertical margins for all the elements in this pattern
+.skatepark-mixed-media-in-container,
+.columns-in-container {
+	hr, p, h1, h2, h3, h4, h5, h6 { // Shrink default vertical margins for all the elements in these patterns
```
scruffian: I'm not very keen on this. It makes it harder users to reproduce the same layouts without patterns and seems like it would create inconsistency between different parts of the site. Also, there are suggestions to add spacing back to this pattern under the image. If we didn't have these rules then I think we wouldn't need to add the extra spacing

mikachan: Yeah you're right, if we remove these smaller margin overrides then the space between the paragraphs and the image is correct/looks better (I think - at 45px). However, this means the h4 headings also default to the 45px vertical margin, which is possibly too much for this pattern.

mikachan:
Thanks all for your help so far. I think this is where this is up to:

- The alignment fix is now being applied to this, woop!
- I've removed the margin overrides so the default `45px` vertical spacing is used, which increases the spacing between the paragraphs and the image (which I think is correct). However, it also increases the spacing below the h4 headings. To fix this, I've added another override, but only for paragraphs after h4 headings. I know this still means the user would struggle to replicate this for another pattern though, so I'm not 100% sold (I just wanted to offer another option!)
- The bottom margin of the last paragraph makes the top and bottom container spacing look different, so I've added an override for this as well, which reduces the vertical margin of the last element in this pattern. This has the same issue as above though, where the user would struggle to replicate this elsewhere.

I'm not sure on the overrides here but I also didn't want to suggest something that was too different from the design. I think not having these margin overrides at all is another option (i.e. if you comment out the `_columns-in-container` scss file completely), but it does mean the spacing is bigger than in the design.

skatepark/sass/block-patterns/_mixed-media-in-container.scss
scruffian: @MaggieCabrera what do you think?

scruffian:
IMO it doesn't make sense to have the overrides. Maybe we should reduce margins for the whole site, so that they match?

mikachan:
> IMO it doesn't make sense to have the overrides. Maybe we should reduce margins for the whole site, so that they match?

I was also thinking this, I'm not sure if we're just struggling with the `45px` in general as the default vertical margin. @MaggieCabrera & @melchoyce, what do you think?

MaggieCabrera:
> I was also thinking this, I'm not sure if we're just struggling with the 45px in general as the default vertical margin. @MaggieCabrera & @melchoyce, what do you think?

About this, I think we've mentioned this in another PR. Initially, the spacing was set to 45px for both p tags and headings (this is a rule in Blockbase, that applies the spacing variable as margin equally to p's and hx's). I changed that to 40px (it's what I saw consistently all over the design, instead of 45px) and half of that for the paragraphs (that was also the case of what I saw all over the design files).

I just saw that this change https://github.com/Automattic/themes/commit/45bb8e35e0977ff82c03a47b72a691d49fc201a8#diff-fd9394d7932810a7f4ad78b09ac7112b9c5cb96caea41dd78ec4fee4df3d5730 undoes what I did, bringing us back to 40px margins being inherited from Blockbase...

I think in this case, this pattern is being an exception to what all other headings were doing in other parts of the design, so I'd defer to @melchoyce judgment.

scruffian:
> I just saw that this change 45bb8e3#diff-fd9394d7932810a7f4ad78b09ac7112b9c5cb96caea41dd78ec4fee4df3d5730 undoes what I did, bringing us back to 40px margins being inherited from Blockbase...

Oh sorry looks like we need to bring that back in. I thought it was there for another reason.

mikachan:
I've added @MaggieCabrera's paragraph spacing change back in (is that OK in this PR?), removed the CSS overrides for this pattern and added some spacer blocks to increase the spacing around the image (as per @melchoyce's suggestion). 

Ready for another review!

jffng (approved):
This is looking good to me, I like that we don't need the pattern-specific margin exceptions. I'll do that for the "[mixed media in container](https://github.com/Automattic/themes/blob/trunk/skatepark/sass/block-patterns/_mixed-media-in-container.scss)" pattern as a follow up.

melchoyce (approved):
:shipit:

skatepark/sass/base/_text.scss
```diff
@@ -2,3 +2,8 @@
 ::selection {
 	color: var(--wp--preset--color--background);
 }
+
+p {
```
scruffian, line 6: @MaggieCabrera we added it back!

PR #4454, opened 2021-08-23
Blockbase: Remove alignment styles

scruffian:
This is an alternate approach to https://github.com/Automattic/themes/pull/4448

Gutenberg has changed so that now all group blocks have the `wp-container-*` class. This means that we can no longer rely on this class to target elements that are meant to "inherit layout".

This PR removes all the alignment styles that are trying to change the way that Gutenberg handles alignments. It has become very difficult to modify the default behaviour in the theme.

The one case that I think we should try to handle is this: https://github.com/Automattic/themes/issues/3747

The problem with adding padding to paragraph blocks inside `alignfull` elements is that you can add `alignfull` to elements that "inherit layout", which means that they no longer align to the edge of the layout container.

So instead of setting padding on these elements we can set a max width on this. However the problem with this approach is that elements inside "inherit layout" containers also use `max-width` to constrain them to the correct column size.

So instead this PR proposes that we use `width` to constrain all paragraphs. This will allow "inherit layout" to continue behaving as before, but it will ensure that paragraph blocks don't ever touch the edge of the screen.

To test you can use the markup in https://github.com/Automattic/themes/pull/4448 and https://github.com/Automattic/themes/issues/4412.

### DIFF
--- a/blockbase/sass/base/_alignment.scss
+++ b/blockbase/sass/base/_alignment.scss
@@ -1,17 +1,10 @@
-.wp-block-group.alignfull,
-*[class*="wp-container-"] //Anything that inherits layout (container)
-{
-	//give it some padding
-	padding-left: var(--wp--custom--post-content--padding--left);
-	padding-right: var(--wp--custom--post-content--padding--right);
-
-	//Any nested containers, and anything that is alignfull
-	*[class*="wp-container-"], // Any nested containers
-	> .alignfull { // Any direct descendant that is alignfull
-		// bust out of the container's padding
-		margin-left: calc(-1 * var(--wp--custom--post-content--padding--left)) !important;
-		margin-right: calc(-1 * var(--wp--custom--post-content--padding--right)) !important;
-		width: calc( 100% + var(--wp--custom--post-content--padding--left) + var(--wp--custom--post-content--padding--right) ) !important;
+[data-align="full"],
+.alignfull {
+	> p {
+		//give it some padding
+		margin-left: auto;
+		margin-right: auto;
+		width: calc( 100vw - var(--wp--custom--post-content--padding--left) - var(--wp--custom--post-content--padding--right) );
 	}
 }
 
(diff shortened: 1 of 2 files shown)


MaggieCabrera:
I tried this one with the same demo content I pasted over at #4448 and this is what I get:

Screenshot 2021-08-23 at 15-56-59 Alignments – free
Screenshot 2021-08-23 at 15-56-46 Alignments – free

blockbase/sass/base/_alignment.scss
```diff
@@ -1,18 +1,8 @@
...
-		// bust out of the container's padding
-		margin-left: calc(-1 * var(--wp--custom--post-content--padding--left)) !important;
-		margin-right: calc(-1 * var(--wp--custom--post-content--padding--right)) !important;
-		width: calc( 100% + var(--wp--custom--post-content--padding--left) + var(--wp--custom--post-content--padding--right) ) !important;
-	}
+p {
+	// An alternative to padding.
+	margin-left: auto;
+	margin-right: auto;
+	width: calc( 100vw - var(--wp--custom--post-content--padding--left) - var(--wp--custom--post-content--padding--right) );
```
jffng, line 5: Setting a width here causes paragraphs to sometime overflow out of their containers:

Screen Shot 2021-08-23 at 2 28 56 PM

jffng (commented):
> Gutenberg has changed so that now all group blocks have the wp-container-* class. This means that we can no longer rely on this class to target elements that are meant to "inherit layout".

Especially given this ^, I am generally in favor of removing alignment styles from Blockbase.
 
That said, I don't think the width approach will work because of the comment I made below. I tested with this content: https://theamdemo.wordpress.com/2021/08/23/gutenberg-alignment/

blockbase/sass/base/_alignment.scss
scruffian: yeah that makes sense. i think we should just remove it all

pbking:
Closing in lieu of #4459

PR #4455, opened 2021-08-23
Lodestar: Removing italics in verse block to respect the block toolbar

### LINKED ISSUE #4363: Lodestar: Verse Block is always in Italics
## Steps to replicate

 On the Lodestar theme, create a post with a verse block and add text.

Screenshot on 2021-08-03 at 18-37-25

## Result

It will be in italics. There seems to be styling in the theme's CSS to make the verse block always italic:

```
.wp-block-verse {
	background: 0 0;
	font-family: inherit;
	font-style: italic;
	padding: 0
}

```
## Expected

In other themes, Twenty-Twenty-One for instance, the verse block does not have a default style, so a user can choose to make the text italic with the block settings as they want.

* User report: [30670203-hc]
* Follow-up ticket: [4196336-zen]

aetta:
Removed font-style: italic from:

blocks.css 
editor-blocks.css

This also fix #4363


MaggieCabrera:
This will change users' sites who may be happy with the italic version of the block. Can we find a solution that provides italic as default but without overriding the user selection from the toolbar?

aetta:
Hi @MaggieCabrera, that makes sense! A possible solution would be to use the emphasis element <em>, adding the font-face: italic; to it, and in another verse block, the font-face: normal;

Another possible solution is to change the font-face of the block by the element inherited from the parent.

Both should show the same result!

**1**
```
.wp-block-verse em {
	background: transparent;
	font-family: inherit;
	font-style: italic;
	padding: 0;
}

.wp-block-verse {
	background: transparent;
	font-family: inherit;
	font-style: normal;
	padding: 0;
}
```

**2**

```
.wp-block-verse {
	background: transparent;
	font-family: inherit;
	font-style: inherit;
	padding: 0;
}
```

What do you think? :)

MaggieCabrera:
> Hi @MaggieCabrera, that makes sense! A possible solution would be to use the emphasis element _, adding the font-face: italic; to it, and in another verse block, the font-face: normal;_
> _ Another possible solution is to change the font-face of the block by the element inherited from the parent. Both should show the same result! **1** ``` .wp-block-verse em { background: transparent; font-family: inherit; font-style: italic; padding: 0; } .wp-block-verse { background: transparent; font-family: inherit; font-style: normal; padding: 0; } ``` **2** ``` .wp-block-verse { background: transparent; font-family: inherit; font-style: inherit; padding: 0; } ``` What do you think? :)_

The first solution requires the user to make a change to achieve what the theme was doing by default. We want to try and not affect user's existing sites, in case they have this in multiple places, so they don't have to edit all of their posts to revert our change.

I'm not sure if the second solution would work, since I'm not sure if the parent has the correct rules to inherit from? I'd have to test that

PR #4461, opened 2021-08-23
fix issue #4460

### LINKED ISSUE #4460: Blockbase: undefined variable location in wp-customize-fonts.php
## Steps to replicate

Creating a child theme, setting typography fontFamily to `var(--wp--preset--font-family--spartan)` (a font not present in the default Google font list) 

## Result

`Undefined variable: location in /xxx/xxx/wp-content/themes/blockbase/inc/customizer/wp-customize-fonts.php on line 321`

## Expected

Unused $location variable should not cause a notice.

RavanH:

Remove undefined variable $location

#### Related issue(s):

Fixes Automattic#4460


scruffian:
Thanks for the fix. It does work, but I think we can remove $location altogether: https://github.com/Automattic/themes/pull/4621

RavanH:
Yes, I know :) I just didn't choose to remove it completely from build_font_from_theme_data because of any child theme incompatibilities or future plans for this mysterious "location" parameter ;)

I can adapt the PR to completely get rid of the parameter in build_font_from_theme_data all together, or make it backward compatible. Just did't want to go and do that without more information...

scruffian:
Thanks. Does #4621 solve it?

RavanH:
I added a little comment at line 341 but other than that it should indeed fix the undefined variable notice :)

RavanH:
As long as no other child themes call build_font_from_theme_data() directly, expecting three parameters instead of two...

RavanH:
To consider: I was actually going to propose dropping the whole "fontsToLoadFromGoogle" thing all together as it looks a bit hacky to me. Any font that is needed in a child theme can just be added to the main "fontFamilies" with an extra "google" variable if needed to be loaded from Google, just like the others.

The much simpler code would then be what I have running on my local dev fork of Blockbase:

```
	function get_font_family( $array, $configuration ) {
		$variable = get_settings_array( $array, $configuration );
		$slug     = preg_replace( '/var\(--wp--preset--font-family--(.*)\)/', '$1', $variable );
		if ( ! isset( $this->fonts[ $slug ] ) ) {
			$this->fonts[ $slug ] = $this->build_font_from_theme_data( $slug, $configuration );
		}
		return $this->fonts[ $slug ];
	}

	function build_font_from_theme_data( $slug, $configuration ) {
		$new_font      = array();
		$font_families = $configuration['settings']['typography']['fontFamilies']['theme'];
		foreach ( $font_families as $font_family ) {
			if ( $font_family['slug'] === $slug ) {
				$new_font['fontFamily'] = $font_family['fontFamily'];
				$new_font['name']       = $font_family['name'];
				if ( ! empty( $font_family['google'] ) ) {
					$new_font['google'] = $font_family['google'];
				}
			}
		}
		$new_font['slug'] = $slug;
		return $new_font;
	}
```

The only thing is that this would need to be communicated to all child theme devs before merging as some themes rely on "fontsToLoadFromGoogle" ... Not sure if there is room for tweaks like this at this point. Maybe others like @pbking and @MaggieCabrera  could say more about that? :)

RavanH:
See PR #4633

scruffian:
Closing in favour if https://github.com/Automattic/themes/pull/4633

PR #4459, opened 2021-08-23
Another attempt at alignments

pbking:
WIP...

PR #4464, opened 2021-08-24
Skatepark: apply wide width to header

### LINKED ISSUE #4407: Skatepark: Header should max out at the wide-width value
## Steps to replicate

Look at this beautiful header:

image

## Result

It is nice, but very wide.

## Expected

It should max out at the wide-width value. So, a max of 1194px on desktop, with no margin added:

image

We should keep the existing margins on smaller screens though :+1:

jffng:

This PR adds some styles to ensure the Skatepark header does not extend beyond the wide width. Now that #4459 has merged, this is ready for a review.

#### Related issue(s):
Closes #4407

### DIFF
--- a/skatepark/block-template-parts/header.html
+++ b/skatepark/block-template-parts/header.html
@@ -1,4 +1,4 @@
-<!-- wp:group { "layout":{"type":"flex"},"align":"full"} -->
+<!-- wp:group { "layout":{"type":"flex"}} -->
 <div class="wp-block-group">
 <!-- wp:group -->
 <div class="wp-block-group">
--- a/skatepark/assets/theme.css
+++ b/skatepark/assets/theme.css
@@ -533,6 +533,12 @@ a:not(.ab-item):not(.screen-reader-shortcut):focus {
 	mask-image: url(svg/post-tag.svg);
 }
 
+header.wp-block-template-part {
+	max-width: var(--wp--custom--layout--wide-size);
+	margin: 0 auto;
+	width: 100%;
+}
+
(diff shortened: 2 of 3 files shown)


jffng:
I tested this with #4448 or #4459, and both work :)

skatepark/block-template-parts/header.html
```diff
@@ -1,4 +1,4 @@
-<!-- wp:group { "layout":{"type":"flex"},"align":"full"} -->
+<!-- wp:group { "layout":{"type":"flex"}} -->
```
jffng, line 1: Layouts can be flex OR flow/default (inherit), not both. So align full has no purpose in this context.

pbking: According to the comment on #4407 the header should be `"align":"wide"` and, since as you point out that can't be done in one group I think that means another group is needed to wrap the entirety of the header.

pbking (approved):
This looks OK.
I think we could do the same with another group wrapping the header.  I don't believe either is wrong but wrapping with a "full width" group would be "reproducable" in the template editor.  With the state of flex controls (currently lacking) I don't think that matters now. Once flex is more powerful this will need to be changed again then and can be re-evaluated then.

melchoyce (approved):

jffng:
> I think we could do the same with another group wrapping the header. 

I opened a PR to try that idea https://github.com/Automattic/themes/pull/4466, it works but yeah I'm not sure which one is preferred. Going to merge this one for now to get the design in place.

PR #4465, opened 2021-08-24
Skatepark: Add custom padding to cover block

### LINKED ISSUE #4413: Skatepark: Adjust spacing in cover blocks
## Steps to replicate

Copy this into your editor:

```
<!-- wp:cover {"url":"http://themes.local/wp-content/uploads/2021/07/guy-man-FSXSYPD4ZV.png","id":11,"minHeight":100,"minHeightUnit":"vh","contentPosition":"bottom left","align":"full"} -->
<div class="wp-block-cover alignfull has-background-dim has-custom-content-position is-position-bottom-left" style="min-height:100vh"><img class="wp-block-cover__image-background wp-image-11" alt="" src="https://cldup.com/ClNs4svA5W.png" data-object-fit="cover"/><div class="wp-block-cover__inner-container"><!-- wp:heading {"textColor":"background","className":"is-style-default"} -->
<h2 class="is-style-default has-background-color has-text-color">Ready to skate?</h2>
<!-- /wp:heading -->

<!-- wp:buttons -->
<div class="wp-block-buttons"><!-- wp:button -->
<div class="wp-block-button"><a class="wp-block-button__link">Sign up</a></div>
<!-- /wp:button --></div>
<!-- /wp:buttons --></div></div>
<!-- /wp:cover -->
```

## Result

image

## Expected

The spacing to be similar to this:

image

Can we somehow limit that inner content to be constrained to the wide-width by default?

mikachan:
This PR adds custom padding to the cover block for Skatepark:

Screenshot 2021-08-24 at 17 41 23

I've tried this using `clamp()`, mainly so the padding isn't too large at smaller resolutions. But we could also use a `calc()` based on one of the spacing settings maybe, or use a percentage?

#### Related issue(s):
Closes #4413.

### DIFF
--- a/skatepark/sass/blocks/_cover.scss
+++ b/skatepark/sass/blocks/_cover.scss
@@ -1,17 +1,10 @@
 .wp-block-cover {
-	.wp-block-cover__inner-container {
-		width: auto;
-	}
-
-	&.has-custom-content-position {
-		&.is-position-top-left, &.is-position-top-center, &.is-position-top-right {
-			padding-top: 5.75%;
-			padding-bottom: 17.25%;
-		}
-
-		&.is-position-bottom-left, &.is-position-bottom-center, &.is-position-bottom-right {
-			padding-top: 17.25%;
-			padding-bottom: 5.75%;
+	&.alignfull {
+		padding-left: 0;
+		padding-right: 0;
+		.wp-block-cover__inner-container {
+			padding-left: calc( max( 0.5 * ( 100% - var(--wp--custom--layout--wide-size) ), var(--wp--custom--post-content--padding--left) ));
+			padding-right: calc( max( 0.5 * ( 100% - var(--wp--custom--layout--wide-size) ), var(--wp--custom--post-content--padding--left) ));
 		}
 	}
 }
--- a/skatepark/sass/blocks/_cover.scss
+++ b/skatepark/sass/blocks/_cover.scss
@@ -1,3 +1,0 @@
-.wp-block-cover {
-	padding: clamp(2em, 15%, 8em);
-}
--- a/skatepark/sass/blocks/_cover.scss
+++ b/skatepark/sass/blocks/_cover.scss
@@ -1,0 +1,19 @@
+.wp-block-cover {
+	padding: 11.5%;
+
+	.wp-block-cover__inner-container {
+		width: auto;
+	}
+
+	&.has-custom-content-position {
+		&.is-position-top-left, &.is-position-top-center, &.is-position-top-right {
+			padding-top: 5.75%;
+			padding-bottom: 17.25%;
+		}
+
+		&.is-position-bottom-left, &.is-position-bottom-center, &.is-position-bottom-right {
+			padding-top: 17.25%;
+			padding-bottom: 5.75%;
+		}
+	}
+}
--- a/skatepark/sass/blocks/_cover.scss
+++ b/skatepark/sass/blocks/_cover.scss
@@ -1,6 +1,6 @@
 .wp-block-cover {
 	&.alignfull {
-		padding-left: calc( max( 0.5 * ( 100% - var(--wp--custom--layout--wide-size) ) + var(--wp--custom--post-content--padding--left), var(--wp--custom--post-content--padding--left) ));
+		padding-right: calc( max( 0.5 * ( 100% - var(--wp--custom--layout--wide-size) ) + var(--wp--custom--post-content--padding--left), var(--wp--custom--post-content--padding--left) ));
 		padding-left: calc( max( 0.5 * ( 100% - var(--wp--custom--layout--wide-size) ) + var(--wp--custom--post-content--padding--left), var(--wp--custom--post-content--padding--left) ));
 	}
 }
--- a/skatepark/sass/blocks/_cover.scss
+++ b/skatepark/sass/blocks/_cover.scss
@@ -1,0 +1,3 @@
+.wp-block-cover {
+	padding: clamp(2em, 15%, 8em);
+}
--- a/skatepark/sass/blocks/_cover.scss
+++ b/skatepark/sass/blocks/_cover.scss
@@ -1,10 +1,6 @@
 .wp-block-cover {
 	&.alignfull {
-		padding-left: 0;
-		padding-right: 0;
-		.wp-block-cover__inner-container {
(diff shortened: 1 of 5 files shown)


skatepark/sass/blocks/_cover.scss
```diff
@@ -0,0 +1,3 @@
+.wp-block-cover {
+	padding: clamp(2em, 15%, 8em);
+}
```
MaggieCabrera: I think we can do this on theme.json with something like 

```
"core/cover": {
  "spacing": {
	  "padding": {
		  "left": "clamp(2em, 15%, 8em)"
...
	  }
  }
},
```

mikachan: Oh cool, thanks! Will give this a try.

mikachan:
@melchoyce, you asked:

> Can we somehow limit that inner content to be constrained to the wide-width by default?

I've just been searching for a way to do this, and it looks like it's only possible by adding some JS that runs while using the block editor. Have we done something like this before, or am I possibly over-complicating this and there is an easy way to set a default block alignment?

MaggieCabrera:
> @melchoyce, you asked:
> 
> > Can we somehow limit that inner content to be constrained to the wide-width by default?
> 
> I've just been searching for a way to do this, and it looks like it's only possible by adding some JS that runs while using the block editor. Have we done something like this before, or am I possibly over-complicating this and there is an easy way to set a default block alignment?

yeah, I think JS is overkill. I managed this without any CSS:

Screenshot 2021-08-26 at 17 47 08

Applying the cover block positioning breaks the possibility of using the wide alignment. Maybe using this markup makes more sense and we should just add some block styles to align it to the bottom?

```
<!-- wp:cover {"url":"http://free.local/wp-content/uploads/2020/03/633e5-calm.jpg","id":647,"minHeight":100,"minHeightUnit":"vh","contentPosition":"center center","align":"full"} -->
<div class="wp-block-cover alignfull has-background-dim" style="min-height:100vh"><img class="wp-block-cover__image-background wp-image-647" alt="" src="http://free.local/wp-content/uploads/2020/03/633e5-calm.jpg" data-object-fit="cover"/><div class="wp-block-cover__inner-container"><!-- wp:group {"layout":{"inherit":true}} -->
<div class="wp-block-group"><!-- wp:group {"align":"wide"} -->
<div class="wp-block-group alignwide"><!-- wp:heading {"textColor":"background","className":"is-style-default"} -->
<h2 class="is-style-default has-background-color has-text-color">Ready to skate?</h2>
<!-- /wp:heading -->

<!-- wp:buttons -->
<div class="wp-block-buttons"><!-- wp:button -->
<div class="wp-block-button"><a class="wp-block-button__link">Sign up</a></div>
<!-- /wp:button --></div>
<!-- /wp:buttons --></div>
<!-- /wp:group --></div>
<!-- /wp:group --></div></div>
<!-- /wp:cover -->
```

MaggieCabrera:
A quick try with the inspector says `margin-top: auto` to `wp-block-cover__inner-container` does the trick

MaggieCabrera:
I think this could be another block pattern since the markup is a bit more involved than a simple cover block with some content in it.

melchoyce:
I'd like to stick to Gutenberg defaults as much as we can so it's as easy as possible to rebuild with or without patterns. Since it sounds like constraining the inner width might introduce other issues, maybe we should ignore the mockup here and go with the default behavior 😕 

@kjellr Would love to hear your thoughts on this when you're back.

skatepark/sass/blocks/_cover.scss
scruffian: I like the idea of keeping it simple with a percentage. Did you try that?

mikachan: Yeah, maybe clamp is getting too fancy! I tried `11.5%` but I wasn't sure if it looked too big at some resolutions (tablet size-ish and super large desktop), but looking at it again it's probably OK.

kjellr:
Yeah, I agree that we should use the defaults here.

mikachan:
Thanks all for your help with this. I've just changed the cover block padding value to `min(8em, 11.5%)` rather than `clamp(2em, 15%, 8em)`.  Just `11.5%` on its own works for everywhere other than 1600px+ ish resolutions, where I thought it looked too big - but this may just be me, happy to change to just `11.5%`.

Do we need to do anything else for this PR?

melchoyce:
Dunno if it's just me, but the positioning setting isn't really working:

https://user-images.githubusercontent.com/2846578/131547426-66040dcd-d2c2-4c9b-a8df-e3608037936b.mp4

mikachan:
> Dunno if it's just me, but the positioning setting isn't really working:

It's not just you, I can see this as well! It's because I've applied the padding equally on all sides. I also didn't realise the position option even existed 🙈  I'll look at this today.

mikachan:
I've improved the different position settings by applying less padding to the side of the alignment (i.e. if it's aligned to the bottom, I've reduced the bottom padding). This seems to work but I don't know if I've over-complicated this, or it's maybe not the best way to do it. Let me know what you think!

melchoyce:
Cool, looking really good 👍 

And just double-checking — there's no way to get these edges to line up consistently, right? I'd proposed a max-width earlier but now I'm thinking about padding... but keeping that consistent across screen sizes would probably make that impossible 🤷 

image

If it's not possible, this one's ready from a design point of view.

MaggieCabrera:
I rebased this and had another go to implement this the same way we are centering the caption for the block pattern. I made this change to affect only the alignfull covers, since the wide or regular aligned ones were looking a bit strange with the existing CSS. This is ready for another review. This is how I tested this:

Screenshot 2021-09-03 at 12-04-58 cover block tests – Skatepark


And our regular cover block test page:

Screenshot 2021-09-03 at 12-06-55 Gutenberg Cover – Skatepark

MaggieCabrera:
Nevermind I found some more issues with this, I'm working on it!

MaggieCabrera:
So I fixed a couple of issues for this. One of the problems that we had using % units is that the horizontal and vertical values of the padding weren't consistent over all the cover variations. Another issue we have is that we were using a very high value of padding because we were assuming the cover was always going to be full viewport height and that makes it look bad when that option is not selected.

I changed to use vh units for this to make it more or less consistent but since the editor doesn't use a class for the "full viewport height" option, we can't have a specific larger padding for that option. I went for an option that would look best for all the possible options of the cover, but I'm happy to tweak as needed, @melchoyce:

Screenshot 2021-09-03 at 12-18-56 cover block tests – Skatepark

melchoyce:
LGTM!

PR #4467, opened 2021-08-25
Blockbase + co: Universal Social menu

### LINKED ISSUE #4378: Skatepark: Social Menu is not universally customizable
Screenshot 2021-08-10 at 16 50 42

The Social menu is a block and can't be edited via the customizer right now like the menu can. We should either remove it from the time being or make a render callback function that lets us add items to the block from what's being edited in the customizer.

MaggieCabrera:

The social link block can't be edited via a customizer only user experience. If our themes (namely Skatepark) include this block on the header, a user with no access to the FSE can't add their social links to it. Here I'm using the block's render callback to inject the social menu from the customizer if it's selected, else it renders the default block's content.

To test this:

- Go to Customizer > menus and add a menu to "Social Navigation"
- Check that the menu is changed to the correct items both in the customizer preview and the frontend.
- Check that when no menu is assigned to "Social Navigation", the default content from the block is shown instead.

Screenshot 2021-08-25 at 10 43 38
Screenshot 2021-08-25 at 10 43 29

A few comments on this:

- This is being built on Blockbase and I'm using skatepark specific settings for it, not all themes will want the block to have it justified right or show in primary color. This is a first pass, I'm not sure what's the best way to customize this
- This doesn't allow for multiple variations of the block. In the case of Skatepark we have it on two places, header and footer block pattern. The footer one is not aligned to the right, I'm not sure how we can account for that.

#### Related issue(s):

Closes #4378

### DIFF
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -162,7 +162,8 @@ function blockbase_social_menu_render( $block_content, $block ) {
 		$block_content = '<!-- wp:social-links {"iconColor":"primary","iconColorValue":"var(--wp--custom--color--primary)","className":"' . $class_name . '"} --><ul class="wp-block-social-links has-icon-color ' . $class_name . '">';
 		$menu = wp_get_nav_menu_items( $social_menu_id );
 		foreach ($menu as $menu_item) {
-			$block_content .= '<!-- wp:social-link {"url":"' . $menu_item->url . '","service":"' . $menu_item->post_name . '"} /-->';
+			$service_name = preg_replace( '/(-[0-9]+)/', '', $menu_item->post_name );
+			$block_content .= '<!-- wp:social-link {"url":"' . $menu_item->url . '","service":"' . $service_name . '"} /-->';
 		}
 
 		$block_content .= '</ul>';
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -131,9 +131,9 @@ function blockbase_social_menu_render( $block_content, $block ) {
 		return $block_content;
 	}
 
-	$return = $block_content;
-
-	if ( has_nav_menu( 'social' ) ) :
+	if ( ! has_nav_menu( 'social' ) ) :
+		return $block_content;
+	else:
 
 		$return  = '<ul class="wp-block-social-links has-icon-color items-justified-right is-style-logos-only">';
 		
@@ -145,7 +145,8 @@ function blockbase_social_menu_render( $block_content, $block ) {
 		}
 
 		$return .= '</ul>';
+		
+		return $return;
 	endif; 
 
-	return $return;
 }
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -156,8 +156,11 @@ function blockbase_social_menu_render( $block_content, $block ) {
 	if ( blockbase_condition_to_render_social_menu( $block ) ) {
 		$nav_menu_locations = get_nav_menu_locations();
 		$social_menu_id = $nav_menu_locations['social'];
-
-		$block_content = '<!-- wp:social-links {"iconColor":"primary","iconColorValue":"var(--wp--custom--color--primary)","className":"items-justified-right is-style-logos-only"} --><ul class="wp-block-social-links has-icon-color items-justified-right is-style-logos-only">';
+		$class_name = 'is-style-logos-only';
+		if( !empty( $block['attrs']['itemsJustification'] ) && $block['attrs']['itemsJustification'] === 'right' ) {
+			$class_name .= ' items-justified-right';
+		}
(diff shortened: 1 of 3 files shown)


blockbase/functions.php
```diff
@@ -119,3 +120,32 @@ function blockbase_restore_customizer() {
...
+ */
+add_filter( 'render_block', 'blockbase_social_menu_render', 10, 2 );
+function blockbase_social_menu_render( $block_content, $block ) {
+	if ( 'core/social-links' !== $block['blockName'] ) {
+		return $block_content;
+	}
+
+	$return = $block_content;
+
+	if ( has_nav_menu( 'social' ) ) :
```
scruffian: you could reverse this conditional and then do an early return.

scruffian:
The main problem I see with this approach is that if I add a social link block to my content it gets taken over...

MaggieCabrera:
> The main problem I see with this approach is that if I add a social link block to my content it gets taken over...

yeah, the only other thing I can think of is having something like the attribute for location that the navigation block has, but I'm not sure if that's going to be accepted into GB

scruffian:
I made a few changes to this:
- I changed this from hijacking the social-links block to hijacking the menu block
- I created a new function which handles the conditions under which to hijack the render for readability
- I output all of the block markup

MaggieCabrera:
Yeah, this is very smart! seems to be working fine and the FSE flow is fine too. I think we need to strip the extra classes from `blockbase_social_menu_render` and add the skatepark specific ones on a skatepark specific renderer function maybe? You may not want the social navigation to be aligned right and with primary color for every child theme.

blockbase/functions.php
```diff
@@ -119,3 +120,51 @@ function blockbase_restore_customizer() {
...
+// We should only change the render of the navigtion block
+// to social links in the following conditions.
+function blockbase_condition_to_render_social_menu( $block ) {
+	// The block should be a navigation block.
+	if ( 'core/navigation' !== $block['blockName'] ) {
+		return false;
+	}
+
+	// The theme should define a social menu.
+	if ( ! has_nav_menu( 'social' ) ) {
```
jffng, line 137: This means that the classic menu the _must_ be named `Social` (or `social`, capitalization doesn't matter apparently) in the Customizer:

Screen Shot 2021-08-25 at 2 55 01 PM

Not great but not terrible UX, we should maybe document that somewhere. Also since we have headstart we can just make sure the social menu is created + populated already, so that helps.

jffng (commented):
This is cool! 

One downside I see to this approach is when you load the site editor, it comes as a navigation block and not social links block, so when you go to select from existing menu, you don't get the icons: 

Screen Shot 2021-08-25 at 2 50 38 PM

blockbase/functions.php
MaggieCabrera: mmh, I think it's not refering to the name of the menu, but the location, and that's defined by the theme, so shouldn't matter what the menu is called, just that it's applied to the social menu location

scruffian: Actually no, here the parameter is the location (see https://developer.wordpress.org/reference/functions/has_nav_menu/)

However you are right that calling `wp_get_nav_menu_items('social')` is making that assumption which is a bit of a problem, but not hard to fix :)

scruffian:
I added a commit so that we render the menu which is saved at the "social" location, rather than just the menu called "social". Ready for another review.

blockbase/functions.php
jffng: Thank you! https://github.com/Automattic/themes/pull/4467/commits/3cc87a98ec591cbc3304a7631bb8c08290628cff fixed it for me, I can name the menu whatever I want now, assign it to the social location, and it works.

jffng:
What should be done about the Skatepark specific classes?

scruffian:
> What should be done about the Skatepark specific classes?

If we can't achieve the same things through settings on the nav block then can we add a class to the nav block and write custom CSS for them?

scruffian:
I added some code to map the itemsJustification from the navigation block to the social icons block. I think this is good to go now.

blockbase/functions.php
```diff
@@ -34,6 +34,7 @@ function blockbase_support() {
 		register_nav_menus(
 			array(
 				'primary' => __( 'Primary Navigation', 'blockbase' ),
+				'social' => __( 'Social Navigation', 'blockbase' )
```
MaggieCabrera: I think if we add this in BB we need to add the navigation block for it on the rest of the children too, else the user will see it in the customizer but won't get a social menu on the frontend. I suggest we add this to Skatepark for now and then handle the rest of the children in a separate issue

MaggieCabrera:
I'm testing on dotcom and some of the icons aren't working for me:

Screenshot 2021-08-27 at 11 08 04
Screenshot 2021-08-27 at 11 08 01

blockbase/functions.php
scruffian: Done

scruffian:
I added a commit to better map menu items to service names.

MaggieCabrera:
I can't ✅ this for you but this LGTM, thanks so much for picking this up!

Issue #4463, opened 2021-08-24
Blockbase + children: header and post meta items need to use the new flex layout


MaggieCabrera:
After https://github.com/WordPress/gutenberg/pull/33812 was merged, there's a new css rule being applied to the default layout support: `$selector > * + * { margin-top: var( --wp--style--block-gap ); margin-bottom: 0; }` This is adding a margin top to every children of the container except the first, causing the post meta groups to break:

Quadrat:
<img width="669" alt="Screenshot 2021-08-24 at 16 31 29" src="https://user-images.githubusercontent.com/3593343/130635953-319e22bb-9e37-45c0-9681-30e9b26afe7b.png">

Seedlet + Mayland:
<img width="398" alt="Screenshot 2021-08-24 at 16 34 38" src="https://user-images.githubusercontent.com/3593343/130636190-8b7bba48-4281-4ab6-8e86-086d325f8a01.png">

Skatepark:
<img width="308" alt="Screenshot 2021-08-24 at 16 35 22" src="https://user-images.githubusercontent.com/3593343/130636265-b667d0b7-132a-47d5-80ab-d46f037fe9b9.png">

<img width="1405" alt="Screenshot 2021-08-24 at 16 36 14" src="https://user-images.githubusercontent.com/3593343/130636431-e3b6c5eb-a29e-4df8-9cc8-a4f85ac0572d.png">

We need to refactor any group block we have added display flex manually on CSS to use the new flex layout instead.

jffng:
I took at the issue in [that last screenshot](https://user-images.githubusercontent.com/3593343/130636431-e3b6c5eb-a29e-4df8-9cc8-a4f85ac0572d.png) of Skatepark's header, and that group is already defining flex support via the block: https://github.com/Automattic/themes/blob/trunk/skatepark/block-template-parts/header.html#L1

The cause of the issue is actually our alignment rules 😅 , so let's double check our fix there against the skatepark header.

MaggieCabrera:
> The cause of the issue is actually our alignment rules 😅 , so let's double check our fix there against the skatepark header.

Doh!

scruffian:
This was fixed by #4468

PR #4468, opened 2021-08-25
Blockbase: Update Blockbase and children to use flex

scruffian:
Since https://github.com/WordPress/gutenberg/pull/33359/files merged we can use the flex property of the group block to achieve this layout. The `blockGap` setting doesn't seem to be working in my tests.

#### Related issue(s):
https://github.com/Automattic/themes/issues/4463

### DIFF
--- a/blockbase/theme.json
+++ b/blockbase/theme.json
@@ -564,7 +564,7 @@
 			}
 		},
 		"spacing": {
-			"blockGap": "var(--wp--custom--margin--horizontal)"
+			"blockGap": "var(--wp--custom--margin--vertical)"
 		},
 		"typography": {
 			"lineHeight": "var(--wp--custom--body--typography--line-height)",
--- a/blockbase/theme.json
+++ b/blockbase/theme.json
@@ -314,7 +314,6 @@
 			"wideSize": "1000px"
 		},
 		"spacing": {
-			"blockGap": "var(--wp--custom--margin--horizontal)",
 			"customPadding": true,
 			"units": [
 				"%",
@@ -563,6 +562,9 @@
 					"text": "var(--wp--custom--color--primary)"
 				}
 			}
+		},
+		"spacing": {
+			"blockGap": "var(--wp--custom--margin--horizontal)"
 		},
 		"typography": {
 			"lineHeight": "var(--wp--custom--body--typography--line-height)",
--- a/blockbase/theme.json
+++ b/blockbase/theme.json
@@ -314,6 +314,7 @@
 			"wideSize": "1000px"
 		},
 		"spacing": {
+			"blockGap": "var(--wp--custom--margin--horizontal)",
 			"customPadding": true,
 			"units": [
 				"%",
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -107,10 +107,7 @@ img {
 }
 
 .site-header {
-	align-items: center;
-	flex-wrap: wrap;
 	justify-content: space-between;
-	display: flex;
 	overflow: inherit;
 }
 
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -814,10 +814,6 @@ p.has-background {
 	margin-bottom: unset;
 }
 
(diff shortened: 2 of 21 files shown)


MaggieCabrera:
Mmmh, I think the gap was only applying initially to the normal layout, not the flex one (so it's actually a vertical gap between blocks inside a container, not horizontal)

scruffian:
`blockGap` should be inside `styles` not `settings`. This now works as expected.

MaggieCabrera:
Looks like Quadrat's header is affected by this ([here](https://github.com/Automattic/themes/blob/45bb8e35e0977ff82c03a47b72a691d49fc201a8/blockbase/sass/base/_header.scss#L6)):

Screenshot 2021-08-26 at 11 16 20

scruffian:
Good catch. Updated it!

blockbase/theme.json
```diff
@@ -563,6 +563,9 @@
 				}
 			}
 		},
+		"spacing": {
+			"blockGap": "var(--wp--custom--margin--horizontal)"
+		},
```
MaggieCabrera: I think this might be too big:

Screenshot 2021-08-26 at 12 22 16
Screenshot 2021-08-26 at 12 23 27

This is also applying vertically to blocks inside layout inherit groups. The rule coming from [here](https://github.com/WordPress/gutenberg/blob/8d1ea0a3ad5440c64e8fae273e3d8125a3f61500/lib/block-supports/layout.php#L66).

I think GB should have two different gap values for the two different layouts, since inherit is going to be separating vertically and flex is horizontal, which usually means smaller blocks of text that don't need such a big gap between them.

scruffian:
I've updated this to use the vertical margin everywhere. Ideally we'd be able to control the horizontal and vertical separately but for now I think this is an OK compromise.

Tracking here: https://github.com/WordPress/gutenberg/issues/34347

MaggieCabrera:
Skatepark's post meta looks like this now:

Screenshot 2021-08-27 at 16 43 48

pbking:
Fixed skatepark's meta:



Had to use an `!important` to beat the Gutenberg center style.  That seems to opinionated to not have a way to configure.

pbking:
NOW it looks like this: 
image

pbking:
I tried to fix the gap issue (as an alternative to just [setting it here](https://github.com/Automattic/themes/pull/4468/commits/d10d84efdf38c8401f35f86e7aef05d681f30c5d)) by adjusting the gap value for the post-meta block:

```
...
  "styles": {
    "blocks": {
      "core/post-meta": {
        "spacing": 
          "blockGap": "0.5em;"
...
```

However it had no effect of the value of `--wp--style--block-gap` for the post-meta block as I had expected.
This is contrary to [this statement](https://github.com/WordPress/gutenberg/issues/34347#issuecomment-907138437) which tells me that it should.

MaggieCabrera:
> I tried to fix the gap issue (as an alternative to just [setting it here](https://github.com/Automattic/themes/pull/4468/commits/d10d84efdf38c8401f35f86e7aef05d681f30c5d)) by adjusting the gap value for the post-meta block:
> 
> ```
> ...
>   "styles": {
>     "blocks": {
>       "core/post-meta": {
>         "spacing": 
>           "blockGap": "0.5em;"
> ...
> ```
> 
> However it had no effect of the value of `--wp--style--block-gap` for the post-meta block as I had expected.
> This is contrary to [this statement](https://github.com/WordPress/gutenberg/issues/34347#issuecomment-907138437) which tells me that it should.

post-meta is not a core block, it's just a class that we use for this bit of the theme :)

pbking:
> post-meta is not a core block, it's just a class that we use for this bit of the theme :)

I just realized that and deleted my comment. :P

MaggieCabrera:
> I just realized that and deleted my comment. :P

No one will know :D

pbking:
I can't find anything else that doesn't look correct.  I believe this is ready to come in.

pbking (approved):
LGTM
🚢

scruffian:
Thanks for finishing this one off @pbking

PR #4469, opened 2021-08-25
Create a Deploy Script

pbking:
This initially began as experiments with git on the sandbox and has morphed into tooling for deploying changes.  There is tooling for BOTH git and svn.

`npm run deploy:svn` if your sandbox is in svn mode
`npm run deploy:git` if your sandbox is in git mode

Doing so will:
	* Gets the last deployed hash from the sandbox
	* Version bump all themes have have changes since the last deployment
	* Commit the version bump change to github
	* Clean the sandbox and ensure it is up-to-date
	* Push all changed files (including removal of deleted files) since the last deployment
	* Update the 'last deployed' hash on the sandbox
	* Create a phabricator diff based on the changes since the last deployment.  The description including the commit messages since the last deployment.
	* Open the Phabricator Diff in your browser
	* Create a tag in the github repository at this point of change which includes the phabricator link in the description

In order to accomplish this the 'last deployed hash' is stored in the file `.pub-git-hash`.  This file represents the git hash last applied to wpcom's repository.  The difference between that hash and HEAD dictates which files are uploaded (or deleted) and what commit logs are included in the resulting Phabricator diff.


pbking:
Found an issue with the SVN version of the deploy script.  Pulling back to in-progress to fix.

scruffian:
This is looking good, thanks for working on it.

pushButtonDeployGit and pushButtonDeploySvn look pretty similar. Could we abstract that to a generic function that takes a source control param to handle the differences?

theme-utils.mjs
```diff
@@ -29,6 +29,591 @@ function showHelp(){
...
+import { spawn } from 'child_process';
+import fs from 'fs';
+import open from 'open';
+
+const remoteSSH = 'wpcom-sandbox';
+const sandboxPublicThemesFolder = '/home/wpdev/public_html/wp-content/themes/pub';
+const sandboxRootFolder = '/home/wpdev/public_html/';
+const isWin = process.platform === 'win32';
+
+(async function start() {
```
scruffian: this whole block seems to be in the file twice?

scruffian:
A note for anyone else testing:

```
 Expects the following to be configured in your ~/.ssh/config file:
Host wpcom-sandbox
	User wpdev
	HostName SANDBOXURL.wordpress.com
	ForwardAgent yes
```

scruffian:
Aside from a few minor comments I think we should merge this!

theme-utils.mjs
pbking: I had a problem yesterday where the entire file was duplicated.  I don't see that now though and thought I fixed and pushed that yesterday.

pbking:
> pushButtonDeployGit and pushButtonDeploySvn look pretty similar. Could we abstract that to a generic function that takes a source control param to handle the differences?

There are other instances of git vs svn (cleaning the sandbox, building the phabricator diff) and those especially feel better contained in repo-specific logic blocks.  But I lean towards agreeing that the top-level handler could be made more generic.  I have also been toying with adding interactivity to the process and as that "top level" logic grows it certainly doesn't make sense to keep such similar logic separate.

I shall combine them.

scruffian (approved):
LGTM

PR #4471, opened 2021-08-26
Skatepark: Logo alignment

### LINKED ISSUE #4419: Skatepark: Site logo appears misaligned
When you add a site logo, the left side of the menu appears very misaligned with the right: 

Screen Shot 2021-08-16 at 3 09 19 PM

@melchoyce do we have a design for how this is supposed to look?

MaggieCabrera:

This PR refactors the logo positioning to fit the new design guidelines provided in https://github.com/Automattic/themes/issues/4419

**Logo + Title + Tagline:**

Desktop | Mobile
--- | ---
Screenshot 2021-08-26 at 11 44 34 | Screenshot 2021-08-26 at 11 37 16

**Logo only:**

Desktop | Mobile
--- | ---
Screenshot 2021-08-26 at 11 45 16 | Screenshot 2021-08-26 at 12 04 18

**Title + Tagline:**

Desktop | Mobile
--- | ---
Screenshot 2021-08-26 at 11 58 03 | Screenshot 2021-08-26 at 11 57 58

#### Related issue(s):

Closes https://github.com/Automattic/themes/issues/4419

### DIFF
--- a/skatepark/sass/templates/_header.scss
+++ b/skatepark/sass/templates/_header.scss
@@ -27,25 +27,32 @@ header.wp-block-template-part {
 		margin-top: calc( 0.5 * var(--wp--custom--margin--vertical) );
 		display: grid;
 		grid-column-gap: var(--wp--custom--margin--horizontal);
+		grid-template-areas: 
+			"logo title"
+			"logo tagline";
 		grid-template-columns: auto 1fr;
+		@include break-small-only(){
+			grid-template-areas: 
+				"logo"
+				"title"
+				"tagline";
+		}
 		.wp-block-site-logo {
-			grid-column-start: 1;
-			grid-column-end: 2;
-			grid-row-start: 1;
-			grid-row-end: 3;
+			grid-area: logo;
 			margin: 0;
 			max-width: 120px;
 			align-self: center;
 			justify-self: flex-end;
+			@include break-small-only(){
+				margin-bottom: calc( 0.5 * var(--wp--custom--margin--vertical) );
+			}
 		}
 		.wp-block-site-title {
-			grid-row-start: 1;
-			grid-row-end: 2;
+			grid-area: title;
 			margin: 0 0 0.5em;
 		}
 		.wp-block-site-tagline {
-			grid-row-start: 2;
-			grid-row-end: 3;
+			grid-area: tagline;
 			margin: 0;
 		}
 	}
--- a/skatepark/sass/templates/_header.scss
+++ b/skatepark/sass/templates/_header.scss
@@ -4,9 +4,9 @@ header.wp-block-template-part {
 	width: 100%;
 	
 	> .wp-block-group {
-		align-items: flex-end; // Needed until theme.json layout lets me specify
+		gap: 0;
 		justify-content: space-between; // Apply a cluster (flex?) layout
-		flex-wrap: wrap-reverse;
+		align-items: flex-start;
 		> * {
 			flex-grow: 1; // Needed to maintain alignment when the containers stack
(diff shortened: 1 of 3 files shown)


skatepark/sass/templates/_header.scss
```diff
@@ -22,6 +22,33 @@ header.wp-block-template-part {
 			}
 		}
 	}
+
+	.site-brand {
+		margin-top: 20px;
```
scruffian: Should this use a variable?

skatepark/sass/templates/_header.scss
```diff
@@ -22,6 +22,33 @@ header.wp-block-template-part {
...
+		display: grid;
+		grid-column-gap: 30px;
```
scruffian: Should this use a variable?

skatepark/sass/templates/_header.scss
```diff
@@ -22,6 +22,33 @@ header.wp-block-template-part {
...
+		grid-column-gap: var(--wp--custom--margin--horizontal);
+		grid-template-columns: auto 1fr;
+		.wp-block-site-logo {
+			grid-column-start: 1;
+			grid-column-end: 2;
+			grid-row-start: 1;
+			grid-row-end: 3;
+			margin: 0;
+			max-width: 120px;
+			align-self: center;
```
scruffian, line 48: I'm not sure about this. I think I'd prefer the logo to align to the top of the screen like this:

Screenshot 2021-08-26 at 15 32 17

scruffian (approved):
One minor comment, but LGTM

skatepark/sass/templates/_header.scss
MaggieCabrera: On figma it was centered, what do you think, @melchoyce ?

MaggieCabrera: ah, it was centered on desktop, true, maybe the logo should be on top and the tagline below? instead of on a two column layout?

MaggieCabrera: Ok I don't know why I didn't follow the design, this is closer to what's on figma now:

Screenshot 2021-08-26 at 16 42 16

Screenshot 2021-08-26 at 16 43 25

MaggieCabrera:
I've made the menu container a bit smaller on mobile, it wasn't looking like the design. I also think that maybe we can make the title and tagline be on top of the menu only if the logo is present, since there's the "custom-logo" class on the body when that's the case, do you agree @melchoyce?

melchoyce:
> I also think that maybe we can make the title and tagline be on top of the menu only if the logo is present, since there's the "custom-logo" class on the body when that's the case, do you agree @melchoyce?

Like this?

image

I like that because I like the logo and title/tagline underneath the menu bar 😂 But I think for consistency we should stick with one, and hierarchically it does makes sense for it to go logo, title, tagline, menu bar, especially since we're not fixing the menu bar positioning.

scruffian:
I think having the menu above the title/tagline looks great :)

MaggieCabrera:
> I like that because I like the logo and title/tagline underneath the menu bar 😂 But I think for consistency we should stick with one, and hierarchically it does makes sense for it to go logo, title, tagline, menu bar, especially since we're not fixing the menu bar positioning.

Then I think this PR is ready to go if @melchoyce thinks it looks fine as it is.

melchoyce:
🚢 :shipit: 🚢

pbking:
I think the flex alignment PR introduced some conflicts so I pulled this into 'in progress' so it could have a bit of a rebase and further evaluation taking those changes into account.

MaggieCabrera:
After the rebase this still looks the same to me so I'm bringing this in

PR #4472, opened 2021-08-26
Geologist: New Theme

scruffian:
This creates a new Blockbase child theme but using the theme.json and template files from Quadrat but with a few changes to the colors. Feel free to iterate...


alaczek:
Hi Ben! Thank you for kickstarting this project! I tried this initial version on my local install, and I noticed a couple of odd style things:

There's no horizontal padding in the header:
image

The post meta in single post view is off to the left:
image

The previous/next post nav is missing the top border and one of the arrows is facing the wrong way (and they don't have the border on top):
image

Does this mean you didn't port all of Quadrat's styles? Because the main change I had in mind was removing the shapes in the background while keeping the rest of the styles intact.

scruffian:
Yeah I didn't add the custom CSS from Quadrat. Let me do that now...

scruffian:
I've added the Quadrat styles and removed the triangles from the top and bottom. I also made a few adjustments to form styles.

kjellr:
I'm noticing some small issues here and there, but in general this is a great start. Later this week I'll take a stab at adding a screenshot and maybe tightening up the space slightly now that the triangle is gone. 👍

kjellr:
I've pushed a few updates to the theme: 

- I renamed it to "Geologist", which I think it sounds better than Bedrock. (That's just my personal opinion sneaking in. 😄)
- Changed the color scheme so that this isn't yet another white-background theme in the showcase. 
- Left-aligned the header and post meta. 
- Tightened up the vertical spacing, made it consistent between the index, page, and single templates. 

Here's a screenshot of how this should look: 

Frame 7

The mobile header width isn't quite right — it's narrower than the rest of the content: 

Screen Shot 2021-09-02 at 12 00 52 PM

Also, header items appear to be stacking instead of being spread out horizontally: 

Screen Shot 2021-09-02 at 12 00 37 PM

Those could use some tightening up if someone from theme dev has time. I'm not sure if they're Quadrats bugs being reproduced here, of if they're something new.

scruffian:
> The mobile header width isn't quite right — it's narrower than the rest of the content:

I think that's a bug coming from Quadrat. I have added a fix....

Now it looks like this:

Screenshot 2021-09-02 at 20 45 23
Screenshot 2021-09-02 at 20 45 09
Screenshot 2021-09-02 at 20 44 20
Screenshot 2021-09-02 at 20 44 13

> Also, header items appear to be stacking instead of being spread out horizontally:

Having trouble replicating this one

alaczek:
@kjellr we wanted to use this theme as the default blogging theme with white background. I agree we need more color variety, but would you consider switching it back to white? Otherwise, _I think_ I'm going to have to spin up (well, generate) yet another theme variation, that won't be all that different from this one (unless there's another solution I'm missing).

kjellr:
> I think that's a bug coming from Quadrat. I have added a fix....

Thanks, that appears to solve it. 👍 I noticed another issue which I've filed in https://github.com/Automattic/themes/issues/4527. 

> Having trouble replicating this one

Yeah, I'm having trouble replicating it on other sites too. Let's just consider this is a bug with my testing site unless it comes again. 👍 

> we wanted to use this theme as the default blogging theme with white background. I agree we need more color variety, but would you consider switching it back to white? Otherwise, I think I'm going to have to spin up (well, generate) yet another theme variation, that won't be all that different from this one (unless there's another solution I'm missing).

I have a couple thoughts: First, I’m not convinced that the default needs to be white. Default doesn’t need to mean bland, and I think this theme in particular (because of its simplicity) is a little bland if it’s just pure white. That, compounded with the  prevolence of other white themes in the showcase, led me to want to avoid a white background here. 

But also, I think it’s important for us to show variants of these themes in the Showcase somehow and the easiest way to do that is to generate multiples of each theme (this was a discussion point early on with the Skatepark theme actually). So I think we should do that either way.*

_* A caveat here: while I think the color variations should be present in the directory, I don't think we'd want to actually surface them all together. The default view would be a curated list, but then if someone were to specify that they want (for example) a white theme, we could show only the white variants of each theme._

scruffian:
What's stopping us shipping this?

kjellr:
Nothing major. I think it just needs a little testing, though I imagine any bugs we find will also be Quadrat bugs. 😄 Might be fine to merge now, and then follow up with any fixes/updates from there?

Either way, @melchoyce, @beafialho, and I will give this one a test next week.

ianstewart:
> … I think it’s important for us to show variants of these themes in the Showcase somehow and the easiest way to do that is to generate multiples of each theme So I think we should do that either way.

Plus one. This is nice, simple, base for blogging. It seems easy to imagine 3–9 variations of it in the showcase. @kjellr are there other colour scheme plus default image choices at hand? These should be *very* easy to launch as generated themes on Dotcom.

> First, I’m not convinced that the default needs to be white. Default doesn’t need to mean bland … That, compounded with the prevalence of other white themes in the showcase, led me to want to avoid a white background here.

For dotcom this is trivial to test by merely picking _any_ default and seeing what most people switch to after signup or before launch with existing tracks events. If we had multiple colour schemes we could even mimic an AB test on default colours by selecting a generated variant randomly and seeing what people switch to from various starting points. With our dotcom defaults we can and should be able to quickly update this.

**Note:** None of the above is a blocker on launching Geologist. It's super-nice and we should launch it ASAP. :) Also the name is cool. :)

kjellr:
> Plus one. This is nice, simple, base for blogging. It seems easy to imagine 3–9 variations of it in the showcase. @kjellr are there other colour scheme plus default image choices at hand? These should be very easy to launch as generated themes on Dotcom.

Not yet, but we can work on creating those once this one is all set to launch. 👍

PR #4474, opened 2021-08-26
Blockbase children: Allow colors to map in patterns and preexisting content

scruffian:
Blockbase defines five colors for themes to use: `primary`, `secondary`, `foreground`, `background` and `selection`. However not all themes have five colors. If I create content in Blockbase using these five colors and then switch to another theme that doesn't use all five colors then it can look pretty bad.

This adds additional CSS to use the mappings in child themes to the Blockbase color variables, so that the colors carry over to child themes in a more predictable way.

To test, create some content like this:
```
<!-- wp:paragraph {"backgroundColor":"primary"} -->
<p class="has-primary-background-color has-background">primary background</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"backgroundColor":"secondary"} -->
<p class="has-secondary-background-color has-background">secondary background</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"backgroundColor":"foreground"} -->
<p class="has-foreground-background-color has-background">foreground background</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"backgroundColor":"background"} -->
<p class="has-background-background-color has-background">background background</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"backgroundColor":"selection"} -->
<p class="has-selection-background-color has-background">selection background</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"textColor":"primary"} -->
<p class="has-primary-color has-text-color">primary text</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"textColor":"secondary"} -->
<p class="has-secondary-color has-text-color">secondary text</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"textColor":"foreground"} -->
<p class="has-foreground-color has-text-color">foreground text</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"textColor":"background"} -->
<p class="has-background-color has-text-color">background text</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"textColor":"selection"} -->
<p class="has-selection-color has-text-color">selection text</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"elements":{"link":{"color":{"text":"var:preset|color|primary"}}}}} -->
<p class="has-link-color">primary <a href="http://scruffian.com">link</a></p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"elements":{"link":{"color":{"text":"var:preset|color|secondary"}}}}} -->
<p class="has-link-color">secondary <a href="http://scruffian.com">link</a></p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"elements":{"link":{"color":{"text":"var:preset|color|foreground"}}}}} -->
<p class="has-link-color"><meta charset="utf-8">foreground <a href="http://scruffian.com">link</a></p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"elements":{"link":{"color":{"text":"var:preset|color|background"}}}}} -->
<p class="has-link-color"><meta charset="utf-8">background <a href="http://scruffian.com">link</a></p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"elements":{"link":{"color":{"text":"var:preset|color|selection"}}}}} -->
<p class="has-link-color"><meta charset="utf-8">selection <a href="http://scruffian.com">link</a></p>
<!-- /wp:paragraph -->
```

In Blockbase this looks as expected but in Quadrat it looks like this:
Screenshot 2021-08-26 at 12 33 06

With this patch applied it looks like this:
Screenshot 2021-08-26 at 12 30 13


MaggieCabrera:
Mmmh, I agree with this, but I'm not super sure this needs to be in Blockbase. If we think the editor should have these rules (as in have a primary, secondary... colors defined for every theme) then it's fine that it's in BB. But I'm not sure that's how things are going to go for semantic colors in the editor and BB doesn't need these rules because it already has 5 colors. 

There's also the matter that the child themes may not want the same specific combination of colors to replace the missing ones, so they may need a copy of similar rules instead.

This is Skatepark:

Screenshot 2021-08-26 at 15 35 32

MaggieCabrera:
I'd like the eyes of the designers on this: @beafialho @kjellr @melchoyce since we are trying to improve the user experience when switching between themes that use our patterns and I don't think this might be enough for that looking at the case of Skatepark. I'm thinking maybe themes should define 5 colors even if they are only actively using 2 of them, specifically for this case (and to give contrasting options to the user in case they want different shades of the colors for their own use).

scruffian:
> There's also the matter that the child themes may not want the same specific combination of colors to replace the missing ones, so they may need a copy of similar rules instead.

This is relying on the mapping that the theme sets in "custom" so if the theme wants to map it differently, it can.

pbking:
I think that most things will also need to provide some additional things to make this worthwhile.  
This is Skatepark with both colors mapped in theme.json and reasonable default colors supplied for those color class choices.

image

skatepark/sass/base/_colors.scss
```diff
@@ -0,0 +1,10 @@
+.has-foreground-background-color,
+.has-selection-background-color,
+.has-primary-background-color,
+.has-secondary-background-color {
+	color: var(--wp--preset--color--background);
+}
+
+.has-background-color {
+	background-color: var(--wp--preset--color--primary);
```
scruffian: Why do we need this?

skatepark/sass/base/_colors.scss
```diff
@@ -0,0 +1,10 @@
+.has-foreground-background-color,
+.has-selection-background-color,
+.has-primary-background-color,
+.has-secondary-background-color {
+	color: var(--wp--preset--color--background);
```
scruffian: I'm not very keen on rules like this. In Gutenberg, if users choose an unreadable color combination we don't try to fix it for them, we just give them a warning. I worry that doing this creates an expectation that the editor will try to fix this for you which isn't the reality.

pbking: Without this (or something like it) patterns and existing content may be less likely to look acceptable when switching themes if they are using semantic colors.  I was thinking less about the application of the colors to the example above and more about how it would apply to an existing pattern.  For instance if the original leveraged the primary color as a background, and the default foreground (text) color was appropriate for that theme's palette.  Sure it's possible to go and change those blocks once the theme has been switched, but it seems like an important goal for that relationship between semantic color combinations be outlined for the themes so that a user wouldn't have to.

However I didn't actually quantify that with an example so it might be a moot point.  I'll see if I can post a good existing example.

pbking: I see what you're saying @scruffian.  I've backed out my commits about "reasonable defaults".

pbking:
I finally got on the same page w/ @scruffian 's original statement.

I used the below as a test against all of the tests (building it helped me better understand the goal).

```
<!-- wp:group {"align":"wide","style":{"border":{"width":"2px","style":"solid"},"spacing":{"padding":{"top":"20px","right":"20px","bottom":"20px","left":"20px"}},"elements":{"link":{"color":{"text":"var:preset|color|primary"}}}},"borderColor":"primary","backgroundColor":"background","textColor":"foreground","layout":{"inherit":true}} -->
<div class="wp-block-group alignwide has-border-color has-primary-border-color has-foreground-color has-background-background-color has-text-color has-background has-link-color" style="border-style:solid;border-width:2px;padding-top:20px;padding-right:20px;padding-bottom:20px;padding-left:20px"><!-- wp:paragraph -->
<p>To begin this group has an outline in PRIMARY color.  The background (of both the group and the page) is BACKGROUND and the text you are reading is FOREGROUND.  The <a href="https://wordpress.org" data-type="URL" data-id="https://wordpress.org">link here</a> is also assigned to PRIMARY.</p>
<!-- /wp:paragraph -->

<!-- wp:group {"style":{"spacing":{"padding":{"top":"5px","right":"5px","bottom":"5px","left":"5px"}}},"backgroundColor":"primary","textColor":"background"} -->
<div class="wp-block-group has-background-color has-primary-background-color has-text-color has-background" style="padding-top:5px;padding-right:5px;padding-bottom:5px;padding-left:5px"><!-- wp:paragraph -->
<p>Here is a group that has a background color of PRIMARY and a foreground of BACKGROUND.  Using background to contrast against FOREGROUND, PRIMARY or SECONDARY is standard behavior, though not automatically done.</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group -->

<!-- wp:group {"style":{"spacing":{"padding":{"top":"5px","right":"5px","bottom":"5px","left":"5px"}}},"backgroundColor":"secondary","textColor":"background"} -->
<div class="wp-block-group has-background-color has-secondary-background-color has-text-color has-background" style="padding-top:5px;padding-right:5px;padding-bottom:5px;padding-left:5px"><!-- wp:paragraph -->
<p>This second group has a background color of SECONDARY with a text color of BACKGROUND.</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group -->

<!-- wp:group {"style":{"border":{"width":"2px","style":"solid"},"spacing":{"padding":{"top":"5px","right":"5px","bottom":"5px","left":"5px"}}},"borderColor":"selection"} -->
<div class="wp-block-group has-border-color has-selection-border-color" style="border-style:solid;border-width:2px;padding-top:5px;padding-right:5px;padding-bottom:5px;padding-left:5px"><!-- wp:paragraph -->
<p>Lastly, you should be able to highlight any of the text and it will highlight with the highlight color (which you can also see as the outline of the group this text is in.</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group --></div>
<!-- /wp:group -->
```

Sample:


Before:


After:

pbking:
> If we think the editor should have these rules (as in have a primary, secondary... colors defined for every theme) then it's fine that it's in BB. But I'm not sure that's how things are going to go for semantic colors in the editor and BB doesn't need these rules because it already has 5 colors.

@MaggieCabrera I believe this is important to standardize in Blockbase.  We do guarantee that there are exactly five semantic color values to be used in each theme (those that are expressed in `custom.colors`).  This guarantees that no matter how many semantic colors the theme makes available to the USER (via the palette) those values will be available in some way, not only to the theme.json files (which is what is consuming the guaranteed five color values from `custom.colors`) but also anything defined in patterns or selected by the user from any Blockbase theme.  The mapping mechanism is [already in place ](https://github.com/Automattic/themes/blob/trunk/skatepark/child-theme.json#L87)this is just making that mapping more available.

I believe this is good to go with the original change but keen to know how your opinion stands today.

MaggieCabrera:
> @MaggieCabrera I believe this is important to standardize in Blockbase. We do guarantee that there are exactly five semantic color values to be used in each theme (those that are expressed in `custom.colors`). This guarantees that no matter how many semantic colors the theme makes available to the USER (via the palette) those values will be available in some way, not only to the theme.json files (which is what is consuming the guaranteed five color values from `custom.colors`) but also anything defined in patterns or selected by the user from any Blockbase theme. The mapping mechanism is [already in place ](https://github.com/Automattic/themes/blob/trunk/skatepark/child-theme.json#L87)this is just making that mapping more available.
> 
> I believe this is good to go with the original change but keen to know how your opinion stands today.

My opinion is that this should come from GB instead of the theme, but I don't disagree that this PR is an improvement from what we started with. I still think it's important from a design perspective to be aware of this situation and I think it's patent when we test it directly against the existing patterns. 

Blockbase:
Screenshot 2021-09-09 at 13-25-11 Patterns – Skatepark

Mayland:
Screenshot 2021-09-09 at 13-23-42 Patterns – Skatepark

Skatepark:
Screenshot 2021-09-09 at 13-22-58 Patterns – Skatepark

Quadrat:
Screenshot 2021-09-09 at 13-22-15 Patterns – Skatepark

Seedlet:
Screenshot 2021-09-09 at 13-21-46 Patterns – Skatepark

The problem in Skatepark is hard to solve without adding extra colors, but this is a real user case that people are going to encounter, these are the themes we are expecting users to swap between that have patterns with these specific names in their colors.

MaggieCabrera:
Check how Mayland Blocks does define 5 colors even though in practice it only uses 2 (this problem predates block themes and the solution was to define 5 always no matter what) and that makes the Quadrat pattern look legible

pbking:
Quadrat's color scheme has the assumption that the SECONDARY color will contrast with the FOREGROUND color (pairing with the BACKGROUND color):


However, all other themes' color schemes assume that the SECONDARY color will contrast with the BACKGROUND color.  Indeed that ALL colors would contrast with the BACKGROUND color.



The solution, I believe, is to document a style guide for the color schemes so that the relationships between them can be assumed across ALL themes and ALL color schemes.

I would argue that Quadrat is "the one of these things is not like the other".  Currently we (kind of) have the following relationship assumption (and all themes but Quadrat follow, even Skatepark which only has two colors)

BACKGROUND: The uh, background
FOREGROUND: Contrasts with Background
PRIMARY: Contrasts with BACKGROUND
SECONDARY: Contrasts with BACKGROUND
SELECTION: Contrasts with BACKGROUND

Only Quadrat:
PRIMARY: Contrasts with FOREGROUND

If instead we had another semantic color that PAIRS with BACKGROUND and CONTRASTS with FOREGROUND (which is what Quadrat's PRIMARY color is doing) then all other themes can follow that (even if it just means assigning the BACKGROUND color to ANOTHERSEMANTICCOLOR for a Theme like Skatepark where exactly two colors is the intended design).

And finally, this is (further) argument (to me) that this kind of "standard semantic color definition and relationship" should be canonized in Gutenberg.

pbking:
For discussion I have pushed a change that introduces `tertiary` semantic color to Blockbase & Co.
The intention is for that color to PAIR with background and CONTRAST with everything that background contrasts with.
It defaults to just being the same color as background in blockbase but provides the opportunity for Quadrat to change it and any other theme to show content and patterns that just work with it.

kjellr:
> I would argue that Quadrat is "the one of these things is not like the other". Currently we (kind of) have the following relationship assumption (and all themes but Quadrat follow, even Skatepark which only has two colors)
> 
> BACKGROUND: The uh, background
> FOREGROUND: Contrasts with Background
> PRIMARY: Contrasts with BACKGROUND
> SECONDARY: Contrasts with BACKGROUND
> SELECTION: Contrasts with BACKGROUND
> 
> Only Quadrat:
> PRIMARY: Contrasts with FOREGROUND

Following up on an offline chat: 

In Seedlet (and prior to that), we had the following colors:

- Foreground
- Background
- Primary (Basically, the accent color)
- Secondary (Another accent color if necessary. Hover state for the Primary color)
- Tertiary (An alternate background color)

Here are the contrast rules we aimed for:

Contrast with Background|Contrast against Foreground
---|---
Foreground|Background
Primary|Tertiary
Secondary

pbking:
Closing in lieu of #4570.

PR #4475, opened 2021-08-26
Adds a small script to fix all warnings raised by npm modules

scruffian:
This is an attempt to fix the warnings raised here: https://github.com/automattic/themes/security/dependabot?page=2&q=is%3Aopen

I created a script which will do it for us automatically...

### DIFF
--- a/alves/package-lock.json
+++ b/alves/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "alves",
-  "version": "1.5.10",
+  "version": "1.5.11",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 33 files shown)


mikachan (commented):
This is really cool! I ran the script locally and it worked for me. Only thing I can think to add is a shortcut to run the script, maybe something like: `"fix:deps": "./fix-dependencies.sh"` - but I'm also not sure how/where we plan to run this script, so maybe a shortcut isn't needed.

Do we have something similar for keeping the dependencies up to date?

scruffian:
> Do we have something similar for keeping the dependencies up to date?

Not to my knowledge

pbking:
Slightly different take on the task in #4478.  Added batch install, build, audit and version-bump.

scruffian:
Closing in favour of https://github.com/Automattic/themes/pull/4478

PR #4476, opened 2021-08-26
Skatepark: fix separator color

MaggieCabrera:

The css for the separator block on skatepark was forcing a black color. I changed it so it doesn't affect the color anymore. To test this check the theme with different body colors and see that the separator matches it and it's still 3px height.

Screenshot 2021-08-26 at 15 20 54


jffng (approved):
Nice, checked with the different color palettes in Skatepark and it's working as described. Thanks!

PR #4477, opened 2021-08-26
Skatepark: remove unneeded CSS for mixed media pattern

jffng:

This PR removes pattern-specific CSS that was overriding the default margins, and adds a spacer in one spot to get things closer to the design. It is a follow up from some changes made in #4449.
 
To test, load the Mixed Media in Container pattern and make sure it looks as expected:

Screen Shot 2021-08-26 at 1 20 20 PM

### DIFF
--- a/skatepark/assets/theme.css
+++ b/skatepark/assets/theme.css
@@ -408,11 +408,6 @@ p {
 	margin-bottom: 0;
 }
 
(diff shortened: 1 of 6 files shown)


scruffian (approved):
LGTM

melchoyce (approved):
LGTM

PR #4478, opened 2021-08-26
Try/add batch utils

pbking:
I ttook the small script from #4475 and made it a bigger script.

This new "theme batch util" script does the following:
* installs all theme project dependencies `npm run batch:install`
* builds all theme projects `npm run batch:build:all`
* builds all theme projects that have changed resources `npm run batch:build:changed`
* audits all theme project dependencies `npm run batch:audit`
* version bumps all projects that have any changes `npm run batch:version-bump`

The first few are pretty straightforward.  
The last one is my favorite.

If ANY changes are in a theme project (from the point of divergence from /trunk to now) then that theme gets a patch version bump.  If the version has changed since that divergence it skips it, for instance major/minor version changes done manually, as well as previous runs of 'version-bump'. The version is then copied to the relevant theme files (style.css, style.scss, etc).

This utility could be leveraged in a number of different places in a deployment workflow.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,8 @@
 		"sandbox:push:ignore": "./sandbox.sh push --ignore",
 		"sandbox:push:force": "./sandbox.sh push --force",
 		"sandbox:watch": "chokidar '**/*' -i '*/node_modules' -i '.git' -c './sandbox.sh push --ignore' --initial",
-		"batch:build": "./theme-batch-utils.sh build",
+		"batch:build:changed": "./theme-batch-utils.sh build",
+		"batch:build:all": "./theme-batch-utils.sh build-all",
 		"batch:install": "./theme-batch-utils.sh install-dependencies",
 		"batch:audit": "./theme-batch-utils.sh audit-dependencies",
 		"batch:version-bump": "./theme-batch-utils.sh version-bump"
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
 		"batch:build:changed": "./theme-batch-utils.sh build",
 		"batch:build:all": "./theme-batch-utils.sh build-all",
 		"batch:install": "./theme-batch-utils.sh install-dependencies",
-		"batch:audit": "./theme-batch-utils.sh audit-dependencies",
+		"batch:audit:fix": "./theme-batch-utils.sh audit-dependencies",
 		"batch:version-bump": "./theme-batch-utils.sh version-bump"
 	},
 	"devDependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -11,7 +11,12 @@
 		"sandbox:push": "./sandbox.sh push",
 		"sandbox:push:ignore": "./sandbox.sh push --ignore",
 		"sandbox:push:force": "./sandbox.sh push --force",
-		"sandbox:watch": "chokidar '**/*' -i '*/node_modules' -i '.git' -c './sandbox.sh push --ignore' --initial"
+		"sandbox:watch": "chokidar '**/*' -i '*/node_modules' -i '.git' -c './sandbox.sh push --ignore' --initial",
+		"batch:build:changed": "./theme-batch-utils.sh build",
+		"batch:build:all": "./theme-batch-utils.sh build-all",
+		"batch:install": "./theme-batch-utils.sh install-dependencies",
+		"batch:audit": "./theme-batch-utils.sh audit-dependencies",
```
scruffian: running npm audit is different to running npm audit fix. This name looks to me more like its going to just do an audit not a fix too.

scruffian:
I get an error on the version bump one: 

```
fatal: Path 'refresh-2019/package.json' exists on disk, but not in '89c08750367610f229021f650c31b6703caa641b'.
undefined:1



SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at [eval]:1:6
    at Script.runInThisContext (vm.js:133:18)
    at Object.runInThisContext (vm.js:310:38)
    at internal/process/execution.js:77:19
    at [eval]-wrapper:6:22
    at evalScript (internal/process/execution.js:76:60)
    at internal/main/eval_string.js:23:3
fatal: Path 'twentynineteen-sb/package.json' exists on disk, but not in '89c08750367610f229021f650c31b6703caa641b'.
undefined:1
```

scruffian:
This looks great thank you. Just a few minor things I noticed...

package.json
pbking: ah, true.  `batch:audit:fix`?  `batch:audit-fix`?

scruffian: Both look good

pbking:
> fatal: Path 'refresh-2019/package.json' exists on disk, but not in '89c08750367610f229021f650c31b6703caa641b'.

This should be addressed now.  Ensuring package.json exists at divergence before trying to parse it.  
This can only version bump projects that are in the repo already.

jffng (commented):
All commands ran successfully in my testing, thank you! 

One comment on the version bump — it was also applied to the root themes directory package.json. If that's expected and desired, then 🚢

pbking:
Yea, bumping the root project was expected as an anticipated part of a deployment workflow.  Thanks for the look see!  Bringing it in.

PR #4479, opened 2021-08-26
corrected package.json version numbers to match style.css (when lower)

pbking:
#4478 illuminated a few cases where the package.json version was lagging behind the style.css version.  This brings all of those instances current (with the expectation that package.json version can be used as source-of-truth henceforth).

### DIFF
--- a/skatepark/package.json
+++ b/skatepark/package.json
@@ -1,6 +1,6 @@
 {
   "name": "skatepark",
-  "version": "1.1.1",
+  "version": "1.0.0",
   "description": "",
   "main": "index.php",
   "dependencies": {},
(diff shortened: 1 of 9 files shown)


jffng (approved):
Thanks for handling this. I wonder why Skatepark is 1.1.1, if we've never released it... Perhaps we can address separately if needed.

skatepark/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "skatepark",
-  "version": "1.0.0",
+  "version": "1.1.1",
```
jffng: Judging from #4492, I actually don't think we should bump this.

pbking: Indeed!  Looks like the incorrect version just came along in the initial commit for Skatepark's style.css file.

pbking:
Skatepark version has been corrected (change style.css instead of package.json which was the correct one)

This should be solid now.

jffng:
Nice, let's bring it in.

PR #4480, opened 2021-08-27
Quadrat: Make the post date a link

scruffian:
When posts don't have a title, there's no way to get to them from the index. This adds a link to the post date, so I can still access posts without a title.


MaggieCabrera:
Oh! this must be new, I don't remember it being an option. I think we should add this [here](https://github.com/Automattic/themes/blob/dc09d115fb36c9cd3bb60ac66d52cfe170c8731c/skatepark/inc/patterns/blog-posts.php#L24) and [here](https://github.com/Automattic/themes/blob/dc09d115fb36c9cd3bb60ac66d52cfe170c8731c/skatepark/block-templates/single.html#L34) too. Thank you!

MaggieCabrera:
oh wait this is Quadrat, I was thinking skatepark! should we add it to both?

MaggieCabrera (approved):
LGTM! Thank you!

PR #4481, opened 2021-08-27
Move the paragraph indent style to the first paragraph of every container

### LINKED ISSUE #4445: Skatepark: Make text-indent the default for first paragraphs
Right now, the text-indent is a block style on the paragraph block. This works, but I wonder if there's a way we could make it the default, where all first paragraphs in a container would be indented. Ideally, it would be something you can toggle on/off in the Global Styles panel. Since that isn't possible yet, maybe we could make it a customizer setting?

[ ] Indent all first paragraphs

Open to all suggestions!

scruffian:
This removes the indented block styles, and replaces them with some CSS to target the first paragraph in each container. The CSS is restricted in scope to only within the post content.

To test, try adding all the block patterns to a post.

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/4445

### DIFF
--- a/skatepark/sass/base/_text.scss
+++ b/skatepark/sass/base/_text.scss
@@ -6,4 +6,18 @@
 p {
 	margin-top: calc( 0.5 * var(--wp--custom--margin--vertical) );
 	margin-bottom: calc( 0.5 * var(--wp--custom--margin--vertical) );
-} 
+}
+
+.wp-block-post-content {
+	p:first-of-type {
+		text-indent: 15%;
+	}
+}
+
+.wp-block-post-author__content,
+.wp-block-post-comments,
+.wp-block-quote {
+	p:first-of-type {
+		text-indent: 0;
+	}
+}
--- a/skatepark/sass/base/_text.scss
+++ b/skatepark/sass/base/_text.scss
@@ -7,17 +7,3 @@ p {
 	margin-top: calc( 0.5 * var(--wp--custom--margin--vertical) );
 	margin-bottom: calc( 0.5 * var(--wp--custom--margin--vertical) );
 }
-
-.wp-block-post-content {
-	p:first-of-type {
-		text-indent: 15%;
-	}
-}
-
-.wp-block-post-author__content,
-.wp-block-post-comments,
-.wp-block-quote {
-	p:first-of-type {
-		text-indent: 0;
-	}
-}
--- a/skatepark/assets/theme.css
+++ b/skatepark/assets/theme.css
@@ -48,16 +48,6 @@ p {
 	margin-bottom: calc( 0.5 * var(--wp--custom--margin--vertical));
 }
 
-.wp-block-post-content p:first-of-type {
-	text-indent: 15%;
-}
-
-.wp-block-post-author__content p:first-of-type,
-.wp-block-post-comments p:first-of-type,
-.wp-block-quote p:first-of-type {
-	text-indent: 0;
-}
-
 .wp-block-post-author__content {
 	display: flex;
 	flex-direction: column;
--- a/skatepark/assets/theme.css
+++ b/skatepark/assets/theme.css
@@ -48,6 +48,16 @@ p {
 	margin-bottom: calc( 0.5 * var(--wp--custom--margin--vertical));
 }
 
+.wp-block-post-content p:first-of-type {
+	text-indent: 15%;
+}
+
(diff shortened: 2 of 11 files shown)


skatepark/sass/base/_text.scss
```diff
@@ -6,4 +6,18 @@
...
+
+.wp-block-post-content {
+	p:first-of-type {
+		text-indent: 15%;
+	}
+}
+
+.wp-block-post-author__content,
+.wp-block-post-comments,
+.wp-block-quote {
```
scruffian: We might need some more exceptions here.

MaggieCabrera:
The original issue mentions adding a customizer option to enable/disable this but I don't see how we can translate that to the FSE. Do you think we should add such a setting even if we don't have an FSE counterpart?

scruffian:
I don't think we should add an option for it.

skatepark/sass/base/_text.scss
MaggieCabrera: having the post comments block inside `wp-block-post-content` would be weird, would it even work?

skatepark/sass/base/_text.scss
```diff
@@ -6,4 +6,18 @@
 p {
 	margin-top: calc( 0.5 * var(--wp--custom--margin--vertical) );
 	margin-bottom: calc( 0.5 * var(--wp--custom--margin--vertical) );
-} 
\ No newline at end of file
+}
+
+.wp-block-post-content {
```
jffng: Should we add an `.is-root-container` selector to target the editor as well?

skatepark/sass/base/_text.scss
jffng: I think cover and media & text also need it...

Screen Shot 2021-08-27 at 10 59 31 AM

skatepark/sass/base/_text.scss
scruffian: I had to use `block-editor-writing-flow` as `is-root-container` is also used in the site editor.

skatepark/sass/base/_text.scss
scruffian: Done

scruffian: > having the post comments block inside wp-block-post-content would be weird, would it even work?

It would be weird but it does work. There could be cases where you wanted to encourage comments so you put the form into the post content? If you let people do it, someone will find a use for it!

skatepark/sass/base/_text.scss
MaggieCabrera: I'm not seeing the indent on the editor right now, I don't see that class on the css of this PR either

MaggieCabrera:
A few cases I found:

Cover block
Screenshot 2021-08-30 at 11 38 45

Media + Text I'm not sure if we should remove it or not. I kinda preferred to have an option to add it as a block style for the user and have it off for this block by default. What do you think?
Screenshot 2021-08-30 at 11 40 38

Pullquote
Screenshot 2021-08-30 at 11 41 39

melchoyce:
I chatted this over with @kjellr. I'm concerned about how complicated this is becoming — either we need a bunch of CSS rules to overwrite the default behavior so it works for all blocks as expected, or we offload the complexity onto users with a block style. I think we should remove the indentation entirely. It was cool to try, and it's a fun aesthetic choice, but unless we can control it natively in Gutenberg we're going to run into lots of edge cases. The theme's color and duotone support should help it stand out even without the indentation. I appreciate all the work y'all did to try this feature out! Maybe we'll be able to implement it natively in the future.

MaggieCabrera:
> I chatted this over with @kjellr. I'm concerned about how complicated this is becoming — either we need a bunch of CSS rules to overwrite the default behavior so it works for all blocks as expected, or we offload the complexity onto users with a block style. I think we should remove the indentation entirely. It was cool to try, and it's a fun aesthetic choice, but unless we can control it natively in Gutenberg we're going to run into lots of edge cases. The theme's color and duotone support should help it stand out even without the indentation. I appreciate all the work y'all did to try this feature out! Maybe we'll be able to implement it natively in the future.

After Mel's feedback, I removed the custom CSS and left the parts of this PR that were removing the text-indent block styles so we have no indentation anywhere

melchoyce:
LGTM 👍

jffng (approved):
Removed the classes from the Columns in Container pattern, this LGTM 🚢

PR #4482, opened 2021-08-27
Blockbase: Add social navigation to blockbase themes

scruffian:
This adds social navigation to all Blockbase themes:

Seedlet blocks
Screenshot 2021-10-12 at 14 09 19
Screenshot 2021-10-12 at 14 09 08

Mayland Blocks
Screenshot 2021-10-12 at 14 08 33
Screenshot 2021-10-12 at 14 08 45

Geologist
Screenshot 2021-10-12 at 14 07 30
Screenshot 2021-10-12 at 14 07 45

Quadrat
Screenshot 2021-10-12 at 14 05 54
Screenshot 2021-10-12 at 14 06 08

Blockbase:
Screenshot 2021-10-12 at 14 03 06
Screenshot 2021-10-12 at 14 03 20

Skatepark
Screenshot 2021-10-12 at 13 25 04
Screenshot 2021-10-12 at 13 25 39
Screenshot 2021-10-12 at 13 25 14

### DIFF
--- a/blockbase/inc/social-navigation.php
+++ b/blockbase/inc/social-navigation.php
@@ -13,22 +13,8 @@ function blockbase_condition_to_render_social_menu( $block_content, $block ) {
 		return false;
 	}
 
-	// The block should have a loction defined.
-	if ( empty( $block['attrs']['__unstableLocation'] ) ) {
-		return false;
-	}
-
-	// The block's location should be 'primary'.
-	if ( $block['attrs']['__unstableLocation'] !== 'primary' ) {
-		return false;
-	}
-
 	// The block should have a social links attribute.
 	if ( empty( $block['attrs']['__unstableSocialLinks'] ) ) {
-		return false;
-	}
-
-	if ( empty( $block_content ) ) {
 		return false;
 	}
 
@@ -57,11 +43,14 @@ function get_social_menu_as_social_links_block( $block ) {
 	return do_blocks( $social_links_content );
 }
 
-function append_social_links_block_to_primary_navigation( $primary_navigation, $social_links_block ) {
+function append_social_links_block( $parent_content, $social_links_block ) {
+	if ( empty( $parent_content ) ) {
+		return $social_links_block;
+	}
 	$dom = new domDocument;
 	// Since the nav block uses HTML5 element names, we need to suppress the warnings it sends when we loadHTML with HTML5 elements.
 	libxml_use_internal_errors( true );
-	$dom->loadHTML( $primary_navigation );
+	$dom->loadHTML( $parent_content );
 	$wp_block_navigation__container = $dom->getElementsByTagName('ul')->item( 0 );
 	$social_links_node = $dom->createDocumentFragment();
 	$social_links_node->appendXML( $social_links_block );
@@ -74,13 +63,13 @@ function blockbase_social_menu_render( $block_content, $block ) {
 	if ( blockbase_condition_to_render_social_menu( $block_content, $block ) ) {
 		$social_links_block = get_social_menu_as_social_links_block( $block );
 
-		return append_social_links_block_to_primary_navigation( $block_content, $social_links_block );
+		return append_social_links_block( $block_content, $social_links_block );
 	}
 
 	return $block_content;
 }
 
 /**
- * Hijack the render of the primary menu to inject a social menu.
+ * Hijack the render of the menu block to inject a social menu.
  */
 add_filter( 'render_block', 'blockbase_social_menu_render', 10, 2 );
--- a/blockbase/inc/social-navigation.php
+++ b/blockbase/inc/social-navigation.php
@@ -45,6 +45,8 @@ function get_social_menu_as_social_links_block() {
 
 function append_social_links_block_to_primary_navigation( $primary_navigation, $social_links_block ) {
 	$dom = new domDocument;
+	// Since the nav block uses HTML5 element names, we need to suppress the warnings it sends when we loadHTML with HTML5 elements.
+	libxml_use_internal_errors( true );
 	$dom->loadHTML( $primary_navigation );
 	$wp_block_navigation__container = $dom->getElementsByTagName('ul')->item( 0 );
 	$social_links_node = $dom->createDocumentFragment();
--- a/blockbase/inc/social-navigation.php
+++ b/blockbase/inc/social-navigation.php
@@ -1,0 +1,70 @@
+<?php
+
+// We should only change the render of the navigtion block
+// to social links in the following conditions.
+function blockbase_condition_to_render_social_menu( $block ) {
+	// The block should be a navigation block.
+	if ( 'core/navigation' !== $block['blockName'] ) {
+		return false;
+	}
+
+	// The theme should have a menu defined at the social location.
+	if ( ! has_nav_menu( 'social' ) ) {
+		return false;
+	}
+
+	// The block should have a loction defined.
+	if ( empty( $block['attrs']['__unstableLocation'] ) ) {
+		return false;
+	}
+
+	// The block's location should be 'primary'.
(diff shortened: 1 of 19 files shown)


jffng:
I rebased this now that #4468 is merged. Seedlet and Mayland look okay, but I have two questions about Blockbase and Quadrat:

1. How should we handle alignment? 

Screenshot 2021-08-27 at 13 13 13

One option is adding a `margin-left: auto;` to the primary nav in CSS. 

2. How should we handle the header when the responsive menu option is enabled? 

Screen Shot 2021-09-01 at 3 52 13 PM
Screen Shot 2021-09-01 at 3 54 13 PM

scruffian:
@kjellr, do you have any thoughts on this?

kjellr:
> How should we handle alignment?
> One option is adding a margin-left: auto; to the primary nav in CSS.

For desktop, I think we should do that. In general, this should look like the Quadrat screenshot above, but with more margin in between the nav + the social links:

Screen Shot 2021-09-02 at 12 13 18 PM
 
The right margin should be equal to the gap between nav items. 

> How should we handle the header when the responsive menu option is enabled?

This is another example of where we would ideally put these in the expanded hamburger menu. Part of me wonders if we should hack some temporary jQuery in there to move it there on small screens. 😅 Let me think about this on a little and get back to you later today/tomorrow.

seedlet-blocks/block-template-parts/header.html
```diff
@@ -11,6 +11,8 @@
 <!-- wp:navigation {"itemsJustification":"center","isResponsive":true,"__unstableLocation":"primary"} -->
 <!-- /wp:navigation -->
 
+<!-- wp:navigation {"orientation":"horizontal","itemsJustification":"center","isResponsive":true,"__unstableLocation":"social"} --><!-- /wp:navigation -->
```
kjellr: For Seedlet, let's use a default icon color of black (the foreground color) to match the social menu color in the classic version of the theme.

kjellr:
@melchoyce @beafialho what do you think of these options for social links placement on small screens in Blockbase and Quadrat? 

Frame 2

None of them seem ideal to me, so I've also offered [a plea for a better solution in the issue here](https://href.li/?https://github.com/WordPress/gutenberg/issues/33278#issuecomment-912743391).

beafialho:
I think the best and less invasive option is definitely the top middle one, it makes most sense for the social icons to be beneath the hamburger. The lines add way too much noise to the top area (in this case, where these themes don't use them anywhere else).

I agree that the solution you propose in [the Navigation block issue](https://github.com/WordPress/gutenberg/issues/33278#issuecomment-876411953) should be possible as a default.

kjellr:
Ok, looping back to this. It _is_ possible to put social links within a navigation block, so we should just aim to do that instead of having two separate navigation menus.

Screen Shot 2021-09-07 at 8 08 12 AM

mobile-nav

scruffian:
I updated the approach here - now the social navigation gets appended to the primary navigation block. This means that it gets wrapped by the hamburger menu when in smaller screens.

We probably need some CSS to fix the alignments of the social menu, but I'm curious to get feedback on the approach before we do that.

pbking:
I **really** like the approach.  

I kind of would prefer to have a secondary attribute to key on for appending the social menu to the block.
(i.e. `<-- wp:navigation {__unstableLocation:"primary",__unstableAppendSocial=true} -->` or the ability to set __unstableLocation multiple times, or to an array or something)

But presently I think maybe that's something the feature potentially could grow into rather than start at.

It feels like the cleanest way to achieve the design target.

Only beef I have with the code is a preference to move the menu rendering stuff to another file.

scruffian:
> Only beef I have with the code is a preference to move the menu rendering stuff to another file.

Good idea, done!

MaggieCabrera:
I'm getting this error when I apply this branch and go to the customizer and add a social menu to the "Social" location:

> ( ! ) Warning: DOMDocument::loadHTML(): Tag nav invalid in Entity, line: 1 in /Users/maggie/A8c/repos/a8c-projects/themes/blockbase/inc/social-navigation.php on line 48

blockbase/inc/social-navigation.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+// We should only change the render of the navigtion block
+// to social links in the following conditions.
+function blockbase_condition_to_render_social_menu( $block ) {
```
MaggieCabrera: A theme may want to define a menu that follows these conditions to show up in the footer too for example, or a pre-footer template part. In some cases we may want to have the social navigation but maybe not always. The perfect example of this is actually Skatepark, with this PR in place, this is what I see when I insert the pre footer block pattern:

Screenshot 2021-10-04 at 10 30 22

scruffian:
> I'm getting this error when I apply this branch and go to the customizer and add a social menu to the "Social" location:

Fixed in 21c1ed4

scruffian:
> I kind of would prefer to have a secondary attribute to key on for appending the social menu to the block.
(i.e. <-- wp:navigation {__unstableLocation:"primary",__unstableAppendSocial=true} --> or the ability to set __unstableLocation multiple times, or to an array or something)

This turned out to be a very good idea, as it means we can control which menus the links get appended to, as I have done in 3a9f0f5.

If y'all like this approach we can add it to the header nav in all blockbase themes.

MaggieCabrera:
Now it's fataling for me:
Screenshot 2021-10-05 at 18 05 27

scruffian:
I fixed the fatal, so you should be able to test now.

This revealed one shortcoming of this approach - if the primary navigation location is empty then no navigation is output, even if something is set in social...

MaggieCabrera:
> I fixed the fatal, so you should be able to test now.
> 
> This revealed one shortcoming of this approach - if the primary navigation location is empty then no navigation is output, even if something is set in social...

Yes! no more errors for me. Now that we have `__unstableSocialLinks`, why do we need to check for the other attribute? I think having `__unstableSocialLinks` is specific enough for us to know that we want to render the social links

RavanH:
Hi all, I've been testing the current social menu implementation in trunk with a local child theme and it seems to work well. Only concern I have is that (unlike item justification) it's not possible to change/unset icon color and icon size.

Is this feature considered already, and if so, is anybody working on that? If not, I could work on a proposal...

scruffian:
Does it work to set icon color and size in the theme.json?

RavanH:
@scruffian, I tried that but do not think it is possible with this block... IS there any documentation on Social Icons specifically?

scruffian:
I think we'll need something like [this](https://github.com/WordPress/gutenberg/issues/35114) for that to work.

scruffian:
For now I think we'll have have to make this a setting for the theme...

pbking:
> Now that we have __unstableSocialLinks, why do we need to check for the other attribute? I think having __unstableSocialLinks is specific enough for us to know that we want to render the social links

Pushed a small change with regard to @MaggieCabrera 's comment.  No longer requires the parent container to be parent menu; __unstableSocialLinks on a navigation block is the only requirement.  Holds true for having content; when there is no content the social links are provided (rather than appended) eliminating the need for the empty() check (and allowing the 'social links' menu target to be exclusively targeted as such:

```
<!-- wp:navigation {"__unstableSocialLinks":"social"} /-->
```

I think the existing color management (leveraging `primary` color for social links) is acceptable and appropriate for the time being.

Otherwise (IMO) this change is good-to-go once the appropriate changes have been made to all of the child themes to both add (and style) the social links.

scruffian:
I have now added social nav to all Blockbase themes, and updated the CSS to accommodate the new navigation. I have updated the screenshots in the PR description


This should also fix #4590

pbking (approved):
Boom!
All the themes look great.  Seems like a great ponyfill to get things into place.  
Diggit
Shippit

PR #4483, opened 2021-08-27
Skatepark: Add title divider

### LINKED ISSUE #4417: Skatepark: Page/post headings without featured images should have a divider

mikachan:
This PR adds a bottom border to the post title block if there is no featured image.

Screenshot 2021-08-27 at 14 09 47

I tried to do this with CSS but ended up adding a PHP function to add a class to the body instead.. not sure how we'd normally handle something like this!

#### Related issue(s):
Closes #4417.

### DIFF
--- a/skatepark/sass/blocks/_post-title.scss
+++ b/skatepark/sass/blocks/_post-title.scss
@@ -1,4 +1,4 @@
 .wp-block-post-title:not(.has-featured-image .wp-block-post-title) {
-	border-bottom: var(--wp--custom--form--border--width) var(--wp--custom--form--border--style) var(--wp--custom--form--border--color);
+	border-bottom: var(--wp--custom--form--border--width) var(--wp--custom--form--border--style) var(--wp--preset--color--primary);
 	padding-bottom: calc( var(--wp--custom--margin--vertical) * 2 );
 }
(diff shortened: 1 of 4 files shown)


MaggieCabrera:
Very clever! My only worry is that the user has no control over this if they want to remove it, but since these kinds of separators are prevalent over the design, it feels like the polygons on Quadrat, just a part of what makes this theme exactly that.

skatepark/sass/blocks/_post-title.scss
```diff
@@ -0,0 +1,4 @@
+.wp-block-post-title:not(.has-featured-image .wp-block-post-title) {
+	border-bottom: var(--wp--custom--form--border--width) var(--wp--custom--form--border--style) var(--wp--custom--form--border--color);
```
MaggieCabrera: These work, but this is not exactly a form. For the color I would directly use the foreground (or was it primary?) color so that we make sure it always matches the color of the text. Maybe the other two variables can stay, I don't think we have a "separator width" variable.

mikachan: That makes sense, I'll swap out the color then 👍

jffng (approved):
Clever indeed, I dig it.

I realized that the thickness of the separator in the mobile header didn't match, so I addressed it over in #4484.

Screen Shot 2021-08-27 at 11 10 40 AM

mikachan:
> I realized that the thickness of the separator in the mobile header didn't match, so I addressed it over in #4484.

Good spot!

Thanks both, I'll bring this in.

melchoyce:
🎉

scruffian:
>  My only worry is that the user has no control over this if they want to remove it

They can always use custom CSS if they really must

Issue #4488, opened 2021-08-29
Blockbase: duplicate font in Google fonts URL


RavanH:
## Steps to replicate

In the Customizer, select both for Body and Header fonts the same Google font and save settings.

## Result

In the page source, the URL parameters in the call to https://fonts.googleapis.com/css2 will contain the same font twice.

## Expected

One font in the URL.

Please note: this does not change the response from Google so it is not a defect, just something that can be fixed easily :)

MaggieCabrera:
Closed via #4489

Issue #4487, opened 2021-08-29
Blockbase: Child theme custom Google font always loaded


RavanH:
## Steps to replicate

Create a child theme with custom Google font set in the theme.json with a "google" attribute. Par example Mayland Blocks, see [mayland-blocks/theme.json](https://github.com/Automattic/themes/blob/e4fd001134863ab16246fbaf47dee42a687372fb/mayland-blocks/theme.json#L342) or [skatepark/theme.json](https://github.com/Automattic/themes/blob/e4fd001134863ab16246fbaf47dee42a687372fb/skatepark/theme.json#L387) for example.

## Result

The child theme font will be loaded ALWAYS instead of only when there is no user data (after first activation of the  theme) or if the user has expressly selected the font in Typography settings.

## Expected

Font to be added to the googlefonts api call only when it is actually used (default settings or user selected).

MaggieCabrera:
Closed via #4489

Issue #4458, opened 2021-08-23
Blockbase: invalid Google Fonts API call when System Font selected


RavanH:
## Steps to replicate

In the Customizer, select "System Font" for both the Body and Header fonts.

## Result

The page head will include `<link rel='stylesheet' id='blockbase-fonts-css'  href='https://fonts.googleapis.com/css2?display=swap' type='text/css' media='all' />` which will generate a 404 response from Google.

## Expected

No blockbase-fonts-css stylesheet at all.

MaggieCabrera:
Closed via #4489

PR #4489, opened 2021-08-29
Blockbase: Improve blockbase_fonts_url handling

RavanH:
This change adresses multiple issues. 

1. prevents the possibility of an "empty" call to fonts.googleapis.com by returning empty instead of `https://fonts.googleapis.com/css2?&display=swap`
2. prevents a child theme custom Google font to ALWAYS be called even if it's is not selected in the user settings
3. prevents duplicate URL parameters when both user body and header fonts are set to the same value with `array_unique`

#### Related issue(s):

1. #4458
2. #4487
3. #4488

### DIFF
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -83,26 +83,28 @@ function blockbase_fonts_url() {
 	}
 
 	$font_families = [];
-	if ( ! empty( $theme_data['typography']['fontFamilies']['theme'] ) ) {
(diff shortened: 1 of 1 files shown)


pbking:
Nice!  Cleans this feature up nicely!  I appreciate the PR very much!

pbking (approved):
Works as intended.
Fixes all the things.
Doesn't break any of the things.
🚢

PR #4492, opened 2021-08-30
Skatepark: readme.txt

MaggieCabrera:
According to the [theme directory requirements](https://make.wordpress.org/themes/handbook/review/required/) I've added the images we use on patterns to the theme and credited them accordingly on the readme.txt. I've corrected some wrong version numbers too.


MaggieCabrera:
I went ahead and passed the theme check plugin to the theme and only found an issue with an instance of blockbase as a textdomain and replaced it here too

skatepark/readme.txt
```diff
@@ -2,7 +2,7 @@
 Contributors: Automattic
 Requires at least: 5.7
```
jffng: Should this be 5.7.2 also?

jffng (commented):
One small comment, other than that LGTM!

PR #4493, opened 2021-08-30
Skatepark: refactor duotone blocks to use the new opacity setting

MaggieCabrera:

This PR applies the changes from [this change](https://github.com/WordPress/gutenberg/pull/34130) to the duotone filter. This changes the block patterns and templates that use duotone so that they don't have the green/black combination but instead they use a black/transparent combination to blend with the background. This means that when we change colors for the theme the duotone images are no longer green, but they blend in with the background.

Screenshot 2021-08-30 at 15-45-21 Duotone tests – fdsfsdf

There's still an issue with the Blue/Cream and Green/Pink color options, since in the design those are expecting the opposite combination of background/foreground colors on their duotone filter. These two combinations are not blending into the background color but instead are doing the opposite:

Screenshot 2021-08-30 at 15-45-47 Duotone tests – fdsfsdf

### DIFF
--- a/skatepark/block-templates/index.html
+++ b/skatepark/block-templates/index.html
@@ -10,7 +10,7 @@
 	<hr class="wp-block-separator is-style-wide"/>
 	<!-- /wp:separator -->
 
-	<!-- wp:post-featured-image {"isLink":true,"style":{"color":{"duotone":["#000","#B9FB9C"]}}} /-->
+	<!-- wp:post-featured-image {"isLink":true,"style":{"color":{"duotone":["#000","rgba(0, 0, 0, 0)"]}}} /-->
 
 	<!-- wp:post-title {"isLink":true,"fontSize":"normal"} /-->
 
--- a/skatepark/block-templates/page.html
+++ b/skatepark/block-templates/page.html
@@ -9,7 +9,7 @@
 
 <!-- wp:group {"align":"full","style":{"spacing":{"padding":{"bottom":"5em"}}}} -->
 <div class="wp-block-group" style="padding-bottom:5em">
-<!-- wp:post-featured-image {"align":"full","style":{"color":{"duotone":["#000","#B9FB9C"]}}} /--></div>
+<!-- wp:post-featured-image {"align":"full","style":{"color":{"duotone":["#000","rgba(0, 0, 0, 0)"]}}} /--></div>
 <!-- /wp:group -->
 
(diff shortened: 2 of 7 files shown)


melchoyce:
What are our technical options for the dark schemes?

MaggieCabrera:
> What are our technical options for the dark schemes?

with what we have right now I don't think we can provide a good solution for them. If/when https://github.com/WordPress/gutenberg/pull/34073 lands, we can use variables for the duotone settings and we could potentially swap the values of the variables when we select the darker color schemes.

kjellr:
@MaggieCabrera Is this basically a halfway step (which works for lighter colors only) until https://github.com/WordPress/gutenberg/issues/33905 is resolved? While I appreciate that it's better than what we have, it still results in a subpar experience for some users. 

_Edit:_ Oh wait, I guess it sounds like https://github.com/WordPress/gutenberg/pull/34073 would basically solve https://github.com/WordPress/gutenberg/issues/33905? Either way, I'm not sure this transparency support is all that consequential... it makes things a little better for right now, but it also doesn't quite solve the problem.

MaggieCabrera:
> @MaggieCabrera Is this basically a halfway step (which works for lighter colors only) until [WordPress/gutenberg#33905](https://github.com/WordPress/gutenberg/issues/33905) is resolved? While I appreciate that it's better than what we have, it still results in a subpar experience for some users.
> 
> _Edit:_ Oh wait, I guess it sounds like [WordPress/gutenberg#34073](https://github.com/WordPress/gutenberg/pull/34073) would basically solve [WordPress/gutenberg#33905](https://github.com/WordPress/gutenberg/issues/33905)? Either way, I'm not sure this transparency support is all that consequential... it makes things a little better for right now, but it also doesn't quite solve the problem.

yeah I agree, it's not a proper solution, but I wanted to explore and share what the transparency PR was bringing anyway, hopefully, we won't need this one. I also wanted to bring attention to the problem with the darker colors, which will definitely need a fix on the theme to switch the duotone color pairs.

melchoyce:
Thanks for exploring this, @MaggieCabrera, and finding those issues. Let's backburner this while y'all and the Gutenberg team flesh out the feature in core.

MaggieCabrera:
Closing this one in favor of the work done over at https://github.com/WordPress/gutenberg/pull/34667

PR #4491, opened 2021-08-29
Blockbase: Fix unchanged font setting reverting to default

RavanH:
#### Changes proposed in this Pull Request:

1. use current setting instead of default to fill in missing post data
2. remove unused $google_font_array
3 correct type-o `elemenets` in unset values

#### Related issue(s):

#4490

PR #4510, opened 2021-08-30
Fix unchanged font setting reverting to default

### LINKED ISSUE #4490: Blockbase: changing one Google font will drop the other
## Steps to replicate

From Blockbase with the default fonts, go to Customizer > Fonts and set both Body and Header fonts to "Arvo" (for example) and Publish. Then hit browser Reload (F5) and go back to Fonts. The Body and Header settings will be as expected. Now, change only the Body font to "Cabin" (for example) and Publish. Now Reload the browser and move into Fonts again. The Body font will now be as expected but the Header font will be back at "System Font".

## Result

The font that was NOT changed is lost after the second Save.

## Expected

Both font options to be saved, not just the one that got changed while the other is reset to default.

pbking:
These changes fix issue #4490

1. use current setting instead of default to fill in missing post data
2. remove unused $google_font_array
3 correct type-o `elemenets` in unset values

---

This was tweaked from the [original PR](https://github.com/Automattic/themes/pull/4491) to further address the "unset" logic (so that the correct values were actually being unset) and to continue leveraging the default value (instead of current) for states when the value is not provided.

#### Related issue(s):
#4490

### DIFF
--- a/blockbase/inc/customizer/wp-customize-fonts.php
+++ b/blockbase/inc/customizer/wp-customize-fonts.php
@@ -386,15 +386,18 @@ class GlobalStylesFontsCustomizer {
 			return;
 		}
 
+		$body_current    = $wp_customize->get_setting( $this->section_key . 'body' )->value();
+		$heading_current = $wp_customize->get_setting( $this->section_key . 'heading' )->value();
+
 		$body_default    = $wp_customize->get_setting( $this->section_key . 'body' )->default;
 		$heading_default = $wp_customize->get_setting( $this->section_key . 'heading' )->default;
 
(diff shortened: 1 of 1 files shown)


jffng (approved):
I tested this on trunk and could replicate the error. Checking out this PR fixes the issue. Thanks!

RavanH:
Excellent! 

I must admit I had my doubts about the unsets there too but did no go into it further than the type-o... However, my doubts were about the logic of the unsets, not about their keys: I was wonderding why only unsetting all when both inputs are "default", not some when one of them equals default?

PR #4511, opened 2021-08-30
Remove block attributes from site-title in header

pbking:
Removed block attributes from site-title in header as they are set in theme.json (and some of those items removed were incorrect)

Before:
image

After:
image


jffng (approved):
This indeed makes things look closer to the design, thanks!

I do remember adding them as attributes because I wasn't sure if we wanted _all_ site title blocks to appear that way (as is the case when we style via theme.json).

That said, I think this is preferred.

pbking:
I actually didn't change theme.json so all instances were already looking that way. :P

PR #4514, opened 2021-08-31
Morden: fix hover color for social navigation icons

### LINKED ISSUE #4513: Morden: social icons invisble on hover
## Steps to replicate

1. Head to https://mordendemo.wordpress.com
2. Hover on a social icon

Screenshot: 
invisible

Video:
https://user-images.githubusercontent.com/36608681/131477806-b58a6e95-80e8-429b-8013-d1d3905a3c22.mp4

## Result

The icon becomes invisible against the background

## Expected

Judging from the CSS, it's supposed to be a slightly transparent white.

* User report: 4256280-zen 
* User does not have access to use custom CSS to override it

I've been poking at it, and so far I see:

1. This is the CSS it uses:
```
.social-navigation a:hover {
	color: rgba(var(--wp--preset--color--background), 0.8);
} 
```
2. The variable as set in --root is fine
3. manually setting the `.social-navigation a:hover` value to `255,255,255,0.8` is fine
4. It's just that for some reason the opacity on the variable on hover just goes NOPE and the result is the social link takes the same color as the `#masthead` element - which, by the way, is using `var(--wp--preset--color--foreground)` as its background color.
5. The SVG and path elements are fine, they use `currentColor` and seem to happily switch if `.social-navigation a:hover` is set to literally anything other than a variable + opacity; even just a variable is fine. 

I found that adding opacity separately works fine:
opacity

I also learned about `filter: brightness(0.8);` which could also fix it - no need to specify either the color variable, or opacity; just throw a filter on it:
filter

I don't know if `filter:` is whitelisted for dotcom yet but it looks compatible enough across browsers
https://caniuse.com/css-filters

sophiegyo:
Issue:
Morden social navigation icons become invisible when hovered on.
Cause: seems that the rgba value using both a variable and opacity causes the theme to nope out - icons become invisible:

invisible

Removing this:

```
.social-navigation a:hover {
    color: rgba(var(--wp--preset--color--background),.8);
}
```

Adding this:

```
.social-navigation a:hover {
	opacity: 0.8;
}
```

opacity

#### Related issue(s):
#4513 4513

### DIFF
--- a/morden/style.css
+++ b/morden/style.css
@@ -3094,7 +3094,7 @@ body:not(.fse-enabled) .main-navigation a {
 }
 
 .social-navigation a:hover {
-	color: rgba(var(--wp--preset--color--background), 0.8);
+	opacity: 0.8;
 }
 
 .social-navigation svg {


scruffian:
LGTM

Issue #4531, opened 2021-09-06
Geologist: enhancements and code quality


MaggieCabrera:
After a bit of discussion on slack with @scruffian there's a few things we can polish for Geologist. Right now the theme is directly duplicating Quadrat's css and that could be improved. There's three immediate areas we could work on:

- The header styling could live in Blockbase by [unifying these themes header styles](https://github.com/Automattic/themes/pull/4482). 
- The buttons styles for Skatepark and Quadrat+Geologist are a design pattern that is very popular. We could include that css in Blockbase without importing it to the main stylesheet, so that the parent theme provides the possibility to import these button styles without adding to the original theme's stylesheet.
- We should remove the Quadrat block pattern's CSS that is unused in Geologist. @kjellr will Geologist have patterns of its own?

kjellr:
> We should remove the Quadrat block pattern's CSS that is unused in Geologist. @kjellr will Geologist have patterns of its own?

It probably will, but we haven't designed them yet. I'll add it to my todo list. 👍

MaggieCabrera:
I'm closing this, since the only thing left is the header issue and that is being tracked already

PR #4532, opened 2021-09-06
Geologist: code quality

MaggieCabrera:

This PR refactors the outline button styles into blockbase so we don't need to repeat the code for every child theme that wants to use these kinds of buttons. We are also removing the block patterns css from Geologist.

To test this check that the buttons still look the same on all child themes. While I was working on this I built all children and only the block patterns css is being removed from Geologist, the rest of the child themes' css are unaffected.

After looking at the rest of the code I think we can remove some css from the post comments block that is already on blockbase, I will be looking into that tomorrow.

#### Related issue(s):

Partially addresses https://github.com/Automattic/themes/issues/4531


scruffian:
Thanks for taking care of this.

PR #4534, opened 2021-09-07
Blockbase: Fix "Blog pages show at most" reading setting

### LINKED ISSUE #4521: Blockbase: Reading setting "Blog pages show at most" is ignored
## Steps to replicate

Go to Settings > Reading and change the "Blog pages show at most" setting to anything other than 10.

## Result

10 latest posts on the blog (or home) page.

## Expected

The number of posts set on Settings > Reading.

This is caused by the perPage in `,"query":{"perPage":10,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":""}` in the wp:query block in block-templates/index.html

I would suggest removing the "query" parameter all together, letting the main blog query default to general WordPress settings. This may also allow plugins to alter the blog query (untested) as users might expect them to do.

mikachan:
This PR addresses the issue described in #4521, where Blockbase is overwriting the "Blog pages show at most" reading setting because `perPage` is set to 10 in the `query` parameter on the index template.

I've removed the `queryId` and `query` parameters as suggested by @RavanH. As far as I can see, this doesn't have any negative effects, and it means that the number saved in the "Blog pages show at most" setting is respected.

Here is Blockbase with a "Blog pages show at most" set to 1, which previously would have been ignored:

Screenshot 2021-09-07 at 11 39 58

#### Related issue(s):
Fixes #4521.

### DIFF
--- a/blockbase/block-templates/index.html
+++ b/blockbase/block-templates/index.html
@@ -1,6 +1,6 @@
 <!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
-<!-- wp:query {"tagName":"main","queryId":1,"query":{"perPage":10,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":""}} -->
+<!-- wp:query {"tagName":"main"} -->
 <main class="wp-block-query">
 <!-- wp:post-template -->
 <!-- wp:group {"layout":{"inherit":true}} -->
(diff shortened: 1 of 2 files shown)


mikachan:
Should the same change be made to the query block on the search.html template? I've tried removing the `queryId` and `query` parameters and it doesn't seem to make a difference (everything still works). Although, it looks like the reading setting is respected in the search template, regardless of the `perPage` setting in the query block.

scruffian:
> Should the same change be made to the query block on the search.html template?

Makes sense to me

scruffian (approved):
LGTM. Feel free to merge once you've updated the search template.

PR #4536, opened 2021-09-07
Quadrat: Add pagination arrows through block settings, not in CSS

scruffian:
Now that https://github.com/WordPress/gutenberg/pull/33656 has merged we can set pagination arrows in block attributes. This is better for RTL and customization.

### DIFF
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -419,21 +419,6 @@ ul ul {
 	border-top: 1px solid var(--wp--custom--color--primary);
 }
 
-.wp-block-query-pagination .wp-block-query-pagination-previous::before {
-	content: "←";
-	margin-right: 0.5em;
-}
-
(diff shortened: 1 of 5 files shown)


MaggieCabrera:
This works as intended! Thank you :D 

Just two things: we should do the same for Geologist and I think we shouldn't merge this on trunk since this will only work on 11.5. Now that we are close to getting these themes on dotcom we should decide on a strategy when we change things that have been recently merged into GB.

MaggieCabrera (approved):

jeffikus:
Great work on this. Solid example of identifying something in a theme, contributing back to Gutenberg, then implementing in the theme! High five! :) 🙌

pbking:
This is working well on wpcom now that 11.5 has been installed.  Removing the DNM tag for inclusion in the next deployment.

PR #4540, opened 2021-09-07
Blockbase: Remove wpcom specific code

scruffian:
In 64-gh-Automattic/analytics we changed the way we load the stats gif so that its always hidden. This means we no longer need CSS in the theme to hide the image.

I plan to do the other themes in another PR


MaggieCabrera (approved):
🚢

PR #4542, opened 2021-09-08
Blockbase: change google url for Fira

### LINKED ISSUE #4541: Blockbase: Fira sans font family not loading
Seedlet blocks' body font-family is not loading at all and it's falling back to a sans-serif system font:

Screenshot 2021-09-07 at 17 33 13

I think this has to do with the shorthand format for the google font import not working for this font in particular, though I didn't find out why.

MaggieCabrera:

The Fira Sans font was not being applied because it doesn't support the v2 google font API format. I used the older format to fix it and commented this for future reference.

To test this check that the correct font is applied. Fira Sans is the default body font for Seedlet blocks.

Screenshot 2021-09-08 at 09 53 50

I went ahead and checked the rest of the fonts on the list and they seemed to be working fine.

#### Related issue(s):

Closes https://github.com/Automattic/themes/issues/4541

### DIFF
--- a/blockbase/inc/customizer/wp-customize-fonts.php
+++ b/blockbase/inc/customizer/wp-customize-fonts.php
@@ -12,6 +12,8 @@ class GlobalStylesFontsCustomizer {
 	private $font_control_default_body;
 	private $font_control_default_heading;
 
+	//Not all fonts support v2 of the API that allows for the shorter URls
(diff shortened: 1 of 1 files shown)


danieldudzic (approved):
Had some caching issues trying to test this, but after a few tries, I was able to verify that it works fine :)

PR #4543, opened 2021-09-08
Mayland Blocks, Blockbase: fix header vertical spacing

MaggieCabrera:
After some of our latest changes we've lost all vertical spacing on the header of Mayland Blocks and Blockbase. I'm opening this PR to fix it separate from https://github.com/Automattic/themes/pull/4482 since that may take a bit longer to fix and this looks pretty bad right now

Before:

Screenshot 2021-09-08 at 10 00 12
Screenshot 2021-09-08 at 09 58 55

After:

Screenshot 2021-09-08 at 10 10 38
Screenshot 2021-09-08 at 10 09 56

/cc @kjellr


danieldudzic (approved):
Works 👍

PR #4546, opened 2021-09-08
Seedlet: Remove wpstats image code

scruffian:
Since we hide the #wpstats image by default now (64-gh-Automattic/analytics) we longer need this CSS. This PR just removes it from Seedlet which will also impact all the Seedlet family themes.

To test, make sure you still can't see the image on wordpress.com sites.

#### Related issue(s):
https://github.com/Automattic/themes/pull/4540


pbking (approved):
👍 
I see Spearhead doesn't have it so no changes need there.
Works as expected.  (As in, I don't see anything different; It's still gone.)
🚢

PR #4548, opened 2021-09-08
Blockbase: Remove wpcom specific CSS file

scruffian:
This removes the wpcom specific CSS file and refactors the post comments block CSS so that it doesn't have an impact on wpcom.

I would rather we didn't have any wpcom specific code in the repo at all, but I can't find another way to remove this line.

To test:
- On a dotorg site, check that the comments form has grid applied
- On a dotcom site, check that the comments form doesn't have any grid code applied

### DIFF
--- a/blockbase/sass/blocks/_post-comments.scss
+++ b/blockbase/sass/blocks/_post-comments.scss
@@ -1,3 +1,48 @@
+body:not(.highlander-enabled) {
+	.wp-block-post-comments {
+		form {
+			display: grid;
+			column-gap: 1em;
+			grid-template-rows: auto;
+			grid-template-areas:
+				"notes notes"
+				"author author"
+				"email url"
+				"comment comment"
+				"cookies-consent cookies-consent"
+				"form-submit form-submit";
+
+			.comment-notes {
+				grid-area: notes;
+			}
+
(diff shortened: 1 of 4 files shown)


blockbase/sass/blocks/_post-comments.scss
```diff
@@ -1,3 +1,48 @@
+body:not(.highlander-enabled) {
+	.wp-block-post-comments {
+		form {
+			display: grid;
```
MaggieCabrera, line 4: I think we only need to cover this line, but I don't think it matters much that we added the rest so I'm not too bothered by it

MaggieCabrera (approved):
this LGTM, the only alternative I can think of is the highlander form not using the same classes as the dotorg comments block but that would break a lot of themes. Maybe the post comments markup will change in the future when it gets reworked and it will have different classes

blockbase/sass/blocks/_post-comments.scss
scruffian: I prefered keeping all the grid CSS together for readability.

MaggieCabrera: makes sense, works for me!

PR #4556, opened 2021-09-08
Simplify Custom margin

### LINKED ISSUE #4527: Quadrat: Mobile header spacing bug
The Quadrat header appears incorrectly placed on mobile viewports:

Current|Intended
---|--- 
current|new

The issue is both the spacing on top of the title/menu button, and also the spacing in between the title and the tagline.

pbking:
While looking into #4527 I got pretty turned around looking at padding.

This is my suggestion for how we could simplify that.

We are using two values to represent "space between stuff"
`--wp--custom--margin--horizontal` (and `--vertical`) as well as `--wp--custom--post-content--padding--left` (and `--right`).  In some themes the value of post-content padding was based off of custom margin (and some not).  Some of the values are responsive and some are not.  I found it all a bit confusing.

This change consolidates the "space between stuff" to JUST `--wp--custom--gap--horizontal` (and `--vertical`).  The default in blockbase is responsive `min(30px, 5vw)`.

I also found the padding introduced to (most of) the headers was more complicated than it needed to be.  Most were align-full which "busted out" of the site margins and then re-introduced those margins unnecessarily. This change simplifies the headers somewhat by removing the align-full.  Since the headers aren't in the content block there's no default behavior to be "content width" so it naturally expands to the (margined) page width which seems to be the default desired effect.

### DIFF
--- a/mayland-blocks/child-theme.json
+++ b/mayland-blocks/child-theme.json
@@ -79,7 +79,8 @@
 				"selection": "var(--wp--preset--color--selection)"
 			},
 			"margin": {
-				"horizontal": "32px"
+				"horizontal": "32px",
+				"vertical": "30px"
 			},
 			"width": {
 				"default": "750px",
--- a/mayland-blocks/child-theme.json
+++ b/mayland-blocks/child-theme.json
@@ -79,8 +79,7 @@
 				"selection": "var(--wp--preset--color--selection)"
 			},
 			"margin": {
-				"horizontal": "32px",
-				"vertical": "30px"
+				"horizontal": "min(32px, 5vw)"
 			},
 			"width": {
 				"default": "750px",
--- a/mayland-blocks/child-theme.json
+++ b/mayland-blocks/child-theme.json
@@ -78,7 +78,7 @@
 				"background": "var(--wp--preset--color--background)",
 				"selection": "var(--wp--preset--color--selection)"
 			},
-			"margin": {
+			"gap": {
 				"horizontal": "min(32px, 5vw)"
 			},
 			"width": {
--- a/quadrat/sass/templates/_header.scss
+++ b/quadrat/sass/templates/_header.scss
@@ -4,11 +4,7 @@
 
 .site-header {
 	overflow: inherit;
-	padding: 10px 0 60px; // TODO: Maybe replace with a responsive custom variable?
-
-	@include break-medium() {
-		padding: var(--wp--custom--post-content--padding--left);
-	}
+	padding-top: var(--wp--custom--margin--vertical);
 
 	.wp-block-site-logo {
 		margin-right: var(--wp--custom--margin--horizontal);
--- a/quadrat/sass/templates/_header.scss
+++ b/quadrat/sass/templates/_header.scss
@@ -2,14 +2,14 @@
 	position: relative; // This is needed so that the polygon is stretched across the whole site.
 }
 
-.site-header {
+.wp-site-blocks .site-header {
 	overflow: inherit;
 	padding-top: var(--wp--custom--margin--vertical);
 
 	// The blockGap is used HORIZONTALLY when the viewport is LARGE (in which case the value defined in theme.json is appropriate) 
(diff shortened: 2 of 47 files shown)


pbking:
This change also forces a smaller gap for the header on small screens now which means this change should fully address #4527

danieldudzic:
This seems comprehensive. Do you have any tips on how to test this thoroughly?

I did a quick test so far, and it does solve the original issue: #4527

quadrat/sass/templates/_header.scss
```diff
@@ -4,12 +4,13 @@
...
 	overflow: inherit;
-	padding: 10px 0 60px; // TODO: Maybe replace with a responsive custom variable?
+	padding-top: var(--wp--custom--margin--vertical);
 
-	@include break-medium() {
-		padding: var(--wp--custom--post-content--padding--left);
+	// The blockGap is used HORIZONTALLY when the viewport is LARGE (in which case the value defined in theme.json is appropriate) 
+	// It needs to be a different value then when the viewport is SMALL and the gap is used VERTICALLY 
+	@include break-small-only() {
+		gap: 2px !important; // !important because Global Styles, which supplies the gap value, is loaded after theme CSS and wins the specificity battle
```
scruffian: Rather than `!important` can we make the selector more specific?

scruffian:
I think the reasoning for adding padding for post content was to handle the case where padding on the post content block is different to the padding around site content. Padding on the post content block can be defined in the theme.json; if you do that then alignfull blocks no longer go to the edge of the window. I'm not sure the best way to handle that case.

quadrat/sass/templates/_header.scss
pbking: I had a higher specificity hack (double-class) initially but the intent didn't seem as obvious.  There's probably a better way than that though.

scruffian: We could use the .wp-site-blocks class

pbking:
> Padding on the post content block can be defined in the theme.json

We weren't actually adding padding to the post-content block though, we were just generating a custom variable we were CALLING post-content and applying it to something that wasn't actually the post-content block.  Which was no small part of my confusion I think (even though I probably was the originator of the confusing naming...)

quadrat/sass/templates/_header.scss
pbking: done

mayland-blocks/child-theme.json
```diff
@@ -79,7 +79,8 @@
 				"selection": "var(--wp--preset--color--selection)"
 			},
 			"margin": {
-				"horizontal": "32px"
+				"horizontal": "32px",
```
jffng: Should mayland blocks get the responsive value also? e.g. 

```suggestion
				"horizontal": "min(32px, 5vw",
```

jffng (commented):
Thank you for revisiting this, the inconsistency of the spacing variable names + how they are applied has bothered me too. 

> "space between stuff"

What do you think about renaming `margin` to `gap` to align with the Gutenberg gap concept? Also since we are using it for both margin and padding in some places.

pbking:
> What do you think about renaming margin to gap to align with the Gutenberg gap concept? Also since we are using it for both margin and padding in some places.

Yes absolutely!  That further clarifies the intent and perhaps the eventual exit strategy.

mayland-blocks/child-theme.json
pbking: Sure.

pbking:
Ok, I have refactored --custom--margin to --custom--gap.  It was a big-ish change as there were quite a few references to --custom--margin but --gap is much clearer everywhere it's used and I like the change.

I believe this is ready for another look see.

jffng (approved):
I removed some of the now defunct margin variables from skatepark and gave this a test across BB & Co. LGTM.

I think this is a nice change that solves the original issue and simplifies our variable naming 👍 

I noticed some issues with Quadrat's padding added directly in the templates, and opened a follow on PR here: https://github.com/Automattic/themes/pull/4574

PR #4557, opened 2021-09-08
Moved gap between site logo and title from GRID to IMAGE

pbking:
Before:
image
image
image
image

After:
image
image
image
image


jffng (approved):
Will need to be reworked to use the new `gap` variable instead.

PR #4562, opened 2021-09-09
Blockbase: Remove code block CSS

scruffian:
Since https://github.com/WordPress/gutenberg/issues/31135 got merged, we can set the typeography for the code block in theme.json.

Monospace is the default for `code` elements in Chrome (and maybe other browsers) so maybe this is unnecessary, but it's probably better to be explicit?


pbking (approved):
Looks good to me.  The issue pointed out [here](https://github.com/WordPress/gutenberg/pull/31742#discussion_r650724928) doesn't really effect things for us.

PR #4563, opened 2021-09-09
Blockbase: Remove wpcom hook

scruffian:
This removes the wpcom hook, since we are moving it to the wpcom codebase.

See D66600-code


pbking:
Are we sure we want to relinquish control of the customizer?  Would this mean that ALL block-based themes have access to the customizer (or, at some point in the future, none do?) 
Will we only have universal block-based themes on wpcom?  Or will there possibly be non-universal themes where access to the customizer isn't necessary relevant. 
This would also eliminate the usage of the customizer for .org usage of Blockbase which I would argue we don't want to do.  That feels like a decision for the theme, not the environment.

Update: Nevermind. :P
I was confusing this with[ this call ](https://github.com/Automattic/themes/blob/trunk/blockbase/functions.php#L114)which handles the menu.

MaggieCabrera:
> This would also eliminate the usage of the customizer for .org usage of Blockbase which I would argue we don't want to do. That feels like a decision for the theme, not the environment.

I'm thinking that maybe we don't want to remove this line after all, just move it to functions.php, and if dotcom wants to remove the customizer link from these themes, then they should override this.

pbking:
Looks like there is still some ongoing discussion regarding the associated diff (and a minor conflict).  Moving back to in-progress until it's resolved.

pbking:
Resolved the conflict by removing the unused file.
Also removed the wpcom CSS file as that was already not being used (the reference was removed previously but the file was not).

Tested on my sandbox and since the Phabricator diff is already applied this worked as expected.

🚢

pbking (approved):

PR #4565, opened 2021-09-09
Blockbase: Add post-meta template part

### LINKED ISSUE #4522: Blockbase: remove unused post meta styling
## Proposal

There are several style rules and files that are aimed at a block with class `.post-meta`. There is no such block used in Blockbase. However, these styles are used in child themes Mayland Blocks and Skatepark. I would argue that these style belong there, not in the parent theme.

Files involved are:
sass/post/_meta.scss
sass/ponyfill.scss (last import)
assets/svg/post-author.svg
assets/svg/post-category.svg
assets/svg/post-date.svg
assets/svg/post-tag.svg (already missing!)

Of course, the child themes in question would need to set up their own version of:

```
	.wp-block-post-author,
	.wp-block-post-date,
	.wp-block-post-tags,
	.wp-block-post-hierarchical-terms {
		display: flex;
		margin-right: calc(2 * var(--wp--custom--margin--baseline) );
	}

	.wp-block-post-author:before,
	.wp-block-post-date:before,
	.wp-block-post-tags:before,
	.wp-block-post-hierarchical-terms:before {
		align-self: center;
		content: '';
		display: inline-block;
		margin-right: calc(0.5 * var(--wp--custom--margin--baseline) );
		height: 16px;
		width: 16px;
		background-color: currentColor;
	}

```

Alternative: fix missing post-tag.svg, fix `.wp-block-post-tags` (should be `.wp-block-post-terms`) and move the scss out of the post-meta class so it will work independently for the Post Author/Date/Categories/Tags blocks that can be added in posts or pages.

mikachan:
This PR adds a post-meta template part to Blockbase.

Screenshot 2021-09-09 at 12 26 52

I've mostly copied the Seedlet Blocks, Mayland Blocks, and Skatepark post-meta. I've used the 'tiny' font size; I did try 'small' as well but I thought 'tiny' probably looked better overall.

@kjellr, I know you mentioned in the original issue that we could maybe just include the post date to start with. Let me know if you think we should use different content - for now, I've gone with: author, date, categories, tags.

We're missing the correct 'post-tag' SVG icon (I think) - I've used an outline version of the 'offer' Material icon, but I think the line thickness is slightly different from the other icons. I'm not sure how we usually create the SVGs, so just pointing out that this one probably needs creating/replacing!

#### Related issue(s):
Closes #4522.


MaggieCabrera:
Blockbase was initially a clone of Blank Canvas, which is a child of Seedlet with a few changes, so I think copying what's on Seedlet blocks should be ok, unless Kjell thinks otherwise

mikachan:
> Blockbase was initially a clone of Blank Canvas, which is a child of Seedlet with a few changes, so I think copying what's on Seedlet blocks should be ok, unless Kjell thinks otherwise

Ok! I've removed the tags as this then matches Seedlet Blocks exactly, but happy to adjust either way.

kjellr:
I think including tags is fine (I'm not sure why Seedlet Blocks doesn't display them? [The original Seedlet does](https://cloudup.com/c4nKFyhe5jc), we just don't use tags on our demo site). You can source the icon [straight from Seedlet](https://github.com/Automattic/themes/blob/1aa5058bcff44951877c2db4f461fd4a46433613/seedlet/classes/class-seedlet-svg-icons.php#L102). 

I'd also remove the separator:

Before|After
---|---
Screen Shot 2021-09-13 at 7 59 28 AM|Screen Shot 2021-09-13 at 7 58 59 AM

Also, in my testing, the comments are showing up really close to this new meta area. Not sure if that's just a bug with my test site? 

Screen Shot 2021-09-13 at 8 00 29 AM

mikachan:
Thanks @kjellr! I've added the post tags back in, copied the post tag SVG over from Seedlet (this looks great, thanks for pointing me in the right direction), and removed the separator.

Seedlet Blocks doesn't include the tags in the post meta template. I could add them here too while I'm working in this area, if you think they should be there?

The comments aren't really close on my local dev, looks like they're inheriting the top margin from this selector: `.wp-site-blocks > * + *` - 

Screenshot 2021-09-13 at 14 21 52

They're using the `--wp--style--block-gap` variable, so I don't know if this is related to a Gutenberg update, or maybe my local version is out of date.. Will have a play with this now.

kjellr:
> Seedlet Blocks doesn't include the tags in the post meta template. I could add them here too while I'm working in this area, if you think they should be there?

Yeah, I think they should be there. Feel free to do it here if you'd like, but it could totally be a separate PR if that's easier.

kjellr:
> The comments aren't really close on my local dev, looks like they're inheriting the top margin from this selector: .wp-site-blocks > * + * -
> They're using the --wp--style--block-gap variable, so I don't know if this is related to a Gutenberg update, or maybe my local version is out of date.. Will have a play with this now.

If we could make this at least `calc(2 * var(--wp--style--block-gap))`, that would be great.

mikachan:
I've updated Seedlet Blocks so the post meta now includes the tags:

Screenshot 2021-09-13 at 16 58 16

And I managed to replicate the lack of spacing above the post comments block (my Gutenberg was out of date). I've added a spacer block above the post comments on Blockbase that's the same height as `calc(2 * var(--wp--style--block-gap))`. I added it as a spacer as all the child themes use a spacer block above the comments, and it also means we don't need to change the CSS for Blockbase and all the children.

Screenshot 2021-09-13 at 17 00 47

This is ready for another review, thanks!

kjellr:
Thanks, @mikachan! Just one minor comment: Can we increase the gap between items to `calc(2 * var(--wp--custom--margin--baseline))` in Blockbase? This helps it match the gap size that Seedlet & Blank Canvas use normally. 

Before|After
---|---
Screen Shot 2021-09-14 at 8 08 56 AM|Screen Shot 2021-09-14 at 8 17 41 AM


After that, I think this is all set for now!

kjellr:
> Can we increase the gap between items...

For what it's worth, I think the theme needs to opt into [the new gap feature](https://href.li/?https://github.com/WordPress/gutenberg/pull/33991) for that to work? I'm not sure how to actually set the gap for the block though. 🤔

danieldudzic:
> > Can we increase the gap between items...
> 
> For what it's worth, I think the theme needs to opt into [the new gap feature](https://href.li/?https://github.com/WordPress/gutenberg/pull/33991) for that to work? I'm not sure how to actually set the gap for the block though. 🤔

So, I have checked the following blocks: `post-author`, `post-date ` and `post-terms` - and they seem to be missing support for `gap` in the core.

If my understanding is correct, gap makes an obvious sense with blocks that will be used in multiple instances together, like `buttons` or `columns`.

With `post-author`, `post-date` and `post-terms` being individual blocks I can see how `gap` might be tricky.

It would almost make sense to have just one block called post meta, and then have different parts inside of the block being `post-author`, `post-date` and `post-terms`. That would make `gap` super useful.

kjellr:
I _think_ you only need gap support on the Group block, right? Since that's the one that has the `flex` style applied to it?

mikachan:
> For what it's worth, I think the theme needs to opt into the new gap feature for that to work?

Thank you! I tried with spacers and then resorted to CSS... but setting `blockGap` to true in theme.json seems to have done the trick. 

I've set the Blockbase gap size to `calc(2 * var(--wp--custom--margin--baseline))`, and then set all the children gap sizes to `var(--wp--custom--gap--vertical)`, as this was the previous value in theme.json. However, `blockGap` wasn't previously enabled, so all the themes were using the default gap size of `0.5em` (which comes from Gutenberg I guess). This means all gap sizes are bigger now - would it be better to set them all to the old default value so there's no change?

Also, I could only seem to get `blockGap` working if it was applied globally (`styles.spacing.blockGap`), and I couldn't get it to work if I tried to set it directly on a block (e.g. `styles.blocks.core/post-meta.spacing.blockGap`). Or I was doing it wrong!

kjellr:
> However, blockGap wasn't previously enabled, so all the themes were using the default gap size of 0.5em (which comes from Gutenberg I guess). This means all gap sizes are bigger now - would it be better to set them all to the old default value so there's no change?

I _think_ it probably makes sense to keep the default size the same for now. I'm not sure how many places that gap property is used, so I think a more conservative approach is warranted here.

mikachan:
Thanks! I've set all of the Blockbase children gap sizes to `0.5em`. Ready for another review.

scruffian:
This now looking a bit odd on Seedlet: 
Screenshot 2021-09-15 at 18 51 04

Could all the children just inherit from Blockbase?

mikachan:
> Could all the children just inherit from Blockbase?

Yeah I think they can. I didn't do this originally as I only tried `var(--wp--custom--gap--vertical)` globally, which looked too big on most themes. I've just changed all themes to use `calc(2 * var(--wp--custom--gap--baseline))`, the same as Blockbase, and this seems to work well across them all. I'm not sure why I didn't try this before, seems like the obvious choice out of the gap sizes! Here's what this is looking like after this change:

Blockbase:

image

Seedlet Blocks:

image

Mayland Blocks:

image

Quadrat:

image

(This doesn't affect Skatepark and Geologist.)

kjellr:
This looks great. One last question: Is it possible to specify a different vertical/horizontal gaps? 

If these meta items wrap, it uses the same horizontal gap for the vertical spacing: 

Screen Shot 2021-09-16 at 7 44 37 AM

I'd prefer if the vertical spacing was half of that: 

Screen Shot 2021-09-16 at 7 44 37 AM

If that's not possible yet, it's ok, and we can merge this as-is for now. Thank you!

mikachan:
> Is it possible to specify a different vertical/horizontal gaps?

From what I can tell, the new blockGap feature only supports one value, and can't yet split out different horizontal and vertical gaps. It looks like there's an ongoing issue for it though: https://github.com/WordPress/gutenberg/issues/34529 (and there's an issue from @scruffian here too: https://github.com/WordPress/gutenberg/issues/34347)

We could set the vertical gap for the post-meta with CSS until this feature is available? Although it looks like it's tricky to overwrite the specific `.wp-container-ID` styles, so it might have to be something like:

```
.post-meta {
	row-gap: var(--wp--custom--gap--baseline) !important;
}
```

kjellr:
Maybe we could set it via CSS just for the post meta?

mikachan:
Yeah I think this works, this is how it looks:

Blockbase:

image

Seedlet Blocks:

image

Mayland Blocks:

image

blockbase/sass/post/_meta.scss
```diff
@@ -1,4 +1,6 @@
 .post-meta {
+	row-gap: var(--wp--custom--gap--baseline) !important;
+
```
kjellr: Might make sense to add a code comment here to explain why we're using `!important`. And we can also probably remove that empty line below.

kjellr (approved):
Left one minor comment, but otherwise I think this is good to go. Thank you! 🙌

blockbase/sass/post/_meta.scss
```diff
@@ -1,4 +1,5 @@
 .post-meta {
+	row-gap: var(--wp--custom--gap--baseline) !important; // !important is needed to override the unique .wp-container-ID classes
```
scruffian, line 4: It would be good to link the GB issue that suggests different vertical and horizontal block gaps, so we can remove this when that gets fixed.

mikachan: Ah yeah good idea, will sort this now

scruffian (approved):
LGTM

PR #4570, opened 2021-09-10
Blockbase + Quadrat: remap quadrat colors, refactor selection to tertiary

MaggieCabrera:

Alternative to https://github.com/Automattic/themes/pull/4474. After our conversation and @kjellr clearing up [the relationship between semantic colors](https://github.com/Automattic/themes/pull/4474#issuecomment-916315963), I think this is the way to go. What I've done:

- I renamed selection to tertiary for all themes, since it doesn't make sense to call it selection (and I never really liked that name since it implied that color was used just for that)
- Remapped secondary to tertiary for both Geologist and Quadrat, since tertiary contrasts with the foreground color, not the background one.
- Remapped tertiary to background instead of primary on Skatepark for the same reason.

And this is the result for the Quadrat patterns on the rest of the themes (yes, apparently Mayland's tertiary color is that light blue, who knew?)

Screenshot 2021-09-10 at 10 17 48
Screenshot 2021-09-10 at 10 17 43
Screenshot 2021-09-10 at 10 17 38
Screenshot 2021-09-10 at 10 17 33
Screenshot 2021-09-10 at 10 17 26

I would like your ideas on the best way to document this for future users of Blockbase. Maybe it can go on the colors.scss comment at the top?


MaggieCabrera:
Also, this won't work on existing patterns, you have to reinsert them again, using the new classes, I'm glad we are doing this before launch, else it would have been a big problem

scruffian (approved):
LGTM. Teamwork FTW!

mikachan:
This looks awesome to me! I really like renaming selection to tertiary too.

> I would like your ideas on the best way to document this for future users of Blockbase. Maybe it can go on the colors.scss comment at the top?

A comment in colors.scss sounds good for this, and maybe also mention it in the main readme file, around the 'Building a Block Child Theme' section? I think it would help to include the contrast rules somewhere for new users.

pbking:
I'm seeing some strangeness with the situation where the colors have already been customized for Quadrat (on trunk) then when switching to this branch the customization options continue to show (and manage) 'Secondary' instead of 'Tertiary'.  It's rectified by using the FSE GS Panel to 'Reset to Defaults' but I suspect this is pointing to something we're not doing quite right with color customizations so would like to further evaluate that before we bring this in.

pbking:
Ok, did a relatively minor refactoring to color customization handling.
Previously USER provided values were used (if available) as the basis for what colors could be configured.  This change instead makes the colors provided by the THEME the source-of-truth.

The use-case this addresses:
On the /trunk branch change Quadrat colors to something other than default.
Switch to this branch.
Attempt to customize colors.  
Note that `Secondary` is no longer an option to customize and selecting colors will allow the user to correct the customized colors.

I'm good with the rest of the changes so once the customizer changes are 👍-ed I think it's good to go.

MaggieCabrera:
> Ok, did a relatively minor refactoring to color customization handling.
> Previously USER provided values were used (if available) as the basis for what colors could be configured. This change instead makes the colors provided by the THEME the source-of-truth.
> 
> The use-case this addresses:
> On the /trunk branch change Quadrat colors to something other than default.
> Switch to this branch.
> Attempt to customize colors.
> Note that `Secondary` is no longer an option to customize and selecting colors will allow the user to correct the customized colors.
> 
> I'm good with the rest of the changes so once the customizer changes are 👍-ed I think it's good to go.

Thank you so much for fixing this!

seedlet-blocks/child-theme.json
```diff
@@ -75,9 +75,9 @@
 					"name": "Background"
 				},
 				{
-					"slug": "selection",
+					"slug": "tertiary",
 					"color": "#f2f2f2",
```
kjellr: Please double check these colors against the original Seedlet theme. [The value of `tertiary` there is different](https://github.com/Automattic/themes/blob/1aa5058bcff44951877c2db4f461fd4a46433613/seedlet/style.css#L112), and should replace this. I'm not sure where this color value came from.

blockbase/theme.json
```diff
@@ -59,9 +59,9 @@
 					"name": "Background"
 				},
 				{
-					"slug": "selection",
+					"slug": "tertiary",
 					"color": "#c2c2c2",
```
kjellr: Similarly, Blockbase should be using [the tertiary color from Blank Canvas](https://github.com/Automattic/themes/blob/1aa5058bcff44951877c2db4f461fd4a46433613/blank-canvas/variables.css#L12). I'm not sure where this value came from.

pbking:
So it seems that SEEDLET doesn't have a configurable SELECTION color. Instead it's [calculated based on the primary color](https://github.com/Automattic/themes/blob/trunk/seedlet/classes/class-seedlet-custom-colors.php#L259).

Initially Blockbase avoided the inability to do the same by allowing the selection color to be selected by the user (at least potentially) and included in the palette collections.

This change co-ops the "selection" color and converts it to "tertiary".  The intent of the original value wasn't tertiary, it was just text selection.  (I don't believe it was used for anything else.) Kjell notes above that Seedlet-Blocks and Mayland-Blocks have tertiary colors different from the original, likely because those were (manually) selected as good text selection colors, but as good (or historical) "secondary background" colors.

In many cases in Quadrat, because "text selection" and "secondary background (tertiary)" are the same value it's impossible to see what has been selected on an element when that tertiary color is used as a background.

Are we sure we want to lose the ability to separately define the selection color from the tertiary color?  
Should we attempt to re-introduce the capacity to calculate a Selection color based on the Primary color?
Is it ok that the selection of things on a tertiary background aren't visibly selected?

blockbase/theme.json
pbking: Using the tertiary color originally defined in blockbase for text selection makes text selection really hard to see:
image

And when tertiary is the background makes it impossible to see:
image

scruffian:
I think its fine for different themes to implement selection colour in different ways, so I don't think we need a separate setting for it.

pbking:
I'm not opposed to themes implementing selection color different ways, but I don't think that one of those ways should be to use the "Tertiary" color as the background.  Since that's (understood to be) a background color that selection is never going to work for an element that has that as the background.  And for those themes that opt to have "tertiary" mirror "background" it wouldn't work for anything.

Given that, what is the solution for Quadrat (and likely Blockbase and others) that are using it as such?  Something closer to Skatepark's implementation?

pbking:
What do we think about standardizing the selection style to match Skateparks?

Skatepark:
image

Blockbase:
image

Quadrat:
image

pbking:
After some discussion with @kjellr it was decided to instead remove any theme-specific text selection styling with the exception of Skatepark.

Quadrat (customized colors):
image
Quadrat (default colors):
image
Blockbase:
image

seedlet-blocks/child-theme.json
pbking: double-checked and changed

blockbase/theme.json
pbking: changed

PR #4575, opened 2021-09-10
Blockbase: A different approach to navigation

scruffian:
In order to add social links to a navigation block (as suggested in https://github.com/Automattic/themes/pull/4482) using the Customizer, I believe we need a different approach to navigation. One problem with the current approach is that if a user edits their navigation in the Site Editor, edits can no longer be made in Customizer. Another issue is that our approach to the social menu needs significant rework by the user when they switch to the Site Editor.

This PR proposes a different approach:
- We create a new template part called "navigation"
- When a user edits their menus in the Customizer, we recreate the same menu in the template part created above

The major benefit of this approach is that we can build a navigation block for the Site Editor in any way we want. We can use the nav walker class to convert a PHP nav menu to a block nav menu. It's surprisingly simple!

The only downside with this approach is that if I edit the template via the Site Editor, or the Template editor, and edit it again in the customizer, or menus screen, my previous changes would be overwritten.

To test:

- Switch to Blockbase (or any blockbase family theme)
- Create a menu at the "social" location
- Confirm that social links are added to the navigation

Still to do:
- Get the social links to display on the same line as the primary navigation
- Get the preview in customizer to update in real time (this is an issue with the current approach)


mikachan:
I've tested this with Blockbase by adding a menu to the 'social' location. Everything seems to work great - it works at all resolutions and I'm able to edit it via the Menus section and the Customizer. Here's what I'm seeing:

Desktop:
Screenshot 2021-09-15 at 11 46 25

Small resolutions:
Screenshot 2021-09-15 at 11 47 34

Sorry I haven't got any more meaningful feedback... I also didn't see any errors or warnings while testing.

kjellr:
I'm getting some errors when trying to run this PR on a fresh site: 

Screen Shot 2021-09-16 at 7 56 37 AM

scruffian:
@kjellr  I can't replicate that, but I've added a commit which I hope will address it. How did you get to an "fresh site" state?

kjellr:
All sorted out now. This works great! The only small thing I noticed is that if you enter an email address/`mailto` link, it creates just a standard custom link block, not a mail one.

draganescu:
Howdy! This is cool. Nevertheless it feels like a serious workaround, a very specific implementation of this theme's need. I wonder how could the navigation block change to allow one to implement "universal theme blocks-in-menus" easier?

There is ongoing effort to change how the navigation block stores its data. It would be lovely if we could chime in with the theme challenges, described in detail, so that the foundational design of the block (which is still evolving) takes them into account as early as possible.

scruffian:
Closing in favour of https://github.com/Automattic/themes/pull/4482

PR #4581, opened 2021-09-14
Geologist: Update tertiary color

kjellr:
Just a minor update: I'd like to update the tertiary color for Geologist slightly so it has better contrast against the primary and secondary colors. 

Screen Shot 2021-09-14 at 8 33 07 AM


scruffian (approved):
LGTM

PR #4582, opened 2021-09-14
Add batch utility to build a .org compatible .zip of themes

pbking:
Use the command `npm run batch:build:zip:all` to build .org compatible zip files ready for upload for all eligible themes (meaning, themes with `package.json` files).

OR use the command 'npm run batch:build:zip:changed` to build the zip files ONLY for those eligible themes that have had changes since the divergence from trunk.

Zip files follow the pattern `/build/THEMENAME_VERSION/THEMENAME.zip`


mikachan (commented):
I don't have much to add but I think this is really cool... I tested it out (both the all and changed options) and everything worked as you described!

PR #4601, opened 2021-09-14
An attempt to fix the theme check sanitization errors in Blockbase

### LINKED ISSUE #4415: Blockbase: sanitization of fonts and color customization
Upon submitting Blockbase 1.2.0 to the dotorg theme repo, the results from the theme check provided these errors: 

```
REQUIRED: Found a Customizer setting called
'color_palette' in wp-customize-color-palettes.php that did not have a sanitization callback function. Every call to the add_setting() method needs to have a sanitization callback function passed.

REQUIRED: Found a Customizer setting called $global_styles_setting ) in wp-customize-colors.php that did not have a sanitization callback function. Every call to the add_setting() method needs to have a sanitization callback function passed.
Line 113: $wp_customize->add_setting( $global_styles_setting );

REQUIRED: Found a Customizer setting called $global_styles_setting ) in wp-customize-fonts.php that did not have a sanitization callback function. Every call to the add_setting() method needs to have a sanitization callback function passed.
Line 356: $wp_customize->add_setting( $global_styles_setting );
```

Note this does not prevent the update from being pushed live, because the theme is already live.

danieldudzic:
Closes #4415

### DIFF
--- a/blockbase/inc/customizer/wp-customize-color-palettes.php
+++ b/blockbase/inc/customizer/wp-customize-color-palettes.php
@@ -64,9 +64,10 @@ class GlobalStylesColorPalettes {
 		$wp_customize->add_setting(
 			'color_palette',
 			array(
-				'default'    => 'default-palette',
-				'capability' => 'edit_theme_options',
-				'transport'  => 'postMessage', // We need this to stop the page refreshing.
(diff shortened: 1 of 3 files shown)


scruffian (approved):
LGTM

PR #4608, opened 2021-09-15
Add theme-experiments as a submodule

scruffian:
This is just a test to see how it would work to add a submodule to the theme repo. If it works well we could make blockbase a submodule.

To test, try running `npm run sandbox:push` and check that the new themes get pushed up.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "theme-experiments"]
+	path = theme-experiments
+	url = https://github.com/WordPress/theme-experiments/


jeffikus:
@scruffian are sub modules still tied to specific revision numbers or does it pull the latest revision from the target repo trunk?

pbking:
Submodules (optionally) point to any kind of reference.  That which is added here isn't pointing to anything so it defaults to that repositories default branch (and the HEAD of that branch).

I've seen submodule configurations that point to a specific TAG (usually a specific released version) as the mechanism to ensure that the version of submodule that a particular branch is referencing is the one that is expected.  (For instance branch /abc needs version 1.2.3 of submodule XXX while the trunk of the project needs version 1.2.2 of submodule XXX.) But you could configure it to point to any branch or hash or whatever.

pbking:
Note that you need to run this in order to get the submodule checked out:
```
git submodule update --init --recursive
```

and in order to UPDATE the submodule run this:
```
git submodule update --recursive
```

Pulling the update to the "containing" project (themes) doesn't automatically update the "contained" project (theme-experiments).

pbking:
Circling back to this.
The PR shows that a separate repo can be added as a sub-module.
I'm not sure how to test the kind of workflow changes that this would mean for Blockbase though.

scruffian:
Closing for now

PR #4621, opened 2021-09-15
Blockbase: Fix notices related to missing properties in theme.json

scruffian:
If a user specifies a font in theme.json that isn't one of our defined fonts we create several notices. This PR escapes in those cases to stop the notices appearing.

To test, check that you don't see any PHP notices with this PR.

#### Related issue(s):
#4460

### DIFF
--- a/blockbase/inc/customizer/wp-customize-fonts.php
+++ b/blockbase/inc/customizer/wp-customize-fonts.php
@@ -320,20 +320,25 @@ class GlobalStylesFontsCustomizer {
 		$variable = get_settings_array( $array, $configuration );
 		$slug     = preg_replace( '/var\(--wp--preset--font-family--(.*)\)/', '$1', $variable );
 		if ( ! isset( $this->fonts[ $slug ] ) ) {
-			$this->fonts[ $slug ] = $this->build_font_from_theme_data( $slug, $location, $configuration );
+			$this->fonts[ $slug ] = $this->build_font_from_theme_data( $slug, $configuration );
 		}
 		return $this->fonts[ $slug ];
 	}
 
-	function build_font_from_theme_data( $slug, $location, $configuration ) {
+	function build_font_from_theme_data( $slug, $configuration ) {
 		$new_font      = array();
-		$google_fonts  = $configuration['settings']['custom']['fontsToLoadFromGoogle'];
 		$font_families = $configuration['settings']['typography']['fontFamilies']['theme'];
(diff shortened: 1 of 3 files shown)


geologist/child-theme.json
```diff
@@ -418,7 +419,7 @@
 			}
 		},
 		"typography": {
-			"fontFamily": "var(--wp--preset--font-family--dm-sans)",
+			"fontFamily": "var(--wp--preset--font-family--spartan)",
```
scruffian, line 422: This change is just for testing purposes.

geologist/child-theme.json
```diff
@@ -160,6 +160,7 @@
 			},
 			"body": {
 				"typography": {
+					"fontFamily": "var(--wp--preset--font-family--spartan)",
```
scruffian, line 163: This change is just for testing purposes.

pbking:
I tried to reproduce this before applying this change and could not (adding only those lines noted here as "added for testing").  As I was unable to reproduce I'm unable to validate that this change fixes the problem (the behavior is the same before and after for me).

Where should I expect to see the error `Undefined variable: location in /xxx/xxx/wp-content/themes/blockbase/inc/customizer/wp-customize-fonts.php on line 321`?

blockbase/inc/customizer/wp-customize-fonts.php
```diff
@@ -320,21 +320,26 @@ function get_font_family( $array, $configuration ) {
...
 			if ( $font_family['slug'] === $slug ) {
 				$new_font['fontFamily'] = $font_family['fontFamily'];
 				$new_font['name']       = $font_family['name'];
 			}
 		}
+
+		// Get data about the font if we have it.
+		$google_fonts = $configuration['settings']['custom']['fontsToLoadFromGoogle'];
+		if ( empty( $google_fonts ) ) {
+			return false;
```
RavanH, line 341: This would mean the method could return FALSE instead of an array, which in turn would set $this->fonts[ $slug ] to FALSE in get_font_family()... have you made sure that does not cause unexpected results?

scruffian:
> Where should I expect to see the error Undefined variable: location in /xxx/xxx/wp-content/themes/blockbase/inc/customizer/wp-customize-fonts.php on line 321?

You have to set typography fontFamily to var(--wp--preset--font-family--spartan) (a font not present in the default Google font list)

scruffian:
Closing in favour of #4633

Issue #4583, opened 2021-09-14
Geologist: Add Block Patterns


kjellr:
Block patterns have been added to the Geologist figma file. These need to be built into the theme next: 

<img width="543" alt="Screen Shot 2021-09-14 at 1 08 14 PM" src="https://user-images.githubusercontent.com/1202812/133302791-d4ba017b-429b-4b8c-8838-593fd7e06f17.png">

PR #4623, opened 2021-09-16
Geologist: Add block patterns

scruffian:
Adds block patterns ala https://github.com/Automattic/themes/issues/4583

localhost_4759__p=1102

#### Related issue(s):
https://github.com/Automattic/themes/issues/4583


scruffian:
@kjellr please can you take a look and see what tweaks we need to make?

kjellr:
This looks great! Just a few small notes:

- For the authors one, I think we should just use heading/paragraph blocks instead of the Author block. I find it's unusual for folks to actually fill in their author bios, so this seems easier. 
- For the image feature block: Let's make the whole thing wide-width (both the image and the columns block that holds the text. And we can replace the text with the sample text from Figma. 
- Also, when I insert the small email updates pattern, it shows me this classic block notice above it at first: 

Screen Shot 2021-09-17 at 8 51 59 AM

scruffian:
I've made all the changes requested, except that I couldn't replicate the issue with the subscriptions block. How were you testing?

kjellr:
This is looking good: Just a few small comments:

1. When I test this one on dotcom, the left/right padding disappears. This does not happen when I test locally:
 
Screen Shot 2021-09-20 at 11 52 26 AM

2. I'm still seeing that classic block situation for the small newsletter pattern. I think the issue here is that [the `<p>` tags](https://github.com/Automattic/themes/blob/7a7aaaacca6044df9ced89b3bdd117fddc133a1b/geologist/inc/patterns/email-updates-small.php#L12) are missing their Gutenberg comment wrappers (`<!-- wp:paragraph -->`):

Screen Shot 2021-09-20 at 11 50 45 AM

3. It looks like we're pointing to the demo site for the images, but I think those will need to be bundled with the theme, right?

scruffian:
1. Looks OK to me, what site were you testing on?
Screenshot 2021-09-21 at 10 45 41

2. I've added the comments. I was assuming that it was the jetpack block that was breaking! 🤦 

3. I've done this, although I'm not sure why its better to do that :)

kjellr (approved):
> Looks OK to me, what site were you testing on?

Whoops, I forgot to copy over a file. 😄 Looks good now. 

> I've done this, although I'm not sure why its better to do that :)

I think it's more for self-hosted sites, to avoid calling and external file. 

Anyway, this looks good now! Thank you!

scruffian:
Deployed in D67205-code

PR #4624, opened 2021-09-16
Skatepark: Fix title spacing

### LINKED ISSUE #4528: Skatepark: Post/page editor text appears closer to the divider than it does on the front end
Editor|Front End
---|---
Screen Shot 2021-09-03 at 12 37 36 PM|Screen Shot 2021-09-03 at 12 37 18 PM

The title and line area also not wide-width. @melchoyce would you mind mocking up the intended spacing there when there's no featured image? I don't think we have a comp for that.

mikachan:
This change improves the spacing above and below the title divider.

| Before      | After |
| ----------- | ----------- |
| image      | image       |

#### Related issue(s):
Fixes #4528.

### DIFF
--- a/skatepark/assets/theme.css
+++ b/skatepark/assets/theme.css
@@ -182,6 +182,11 @@ h1.wp-block-post-title:not(.has-featured-image .wp-block-post-title) {
 	margin-bottom: 0;
 }
 
+.is-root-container,
+.wp-block-post-content {
+	margin-top: calc(var(--wp--custom--gap--vertical) * 3);
+}
+
 .wp-block-query-pagination .wp-block-query-pagination-numbers .current {
 	text-decoration-thickness: 0.07em;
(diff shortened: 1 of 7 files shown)


pbking:
The spacer isn't used in the editor which made the spacing different between editor and front end.
I pushed a small change that replaces that spacer with a `margin-top` on the post content (different classes for editor vs front-end).

Before:


After:

mikachan:
> I pushed a small change that replaces that spacer with a margin-top on the post content 

Brilliant, thank you! This looks good to me. I didn't know whether to use a spacer or CSS and then I couldn't figure out why the spacer wasn't showing in the editor. This fixes all the issues I kept seeing!

pbking:
Cool, if you like the change then I think it's good to go. 👍

pbking (approved):
🚢

scruffian:
Deployed in D67367-code

PR #4629, opened 2021-09-17
Blockbase: Fix padding for the Group block with a background

### LINKED ISSUE #4617: Blockbase: Group block padding
## Steps to replicate

The Group block seems to be missing padding when it has a background colour:

image

Example markup:

```
<!-- wp:group {"style":{"color":{"background":"#d8f6ec"}}} -->
<div class="wp-block-group has-background" style="background-color:#d8f6ec"><!-- wp:paragraph -->
<p>The Layout Elements category includes the following blocks: <em>Group, Button, Columns, Media &amp; Text, separator, spacer, read more, and page break.</em></p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>This group block has a light green background color.</p>
<!-- /wp:paragraph -->

<!-- wp:button -->
<div class="wp-block-button"><a class="wp-block-button__link">A button</a></div>
<!-- /wp:button -->

<!-- wp:paragraph -->
<p>The read more block should be right below this text, but only on list pages of themes that show the full content. It won't show on the single page or on themes showing excerpts.</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group -->
```

danieldudzic:
Closes #4617

Before:
Group_block_–_themes

After:
after

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -379,6 +379,10 @@ input[type=checkbox] + label {
 	font-size: var(--wp--custom--gallery--caption--font-size);
 }
 
+.wp-block-group.has-background {
+	padding: var(--wp--custom--gap--horizontal);
+}
+
 .block-library-html__edit .block-editor-plain-text {
 	color: var(--wp--custom--form--color--text);
 	border: var(--wp--custom--form--border--width) var(--wp--custom--form--border--style) var(--wp--custom--form--border--color);
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -155,10 +155,6 @@ img {
 	text-decoration: none;
 }
 
-h1.has-background, h2.has-background, h3.has-background, h4.has-background, h5.has-background, h6.has-background {
-	padding: var(--wp--custom--padding--vertical) var(--wp--custom--padding--horizontal);
-}
-
 :root {
 	--wpadmin-bar--height: 46px;
 }
@@ -202,6 +198,10 @@ p {
 
 .has-background-no-padding.wp-block-columns.has-background {
 	padding: 0;
+}
+
+.has-background {
+	padding: var(--wp--custom--gap--vertical) var(--wp--custom--gap--horizontal);
 }
 
 /**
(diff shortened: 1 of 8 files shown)


mikachan (commented):
This looks like it works great!

One thing I noticed while testing this (but not related to these changes) is that the [paragraph](https://github.com/Automattic/themes/blob/f43254ac5b97bc79851c13c5304151d3fc8a549b/blockbase/sass/blocks/_paragraph.scss#L13) and [heading](https://github.com/Automattic/themes/blob/f43254ac5b97bc79851c13c5304151d3fc8a549b/blockbase/sass/base/_header.scss#L11) `.has-background` padding settings no longer apply. It looks like this is because we no longer have a custom padding setting (as we've moved to the gap sizing I think?)

image

So I was thinking maybe we could have a global/non-specific `.has-background` padding rule, something like this in one of the base sass files:

```
.has-background {
	padding: var(--wp--custom--gap--vertical) var(--wp--custom--gap--horizontal);
}
```

Rather than setting this per block/element. Or maybe we just need to update the current paragraph and heading styles to use the gap values instead.

scruffian:
I thought this was something that Gutenberg handled?

scruffian:
I added a commit which standardises the approach for all blocks and fixes the regression on the heading and paragraph blocks. I also added a comment which links to a GB issue which when resolved will mean we can remove this code. Thoughts?

scruffian:
I also added a commit so that the last-child in a block with background color doesn't have a bottom margin, which maintains the vertical rhythm...

mikachan:
I think this works well! Looking good on the same blocks I tested earlier.

scruffian (approved):

scruffian:
Deployed in D67205-code

pbking:
[This](https://github.com/Automattic/themes/pull/4629/files#diff-8d19c6d8b05bdcca09ecf489813da09999af47dd7cfafc71b7e263f900b97410R407-R414) seems to have been rather heavy-handed in applying margins to ALL groups (including those without backgrounds, etc).  This seems to be the cause for some new incorrect layout in the theme's headers.

PR #4630, opened 2021-09-17
Blockbase: Fix Gallery caption spacing

danieldudzic:
#4613

Before & After

BEFORE-4

PR #4631, opened 2021-09-17
Gallery: Add margins to the gallery block using theme.json

scruffian:
This is an alternative approach to https://github.com/Automattic/themes/pull/4630. We can use theme.json to set the gap, so we don't need extra CSS for it.

Before:
Screenshot 2021-09-17 at 14 16 52

After:
Screenshot 2021-09-17 at 14 16 28

#### Related issue(s):
4613

### DIFF
--- a/blockbase/theme.json
+++ b/blockbase/theme.json
@@ -396,6 +396,21 @@
 					"fontFamily": "monospace"
 				}
 			},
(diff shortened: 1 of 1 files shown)


danieldudzic:
LGTM

PR #4633, opened 2021-09-17
Drop "fontsToLoadFromGoogle" for child themes

### LINKED ISSUE #4460: Blockbase: undefined variable location in wp-customize-fonts.php

RavanH:
Drop "fontsToLoadFromGoogle" in favor of  the common "google" parameter. Child themes can simply set a "google" key/value for any Google fonts they need, just as the Blockbase does everywhere else.

I see no need (anymore) to have this separate custom section in the child theme json. Or am I missing something?

It just needs to be communicated and coordinated with child theme devs that still use "fontsToLoadFromGoogle"...

This also fixes #4460 and #4621

### DIFF
--- a/blockbase/inc/customizer/wp-customize-fonts.php
+++ b/blockbase/inc/customizer/wp-customize-fonts.php
@@ -320,25 +320,21 @@ class GlobalStylesFontsCustomizer {
 		$variable = get_settings_array( $array, $configuration );
 		$slug     = preg_replace( '/var\(--wp--preset--font-family--(.*)\)/', '$1', $variable );
 		if ( ! isset( $this->fonts[ $slug ] ) ) {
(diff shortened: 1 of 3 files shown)


scruffian:
Yes! This looks great. I think we can just update the child themes to remove this property and define the theme font as suggested.

MaggieCabrera:
@pbking I would like your eyes on this, I seem to remember we discussed this as a possibility and you had opinions about it :D

scruffian (approved):
LGTM

PR #4662, opened 2021-09-20
Skatepark: Add submenu styling

### LINKED ISSUE #4586: Skatepark: Submenu background color is incorrect
Screen Shot 2021-09-14 at 2 05 15 PM

The background color here should be the same as the page background. Also, all menu items are missing hover states. I believe these should just receive an underline, but @melchoyce can you confirm?

danieldudzic:
I wasn't able to find any theme.json supported settings in the navigation block.

I find it hard to believe that no one has created a Gutenberg issue for this so far, but I wasn't able to find anything.

#4586


scruffian (approved):
LGTM

scruffian:
Deployed in D67205-code

Issue #4606, opened 2021-09-15
Quadrat: Add padding to Post Meta


beafialho:
Currently, Post Meta is too close to the Post Title which makes the layout feel more constrained.

Adding 50px padding to the post meta area feels reasonable and more comfortable.

<img width="1031" alt="Captura de ecrã 2021-09-15, às 11 16 43" src="https://user-images.githubusercontent.com/65220155/133416361-33696edb-7f67-42ce-82ac-b8b825bddd56.png">

| Current      | Intended |
| ----------- | ----------- |
| <img width="893" alt="Captura de ecrã 2021-09-15, às 11 17 15" src="https://user-images.githubusercontent.com/65220155/133416240-56a98162-7ee4-46e4-961e-652945e51e64.png">  |  <img width="1031" alt="Captura de ecrã 2021-09-15, às 11 16 52" src="https://user-images.githubusercontent.com/65220155/133416292-f8db9680-4a23-430d-9a89-1ca1151625b6.png">  |

kjellr:
In addition, the category icon shouldn't be there. Instead, we should have a little dot divider when there are categories. Here's an example from the comps: 

<img width="324" alt="Screen Shot 2021-09-21 at 2 39 55 PM" src="https://user-images.githubusercontent.com/1202812/134228797-bd5bf5d3-eead-4844-a293-8aaa59ae880c.png">

EDIT: Tracked here: https://github.com/Automattic/themes/issues/4605

beafialho:
@kjellr also mentioned that here: #4605

kjellr:
Ah thanks, I missed that one.

danieldudzic:
> In addition, the category icon shouldn't be there. Instead, we should have a little dot divider when there are categories. Here's an example from the comps:
> 
> <img alt="Screen Shot 2021-09-21 at 2 39 55 PM" width="324" src="https://user-images.githubusercontent.com/1202812/134228797-bd5bf5d3-eead-4844-a293-8aaa59ae880c.png">
> 
> EDIT: Tracked here: #4605

@kjellr Does this apply to other meta as well? I'm assuming we want dots to replace all the Blockbase icons?

<img width="1260" alt="tags" src="https://user-images.githubusercontent.com/905781/134392057-153421f5-55c8-4dfb-beae-7a78c90df329.png">

Sidenote: the font-size doesn't look correct :)

danieldudzic:
[...]

PR #4663, opened 2021-09-20
Add post meta to Geologist Theme

kjellr:
This follows up on #4565 and adds the new post meta template part to Geologist (both in the archive and single post views). 

The only bug I'm seeing is that the date icon isn't loading? I'm not sure why. 🤔 

Before|After
---|---
Screen Shot 2021-09-20 at 1 03 46 PM|Screen Shot 2021-09-20 at 12 55 31 PM

### DIFF
--- a/geologist/assets/theme.css
+++ b/geologist/assets/theme.css
@@ -398,18 +398,9 @@ textarea:focus {
 	}
 }
 
-.post-meta .wp-block-post-date::before {
-	display: none;
-}
-
 .post-meta > *,
 .post-meta .wp-block-post-date {
 	margin: 0 8px;
-}
-
-.post-meta > *::before,
-.post-meta .wp-block-post-date::before {
-	content: "";
 }
 
 .post-meta .wp-block-post-terms {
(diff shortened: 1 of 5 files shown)


mikachan:
It looks like the date icon isn't showing because it's hidden in Geologist's [_meta.scss file](https://github.com/Automattic/themes/blob/cb0a6944216dff5b26e1111588a9de30ab17fd68/geologist/sass/templates/_meta.scss#L10). I guess this was because the icon wasn't in the original design perhaps? I've un-hidden for now.

kjellr:
Awesome, thanks @mikachan! This works well for me now. 🙌 Could use one more set of eyes just to be sure, but I think it's good to go.

scruffian:
Should we do the same in Quadrat?

kjellr:
> Should we do the same in Quadrat?

As in, should we add the icon to quadrat? If so, the answer is no — Quadrat was designed to have no icons in its post meta: 

Screen Shot 2021-09-21 at 2 39 55 PM

(We should actually remove the categories icon we have up there right now and replace with this. I've updated https://github.com/Automattic/themes/issues/4606 to mention that).

scruffian:
Deployed in D67302-code

PR #4664, opened 2021-09-20
Added additional margin top to achieve design target

### LINKED ISSUE #4585: Skatepark: Header does not have intended top margin
The margin above the header is incorrect on [the WP.com demo site](https://skateparkdemo.wordpress.com). 

Current|Intended
---|---
Screen Shot 2021-09-14 at 2 00 01 PM|Screen Shot 2021-09-14 at 1 59 42 PM

cc @melchoyce I think the design said this should be [45px margin above there](https://cloudup.com/cWVBBK7xNEj), but I think 30px total seems more comfortable since that matches the left/right margins. What do you think?

pbking:
Added additional margin to header to achieve design target.
HALF of the "vertical gap" was already applied elsewhere in the header.  This change adds another half so the gap is the total of the gap.  Attempting to refactor to use apply the entire gap in one place was proving troublesome.
Note that the "virtical gap" is responsive so it will be the expected `30px` at wide screen but at smaller screens that gap diminishes down to `5vw`

#### Related issue(s):
Fixes #4585

### DIFF
--- a/skatepark/assets/theme.css
+++ b/skatepark/assets/theme.css
@@ -490,6 +490,7 @@ header.wp-block-template-part {
 	max-width: var(--wp--custom--layout--wide-size);
 	margin: 0 auto;
 	width: 100%;
+	margin-top: calc( 0.5 * var(--wp--custom--gap--vertical));
 }
 
 header.wp-block-template-part > .wp-block-group {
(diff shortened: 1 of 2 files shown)


scruffian (approved):
LGTM

scruffian:
I'm not sure we actually need this one anymore. The gap seems to have been added in another PR. @kjellr @melchoyce what do you think?

kjellr:
Yeah, this works for me. It does appear necessary when comparing against what's currently in trunk. 

Before|After
---|---
Screen Shot 2021-09-22 at 8 00 47 AM|Screen Shot 2021-09-22 at 8 00 34 AM

PR #4667, opened 2021-09-20
Added font color and size styles to form input placeholder

### LINKED ISSUE #4596: Quadrat: Placeholder text color/size is incorrect
This is happening on two blocks that I've found: 

---

## Jetpack Subscribe Block

1. Add a Jetpack/WP.com "subscribe" block. 
2. Note that the text color is gray. It should be our standard pink color. 

Screen Shot 2021-09-14 at 2 51 59 PM

---

## Search Block

1. Add a search block.
2. Note that the text color is gray, and too small. The text size should be 18px, and it should be pink.

Screen Shot 2021-09-14 at 2 56 12 PM

pbking:

In order to differentiate between prompt and entered text could we set the opacity?  (.66 pictured below)
Before:
image

After:
image
image

Note: Setting the following does not work as I expected it to:
```
"core/search": {
	"typography": {
		"fontSize": "var(--wp--preset--font-size--normal)"
	}
},
```

#### Related issue(s):

Fixes: #4596

### DIFF
--- a/quadrat/sass/elements/_forms.scss
+++ b/quadrat/sass/elements/_forms.scss
@@ -19,4 +19,9 @@ textarea {
 	&:focus {
 		outline: 1px dotted currentColor;
 	}
+	font-size: var(--wp--preset--font-size--normal);
+	&::placeholder {
+		color: var(--wp--custom--form--color--text);
+		opacity: 0.66;
+	}
 }
--- a/quadrat/sass/elements/_forms.scss
+++ b/quadrat/sass/elements/_forms.scss
@@ -20,9 +20,4 @@ textarea {
 		outline: 1px dotted currentColor;
 	}
 	font-size: var(--wp--preset--font-size--normal);
-
-	&::placeholder {
-		color: var(--wp--custom--form--color--text);
-		opacity: 0.66;
-	}
 }
--- a/quadrat/sass/elements/_forms.scss
+++ b/quadrat/sass/elements/_forms.scss
@@ -20,6 +20,7 @@ textarea {
 		outline: 1px dotted currentColor;
 	}
 	font-size: var(--wp--preset--font-size--normal);
+
 	&::placeholder {
 		color: var(--wp--custom--form--color--text);
 		opacity: 0.66;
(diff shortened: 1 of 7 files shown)


scruffian:
> Note: Setting the following does not work as I expected it to:

Is there a Gutenberg issue for this?

quadrat/sass/elements/_forms.scss
```diff
@@ -18,4 +18,9 @@ textarea {
 	&:focus {
 		outline: 1px dotted currentColor;
 	}
+	font-size: var(--wp--preset--font-size--normal);
+	&::placeholder {
```
scruffian:
```suggestion

	&::placeholder {
```

quadrat/sass/elements/_forms.scss
```diff
@@ -18,4 +18,10 @@ textarea {
 	&:focus {
 		outline: 1px dotted currentColor;
 	}
+	font-size: var(--wp--preset--font-size--normal);
+
+	&::placeholder {
+		color: var(--wp--custom--form--color--text);
+		opacity: 0.66;
```
scruffian: Where does this number come from? Should it be shared across all themes?

scruffian:
This is in the editor:
Screenshot 2021-09-21 at 13 41 37

Should we fix the placeholder there too?

quadrat/sass/elements/_forms.scss
MaggieCabrera: Browsers don't set up transparency for placeholders but set a lighter font color than the input value text has. I think it's fair enough to use opacity here. I'm not sure about adding another variable just for this. I don't have a strong opinion about this case in particular, but I worry that we may be adding a bit too many sometimes.

scruffian:
I moved this code to Blockbase. I think it's ready for a review.

MaggieCabrera:
This is mostly fine except for one little thing:

dotcom comment form :D 
Screenshot 2021-10-05 at 17 39 36

scruffian:
I fixed the dotcom issue here: D67963-code

MaggieCabrera (approved):
I'm fine with this PR, but I wouldn't land it without https://href.li/?D67963-code

Issue #4597, opened 2021-09-14
Quadrat: Site Logo right margin should match spacing between title & tagline


kjellr:
On desktop, the right margin on the site logo should match the spacing between the title and tagline (20px). 

Before|After
---|---
<img width="519" alt="Screen Shot 2021-09-14 at 3 04 37 PM" src="https://user-images.githubusercontent.com/1202812/133318643-a296b76e-d97c-4352-b3bd-e468d2a13558.png">|<img width="496" alt="Screen Shot 2021-09-14 at 3 04 29 PM" src="https://user-images.githubusercontent.com/1202812/133318649-7f059739-e63f-4ed4-b822-f1487e20c91b.png">

pbking:
This was addressed successfully in #4692

PR #4668, opened 2021-09-20
removed non-gap margins from heading

pbking:
Removed all gaps in the header not controlled by the gap (flex) attribute.

Before:
image
image

After:
image
image
image

#### Related issue(s):
#4597

### DIFF
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -501,10 +501,6 @@ textarea:focus {
 	}
 }
 
-.wp-site-blocks .site-header .wp-block-site-logo {
-	margin-right: var(--wp--custom--gap--horizontal);
-}
-
 @media (max-width: 599px) {
 	.wp-site-blocks .site-header .wp-block-site-logo {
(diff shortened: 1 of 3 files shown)


danieldudzic:
I like this approach, but I'm getting some odd behavior for smaller screen sizes:

ouYUIsKfVM

pbking:
Thank you @danieldudzic.  I see that I was overly aggressive removing the auto from the right margin of that element.  Returned it looks correct again:

image

danieldudzic:
@pbking it looks better now, but still a bit odd. Is that correct? The site title being on the right and the subtitle under the logo on the left?
1h557y9O5s

pbking:
I've combined this behavior with #4692.  Can you check there?  The behavior should be different but I've focused my efforts there to make sure it's right:
image
image

danieldudzic:
Yep! Header in Quadrat in #4692 seems to be behaving correctly.

pbking:
Closing this, fixed by #4597 instead

PR #4669, opened 2021-09-20
Add a 'blank' default block template for blockbase themes

### LINKED ISSUE #4260: Blockbase + Co: Default block template could be improved
When we are creating a new block template, the default one has a header that is not consistent with the rest of the site, and the content has no layout. We should define a better default template ourselves like we are doing for our classic themes by just using the single.html template as a base. I wonder if defining this on Blockbase will be enough for the child themes too.

Screenshot 2021-07-21 at 13 37 53

pbking:
This sets the default template content to just a group (inheriting layout) containing the post content.

image

#### Related issue(s):
Fixes: #4260

### DIFF
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -37,6 +37,14 @@ if ( ! function_exists( 'blockbase_support' ) ) :
 			)
 		);
 
+		add_filter(
+			'block_editor_settings_all',
+			function( $settings ) {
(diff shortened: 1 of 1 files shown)


scruffian (approved):
LGTM

scruffian:
Deployed in D67205-code

PR #4673, opened 2021-09-21
Blockbase: Fix alignments in the Query Pagination block

scruffian:
This fixes some alignment issues with the Query Pagination block, when there aren't many posts on a site:

Before:
Screenshot 2021-09-21 at 11 07 19
Screenshot 2021-09-21 at 11 07 12

After:
Screenshot 2021-09-21 at 11 33 41
Screenshot 2021-09-21 at 11 33 36

#### Related issue(s):
Ideally this would be fixed in Gutenberg: https://github.com/WordPress/gutenberg/issues/34997

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -629,7 +629,7 @@ p.has-background {
 	color: var(--wp--custom--color--background);
 }
 
-.wp-block-query-pagination {
+div.wp-block-query-pagination {
 	padding-top: 1.5em;
 	justify-content: space-between;
 	display: grid;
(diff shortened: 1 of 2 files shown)


pbking:
I wasn't able to reproduce the original issue noted in the description screenshots.  The vertical position of the pagination control was a bit different for me (true for both /trunk and the branch) however the numbers were always vertically aligned as expected.

image

image

mikachan (commented):
I think I can reproduce the original issue:

Page 1:
image

Page 2:
image

Page 3:
image

And then with this fix:

Page 1:
image

Page 2 is very similar.

Page 3:
image

So I think this looks good to me 👍.  But I'm not sure why you can't reproduce it @pbking!

pbking:
FWIW I did finally see this reproduce.  I also saw that this change did indeed fix it.  Not sure why I couldn't before!  Though I suspect it was a case of an older version of Gutenberg I was using locally where things worked fine without the fix.

PR #4675, opened 2021-09-21
Blockbase: Add more styles to .aligncenter

### LINKED ISSUE #4619: Blockbase: Missing centered image alignment (classic editor)
## Steps to replicate

This image appears on the left instead of correctly being centered:

image

Example markup (classic editor):

```
<p><img class="size-full wp-image-906 aligncenter" title="Image Alignment 580x300" src="https://sarahnorrissandbox2.files.wordpress.com/2021/09/8a21e-image-alignment-580x300-1.jpg" alt="Image Alignment 580x300" width="580" height="300" /></p>
<p>The image above happens to be <em><strong>centered</strong></em>.</p>
```

mikachan:
This adds a few more styles to the existing `.aligncenter` class to support centering an image from the classic editor.

#### Related issue(s):
Fixes #4619.

### DIFF
--- a/blockbase/sass/base/_alignment.scss
+++ b/blockbase/sass/base/_alignment.scss
@@ -47,8 +47,12 @@
 // This was added for the 'site-logo' block which centers with an 'align:center' attribute
 // instead of 'textAlign' center which sets an .aligncenter class instead of a has-text-align-center
 // class which would do this for us.  I'm not sure why but this centers things appropriately.
+// Display and margin properties added to support image alignment from the classic editor.
 .aligncenter {
 	text-align: center;
+    display: block;
+    margin-right: auto;
+    margin-left: auto;
 }
 
 // Mimick the Gutenberg-generated rules of the layout container on the 404 template
--- a/blockbase/sass/base/_alignment.scss
+++ b/blockbase/sass/base/_alignment.scss
@@ -50,9 +50,9 @@
 // Display and margin properties added to support image alignment from the classic editor.
 .aligncenter {
 	text-align: center;
-    display: block;
-    margin-right: auto;
-    margin-left: auto;
+	display: block;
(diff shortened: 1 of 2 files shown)


blockbase/sass/base/_alignment.scss
```diff
@@ -47,8 +47,12 @@
 // This was added for the 'site-logo' block which centers with an 'align:center' attribute
 // instead of 'textAlign' center which sets an .aligncenter class instead of a has-text-align-center
 // class which would do this for us.  I'm not sure why but this centers things appropriately.
+// Display and margin properties added to support image alignment from the classic editor.
 .aligncenter {
 	text-align: center;
+    display: block;
```
scruffian: Nitpick: This looks like spaces not tabs

mikachan: Ha yes it definitely does! I'll change it 😁

jffng: Are there any cases where adding `display: block` to an `aligncenter` block would result in unwanted behavior? 

It's pretty low specificity so I think it's okay, but it gave me pause.

jffng (commented):
This fixes the issue as described, if we're okay with the display block property being added then 🚢

blockbase/sass/base/_alignment.scss
mikachan: > Are there any cases where adding display: block to an aligncenter block would result in unwanted behavior?

Yeah this is definitely valid. I guess the possible issue would be that content would be stacked that you may have intended to be inline, from applying this class? But I can't think of an example where this would happen when centering (i.e. you'd probably expect things to stack and center, and only expect left/right alignment to be inline, I think..)

scruffian:
Deployed in D67367-code

Issue #4656, opened 2021-09-20
videomaker: Navigation (if needed)


scruffian:
The Videomaker navigation looks like this:
<img width="280" alt="Screenshot 2021-09-22 at 14 29 04" src="https://user-images.githubusercontent.com/275961/134352737-0866d059-fb99-445e-948d-da873f1a3177.png">

<img width="441" alt="Screenshot 2021-09-22 at 14 29 18" src="https://user-images.githubusercontent.com/275961/134352733-a1d5f2ab-81c1-408e-a672-854ea9812f11.png">

We need to wait for https://github.com/WordPress/gutenberg/issues/34783 to merge before we can achieve all of this.

scruffian:
This will be fixed by https://github.com/Automattic/themes/pull/4832

kjellr:
I don't think the mockups above are relevant any longer, but I do think we need to make some adjustments to the way the navigation appears on desktop. Currently, the menu items are small and way over on the left edge of the page. This is easy to miss on medium-to-large screens. 

![menu](https://user-images.githubusercontent.com/1202812/141474872-97ab8696-c4c9-4c1c-8700-175eabb0e1e2.gif)

The theme mockups have these aligned more to the center column, so I think we should give that a try:

![Menu Expanded-1](https://user-images.githubusercontent.com/1202812/141475054-680048a8-8d9e-4692-8207-ee89c363956e.jpg)
![Menu Expanded](https://user-images.githubusercontent.com/1202812/141475057-d091e16e-1fc6-4c04-9363-4f180233be92.jpg)

cc @beafialho

beafialho:
> The theme mockups have these aligned more to the center column, so I think we should give that a try:

Definitely, this looks more obvious 👍

PR #4680, opened 2021-09-22
Videomaker: Add navigation styles

scruffian:
Adds navigation styles for Videomaker:
Screenshot 2021-09-23 at 10 30 57
Screenshot 2021-09-23 at 10 30 45

#### Related issue(s):
https://github.com/Automattic/themes/issues/4656


kjellr:
Sorry for not having caught this sooner — let's leave this vertical menu out entirely for the initial launch. It's not viable for a bunch of users, and the existing horizontal menu is entirely fine. We should turn on the hamburger by default at all times (once that's possible), but that's all we should do with the menu. 

If we'd like, we can decide to bundle a vertical option like this as an optional header block style. But I don't think it's a priority at all, and I think we should launch without it to keep things streamlined.

beafialho:
This could work if menu links didn't have sublinks, but I agree we can launch without it.

scruffian:
I've updated the code and the screenshot. Ready for another review.

beafialho:
Thanks, this looks good!

kjellr:
The only other feedback I have is that this suffers from the same issue noted in https://github.com/Automattic/themes/issues/4687. The spacing on the sides + top of the header should be identical. Not sure if you want to fix that here or in a separate PR though — it's also a blockbase issue.

scruffian:
Deployed in D67302-code

PR #4685, opened 2021-09-22
Videomaker: add footer

### LINKED ISSUE #4660: videomaker: Templates: Footer template part (including mobile)

jffng:

This PR adds a footer template part. It also registers a new nav area `footer` so that users can place links there. I thought this was simpler than adding an opt-in for social menu icons vs text. 

Screen Shot 2021-09-23 at 2 30 37 PM

I'm not sure what to do about the credit.

To test this:
- Add a menu, assign it to the "Footer" display location
- Verify it appears in the footer

#### Related issue(s):
#4660

### DIFF
--- a/videomaker/sass/templates/_footer.scss
+++ b/videomaker/sass/templates/_footer.scss
@@ -1,0 +1,3 @@
+footer > .wp-block-group {
+	justify-content: space-between; // Needed until we have a way to set this via block attributes
+}
--- a/videomaker/sass/templates/_footer.scss
+++ b/videomaker/sass/templates/_footer.scss
@@ -1,3 +1,8 @@
 footer > .wp-block-group {
+	align-items: center;
 	justify-content: space-between; // Needed until we have a way to set this via block attributes
+	padding: var(--wp--custom--gap--horizontal);
+	p {
+		margin-bottom: 0;
+	}
 }
--- a/videomaker/sass/templates/_footer.scss
+++ b/videomaker/sass/templates/_footer.scss
@@ -1,7 +1,6 @@
 footer > .wp-block-group {
 	align-items: center;
 	justify-content: space-between; // Needed until we have a way to set this via block attributes
-	padding: var(--wp--custom--gap--horizontal);
 	p {
 		margin-bottom: 0;
 	}
--- a/videomaker/assets/theme.css
+++ b/videomaker/assets/theme.css
@@ -59,7 +59,6 @@ a:active {
 footer > .wp-block-group {
 	align-items: center;
 	justify-content: space-between;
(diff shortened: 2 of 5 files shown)


scruffian:
I think for the social menu, the thing we need to do is make the icons an opt in setting..

videomaker/sass/templates/_footer.scss
```diff
@@ -0,0 +1,8 @@
+footer > .wp-block-group {
+	align-items: center;
+	justify-content: space-between; // Needed until we have a way to set this via block attributes
+	padding: var(--wp--custom--gap--horizontal);
```
scruffian: Can we set this in the block attributes?

jffng: Yes good call, also I realized I probably just need left and right, not top and bottom because the vertical spacing is already provided by the group block.

jffng:
I believe this is ready for a design review. A few things to note:

- ~The font size for the links will be updated / correct when #4699 lands.~
- The vertical spacing between the links is slightly bigger than the comps, but I wasn't sure if it was worth targeting in CSS, let me know. 
- I wasn't sure what to do about the credit text, since it won't be editable for non-FSE users. So I left the Proudly Powered by WordPress in there for now.

videomaker/sass/templates/_footer.scss
jffng: Actually based on https://github.com/Automattic/themes/pull/4698 I think we will need the top and bottom padding, I added it back in the block attributes.

scruffian: I think that's going to be removed in https://github.com/Automattic/themes/pull/4698

scruffian:
> The font size for the links will be updated / correct when Updates font sizes. #4699 lands.

Isn't that for Geologist?

scruffian:
> I wasn't sure what to do about the credit text, since it won't be editable for non-FSE users. So I left the Proudly Powered by WordPress in there for now.

👍

jffng:
> Isn't that for Geologist?

Oops, you're right. It doesn't look this a 20px font size is in the type styles — @beafialho should I set a custom font size for the footer links, or would you want them to be the default font size 18px?

kjellr:
> Oops, you're right. It doesn't look this a 20px font size is in the type styles — @beafialho should I set a custom font size for the footer links, or would you want them to be the default font size 18px?

Those can just be the default size. 

> I think for the social menu, the thing we need to do is make the icons an opt in setting..

I think just using a nav block here is fine for now.

scruffian (approved):
LGTM.

The navigation should be IN UPPERCASE but that will happen in https://github.com/Automattic/themes/pull/4701

Issue #4686, opened 2021-09-22
Geologist: Site Logo option is missing from the customizer. 


kjellr:
Geologist is missing the ability to customize the site logo. This should be there, identical to how it works for Quadrat.

<img width="926" alt="Screen Shot 2021-09-22 at 2 19 15 PM" src="https://user-images.githubusercontent.com/1202812/134399798-c940c271-5837-4b9a-8f42-7405fc89863b.png">

Quadrat, for comparison: 

<img width="683" alt="Screen Shot 2021-09-22 at 2 21 21 PM" src="https://user-images.githubusercontent.com/1202812/134399968-788d2070-b093-4bb3-81f5-fa4fc35ebe3f.png">

PR #4692, opened 2021-09-23
Blockbase: Add site logo

### LINKED ISSUE #4597: Quadrat: Site Logo right margin should match spacing between title & tagline

scruffian:
This adds support for a site logo to Blockbase and all children

@kjellr we'll probably want to tweak some of the spacing here:

Blockbase:
Screenshot 2021-09-23 at 09 55 27

Quadrat:
Screenshot 2021-09-23 at 09 55 48

Geologist:
Screenshot 2021-09-23 at 09 56 11

Mayland Blocks:
Screenshot 2021-09-23 at 10 03 31

Seedlet Blocks:
Screenshot 2021-09-23 at 10 03 52

Skatepark:
Screenshot 2021-09-23 at 09 56 35

Videomaker:
Screenshot 2021-09-23 at 09 54 50

#### Related issue(s):
https://github.com/Automattic/themes/issues/4686


kjellr:
> @kjellr we'll probably want to tweak some of the spacing here:

Blockbase, Mayland Blocks, and Videomaker can use the spacing used in Geologist.

pbking:
Pushed a change so that the navigation flexes to fill the space and the rest of the elements flow from the start.

Videomaker:
image

Blockbase:
image

Mayland (seems unchanged):
image

kjellr:
Can we set the default size to 64px to match Quadrat? It's currently 120px wide. I also think it's better to set the height and let the width be auto, so it better supports wide text-based logos. 

Currently: 

Screen Shot 2021-09-23 at 15 30 41

Recommended: 

Screen Shot 2021-09-23 at 15 34 37

kjellr:
On small screens — if there's a logo _and_ a title — the logo should sit above like it does for Quadrat.

Currently, it just stays to the left: 

Screen Shot 2021-09-24 at 10 58 58 AM

Here's quadrat for reference: 

Screen Shot 2021-09-24 at 11 00 16 AM


---

Also, does Blockbase not support display of the site tagline via the customizer?

pbking:
> Also, does Blockbase not support display of the site tagline via the customizer?
The tagline isn't in the header, but the field is available to edit in the customizer:


I don't believe it was ever in the header.

It sounds like your suggestion is that Blockbase and all children (with the exception of Seedlet-blocks and Skatepark) should have the exact same header design (the design of Quadrat).  Is that correct?

pbking:
Considering the standard we seem to be achieving of "64px" for the site logo in the header are these values appropriate?

image

pbking:
I pushed some changes that bring all of the non-triangle-specific styling from Quadrat to Blockbase.  I am in the process of evaluating all of the themes with the following scenarios: "with/without tagline", "with/without logo", "with/without navigation", "mobile/desktop"
Have also brought in the change from https://github.com/Automattic/themes/pull/4668 for a more complete PR.

kjellr:
> It sounds like your suggestion is that Blockbase and all children (with the exception of Seedlet-blocks and Skatepark) should have the exact same header design (the design of Quadrat). Is that correct?

That's right, thanks. 

> Considering the standard we seem to be achieving of "64px" for the site logo in the header are these values appropriate?

Height should be at least 2x the expected height, so `192px` (so the logo appears sharp on high-res screens). Not sure about width, since we aren't strict about that. From the docs it looks like you can [specify that a flexible width is ok](https://developer.wordpress.org/themes/functionality/custom-logo/#what-is-a-custom-logo), so maybe we do that?

pbking:
Adjusted to `192`.  A height value is needed (otherwise it defaults to 150px) so I set that the same as width.  It's just a suggestion though (as far as I can tell) and the behavior when selecting and cropping the image is the only place that effects things.
If we want to suggest that the image be dimensions other than a square LMK what that size should be.

scruffian:
This is looking good. One concern I have, which I also have about the post meta, is that these rules are starting to be quite opinionated, and other themes might want to opt out of them (Mayland Block and Seedlet Blocks for example). I am wondering if we need to split Blockbase's CSS into two files - a ponyfill, and then some extra opinions, which a child theme can either conditionally add, or dequeue.

pbking:
Skatepark and Seedlet Block both have an opinionated header that's different than Blockbase.  All of the "opinion" is in the '.site-header' class (and the gap variables it leverages) and the conditional a theme can act on is to use the class or not.

I can see how that opinion isn't NEEDED in something like Skatepark.  The overhead (IMO) is negligible and my 'druthers (for this PR in this moment in time) is not to work out a more complicated solution.  

I see potential value in the idea though, especially since that "opinion" deviates from the idea of Blockbase ponyfilling in the things that Gutenberg doesn't yet offer and instead "frameworks" an opinionated thing for child themes.

scruffian:
Yeah it might not be needed yet, but it's on my mind :)

kjellr:
> Skatepark and Seedlet Block both have an opinionated header that's different than Blockbase. All of the "opinion" is in the '.site-header' class (and the gap variables it leverages) and the conditional a theme can act on is to use the class or not.

At the very least, would it make sense to assign the more opinionated rules to a second class (`.site-header-opinionated`  or something), to divide them out from the more important structural ones? That way a child theme could remove them by removing the class in the header instead of overriding them if they still want to use a more generic `.site-header` classname.

pbking:
Another thing to consider...
As I was building this out in some cases I was moving the configuration of gap from `template-part` to `css`.  It was easy to make it consistent in that way, but not exactly re-useable. 

At the same time... CSS does look and feel a whole lot cleaner for that kind of opinion.  Might this just be most appropriate for a block style?  We're targeting a group block after all.  Perhaps the relevant opinion could be extracted into a Group Block Style and applied that way?

Regardless, I'm in favor of extracting opinion in some way.  I prefer to do that in a [separate](https://github.com/Automattic/themes/issues/4704) change.

jffng (approved):
> Blockbase and all children (with the exception of Seedlet-blocks and Skatepark) should have the exact same header design (the design of Quadrat).

This is exactly what I get in my testing, so I think this is good to go. 🚢 

> Regardless, I'm in favor of extracting opinion in some way. I prefer to do that in a separate change.

👍

PR #4693, opened 2021-09-23
Quadrat+Skatepark: Remove code that removes support for block templates

scruffian:
I think these themes should support block templates, I'm not sure why we removed the support.


jffng (approved):
Confirmed this works as expected, enabling template editing mode in the post editor.

scruffian:
Deployed in D67367-code

Issue #4641, opened 2021-09-20
videomaker: theme.json: Margin/spacing settings + content layout


scruffian:

scruffian:
I think we can close this with #4697 merged.

PR #4697, opened 2021-09-23
Videomaker: Add content width settings

scruffian:
This updates the layout settings for Videomaker:

Content width: 886px
Wide size: 1000px

This also uses the same gap size as Blockbase:
```
			"gap": {
				"baseline": "10px",
				"horizontal": "min(30px, 5vw)",
				"vertical": "min(30px, 5vw)"
			},
```

#### Related issue(s):
https://github.com/Automattic/themes/issues/4641

### DIFF
--- a/videomaker/child-theme.json
+++ b/videomaker/child-theme.json
@@ -98,6 +98,10 @@
 				}
 			]
 		},
+		"layout": {
+			"contentSize": "886px",
+			"wideSize": "1342px"
+		},
 		"typography": {
 			"fontFamilies": [
 				{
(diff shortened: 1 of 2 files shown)


pbking:
Looks like the wide width should be "1342px" per figma and discussion.  Also the values should be in "settings" rather than "style".

kjellr:
> Looks like the wide width should be "1342px" per figma and discussion. Also the values should be in "settings" rather than "style".

It should actually be 800px and 1300px so the numbers are less random. (@beafialho and I had a followup chat). I pushed a quick commit, so this should be ready to go!

pbking (approved):
🚢

scruffian:
Deployed in D67367-code

Issue #4687, opened 2021-09-22
Geologist: Use identical spacing above and to the left/right of the header. 


kjellr:
**The padding on the top, left, and right of the header should always be identical.** 
Right now there's more space on the top than there is on the sides. This has come up for multiple themes — I think this might be an issue with Blockbase. 

On desktop, please use the larger measurement here (`min(30px, 5vw)`). On small screens you can use the smaller measurement (`min(20px, 5vw)`). The left/right padding here would also apply to the entire page. 

Current:
<img width="1136" alt="Screen Shot 2021-09-22 at 2 31 16 PM" src="https://user-images.githubusercontent.com/1202812/134401399-458a49d3-931e-4b55-bf96-7a348b27f3f3.png">

Preferred:
<img width="1134" alt="Screen Shot 2021-09-22 at 2 31 06 PM" src="https://user-images.githubusercontent.com/1202812/134401406-d1e2bb90-b6bb-4e02-81fe-ad474bf06b1f.png">

danieldudzic:
The current setting for the `padding-top` is `min(30px, 5vw)`. Making spacing identical above and to the left/right of the header would require changing the left and right padding. Is this the correct way of approaching this?

kjellr:
> The current setting for the padding-top is min(30px, 5vw). Making spacing identical above and to the left/right of the header would require changing the left and right padding. Is this the correct way of approaching this?

Yes! On small breakpoints though, we should continue to use `min(20px, 5vw)` like we do now, but also apply that to the top.

PR #4698, opened 2021-09-23
Removed block-level margin styles from all themes

pbking:
[This change](https://github.com/Automattic/themes/pull/4629) introduced top and bottom margins for all [group blocks](https://github.com/Automattic/themes/pull/4629/files#diff-8d19c6d8b05bdcca09ecf489813da09999af47dd7cfafc71b7e263f900b97410R407-R414).  That introduced undesirable margins in other places (such as the header).  I don't believe that change was necessary to achieve the goal; colored background groups seem to continue to look correct with that setting absent.

This change removes that block-level configuration for all Blockbase themes.

Before:
image

After:
image

### DIFF
--- a/blockbase/theme.json
+++ b/blockbase/theme.json
@@ -404,14 +404,6 @@
 					}
 				}
 			},
-			"core/group": {
-				"spacing": {
-					"margin": {
-						"top": "var(--wp--custom--gap--vertical)",
-						"bottom": "var(--wp--custom--gap--vertical)"
-					}
-				}
-			},
 			"core/navigation": {
 				"typography": {
 					"fontSize": "var(--wp--preset--font-size--normal)"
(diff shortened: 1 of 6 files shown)


scruffian (approved):
LGTM

kjellr:
This definitely removes a little space, but the space on top still doesn't match the sides as detailed in https://github.com/Automattic/themes/issues/4687. Are we going to handle that part separately? 

Blockbase: 

Before|After
---|---
Screen Shot 2021-09-23 at 15 25 32|Screen Shot 2021-09-23 at 15 24 24


Quadrat: 

Before|After
---|---
Screen Shot 2021-09-23 at 15 25 58|Screen Shot 2021-09-23 at 15 25 02

Geologist (No change): 

Before|After
---|---
Screen Shot 2021-09-23 at 15 25 48|Screen Shot 2021-09-23 at 15 24 05

scruffian:
Deployed in D67367-code

PR #4709, opened 2021-09-26
Deploy Themes 1.0.2 to wpcom

pbking:
version bump
Added additional margin top to achieve design target (#4664)
Quadrat: Add pagination arrows through block settings, not in CSS (#4536)
Videomaker: add footer (#4685)

 Phabricator: D67421-code

(update)
Due to some errors in my workflow that diff wasn't working for me like I wanted to.  Might not be anything wrong with the diff but this is the one that wound up landing: D67473-code

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "themes",
-	"version": "1.0.1",
+	"version": "1.0.2",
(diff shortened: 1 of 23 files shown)


pbking:
The diff has landed and has been deployed.

PR #4711, opened 2021-09-27
Blockbase: Update font size names

scruffian:
Standardises our font names to match those used in core.

The font sizes names in core are: `small`, `normal`, `medium`, `large` and `huge`. Blockbase uses 5 font sizes, 14px, 16px, 18px, 24px and 28px. 

We _could_ map them like this:

14px - small
16px - normal
18px - medium
24px - large
28px - huge

However this would mean that "normal" mapped to 16px, whereas the body font size is 18px.

We _could_ map them like this:

14px - small
16px - medium
18px - normal
24px - large
28px - huge

However this would mean that "medium" is smaller than "normal" which is different to core.

Instead I've gone for this mapping:

14px - custom
16px - small
18px - normal
24px - large
28px - huge

This isn't ideal as it means that any theme that doesn't want the custom sizes will have to override them manually (see the additions to the child-theme.json files in Geologist and Quadrat).

Another option would be to remove the 14px custom size from Blockbase and remap them to small, which plays better with the child themes. This is my preferred approach but it will mean a change to the design of Blockbase, so we'll need to run it past some designers.

If we did that we'd see the following visual changes:

Footer credit before:
Screenshot 2021-09-27 at 11 51 50

Footer credit after:
Screenshot 2021-09-27 at 11 51 45

Post meta before:
Screenshot 2021-09-27 at 11 59 16

Post meta after:
Screenshot 2021-09-27 at 11 59 28

Form labels before:
Screenshot 2021-09-27 at 12 01 01

Form labels after:
Screenshot 2021-09-27 at 12 01 41

Pullquote, quote and table captions before:
Screenshot 2021-09-27 at 12 04 43

Pullquote, quote and table captions after:
Screenshot 2021-09-27 at 12 05 49

To be clear I haven't yet made the above changes, but I think we should :)

#### Related:
pNEWy-en2-p2


MaggieCabrera:
I'm thinking the post meta, and footer can be solved using the custom value on the block markups. The form labels should use css (the block doesn't allow for any customization at all atm, so that can be moved to the block in the future when it allows it). For the pullquote, quote and table I'd say the same as the table, in the future they most probably will provide a way to customize the font size of those elements. Are medium and normal both 18px for blockbase then?

scruffian:
> Are medium and normal both 18px for blockbase then?

Blockbase doesn't use medium at all

MaggieCabrera:
> > Are medium and normal both 18px for blockbase then?
> 
> Blockbase doesn't use medium at all

No, but core does, I just checked and it's not there so it makes sense for Blockbase not to have it. we could have a custom theme.json variable for the 14px value, what do you think?

scruffian:
Yeah that's a good idea :)

kjellr:
> Another option would be to remove the 14px custom size from Blockbase and remap them to small, which plays better with the child themes. This is my preferred approach but it will mean a change to the design of Blockbase, so we'll need to run it past some designers.

I'd prefer not to change the post meta (etc.) to 16px, since it doesn't have enough of a scale difference compared to the normal body text.

scruffian:
> I'd prefer not to change the post meta (etc.) to 16px, since it doesn't have enough of a scale difference compared to the normal body text.

We could change all the 16px to 14px...

blockbase/theme.json
```diff
@@ -465,6 +463,12 @@
 					"style": "solid",
 					"width": "0 0 0 1px"
 				},
+				"citation": {
```
scruffian, line 472: I took this from Quadrat as it seemed like a good rule to apply to all themes...

scruffian:
I updated the approach here to use a custom variable in place of "tiny".

kjellr:
> I updated the approach here to use a custom variable in place of "tiny".

That seems like a reasonable approach to me. 👍

jeffikus:
@scruffian i think that commit of mine to the PR happened by accident when I was pulling the latest changes to review. Feel free to revert it.

pbking (approved):
w00t!
That was a hefty one but things look pretty good I think.
There may be some things to shake out from this but I don't think they would be critical if there is anything amiss.

Checked out all of the BB themes and they all seem good.  Design targets seem to remain correct with a cleaner implementation.

pbking:
I think the commit is fine @jeffikus.  This is gonna get squash-merged anyway and I don't see anything troubling coming in from it.

PR #4712, opened 2021-09-27
Blockbase: Remove padding from the separator

### LINKED ISSUE #4710: Blockbase: Separator block height issue
I noticed that when I use a custom color on the separator block its height increases significantly.

## Steps to replicate
1. On a site using Blockbase theme (on WordPress.com) add the separator block to a post or a page content.
2. Note the separator height - it's 1px tall (although in the style picker preview it is much thicker).
3. Go to the settings panel and change the color to any custom color (from the color palette or random).
4. Note how the height of the block changes.

## Expected

I expected that the height of the separator block would not change when I picked a custom color.

Here's a quick video: 
https://cloudup.com/cxGWiqhSJ8m

danieldudzic:
Before:

before-5

After:

HR_–_themes

Closes #4710

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -207,7 +207,7 @@ p {
 	padding: 0;
 }
 
-.has-background {
+.has-background:not(.wp-block-separator) {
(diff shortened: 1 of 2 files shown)


mikachan (commented):
This LGTM 👍

PR #4714, opened 2021-09-27
Blockbase: add json schema to theme.json

MaggieCabrera:

Props to @ajlende for adding theme.json to https://json-schema.org/ This PR adds it to Blockbase and children. 

To test it, open theme.json and check and add an error to it, vscode should warn you about it:

Screenshot 2021-09-27 at 16 17 31

### DIFF
--- a/blockbase/theme.json
+++ b/blockbase/theme.json
@@ -1,4 +1,5 @@
 {
+	"$schema": "https://json.schemastore.org/theme-v1.json",
 	"version": 1,
 	"templateParts": [
 		{
--- a/geologist/theme.json
+++ b/geologist/theme.json
@@ -1,4 +1,5 @@
 {
+	"$schema": "https://json.schemastore.org/theme-v1.json",
 	"version": 1,
 	"templateParts": [
 		{
--- a/mayland-blocks/theme.json
+++ b/mayland-blocks/theme.json
@@ -1,4 +1,5 @@
 {
+	"$schema": "https://json.schemastore.org/theme-v1.json",
 	"version": 1,
 	"templateParts": [
 		{
--- a/quadrat/theme.json
+++ b/quadrat/theme.json
@@ -1,4 +1,5 @@
 {
+	"$schema": "https://json.schemastore.org/theme-v1.json",
 	"version": 1,
 	"templateParts": [
 		{
--- a/seedlet-blocks/theme.json
+++ b/seedlet-blocks/theme.json
@@ -1,4 +1,5 @@
 {
+	"$schema": "https://json.schemastore.org/theme-v1.json",
 	"version": 1,
 	"templateParts": [
 		{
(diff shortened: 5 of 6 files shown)


mikachan:
Looks awesome! Thanks for sorting this.

ajlende:
I should note that there are a couple bugs that are being fixed in https://github.com/SchemaStore/schemastore/pull/1846. Let me know if you find any more.

One thing that I wasn't sure if I should include in the official schema are arrays in `settings.custom`.

They are used in Blockbase and children.

```jsonc
{
	"settings": {
		"custom": {
			"colorPalettes": [
				{
					"label": "Featured",
					"slug": "palette-1",
					"colors": {
						"primary": "#C8133E",
						"secondary": "#4E2F4B",
						"foreground": "#1D1E1E",
						"background": "#FFFFFF",
						"tertiary": "#F9F9F9"
					}  // ... 
				}
			]
		}
	}
}
```

These generate the following CSS custom properties.

```css
body {
--wp--custom--color-palettes--0--label: Featured;
--wp--custom--color-palettes--0--slug: palette-1;
--wp--custom--color-palettes--0--colors--primary: #C8133E;
--wp--custom--color-palettes--0--colors--secondary: #4E2F4B;
--wp--custom--color-palettes--0--colors--foreground: #1D1E1E;
--wp--custom--color-palettes--0--colors--background: #FFFFFF;
--wp--custom--color-palettes--0--colors--tertiary: #F9F9F9;
/* ... */
}
```

I didn't add arrays to the PR above because it seemed like something that could be a mistake: someone accidentally putting `settings.color.palette` under `settings.custom.color.palette` and not realizing that its generating these properties instead of the `--wp--preset--color--{slug}` ones. I didn't see these properties used anywhere, so I'm curious of the intentions behind these settings and if I should allow arrays in the PR above.

MaggieCabrera:
> One thing that I wasn't sure if I should include in the official schema are arrays in `settings.custom`.

I don't think we should. `settings.custom` is reserved for the theme authors to add extra CSS variables when needed and these could be whatever the theme developer decides them to be. The way we work on Blockbase is that, whenever we find something we can't do using theme.json as it stands now, we create a new variable in custom that will serve its purpose, and in the future, we will absorb these custom variables into preset ones when we can.

In the case of the color palettes, this is a feature that Gutenberg has not built-in. We can define colors for a theme, but we can't define alternatives to said colors and map them to things (like in Skatepark, a Background/foreground color combination) to easily swap in and out. We do build this feature using the customizer and intend to adapt this in the future to work with Global Styles once this is supported. The customizer changes the theme.json variables depending on what the user selects (the code lives [here](https://github.com/Automattic/themes/blob/2ea629428164d95843b80bfb8144fa840cbc499d/blockbase/inc/customizer/wp-customize-color-palettes.php#L35)).

pbking (approved):
Tickled to have this in.  Thanks for the PR and the Schema!

PR #4720, opened 2021-09-28
Quadrat: Fix post title spacing

### LINKED ISSUE #4592: Quadrat: If a post has no title, we should collapse the space that the title would usually take up
Before|After
---|---
Screen Shot 2021-09-14 at 2 40 03 PM|Screen Shot 2021-09-14 at 2 41 11 PM

mikachan:
This PR should remove the spacing if the post title is empty.

Before:
image

After:
image

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/4592


quadrat/sass/blocks/_post-title.scss
```diff
@@ -0,0 +1,7 @@
+.wp-block-post-title {
+	margin-bottom: var(--wp--custom--gap--vertical);
+
+	&:empty {
```
jffng: I understand why you did this but IMO this shouldn't be needed — if the element is empty, why render it? That would need to be addressed in GB though.

quadrat/sass/blocks/_post-title.scss
```diff
@@ -0,0 +1,7 @@
+.wp-block-post-title {
+	margin-bottom: var(--wp--custom--gap--vertical);
```
scruffian: This looks like it belongs in Gutenberg. What do you think about opening an issue for it there?

jffng (commented):
This looks better in the case where there is no title. However when there is a title, the spacing has changed quite a bit: 

Trunk | This PR
-------- | ---------
Screen Shot 2021-09-28 at 1 56 15 PM | Screen Shot 2021-09-28 at 1 55 42 PM
 
I think we want to maintain the spacing when there is a post title / featured image, is that right @kjellr ?

kjellr:
> I think we want to maintain the spacing when there is a post title / featured image, is that right @kjellr ?

That's right, we don't want to change the spacing in any other scenario. We just want to hide the post title space if there's no post title.

mikachan:
Thanks all! I think it makes sense to open an issue in Gutenberg for this, or specifically for not rendering an empty title element (or any element actually, I think this would help with empty featured images too). Have I understood that correctly? I'm happy to open the issue too 👍

mikachan:
I've created an issue in GB to address not rendering an empty element when there is no post title: https://github.com/WordPress/gutenberg/issues/35224

I've also just pushed a commit to increase the spacing below the title on this PR, just in case we need to continue with this as a fix for now.

scruffian:
Thanks. It would be good to link to that issue in the SASS so we know when we can remove the code...

mikachan:
Yep I added that! https://github.com/Automattic/themes/pull/4720/commits/bfaca58981070fcbddde5f3979b3d076d7bc393e

quadrat/sass/blocks/_post-title.scss
```diff
@@ -0,0 +1,9 @@
+.wp-block-post-title {
+	margin-bottom: calc( var(--wp--custom--gap--vertical) * 3.8 );
```
jffng: When there is a featured image, this causes more space between the title and featured image:

Trunk | This PR
------ | -----
Screen Shot 2021-09-30 at 2 22 13 PM | Screen Shot 2021-09-30 at 2 22 22 PM

I don't know if that's desirable cc @kjellr @beafialho

kjellr: That's not desirable. 😄 We should have equal spacing above and below the title. (Ideally the spacing would below the title would match the spacing that's currently above the title).

mikachan:
Thanks @jffng and @kjellr. I've pushed a fix for the spacing when there's a featured image. I've used a negative top margin on the image container to reduce the spacing between the title and the image. And then I've added a bottom margin to the image as well, as I'd missed this from the original spacing too 🙈

image

kjellr:
This is looking pretty off to me at the moment:

Post Archive pages: [Before](https://cloudup.com/i8k7TlE8dQ7) / [After](https://cloudup.com/ieHQmYUZAoD)
Single Posts: [Before](https://cloudup.com/iExF3RwNEyx) / [After](https://cloudup.com/i9p8uSHjwpc)

(Sorry for the links. For some reason, I'm having trouble with embedding images on GitHub at the moment. 🤷)

I prefer the before states in both cases. I did some digging myself, and I don't think there's a non-hacky way to solve this. We can use `:empty` to hide the div itself (except on the archive pages, where the empty `<a>` tag has an extra space in it for some reason. 😩). The root of the problem is just that Gutenberg shouldn't be rendering a post title if there's no post title. This isn't a huge priority bug, so I suggest we table it for the moment and see if there's movement on the Gutenberg end first.

mikachan:
Thanks @kjellr. I agree that the before states look better. I guess we're adding more and more CSS just to get around one issue, especially if we fix the archive and single post templates too. But I'm super happy to look into this if anyone thinks we should.

Should we close this issue for now? Or I've just spotted we could show a placeholder when there is no title by using a filter like this (from here https://github.com/WordPress/gutenberg/issues/29520):

```
add_filter(
	'the_title',
	function( $title ) {
		if ( ! is_admin() && empty( $title ) ) {
			$title = __( '(no title)', 'textdomain' );
		}

		return $title;
	}
);
```

We could do that instead for now?

image

kjellr:
I'm not a big fan of showing a placeholder in general — there are many cases where you might legitimately want no post title (For example, if your post is just an image post, or a video). 

I think it's ok to keep the issue open, but say that it's waiting for an upstream fix.

mikachan:
OK, thank you!

scruffian:
Since we're waiting for this to be fixed in Gutenberg I'm going to move it to Blocked for now.

mikachan:
This should be ready for another review. https://github.com/WordPress/gutenberg/issues/35224 has now been merged so there's no longer any markup for an empty title, and I've adjusted the spacing to hopefully match what we originally wanted.

scruffian:
This is looking good. I moved some of that CSS into the templates, but we still need that negative margin...

kjellr:
I wrestled with a really similar problem last week for Twenty Twenty-Two, and that helped clear my mind a bit. I just pushed a commit that sorts this out without the need for negative margins.

I also: 

- Stopped using the `100px` value in favor of the `blockGap` value, since that should scale a little better. 
- Tidied up the spacing beneath the post titles so it's more in line with [my recommendation here](https://github.com/Automattic/themes/pull/4720#discussion_r719665103).

Here's how posts look in the main feed with various combinations of featured images, titles, and content. The spacing for all these elements should be identical on the `single.html` template. 

Posts with Titles:

Featured Image|No Featured Image
---|---
Screen Shot 2021-10-27 at 3 13 32 PM|Screen Shot 2021-10-27 at 3 13 20 PM

Posts without Titles: 

Featured Image|No Featured Image
---|---
Screen Shot 2021-10-27 at 3 13 47 PM|Screen Shot 2021-10-27 at 3 14 01 PM

scruffian:
This looks good. We need to make the same changes to:
- Zoologist
- Geologist
- Payton

Maybe some others...

mikachan:
Awesome, thanks for this @kjellr!

Here are these changes in Geologist/Zoologist:

| Before      | After |
| ----------- | ----------- |
|  image    | image       |
|  image  |  image  |
| image  | image   |
| image  | image  |

mikachan:
Same changes to Payton:

| Before      | After |
| ----------- | ----------- |
|  image    | image     |
|   image   |   image     |
|  image    |   image    |
|  image   |   image   |

In some cases here I used just `var(--wp--style--block-gap)` on its own, as I thought the spacing looked too big with `calc(2 * var(--wp--style--block-gap))`.

payton/block-templates/page.html
```diff
@@ -1,7 +1,7 @@
 <!-- wp:template-part {"slug":"header"} /-->
 
-<!-- wp:group {"style":{"spacing":{"padding":{"right":"20px","left":"20px"}}},"layout":{"inherit":true}} -->
```
scruffian, line 3: Why did you remove the side padding?

scruffian:
In Geologist and Zoologist, for posts without a title, the content starts at a different place on the index and on the page itself:
Screenshot 2021-11-03 at 13 24 40
Screenshot 2021-11-03 at 13 24 31

scruffian:
I pushed a fix to align the no title content in Geologist and Zoologist

payton/block-templates/page.html
mikachan: It didn't seem to make a difference to the layout, so I removed it. Although I've just noticed that if the title goes onto multiple lines, this padding adds a bit more space. I'll add it back!

scruffian (approved):
@mikachan if you're happy with this I think we can merge when 11.9 lands.

mikachan:
Yeah that sounds good to me! Thanks for the help with this.

PR #4737, opened 2021-09-29
Quadrat: created new query block pattern

MaggieCabrera:

After having a conversation with @kriskorn and the team about what users want from a theme and mentioning static vs dynamic content, I wanted to try and see how hard it would be to make the episodes patterns into a dynamic query block. Turns out it's complex but not difficult. I had to use 2 query loops since the design is asymmetric and I ended adding a bunch of CSS to remove gaps and force the aspect ratio (TIL: [aspect-ratio](https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio) is a thing and it's SWEET) of the images since the featured image offers no control over the image sizes and we can't use them inside the media+text pattern either. Turned out looking quite ok to use as a pattern:

Screenshot 2021-09-29 at 10 23 31

/cc @beafialho @kjellr

### DIFF
--- a/quadrat/inc/patterns/query-episodes.php
+++ b/quadrat/inc/patterns/query-episodes.php
@@ -25,12 +25,12 @@ return array(
 	<!-- wp:query {"queryId":34,"query":{"perPage":"1","pages":0,"offset":"2","postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"exclude","inherit":false},"align":"wide"} -->
 	<div class="wp-block-query alignwide"><!-- wp:post-template -->
 	<!-- wp:columns {"className":"horizontal-query-pattern"} -->
-	<div class="wp-block-columns horizontal-query-pattern"><!-- wp:column {"width":"33%"} -->
-	<div class="wp-block-column" style="flex-basis:33%"><!-- wp:post-featured-image {"className":"image-no-margin mb-0"} /--></div>
+	<div class="wp-block-columns horizontal-query-pattern"><!-- wp:column {"width":"44%"} -->
+	<div class="wp-block-column" style="flex-basis:44%"><!-- wp:post-featured-image {"className":"image-no-margin mb-0"} /--></div>
 	<!-- /wp:column -->
 	
-	<!-- wp:column {"width":"66%"} -->
-	<div class="wp-block-column" style="flex-basis:66%"><!-- wp:cover {"overlayColor":"tertiary","minHeight":360,"className":"mt-0","style":{"spacing":{"padding":{"top":"30px","right":"40px","bottom":"20px","left":"40px"}}}} -->
+	<!-- wp:column {"width":"56%"} -->
+	<div class="wp-block-column" style="flex-basis:56%"><!-- wp:cover {"overlayColor":"tertiary","minHeight":360,"className":"mt-0","style":{"spacing":{"padding":{"top":"30px","right":"40px","bottom":"20px","left":"40px"}}}} -->
 	<div class="wp-block-cover has-tertiary-background-color has-background-dim mt-0" style="padding-top:30px;padding-right:40px;padding-bottom:20px;padding-left:40px;min-height:360px"><div class="wp-block-cover__inner-container"><!-- wp:post-title {"isLink":true,"linkTarget":"_blank","textColor":"primary"} /-->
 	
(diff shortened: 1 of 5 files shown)


MaggieCabrera:
I changed the column widths on the horizontal query to fit the original design:

Screenshot 2021-09-29 at 11 27 20

quadrat/inc/patterns/query-episodes.php
```diff
@@ -0,0 +1,42 @@
...
+	<div class="wp-block-query alignwide"><!-- wp:post-template -->
+	<!-- wp:group {"style":{"spacing":{"padding":{"top":"0px","right":"0px","bottom":"0px","left":"0px"}}},"className":"vertical-query-pattern","layout":{"inherit":false}} -->
+	<div class="wp-block-group vertical-query-pattern" style="padding-top:0px;padding-right:0px;padding-bottom:0px;padding-left:0px"><!-- wp:post-featured-image {"className":"image-no-margin mb-0"} /-->
+	
+	<!-- wp:cover {"overlayColor":"tertiary","minHeight":360,"className":"mt-0","style":{"spacing":{"padding":{"top":"30px","right":"40px","bottom":"20px","left":"40px"}}}} -->
+	<div class="wp-block-cover has-tertiary-background-color has-background-dim mt-0" style="padding-top:30px;padding-right:40px;padding-bottom:20px;padding-left:40px;min-height:360px"><div class="wp-block-cover__inner-container"><!-- wp:post-title {"isLink":true,"linkTarget":"_blank","textColor":"primary"} /-->
```
scruffian, line 18: Why are you making the linkTarget "_blank"?

MaggieCabrera: mmh, I didn't think much of it, I think I selected the option automatically. I can remove it

scruffian:
The third episode doesn't look great when there's no featured image:

Screenshot 2021-09-29 at 11 01 14

Let me try something...

scruffian:
What do you think of this?
Screenshot 2021-09-29 at 11 26 17

MaggieCabrera:
> What do you think of this? 

it would need extra css to fit the design. i just pulled from the branch and I don't see the image to the left of the text but on top of it, are we missing something?

scruffian:
>  i just pulled from the branch and I don't see the image to the left of the text but on top of it, are we missing something?

was that in the editor, or in the front end? there seems to be a bug in the editor where the featured image doesn't appear left aligned, but it is in the front end....

MaggieCabrera:
Not going to add it myself right now since you are working on it, but the horizontal query block needs to add `aspect-ratio: 1.12;` and `object-fit: cover;` to the image to fit the original design, regardless of the blocks we are using (cover or columns blocks)

MaggieCabrera:
> was that in the editor, or in the front end? there seems to be a bug in the editor where the featured image doesn't appear left aligned, but it is in the front end....

Either, this is the frontend:
Screenshot 2021-09-29 at 12 31 46

scruffian:
I pushed a different approach:
localhost_4759__p=402

scruffian:
Another iteration: This is responsive:
Screenshot 2021-09-29 at 13 18 45
Screenshot 2021-09-29 at 13 18 08
Screenshot 2021-09-29 at 13 17 52

MaggieCabrera:
Thanks so much for picking this up and improving it, @scruffian 

For me this is fine, I found a case where the image doesn't properly fit the container:

Screenshot 2021-09-29 at 14 56 41

This is fixed adding: `align-self: stretch` to the figure and `height: 100%` to the img tag. Ideally we won't need this when we can do this kind of pattern using media+text blocks.

The image I used for testing: 
a8d2a-8b6cc-shifaaz-shamoon-or0uertvyd0-unsplash

kjellr:
Thanks for exploring. I agree this should be a query pattern. 

My overarching comment is that I don't love the (admittedly small) amount of CSS that we need to accomplish this at the moment. Is there any way we can accomplish this without these styles? If not, what features need to exist in Gutenberg to make this possible without additional CSS?

MaggieCabrera:
the main problem comes from the very limited controls for the featured image sizing within GB and a few margin options related to the new gap in blocks. That and the lack of support for the featured image on the media + text block. I think we could ship this with the present CSS and then refactor it later even if we maintain the CSS in the theme for backwards compatibility.

MaggieCabrera:
I'm bringing this in and hopefully, we can refactor it in the future when media+text block supports featured images!

PR #4733, opened 2021-09-28
Blockbase: fix post title + content vertical spacing

jffng:
#### Changes proposed in this Pull Request:

This PR overrides the additional margin provided by Gutenberg's default vertical spacing between blocks ("gap"). 

Before | After
-------- | --------
Screen Shot 2021-09-28 at 4 03 48 PM | Screen Shot 2021-09-28 at 4 02 32 PM

#### Related issue(s):

Fixes #4727

Issue #4722, opened 2021-09-28
Blockbase: Under-header margin is incorrect


kjellr:
The spacing between a the site header and the page/post title (when there is one title) should be 80px. When you click through from the first post in a blog feed to the post itself, the title should remain in the same place. 

Blog/Archive:

Current|Intended
---|---
![Screen Shot 2021-09-28 at 09 44 04](https://user-images.githubusercontent.com/1202812/135099251-4c1c8a11-c666-4118-ac36-0eac9824cbe4.png)|![Screen Shot 2021-09-28 at 09 42 35](https://user-images.githubusercontent.com/1202812/135099303-8f90f40a-9239-4789-8163-74de8f1172b6.png)

Single Post: 

Current|Intended
---|---
![Screen Shot 2021-09-28 at 09 43 44](https://user-images.githubusercontent.com/1202812/135099387-cae1c008-75e4-4c23-83fa-7638f7b60569.png)|![Screen Shot 2021-09-28 at 09 43 13](https://user-images.githubusercontent.com/1202812/135099532-1a4fbfea-2828-4675-a98b-85f75a66def0.png)

kjellr:
I'm still seeing a vertical jump in the page title position when I visit a single post vs. the archive page. Then title should be in the same spot in both places. (I believe the spacing on the single post page looks more correct to me)

![jump](https://user-images.githubusercontent.com/1202812/135276050-0e11e8f0-eb19-4a99-aca0-5f2b28f37006.gif)

scruffian:
Closed by #4738.

PR #4738, opened 2021-09-29
Blockbase: Remove CSS to bring up the page content and instead remove the bottom margin from post titles

scruffian:
In https://github.com/Automattic/themes/pull/4733 we removed the margin on top of all `<main>` elements, but this is actually needed for the index template. This PR removes that change and instead removes the bottom margin on Post Title blocks, so that the content sits closer to the post title. This means that https://github.com/Automattic/themes/issues/4722 is now fixed again.

#### Related issue(s):
 https://github.com/Automattic/themes/pull/4733
https://github.com/Automattic/themes/issues/4722

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -205,10 +205,6 @@ pre {
 	flex-direction: column;
 }
 
-.wp-site-blocks > main {
-	margin-top: 0;
-}
-
 .site-footer-container {
 	margin-top: auto;
 }
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -657,6 +657,10 @@ p.has-drop-cap:not(:focus):first-letter {
 	margin-bottom: 0;
 }
 
+.wp-block-post-template .wp-block-post-featured-image {
+	line-height: 0;
+}
+
 .wp-block-pullquote.is-style-solid-color,
 .wp-block-pullquote {
 	text-align: var(--wp--custom--pullquote--typography--text-align);
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -219,7 +219,8 @@ body.admin-bar .wp-site-blocks {
 	min-height: calc( 100vh - var(--wpadmin-bar--height));
 }
 
-.entry-content > p {
+p.wp-block.wp-block-paragraph,
+p {
 	margin-top: 1em;
 	margin-bottom: 1em;
 }
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -220,8 +220,8 @@ body.admin-bar .wp-site-blocks {
 }
 
 p {
-	margin-top: var(--wp--custom--gap--vertical);
-	margin-bottom: var(--wp--custom--gap--vertical);
+	margin-top: var(--wp--style--block-gap);
+	margin-bottom: var(--wp--style--block-gap);
 }
 
 .wp-block-column :first-child, .wp-block-group :first-child {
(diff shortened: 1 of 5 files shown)


kjellr:
This _almost_ fixes it, but the spacing below the title is now very slightly less than it was before (and no longer exactly equal to the space between the post content and the post meta):

Before|After
---|---
Screen Shot 2021-09-29 at 9 46 43 AM|Screen Shot 2021-09-29 at 9 46 34 AM

scruffian:
I think that's caused by paragraphs not using block gap. I pushed a commit so paragraph spacing now uses the block gap settings.

kjellr:
We are really close! The only small issue I'm seeing now is that the gap between the title and the featured image (when there is one) is shorter than the the other gaps used: 

Screen Shot 2021-09-29 at 1 05 36 PM

If we can add more space above the top of the featured image so it matches the post meta gap (and let everything else stay the way it is right now), we're good to go.

jffng:
> The only small issue I'm seeing now is that the gap between the title and the featured image (when there is one) is shorter than the the other gaps used

I looked into this and it's caused by some extra space on the featured image:

Screen Shot 2021-09-29 at 4 56 02 PM

I added a commit https://github.com/Automattic/themes/pull/4738/commits/54234ac4b3798ef363ec4e9ef8e7760340d27ed0 to remove it. I am not sure if this should go in Gutenberg.

scruffian:
> I am not sure if this should go in Gutenberg.

I think so probably

kjellr:
Any objections to changing the custom gap baseline to 15px instead of 10px? I just pushed a quick commit to try that out, and I like it better than the previous iteration. Feels much more comfortable, and sort of minimizes any other spacing inconsistencies: 

Screen Shot 2021-09-30 at 7 44 38 AM

Screen Shot 2021-09-30 at 7 44 57 AM

Screen Shot 2021-09-30 at 7 45 06 AM

The only thing I _don't_ like with this extra space is that it should not be applied to paragraphs. I reset that back to using `1em`, which is a better default: 

1em|block-gap
---|---
Screen Shot 2021-09-30 at 7 47 04 AM|Screen Shot 2021-09-30 at 7 46 58 AM

The main issue is that that change doesn't get picked up in the editor unless I add some weird extra specific rules. 😕 Any ideas there?

scruffian:
This is looking good. I changed the selectors on the paragraph to make it work in the editor. It's not pretty but I couldn't think of a better option....

I'll open a Gutenberg issue...

kjellr (approved):
Excellent. I imagine in the future it would be nice to be able to just specify a `blockGap` value for the paragraph block itself to take care of that. 🤔

Anyway, in the meantime I think this is good to go!

scruffian:
Tracked here: https://github.com/WordPress/gutenberg/issues/35267

jffng:
If I'm following correctly, the space between paragraphs should be `1em` in the editor and front-end. This is not reflected on the front-end because the blockGap selector specificity is still winning: 

Screen Shot 2021-09-30 at 1 49 54 PM

kjellr:
> If I'm following correctly, the space between paragraphs should be 1em in the editor and front-end. This is not reflected on the front-end because the blockGap selector specificity is still winning:

Ugh yes. I just pushed another commit that specifically overrides that. It's pretty specific, but it's the only one I found that generally worked. 

(I tried doing just `.wp-site-blocks p`, but the specificity there messed with static `p` tags inside of other blocks like the Author block. So instead I focused on just overriding this single rule. 😕)

jffng (approved):
Since we've documented the paragraph spacing issue in GB, I think this is good to bring in. 

I also opened a PR about the extra space on the featured image block: https://github.com/WordPress/gutenberg/pull/35273

PR #4740, opened 2021-09-29
Skatepark: dynamic duotone support

### LINKED ISSUE #4411: Skatepark: Apply duotone effect to Featured Images by default
_Related to #4329 and #4406, but specifically for featured images._ 

---

Skatepark's featured images should all have the duotone feature applied by default:

Screen Shot 2021-08-16 at 11 06 11 AM

For FSE users should be able to turn this on and off by editing a template (there's [a pending issue in the Gutenberg repo](https://href.li/?https://github.com/WordPress/gutenberg/issues/34088) to add this functionality to the Featured Image block). We'd theoretically need to turn this into a customizer setting for folks without FSE as well.

MaggieCabrera:

This PR applies the changes from https://github.com/WordPress/gutenberg/pull/34667 to Skatepark. I have removed the duotone markup from block patterns and templates and defined it directly on theme.json for the affected blocks.

To test this, check that all the images in Skatepark show the duotone filter. Try inserting new ones too :D I went ahead and removed some markup issues from block patterns, check that none of the blocks throw any errors too.

TODO: we need to implement the change to the default filter when the user selects a new color palette on the customizer.

Closes #4411 #4329


scruffian:
>  we need to implement the change to the default filter when the user selects a new color palette on the customizer.

Should we...

1) Update the definition of the default filter in the theme.json

or

2) Create a new custom duotone definition and update the theme.json to point to the new custom duotone definition?

I think option 1 makes sense, but I thought it was worth discussing :)

MaggieCabrera:
> Should we...
> 
>     1. Update the definition of the default filter in the theme.json
> 
> 
> or
> 
>     1. Create a new custom duotone definition and update the theme.json to point to the new custom duotone definition?
> 
> 
> I think option 1 makes sense, but I thought it was worth discussing :)


Option 1 makes more sense to me because the number of blocks where the filter may be used can vary in the future and we'd have to keep maintaining it or look for every instance of the filter all over the json.

If we change the default filter though we would be losing the green option on the filter menu for the users to chose and duplicating one of the other options. Maybe we should define the green-black filter and the default one, so we don't lose the green one but we'd still always have a duplicate one. Now I'm not so sure!

MaggieCabrera:
We should consider if we want to add an option to disable duotone from the customizer as a default option for all the blocks.

scruffian:
I added a commit to add a updated default filter to the user CPT for Global Styles. This revealed a bug in the way Duotone works: https://github.com/WordPress/gutenberg/issues/35650

I think we're going to get blocked by this :/

MaggieCabrera:
> I added a commit to add a updated default filter to the user CPT for Global Styles. This revealed a bug in the way Duotone works: [WordPress/gutenberg#35650](https://github.com/WordPress/gutenberg/issues/35650)
> 
> I think we're going to get blocked by this :/

We could try option 2, instead of changing the default filter, we just assign a different one to the blocks that need it. It will also help with adding an option to turn duotone off altogether.

MaggieCabrera:
I've been thinking about this and option 2 is not a good idea because it wouldn't work with custom colors anyway, just with the preset palettes from the theme :(

scruffian:
> I've been thinking about this and option 2 is not a good idea because it wouldn't work with custom colors anyway, just with the preset palettes from the theme :(

What about an option 3 where we create a new filter?

MaggieCabrera:
> > I've been thinking about this and option 2 is not a good idea because it wouldn't work with custom colors anyway, just with the preset palettes from the theme :(
> 
> What about an option 3 where we create a new filter?

We should pair on this, I think!

MaggieCabrera:
Sorry in advance because this code is probably horrible, but it actually works in the frontend! 🎉 

Screenshot 2021-10-21 at 18 01 32

What's left to do:

- Make the preview work. We couldn't figure out how, since we don't really know which selectors we need to target for a given theme. Maybe we just need to force a reload?
- The way this works now we are replacing the theme's duotone palettes with the user selected one. That is mostly ok but ideally we would merge that with the theme ones.
- This is assuming that both primary and background colors are defined, we may need to double check that those exist. Maybe not all themes want to use those two as the duotone combination, but that's out of the scope of this PR..

Thanks @mikachan for your help with this!

MaggieCabrera:
Also, I don't think this PR works for testing the GB PR anymore, sorry!

blockbase/inc/customizer/wp-customize-colors-preview.js
```diff
@@ -1,15 +1,17 @@
...
+				value.bind( ( newValue ) => {
+					paletteItem.color = newValue;
+					blockBaseUpdateColorsPreview( userColorPalette );
+				} );
 			} );
 		} );
-	} );
-}
+	}
+} )();
```
scruffian: Why did you add this closure?

blockbase/inc/customizer/wp-customize-colors-preview.js
```diff
@@ -24,4 +26,45 @@ function blockBaseUpdateColorsPreview( palette ) {
...
+	if ( userColorDuotone ) {
+		let colors = palette.map( ( paletteItem ) => paletteItem.color );
+		//we are inverting the order when we have a darker background so that duotone looks good.
+		colors = colors.sort( ( first, second ) => {
+			return colord( first ).brightness() > colord( second ).brightness();
+		} );
+		const colorValues = getValuesFromColors( colors );
+
+		if ( document.querySelector( '#wp-duotone-default-filter' ) ) {
+			updateDuotoneFilter( 'wp-duotone-default-filter', colorValues );
```
scruffian: It's a little tricky to understand why there is no hash on this string but there is on line above. It might be better to just move the hash here for consistency.

blockbase/inc/customizer/wp-customize-colors-preview.js
```diff
@@ -24,4 +26,45 @@ function blockBaseUpdateColorsPreview( palette ) {
 		'global-styles-colors-customizations-inline-css'
 	);
 	styleElement.innerHTML = innerHTML;
+
+	if ( userColorDuotone ) {
+		let colors = palette.map( ( paletteItem ) => paletteItem.color );
+		//we are inverting the order when we have a darker background so that duotone looks good.
+		colors = colors.sort( ( first, second ) => {
```
scruffian: We can write this in one line

blockbase/inc/customizer/wp-customize-colors-preview.js
```diff
@@ -24,4 +26,45 @@ function blockBaseUpdateColorsPreview( palette ) {
...
+			return colord( first ).brightness() > colord( second ).brightness();
+		} );
+		const colorValues = getValuesFromColors( colors );
+
+		if ( document.querySelector( '#wp-duotone-default-filter' ) ) {
```
scruffian: We could move this check inside the function

blockbase/inc/customizer/wp-customize-colors-preview.js
MaggieCabrera: I think I did to try and make the refresh work, we don't need it anymore

blockbase/inc/customizer/wp-customize-colors-preview.js
```diff
@@ -24,4 +26,45 @@ function blockBaseUpdateColorsPreview( palette ) {
...
+		.setAttribute( 'tableValues', colors.r.join( ' ' ) );
+	document
+		.querySelector( '#' + filterID + ' feFuncG' )
+		.setAttribute( 'tableValues', colors.g.join( ' ' ) );
+	document
+		.querySelector( '#' + filterID + ' feFuncB' )
+		.setAttribute( 'tableValues', colors.b.join( ' ' ) );
+}
+
+function getValuesFromColors( colors = [] ) {
```
scruffian, line 63: It would be good to add a comment that this line is from Gutenberg

scruffian:
This is looking good, but I don't see all the filters in the editor:
Screenshot 2021-10-26 at 21 49 51

MaggieCabrera:
> This is looking good, but I don't see all the filters in the editor: Screenshot 2021-10-26 at 21 49 51

I tried to merge the user's custom filter with theme preset ones and it's mostly working but in some cases some of the colors are not saving correctly if you var_dump the result of the array_merge from my commit. I can't seem to find what's causing that!

scruffian (approved):
LGTM. Don't forget to update the changelog. This should probably be a new minor version!

PR #4743, opened 2021-09-29
Blockbase: remove generic has-background padding

### LINKED ISSUE #4735: Quadrat: episode pattern regression
The episode pattern now shows an extra padding that shouldn't be there:

Screenshot 2021-09-29 at 09 31 39

This is being caused by this rule:

```
.has-background:not(.wp-block-separator) {
    padding: var(--wp--custom--gap--vertical) var(--wp--custom--gap--horizontal);
}
```

I don't know if I agree that everything with background color needs a padding. And if some blocks do, shouldn't this be something that GB does?

jffng:

In #4629, a standard padding was added to all blocks with the `has-background` class. Based on discussion [here](https://github.com/Automattic/themes/issues/4735#issuecomment-930155711), I don't think that's the best approach. This PR removes that CSS and fixes #4735. 

Before | After
-------- | -------
Screenshot 2021-09-29 at 09 31 39 | Screen Shot 2021-09-29 at 4 26 07 PM

Will removing this CSS cause regressions in any other Blockbase & co patterns?

#### Rationale

The [original issue](https://github.com/Automattic/themes/issues/4617) suggested that group blocks with a background color were missing padding. I think it's better to not supply any padding by default, since padding on group blocks is customizable via the editor.

In addition, we could consider improving how this padding is added / calculated in Gutenberg, as suggested here https://github.com/WordPress/gutenberg/issues/34588

#### Related issue(s):

Solves https://github.com/Automattic/themes/issues/4735

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -244,14 +244,6 @@ p {
 	padding: 0;
 }
 
-.has-background:not(.wp-block-separator) {
-	padding: var(--wp--custom--gap--vertical) var(--wp--custom--gap--horizontal);
-}
-
-.has-background :last-child {
-	margin-bottom: 0;
-}
-
 /**
  * Elements
  * - Styles for basic HTML elemants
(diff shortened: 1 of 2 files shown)


scruffian:
I say lets bring this in sooner rather than later as it will mess with people's sites.

scruffian (approved):

PR #4744, opened 2021-09-29
Replace query-loop with post-template

### LINKED ISSUE #4731: Update template references to query-loop
Some templates still reference the "query-loop" block. We should update them to reference its new name "post-template" as well as respective CSS classes:

Screen Shot 2021-09-28 at 2 35 38 PM

jffng:

Renames instances of query-loop block to post-template.

#### Related issue(s):

Fixes #4731

### DIFF
--- a/geologist/assets/theme.css
+++ b/geologist/assets/theme.css
@@ -218,7 +218,7 @@ ul ul {
 	margin-left: 20%;
 }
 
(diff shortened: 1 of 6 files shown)


mikachan (commented):
This looks good to me. I can't see any other query-loops 👍

PR #4745, opened 2021-09-29
Quadrat: Fix Code block text color in the editor

### LINKED ISSUE #4625: Quadrat: Text unreadable in Code Block while editing
Though it shows up correctly on the front end, in the editor the text is currently looking too dark. It should be pink to pass contrast.

Captura de ecrã 2021-09-16, às 15 27 17

danieldudzic:
Closes #4625

### DIFF
--- a/quadrat/child-theme.json
+++ b/quadrat/child-theme.json
@@ -265,7 +265,8 @@
 					"width": "0px"
 				},
 				"color": {
-					"background": "var(--wp--custom--color--tertiary)"
+					"background": "var(--wp--custom--color--tertiary)",
+					"text": "var(--wp--custom--color--primary)"
 				},
 				"typography": {
 					"fontSize": "var(--wp--preset--font-size--normal)",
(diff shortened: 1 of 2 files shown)


scruffian:
The problem here seems to be that WordPress.com is loading the theme.css files for blocks even though we don't add support for that in the theme. I think that's the problem we should address as I expect it causes other issues.

jeffikus:
@scruffian is this something that wpcom does globally for all themes?

danieldudzic:
@scruffian What's the process of including theme.css usually? What _should_ be causing the theme.css file to be enqueued?

scruffian:
There's two theme.css files, one that is used by the child theme, in this case Quadrat. That file is being enqueued by functions.php in Quadrat.

There's also a theme.css file that is used by blocks, which Gutenberg adds. This provides opinionated styles for blocks. This file is only used when a theme adds support for it via `add_theme_support('wp-block-styles')`, which we don't do for Blockbase themes. I'm not sure why this file is being added on wpcom... yet!

scruffian:
Fixed by D67673-code

Issue #4704, opened 2021-09-24
Separate Blockbase "opinion" from "ponyfill"


pbking:
There are a number of things Blockbase is doing that are opinion rather than polyfill.

This issue covers the idea of extracting and leveraging that opinion in the best way.

See: https://github.com/Automattic/themes/pull/4692#issuecomment-926833152

MaggieCabrera:
We already have something like this with the button styling. We provide a Blockbase base and an alternative option (the outline version that Quadrat, Skatepark, Geologist, and I believe Videomaker are using). In the case of the button styles, we have an extra sass partial that is added by the theme that needs it on top of the base styles that Blockbase provides but it's not being added by Blockbase itself to avoid bloating its CSS.

Ideally, we would have Blockbase be the simplest it can be and add extra options as opt-in for the child themes but that may not be the case every time.

In the case of the header, since most of the time it's going to require extra css, I think it would be a good thing to provide multiple options for it from the parent theme perspective. Have multiple sass partials for each header style and have the child theme decide which one they are going to implement, as well as a different template part provided too. Kind of like a block style but for the whole template part?

jeffikus:
Closing this, can we separate out the tasks into individual issues?

PR #4752, opened 2021-09-30
Quadrat: Fix post meta offset

### LINKED ISSUE #4748: Quadrat: Slight offset in tagline and post meta
I was checking Quadrat on mobile and there's a slight offset in tagline and post meta. 

The spacing is also too tight between post meta and the post title.

Captura de ecrã 2021-09-30, às 12 34 59

The misalignment is also happening in larger screens:

Captura de ecrã 2021-09-30, às 13 02 54

danieldudzic:
Closes #4748

### DIFF
--- a/quadrat/block-template-parts/header.html
+++ b/quadrat/block-template-parts/header.html
@@ -2,7 +2,7 @@
 <header class="wp-block-group site-header" style="padding-bottom:170px">
 	<!-- wp:site-logo /-->
 	<!-- wp:site-title /-->
-	<!-- wp:site-tagline {"style":{"spacing":{"padding":{"top":"0px","right":"0px","bottom":"0px","left":"0px"},"margin":{"right":"auto"}},"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
+	<!-- wp:site-tagline {"style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
 	<!-- wp:navigation {"isResponsive":true,"__unstableLocation":"primary"} /-->
 </header>
 <!-- /wp:group -->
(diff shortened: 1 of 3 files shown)


danieldudzic:
@jffng Do you think this approach is correct, here?

quadrat/sass/templates/_meta.scss
```diff
@@ -28,6 +28,9 @@
 			background: transparent;
 			-webkit-mask-image: none;
 			mask-image: none;
+			height: 0;
```
jffng: Could we just use `content: none;` here and remove the rest of CSS specific to the `:before` element?

quadrat/block-template-parts/header.html
```diff
@@ -2,7 +2,7 @@
 <header class="wp-block-group site-header" style="padding-bottom:170px">
 	<!-- wp:site-logo /-->
 	<!-- wp:site-title /-->
-	<!-- wp:site-tagline {"style":{"spacing":{"padding":{"top":"0px","right":"0px","bottom":"0px","left":"20px"},"margin":{"right":"auto"}},"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
+	<!-- wp:site-tagline {"style":{"spacing":{"padding":{"top":"0px","right":"0px","bottom":"0px","left":"0px"},"margin":{"right":"auto"}},"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
```
jffng: Pushed a commit to solve the other part of the issue.

Before | After
------- | ------
Screen Shot 2021-09-30 at 11 11 04 AM | Screen Shot 2021-09-30 at 11 10 48 AM

scruffian: Can we just remove the padding definition altogether?

quadrat/sass/templates/_meta.scss
scruffian: I think this makes a good case for making the post meta rules opt in in blockbase, as suggested in https://github.com/Automattic/themes/issues/4704

MaggieCabrera:
I'm bringing this in, we can discuss making the post meta opt-in in https://github.com/Automattic/themes/issues/4704

Issue #4768, opened 2021-10-02
Blockbase: fatal error previewing theme without Gutenberg


RavanH:
## Steps to replicate

On an install with default theme and Gutenberg disabled (or not installed); activate the Blockbase live preview.

## Result

A fatal error:
```
Fatal error: Uncaught Error: Class 'WP_Theme_JSON_Resolver_Gutenberg' not found in /xxxxxxxx/wp-content/themes/blockbase/inc/customizer/wp-customize-colors.php:56
```

## Expected

Empty preview or maybe a visual message "Error: This theme (Blockbase) uses Full Site Editing, which requires the Gutenberg plugin to be activated." as is shown when a block theme is immediately activated without Gutenberg.

## Proposal

Only include Customizer files (colors/palette/fonts) when class WP_Theme_JSON_Resolver_Gutenberg is available...

PR #4769, opened 2021-10-02
fix WP_Theme_JSON_Resolver_Gutenberg not found

RavanH:

Only include Customizer files when Gutenberg is active.

Also: move `add_action( 'after_setup_theme', 'blockbase_support', 9 );` outside conditional.

#### Related issue(s):
https://github.com/Automattic/themes/issues/4768

### DIFF
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -57,8 +57,8 @@ if ( ! function_exists( 'blockbase_support' ) ) :
 		);
 
 	}
-	add_action( 'after_setup_theme', 'blockbase_support', 9 );
 endif;
+add_action( 'after_setup_theme', 'blockbase_support', 9 );
 
 /**
  *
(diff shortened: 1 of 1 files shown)


blockbase/functions.php
```diff
@@ -137,9 +137,11 @@ function blockbase_restore_customizer() {
 /**
  * Customize Global Styles
  */
-require get_template_directory() . '/inc/customizer/wp-customize-colors.php';
-require get_template_directory() . '/inc/customizer/wp-customize-color-palettes.php';
-require get_template_directory() . '/inc/customizer/wp-customize-fonts.php';
+if ( class_exists( 'WP_Theme_JSON_Resolver_Gutenberg' ) ) {
```
scruffian, line 140: This is a good idea, thanks.

blockbase/functions.php
```diff
@@ -57,8 +57,8 @@ function( $settings ) {
 		);
 
 	}
-	add_action( 'after_setup_theme', 'blockbase_support', 9 );
 endif;
+add_action( 'after_setup_theme', 'blockbase_support', 9 );
```
scruffian, line 61: I'm not sure we want to do this. The way I see it, if that function is defined elsewhere, its probably by some totally unrelated code, in which case we probably want to leave it to run itself...

RavanH: Ah, okay? I assumed it to be a "pluggable" function for child themes, like many other Automattic themes use ;)

scruffian: Oh yes, I see what you mean. 👍

PR #4773, opened 2021-10-04
Geologist: Update header spacing

scruffian:
This updates the spacing for Geologist so that the vertical and horizontal spacing is the same. We don't currently have a way to set responsive margins for the header, but I'm not sure we need it. With the variable set to min(30px, 5vw), on smaller devices 5vw will be used instead of the pixel value anyway.

Before:
localhost_4759__p=371 (2)

After:
localhost_4759__p=371 (3)

#### Related issue(s):
https://github.com/Automattic/themes/issues/4687

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -832,6 +832,10 @@ div.wp-block-query-pagination .wp-block-query-pagination-numbers .current {
 	width: var(--wp--custom--separator--width);
 }
 
+p.wp-block-site-tagline {
+	margin: 0;
+}
+
 .wp-block-file .wp-block-file__button {
 	border-width: 0;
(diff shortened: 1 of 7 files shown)


kjellr:
This looks  great on a fresh site. I am seeing this tiny weird indent on the site tagline though:
 
Screen Shot 2021-10-04 at 11 58 12 AM

kjellr:
That last commit seems to fix the issue: 
Screen Shot 2021-10-04 at 2 25 37 PM

... but I'm seeing a bunch of errors all of a sudden: 

Screen Shot 2021-10-04 at 2 25 28 PM

Those weren't there earlier. 🤔

scruffian:
Ah sorry that's my fault, I missed a `{`. It should be ok now...

kjellr (approved):
This works well now, thanks!

Issue #4649, opened 2021-09-20
videomaker: Templates: Page templates - Search


scruffian:
@beafialho how's this for the search template?
<img width="858" alt="Screenshot 2021-09-20 at 17 05 52" src="https://user-images.githubusercontent.com/275961/134035594-1d308f3c-0ed4-4bdc-a567-d9b4c9f52f3f.png">

beafialho:
Can we increase the spacing between the header and the content and make it the same as in post pages? Let's try 170px.

<img width="1166" alt="Captura de ecrã 2021-09-30, às 10 00 50" src="https://user-images.githubusercontent.com/65220155/135421969-819f955e-3fe4-4b42-ab80-a89e6bdac39c.png">

mikachan:
I can see the 170px gap locally on my search results page:

![image](https://user-images.githubusercontent.com/1645628/136775109-be20fc78-ed42-4de2-b754-489d1e99d240.png)

There's some additional space (`--wp--style--block-gap` at 30px) above the main content coming from GB:

![image](https://user-images.githubusercontent.com/1645628/136775797-dfd16ea7-c397-4fd8-be19-f9b407210c75.png)

I don't know if either the 170px gap should take this into account (so the space is exactly 170px) or if we're OK with it like this?

scruffian:
The title on the search page matches the index, so I think that's fine, I'm going to close this.

PR #4774, opened 2021-10-04
Videomaker: Update header and footer

scruffian:
This adds a gap to the bottom of the header and the top of the footer, to match the design in https://github.com/Automattic/themes/issues/4649

Screenshot:
localhost_4759__p=371 (4)

The footer was reverted by accident, so this reinstates that code.

#### Related issue(s):
https://github.com/Automattic/themes/issues/4649


mikachan (commented):
This is looking good to me when I test locally, and it looks like it matches what @beafialho requested.

Would we need to update the [heading gap variable](https://github.com/Automattic/themes/pull/4777) based on these changes?

scruffian:
> Would we need to update the heading gap variable based on these changes?

Yeah I'll update the other PR when this one merges. Thanks!

beafialho:
Sorry for the delayed response, this looks good to me :+1:

PR #4777, opened 2021-10-04
Blockbase: Make the heading gap a variable

scruffian:
This adds a new variable to custom, making the gap between the header and the content a value which can be configured using theme.json. In the future we hope that we'll be able to inherit the markup for headers from the parent theme, so that child themes can customize this gap without needing to create a new header.html file.

The look of the themes should be unchanged.

### DIFF
--- a/blockbase/block-template-parts/header.html
+++ b/blockbase/block-template-parts/header.html
@@ -1,4 +1,4 @@
-<!-- wp:group {"tagName":"header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"var(--wp--custom--header--spacing--padding--bottom)","top","var(--wp--custom--header--spacing--padding--top)"}}},"className":"site-header"} -->
+<!-- wp:group {"tagName":"header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"var(--wp--custom--header--spacing--padding--bottom)","top":"var(--wp--custom--header--spacing--padding--top)"}}},"className":"site-header"} -->
 <header class="wp-block-group site-header" style="padding-bottom:var(--wp--custom--header--spacing--padding--bottom);padding-top:var(--wp--custom--header--spacing--padding--top);">
 	<!-- wp:site-logo /-->
 	<!-- wp:site-title /-->
--- a/blockbase/block-template-parts/header.html
+++ b/blockbase/block-template-parts/header.html
@@ -1,5 +1,5 @@
-<!-- wp:group {"className":"site-header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"var(--wp--custom--gap--header)"}}}} -->
-<div class="wp-block-group site-header" style="padding-bottom:var(--wp--custom--gap--header)">
+<!-- wp:group {"tagName":"header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"var(--wp--custom--header--spacing--padding--bottom)","top","var(--wp--custom--header--spacing--padding--top)"}}},"className":"site-header"} -->
+<header class="wp-block-group site-header" style="padding-bottom:var(--wp--custom--header--spacing--padding--bottom);padding-top:var(--wp--custom--header--spacing--padding--top);">
 	<!-- wp:site-logo /-->
 	<!-- wp:site-title /-->
 	<!-- wp:site-tagline {"fontSize":"tiny"} /-->
--- a/blockbase/block-template-parts/header.html
+++ b/blockbase/block-template-parts/header.html
@@ -1,5 +1,5 @@
-<!-- wp:group {"className":"site-header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"80px"}}}} -->
-<div class="wp-block-group site-header" style="padding-bottom:80px">
+<!-- wp:group {"className":"site-header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"var(--wp--custom--gap--header)"}}}} -->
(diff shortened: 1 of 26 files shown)


mayland-blocks/theme.json
```diff
@@ -246,7 +246,8 @@
 				}
 			},
 			"gap": {
-				"baseline": "10px",
+				"baseline": "15px",
```
MaggieCabrera: why the change from 10 to 15px?

scruffian: I think that's coming from blockbase - the theme.json must not have been rebuild when that change was made. I'm not sure if we should reset it or not - often baseline is about 50% of the horizontal/vertical spacing...

mikachan:
I really like this idea. +1 for anything that makes alignment slightly easier!

mayland-blocks/theme.json
scruffian: I've set the baseline back to 10px in every theme where that was the case before...

scruffian:
I've updated this to include the Videomaker header too, so it's ready for another review.

mikachan (commented):
Looking good all round to me. The only one I spotted that didn't look right was Videomaker, and I think it was just because `gap` wasn't sitting under `custom`, so I pushed a commit for that.

blockbase/theme.json
```diff
@@ -239,6 +239,7 @@
 			},
 			"gap": {
 				"baseline": "15px",
+				"header": "80px",
```
MaggieCabrera: Maybe it's a nitpick, but I feel like it's hard to guess here that we are only talking about the bottom padding without looking at the template part. Maybe this should be something like:

```
"header": {
    "bottom": "80px",
    "top": "var(--wp--custom--gap--vertical)"
}
```

and remove [this line from the CSS](https://github.com/Automattic/themes/blob/029c1c316f348d02cfb18356b24632b6f242479c/blockbase/sass/base/_header.scss#L4). It feels closer to what GB may do in the future and it's customizable by the user from the FSE.

MaggieCabrera: what do you think, @scruffian and @mikachan ?

mikachan: I think that's a good idea, I didn't realise how ambiguous it might look with just 'header'. Plus if we add the 'top' spacing in there too, that means we can remove it from the CSS as you say.

scruffian: My only hesitation is that I this same value could (maybe should) be used at the bottom of the content too. I did originally call it "content" for that reason. On balance though I think your suggestion makes sense.

scruffian:
I updated this to use the following format:

```
			"header": {
				"spacing": {
					"padding": {
						"bottom": "80px",
						"top": "var(--wp--custom--gap--vertical)"
					}
				}
			},
			
```

Ready for another review

MaggieCabrera:
This is what I see:

Skatepark's header broke a bit:

Trunk | This PR 
--- | ---
Screenshot 2021-10-11 at 10 50 26 | Screenshot 2021-10-11 at 10 50 08 

Geologist:

Trunk | This PR 
--- | ---
Screenshot 2021-10-11 at 10 52 22 | Screenshot 2021-10-11 at 10 52 09

Seedlet looks good but the header sits a tiny bit lower than originally:

Trunk | This PR 
--- | ---
Screenshot 2021-10-11 at 10 52 55 | Screenshot 2021-10-11 at 10 53 06

Mayland is broken and it's pushing the content all the way to the bottom:

Trunk | This PR 
--- | ---
Screenshot 2021-10-11 at 10 53 50 | Screenshot 2021-10-11 at 10 54 06

Quadrat:
Trunk | This PR 
--- | ---
Screenshot 2021-10-11 at 10 54 49 | Screenshot 2021-10-11 at 10 54 55

Blockbase is also pushing the content down:

Trunk | This PR 
--- | ---
Screenshot 2021-10-11 at 10 55 25 | Screenshot 2021-10-11 at 10 55 40

I'm running the latest GB trunk here

scruffian:
Eugh, sorry there was a `,` instead of a `:`. It should be working now...

blockbase/block-template-parts/header.html
```diff
@@ -1,8 +1,8 @@
-<!-- wp:group {"className":"site-header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"80px"}}}} -->
-<div class="wp-block-group site-header" style="padding-bottom:80px">
+<!-- wp:group {"tagName":"header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"var(--wp--custom--header--spacing--padding--bottom)","top":"var(--wp--custom--header--spacing--padding--top)"}}},"className":"site-header"} -->
+<header class="wp-block-group site-header" style="padding-bottom:var(--wp--custom--header--spacing--padding--bottom);padding-top:var(--wp--custom--header--spacing--padding--top);">
```
pbking, line 2: This gives us "double nested header" elements as the template part is already expressed as a header.

pbking:
I'm actually not terribly keen on the concept.  It's cool for "theme makin'" but less so for "theme customizin'".  Before the change as a user I have the option to control that padding in the header with the current value being shown.



Admittedly it's not terribly easy to find, but it's available to me to adjust as I see fit.

With this change the easiest means to adjust this value is via JSON configuration, not block configuration.  The block IS adjustable, but it's not obvious that the block actually has a value that can/should be adjusted.  (Note that CHANGING the top/bottom will overwrite the 'hidden' css variables but it's a questionable user experience.)

scruffian:
You have a good point. Kjell said the same thing about TT2. I do think that it's worth moving the header padding out of the CSS though, so maybe what we should do is just replace the variables with the explict values in the template files...

scruffian:
On reflection I'm not sure this is a good idea. If we keep the values in the HTML it makes the templates more reusable across themes...

Issue #4776, opened 2021-10-04
Blockbase: paragraph custom link color ignored


RavanH:
## Steps to replicate

Create a paragraph block with a custom text color, and set a custom link color too.

## Result

Link color is forced to either --wp--style--color--link or --wp--custom--color--primary by ponyfill.css

## Expected

The selected custom link color to be used...

RavanH:
The exact style rule in question is:

```
p.has-text-color a {
	color: var(--wp--style--color--link, var(--wp--custom--color--primary));
}
```

But when is the style variable --wp--style--color--link ever created? What mystical theme.json node could be responsible for that? I have not found it... So maybe the whole rule makes no sense?

danieldudzic:
Hi @RavanH,

Thanks for flagging this. It seems that this rule was present to overwrite some core styles. I believe it's no longer necessary, so I've removed it.

pbking:
The `--wp--style--color--link` variable was created by Gutenberg based on the link color settings in `theme.json` but the links are now colored (and configured) a different way so you're right @danieldudzic  this is no longer relevant.  That bit was in place because paragraphs with a custom color would overwrite the link color with the custom color in earlier versions.

This change LGTM, thank's ya both.

PR #4779, opened 2021-10-04
Fix ignored link color

RavanH:

Make style rule for paragraph links with custom text color not apply when the paragraph also has a custom link color set.

#### Related issue(s):

#4776


MaggieCabrera:
Thank you for your contribution, sorry for such a late reply.

This needs to be fixed on the sass and then build the CSS with `npm run build`

PR #4781, opened 2021-10-04
Blockbase: Fix header alignment

scruffian:
This ensures that the text in the blockbase header is aligned to the baseline even if the font sizes are different:

Before:
Screenshot 2021-10-04 at 17 16 56
Screenshot 2021-10-04 at 17 16 02

After:
Screenshot 2021-10-04 at 17 14 58
Screenshot 2021-10-04 at 17 14 37

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -154,6 +154,7 @@ pre {
 }
 
 .wp-site-blocks .site-header {
+	align-items: baseline;
 	justify-content: start;
 	overflow: inherit;
 	padding-top: var(--wp--custom--gap--vertical);
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -174,6 +174,10 @@ pre {
 	}
 }
 
+.wp-site-blocks .site-header .wp-block-site-logo {
+	align-self: center;
+}
+
 @media (max-width: 599px) {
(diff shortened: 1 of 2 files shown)


kjellr:
Things look just very slightly off in Quadrat for me still: 

Screen Shot 2021-10-04 at 2 38 16 PM

Blockbase is _super_ close, but is just a tiny bit off for some reason:

Screen Shot 2021-10-04 at 2 39 26 PM


Also in Blockbase + Quadrat, the nav doesn't appear to line up correctly with the baseline: 

Screen Shot 2021-10-04 at 2 41 49 PM

scruffian:
I'm really sorry I had pushed the wrong code to this branch - try now...

mikachan:
This looks good without a logo:

image

But then with a logo, the baseline alignment looks like this:

image

Changing the alignment to `align-items: center` looks more similar to what's on trunk, but I don't know if I'm missing something like a design decision around `baseline`..

scruffian:
Thanks for testing @mikachan. The problem I am trying to deal with is when the items are different font sizes:
Screenshot 2021-10-11 at 16 39 35

In this case the bottom of the text doesn't line up.

The only way round this I can find is to set everything to baseline and the site logo to center, which looks like this:

Screenshot 2021-10-11 at 16 39 48

However, although now all the text is aligned, the logo alignment is still off. I think the situation on trunk is preferable so I'm going to close this. Feel free to open if you can think of a solution!

PR #4771, opened 2021-10-04
Quadrat/Geologist: Match index spacing to single post & page

mikachan:
#### Changes proposed in this Pull Request:
This PR adjusts some margins and spacer blocks in order to make the post meta and title spacing consistent across the index, post, and page templates. 

I'm not sure about the way I've approached this, especially with the spacer block changes. 

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/4757

Issue #4757, opened 2021-10-01
Quadrat: Vertical position of the post meta + title should be consistent


kjellr:
When navigating from the main blog page to a single post, the vertical position of the first title + post meta should be identical. Currently, it is not. The title on single post pages should align to this same vertical position as well. \

The spacing on the single post page is the correct spacing to use in all three cases. 

## Screenshots of the issue currently:

Blog Index: 
<img width="1293" alt="Screen Shot 2021-10-01 at 9 32 38 AM" src="https://user-images.githubusercontent.com/1202812/135628532-d04c6b91-ddd9-4a75-8632-deb880d19f8f.png">

Single Post:

<img width="1294" alt="Screen Shot 2021-10-01 at 9 32 16 AM" src="https://user-images.githubusercontent.com/1202812/135628501-1271f6bc-900f-4f66-9af7-4789cac09da1.png">

Page: 

<img width="1293" alt="Screen Shot 2021-10-01 at 9 32 52 AM" src="https://user-images.githubusercontent.com/1202812/135628559-82298c1a-3ee9-4978-9aa7-212782d67689.png">

(This is also happening in Geologist: https://github.com/Automattic/themes/issues/4758)

MaggieCabrera:
Make this change to Geologist too :D

mikachan:
Fixed with https://github.com/Automattic/themes/pull/4783.

PR #4783, opened 2021-10-04
Quadrat: Ensure that the position of the post title is the same on all pages

scruffian:
This updates the header for Quadrat so that the Post/Page title is the same on the index, single and page templates.

An alternative to https://github.com/Automattic/themes/pull/4771, which I think is a bit simpler. If this is preferred then I can make the same changes to Geologist too.

After:
localhost_4759__page_id=2 preview_id=2 preview_nonce=01dd0db364 preview=true
localhost_4759__p=167 (1)
localhost_4759__paged=8

#### Related issue(s):
https://github.com/Automattic/themes/issues/4757


mikachan:
Thanks for this @scruffian, I definitely prefer it to my PR because it's simpler and doesn't require any CSS changes. The titles and post meta look to be in the same position to me.

mikachan (approved):

PR #4795, opened 2021-10-06
Blockbase + co: Fix navigation alignment

MaggieCabrera:

After https://github.com/WordPress/gutenberg/pull/35077 was merged the navigation for Blockbase, Quadrat, Geologist, and Videomaker were broken. This PR fixes this by applying the alignment correctly in the block markup and removing the unused CSS. The rest of the children were behaving correctly.

Before:

Screenshot 2021-10-06 at 16 25 47

After:

Screenshot 2021-10-06 at 16 25 21

We need to hold on to this PR before we merge because dotcom doesn't have the GB just yet.

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -165,7 +165,6 @@ pre {
 
 .wp-site-blocks .site-header .wp-block-navigation {
 	flex-grow: 1;
-	justify-content: flex-end;
 }
 
 @media (max-width: 599px) {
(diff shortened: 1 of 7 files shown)


scruffian (approved):
This fixes it for me

scruffian:
In my testing this works fine on dotcom, so we don't need to to wait for the next GB release...

MaggieCabrera:
> In my testing this works fine on dotcom, so we don't need to to wait for the next GB release...

You are totally right! we can merge then

PR #4796, opened 2021-10-06
Quadrat: Allow long titles to wrap the same way on post pages and on the index

scruffian:
This keeps the look of the page titles the same on the post/page templates and the index template.

One side effect is that the featured image is in a different place on each page. Which one is correct?

Also, should we make this change in Geologist?

Post:
Screenshot 2021-10-06 at 16 21 27

Index:
Screenshot 2021-10-06 at 16 21 37

### DIFF
--- a/quadrat/block-templates/index.html
+++ b/quadrat/block-templates/index.html
@@ -2,13 +2,17 @@
 
 <!-- wp:query {"tagName":"main","layout":{"inherit":true},"queryId":1,"query":{"perPage":5,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":"","inherit":true}} -->
 <main class="wp-block-query">
-	<!-- wp:post-template -->
-		<!-- wp:group {"layout":{"inherit":true}} -->
+	<!-- wp:post-template {"align":"wide"} -->
+		<!-- wp:group -->
 		<div class="wp-block-group">
 		<!-- wp:post-date {"isLink":false,"textAlign":"center","style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
 		<!-- wp:post-title {"isLink":true, "level": 3, "textAlign":"center"} /-->
+		<!-- wp:group {"layout":{"inherit":true}} -->
+		<div class="wp-block-group">
 		<!-- wp:post-featured-image {"isLink":true} /-->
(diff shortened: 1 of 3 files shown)


beafialho:
> One side effect is that the featured image is in a different place on each page. Which one is correct?

The image width should be like the one shown in the Index:

Captura de ecrã 2021-10-06, às 16 27 05

Another detail - the spacing between the post title and featured image should be as shown in the Post:

Captura de ecrã 2021-10-06, às 16 27 10

kjellr:
> The image width should be like the one shown in the Index:

I think the image width changing is fine, personally. Conceptually it gets wider when you're on the post page because you're taking a closer look at the post.

> Another detail - the spacing between the post title and featured image should be as shown in the Post:

Yeah, on the post pages, it seems like slightly too much space to me? The space below the titles should be visually equal to the spacing above the titles. The space above the titles is correct in both places.  

---

Here's how the change currently looks on the demo:

Screen Shot 2021-10-06 at 11 30 35

beafialho:
> I think the image width changing is fine, personally. Conceptually it gets wider when you're on the post page because you're taking a closer look at the post.

That makes sense, let's have the image width changing then.

> Yeah, on the post pages, it seems like slightly too much space to me? The space below the titles should be visually equal to the spacing above the titles. The space above the titles is correct in both places.

I see what you mean. I thought it looked a bit too tight, especially with all those "y's", so the spacing on the Post looked more comfortable to me.

scruffian:
I pushed a commit to remove the spacer so the image is always in the same place. @mikachan does this break anything you did recently?

mikachan:
Nope, this doesn't break anything I've done recently. 

I removed the same spacer in https://github.com/Automattic/themes/pull/4720, which is waiting for a fix in GB for the empty title. This wouldn't break either, but it might be what you were thinking about.

kjellr:
What did that last commit do? It appears mostly the same to me: 

index

scruffian:
That's strange, for me the images are in the same place:
Oct-08-2021 15-59-52

Is it possible that you have a saved template?

pbking:
My experience is the same as @scruffian 's, the images vertical position is the same in both scenarios.

pbking (approved):
Going to approve this one and plan to bring it in.  As best I can tell this achieves the design target.

The only outstanding comment is the "jumping image" which I suspect has a cause elsewhere.  We can work through that as a separate item if it persists.

PR #4799, opened 2021-10-07
Set Geologist version to 1.0.0

scruffian:
I think we should do this before launch


MaggieCabrera:
Package-lock.json needs the bump too. Doesn't Jason's script do this too?

PR #4803, opened 2021-10-07
Quadrat + Geologist: Remove invalid site-logo rule

scruffian:
This was added in https://github.com/Automattic/themes/pull/3776 originally, but it's in the wrong place in the file and doesn't seem to do anything, so I think we can just remove it.

To test, make sure that the header with a logo looks the same:
Screenshot 2021-10-07 at 16 22 30
Screenshot 2021-10-07 at 16 22 22

### DIFF
--- a/quadrat/assets/theme.css
+++ b/quadrat/assets/theme.css
@@ -502,6 +502,7 @@ input[type="submit"],
 input[type="datetime"],
 input[type="datetime-local"],
 input[type="color"],
+input[type="checkbox"],
 textarea {
 	font-size: var(--wp--preset--font-size--normal);
 }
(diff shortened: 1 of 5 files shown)


quadrat/assets/theme.css
```diff
@@ -502,6 +502,7 @@ input[type="submit"],
 input[type="datetime"],
 input[type="datetime-local"],
 input[type="color"],
+input[type="checkbox"],
```
scruffian, line 505: I guess this is a blockbase change that didn't get rebuilt?

mikachan (approved):
This works well for me in both themes 🚢

PR #4804, opened 2021-10-07
Blockbase: Rebuild all blockbase children

scruffian:
It seems like we've made changes to Blockbase and not always rebuilt the children when we have made those changes.

This PR rebuilds all the children and makes changes to some of the child themes so that the rebuilds don't change anything.


MaggieCabrera (approved):
LGTM!

PR #4806, opened 2021-10-08
Quadrat + Geologist: fix query loop not working

MaggieCabrera:

On Geologist and Quadrat we've encountered a bug when the gutenberg-edge sticker is applied to a site on dotcom. To reproduce this go to either of the demo sites and check the blog page while the sticker is active, you will see something like this:

Screenshot 2021-10-08 at 09 39 36

This is caused by a duplicated id in the query block apparently. Since that id is not needed by the markup to function, I've removed all instances of it on our themes. The problem this unveils is that now that the themes are activated, those that had this markup applied via headstart will not receive this fix. My test site wasn't fixed by this PR, while the demo site was:

Screenshot 2021-10-08 at 09 38 16

I'm going to fix the headstart file so that it doesn't happen in the future but this is a problem we'll encounter in the future with any changes that we make to the markup and it's going to be a tough one.

### DIFF
--- a/geologist/block-templates/index.html
+++ b/geologist/block-templates/index.html
@@ -1,6 +1,6 @@
 <!-- wp:template-part {"slug":"header"} /-->
 
-<!-- wp:query {"tagName":"main","layout":{"inherit":true},"queryId":1,"query":{"perPage":5,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":"","inherit":true}} -->
+<!-- wp:query {"tagName":"main","layout":{"inherit":true},"query":{"perPage":5,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":"","inherit":true}} -->
 <main class="wp-block-query">
 	<!-- wp:post-template -->
 		<!-- wp:group {"layout":{"inherit":true}} -->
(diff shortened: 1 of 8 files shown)


scruffian (approved):
LGTM

mtias:
Should the query block be more resilient here? Any ideas?

MaggieCabrera:
> Should the query block be more resilient here? Any ideas?

I don't know why we need the queryId, but at least an error message would be nice. The funny thing is, the query loop was showing up in the site editor, so I had little clue as to what was going on

mtias:
@ntsekouras any thoughts on this?

MaggieCabrera:
I was wrong about the headstarts, it doesn't contain the query loop (I thought we were pasting it in a page, but turns out we weren't). I'll check on other test sites to see if the fix will cover all user existing sites or not.

MaggieCabrera:
I'm bringing this in to prevent this from happening again for users

ntsekouras:
>I don't know why we need the queryId, but at least an error message would be nice. The funny thing is, the query loop was showing up in the site editor, so I had little clue as to what was going on

Sorry for the super late reply but this fell through the cracks.. 😞 .

`queryId` needs to be unique because it's tied to the `Query Pagination` functionality. This way we can properly support pagination in multiple queries in a single page.

Having said that, I don't think I understand what the error was and if is something still happening.  If you have multiple `Query Loop` blocks with the same `queryId`, their paginations wouldn't work properly but it shouldn't result in any error.

PR #4807, opened 2021-10-08
Try: Removing child-theme.json

scruffian:
This works with https://github.com/WordPress/gutenberg/pull/35459. With that PR checked out, it should no longer be necessary to build the theme.json, Gutenberg will do it for you.

The theme should behave the same as before.


MaggieCabrera:
not relevant for testing purposes but when this lands and we change all themes, we should add `"$schema": "https://json.schemastore.org/theme-v1.json"` to all of them too

pbking:
One thing I noticed tinkering with this is that our build process has a [special scenario for the custom templates](https://github.com/Automattic/themes/blob/trunk/blockbase/build.js#L38) (where they are merged instead of replaced).  The rest of the properties seem to be following standard procedure ok, but in order to keep the custom templates definitions consistent across this change it will need to define all of the templates expected to be available.

scruffian:
Closing in favour of https://github.com/Automattic/themes/pull/4861

PR #4808, opened 2021-10-08
Blockbase:  remove unneeded margin rules

MaggieCabrera:

This PR removes CSS that is no longer necessary after block gap was implemented. After gap was introduced, these rules were being overriden.

Check that the table and separator blocks look evenly separated from other blocks:

Screenshot 2021-10-08 at 17 15 45

Same is happening for Post template block:

Screenshot 2021-10-08 at 17 21 06

And with the columns block:

Screenshot 2021-10-08 at 17 23 59

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -902,10 +902,6 @@ p.wp-block-site-tagline {
 	text-align: var(--wp--custom--video--caption--text-align);
 }
 
-.wp-block-columns {
-	margin-bottom: unset;
-}
-
 .post-meta {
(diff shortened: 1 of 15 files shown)


MaggieCabrera:
Skatepark's separators need a look and probably custom CSS, since the original styles that this PR removes was being overridden, but I think we can make it in a separate PR or we'd have to ponyfill Blockbase more specifically while we wait for https://github.com/WordPress/gutenberg/pull/30609 or a similar alternative from Gutenberg.

scruffian (approved):
LGTM

PR #4818, opened 2021-10-12
Blockbase: Allow themes to inherit core gradients unless they define their own

scruffian:
By defining gradients as an empty array, we disable the default gradients. This removes that definition, allowing Blockbase themes to use the core gradients, unless they supply their own.


pbking (approved):
LGTM.
Seedlet-blocks has custom gradients.  The rest have the defaults.
👍

PR #4819, opened 2021-10-12
Blockbase: Add theme.json meta data

scruffian:
Once child theme's theme.json inherit from the parent theme.json, they will need a version and schema defining on them. There's no harm in adding this now ready for that change.

Hopefully this also makes it easier to edit these files as they will have better syntax highlighting.


pbking (approved):
👍 
I hadn't thought to add these to the child-theme.json files.  Thankee.

PR #4820, opened 2021-10-12
Videomaker: Fix post meta spacing

scruffian:
The spacing around post meta should be smaller.

Before:
Screenshot 2021-10-12 at 15 04 48

After:
Screenshot 2021-10-12 at 15 04 12

### DIFF
--- a/videomaker/child-theme.json
+++ b/videomaker/child-theme.json
@@ -101,6 +101,9 @@
 			],
 			"fontSizes": {
 				"tiny": "16px"
+			},
+			"gap": {
+				"baseline": "10px"
 			}
 		},
 		"layout": {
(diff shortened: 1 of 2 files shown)


beafialho:
This looks good 👍 

I'm not sure if these changes have been made, but just wanted to flag out that I see different spacings in different occasions:

https://user-images.githubusercontent.com/65220155/136987975-80ba1a96-7ce8-4060-b5ef-69186fd42769.mp4

pbking:
I verified that the spacing I observe in the two scenarios you illustrate @beafialho look the same for me.  There are no changes in this PR that would correct the behavior you have captured, however nothing that should break it.  Seems consistent to me and I think this is good-to-go.

pbking (approved):

PR #4822, opened 2021-10-12
Adjusted menu CSS to match recent changes in the block

### LINKED ISSUE #4821: Blockbase & Co: Submenu color is incorrect.
This seems to have been caused by changes to the structure and classes of the navigation block.

Populate menus with submenus and note their style on desktop:

The colors for the responsive version of the menu are correct, however the layout no longer matches the design target:

Current:

Design:

pbking:

Below are the mobile and desktop changes that I noticed were effected.  There may be other changes that I missed and didn't note below.  

Developed and tested on Gutenberg Trunk.  I'm unsure which commit (or version) the changes that broke the styles were introduced so I can't offer that for comparison. 

Quadrat Desktop Before:

Quadrat Desktop After:

Quadrat Mobile Before:

Quadrat Mobile After:

Geologist Desktop Before:

Geologist Desktop After:

Geologist Mobile Before:

Geologist Mobile After:

Blockbase Mobile Before:

Blockbase Mobile After:

Skatepark Desktop Before:

Skatepark Desktop After:

Seedlet Mobile Before:

Seedlet Mobile After:

Adjusts menu stylin' CSS to fix recent changes in the navigation block

#### Related issue(s):
Fixes #4821

### DIFF
--- a/blockbase/inc/block-styles.php
+++ b/blockbase/inc/block-styles.php
@@ -1,0 +1,31 @@
+<?php
+/**
+ * Blockbase Theme: Block Styles
+ *
+ * @package Blockbase 
+ * @since 1.2.8
+ */
+
+if ( ! function_exists( 'blockbase_register_block_styles' ) ) :
+
+	function blockbase_register_block_styles() {
+
+		if ( function_exists( 'register_block_style' ) ) {
+
+			/**
+			 * Register block styles
+			 */
+			register_block_style(
+				'core/navigation',
+				array(
+					'name'         => 'blockbase-navigation-clean',
+					'label'        => __( 'Clean', 'blockbase' ),
+					'style_handle' => 'blockbase-navigation-clean',
+				)
+			);
+
+		}
+	}
+endif;
+
+add_action( 'after_setup_theme', 'blockbase_register_block_styles' );
--- a/blockbase/inc/block-styles.php
+++ b/blockbase/inc/block-styles.php
@@ -18,9 +18,9 @@ if ( ! function_exists( 'blockbase_register_block_styles' ) ) :
 			register_block_style(
 				'core/navigation',
 				array(
-					'name'         => 'blockbase-navigation-clean',
-					'label'        => __( 'Clean', 'blockbase' ),
-					'style_handle' => 'blockbase-navigation-clean',
+					'name'         => 'blockbase-navigation-improved-responsive',
+					'label'        => __( 'Improved Responsive Navigation', 'blockbase' ),
+					'style_handle' => 'blockbase-navigation-improved-responsive',
 				)
 			);
 
--- a/blockbase/sass/blocks/_navigation.scss
+++ b/blockbase/sass/blocks/_navigation.scss
@@ -23,6 +23,8 @@
 		.wp-block-navigation__container {
 			row-gap: 0.5rem !important;
 			align-items: flex-start !important;
+			flex: unset;
+			padding-bottom: 0;
 		}
 		.wp-block-navigation-item {
 			align-items: flex-start !important;
--- a/blockbase/sass/blocks/_navigation.scss
+++ b/blockbase/sass/blocks/_navigation.scss
@@ -17,7 +17,7 @@
 	}
 }
 
-.wp-block-navigation.is-style-blockbase-navigation-clean {
+.wp-block-navigation.is-style-blockbase-navigation-improved-responsive {
 	&.is-responsive .is-menu-open {
 		font-size: var(--wp--preset--font-size--medium) !important;
 		.wp-block-navigation__container {
--- a/blockbase/sass/blocks/_navigation.scss
+++ b/blockbase/sass/blocks/_navigation.scss
@@ -1,19 +1,41 @@
 // See https://github.com/WordPress/gutenberg/issues/34648
-.wp-block-navigation.is-responsive {
-	.wp-block-navigation__responsive-container.is-menu-open {
+.wp-block-navigation {
+	// See https://github.com/WordPress/gutenberg/issues/34648
+	.has-child .wp-block-navigation__submenu-container {
 		background-color: var(--wp--custom--color--background);
-		color: var(--wp--custom--color--foreground);
+		border: var(--wp--custom--form--border--width) var(--wp--custom--form--border--style) var(--wp--preset--color--primary);
 	}
-	.wp-block-navigation-link__content {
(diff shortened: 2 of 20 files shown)


blockbase/sass/blocks/_navigation.scss
```diff
@@ -7,7 +7,7 @@
 	.wp-block-navigation-link__content {
 		color: var(--wp--custom--color--foreground) !important;
 	}
-	.has-child .wp-block-navigation-link__container{
-		display: revert;
+	.has-child .wp-block-navigation__submenu-container{
+		gap: 1rem;
```
scruffian: What does this do?

seedlet-blocks/sass/blocks/_navigation.scss
```diff
@@ -1,9 +1,19 @@
...
+		}
+		.wp-block-navigation__submenu-container {
+			padding-top: 0 !important;
+		}
 		.wp-block-navigation-link {
-			//NOTE: For reasons I cannot explain... if I set this to use the --wp--preset--font-family--headings
-			//these values (which are what are set) are not respected in the modal.
-			font-family: var(--wp--preset--font-family--headings);
+			//NOTE: We lost the ability to tap into "heading font family" when we "inverted" the mechanism of customizing fonts in universal themes.
+			//I'm unsure how to make this work as designed...
```
scruffian: I think this is a limitation of the approach. We have decided what blocks have heading styles in blockbase, and we can't implement that differently on a per theme basis. The upshot of which is that I think we should just remove this line.

scruffian:
~~I wasn't very keen on the position of the social icons in the mobile menu, I'm not sure where that came from. I've updated it to be as below, but this probably needs a designer's eye too @kjellr.~~ Edit: Sorry I just check the issue and saw the design 🤦

scruffian:
OK I made one small change to remove the margins on the social links block when its used inside a navigation block. This allows the social icons to have the correct spacing when they are at the bottom of the mobile menu like this:
Screenshot 2021-10-13 at 12 00 50
Screenshot 2021-10-13 at 12 00 29
Screenshot 2021-10-13 at 11 54 21
Screenshot 2021-10-13 at 11 50 40

The rest of the changes look good to me.

kjellr:
> <strike>I wasn't very keen on the position of the social icons in the mobile menu, I'm not sure where that came from. I've updated it to be as below, but this probably needs a designer's eye too @kjellr</strike>. Edit: Sorry I just check the issue and saw the design 🤦

We should generally default to a simple left-aligned layout like this for all of these themes except Quadrat: 

menu

That was the original intention for Blockbase (and all its children), but I think we dug into old comps and got it way more complicated than it needs to be.

blockbase/sass/blocks/_navigation.scss
pbking: It minimized the (very large) gap between sub-menu items.  Default is 2rem which is quite large.

seedlet-blocks/sass/blocks/_navigation.scss
pbking: I think we should remove the comment... and let the default menu selection be what's there.  If the user changes the fonts then it picks up the BODY font instead of the HEADER font but I think that's the only downside to just setting it to the default header font family in the CSS.

pbking:
> We should generally default to a simple left-aligned layout like this for all of these themes except Quadrat:

What blockbase is showing is very close to the Gutenberg default though (aside from minor spacing and color assignment) and I think that's why it's as it is right now. 

My 'druthers is to leave Blockbase as-is (keeping it more polyfill than parentstyle) though this definitely seems like another opportunity to [provide opinion](https://github.com/Automattic/themes/issues/4704).  Maybe blockbase could supply that menu style as a block style?

blockbase/sass/blocks/_navigation.scss
pbking: (see [comment](https://github.com/Automattic/themes/pull/4822#issuecomment-942335893))
Considering this is more opinion than function perhaps even this should be expressed as a block style?

pbking:
Keen to get some feedback on this direction:

The default navigation block styles for blockbase are pretty minimal.  They fix a couple of color problems (one fix for submenu colors on desktop and another for mobile colors) both of which should ultimately be fixed upstream, and a couple of tweaks for the social menu (which I think are theme-specific).  It's otherwise exactly how Gutenberg styles the responsive menu.




Additionally I have added a 'clean' (the name should be changed but that's just what came out when I named it) Block Style for the navigation block.  That style achieves the design that @kjellr posted above.



All of the child themes apply that style with the exception of Quadrat (and it's offspring Geologist) which have a much more customized mobile menu style (unchanged from the screenshots above).

Videomaker and Skatepark use it unaltered:




and Seedlet just changes font/typeface



It would be clean enough to move that block style CSS into the default blockbase navigation block styles, but it feels like a lot of "opinion" so wanted to see how y'all thought about handling that.

scruffian:
I love the idea of using a block style for this

kjellr:
Seems like a decent workaround, until we can control this in the block itself. Can we move the social icons up right below the menu items like there are [in that mockup](https://github.com/Automattic/themes/pull/4822#issuecomment-942243422)?

pbking:
Like this @kjellr ?

Seedlet-Blocks:


Skatepark:


Videomaker:


Mayland-Blocks:

kjellr:
Yep, looks good to me! Thanks

blockbase/inc/block-styles.php
```diff
@@ -0,0 +1,31 @@
...
+		if ( function_exists( 'register_block_style' ) ) {
+
+			/**
+			 * Register block styles
+			 */
+			register_block_style(
+				'core/navigation',
+				array(
+					'name'         => 'blockbase-navigation-clean',
+					'label'        => __( 'Clean', 'blockbase' ),
```
scruffian: Don't forget to change the name here.

pbking: You mean the name of the style?  `Clean` doesn't seem like a very good name.  Do y'all have a suggestion as to what it should be?    It's entirely changes to the responsive styling... maybe `Improved Responsive Navigation` or something?

scruffian: I like that

pbking: Done:

scruffian (approved):
LGTM

PR #4823, opened 2021-10-12
Added block-templates tag to blockbase themes

pbking:
It seems that our blockbase themes need the block-templates tag to play nice.  This change adds that tag to all of the themes.


pbking:
Oops.  Seems the tag is only needed on CLASSIC themes...

PR #4824, opened 2021-10-12
Blockbase: Fix form background color contrast

### LINKED ISSUE #4718: Blockbase: Transparent background on input elements
## Steps to replicate
1. Insert a group block and pick a background color.
2. Inside the group block add a contact form block.
3. Save & preview. Note that both in the editor and in preview the form input elements have a transparent background and depending on background color can be barely visible.

## Result
Preview:
image

Editor:
image

mikachan:
This fixes an issue on Blockbase & Co where form inputs can be difficult to see due to the color contrast, depending on the background color.

I went with @RavanH and @MaggieCabrera suggestion of setting `"background": "var(--wp--custom--color--background)"`, which seems to work well:

Before:
image

After:
image

#### Related issue(s):
Closes https://github.com/Automattic/themes/issues/4718


scruffian (approved):
My only hesitation with this is that if we force the background color, this makes it harder to customize the form. At the moment I can put the form on a different color background and the fields will change. However I think the risk of poor contract outweighs that consideration. It will be create when users can customize forms in in Global Styles.

RavanH:
Hence my alternative suggestion: setting `"text": "inherit"` and maybe also `"border": { "color": "currentColor",` or `var(--wp--custom--button--border--color)` which would allow either users or child themes to let the parent block text color (surely in contrast with the background color) determine the form text and field border color.

mikachan:
Thanks @RavanH, I've tried out your other suggestion now instead. I think this feels more flexible from a users' perspective, and if for some reason the parent block colours don't have good contrast with a custom background colour, it looks like these can all be overwritten per block. This worked for me with the search block.

| Before      | After |
| ----------- | ----------- |
| image        |  image      |

blockbase/theme.json
```diff
@@ -77,8 +77,8 @@
 					"width": "2px"
 				},
 				"color": {
-					"background": "var(--wp--custom--color--primary)",
-					"text": "var(--wp--custom--color--background)"
+					"background": "transparent",
+					"text": "inherit"
```
pbking: I don't believe buttons need to change.  This makes button background invisible by default:

blockbase/theme.json
```diff
@@ -151,7 +151,7 @@
 			"form": {
 				"padding": "calc( 0.5 * var(--wp--custom--gap--horizontal) )",
 				"border": {
-					"color": "#EFEFEF",
+					"color": "var(--wp--custom--color--primary)",
```
pbking: I'm not sure I'm sold on changing the border color of forms to primary.  My 'druthers is the non-color grey that's used already (at least as the Blockbase default).

blockbase/theme.json
```diff
@@ -184,7 +184,7 @@
 				"color": {
 					"background": "transparent",
 					"boxShadow": "none",
-					"text": "var(--wp--custom--color--foreground)"
+					"text": "inherit"
```
pbking, line 187: This bit fixes what ailes us.  I'm keen to bring this line in. Seems to address what's discussed in [the issue](https://github.com/Automattic/themes/issues/4718).

mikachan:
Thanks so much for testing this out, @pbking. I've left these changes as just the form color change, all your comments made sense to me.

blockbase/theme.json
```diff
@@ -78,7 +78,7 @@
 				},
 				"color": {
 					"background": "var(--wp--custom--color--primary)",
-					"text": "var(--wp--custom--color--background)"
+					"text": "inherit"
```
pbking: I just noticed that this changes the color of button labels.  I don't believe this change is necessary to fix this either.

pbking:
Bringing this in with the change to button removed.  @mikachan if I misunderstood the intent please LMK and we can fix it in another change.

PR #4832, opened 2021-10-13
Blockbase: Use new attributes for navigation block

scruffian:
When https://github.com/WordPress/gutenberg/pull/35568/ merges we will need to update the attributes of the navigation block from:

`"isResponsive": true`

to:

`"overlayMenu":"mobile"`.

Except in videomaker where it should be:
`"overlayMenu":"always"`.

#### Related issue(s):
https://github.com/WordPress/gutenberg/pull/35568/

### DIFF
--- a/blockbase/block-template-parts/header.html
+++ b/blockbase/block-template-parts/header.html
@@ -9,6 +9,6 @@
 	</div>
 	<!-- /wp:group -->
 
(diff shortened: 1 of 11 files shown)


mikachan (approved):
LGTM 🚀

pbking:
Looks like wpcom is ready for this change (Gutenberg has been updated enough to support it).  However there are a few conflicts to address first.

pbking:
Rebased and addressed conflicts.
Retested on wpcom; works as expected.
The CONTROLS are missing (and have been afaict) for the Navigation Block for a user to select these options.  But the attributes are working as expected so this seems ready to 🚢

PR #4835, opened 2021-10-14
Deploy script: Catch errors when adding new themes

scruffian:
When trying to deploy a new theme, the version bump script throws an uncaught error, halting the script. This catches the error and allows the script to proceed. There is no need to bump version numbers for new themes.


pbking (approved):
Looks like the real change happened [here](https://github.com/Automattic/themes/commit/513d7937b85119ec33699544705f20b989adc0d9) and this just cleans it up.
Looks good though, thanks for adding it.

PR #4838, opened 2021-10-14
Russell: Add a new Blockbase child called Russell

scruffian:
This creates a new child of Blockbase using a design from @alaczek.

Here's a screenshot but with TT2 patterns!
localhost_4759__p=559 (1)


russell/child-theme.json
```diff
@@ -0,0 +1,145 @@
...
+                    "slug": "huge"
+                }
+            ]
+        }
+    },
+    "styles": {
+        "blocks": {
+            "core\/button": {
+                "typography": {
+                    "fontFamily": "var(--wp--preset--font-family--dm-sans)"
```
scruffian, line 89: When the user edits their fonts using the customizer, these ones won't change.

russell/child-theme.json
```diff
@@ -0,0 +1,145 @@
...
+            },
+            "core\/post-title": {
+                "typography": {
+                    "fontFamily": "var(--wp--preset--font-family--libre-baskerville)",
+                    "fontSize": "var(--wp--preset--font-size--huge)"
+                }
+            },
+            "core\/pullquote": {
+                "typography": {
+                    "fontFamily": "var(--wp--preset--font-family--libre-baskerville)"
```
scruffian, line 100: When the user edits their fonts using the customizer, these ones won't change.

russell/child-theme.json
alaczek: Hmmm... I thought this is just setting the default. So you're saying if they changed DM Sans to say, Roboto in the Customizer, the button would still use DM Sans? If that's the case this can be removed (same goes for the pullquote block).

alaczek (approved):
Looks good to me! Thanks for generating the theme!

russell/child-theme.json
scruffian: Sorry I was wrong about this, pullquote is treated like a heading, so we can leave this in.

russell/child-theme.json
scruffian: Sorry I was wrong about this, button typography is treated like a heading, so we can leave this in.

Issue #4812, opened 2021-10-11
Geologist: .org submission issues


MaggieCabrera:
This is the .org feedback:

Keyboard Navigation

    Theme authors must provide visual keyboard focus highlighting in navigation menus and for form fields, submit buttons, and text links.
    Some resources to help you: Information on the requirement
    Video help for Modals, Links, Forms, Menus
    Check TwentyTwenty using tab and shift+tab as an example 

Mobile menu is not accessible using keyboard navigation.
Extras

    Tested up to: 5.7 test your theme with latest WP version. 

Prefixing

    Provide a unique prefix for everything the Theme defines in the public namespace, including options, functions, global variables, constants, post meta, wp_enqueue_script/style handle names, add_image_size names, wp_script_add_data keys, slugs/ids for new categories created with register_block_pattern_category etc. Theme nav menu locations and sidebar IDs are exceptions. 

Fix the issues and resubmit your theme again.

MaggieCabrera:
The trac ticket: https://themes.trac.wordpress.org/ticket/106674

scruffian:
Did these come up with Quadrat? Should we make the fixes there anyway?

MaggieCabrera:
> Did these come up with Quadrat? Should we make the fixes there anyway?

I don't think they came for Quadrat, no. And some, like the keyboard navigation, are not even directly related to the child, are they? I'm not sure what we should do here.

scruffian:
Fix them in Blockbase if we can I suppose.

pbking:
I went through Geologist (and blockbase) and found that I was able to navigation with the keyboard (both in desktop and responsive sizes) as well as I had expected.  I'm not sure what else is expected there.  There certainly don't seem to be any restrictions in Geologist that aren't in Blockbase, so I agree if something is found it should be fixed there.  I just don't know what that would be.  I didn't test deeply in all browsers (only chrome), but I have validated keyboard usage for Blockbase in Chrome Firefox and Edge prior.

pbking:
With the three (very minor) PR's noted above I think I have fixed everything addressable from the TRAC ticket.  With those changes I believe we can/should resubmit this.

jeffikus:
@pbking it looks like our TRAC ticket has been closed, would we need to resubmit a new ticket?

MaggieCabrera:
New trac ticket: https://themes.trac.wordpress.org/ticket/107158

MaggieCabrera:
[...]

PR #4839, opened 2021-10-14
Updated tested version of wordpress for all blockbase themes

pbking:
I'm not sure what process is needed to officially have the themes tested in a WP version but I've been using 5.8 locally to develop as long as it's been released so I feel comfortable saying it's a go...

Fixes part of #4812


scruffian (approved):
LGTM. We will also need to update the supported version when we build the zips.

PR #4840, opened 2021-10-14
renamed a couple of functions from 'newtheme' to 'geologist'

pbking:
Seems either the theme was initially generated with the name 'newtheme', the script didn't change those values, copypasta error or something else.

I'm not sure if this was the 'namespacing' problems noted in #4812 but it's the only instance I could find that didn't seem correct.

### DIFF
--- a/blockbase/create-child.js
+++ b/blockbase/create-child.js
@@ -58,7 +58,7 @@ const functions_php = `<?php
 /**
  * Add Editor Styles
  */
-function newtheme_editor_styles() {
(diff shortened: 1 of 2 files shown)


MaggieCabrera (approved):
yeah, this looks exactly like what's being reported, thank you!

scruffian:
I think this comes from the theme gen script. Can we update it to use the slug?

scruffian:
I pushed an update to the theme gen script.

PR #4841, opened 2021-10-14
Add copy of Quadrat's index.php to Geologist

pbking:
I'm not entirely sure what the index.php file's purpose would be for Geologist, nor if it should copy the sparse instance in Quadrat or the blank instance in Blockbase.

This should address the `REQUIRED: Could not find the file index.php in the theme.` comment in the [track](https://themes.trac.wordpress.org/ticket/106674) ticket noted in #4812


mikachan (approved):
This LGTM!

scruffian:
I think actually what we should do is copy this to Blockbase and remove it from all the children, since they will inherit it.

scruffian:
I pushed a commit to move the index.php to Blockbase, so now all the children will inherit it. I'm not sure if the theme check will like this though...

MaggieCabrera:
> I pushed a commit to move the index.php to Blockbase, so now all the children will inherit it. I'm not sure if the theme check will like this though...

Yeah I'm not so sure myself. I think we only added this to pass that check. I just tested the check against this branch and it passes, but I never saw that warning on the first place myself so I don't know... I'm running Version 20210921 of the theme check plugin

Screenshot 2021-10-15 at 09 55 18

scruffian:
Maybe we should just give it a go?

pbking:
I agree with the change, it's clear that this at least is the "blockbase index" way.  Also think giving it a go with this change is good too, though I agree it might not pass inspection but then at least we'll know.

PR #4845, opened 2021-10-14
Removed paragraph margin gap styles.  They no longer seem to be needed.

### LINKED ISSUE #4842: Blockbase: Broken margins for buttons in a group.
## Steps to replicate
When buttons are in a group block their alignment is off:
Screenshot 2021-10-14 at 18 49 59

This is probably caused by this line:
```
.wp-block-group :first-child {
	 margin-top: 0;
}
```

This might not be needed since we added block gap fixes to Gutenberg.

pbking:
Before:

Editor:

View:

After:

Editor:

View:

Brought to light via #4842


scruffian:
I think we can remove the margin-top for :first-child

However, it looks like the paragraph margins are still needed, see https://github.com/Automattic/themes/pull/4738. We need the gaps between paragraph blocks to be 1em, rather than whatever block gap is.

scruffian (approved):
LGTM

pbking:
As it exists now the gap between paragraphs in a group also includes a bottom gap which is not the case by default.


If using the same block gap isn't desirable then I think we may still need to tweak the styles to keep it more correct in that scenario.

scruffian:
Yeah good catch

PR #4847, opened 2021-10-15
Quadrat: removed the utility class from block patterns and instead used margin

MaggieCabrera:

This PR removes the utility class `mt-0` from Quadrat's patterns (I hesitate to remove it from Blockbase's CSS in case other users are using it on their themes) and I am using GB margin instead. I was under the impression that only Group blocks supported margins but it appears that is [no longer the case](https://github.com/WordPress/gutenberg/issues/28356#issuecomment-943929002). Hopefully, this can help solve the issues over at https://github.com/Automattic/wp-calypso/issues/56833

To test, check that there's no gap between the darker block and the images on the Latest episodes and Query episodes patterns.

Screenshot 2021-10-15 at 09 46 36


scruffian (approved):
LGTM

PR #4849, opened 2021-10-15
Russell: Update the navigation to match Blockbase

scruffian:
This takes the navigation from Blockbase to Russell, bringing in Social Links, right alignment, and a smaller font for the tag line.


pbking:
Should we apply the navigation block style?

Before:


After:

pbking:
The other bits look good. If you think the block style is 👍 I think it's GTG.

scruffian:
Oh yeah we should, nice one.

alaczek:
Will you be pushing this change to .com

scruffian:
Yes, this will get pushed when we next deploy

PR #4857, opened 2021-10-18
Blockbase + children: cleanup errors on theme.json files and rebuild themes

MaggieCabrera:

This PR removes errors on the theme.json files of Blockbase and all its children. All themes should look the same after this change. There's a couple of errors showing up that are not errors at all (such as the `link` elements inside some core blocks), and need to be reported over at the SchemaStore/schemastore repo. I'll get to it as soon as possible.


scruffian (approved):
Thanks for doing this.

PR #4861, opened 2021-10-18
Remove/child theme

pbking:
Created as an alternative to #4807

This change:

Copies the Custom Templates portion to any themes that don't have them.  (Our build script made an exception for that collection that Gutenberg doesn't.

Removes the JSON merging step from Blockbase and all children.

Modifies the `rebuild-child-themes.sh` tool to copy `child-theme.json` to `theme.json` for all child themes.

This modification is done in a script so that this change can sit at-the-ready until all environments are ready to accept this change and won't (or at least, isn't likely to) conflict with any `child-theme.json` made prior to that readyness.  To execute run `./rebuild-child-themes.sh` (then delete that file) and then commit THOSE changes in the change as well.  (But only when it's time to ship!)


zoologist/child-theme.json
```diff
@@ -1,5 +1,21 @@
 {
 	"customTemplates": [
+		{
```
scruffian: I feel like this should also be part of the child/parent inheritance in Gutenberg

scruffian (approved):
LGTM. I think it would be interesting to explore moving our SASS to CSS too, to totally remove the need for a build

MaggieCabrera:
> LGTM. I think it would be interesting to explore moving our SASS to CSS too, to totally remove the need for a build

we would need to move things like the extra button styles to a separate stylesheet then

PR #4862, opened 2021-10-18
Blockbase: Add a way to create a theme

scruffian:
This is a proposal for a mechanism to create Blockbase child themes.

The way it works is like this:
1. Switch to a Blockbase theme that you want to clone
2. Make any changes you want to see to the block templates or theme.json settings. You can do this using Global Styles or using the customizer
3. Go to the "Create theme" in the menu, and fill in the details:
Screenshot 2021-10-18 at 20 47 41
4. When you click "Create theme" a zip will be generated containing all your theme files.
5. You can install this like any other theme

This works without https://github.com/Automattic/themes/pull/4807 but it assumes we are taking that approach for Blockbase children, so that PR will need to be merged before this one is usable.

The logic is built on top of a REST API endpoint in core. I think eventually this should be part of core.

### DIFF
--- a/blockbase/inc/create-theme.php
+++ b/blockbase/inc/create-theme.php
@@ -1,0 +1,228 @@
+<?php
+/**
+ * REST endpoint for exporting the contents of the Edit Site Page editor.
+ *
+ * @package gutenberg
+ */
+
+function gutenberg_edit_site_get_theme_json_for_export() {
+	$child_theme_json = json_decode( file_get_contents( get_stylesheet_directory() . '/theme.json' ), true );
+	$child_theme_json_class_instance = new WP_Theme_JSON_Gutenberg( $child_theme_json );
+	$user_theme_json = WP_Theme_JSON_Resolver_Gutenberg::get_user_data();
+	// Merge the user theme.json into the child theme.json.
+	$child_theme_json_class_instance->merge( $user_theme_json );
+
+	// I feel like there should be a function to do this in Gutenberg but I couldn't find it
+	function remove_theme_key( $data ) {
+		if ( is_array( $data ) ) {
+			if ( array_key_exists( 'theme', $data ) ) {
+				if ( array_key_exists( 'user', $data ) ) {
+					return $data['user'];
+				}
+
+				return $data['theme'];
+			}
+			foreach( $data as $node_name => $node_value  ) {
+				$data[ $node_name ] = remove_theme_key( $node_value );
+			}
+		}
+
+		return $data;
+	}
+
+	return remove_theme_key( $child_theme_json_class_instance->get_raw_data() );
+}
+
+function blockbase_get_style_css( $theme ) {
+	$slug = $theme['slug'];
+	$name = $theme['name'];
+	$description = $theme['description'];
+	$uri = $theme['uri'];
+	$author = $theme['author'];
+	$author_uri = $theme['author_uri'];
+
+	return "/*
+Theme Name: {$name}
+Theme URI: {$uri}
+Author: {$author}
+Author URI: {$author_uri}
+Description: {$description}
+Requires at least: 5.8
+Tested up to: 5.8
+Requires PHP: 5.7
+Version: 0.0.1
+License: GNU General Public License v2 or later
+License URI: https://raw.githubusercontent.com/Automattic/themes/trunk/LICENSE
+Template: blockbase
+Text Domain: {$slug}
+Tags:
+*/";
+}
+
+function blockbase_get_functions_php( $theme ) {
+	$slug = $theme['slug'];
+	return "<?php
+/**
+ * Add Editor Styles
+ */
+function {$slug}_editor_styles() {
+	// Enqueue editor styles.
+	add_editor_style(
+		array(
+			'/assets/theme.css',
+		)
+	);
+}
+add_action( 'after_setup_theme', '{$slug}_editor_styles' );
+
+/**
+ *
+ * Enqueue scripts and styles.
+ */
+function {$slug}_scripts() {
+	wp_enqueue_style( '{$slug}-styles', get_stylesheet_directory_uri() . '/assets/theme.css', array( 'blockbase-ponyfill' ), wp_get_theme()->get( 'Version' ) );
+}
+add_action( 'wp_enqueue_scripts', '{$slug}_scripts' );";
+}
+
+function blockbase_get_theme_css( $theme ) {
+	return file_get_contents( get_stylesheet_directory() . '/assets/theme.css' );
+}
+
+/**
+ * Creates an export of the current templates and
+ * template parts from the site editor at the
+ * specified path in a ZIP file.
+ *
+ * @param string $filename path of the ZIP file.
+ */
+function gutenberg_edit_site_export_theme_create_zip( $filename, $theme ) {
+	if ( ! class_exists( 'ZipArchive' ) ) {
+		return new WP_Error( 'Zip Export not supported.' );
+	}
+
+	$zip = new ZipArchive();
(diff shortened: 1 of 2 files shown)


blockbase/inc/create-theme.php
```diff
@@ -0,0 +1,228 @@
+<?php
+/**
+ * REST endpoint for exporting the contents of the Edit Site Page editor.
+ *
+ * @package gutenberg
+ */
+
+function gutenberg_edit_site_get_theme_json_for_export() {
+	$child_theme_json = json_decode( file_get_contents( get_stylesheet_directory() . '/theme.json' ), true );
```
scruffian, line 9: In Blockbase this is currently a generated file, but once https://github.com/Automattic/themes/pull/4807 is merged this will make more sense.

blockbase/inc/create-theme.php
```diff
@@ -0,0 +1,228 @@
...
+ * Enqueue scripts and styles.
+ */
+function {$slug}_scripts() {
+	wp_enqueue_style( '{$slug}-styles', get_stylesheet_directory_uri() . '/assets/theme.css', array( 'blockbase-ponyfill' ), wp_get_theme()->get( 'Version' ) );
+}
+add_action( 'wp_enqueue_scripts', '{$slug}_scripts' );";
+}
+
+function blockbase_get_theme_css( $theme ) {
+	return file_get_contents( get_stylesheet_directory() . '/assets/theme.css' );
```
scruffian, line 89: Ideally this should output all enqueued CSS for the theme.

blockbase/inc/create-theme.php
```diff
@@ -0,0 +1,228 @@
...
+
+add_action( 'admin_menu', 'blockbase_create_theme_menu' );
+
+function blockbase_save_theme() {
+	// I can't work out how to call the API but this works for now.
+	if ( ! empty( $_GET['page'] ) && $_GET['page'] === 'create-theme' && ! empty( $_GET['theme'] ) ) {
+		gutenberg_edit_site_export_theme( $_GET['theme'] );
+	}
+}
+add_action( 'init', 'blockbase_save_theme');
```
scruffian, line 228: This is way too heavy handed, but it works for now.

pbking:
This is brilliant!  Per your comment this pairs really nice with the [removal of `child-theme.json`](https://github.com/Automattic/themes/pull/4861) and makes a nice, compact theme.json.

I haven't gone through the resulting theme carefully yet but as to the concept I believe it's a great looking tool.

alaczek:
This is awesome!

Would it be possible to create a child theme not from an existing child (Geologist > Zoologist), but directly from Blockbase? I think that would be the most useful in the context of the work I'm doing now.

Also, I expected was to get child-theme.json starter file and it wasn't there in the generated theme.

Lastly, I thought the generated theme would include package.json?

pbking:
child-theme.json is ABOUT to go away, and that whole part of the build step too.  Since there's also no scss in the generated theme there's also no css build step for these themes so the result is pretty simple.

I just tested building straight from Blockbase and it looks like it will need a few tweaks to do so efficiently (It creates a working theme, but with more stuff than needed).  But no reason that wouldn't also work.

MaggieCabrera:
I love this! in fact, I like it so much that I think it should be the theme generator we were thinking about for https://blockbasetheme.com/

scruffian:
> Would it be possible to create a child theme not from an existing child (Geologist > Zoologist), but directly from Blockbase? I think that would be the most useful in the context of the work I'm doing now.

Yes that should work fine.

> Also, I expected was to get child-theme.json starter file and it wasn't there in the generated theme.

This is because we are getting rid of child-theme.json (see https://github.com/Automattic/themes/pull/4807)

> Lastly, I thought the generated theme would include package.json?

Again, because we don't need the build step anymore we shouldn't need a package.json. I'm curious what @MaggieCabrera and @pbking think about this, since we might still want to build SASS in these child themes...

MaggieCabrera:
> Again, because we don't need the build step anymore we shouldn't need a package.json. I'm curious what @MaggieCabrera and @pbking think about this, since we might still want to build SASS in these child themes...

I think it depends! Generally, I don't want to get rid of SASS for Blockbase at all for our own benefit. Thinks like the Buttons stylesheet using mixins and stuff would be very ugly to do using only CSS. But I agree that if the theme is based on Blockbase and it's only adding some theme.json changes, there's no point to do any kind of build process, and the extra CSS can be added without SASS. If you are basing on Quadrat, however, you are loading the extra import for the outlined buttons for example. That requires SASS unless we are refactoring Blockbase

pbking:
>Again, because we don't need the build step anymore we shouldn't need a package.json. I'm curious what @MaggieCabrera and @pbking think about this, since we might still want to build SASS in these child themes...

Don't build it 'till ya' need it.

If a generated theme advances such that it needs a sass build step then that's a decision for the future.

Ideally a theme based on Blockbase wouldn't HAVE any css either which is indeed the best scenario; when changes come down the pipe the child theme gets them without need of alteration whereas if it were build from something like Quadrat and pulls that theme's CSS in there may be reason for that CSS to change in the future based on changes to Gutenberg.  Ideally most of that kind of stuff happens in Blockbase, that's the idea.  

Definitely agree with Maggie though that we can't lose SCSS for Blockbase itself.  

> If you are basing on Quadrat, however, you are loading the extra import for the outlined buttons for example. That requires SASS unless we are refactoring Blockbase

I do think that if it's based on Quadrat then the Quadrat CSS is needed (for the types of reasons you state @MaggieCabrera ) , but I don't think it needs to have the SASS source come along for the ride.  Though... if it were a child of QUADRAT instead then there's no need to include that CSS in the bundle either... Grandchildren Block Themes?  🤔

alaczek:
No build process sounds great to me! :) 

> This is because we are getting rid of child-theme.json

Oh ok, I wasn't aware of that. What does that mean to themes I'm currently developing from Blockbase - do I keep using child-theme.json?

scruffian:
> Oh ok, I wasn't aware of that. What does that mean to themes I'm currently developing from Blockbase - do I keep using child-theme.json?

Yeah for now that's fine. We'll migrate all the themes over to the new process when the time comes, just make sure its in the list of child themes in rebuild-child-themes.js

jeffikus:
@scruffian was thinking about this "The logic is built on top of a REST API endpoint in core. I think eventually this should be part of core." - should we place it under the "Tools" menu item perhaps then? Or "Appearance"?

MaggieCabrera:
I'm thinking we may want to hide this for dotcom users if we ever merge this, and this kind of thing is probably not allowed on .org either as part of the theme

blockbase/inc/create-theme.php
```diff
@@ -0,0 +1,228 @@
...
+Description: {$description}
+Requires at least: 5.8
+Tested up to: 5.8
+Requires PHP: 5.7
+Version: 0.0.1
+License: GNU General Public License v2 or later
+License URI: https://raw.githubusercontent.com/Automattic/themes/trunk/LICENSE
+Template: blockbase
+Text Domain: {$slug}
+Tags:
```
MaggieCabrera, line 58: can we pre populate this with Blockbase's tags? We never really change them for child themes. At most you would had a "photography", "blogging", etc tag that would talk about the content, but most children will have the same functionalities as the parent

scruffian: one issue is that our tags contain wpcom specific tags, which would be annoying for dotorg users.

MaggieCabrera: well, we can leave those for now and add them manually, but I think adding the blockbase ones is better than leaving it empty

scruffian:
I moved this to a plugin: https://github.com/Automattic/create-blockbase-theme

PR #4863, opened 2021-10-19
Payton: Initial commit.

alaczek:
Add a new Blockbase child called Payton.

### DIFF
--- a/payton/functions.php
+++ b/payton/functions.php
@@ -1,0 +1,24 @@
+<?php
+
+/**
+ * Add Editor Styles
+ */
+function newtheme_editor_styles() {
+	// Enqueue editor styles.
+	add_editor_style(
+		array(
+			'/assets/theme.css',
+		)
+	);
+}
+add_action( 'after_setup_theme', 'newtheme_editor_styles' );
+
+/**
+ *
+ * Enqueue scripts and styles.
+ */
+function newtheme_scripts() {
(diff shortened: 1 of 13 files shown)


payton/functions.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+/**
+ * Add Editor Styles
+ */
+function newtheme_editor_styles() {
```
scruffian: This should say payton, not newtheme.

payton/functions.php
```diff
@@ -0,0 +1,24 @@
...
+	// Enqueue editor styles.
+	add_editor_style(
+		array(
+			'/assets/theme.css',
+		)
+	);
+}
+add_action( 'after_setup_theme', 'newtheme_editor_styles' );
```
scruffian: This should say payton, not newtheme.

payton/functions.php
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ *
+ * Enqueue scripts and styles.
+ */
+function newtheme_scripts() {
```
scruffian: This should say payton, not newtheme.

scruffian (approved):
Just a few instances of newtheme->payton and then we can merge this

payton/functions.php
alaczek: Good catch! I think this was generated by the script and I didn't notice - should the script be inserting theme name here?

scruffian: yeah, I thought i'd fixed it but looks like it's not working, or you're using an older version

alaczek: That's entirely possible :) I'll keep an eye out for it next time I'm generating a theme.

PR #4865, opened 2021-10-19
Blockbase: search block font size

MaggieCabrera:

This PR https://github.com/WordPress/gutenberg/pull/35723 is about to be merged and it will affect all Blockbase themes. Previously, applying font-size to the search block via theme.json didn't work, after the GB PR is merged it will start working. Before the PR, the search block on all BB themes was simply applying the browser default rules for inputs (about 13.333px on FF). I decided on using --wp--preset--font-size--small instead, it will make the block look a little bit bigger don't he themes but I think it makes it more consistent with the rest of the elements of the theme than the random browser size.

I'm also changing the line-height so that the difference in size is not as big.

Skatepark 

Before | After
--- | ---
Screenshot 2021-10-19 at 13 01 22 | Screenshot 2021-10-19 at 13 06 08

After the related PR is merged, the font size of the block will inherit directly from the body font, which will be much bigger than this.

@kjellr @beafialho @melchoyce I'm tagging you since this will introduce a slight visual change.

#### Related issue(s):

https://github.com/WordPress/gutenberg/pull/35723

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -782,7 +782,6 @@ div.wp-block-query-pagination .wp-block-query-pagination-numbers .current {
 }
 
 .wp-block-search .wp-block-search__input {
-	line-height: var(--wp--custom--button--typography--line-height);
 	padding: var(--wp--custom--form--padding);
 	border: var(--wp--custom--form--border--width) var(--wp--custom--form--border--style) var(--wp--custom--form--border--color);
 }
--- a/blockbase/sass/blocks/_search.scss
+++ b/blockbase/sass/blocks/_search.scss
@@ -14,7 +14,6 @@
 	}
 
 	.wp-block-search__input {
-		line-height: var(--wp--custom--button--typography--line-height);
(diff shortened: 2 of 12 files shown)


pbking:
Change looks good.  I'll leave the design 👍 to others but if it looks good it's good to bring in.

Also I see no negatives to bringing this in before the change lands in Gutenberg besides the lineheight difference.

kjellr:
What is the font size that's used for Buttons by default in Blockbase? I think it might make sense to match that.

MaggieCabrera:
> What is the font size that's used for Buttons by default in Blockbase? I think it might make sense to match that.

ok, then it's even bigger than the original font-size and my change, I'll make the change if you are fine with that:

Before:

Screenshot 2021-10-20 at 10 05 31
After:

Screenshot 2021-10-20 at 10 05 19

kjellr:
Using the same size is great. Thanks, @MaggieCabrera!

pbking (approved):
Now with design approval this change LGTM.

PR #4868, opened 2021-10-19
Videomaker: rename functions

scruffian:
These functions need namespacing to the theme.


mikachan (commented):
It looks like this includes the tag style changes as well, but other than that the function namespacing looks good!

PR #4869, opened 2021-10-19
Videomaker: Add dynamic template references

### LINKED ISSUE #4858: Videomaker: Post Meta template part
## Steps to replicate
On WordPress.com, the block patterns have an error because they try to use the template part block, which isn't supported in WordPress.com yet. We'll need to find a different solution for now.

scruffian:
The theme slug needs to be dynamic in the block-template block, as it depends where the theme is installed.

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/4858

### DIFF
--- a/videomaker/assets/theme.css
+++ b/videomaker/assets/theme.css
@@ -64,11 +64,6 @@ div.post-meta a {
 	align-items: flex-end;
 }
 
-.wp-block-navigation.is-responsive .has-child .wp-block-navigation__submenu-container {
-	background-color: var(--wp--custom--color--background);
-	border: var(--wp--custom--form--border--width) var(--wp--custom--form--border--style) var(--wp--preset--color--primary);
-}
-
 footer > .wp-block-group {
 	align-items: center;
 	justify-content: space-between;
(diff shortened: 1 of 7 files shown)


pbking:
Looks like it's writing the markeup:
```
<!-- wp:template-part {"slug":"post-meta","theme":"pub/blockbase"} /-->
```

which doesn't seem to work.  Manually changing the block markup to this does:

```
<!-- wp:template-part {"slug":"post-meta","theme":"pub/videomaker"} /-->
```

I think the concept will work, but should be writing the current them not parent theme?

MaggieCabrera:
you'd want [get_stylesheet_directory()](https://developer.wordpress.org/reference/functions/get_stylesheet_directory/) maybe? it doesn't work without the theme attribute??

scruffian:
Oh yeah sorry, it's actually `get_stylesheet`. Updated.

pbking:
> it doesn't work without the theme attribute??

From what I understand (and can tell tinkering with this branch) it works without the theme attribute in a block template (i.e. the index.html block template) but not in the content.

pbking (approved):
Creates the pattern w/ reference to block template part as expected:
```
<!-- wp:template-part {"slug":"post-meta","theme":"pub/videomaker"} /-->
```

Tested locally and on wpcom.

🚢

PR #4870, opened 2021-10-19
Videomaker: Update the single post template

scruffian:
This makes some changes to the single post template to match figma.

@beafialho do you think the next and previous buttons should go above or below the comment form?

Figma:
Screenshot 2021-10-19 at 15 20 14

This PR:
Screenshot 2021-10-19 at 14 38 01

### DIFF
--- a/videomaker/block-templates/index.html
+++ b/videomaker/block-templates/index.html
@@ -23,8 +23,6 @@
 	<!-- wp:query-pagination -->
 	<div class="wp-block-query-pagination"><!-- wp:query-pagination-previous /-->
 
-		<!-- wp:query-pagination-numbers /-->
-
 		<!-- wp:query-pagination-next /--></div>
 	<!-- /wp:query-pagination -->
 	</div>
--- a/videomaker/sass/theme.scss
+++ b/videomaker/sass/theme.scss
@@ -4,4 +4,5 @@
 @import "navigation";
 @import "templates/footer";
 @import "blocks/post-terms";
+@import "blocks/post-navigation-link";
 @import "blocks/query-pagination";
--- a/videomaker/sass/theme.scss
+++ b/videomaker/sass/theme.scss
@@ -4,3 +4,4 @@
 @import "navigation";
 @import "templates/footer";
 @import "blocks/post-terms";
+@import "blocks/query-pagination";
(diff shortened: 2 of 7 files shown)


videomaker/block-templates/index.html
```diff
@@ -25,8 +25,6 @@
 	<!-- wp:query-pagination -->
 	<div class="wp-block-query-pagination"><!-- wp:query-pagination-previous /-->
```
scruffian, line 24: To test this on trunk you'll need to remove this div and the one on line 30.

beafialho:
I think they should go above the comment form:

Captura de ecrã 2021-10-19, às 16 36 01

scruffian:
I updated this to use the next and previous page blocks:
Screenshot 2021-10-19 at 18 03 13

videomaker/sass/theme.scss
```diff
@@ -2,4 +2,6 @@
 @import "links";
 @import "meta";
 @import "navigation";
-@import "templates/footer";
\ No newline at end of file
+@import "templates/footer";
+@import "blocks/post-navigation-link";
```
mikachan, line 7: I can't see the `blocks/post-navigation-link` SASS file when I pull this branch (my build is failing because of it).

mikachan (commented):
On the single post template, the prev/next links look good when they both exist:

image

But they look wonky when one is missing:

image

image

I'm guessing I might be missing some CSS?

videomaker/sass/theme.scss
MaggieCabrera: there's two things we can do: 

- use `:empty` to check if we want the border to show or not (https://developer.mozilla.org/en-US/docs/Web/CSS/:empty)
- use the columns block so that they stay in the correct positions too

jeffikus: Nice catch @mikachan - I see the same, once we've tested with that file we could do a CSS fix by moving the styles to the anchor tag, or use empty like @MaggieCabrera suggested as well.

mikachan:
Those missing styles have fixed the misalignment when either prev/next is missing, although they look like this for me:

image

Is this right? We could make the links take up the full width with `display: block`.

jeffikus:
@mikachan I agree with adding the `display: block` - I get the same on my dev environment.

scruffian:
I added display: block, I think this can be merged.

PR #4874, opened 2021-10-19
Update style.css

jeffikus:

Adds auto-loading-homepage tag to stylesheet. Fixes homepage switching on WordPress.com

Themes:
- Blockbase
- Geologist
- Mayland Blocks
- Payton
- Quadrat
- Russell
- Seedlet Blocks
- Skatepark
- Videomaker
- Zoologist

#### Related issue(s):

None.

### DIFF
--- a/payton/style.css
+++ b/payton/style.css
@@ -12,6 +12,6 @@ License: GNU General Public License v2 or later
 License URI:
 Template: blockbase
 Text Domain: payton
-Tags:
+Tags:auto-loading-homepage
 
 */
--- a/payton/style.css
+++ b/payton/style.css
@@ -12,6 +12,6 @@ License: GNU General Public License v2 or later
 License URI:
 Template: blockbase
 Text Domain: payton
-Tags:auto-loading-homepage
+Tags: editor-style, auto-loading-homepage
 
 */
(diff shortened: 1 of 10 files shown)


payton/style.css
```diff
@@ -12,6 +12,6 @@ License: GNU General Public License v2 or later
 License URI:
 Template: blockbase
 Text Domain: payton
-Tags:
+Tags:auto-loading-homepage
```
pbking: The .zip building script is specifically looking for `, auto-loading-homepage` thus if it is the first (or only) item it will not be removed.  Need to either add others so it's not the last or update (and complicate) the script.

videomaker/style.css
```diff
@@ -12,6 +12,6 @@ License: GNU General Public License v2 or later
 License URI: 
 Template: blockbase
 Text Domain: videomaker
-Tags: 
+Tags: auto-loading-homepage
```
pbking: same problem as payton, needs `, auto-loading-homepage` to be removed.

payton/style.css
jeffikus: Added editor-styles tag.

videomaker/style.css
jeffikus: Added editor-styles tag.

pbking (approved):
Does the job.
.org script will strip it out finely
🚢

PR #4876, opened 2021-10-20
Videomaker: Navigation styles

### LINKED ISSUE #4651: videomaker: Templates: Header template part (including mobile)
The videomaker nav should look like this:
Screenshot 2021-10-14 at 19 16 30
Screenshot 2021-10-19 at 15 24 10

Also the navigation in the header shouldn't be UPPERCASE, only in the patterns.

mikachan:
This PR tweaks the Videomaker navigation styles to match the Figma designs.

| Before      | After |
| ----------- | ----------- |
| image     | image     |

It looks like GB doesn't pull through the `current-menu-item` class, so we're not able to style the active menu item differently at the moment. I've made a change to the navigation attributes PR on GB that fixes this, so it's possible to test on this branch: https://github.com/WordPress/gutenberg/pull/35634. 

Without the GB PR, I think we'll need to remove the submenu styles. At the moment I've hidden all child menu items unless the parent is the current menu item, which means they'll be hidden by default without the `current-menu-item` class. Let me know if you want me to remove these now - or maybe comment them out while we wait for a fix?

I've made sure the menus in the patterns are still uppercase by adding an inline style to the nav blocks in the patterns.

I've also only styled the responsive menu, as it looks like Videomaker should always use the responsive menu when this can be merged: https://github.com/Automattic/themes/pull/4832

#### Related issue(s):
Closes https://github.com/Automattic/themes/issues/4651

### DIFF
--- a/blockbase/block-template-parts/header.html
+++ b/blockbase/block-template-parts/header.html
@@ -3,6 +3,6 @@
 	<!-- wp:site-logo /-->
 	<!-- wp:site-title /-->
 	<!-- wp:site-tagline {"style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
-	<!-- wp:navigation {"itemsJustification":"right","isResponsive":true,"__unstableLocation":"primary","__unstableSocialLinks":"social"} /-->
+	<!-- wp:navigation {"className":"is-style-blockbase-navigation-improved-responsive","itemsJustification":"right","isResponsive":true,"__unstableLocation":"primary","__unstableSocialLinks":"social"} /-->
 </div>
 <!-- /wp:group -->
--- a/videomaker/sass/_navigation.scss
+++ b/videomaker/sass/_navigation.scss
@@ -16,6 +16,12 @@
 			&.current-menu-item > a {
 				text-decoration: underline;
 			}
+			&.has-child {
+				.wp-block-navigation__submenu-container {
+					padding-left: 0 !important;
+					padding-right: 0;
+				}
+			}
 			&.has-child:not(.current-menu-item) {
 				.wp-block-navigation__submenu-container {
 					display: none;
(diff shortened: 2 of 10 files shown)


videomaker/sass/_navigation.scss
```diff
@@ -1,4 +1,39 @@
 // Needed until this is merged: https://github.com/WordPress/gutenberg/issues/35040
 .is-vertical.items-justified-right ul.wp-block-navigation__container {
 	align-items: flex-end;
+}
+
+.wp-block-navigation.is-style-blockbase-navigation-improved-responsive {
+	&.is-responsive .is-menu-open {
+		.wp-block-navigation__container {
+			align-items: center !important;
```
scruffian: We need the `!important` to overcome some rules in blockbase, but I'm not sure that the blockbase rules need `!important` either. We should look into removing them.

blockbase/block-template-parts/header.html
```diff
@@ -3,6 +3,6 @@
 	<!-- wp:site-logo /-->
 	<!-- wp:site-title /-->
 	<!-- wp:site-tagline {"style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
-	<!-- wp:navigation {"itemsJustification":"right","isResponsive":true,"__unstableLocation":"primary","__unstableSocialLinks":"social"} /-->
+	<!-- wp:navigation {"className":"is-style-blockbase-navigation-improved-responsive","itemsJustification":"right","isResponsive":true,"__unstableLocation":"primary","__unstableSocialLinks":"social"} /-->
```
scruffian, line 6: I added this for testing, but I do think it looks better too. What do you think @kjellr ?

videomaker/sass/_navigation.scss
scruffian: Done

blockbase/block-template-parts/header.html
kjellr: Yeah, I think this can use the standard block style. 👍

PR #4877, opened 2021-10-20
dotorg package script: remove tags for themes that don't build their css

MaggieCabrera:

The script removes the tags only on sass files. This worked fine for Varia based themes, but Blockbase children don't have the tags on the sass files, just the CSS file

This change removes the tags from the CSS files as well


scruffian (approved):
LGTM

PR #4878, opened 2021-10-20
Blockbase: Add query title

scruffian:
This adds the query title to all Blockbase themes, with the exception of Skatepark which already uses it. This is useful for displaying the context on Archive pages.

Fixes https://github.com/Automattic/themes/pull/4878

### Screenshots:

Screenshot 2021-10-20 at 14 44 17
Screenshot 2021-10-20 at 14 42 23
Screenshot 2021-10-20 at 14 42 21
Screenshot 2021-10-20 at 14 42 19
Screenshot 2021-10-20 at 14 40 14
Screenshot 2021-10-20 at 14 40 07
Screenshot 2021-10-20 at 14 40 04
Screenshot 2021-10-20 at 14 39 36
Screenshot 2021-10-20 at 14 39 34
Screenshot 2021-10-20 at 14 39 32
Screenshot 2021-10-20 at 14 38 42ea
Screenshot 2021-10-20 at 14 38 28
Screenshot 2021-10-20 at 14 44 19
Screenshot 2021-10-20 at 14 46 06
Screenshot 2021-10-20 at 14 46 01
Screenshot 2021-10-20 at 14 44 21
Screenshot 2021-10-20 at 14 46 02
Screenshot 2021-10-20 at 14 46 07
Screenshot 2021-10-20 at 14 48 01
Screenshot 2021-10-20 at 14 48 03
Screenshot 2021-10-20 at 14 48 05

### DIFF
--- a/blockbase/block-templates/index.html
+++ b/blockbase/block-templates/index.html
@@ -1,13 +1,10 @@
 <!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
-<!-- wp:group {"layout":{"inherit":true}} -->
-<div class="wp-block-group"><!-- wp:query-title {"type":"archive"} /--></div>
-<!-- /wp:group -->
-
-<!-- wp:query {"tagName":"main"} -->
+<!-- wp:query {"tagName":"main","layout":{"inherit":true}} -->
 <main class="wp-block-query">
+<!-- wp:query-title {"type":"archive","style":{"spacing":{"margin":{"bottom":"100px"}}}} /-->
 <!-- wp:post-template -->
-<!-- wp:group {"layout":{"inherit":true}} -->
+<!-- wp:group -->
 <div class="wp-block-group">
 	<!-- wp:post-title {"isLink":true} /-->
 	<!-- wp:post-featured-image {"isLink":true} /-->
--- a/blockbase/block-templates/index.html
+++ b/blockbase/block-templates/index.html
@@ -1,5 +1,9 @@
 <!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
+<!-- wp:group {"layout":{"inherit":true}} -->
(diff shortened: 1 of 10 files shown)


kjellr:
@scruffian can you add some bottom margin or padding to the archive title block here? you can use the same amount of spacing that exists between each post in the list below.

scruffian:
I made a big change to the way this works. Currently we use a spacer block to separate posts in the query loop. We can't use a spacer block beneath the Query Title block, since we need there to be no space when the query title block has no output (like on the homepage). This means we need to find another way to add space beneath the Query Title block - the only option we have right now is to use CSS.

We could just add CSS to the Query Title block, but the problem with this is that if we want the gap beneath the Query Title block to match the gap between the blocks then we also need to control the gap between the posts with CSS, which is what this change does. I added a custom variable for this in the theme.json which means that we can use the same templates in all themes. The problem with this is that it makes it impossible for users to change this gap without editing the code.

scruffian:
Based on feedback from @MaggieCabrera I am going to try doing this through theme.json

scruffian:
I added spacing to the query title block
Screenshot 2021-10-22 at 15 13 29
Screenshot 2021-10-22 at 15 11 27
Screenshot 2021-10-22 at 15 07 03
Screenshot 2021-10-22 at 15 05 21
Screenshot 2021-10-22 at 15 01 15
Screenshot 2021-10-22 at 14 55 38
Screenshot 2021-10-22 at 14 54 09
Screenshot 2021-10-22 at 14 53 08
:

mikachan (approved):
This looks great! I've tested each theme on a range of archive pages and they all work well. The spacing also looks good to me, so I'm going to tentatively approve... If no one else gets a chance to review, we can always come back and adjust any spacing if needed.

PR #4879, opened 2021-10-20
Videomaker: Update navigation to be left-aligned

### LINKED ISSUE #4651: videomaker: Templates: Header template part (including mobile)

mikachan:
This PR makes the Videomaker nav left-aligned, plus some other small spacing tweaks.

Here's an example of when we can style using `.current-menu-item` too:

#### Related issue(s):
Closes https://github.com/Automattic/themes/issues/4651

### DIFF
--- a/videomaker/sass/_navigation.scss
+++ b/videomaker/sass/_navigation.scss
@@ -22,6 +22,7 @@
 			justify-content: space-between;
 		}
 		ul.wp-block-social-links {
+			gap: calc( var(--wp--custom--gap--horizontal) * 1.5 );
 			padding-bottom: calc( var(--wp--custom--gap--vertical) * 2 );
 		}
 	}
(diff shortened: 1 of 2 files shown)


videomaker/sass/_navigation.scss
```diff
@@ -6,32 +6,23 @@
...
-					}
-				}
 			}
 		}
 		.wp-block-navigation__responsive-container-content {
 			justify-content: space-between;
 		}
 		ul.wp-block-social-links {
-			justify-content: center;
+			gap: calc( var(--wp--custom--gap--horizontal) * 1.5 );
```
scruffian: This looks too big to me. I'd stick with the same as we have in Blockbase

mikachan: Yep ok, I'll change it.

scruffian (approved):
LGTM

PR #4887, opened 2021-10-21
Payton (work in progress)

alaczek:
I've been working on a Blockbase child theme - Payton - and I believe I have it in a reasonable place to deploy to .com and test some more.

### DIFF
--- a/payton/block-template-parts/header.html
+++ b/payton/block-template-parts/header.html
@@ -1,8 +1,19 @@
-<!-- wp:group {"className":"site-header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"40px"}}}} -->
-<div class="wp-block-group site-header" style="padding-bottom:40px">
-	<!-- wp:site-logo /-->
-	<!-- wp:site-title /-->
-	<!-- wp:site-tagline {"style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
-	<!-- wp:navigation {"className":"is-style-blockbase-navigation-improved-responsive","itemsJustification":"right","isResponsive":true,"__unstableLocation":"primary","__unstableSocialLinks":"social"} /-->
-</div>
+<!-- wp:group {"tagName":"header","style":{"spacing":{"padding":{"top":"10px"}}},"className":"site-header"} -->
+<header class="wp-block-group site-header" style="padding-top:10px">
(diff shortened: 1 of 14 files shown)


payton/child-theme.json
```diff
@@ -125,37 +125,43 @@
...
-                    "fontFamily": "var(--wp--preset--font-family--overpass)"
+                    "fontFamily": "var(--wp--preset--font-family--overpass)",
+                    "fontSize": "16px"
                 }
             },
             "h6": {
                 "typography": {
-                    "fontFamily": "var(--wp--preset--font-family--overpass)"
+                    "fontFamily": "var(--wp--preset--font-family--overpass)",
+                    "fontSize": "16px"
```
scruffian, line 159: You might want to use the preset variables for this, so if they change these will update automatically.

payton/style.css
```diff
@@ -7,11 +7,11 @@ Description:
 Requires at least: 5.7
 Tested up to: 5.7
 Requires PHP: 5.7
-Version: 0.0.3
+Version: 0.0.1
```
scruffian, line 10: Probably better to increment this rather than go back to 0.0.1 although it probably makes no difference.

payton/block-template-parts/header.html
```diff
@@ -1,8 +1,19 @@
...
-	<!-- wp:site-logo /-->
-	<!-- wp:site-title /-->
-	<!-- wp:site-tagline {"style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
-	<!-- wp:navigation {"className":"is-style-blockbase-navigation-improved-responsive","itemsJustification":"right","isResponsive":true,"__unstableLocation":"primary","__unstableSocialLinks":"social"} /-->
-</div>
-<!-- /wp:group -->
\ No newline at end of file
+<!-- wp:group {"tagName":"header","style":{"spacing":{"padding":{"top":"10px"}}},"className":"site-header"} -->
+<header class="wp-block-group site-header" style="padding-top:10px">
+	<!-- wp:navigation {"itemsJustification":"center","isResponsive":true,"__unstableLocation":"primary","style":{"typography":{"fontFamily":"var:preset|font-family|overpass"}},"fontSize":"small"} -->
```
scruffian, line 3: You might want to use the block style we use for most blockbase children to make the mobile navigation a bit nicer, but that's up to you.

scruffian (approved):
This is looking great. The only thing you might want to change is to use the preset values for font sizes rather than fixed values.

pbking:
Bringing this in.  Per comments additional work noted above can happen after.

PR #4889, opened 2021-10-21
Videomaker: Fix category trailing slash 

### LINKED ISSUE #4882: Videomaker: Trailing slash when no category set
## Steps to replicate

When posts don't have a category, the slash between category and date looks kind of weird:
Screenshot 2021-10-20 at 20 41 29

Any thoughts on this one @beafialho

mikachan:
Co-authored-by: @danieldudzic

This PR fixes the trailing slash if a post does not have categories.

image

This also removes the tag styles if tags are added to the `post-meta` group:

image

#### Related issue(s):
Closes #4882


videomaker/sass/_meta.scss
```diff
@@ -4,12 +4,29 @@ div.post-meta {
...
+
+		&::before {
+			background-color: transparent;
+			content: '/';
+			font-size: var(--wp--preset--font-size--normal);
+			margin-right: var(--wp--custom--gap--baseline);
+			-webkit-mask-image: none;
+			mask-image: none;
+			width: auto;
+			height: auto;
```
MaggieCabrera: Do we need this because you are using before for the slash? or do we need this anyway? Beacause I'm thinking, a different way of doing this would be something like: `.post-meta p:last-child { display: none; }` it will only happen when the post-terms blocks is not present because afaik when it's not there there is no empty div.

mikachan: Oh yeah that sounds good! All this was because I was using the before for the slash. Let me try this out now.

mikachan: Ah, it was also added to handle multiple post meta blocks being added by the user (or the user just moving these blocks around in general). At the moment, I think users would need to manually add a slash themselves between each item.

MaggieCabrera: I'm going to start a PR to move the icons outside of the general post meta styles

MaggieCabrera: [This](https://github.com/Automattic/themes/pull/4891) removes all the extra CSS from the icons if the class is not applied, should reduce the needed CSS here

MaggieCabrera: Also, Quadrat does the same thing with a dot as a separator, so maybe you want to grab inspiration from that.

mikachan: I've updated this so Videomaker is tweaking the styles in the same way as Quadrat. I think it looks good, the styles definitely look better!

This means that this change needs [your PR](https://github.com/Automattic/themes/pull/4891) before it's merged.

MaggieCabrera: Just did! now I think this needs a rebase

mikachan: Woop, this is now up to date with trunk and still working 😁

videomaker/sass/_meta.scss
```diff
@@ -4,4 +4,12 @@ div.post-meta {
 	a {
 		text-decoration: none;
 	}
+
+	> * {
+		&:not(:first-child):before {
+			color: var(--wp--custom--color--foreground);
+			content: "/";
```
scruffian, line 11: what happens with i18n and RTL if we take this approach? I guess we had that problem before :)

videomaker/block-template-parts/post-meta.html
```diff
@@ -1,10 +1,12 @@
-<!-- wp:group {"className":"post-meta","layout":{"type":"flex"}} -->
-<div class="wp-block-group post-meta">
+<!-- wp:group {"className":"post-meta","layout":{"type":"flex"},"style":{"spacing":{"padding":{"top":"1em","bottom":"1em"}}}} -->
```
scruffian: Should we use the block-gap variable here?

mikachan: Yeah good idea! I think variables usually make more sense than fixed values.

scruffian:
In RTL this did look like this
Screenshot 2021-10-22 at 09 55 41

With the changes I just made it now looks like this:
Screenshot 2021-10-22 at 09 55 52

scruffian:
@mikachan I updated this to use the --wp--style--block-gap variable as that's what we use for spaces between blocks. I also added 1em as a fallback.

scruffian (approved):
LGTM if you're happy with my changes

mikachan:
> @mikachan I updated this to use the --wp--style--block-gap variable as that's what we use for spaces between blocks. I also added 1em as a fallback.

Ah thanks, I didn't realise. These changes look good to me, thanks for your help!

PR #4894, opened 2021-10-22
Minor changes to get the deploy script to actually deploy themes

pbking:
The `deploy` action is an alias and doesn't with on a non-interactive shell, this changes how the command is called so that it will run from our script.

If the response to the command doesn't have the text "successfully deployed to" then the script waits 10 seconds and attempts to deploy again.  (Indefinitely)


pbking:
Because automation I have to bring this in before I can use it.

🤞

pbking:
It worked!

PR #4895, opened 2021-10-22
Blockbase: Fix fatal error with social navigation

scruffian:
We have had reports of this error on Blockbase:
`PHP Fatal error: Uncaught Error: Call to a member function appendChild() on null`

I can't reproduce it, but hopefully this code will prevent it happening.


pbking:
I also cannot reproduce (tried a number of ways) but also this change doesn't break anything and seems work a ship based on the reports.

LGTM

PR #4896, opened 2021-10-22
Payton: Fix query pagination containers

scruffian:
This removes markup in the query pagination container that is no longer required by Gutenberg.

Before:
Screenshot 2021-10-22 at 20 15 11

After:
Screenshot 2021-10-22 at 20 15 41


pbking (approved):
LGTM
🚢

PR #4897, opened 2021-10-22
Blockbase: Add a mixin for post meta

scruffian:
I noticed that when we updated the CSS for post meta in Quadrat, the same changes didn't get made to Payton so it looked like this:

Screenshot 2021-10-22 at 21 07 47

We have very similar code also in Videomaker. To make it easier to share this code I have created a mixin in Blockbase for it. Ideally this should be possible in Gutenberg, but for now the mixin will help.

Payton:
Screenshot 2021-10-22 at 20 37 55

Quadrat:
Screenshot 2021-10-22 at 20 37 29

Videomaker
Screenshot 2021-10-22 at 21 01 25


blockbase/sass/base/_mixins.scss
```diff
@@ -29,3 +29,18 @@
 		background-color: currentColor;
 	}
 }
+
+@mixin post-meta-with-separator( $content ) {
```
mikachan, line 33: This is cool!

mikachan (approved):
This looks good to me. I tested Payton, Quadrat, and Videomaker and they all look the same as before. I also think this is a great way to share these styles.

MaggieCabrera:
Yes, great idea, thank you!

MaggieCabrera (approved):

PR #4041, opened 2021-06-10
Coutoire: Include Menu Enhancements

Aurorum:
#### Changes proposed in this Pull Request:

1) Ensures that the logo doesn't overlap with the text content

https://user-images.githubusercontent.com/43215253/121583945-d40c2380-ca28-11eb-8deb-1acda6a321d1.mov

2) Allows users to ensure their menu doesn't exist. As mentioned in https://github.com/Automattic/themes/issues/1901#issuecomment-630361066, this will automatically be enabled for people who active the theme for the first time, but not for those who already have it active.

Screenshot 2021-06-10 at 20 18 27

3) Adds some space around the footer links

Screenshot 2021-06-10 at 20 19 11

#### Related issue(s):

Fixes #1901
Fixes #1838
Fixes #2072 
Fixes #4022

PR #4899, opened 2021-10-25
Coutoire: Menu Enhancements

scruffian:
This is an alternative to https://github.com/Automattic/themes/pull/4041, which I encountered a few issues with. (I couldn't put to that PR as it was created in a fork).

The main things I noticed were:
1. When you scrolled down the content would jump up before animating back into position
2. Sometimes when the page loaded the site logo animation would fire unexpectedly
3. The first time the site logo animation fired it was often jumpy.

co-authored with @Aurorum

### DIFF
--- a/coutoire/js/fixed-header-spacing.js
+++ b/coutoire/js/fixed-header-spacing.js
@@ -1,5 +1,4 @@
-(function() {
-
+( function () {
 	/**
 	 * Debounce
 	 *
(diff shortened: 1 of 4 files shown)


coutoire/js/fixed-header-spacing.js
```diff
@@ -42,46 +41,88 @@
...
+
 	/**
 	 * Run our page header height function
 	 */
 	pageHeaderHeight();
-})();
+
+	setTimeout( function () {
+		window.addEventListener( 'scroll', logoWidth );
+	}, 26 ); // We need to wait for the margin top to be set before we attach the scroll handler.
```
scruffian, line 127: This is really nasty, but I found that sometimes the scroll event was firing when the page loaded, and before the other things above had finished running.

coutoire/sass/_extra-child-theme.scss
```diff
@@ -385,4 +385,27 @@ a.wp-block-file__button {
...
+
+/**
+ * Animate logo resizing.
+ */
+ .site-logo {
+	width: 100%;
+	-webkit-transition: 0.25s linear;
+	-moz-transition: 0.25s linear;
+	-o-transition: 0.25s linear;
+	-ms-transition: 0.25s linear;
```
MaggieCabrera, line 398: super minor nitpick but I think the prefixes are added automatically during the build process

MaggieCabrera:
This looks like it's working fine. The only thing I see is that the tagline/site title can overlap the content if it's too long (and the font size is quite big, so it's bound to happen):

Screenshot 2021-10-27 at 11 30 50

kjellr:
(I am using a laughably huge image here, but) I'm seeing some weird double-animation when I scroll back up to the top of the page in Safari: 

scrolling


This doesn't happen in Chrome.

scruffian:
@jeffikus do you think this is still worth us investing time in?

scruffian:
We're putting this on hold for now

Issue #4859, opened 2021-10-18
Videomaker: Additional Block Patterns


beafialho:
In the meantime, after development started, we created a few more block patterns [in Figma here](https://www.figma.com/file/XnDSyzD9cBG5kbDaVZn9yG/Videomaker?node-id=183%3A563):

<img width="706" alt="Captura de ecrã 2021-10-18, às 16 28 40" src="https://user-images.githubusercontent.com/65220155/137774004-1ae81642-93e2-42d4-a05e-0d2f7f23e097.png">

I've built them in the editor. Here's the code for each of them:

**Film Credits**
```
{<!-- wp:spacer {"height":50} -->
<div style="height:50px" aria-hidden="true" class="wp-block-spacer"></div>
<!-- /wp:spacer -->

<!-- wp:paragraph {"align":"center","style":{"typography":{"lineHeight":"0.5"}},"fontSize":"medium"} -->
<p class="has-text-align-center has-medium-font-size" style="line-height:0.5">Based on the novel “Man Working at Dawn” by</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"align":"center","style":{"typography":{"lineHeight":"0.5"}},"fontSize":"large"} -->
<p class="has-text-align-center has-large-font-size" style="line-height:0.5">LOIS HERBERT</p>
<!-- /wp:paragraph -->

<!-- wp:spacer {"height":20} -->
<div style="height:20px" aria-hidden="true" class="wp-block-spacer"></div>
<!-- /wp:spacer -->

<!-- wp:paragraph {"align":"center","style":{"typography":{"lineHeight":"0.5"}},"fontSize":"medium"} -->
<p class="has-text-align-center has-medium-font-size" style="line-height:0.5">Screenplay by</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"align":"center","style":{"typography":{"lineHeight":"1.3"}},"fontSize":"large"} -->
<p class="has-text-align-center has-large-font-size" style="line-height:1.3">AUGUSTA SMITH<br>DAN ANDREWS<br>GARY HAN</p>
<!-- /wp:paragraph -->

<!-- wp:spacer {"height":20} -->
<div style="height:20px" aria-hidden="true" class="wp-block-spacer"></div>
<!-- /wp:spacer -->

<!-- wp:paragraph {"align":"center","style":{"typography":{"lineHeight":"0.5"}},"fontSize":"medium"} -->
[...]

PR #4900, opened 2021-10-25
Update/videomaker patterns

### LINKED ISSUE #4885: Videomaker: Block template preview broken on WordPress.com
## Steps to replicate
When you add a pattern that contains a block template block, the preview in the editor looks like this:
Screenshot 2021-10-20 at 20 41 15

However it does work in the front end:
Screenshot 2021-10-20 at 20 41 29

I'm not sure if there's anything we can do about this except wait for the new version of Gutenberg to land.

scruffian:
The template part block doesn't work in the frontend on WordPress.com, so this replaces references to that with the markup itself. This also makes it easier to adjust the spacing in these patterns.

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/4885

### DIFF
--- a/videomaker/block-template-parts/post-meta.html
+++ b/videomaker/block-template-parts/post-meta.html
@@ -1,12 +1,12 @@
 <!-- wp:group {"className":"post-meta","layout":{"type":"flex"},"style":{"spacing":{"padding":{"top":"var(--wp--style--block-gap,1em)","bottom":"var(--wp--style--block-gap,1em)"}}}} -->
 <div class="wp-block-group post-meta" style="padding-top:var(--wp--style--block-gap,1em);padding-bottom:var(--wp--style--block-gap,1em)">
-    <!-- wp:post-date {"isLink":true,"style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
-    <!-- wp:post-terms {"term":"category","style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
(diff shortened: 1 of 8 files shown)


beafialho:
This is not related to #4859, right?

I see an "unexpected or invalid content" message above the first video and an area with post tags below it, which we can remove.

Captura de ecrã 2021-10-26, às 11 59 28

We can also remove tags here, and in all the other "grid" patterns. Let's leave the post tags just for the "post" pattern.

Captura de ecrã 2021-10-26, às 12 01 57

pbking:
Seems (unrelated?) spacing changes in this diff (shifting from blockbase's custom-vertical to global styles block-gap for header element's margins).  Is that intentional?  Is that a change that should happen in Blockbase?

scruffian:
>  Is that intentional? Is that a change that should happen in Blockbase?

It is intentional - the gaps around the heading blocks looked too big to me. I also wonder if it should happen in Blockbase, but I've not investigated yet.

scruffian:

beafialho:
I still see that unexpected content message, it shows up as a repeated video in the front end.

Captura de ecrã 2021-10-27, às 10 16 30

scruffian:
> I still see that unexpected content message, it shows up as a repeated video in the front end.

Strange. Did you pull the latest version of this branch?

beafialho:
You're right, I wasn't testing with the latest version of this branch. LGTM!

pbking (approved):
Design approved it.
Changes look good. 
🚢

PR #4901, opened 2021-10-25
Videomaker: Change videos in patterns to point to VideoPress

### LINKED ISSUE #4884: Videomaker: Playback button
## Steps to replicate

On this (pdmazJ-O-p2) demo site the playback button is big:
Screenshot 2021-10-20 at 20 47 49

However on WordPress.com it looks like this:
Screenshot 2021-10-20 at 20 49 07

I'm not sure of the reasons for the difference but it would be good if it looked like the first example.

scruffian:
This updates all the patterns in Videomaker to use VideoPress videos rather than local ones. This means that they load with a big playback button.

#### Related issue(s):
Fixes #4884 and #4882

### DIFF
--- a/videomaker/inc/patterns/full-width-homepage.php
+++ b/videomaker/inc/patterns/full-width-homepage.php
@@ -19,7 +19,9 @@ return array(
 	<!-- /wp:column --></div>
 	<!-- /wp:columns -->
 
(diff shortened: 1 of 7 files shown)


mikachan (commented):
I've tested all these patterns and I can see the videos loading with the big playback button, so this is looking good. I'll leave final approval for @beafialho.

beafialho:
This looks good to me, all videos are working now.

PR #4908, opened 2021-10-26
Refactor Paragraph margin rules for consistent, non block-gap values

### LINKED ISSUE #4851: Blockbase: ponyfill p margins rules create unexpected behavior
The following rule in ponyfill.css which creates unexpected top margins for paragraphs:
```
p.wp-block.wp-block-paragraph, *[class^="wp-container"] > * + p, p {
    margin-top: 1em;
    margin-bottom: 1em;
}
```

## Steps to replicate

Create a new post with Title and three paragraphs. Then place one of the paragraphs inside a group block and move it to different positions in respect to the other two paragraphs.

## Result

When the group block is in position two or three, the margins are different from when in first position or when no group block is used:

top-margin-ponyfill

## Expected

Margins to be constant, as happens when the above ponyfill style rule is disabled:

top-margin-default

pbking:

This change:
* Removes the bottom margin from paragraphs
* Adjusts the top-margin of any element that FOLLOWS a paragraph to be the "paragraph gap" size.

Thus the "paragraph gap" rules are as such:
```
Inside any wp-container,
Any paragraph element that IS NOT THE FIRST thing and
Any element immediately after a paragraph element
use paragraph gap rules
```

This allows us to define the paragraph gap rules differently than the general block-gap rules and those rules will be followed on both sides of the paragraph.

Before:
image

(note that: the gap above the top of the group is 30px while the gap below is 1em)
After:

image
(note that: the gap above and below the group is 1em)

The gap between the bottom button and image remains 30px.

#### Related issue(s):
Closes: #4851

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -221,9 +221,9 @@ body.admin-bar .wp-site-blocks {
 
 p.wp-block.wp-block-paragraph,
 *[class^="wp-container"] > * + p,
+*[class^="wp-container"] > p + *,
 p {
 	margin-top: 1em;
-	margin-bottom: 1em;
 }
 
 .image-no-margin {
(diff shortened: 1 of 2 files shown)


MaggieCabrera (approved):
this looks great to me, thank you!

PR #4912, opened 2021-10-26
Videomaker: Fix post tags location. Closes #4888

### LINKED ISSUE #4888: Videomaker: Tags should go after Featured Image
| Currently      | Intended |
| ----------- | ----------- |
| Captura de ecrã 2021-10-21, às 10 37 14     |  Captura de ecrã 2021-10-21, às 10 38 48 |

danieldudzic:
Before:
Screenshot_26_10_2021__18_39

After:
Screenshot_26_10_2021__18_37

Co-authored-by: @mikachan

### DIFF
--- a/videomaker/block-template-parts/post-meta.html
+++ b/videomaker/block-template-parts/post-meta.html
@@ -4,3 +4,10 @@
     <!-- wp:post-terms {"term":"category","style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
 </div>
 <!-- /wp:group -->
+
+
+<!-- wp:group {"className":"post-tags-container"} -->
+<div class="wp-block-group post-tags-container">
+    <!-- wp:post-terms {"term": "post_tag","style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
+</div>
+<!-- /wp:group -->
--- a/videomaker/block-template-parts/post-meta.html
+++ b/videomaker/block-template-parts/post-meta.html
@@ -5,7 +5,6 @@
 </div>
 <!-- /wp:group -->
 
-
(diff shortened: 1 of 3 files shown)


scruffian:
I think we use the post-meta template in other places, so if we modify it here we'll break those other locations. Instead I think we need to just not use the post-meta template on the single.html template.

danieldudzic:
> I think we use the post-meta template in other places, so if we modify it here we'll break those other locations. Instead I think we need to just not use the post-meta template on the single.html template.

So, add specific meta parts inside of single.html then?

scruffian:
> So, add specific meta parts inside of single.html then?

Yeah I think we'll have to

videomaker/block-template-parts/post-meta.html
```diff
@@ -5,6 +5,7 @@
 </div>
 <!-- /wp:group -->
 
+
```
scruffian:
```suggestion
```

videomaker/inc/patterns/post-i-love-you.php
```diff
@@ -13,9 +13,6 @@
 	https://videopress.com/v/e7QDnuzH
 	</div></figure>
 	<!-- /wp:embed -->
-
-	<!-- wp:post-terms {"term":"post_tag"} /-->
```
scruffian, line 17: I'm not sure if we want to remove the tags from this pattern or not. @beafialho what do you think?

beafialho: If it's possible, we should leave the tags in this one.

danieldudzic: This has been removed because single.html will already load **tags** under the featured image on the single post page, plus the styling of tags inside of the content tag is broken:
themes
 
(this should be fixed either way)

scruffian (approved):
LGTM

PR #4916, opened 2021-10-27
Add a new theme named Kerr.

alaczek:
This is a work in progress, but in good enough shape to bring it to .com for testing.


scruffian (approved):
LGTM

PR #4920, opened 2021-10-27
Handle posts without titles

scruffian:
This needs design input!

When posts don't have a title they need another way to access them from the index. Conventionally this is done with a link on the post date, so I have modified the following themes to add links on post date where it already exists (Quadrat, Payton and Kerr), and added a post meta to the index for themes that don't contain the date at all (Blockbase and Russell)

Screenshot 2021-10-27 at 12 31 22
Screenshot 2021-10-27 at 12 25 55
Screenshot 2021-10-27 at 12 24 00
Screenshot 2021-10-27 at 12 22 00
Screenshot 2021-10-27 at 12 21 44


kjellr:
Yep, linking the post date is ok here. 👍

PR #4921, opened 2021-10-27
Videomaker: Reduce header spacing on header-footer-only template

### LINKED ISSUE #4871: Videomaker: Header bottom padding for pages without titles
While editing the [demo site](https://videomakerdemo.wordpress.com) and recreating the homepage, I noticed the spacing in the `header-footer-only` template is the same as in other pages (170px). 

In this template though, can we have less spacing (50px) to make it identical to the Figma mockup?

| Demo Site      | Figma |
| ----------- | ----------- |
|   Captura de ecrã 2021-10-19, às 15 40 20   |   Captura de ecrã 2021-10-19, às 15 46 21 |

mikachan:
This PR changes the bottom padding on the header from `170px` to `50px` on the header-footer-only template.

I did this by replacing the header template part reference with the contents of the header and then changing the padding. I'm not sure if this is the best way..

Before:
image

After:
image

#### Related issue(s):
Fixes #4871

### DIFF
--- a/videomaker/block-template-parts/header.html
+++ b/videomaker/block-template-parts/header.html
@@ -1,5 +1,5 @@
-<!-- wp:group {"tagName":"header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"170px"}}},"className":"site-header"} -->
-<header class="wp-block-group site-header" style="padding-bottom:170px">
+<!-- wp:group {"tagName":"header","layout":{"type":"flex"},"className":"site-header"} -->
+<header class="wp-block-group site-header">
 	<!-- wp:site-logo /-->
 	<!-- wp:site-title /-->
 	<!-- wp:site-tagline /-->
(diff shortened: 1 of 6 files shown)


scruffian:
I'm not sure about this. The reason for having the padding on the header was to make the space consistent across all the templates. If we don't want that then I think it's probably better to remove the padding and add spacers to each template which can be adjusted on a per-template basis.

beafialho:
@mikachan my suggestion of 50px was not accurate to achieve what's intended here. Can we make the header's bottom padding the same as top padding?

| Currently      | Intended |
| ----------- | ----------- |
| Captura de ecrã 2021-10-28, às 10 14 58   |  Frame 11 |

mikachan:
@scruffian Yeah I think I agree, it didn't feel right copying the header template contents just to make the spacing different. I've swapped the padding out for spacers in all the templates now.

@beafialho I've changed the spacing above the main content on this template to match the spacing above the header.

Ready for another review. Thanks both!

scruffian (approved):
LGTM

beafialho:
Looks perfect, thanks!

PR #4924, opened 2021-10-27
Add theme zips to deploy script

scruffian:
This is a first step towards generating our zips from the deploy script

Co-authored with @mikachan.


videomaker/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.0.21",
   "description": "",
   "main": "style.css",
+  "wpversioncompat": "5.7",
```
scruffian: Not that it matters for this PR, but I think this should be 5.8

theme-utils.mjs
```diff
@@ -35,10 +35,19 @@ function showHelp(){
...
+/*
+ Build .zip files for .com
+*/
 async function buildZip( themeSlug ) {
-	const currentPackage = JSON.parse(fs.readFileSync(`${themeSlug}/package.json`))
-	const themeVersion = currentPackage.version;
-	const wpVersionCompat = currentPackage.wpversioncompat;
+	let themeVersion;
+	let wpVersionCompat;
+	let styleCss = fs.readFileSync(`${themeSlug}/style.css`, 'utf8');
```
scruffian: Nice one

theme-utils.mjs
```diff
@@ -161,10 +162,47 @@ async function pushButtonDeploy(repoType) {
...
+	}
+
+	if (themeVersion && wpVersionCompat) {
+		response = await executeOnSandbox(`php ${sandboxRootFolder}bin/themes/theme-downloads/build-theme-zip.php --stylesheet=pub/${themeSlug} --themeversion=${themeVersion} --wpversioncompat=${wpVersionCompat}`, true);
+		
+		try {
+			console.log( response );
+		} catch( error ) {
+			console.log( error );
+		}
```
pbking: We wouldn't need to try/catch the log.  Not sure we even need to capture and log the `response` since `true` was passed to 'executeOnSandbox()` the output is already sent to stdout and no logic is done on the response.

theme-utils.mjs
```diff
@@ -161,10 +162,47 @@ async function pushButtonDeploy(repoType) {
...
+	} else {
+		console.log('Unable to build theme .zip.');
+		if (!themeVersion) console.log('Could not find theme version (Version:) in the theme style.css.');
```
pbking: I'm not sure if WP has a standard for it but if-statements without `{}`'s give me the heeby jeebies.

theme-utils.mjs
```diff
@@ -161,10 +162,47 @@ async function pushButtonDeploy(repoType) {
...
+		open('https://mc.a8c.com/themes/downloads/');
+	}
+}
+
 async function buildComZips(themes) {
-	//TODO: Figure out how to create these zip files automatically.
-	console.log('Please build the .zip files for the themes manually.', themes);
-	open('https://mc.a8c.com/themes/downloads/');
+	for ( let theme of themes ) {
+		buildComZip(theme);
```
pbking: I think this should be `await buildComZip(theme)` to ensure the operation has finished before starting the next one.  It PROBABLY operates fine concurrently but seems safer to let each operation completely finish before triggering the next.

pbking:
Looks really good.  I was really hoping we could build these assets like this!  I haven't ran it yet but would love to give it a spin on the next deploy!

theme-utils.mjs
mikachan: Oh yeah of course! This works and looks much better, thanks.

mikachan:
Thanks @pbking! I've just pushed the changes you suggested, fingers crossed for when we try it out.

pbking (approved):
Considering it has to be merged to `/trunk` in order to try this out on a legit deploy I'm keen to go ahead and bring this in.

Looks great!  Thanks y'all for getting this figured out!

PR #4926, opened 2021-10-27
Added a dotorg deploy script

pbking:
This is the first step to automating theme releases to dotorg.

This script can be ran manually, could be integrated into the manual deploy process, or handled by a github action.

For this change we are evaluating running it manually.

To test:
From the root of the project execute the script.  Note that the script needs to have the SVN_USERNAME and SVN_PASSWORD variables set.  Example:
```
%> SVN_USERNAME=test SVN_PASSWORD='test' ./deploy-dotorg.sh
```

Note that legit credentials aren't necessary since the script doesn't actually commit the change (yet).  Once [this line](https://github.com/Automattic/themes/compare/deploy/dotorg?expand=1#diff-8b2f23c53ecb2badc86c5f0f6d2083fa76616033bf4a04f65600ab36ba30527aR66) is uncommented legit credentials will be necessary.

Once the script has completed you will notice that there is a new `/deploy` folder in which you will find a subfolder with the theme slug that will be deployed.  In that folder you will see (empty) subfolders of ALL released versions, the latest of which will have the newly added version.

The script cycles through ALL folders and operates on those that have a package.json file.
If the version in that file doesn't already have a matching version in SVN it creates one, copies the resources for that theme into it (excluding those noted in `dotorg-exclude.txt`) and does a few minor transformations on the resulting files (removing wpcom-specific tags from style.css, etc).

Once we have verified that all of the information in the generated theme folders is correct (and any wpcom>wporg transformations are correct) we can re-run the script with the "commit" line uncommented and all of the themes will be updated in SVN deploying a new version of the theme.

All done!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 node_modules
+deploy/
 twentynineteen/
 theme-dev-utils/
 theme-dev-utils
--- a/deploy-dotorg.sh
+++ b/deploy-dotorg.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+
+# Make sure SVN credentials and project slug are set
+if [[ -z "$SVN_USERNAME" ]]; then
+	echo "Set the SVN_USERNAME secret"
+	exit 1
+fi
+
+if [[ -z "$SVN_PASSWORD" ]]; then
+	echo "Set the SVN_PASSWORD secret"
+	exit 1
+fi
+
+rm -rf ./deploy
+
+# Look into all the subfolders
+for THEME_SLUG in */ ; do
+
+	# And for those that have a package.json file...
+	if test -f "./${THEME_SLUG}/package.json"; then
+
+		THEME_VERSION=$(cat ./${THEME_SLUG}/package.json \
+		  | grep version \
+		  | head -1 \
+		  | awk -F: '{ print $2 }' \
+		  | sed 's/[",]//g' \
+		  | sed 's/-wpcom//g' \
+		  | tr -d '[[:space:]]')
+
+		# TODO: This does take into account the -wpcom appended to some theme versions.
+		# Ideally that can be removed from all of the themes versioning in this repository.
+		# I'm not convinced it's helpful...
+
+		printf "\n\nAttempting to deploy theme ${THEME_SLUG} version ${THEME_VERSION}\n"
+
+		SVN_URL="https://themes.svn.wordpress.org/${THEME_SLUG}/"
+		SVN_DIR="$PWD/deploy/${THEME_SLUG}"
+
+		svn checkout --depth immediates "$SVN_URL" "$SVN_DIR" --no-auth-cache --non-interactive > /dev/null
+
+		if [ ! -d "$SVN_DIR" ]; then
+			echo "No theme by that slug to be checked out. Probably not a theme.  Moving on."
+			continue;
+		fi
+
+		if [ -d "$SVN_DIR/$THEME_VERSION" ]; then
+			rm -rf $SVN_DIR
+			echo "Release already exists.  Moving on."
+			continue;
+		fi
+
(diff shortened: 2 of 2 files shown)


pbking:
Note: I just pushed a change that leverages the version stored in style.css instead of package.json.  Doing so revealed a lot more themes that needed to be pushed to .org.  `package.json` versioning IS the 'source of truth' for versioning, but that may be more important for the other deploy stuff we do.

I'm open to using either, and hope to have package.json file for ALL theme projects, but thought that idea was worth exploring.

Thoughts?

scruffian:
Since WordPress uses style.css as the source of truth, I'd say it's better to use that. Also, if we go that route, do we need to create package.json files for every theme?

scruffian:
This made my fans go wild!

scruffian:
There are many themes that haven't been updated on dotorg for over 2 years. If that's the case then they have to undergo a more thorough review and many would fail based on today's guidelines. We might want to make this work with a whitelist...

scruffian:
I'm a little uncomfortable with our approach of having wordpress.com tags on the themes by default and removing them for dotorg. I would rather we did it the other way round, but I'm not sure yet how that might work...

pbking:
> There are many themes that haven't been updated on dotorg for over 2 years. If that's the case then they have to undergo a more thorough review and many would fail based on today's guidelines. We might want to make this work with a whitelist...

Perhaps the existence of the package.json can be our 'whitelist'?  Or perhaps put a flag in said package.json that prevents the theme from being deployed to .org?

> I'm a little uncomfortable with our approach of having wordpress.com tags on the themes by default and removing them for dotorg. I would rather we did it the other way round, but I'm not sure yet how that might work...

It might not be appropriate to tag ALL of them in that way in the wpcom deploy process.  If so then the inclusion of those tags would need to be stored in some other way;  The style.css file seems the most relevant place to store that metadata to me.  Either way I see the repo as "what we put onto wpcom, but built in the open" with a relevant deploy process that takes care of the transformation to prepare it for wporg.  Which also includes the removal of the wpcom-specific PHP code (during the rsync process).  So it's not just those tags, but other things as well.

pbking:
I returned the version parsing to package.json with the hope that we can use its existence (or a flag in it) as our whitelist.

pbking:
I went ahead and did the above (leverage package.json as the determinator for deploying to wporg) and added package.json file to all the themes currently missing them, defaulting all of them to `"wporg": "ignore"`.  They could be selectively enabled as needed.

pbking:
Ok, undid the package.json parsing noted above.

After thinking a bit I agree with you and with https://github.com/Automattic/themes/pull/4943 there is a change for deploy process to use style.css instead of package.json as well.

pbking:
And now it operates on a list too.

@scruffian you're always right, it just takes me a little while to get there too. :P

scruffian:
One thing to note is that this currently will only work for Blockbase anyway as the other themes aren't enabled on dotorg yet.

scruffian:
This is looking good. Since only Blockbase has this enabled, I suggest we go ahead and add this as a GitHub action to push Blockbase changes, so we can test it out, and then we can add the other themes as that becomes possible.

pbking:
I would like to run this manually at least once before we set it up with an action.  I've already got the action part figured out (tested it deploying another github project to another svn repo so I'm pretty confident in it).

Do you JUST want blockbase in the list?  Or also all of Blockbase's (submitted) children as well?  My 'druthers is to ship it with [this list](https://github.com/Automattic/themes/pull/4926/files#diff-8b2f23c53ecb2badc86c5f0f6d2083fa76616033bf4a04f65600ab36ba30527aR18-R26).

scruffian:
I think only Blockbase will work for now...

scruffian:
This has now changed and all themes will work!

Issue #4881, opened 2021-10-20
Change Blockbase Header Design Target


pbking:
The design target for "Stock" Blockbase Headers has two different layouts for Desktop vs Mobile.

Desktop:
![image](https://user-images.githubusercontent.com/146530/138140081-62c785b5-c4d3-469c-ba2d-ce09b9384427.png)

Mobile:
<img src="https://user-images.githubusercontent.com/146530/138140203-d9c2f518-127a-4d40-80dc-2b69a8834437.png" width="300">

This has always been tricky but is getting increasingly [more difficult](https://github.com/Automattic/themes/issues/4860). (Using techniques outlined in that linked issue the desktop design target could be achieved (mostly) with just blocks and not special CSS, but how those techniques COULD result in mobile-only CSS differences is something I haven't figured out yet.).

TwentyTwentyTwo has similar header designs, but they remain consistent in the responsive view; a much easier feat.

Considering the "polyfill" goal of Blockbase shoehorning a design target like the current one seems the incorrect way to go.

I propose that instead we adopt a single design for Desktop/mobile and just use that.  Quadrat and Skatepark (etc) seem reasonable to leverage CSS to achieve a particular design target like that but for the Blockbase I propose we simplify instead.

My 'Druthers:
![image](https://user-images.githubusercontent.com/146530/138140030-f97d8230-2851-4b06-bf89-ff38725e2390.png)

scruffian:
FWIW I much prefer the stacked site title and site tagline, and I think we should do that on all our themes, except maybe Videomaker.

MaggieCabrera:
I kind of agree here. I also think it would be useful if Blockbase did like TT2 and had a couple of different options of headers that child themes can choose from, but we are writing way too much CSS to modify the mobile headers at the moment, the baseline one should be simpler.
@kjellr I'd love your opinion on this

jasmussen:
> we are writing way too much CSS to modify the mobile headers at the moment, the baseline one should be simpler

Happy to help here if I can. In cases like these, I'd agree you should ideally only need to write code to customize superficial stuff. As soon as structural stuff is changed, it increases the chance of drift.

MaggieCabrera:
> Happy to help here if I can. In cases like these, I'd agree you should ideally only need to write code to customize superficial stuff. As soon as structural stuff is changed, it increases the chance of drift.

[...]

PR #4928, opened 2021-10-28
Blockbase: Update the stacking of site title and tagline in Blockbase and co

scruffian:
This updates the way that site title and site tagline blocks stack in the header. I realised we could do this easily without any CSS. I think this is preferable as it will give a lot more space to the navigation before it stacks.

Blockbase:
Screenshot 2021-10-28 at 08 43 48

Quadrat:
Screenshot 2021-10-28 at 08 44 43

Geologist:
Screenshot 2021-10-28 at 08 52 35

Zoologist:
Screenshot 2021-10-28 at 09 06 30

Mayland blocks:
Screenshot 2021-10-28 at 08 53 45

Russell:
Screenshot 2021-10-28 at 08 55 11

Kerr:
Screenshot 2021-10-28 at 08 52 54

#### Related issue(s):
https://github.com/Automattic/themes/issues/4881

### DIFF
--- a/blockbase/block-template-parts/header.html
+++ b/blockbase/block-template-parts/header.html
@@ -1,8 +1,14 @@
 <!-- wp:group {"className":"site-header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"80px"}}}} -->
 <div class="wp-block-group site-header" style="padding-bottom:80px">
 	<!-- wp:site-logo /-->
-	<!-- wp:site-title /-->
-	<!-- wp:site-tagline {"style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
+
+	<!-- wp:group -->
+	<div class="wp-block-group">
+		<!-- wp:site-title /-->
+		<!-- wp:site-tagline {"style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
+	</div>
+	<!-- /wp:group -->
+
 	<!-- wp:navigation {"className":"is-style-blockbase-navigation-improved-responsive","itemsJustification":"right","isResponsive":true,"__unstableLocation":"primary","__unstableSocialLinks":"social"} /-->
 </div>
 <!-- /wp:group -->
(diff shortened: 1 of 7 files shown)


MaggieCabrera:
I agree with this if @beafialho and @kjellr give the thumbs up, it helps when either the site title, tagline or the menu are very long.

Skatepark doesn't seem to be aligned though, although this PR is not affecting it we should make it look like the rest:
Screenshot 2021-10-28 at 10 11 06

Looks like mobile works fine too with this change:

With tagline | Without
--- | ---
Screenshot 2021-10-28 at 10 13 29 | Screenshot 2021-10-28 at 10 13 46

beafialho:
This looks good, except for Skatepark which is misaligning on mobile:

Captura de ecrã 2021-10-28, às 10 45 03

Here's how it should align, according to the Figma file:

Captura de ecrã 2021-10-28, às 10 48 13

scruffian:
I don't see that problem on Skatepark:
Screenshot 2021-10-28 at 13 25 25

Either way this PR doesn't touch skatepark, so I don't think that should stop us merging.

pbking:
I tried to make it even more "blocky" by wrapping things in another row:

```
<!-- wp:group {"className":"site-header","layout":{"type":"flex","justifyContent":"space-between"},"style":{"spacing":{"padding":{"bottom":"80px"}}}} -->
<div class="wp-block-group site-header" style="padding-bottom:80px">

	<!-- wp:group {"layout":{"type":"flex"}} -->
	<div class="wp-block-group">
		<!-- wp:site-logo /-->

		<!-- wp:group -->
		<div class="wp-block-group">
			<!-- wp:site-title /-->
		</div>
		<!-- /wp:group -->
	</div>
	<!-- /wp:group -->

	<!-- wp:navigation {"className":"is-style-blockbase-navigation-improved-responsive","itemsJustification":"right","isResponsive":true,"__unstableLocation":"primary","__unstableSocialLinks":"social"} /-->
</div>
<!-- /wp:group -->
```

Doing so allowed more of the header.scss to be cleaned out, however the design changing so that the logo was above and in-the-center wasn't do-able that way.

As it stands this change just updates the design, it doesn't simplify the header like I was [hoping we could](https://github.com/Automattic/themes/issues/4881).  



My 'druthers would be to drop the design (at least for Blockbase) where the mobile layout is different than desktop allowing the design to be more fully represented in block markup rather than CSS.   But that's a bigger change then the design change done here and more impactful I think.

If the image needs to stay where it is this change looks good and we should ship it.  All the themes I checked out looked correct.

kjellr:
I think we should keep the image where it is today on mobile. 👍

pbking (approved):

kjellr:
@scruffian, @pbking: Let's please revert this change. This has a very noticeable impact to users of these themes, and it also does not look right today anyway: 

Screen Shot 2021-10-29 at 4 22 40 PM

PR #4936, opened 2021-10-28
Add/package files

pbking:
The adds package file to all of the themes that are missing package.json file.  Name is pulled from directory, version from style.css file.

Adding these will cause these themes to be included in the "version bump" part of the deploy process.


pbking:
Closing with preference of #4943

PR #4937, opened 2021-10-28
Geologist: Remove blockTypes from Geologist patterns

scruffian:
This removes `blockTypes` from some of the Geologist themes - this should be used to present variations of block templates, which these aren't.


jeffikus:
@scruffian code looks fine, for testing purposes, testing this with GB 11.8?

scruffian:
>  for testing purposes, testing this with GB 11.8?

The results should be the same with any GB version.

PR #4939, opened 2021-10-28
set wide with to 'full width but with margins'

### LINKED ISSUE #4938: Videomaker: "Wide" width should align with the header edges
Right now, when we set content to be wide, it has a set width limit. Can we change that to make it as the example shown?

Currently:

Captura de ecrã 2021-10-28, às 15 33 09    

Intended:

Captura de ecrã 2021-10-28, às 15 40 49

pbking:
Fixes: #4938

Changes wide with content to be "full width but with margins".

Before:
image

After:
image

### DIFF
--- a/videomaker/child-theme.json
+++ b/videomaker/child-theme.json
@@ -108,7 +108,7 @@
 		},
 		"layout": {
 			"contentSize": "800px",
-			"wideSize": "1300px"
+			"wideSize": "100%"
 		},
 		"typography": {
 			"fontFamilies": [
--- a/videomaker/theme.json
+++ b/videomaker/theme.json
@@ -312,7 +312,7 @@
 		},
 		"layout": {
 			"contentSize": "800px",
-			"wideSize": "1300px"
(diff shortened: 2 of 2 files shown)


scruffian (approved):

kjellr:
👋 [As noted over in the issue](https://github.com/Automattic/themes/issues/4938#issuecomment-953984661), I didn't think we needed this PR. It seemed like @beafialho was able to achieve what we wanted without it?

pbking:
Looks like we should revert and close this idea out...

beafialho:
Yes, I was able to achieve it! Though I must say, after seeing it in the other pages as well, I kind of like the way elements align in this "full width but with margins" way.

kjellr:
It might be an effect that we re-use in a later theme, but I think we should leave it out for this one. Videographers and their peers will be using large monitors more frequently than our typical users, and this results in some really stretched out layouts on large screens.

beafialho:
That sounds reasonable 👍

PR #4940, opened 2021-10-28
Independent Publisher 3: New Blockbase Child

scruffian:
I tried creating a Blockbase version of Indpendent Publisher 3. It was mostly very simple, I only needed to modify a few things in Blockbase. There are a couple of things that we can't do, which would be good:

1. It would be great to have the featured image in the header, but we need this PR first: https://github.com/WordPress/gutenberg/issues/24660
2. The split between body and heading fonts for this themes is quite different to other themes. For this to be customizable we'd need to change the approach to font customization in Blockbase somewhat.

| IP2 | IP3 |
| --- | --- |
| Screenshot 2021-10-28 at 17 48 44 | Screenshot 2021-10-28 at 17 49 19 | 
|independentpublisher2demo wordpress com_2016_07_02_start-of-a-new-journey_ | localhost_4759__p=849 |
| independentpublisher2demo wordpress com_about-me_ | localhost_4759__page_id=918 |
| independentpublisher2demo wordpress com_tag_journey_ | localhost_4759__tag=journey |
| independentpublisher2demo wordpress com_category_artwork_ | localhost_4759__cat=56 |


MaggieCabrera:
I think number 1 could be addressed with CSS using block styles until the upstream issue is resolved. What does number 2 entail?

blockbase/sass/blocks/_navigation.scss
```diff
@@ -2,8 +2,12 @@
 .wp-block-navigation {
 	// See https://github.com/WordPress/gutenberg/issues/34648
 	.has-child .wp-block-navigation__submenu-container {
-		background-color: var(--wp--custom--color--background);
-		border: var(--wp--custom--form--border--width) var(--wp--custom--form--border--style) var(--wp--preset--color--primary);
+		background-color: var(--wp--custom--navigation--submenu--color--background);
+		border: var(--wp--custom--navigation--submenu--border--width) var(--wp--custom--navigation--submenu--border--style) var(--wp--custom--navigation--submenu--border--color);
+
+		a {
+			color: var(--wp--custom--navigation--submenu--color--text);
```
MaggieCabrera, line 9: this is not defined in theme.json

scruffian:
If we do this it should be as a block theme.

PR #4943, opened 2021-10-28
Change deploy to use style.css not package.json

pbking:
Because of course it's more proper to use what ALL themes are guaranteed to have as the metadata source-of-truth.  Even if it means doing a little work refactoring the utility...

Pulls version information for deploy process from style.css instead of package.json for deployment processing.  Version bumps no longer require package.json file.

Sped up version bump processing.

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -109,9 +109,6 @@ async function pushButtonDeploy(repoType) {
 
 		let changedThemes = await getChangedThemes(hash);
 
-		//TODO: Can these be automagically uploaded?
-		//await buildChangedOrgZips();
-
 		await pushChangesToSandbox();
 
 		await updateLastDeployedHash();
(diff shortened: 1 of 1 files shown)


theme-utils.mjs
```diff
@@ -327,41 +323,65 @@ async function versionBumpThemes() {
...
 }
 
+function getThemeMetadata(styleCss, attribute) {
+	if ( !styleCss || !attribute ) {
+		return null;
+	}
+	switch ( attribute ) {
+		case 'Version':
+			return styleCss
+				.match(/(?<=Version:\s*).*?(?=\s*\r?\n|\rg)/gs)[0]
```
scruffian, line 364: I think we do something very similar when building the zips. Maybe we could share the code?

pbking: 👍 That was exactly the idea.  I took the code from the .zips branch.

pbking: That branch wasn't in at the time I wrote this so I couldn't consolidate.  I'll rebase and change this branch so that it does.

PR #4946, opened 2021-10-29
Deploy script: Use aliases when committing

scruffian:
See p1635495934300200-slack-C029FM1EH


jeffikus (approved):

pbking:
Good catch.  Yes that seems right.

I don't believe it will be an issue when we switch to git... But I would have to evaluate the git deploy process again to be sure.  The process of "landing the change" is different with git so there's the potential but if that's the case at least we know the fix.

PR #4948, opened 2021-10-29
blockbase checkbox for duotone

### LINKED ISSUE #4434: Add User-Configurable Duotone option to Customizer
So that a user can control the Global Duotone options they should be presented with an option to enable/disable the duotone usage on all of the images.

* Enabling it applies the primary/background colors to the duotone attribute of all relevant blocks.
* Disabling it removes all colors from the duotone attribute of all relevant blocks.
* When a user changes Primary/Background colors using the customizer (and the option is enabled) those Primary/Background colors should be applied to all relevant blocks' duotone attributes. (Note: If we are able to accomplish [this issue](https://github.com/WordPress/gutenberg/issues/33905) then this item can be simplified by just using the `--primary` and `--background` CSS variables.)
* Determining if the control is enabled or disabled should be done comparing values in the User Theme JSON rather than via theme_mod values.

The "FSE" equivalent will be to manually set the duotone values of the blocks in question in the FSE Global Styles panel.

MaggieCabrera:
### Do not Merge
Can be merged once this issue has been brought in and has landed on wpcom : https://github.com/WordPress/gutenberg/pull/36236

This PR adds a checkbox for themes that have duotone support that allows the user to turn it off for any blocks that have it set in theme.json directly from the customizer.

Screenshot 2021-11-04 at 13 12 33

There are two known issues with this approach:

- If we uncheck the duotone checkbox and reload the customizer page and check it back again, the preview will not show the filter (but it will work in the frontend). There's no easy way around this because when the filter is off we don't know which selectors to apply it back on.
- On non dev environments (with no debug settings on) the transient cache from Gutenberg removes the SVG filters, Ben has [opened an issue upstream](https://github.com/WordPress/gutenberg/issues/36208) but this effectively breaks the duotone filters in the frontend. That's happening regardless of this PR too, we just uncovered that while testing this.
EDIT: the cache issue is solved by https://github.com/WordPress/gutenberg/pull/36236

#### Related issue(s):

Closes https://github.com/Automattic/themes/issues/4434

### DIFF
--- a/blockbase/inc/customizer/wp-customize-colors-preview.js
+++ b/blockbase/inc/customizer/wp-customize-colors-preview.js
@@ -1,3 +1,13 @@
+var enabledDuotone = duotoneVars[ 'duotoneControl' ] === '1' ? true : false;
+
+wp.customize( 'duotone_control', ( value ) => {
+	value.bind( ( newValue ) => {
+		enabledDuotone = newValue;
+		toggleDuotoneFilter( '#wp-duotone-default-filter', enabledDuotone );
+		toggleDuotoneFilter( '#wp-duotone-custom-filter', enabledDuotone );
+	} );
+} );
+
 if ( userColorPalette && userColorSectionKey ) {
 	// For each of the palette items add a listener
 	userColorPalette.forEach( ( paletteItem ) => {
@@ -6,6 +16,7 @@ if ( userColorPalette && userColorSectionKey ) {
 			value.bind( ( newValue ) => {
 				paletteItem.color = newValue;
 				blockBaseUpdateColorsPreview( userColorPalette );
+				console.log( userColorPalette );
 			} );
 		} );
 	} );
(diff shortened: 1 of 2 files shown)


blockbase/inc/customizer/wp-customize-colors.php
```diff
@@ -121,21 +131,46 @@ function register_color_controls( $wp_customize, $palette ) {
 		}
 	}
 
+	function register_duotone_controls( $wp_customize ) {
+
```
scruffian:
```suggestion
```

blockbase/inc/customizer/wp-customize-colors.php
```diff
@@ -121,21 +131,45 @@ function register_color_controls( $wp_customize, $palette ) {
 		}
 	}
 
+	function register_duotone_controls( $wp_customize ) {
+		$wp_customize->add_setting(
+			'duotone_control',
+			array(
+				'default'           => true,
+				'capability'        => 'edit_theme_options',
+				'sanitize_callback' => array( __CLASS__, 'sanitize_boolean' ),
```
scruffian: Can we change this to https://developer.wordpress.org/reference/functions/rest_sanitize_boolean/ ?

blockbase/inc/customizer/wp-customize-colors.php
```diff
@@ -121,21 +131,45 @@ function register_color_controls( $wp_customize, $palette ) {
...
+				'transport'         => 'postMessage', // We need this to stop the page refreshing.
+			)
+		);
+
+		$wp_customize->add_control(
+			'duotone_control',
+			array(
+				'type'    => 'checkbox',
+				'section' => $this->section_key,
+				'label'   => __( 'Apply these colors to the theme\'s images', 'blockbase' ),
```
scruffian: This description isn't quite accurate - it depends what blocks the duotone is already applied to. How about just "enable duotone"?

scruffian:
This is looking great, just a few small comments.

pbking:
Works great for me locally.  Wasn't working as expected on wpcom.  

May be due to the cacheing issue.  If so suggest we just park this until that has been fixed and shipped.  Not sure how to determine that that is the problem though.

MaggieCabrera:
> Works great for me locally. Wasn't working as expected on wpcom.
> 
> May be due to the cacheing issue. If so suggest we just park this until that has been fixed and shipped. Not sure how to determine that that is the problem though.

yes, you can reproduce the problem locally! you just need to remove [the dev env variables](https://github.com/WordPress/gutenberg/blob/3e84b5671554281453ea0c635742d6a9e2994351/lib/compat/wordpress-5.9/get-global-styles-and-settings.php#L79) that are circumventing the cache in the first place. We definitely need [the fix](https://github.com/WordPress/gutenberg/pull/36236) before we launch the theme, but with it, the problem goes away (at least for me, locally)

pbking (approved):
👍 

Thanks @MaggieCabrera 
> yes, you can reproduce the problem locally! you just need to remove the dev env variables that are circumventing the cache in the first place. We definitely need the fix before we launch the theme, but with it, the problem goes away 

Confirmed the problem exists and is fixed by the linked issue.  Approving these changes to ship once that lands.

MaggieCabrera:
I had to add a change because I had JS errors on Blockbase with the original PR, care to have a quick look again @scruffian?

blockbase/inc/customizer/wp-customize-colors-preview.js
```diff
@@ -1,4 +1,8 @@
-var enabledDuotone = duotoneVars[ 'duotoneControl' ] === '1' ? true : false;
+var enabledDuotone = false;
+
+if ( window.duotoneVars ) {
```
scruffian, line 3: does this need to be window.duotoneVars? Couldn't it just be `duotoneVars`?

MaggieCabrera: it didn't work that way!

> Uncaught ReferenceError: duotoneVars is not defined

scruffian: I think you need to check the type of duotoneVars:

```
if ( typeof duotoneVars !== 'undefined' ) {
```

but I suppose your way works too

MaggieCabrera:
I think we should rebase this and check if it still works and then deploy to .org!

scruffian:
This needs https://github.com/WordPress/gutenberg/pull/38055 to be merged before it will work.

scruffian:
I rebased this and gave it another test and it still works. If someone else could confirm we can bring it in!

MaggieCabrera:
> I rebased this and gave it another test and it still works. If someone else could confirm we can bring it in!

I don't know if I'll have time today, but did you check dotcom? last time I tried it didn't work in there

scruffian:
It's hard to test on wpcom because the Gutenberg fix isn't there, but the customizer preview is working there.

pbking:
I just checked on edge sites on dotcom.  Seems 12.5 just landed there.  However the ability to enable/disable the duotone still didn't seem to be working.  It does work locally as described.

scruffian:
The fix only landed in GB today so it won't be in 12.5

MaggieCabrera:
I think we should close this after we discussed [this](https://github.com/Automattic/themes/issues/5471)

PR #4956, opened 2021-10-29
Videomaker: Limit the width of the header to be wide

scruffian:
Limits the header to be "wide" width:

Screenshot 2021-10-29 at 21 54 01

#### Related issue(s):
#4938


beafialho (approved):

kjellr (approved):
Yes, looks good to me. 👍 I've adjusted the top of the homepage demo to be wide width again too, so it should work well with this PR once it's merged. 

Thanks!

PR #4958, opened 2021-11-02
Kerr

alaczek:
This is a work in progress branch with the latest updates:
* added screenshot and readme files
* updated style.css with theme info and image credits
* updated default color settings for buttons and links
* updated header spacing.

I would appreciate a review and merge. Not sure if this fits the workflow, but could we leave this branch after merging so that I can push more changes to it? I'd delete it as a part of the theme launch (or I'm open to other suggestions).


kerr/readme.txt
```diff
@@ -0,0 +1,69 @@
+=== Kerr ===
+Contributors: Automattic
+Requires at least: 5.7
```
scruffian: I think this needs to be 5.8

kerr/readme.txt
```diff
@@ -0,0 +1,69 @@
...
+Tested up to: 5.7.2
```
scruffian: So does this.

scruffian (approved):

scruffian:
This has now been deployed.

PR #4966, opened 2021-11-02
Skatepark: Increase menu spacing & style current menu item

### LINKED ISSUE #4963: Skatepark: Spacing between menu items is incorrect
Current Dev|Mockup
---|---
Screen Shot 2021-11-02 at 3 15 33 PM|Screen Shot 2021-11-02 at 3 30 37 PM

mikachan:
This PR increases the space between the Skatepark menu items to match the mocks, and styles the current menu item:

| Before      | After |
| ----------- | ----------- |
| image      | image       |

(The current menu item styling is dependent on GB 11.9, but nothing will break if this gets merged on our side sooner.)

#### Related issue(s):
Fixes #4963

### DIFF
--- a/skatepark/sass/blocks/_navigation.scss
+++ b/skatepark/sass/blocks/_navigation.scss
@@ -2,4 +2,12 @@
 //preventing mobile menu items (and sub-menu items) from having a different size value
 .wp-block-navigation {
 	font-size: var(--wp--preset--font-size--small);
+
+	.wp-block-navigation-item {
+		&.current-menu-item > a {
+			text-decoration-line: underline;
+			text-decoration-thickness: 0.2em;
+			text-underline-offset: 0.35em;
+		}
+	}
 }
--- a/skatepark/assets/theme.css
+++ b/skatepark/assets/theme.css
@@ -130,6 +130,13 @@
 	font-size: var(--wp--preset--font-size--small);
 }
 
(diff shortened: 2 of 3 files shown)


danieldudzic (approved):
Very elegant. LGTM!

scruffian (approved):

skatepark/sass/blocks/_navigation.scss
```diff
@@ -2,4 +2,12 @@
 //preventing mobile menu items (and sub-menu items) from having a different size value
 .wp-block-navigation {
 	font-size: var(--wp--preset--font-size--small);
+
+	.wp-block-navigation-item {
+		&.current-menu-item > a {
+			text-decoration-line: underline;
+			text-decoration-thickness: 0.2em;
+			text-underline-offset: 0.35em;
```
MaggieCabrera, line 10: I wonder if we should use the [mixin](https://github.com/Automattic/themes/blob/trunk/skatepark/sass/base/_mixins.scss) that is used for this for the rest of the site. It won't look the same as the design, but it would be consistent with the rest of the underlines

mikachan: Ah yeah I like this idea, although it does look quite different from the design:

Custom for the design:

image

With this mixin:

image

MaggieCabrera: mmh yeah, don't make the change then

PR #4968, opened 2021-11-02
Videomaker: Header navigation styles

### LINKED ISSUE #4959: Videomaker: Navigation is missing underlines on hover, is unreadable in its active state
These links should be underlined on hover, and their active states should have black text against the white background. 

active-states

(Also, this menu should use the hamburger on desktop. It didn't look like there's another issue for that?)

cc @beafialho

mikachan:
This PR adjusts the Videomaker header navigation by underlining the menu items on hover and ensuring a contrasting text color is used in the active state.

vm-nav

I tested this with `"overlayMenu":"always"` set on the navigation block, as I think this should be how the final version of this menu will look when we can merge #4832.

#### Related issue(s):
Fixes #4959

### DIFF
--- a/videomaker/sass/_links.scss
+++ b/videomaker/sass/_links.scss
@@ -1,4 +1,4 @@
 a:active {
-	background: var(--wp--custom--color--foreground);
+	background: var(--wp--custom--color--foreground) !important;
 	color: var(--wp--custom--color--background) !important;
 }
--- a/videomaker/sass/_links.scss
+++ b/videomaker/sass/_links.scss
@@ -1,4 +1,4 @@
 a:active {
 	background: var(--wp--custom--color--foreground);
-	color: var(--wp--custom--color--background);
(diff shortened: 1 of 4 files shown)


videomaker/sass/_links.scss
```diff
@@ -1,4 +1,4 @@
 a:active {
 	background: var(--wp--custom--color--foreground);
```
scruffian: I wonder if we should add the !important to this line too, otherwise it's possible that one declaration could work and the other could not, leading to the same problem again. What do you think?
```suggestion
	background: var(--wp--custom--color--foreground) !important;
```

mikachan: I also thought this actually! I didn't think it was as likely that a background would be added to active links from GB, but it's definitely possible, and then we'd have the same problem again like you say. Maybe it's best to add it, and then we're doing our best to ensure contrasting colours here.

scruffian (approved):

PR #4972, opened 2021-11-03
Blockbase: Fix warning when we try to set properties on arrays

### LINKED ISSUE #4970: Blockbase: warnings when saving color or font changes
## Steps to replicate

I've seen this on trunk, using skatepark and changing the color palette, on my PHP error_log file I see warnings like this:

Screenshot 2021-11-03 at 12 28 01

/cc @scruffian

scruffian:
Sometimes the settings in theme.json are arrays not objects. This converts them to objects so that we can set properties on them.

#### Related issue(s):
Fixes #4970


MaggieCabrera (approved):
Yes, this solves my issue, thank you!

PR #4980, opened 2021-11-03
Blockbase: Simplify the approach to font customization

scruffian:
This changes our approach to font customization in Blockbase. The current implementation restricts our division of body and heading settings to specific blocks. This approach means that themes can use heading or body fonts on any blocks and the customization will still work.

This approach is much simpler than our previous implementation. We are using semantic rather than descriptive names for our typography settings (heading/body instead of Times/Arial). This means that in order to edit a font we just need to define a new typography setting with the same key.

This was made more complex by the fact that we have to deal with sites that are using the old settings. In this case we only show the reset button, and then when they reset we allow them the select a font again.

To test:
1. Switch to trunk
2. Select a Blockbase theme
3. Go to the Customizer and select a custom font
4. Confirm that you see the new font in the front end
5. Switch to this branch
6. Go to the Customizer
7. Confirm that you see the reset button, but not your selected fonts
8. Press the reset button
9. Confirm that the preview reloads and shows the default fonts
10. Confirm that the frontend shows the default fonts
11. Select a custom font
12. Confirm that the preview shows the correct font
13. Save changes
14. Confirm that the frontend shows the correct font

### DIFF
--- a/arbutus/theme.json
+++ b/arbutus/theme.json
@@ -47,66 +47,28 @@
 			"fontFamilies": [
 				{
 					"fontFamily": "\"Roboto\", sans-serif",
-					"slug": "roboto",
-					"name": "roboto",
-					"google": "family=Open+Sans:ital,wght@0,300;0,400;0,500;0,700;0,900;1,300;1,400;1,500;1,700;1,900"
+					"fontSlug": "roboto",
+					"slug": "body-font",
+					"name": "Body (Roboto)",
+					"google": "family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,800;1,900"
+				},
+				{
+					"fontFamily": "\"Roboto\", sans-serif",
+					"fontSlug": "roboto",
+					"slug": "heading-font",
+					"name": "Headings (Roboto)",
+					"google": "family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,800;1,900"
 				}
 			]
 		}
 	},
 	"styles": {
 		"blocks": {
-			"core/button": {
-				"typography": {
-					"fontFamily": "var(--wp--preset--font-family--roboto)"
-				}
-			},
 			"core/post-title": {
 				"typography": {
-					"fontFamily": "var(--wp--preset--font-family--roboto)",
 					"fontSize": "48px"
 				}
-			},
-			"core/pullquote": {
-				"typography": {
-					"fontFamily": "var(--wp--preset--font-family--roboto)"
-				}
 			}
-		},
-		"elements": {
-			"h1": {
-				"typography": {
-					"fontFamily": "var(--wp--preset--font-family--roboto)"
-				}
-			},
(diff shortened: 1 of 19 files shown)


pbking:
Isn't this essentially what we did the first time?  We moved from something similar to this and "inverted" the dependency configuring the values on the blocks directly.

I like the font-family dependency expressed this way, but what was the issue we were trying to overcome when we did this the first time?  And does this implementation overcome that?  (Going to poke around at the history to see if I can answer that.)

scruffian:
The issue before was that we stored the font settings in custom, which the user wasn't able to modify because of wp_kses restrictions on WordPress.com. This approach avoids the custom property.

pbking:
This is better than adjusting values in `.custom` for sure, and I was never very comfortable with the customizer overriding individual block anyway so I'm in favor of this change.

I poked around at it for a while and things seem good.  The child themes still need adjusting but otherwise I think this will be a good and better change.

MaggieCabrera:
Testing on Seedlet Blocks I get the following warning when I change fonts in the customizer:

`PHP Warning:  First parameter must either be an object or the name of an existing class in /Users/maggie/A8c/repos/a8c-projects/themes/blockbase/inc/customizer/wp-customize-colors.php on line 168`

MaggieCabrera:
This is working for me only in the customizer, the frontend is not getting the changes:

Frontend:

Screenshot 2021-11-23 at 10 26 36

Customizer:

Screenshot 2021-11-23 at 10 26 29

The CSS variables have the right values in the front but no fonts are loading at all in the console (no 404s or 200s, just no fonts at all in the network tab of the browser inspector)

arbutus/theme.json
```diff
@@ -47,66 +47,28 @@
 			"fontFamilies": [
 				{
 					"fontFamily": "\"Roboto\", sans-serif",
-					"slug": "roboto",
-					"name": "roboto",
-					"google": "family=Open+Sans:ital,wght@0,300;0,400;0,500;0,700;0,900;1,300;1,400;1,500;1,700;1,900"
+					"fontSlug": "roboto",
+					"slug": "body-font",
+					"name": "Body (Roboto)",
+					"google": "family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,800;1,900"
```
MaggieCabrera, line 53: what happens to child themes that don't make these changes? will this update break their sites?

scruffian: If they update Blockbase and not their child then it _may_ break their sites depending on what the child theme.json file looks like. In most cases the font declarations at present are more specific than this proposal, so they should continue working the same.

MaggieCabrera:
This seems to be working fine, maybe we should add a notice when the child theme.json is not updated but Blockbase is running this code so that the user knows that they need to update the theme.json file.

And as we talked about, this is going to be a major version change since it includes a non-backwards compatible change.

scruffian:
I added a notice in case some child themes aren't updated:

Screenshot 2021-11-23 at 13 45 49

To test this, try adding the version of the theme.json file from trunk for a child theme...

mikachan (approved):
This works for me!

PR #4984, opened 2021-11-04
Make it possible to configure the color of the social icons

scruffian:
This adds the option to configure the color of social links in the navigation via theme.json.

I'm not 100% sure that this is the right place to put this in the theme.json file - it might make more sense to put it under navigation...

### DIFF
--- a/blockbase/theme.json
+++ b/blockbase/theme.json
@@ -294,6 +294,11 @@
 				"opacity": 1,
 				"width": "150px"
 			},
+			"social-links": {
+				"color": {
+					"text": "pink"
+				}
+			},
 			"table": {
 				"figcaption": {
 					"typography": {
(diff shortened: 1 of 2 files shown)


blockbase/theme.json
```diff
@@ -294,6 +294,11 @@
 				"opacity": 1,
 				"width": "150px"
 			},
+			"social-links": {
```
scruffian: This is just for testing

MaggieCabrera: this is only being applied to social links inside navigation blocks so maybe we'd better change the custom variable name to reflect that?

scruffian: good idea

MaggieCabrera (approved):
besides deciding on the name of the variable, this works for me.

blockbase/theme.json
scruffian: changed to navigaiton/social-links, what do you think?

MaggieCabrera: yeah that sounds about right!

PR #4985, opened 2021-11-04
Make it possible to customize submenus in Blockbase

scruffian:
This makes it possible to customize the look of the navigation block independently from the form blocks, and add different colors to submenus.

The navigation block in Blockbase should look the same with these changes, but I think the design could be improved.

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -461,8 +461,12 @@ ol {
 }
 
 .wp-block-navigation .has-child .wp-block-navigation__submenu-container {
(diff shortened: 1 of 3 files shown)


pbking:
Made a small change to default submenu color to FOREGROUND instead of `inherit` so that it would match theme styles.  `inherit` in that context always seems to resolve to `black`.

Otherwise this both fixes the incorrect submenu coloring as well as allows for flexibility in defining the styles in other areas.  I'm not sure how this will resolve in Gutenberg but the need is there.

👍 to bring this in.

pbking (approved):

PR #4997, opened 2021-11-05
Arbutus: Initial commit.

alaczek:
This is an initial commit of the Arbutus theme generated with the Create Blockbase Theme plugin, with minimal changes.


arbutus/block-template-parts/post-meta-icons.html
```diff
@@ -0,0 +1,8 @@
+<!-- wp:group {"className":"post-meta","layout":{"type":"flex"}} -->
```
scruffian: I think we can delete this file

arbutus/block-template-parts/post-meta.html
```diff
@@ -0,0 +1,8 @@
+<!-- wp:group {"className":"post-meta","layout":{"type":"flex"}} -->
```
scruffian: I think we can delete this file

scruffian:
This looks good, but I think we can remove quite a bit of code which is duplicated from Blockbase.

I made a PR to update the approach in the plugin here: https://github.com/Automattic/create-blockbase-theme/pull/4

For this PR I think there are a couple of options:
1. Remove the templates that are the same as Blockase and remove the theme.json setting that match Blockbase
2. Try recreating the child theme with the PR above.

Whether you do 1 or 2 probably depends on how many changes you made. I'm sorry that this has caused you extra work. I'm AFK next week but someone on @Automattic/theam should be able to help.

MaggieCabrera:
I removed the duplicate templates and I think this is good to go. I did it myself because I don't want to block this because of that, but we'll find a fix so that we don't create duplicated templates from the generator.

MaggieCabrera (approved):
LGTM

alaczek:
Thanks Maggie! This is now deployed!

scruffian:
I think this also has a lot more stuff in the theme.json file than it needs. I'll spin up a PR...

scruffian:
Done in https://github.com/Automattic/themes/pull/5030

PR #4999, opened 2021-11-05
Add empty blockbase child

scruffian:
This creates a new empty blockbase child theme. This might be useful for people building a simple theme on top of blockbase as it avoids duplicating things that can be inherited from the parent theme. At the moment, even with this theme all the templates are duplicated.... 

To test, switch to this theme, and then try creating a child theme using the create blockbase theme plugin: https://github.com/Automattic/create-blockbase-theme

### DIFF
--- a/empty-blockbase-child/functions.php
+++ b/empty-blockbase-child/functions.php
@@ -2,7 +2,7 @@
 /**
  * Add Editor Styles
  */
-function typewriter_editor_styles() {
+function empty_blockbase_child_editor_styles() {
 	// Enqueue editor styles.
 	add_editor_style(
 		array(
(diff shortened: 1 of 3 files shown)


empty-blockbase-child/functions.php
```diff
@@ -0,0 +1,22 @@
+<?php
+/**
+ * Add Editor Styles
+ */
+function typewriter_editor_styles() {
```
MaggieCabrera: typewriter?

scruffian: Ooops :)

pbking:
Rather than this I would prefer the theme generator be smart enough to exclude the Blockbase stuff if the tweaks are on Blockbase, rather that require they be made on an interim theme.

scruffian:
This is no longer needed.

PR #5006, opened 2021-11-09
Zoologist: Replace theme screenshot

kjellr:
This PR replaces the theme screenshot for Zoologist so that it matches the screenshot + demo used on dotcom. 

We should replace the screenshot shown on dotcom with this one, since it also has consistent spacing when paired with its Geologist twin. (It looks like dotcom uses a separate screenshot right now?)

## Before

Geologist|Zoologist
---|---
Frame 11|screenshot

## After

Geologist|Zoologist
---|---
Frame 11|screenshot


MaggieCabrera (approved):
LGTM

PR #5012, opened 2021-11-10
Videomaker: Specify grid pattern header spacing

### LINKED ISSUE #4950: Videomaker: Tighten up space between the H4 and paragraph on the homepage pattern.
In the block pattern used on the homepage, we should tighten up the space between the video titles and the date below them: 

Before|After
---|---
current|intended

`calc(0.5 * var(--wp--style--block-gap))` looks right to me.

(Also, I'm not sure if this is what happens in the pattern natively, but on the demo site it looks like we're using the post date block for the dates, but that doesn't make sense since these are not individual posts? cc @beafialho)

mikachan:
This PR adds custom spacing to the headings in the grid pattern on Videomaker.

Before:
image

After:
image

#### Related issue(s):
Closes #4950

### DIFF
--- a/videomaker/inc/patterns/pattern-grid.php
+++ b/videomaker/inc/patterns/pattern-grid.php
@@ -8,8 +8,8 @@
 return array(
 	'title'      => __( 'Pattern: Grid', 'videomaker' ),
 	'categories' => array( 'videomaker' ),
-	'content'    => '<!-- wp:columns {"align":"wide","className":"pattern-grid"} -->
-	<div class="wp-block-columns alignwide pattern-grid"><!-- wp:column -->
+	'content'    => '<!-- wp:columns {"align":"wide"} -->
+	<div class="wp-block-columns alignwide"><!-- wp:column -->
 	<div class="wp-block-column">
 
 	<!-- wp:embed {"url":"https://videopress.com/v/bjvmxiQS","type":"video","providerNameSlug":"videopress","responsive":true,"className":"wp-embed-aspect-16-9 wp-has-aspect-ratio"} -->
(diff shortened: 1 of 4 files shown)


videomaker/inc/patterns/pattern-grid.php
```diff
@@ -18,8 +18,8 @@
 	</div></figure>
 	<!-- /wp:embed -->
 
-	<!-- wp:heading {"level":4} -->
-	<h4>' . esc_html__( 'Man Working at Dawn', 'videomaker' ) . '</h4>
+	<!-- wp:heading {"level":4,"style":{"spacing":{"margin":{"bottom":"calc(0.5 * var(--wp--style--block-gap))"}}}} -->
+	<h4 style="margin-bottom:calc(0.5 * var(--wp--style--block-gap))">' . esc_html__( 'Man Working at Dawn', 'videomaker' ) . '</h4>
```
pbking: Should we be calculating values for things in block markup?

I know it works, and we've done it elsewhere, and it helps makes things consistent.  However, the value can't be expressed to the user in that form (it's just blank in the Block details panel). 

I think it makes sense to use calc() and var() in CSS and in theme.json definitions, but I'm not convinced it's the most bestest in block markup.

Thoughts?

mikachan: Ah yes, I agree, and I didn't realize how this was expressed to the user. I guess even if these values didn't show as blank in the details panel, they're still a lot more confusing to read than a fixed value.

I'll update this PR and remember this for the future. Thanks!

pbking: It's not just you @mikachan!  I've been using them too; that's what they were made for after all. :P

I'm glad we've got consensus on limiting their use though.

pbking (approved):
👍 🚢

PR #5014, opened 2021-11-10
Fixed null-ref error when previewing colors for any theme but Skatepark

### LINKED ISSUE #4996: Blockbase: Color previews in customizer are broken
## Steps to replicate
1. Switch to Blockbase
2. Open the customizer
2. Try changing primary color
3. Notice that the preview updates the first time
4. Try changing the background color
5. Notice that the preview no longer updates

pbking:
The PR [enabling Duotone customizations](https://github.com/Automattic/themes/pull/4740) via the customizer seems to have introduced a null-ref error.  It's checking for the existence of a global variable so I thought the original code was fine but Chrome was throwing an error and made color customization previews stop working.

To reproduce try to change the colors in Blockbase.  Note the console error.

This change fixes that by referencing `window.` where the variable actually resides.

### DIFF
--- a/blockbase/inc/customizer/wp-customize-colors-preview.js
+++ b/blockbase/inc/customizer/wp-customize-colors-preview.js
@@ -25,7 +25,7 @@ function blockBaseUpdateColorsPreview( palette ) {
 	);
 	styleElement.innerHTML = innerHTML;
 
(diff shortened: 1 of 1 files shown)


MaggieCabrera (approved):
thanks for fixing this!

PR #5021, opened 2021-11-11
Skatepark: Fix multiline tags and categories

### LINKED ISSUE #4979: Skatepark: Tags broken when they are too many
When we use a lot of tags, here's how they behave:
Captura de ecrã 2021-11-03, às 17 10 45

It would be great if we could have them line up horizontally, as shown below:
Captura de ecrã 2021-11-03, às 17 11 32

cc: @kjellr

mikachan:
This PR ensures that the tags and categories on Skatepark line up horizontally, without any gaps. I'm not sure if I've done this in the best way. I tried to only use `flex` properties but I didn't get very far..

I've also slightly increased the gap between the icon and the text, as I noticed it was slightly bigger on the design.

Before:
image

After:
image

#### Related issue(s):
Closes #4979

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -914,6 +914,7 @@ p.wp-block-site-tagline {
 .is-style-post-category-icon,
 .is-style-post-tag-icon {
 	display: flex;
+	flex-wrap: wrap;
 }
 
 .is-style-post-author-icon:before,
--- a/blockbase/sass/base/_mixins.scss
+++ b/blockbase/sass/base/_mixins.scss
@@ -15,6 +15,7 @@
 
 @mixin post-meta-icon {
 	display: flex;
+	flex-wrap: wrap;
 	&:before {
 		align-self: center;
 		content: '';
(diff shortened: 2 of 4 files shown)


pbking:
Similarly strange behavior from Blockbase, I think this fix should be there rather than Skatepark.



And perhaps raised as an issue in Gutenberg.  I'm not sure why anyone would prefer (and certainly default) to having tags stacked in that way...

I made one small tweak to bring the icons down just a few pixels...
Before:
image

After:
image

This is good to ship for Skatepark, or could be refactored to be done in Blockbase instead.

pbking:
Went ahead and moved the `flex: wrap` bit up into blockbase to take care of that one.  There's no Gutenberg block up bubble this up to so I think this is good to go.

Before: (above)
After:
image

mikachan:
Thanks @pbking! Blockbase is looking good to me, and Skatepark still looks good.

I double-checked all the other child themes and they're also looking good, so I think this is ready ✨

pbking (approved):

PR #5023, opened 2021-11-11
Removed child creation tool and documention referencing it.  

pbking:
Removed child creation tool and documentation referencing it.  
Pointed instead to Create Blockbase Theme project.

Added a blurb about building Blockbase (scss compilation)

[Formatted Version of README](https://github.com/Automattic/themes/blob/fix/child-docs/blockbase/README.md)


MaggieCabrera (approved):
Thanks so much for doing this, love your words :D

Issue #5020, opened 2021-11-11
Blockbase refactor 404 page to use pattern block


MaggieCabrera:
Now that the pattern block is in Gutenberg, we should refactor de 404 page to use it instead of being a PHP template

PR #5031, opened 2021-11-15
Blockbase: Change references to gutenberg_block_template_part, to block_template_part

scruffian:
This function was renamed in https://github.com/WordPress/gutenberg/pull/36201, so we need to update our references to it.

### DIFF
--- a/blockbase/footer.php
+++ b/blockbase/footer.php
@@ -1,6 +1,10 @@
 <footer class="wp-block-template-part site-footer-container">
 		<?php
-			echo block_template_part( 'footer' );
+			if ( function_exists( 'block_template_part' ) ) {
+				echo block_template_part( 'footer' );
+			} else {
+				echo gutenberg_block_template_part( 'footer' );
+			}
 		?>
 	</div>
 </footer>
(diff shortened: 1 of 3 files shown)


jeffikus:
@scruffian can we not move this into an action, where we can check for the presence of the function? That way we can avoid the fatal in future if this changes again.

scruffian:
I don't see how we could make it an action, but we could create an abstraction around it. I don't think the name will change again though as this is going into core.

david-binda:
My two cents: The `block_template_part` function is making it's way to core, and is being backported to Gutenberg 11.9, but is not present in prior versions of Gutenberg, nor in latest WordPress version ( 5.8.2 ).

So, this change would make the theme only work with Gutenberg 11.9.0 at the moment.

So, perhaps a `function_exist` check, as suggested, might make it fully backward compatible? Eg.:

```php
if ( function_exists( 'block_template_part' ) ) {
    echo block_template_part( 'header' );
} else {
   echo gutenberg_block_template_part( 'header' );
}
```

The check on whether the original `gutenberg_block_template_part` function exists is, IMHO, not needed, as the theme works well with older versions of Gutenberg now, so it feels like a good fallback for cases when `block_template_part` function does not exist yet.

MaggieCabrera:
I agree with @david-binda here, and also I think this will stop being an issue when we implement https://github.com/Automattic/themes/issues/5020 and just simply remove the PHP templates.

I can't seem to reproduce the original issue on a gutenberg-edge sandboxed test site, though

Issue #4602, opened 2021-09-15
Quadrat: Buttons Hover and Colors Incorrect in Several Occasions


beafialho:
In the WP.com demo site, some buttons aren't looking as intended:

- The "Learn More" button has the correct hover effect, but the default text is white
- The "Subscribe" button has an incorrect hover effect, it should turn into an outline button

https://user-images.githubusercontent.com/65220155/133407456-407bc632-75e7-4d7f-bf64-034c18b16813.mov

- In the contact form, the "Send Message" button has a Regular font weight, it should be the same as others (Bold)

| Editor      | Demo Site |
| ----------- | ----------- |
| <img width="465" alt="editor" src="https://user-images.githubusercontent.com/65220155/133408276-6ed3381d-6246-4763-923a-bac5a05b2379.png">   | <img width="411" alt="site" src="https://user-images.githubusercontent.com/65220155/133408308-d1453fcb-1165-4ddf-8e08-a0c207223302.png"> |

danieldudzic:
I believe the above issues are caused by Jetpack styles.

@scruffian has opened an issue related to this: https://github.com/Automattic/jetpack/issues/15227

The problem with the Contact Form button is that the styles are getting overwritten by:
_wp-content/mu-plugins/grunion-contact-form/css/grunion.css_
```
.wp-block-jetpack-button button {
    padding: 20px 30px;
    border: inherit;
    font: inherit;
}
```

PR #5033, opened 2021-11-15
Quadrat: fixing hover styles for buttons inside the subscribe form

matiasbenedetto:
Hi :wave: 

Adding some CSS definitions to:
- Fix the hover style of the buttons rendered by the Jetpack subscribe form
- Fix font-weight in the submit button rendered by the Jetpack contact form

Hover styles fix demo:
hover button

Font-weight fix demo:
font weight

#### Related issue(s):
https://github.com/Automattic/themes/issues/4602

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -347,19 +347,24 @@ input[type=checkbox] + label {
  * Block Options
  */
 .wp-block-button.wp-block-button__link,
-.wp-block-button .wp-block-button__link {
+.wp-block-button .wp-block-button__link, .wp-block-button__link.wp-block-button__link,
+.wp-block-button__link .wp-block-button__link {
 	border-width: 0;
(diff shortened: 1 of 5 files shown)


pbking:
I think there is a jetpack issue opened that will ultimately fix this and allow us to remove these changes.  Make a comment pointing future-us to that change so that we can remove this when appropriate.

scruffian:
Do you think these rules should be in Blockbase?

matiasbenedetto:
I reverted the Quadrat changes and add the fix in Blockbase as @scruffian suggested. I think it looks much better now and it's not a hack we have to revert later.

scruffian:
> it's not a hack we have to revert later.

I still think this is something we'll want to remove from Blockbase eventually, when the Jetpack block is updated.

scruffian:
LGTM

scruffian (approved):

PR #5035, opened 2021-11-15
Videomaker: Make footer group block align wide

### LINKED ISSUE #4960: Videomaker: Footer should be wide width
Currently, the footer is sticked to the edges of the screen, but it should behave similarly to the header and be wide width too.

Captura de ecrã 2021-11-02, às 15 38 39

mikachan:
This PR changes the footer group block width to align wide rather than align full. I've also removed the left and right padding as I don't think we need it with align wide.

image

#### Related issue(s):
Closes https://github.com/Automattic/themes/issues/4960


MaggieCabrera (approved):
Looks good on desktop and mobile too, thanks!

PR #5038, opened 2021-11-16
Update all navigation blocks to new markup

mikachan:
This PR updates all navigation blocks to the new markup. This fixes a couple of issues ready for GB 11.9:

- Fixes broken vertical navigation blocks
- Fixes right and center alignments
- Removes a `margin-top` setting that's being inherited by the `wp-block-navigation` container*

&ast; This is only an issue if there's a paragraph block next to the navigation. This isn't needed if the site title and logo are inside a group block, as the navigation then sits next to a div. I saw it happening in quite a few themes so I thought I'd add a CSS fix in Blockbase.

To test, check each navigation block in Blockbase & Co still renders with its original settings, especially alignment and orientation.

### DIFF
--- a/arbutus/block-template-parts/header.html
+++ b/arbutus/block-template-parts/header.html
@@ -5,5 +5,5 @@
 
 <!-- wp:site-tagline {"fontSize":"small","fontFamily":"open-sans"} /-->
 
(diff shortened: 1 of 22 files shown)


scruffian (approved):
LGTM

PR #5045, opened 2021-11-17
Videomaker: Awards pattern redone to not use images for text

### LINKED ISSUE #4953: Videomaker: Try allowing more flexibility in the laurel patterns
Screen Shot 2021-10-29 at 3 26 08 PM

Let's explore to see if we can we create these patterns in a way that lets users insert their own image or text in between the laurels?

MaggieCabrera:

I took the liberty to change Bea's markup so that instead of cover blocks we are using rows. This avoids the cropping on mobile without any extra CSS and I think it's a bit more flexible. I also adapted the middle laurel to not use images too for the NFF logo. 

Screen Capture on 2021-11-16 at 18-16-15

#### Related issue(s):

Closes #4953


beafialho:
Thanks Maggie! Not sure why I'm not seeing the same on my end, but the behaviour in the video you uploaded looks good 👍

MaggieCabrera:
> Thanks Maggie! Not sure why I'm not seeing the same on my end, but the behaviour in the video you uploaded looks good 👍

What do you see? this one took me some time to make it work, maybe something broke along the way

beafialho:
I see this: 

https://user-images.githubusercontent.com/65220155/142183186-019a6ec0-39aa-452e-b9ea-c4032f7d3aa0.mp4

MaggieCabrera:
> I see this:
> laurels.mp4

Mmmh, that could be related to blockgap, I was using the latest version from Gutenberg. I'll keep testing and see if I can reproduce. Keen to see if this happens to anyone else too.

mikachan:
I'm seeing the same as @beafialho, where it seems blockgap is pushing the content into a tiny column. If I remove the gap from the group blocks inside the columns, then I see this:

https://user-images.githubusercontent.com/1645628/142193576-8a81f173-9e51-47bf-9924-b647ac296d68.mov

I'm using the latest GB trunk and I've tried this in Chrome and Firefox.

(Apart from this issue, I think this refactor is great.)

MaggieCabrera:
yeah, I had that issue initially where the rest of blockgap on the inner groups wouldn't work but at some point it worked again and I couldn't reproduce it anymore. I'm going to remove the block gap from the columns and investigate because that would be a GB issue when nesting blockgap rules

MaggieCabrera:
this is ready for another review

scruffian (approved):
LGTM. It might also be good to have these as separate patterns.

kjellr:
> It might also be good to have these as separate patterns.

How do you mean? One with editable text, and one that's just a set of images? If so, I'd advise against it, since they look so similar in the preview state. I think this new version is the one that provides more value, so I think we should stick with it alone.

PR #5047, opened 2021-11-17
Blockbase: Add child CSS if it exists

scruffian:
To make Blockbase children simpler we should add the child theme CSS in Blockbase itself, if it exsits. This means that child themes don't need any PHP to work, which will make it much easier to develop and review them since there are much fewer security concerns if the theme is only CSS and HTML.


blockbase/functions.php
```diff
@@ -72,6 +72,11 @@ function blockbase_editor_styles() {
 			blockbase_fonts_url(),
 		)
 	);
+	if ( file_exists( get_stylesheet_directory() . '/assets/theme.css' ) ) {
+		add_editor_style(
+			'/assets/theme.css'
+		);
+	}
```
MaggieCabrera, line 81: we should add a comment here

blockbase/functions.php
```diff
@@ -83,6 +88,10 @@ function blockbase_scripts() {
 	// Enqueue Google fonts
 	wp_enqueue_style( 'blockbase-fonts', blockbase_fonts_url(), array(), null );
 	wp_enqueue_style( 'blockbase-ponyfill', get_template_directory_uri() . '/assets/ponyfill.css', array(), wp_get_theme()->get( 'Version' ) );
+
+	if ( file_exists( get_stylesheet_directory() . '/assets/theme.css' ) ) {
+		wp_enqueue_style( 'blockbase-child-styles', get_stylesheet_directory_uri() . '/assets/theme.css', array('blockbase-ponyfill'), wp_get_theme()->get( 'Version' ) );
+	}
```
MaggieCabrera, line 97: and here too I think

pbking:
LOVE it.

PR #5054, opened 2021-11-18
Paxton 2: Initial commit.

alaczek:
There's been a discussion of launching simple, one-page designs as block-based themes, in an effort to increase the number of FSE compatible themes. This is an experiment doing just that :) 
I took an updated Paxton design and used the Create Blockbase theme plugin to generate the theme. It's super simple and includes only color/font config. The homepage content will be built in the editor and distributed via headstart. Please let me know if this is the right approach!

screenshot

cc @ianstewart


alaczek:
Oh, the plugin generated the block-templates and block-template-parts dirs, but since they are empty I couldn't commit them... do we still want to include them in the theme, even if empty? If so, how do I do it? :)

alaczek:
I added a custom "Footer only" template so that we can display WordPress.com credits at the bottom.

alaczek:
@scruffian I removed the header.html, post-meta-icons.html, post-meta.html from the `block-template-parts` dir (and also all the page templates inside the `block-templates` dir), since I want to fall back on Blockbase - is that the right approach, or should I have left them in place?

scruffian:
> @scruffian I removed the header.html, post-meta-icons.html, post-meta.html from the block-template-parts dir (and also all the page templates inside the block-templates dir), since I want to fall back on Blockbase - is that the right approach, or should I have left them in place?

That's perfect.

scruffian:
> I added a custom "Footer only" template so that we can display WordPress.com credits at the bottom.

I believe the WordPress.com footer credit should work with the other Blockbase themes, so this should not be necessary.

jeffikus:
Yes, @MaggieCabrera worked on the footer credits for block themes in wpcomsh issue 781

kjellr:
This is cool, thanks @alaczek! 

Can we add a specific tag and/or mention me + @beafialho on these and other generated themes so we can keep an eye on them as they go out?

alaczek:
Sure @kjellr, will do!

> I believe the WordPress.com footer credit should work with the other Blockbase themes,

@jeffikus Could you explain how this works? (Or what does it need to work?)
When I use the "Blank" page template (on a site running Blockbase), there is no footer and no credits.

alaczek:
Closing as we're not going to launch Paxton 2 as a free theme after all.

Issue #5042, opened 2021-11-16
Videomaker: Remove tags and increase spacing in Blog


beafialho:
In the default template let's adjust two things, as [shown](https://www.figma.com/file/XnDSyzD9cBG5kbDaVZn9yG/Videomaker?node-id=68%3A1223) in the Figma file:
- Omit tags and only show them in the posts
- Increase spacing between post title, featured image and excerpt to match the mockups (perhaps `45 px`)

Currently:
<img width="504" alt="Captura de ecrã 2021-11-16, às 17 07 37" src="https://user-images.githubusercontent.com/65220155/142031937-75b6d600-fb1a-40a5-a30c-b914ff27da27.png">

Intended:
<img width="616" alt="Captura de ecrã 2021-11-16, às 17 05 47" src="https://user-images.githubusercontent.com/65220155/142031835-a8c34ea9-ae40-4cf2-a1f3-5909348837bd.png">

mikachan:
Closed with #5057

PR #5057, opened 2021-11-18
Videomaker: Remove tags from the index and increase margin under post title

scruffian:
This remove post tags from the index template. This now means that the post-meta template is only used once, so we should just remove it.

Also adds some space under the post title on the index.

#### Related issue(s):
#5042


MaggieCabrera (approved):
LGTM!

PR #5059, opened 2021-11-18
Videomaker: Update Videomaker navigation

### LINKED ISSUE #4656: videomaker: Navigation (if needed)

scruffian:
Implements the idea in https://github.com/Automattic/themes/issues/4656#issuecomment-967120686

Screenshot 2021-11-18 at 17 29 33

#### Related issue(s):
#4656

### DIFF
--- a/videomaker/block-template-parts/header.html
+++ b/videomaker/block-template-parts/header.html
@@ -7,6 +7,6 @@
 		<!-- wp:site-tagline /-->
 	</div>
 	<!-- /wp:group -->
-	<!-- wp:navigation {"overlayMenu":"always","__unstableLocation":"primary","className":"is-style-blockbase-navigation-improved-responsive","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"}} /-->
+	<!-- wp:navigation {"overlayMenu":"always","__unstableLocation":"primary","__unstableSocialLinks":"social","className":"is-style-blockbase-navigation-improved-responsive","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"}} /-->
 </header>
 <!-- /wp:group -->
--- a/blockbase/sass/blocks/_navigation.scss
+++ b/blockbase/sass/blocks/_navigation.scss
@@ -35,6 +35,7 @@
 		.wp-block-navigation-item {
 			align-items: flex-start;
 		}
+
 		.wp-block-navigation__responsive-container-content .has-child { // Needed for specificity to beat the navigation block CSS
 			.wp-block-navigation__submenu-container {
 				font-size: var(--wp--preset--font-size--normal);
--- a/videomaker/assets/theme.css
+++ b/videomaker/assets/theme.css
@@ -103,6 +103,11 @@ div.post-meta a {
 	align-items: flex-end;
 }
 
+.wp-block-navigation.is-style-blockbase-navigation-improved-responsive.is-responsive .is-menu-open.wp-block-navigation__responsive-container {
+	font-size: var(--wp--preset--font-size--large);
+	font-weight: 200;
+}
+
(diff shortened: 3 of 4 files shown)


pbking:
The main link font sizes seem pretty far off from target.  Shouldn't they be much larger?

This change also has changes to the awards block patterns.  Perhaps branched from another branch with unnecessary changes?

scruffian:
You're right, I updated the font sizes.

The menu items should also be vertically centered. This is pretty hard to do with the currrent implementation and would take significant reworking, so if we want to do that I suggest we do it in a followup PR.

kjellr:
What is the best way to get a social navigation set up so that I can test that aspect of this? Using the customizer doesn't seem to do it: 

Screen Shot 2021-11-22 at 3 16 52 PM

... and doing this in the site editor gets weird results: 

Editor|Front End
---|---
Screen Shot 2021-11-22 at 3 19 06 PM|Screen Shot 2021-11-22 at 3 18 49 PM

kjellr:
Oh wait, my mistake. I can get it working in the site editor (But not in the customizer? 🤷)

kjellr:
I'm seeing very bizarre vertical spacing here (and in `trunk`) when I update this in the site editor:

Screen Shot 2021-11-22 at 3 24 07 PM



> The menu items should also be vertically centered. This is pretty hard to do with the currrent implementation and would take significant reworking, so if we want to do that I suggest we do it in a followup PR.

Not sure if this is viable, but this spacing bug would actually be solved if we adopted `justify-content: center` instead of `space-between` inside of the expanded menu:

Screen Shot 2021-11-22 at 3 29 11 PM

scruffian:
Oh yeah nice one:
Screenshot 2021-11-24 at 13 51 44

kjellr:
I'm seeing some weirdness when there are a lot of items in the menu. They start flowing (and scrolling) horizontally: 
Screen Shot 2021-11-24 at 9 01 48 AM
Screen Shot 2021-11-24 at 9 01 31 AM

scruffian:
OK I worked out what was going on here - the social navigation was missing as you said above - you shouldn't need to add it from the site editor.

The problem is that if you have justify content center then when there's not many menu items it looks like this:
Screenshot 2021-11-24 at 17 31 57

The alternative with space-between looks like this:
Screenshot 2021-11-24 at 17 35 29

@beafialho @kjellr which do you prefer?

kjellr:
I definitely prefer the top one. That looks right to me:

scruffian:
In that case can we merge this? :)

videomaker/block-template-parts/header.html
```diff
@@ -7,6 +7,6 @@
 		<!-- wp:site-tagline /-->
 	</div>
 	<!-- /wp:group -->
-	<!-- wp:navigation {"overlayMenu":"always","__unstableLocation":"primary","className":"is-style-blockbase-navigation-improved-responsive","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"}} /-->
+	<!-- wp:navigation {"overlayMenu":"always","__unstableLocation":"primary","__unstableSocialLinks":"social","className":"is-style-blockbase-navigation-improved-responsive","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"}} /-->
```
kjellr, line 10: What is this `__unstableSocialLinks` attribute about?

kjellr (approved):
Left one minor comment, but this is working on my end. 👍

videomaker/block-template-parts/header.html
scruffian: This is how we show a social navigation inside a navigation block. It allows us to hide the social links inside the overlay, while still allowing menus to be updated via the customizer.

PR #5061, opened 2021-11-18
Skatepark: Add space below the Featured Image on the single post page

### LINKED ISSUE #4976: Skatepark: Add more spacing between featured image and post content
There should be more spacing between the featured image and the post content. According to the [Figma file](https://www.figma.com/file/CA8Gb16mt9yYNWOiVLAWak/Skatepark?node-id=2691%3A2220), the first paragraphs on posts should also have `184px` indentation. Here's how it should look:

| Currently      | Mockups |
| ----------- | ----------- |
| Captura de ecrã 2021-11-03, às 17 08 13    |  Captura de ecrã 2021-11-03, às 17 08 45 |    

cc: @kjellr

danieldudzic:
The logic behind the `calc()` code is trying to mimic the way the space gets calculated when a Featured Image isn't set.

Before:
space

After:
star

Closes #4976

### DIFF
--- a/skatepark/block-templates/single.html
+++ b/skatepark/block-templates/single.html
@@ -9,7 +9,7 @@
 
 <!-- wp:group {"align":"full"} -->
 <div class="wp-block-group alignfull">
-<!-- wp:post-featured-image {"align":"full"} /--></div>
+<!-- wp:post-featured-image {"align":"full", "style":{"spacing":{"margin":{"bottom":"calc(2*var(--wp--style--block-gap) + 3*var(--wp--custom--gap--horizontal))"}}}} /--></div>
 <!-- /wp:group -->
 
 <!-- wp:post-content {"layout":{"inherit":true}} /-->


skatepark/block-templates/single.html
```diff
@@ -9,7 +9,7 @@
 
 <!-- wp:group {"align":"full"} -->
 <div class="wp-block-group alignfull">
-<!-- wp:post-featured-image {"align":"full"} /--></div>
+<!-- wp:post-featured-image {"align":"full", "style":{"spacing":{"margin":{"bottom":"calc(2*var(--wp--style--block-gap) + 3*var(--wp--custom--gap--horizontal))"}}}} /--></div>
```
pbking, line 12: I don't think we should be using `var()` or `calc()` in block markup; It doesn't work in the FSE when you load that block template.  It makes the block markup more verbose, but I'm of the opinion that we should be verbose there, even if it means more manual changes when things like "block gap" needs to change.

MaggieCabrera: About this, looks like there's a proposal so that those values are [not empty](https://github.com/WordPress/gutenberg/issues/35558) on the block sidebar

danieldudzic: @MaggieCabrera Perhaps this PR could be moved to `Approved but not mergable`?

MaggieCabrera: no, I think we can merge this anyway, since only the UI is affected by the future change and not the actual looks of the site

MaggieCabrera (approved):
I think this is good to go

PR #5063, opened 2021-11-18
Fix Git Deploy

pbking:
### Fixes the deploy process to work correctly in GIT.  
Previously the archanist commands weren't properly mapped and the command node was executing wasn't actually interactive.  This change fixes both of those problems.

### Corrects release tagging mechanism 
by improving root project version bumping (it was broken since moving to style.css for version information)

NOTE:  There is STILL CURRENTLY a problem deploying wpcom zips of the themes using the script when deploying with git.

### To Test
Make sure your sandbox is in git mode and deploy changes with the `npm run deploy:git` command

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -15,19 +15,14 @@ const premiumThemes = [ 'videomaker', 'videomaker-white' ];
 	let command = args?.[0];
 	switch (command) {
 		case "push-button-deploy-git": return pushButtonDeploy('git');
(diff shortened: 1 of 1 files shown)


pbking:
These changes (and a further one correcting some paths) have already been included in [another change](https://github.com/Automattic/themes/commit/855e4a4072a4e9fcc81c89e88e5df3410356533e).  Closing this.

PR #5064, opened 2021-11-19
Arbutus: Update Header

scruffian:
I think its better to stack the site title and tagline vertically like this, as it gives more room to the navigation:

Screenshot 2021-11-19 at 10 15 10


alaczek:
Looks good, let's do it!

PR #5080, opened 2021-11-23
Push themes to premium

scruffian:
This updates our deploy script to push themes to the premium repo as well as the public one.


theme-utils.mjs
```diff
@@ -4,26 +4,30 @@ import open from 'open';
 import inquirer from 'inquirer';
 
 const remoteSSH = 'wpcom-sandbox';
-const sandboxPublicThemesFolder = '/home/wpdev/public_html/wp-content/themes/pub';
+const sandboxThemesRootFolder = '/home/wpdev/public_html/wp-content/themes/';
 const sandboxRootFolder = '/home/wpdev/public_html/';
 const isWin = process.platform === 'win32';
+const premiumThemes = [
+	'videomaker'
```
scruffian, line 11: Themes in this array will get pushed to premium not pub

jeffikus:
This makes sense to me - is this pushing Videomaker to both /pub/ and /premium/ ?

scruffian:
No, it should push videomaker to premium and everything else to pub.

scruffian:
@pbking I'd appreciate your input on this. Think all that remains to do is to put the hash in the premium directory.

blockbase/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "blockbase",
-  "version": "2.0.0",
+  "version": "2.0.1",
```
pbking, line 3: Would like to pull out the version changes from this change.  There shouldn't be a reason for those right?  Just something committed testing out the script?

theme-utils.mjs
```diff
@@ -219,17 +230,19 @@ async function checkForDeployability(){
...
 */
-async function landChangesSvn(diffId){
+async function landChangesSvn(diffId, repo){
+	const sandboxThemesFolder = getThemesFolderFromRepo( repo );
 	return await executeOnSandbox(`
-		cd ${sandboxPublicThemesFolder};
+		cd ${sandboxThemesFolder};
 		svn ci -m ${diffId}
 	`, true );
 }
```
pbking, line 248: getChangedThemes will also have to operate with the pub/premium separation

pbking:
The "changes" that are recorded are all of the changes between HASH...HEAD and those changes are going to be noted no matter which repo is targeted.  What will be reported is the entirety of the changes listed in the diffs in both repos. Which might not be a big deal, but I did want to bring it up.

theme-utils.mjs
pbking: Working on that change now...

pbking:
I have thus-far been unable to get the script to operate on the /premium repo via SVN.  AFAIK I have my credentials stored and the same commands are working if I run the same commands in an SSH session on my sandbox they work fine.  But in the script /pub works and /premium does not.

```
% node theme-utils.mjs clean-sandbox-svn premium

Updating '.':

svn: E170013: Unable to connect to a repository at URL 'https://wpcom-premium-themes.svn.automattic.com'
svn: E215004: No more credentials or we tried too many times.
Authentication failed

(node:38691) UnhandledPromiseRejectionWarning: svn: E170013: Unable to connect to a repository at URL 'https://wpcom-premium-themes.svn.automattic.com'
svn: E215004: No more credentials or we tried too many times.
Authentication failed
(Use `node --trace-warnings ...` to show where the warning was created)
(node:38691) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:38691) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Unfortunately as I have been as of yet unable to get past the cleaning portion of the workflow I'm unable to test this completely...

UPDATE: turns out in switching back and forth from GIT <> SVN mode I think I lost my SVN stored credentials.  `SVNAUTHCACHE=1 svn up .` ran in my `~/public_html` folder and again storing my credentials seems to have put me where I needed to be.

jeffikus:
@pbking I checked your permissions for those folders and all seems in order for SVN access. Can you try committing something trivial on your sandbox, even if it's updating a readme of a retired theme and let me know if you can do that?

pbking:
As it exists our script is pushing ALL changes that happened between HASH...HEAD to the /repo.  Prior to this change it's just /pub but we're trying to add /premium.

Unfortunately the way we resolve those changes is to copy (rsync) the files added in that changeset and remove those removed.  To do that in BOTH repositories means that we'll be trying to add/remove ALL of the changes in the changeset to BOTH repositories.  "Pushing changes" isn't "theme specific" so all of the changes made in the '/videomaker' folder will still be pushed to /pub and all of the OTHER changes will still be pushed to /premium.

As complicated as this is I don't think achieving this is going to be "this easy".

theme-utils.mjs
```diff
@@ -95,69 +99,75 @@ async function pushButtonDeploy(repoType) {
...
+	let diffUrl;
 
-		let changedThemes = await getChangedThemes(hash);
+	await versionBumpThemes( repo );
 
-		await pushChangesToSandbox();
+	let changedThemes = await getChangedThemes(hash);
 
-		await updateLastDeployedHash();
+	await pushChangesToSandbox( repo );
```
pbking, line 125: As noted elsewhere this will still push ALL the changes in the changeset between HASH...HEAD to the given repository.  Copying and deleting everything from both repositories.  Which isn't very helpful.

scruffian:
I'm not sure its worth pursuing this. It doesn't look like any other premium themes will end up in this repo anyway. For now I suggest we close this.

PR #5081, opened 2021-11-23
Videomaker: Update screenshot

kjellr:
@beafialho and I updated the video on the top of [the demo site](https://videomakerdemo.wordpress.com). This new screenshot reflects that change.


kjellr:
The old screenshot was double the size required by the theme directory, so this also supplies a much smaller one.

Issue #5001, opened 2021-11-05
Blockbase: Color schemes in customizer are broken


sublimese:
### Steps to replicate

1. Switch to Blockbase
2. Open the customizer
3. Select one of the non-default color schemes
4. Press publish
5. Notice an error message appears:  Looks like something’s gone wrong. Wait a couple seconds, and then try again.

## Result

Error message on screen:
Looks like something’s gone wrong. Wait a couple seconds, and then try again.

Error message in logs:
PHP Fatal error:  Uncaught TypeError: call_user_func_array(): Argument #1 ($function) must be a valid callback, non-static method GlobalStylesColorPalettes::sanitize_color_palette() cannot be called statically in C:\inetpub\wwwroot\someSite\wp-includes\class-wp-hook.php:303

## Expected

Color scheme change successfully

* User report: [shortlink to chat/forum thread/ticket, ie 12345-hc]
* Follow-up ticket: [shortlink to ticket, ie 12345-zen]

pbking:
I was not able to reproduce this issue (on /trunk as of today) in either .com or .org environment.  However while testing I did find and fix [something similar](https://github.com/Automattic/themes/pull/5014).  Not likely related but maybe worth mentioning.

sublimese:
I am happy to test something out if you would like.  Can also do screen grabs of each step that I am doing to get the issue.

scruffian:
> I am happy to test something out if you would like. Can also do screen grabs of each step that I am doing to get the issue.

Yes please!

sublimese:
WP Version:
![Screen Shot 2021-11-19 at 1 40 36 PM](https://user-images.githubusercontent.com/47400870/142689707-f473c671-9804-432a-bd06-77de6d91faa7.png)

Blockbase Version:
![Screen Shot 2021-11-19 at 1 40 48 PM](https://user-images.githubusercontent.com/47400870/142689743-c9272c12-b1d7-4860-9081-c8ea8a838de2.png)

Gutenberg (And all plugins)
![Screen Shot 2021-11-19 at 1 41 01 PM](https://user-images.githubusercontent.com/47400870/142689820-0f987f0d-6709-4469-b052-e74309361db1.png)

PHP Version:
![Screen Shot 2021-11-19 at 1 41 57 PM](https://user-images.githubusercontent.com/47400870/142689839-76b7981c-8a9a-4405-9f88-eaeedee5f56b.png)

Pages on Site:
[...]

PR #5082, opened 2021-11-23
Adding static keyword to avoid error in PHP 8

matiasbenedetto:
Adding `static` keyword to class method to avoid error in PHP 8.

Declaring class properties or methods as static makes them accessible without needing an instantiation of the class.

#### Related issue(s):
https://github.com/Automattic/themes/issues/5001


scruffian (approved):
LGTM

Issue #507, opened 2019-01-23
Independent Publisher 2: The extra margins above/below Vimeo video


hacchism:
## Steps to replicate
1. Activate "Independent Publisher 2"
2. Create a new post or page
3. Embed a Vimeo video with a URL or shortcode
4. Publish the post or the page

Please note that this is with Classic Editor (Not Gutenberg).

## Result

A Vimeo video shows up with unnecessary margins above and below.

![](http://cld.wthms.co/ynKTQK+)

* User report: [#1735249-zen](https://woothemes.zendesk.com/agent/tickets/1735249)
* User's site: https://johnjronline.wordpress.com/2018/03/14/strings-2015/

laurelfulford:
Thanks for this report, @hacchism!

This isn't actually a theme-specific problem, though it's more noticeable in some themes than others -- it can be reproduced very consistently on mobile with any of the themes in this repo. 

The issue is that Jetpack currently 'turns off' its own responsive video sizing when a theme support's Gutenberg's responsive embeds, because at the moment, the two clash. This in turn causes any embeds like videos, added through the classic block or editor, not to be responsive. 

More info here: https://github.com/Automattic/jetpack/issues/11097

I'm going to close this issue as a fix is currently being worked on in the Jetpack end of things; I'll add a note to the ZD ticket for follow up!

hacchism:
Thanks @laurelfulford. It makes sense. The user reported another issue for their Youtube video on this post: https://johnjronline.wordpress.com/2019/01/23/penny-dreadful-seasons-1-3/. And, it looks like the issue is also related to [https://github.com/Automattic/jetpack/issues/11097](https://github.com/Automattic/jetpack/issues/11097).

PR #5083, opened 2021-11-23
Videomaker: Fix "Film Poster" and "Background Video" patterns

kjellr:
Fixes these patterns as described in #507. 

"Background Video" should be pretty self-explanatory. For "Film Poster", I felt that it was simpler to just add the image versus adding the editable text w/laurels treatment from our other patterns. In this case, I think users will just swap that out with a different image, or delete it entirely. 

## Screenshots
Screen Shot 2021-11-23 at 3 22 59 PM
Screen Shot 2021-11-23 at 3 23 11 PM


scruffian (approved):
LGTM

PR #5090, opened 2021-11-24
Videomaker: Cleanup grid patterns, replace "I Love You" video

### LINKED ISSUE #5070: Videomaker: Patterns issues
## Background Video

I recommend we replace this with standard text instead, just like we did with the grid patterns. I think we can simplify it too, so it's more of a general-use pattern and less of a site-building one. 

- Change Post Title block to a H2 that says "Plenty of Fish In The Sea". 
- Change date and categories to static text that reads "A film about finding what you’re looking for."
- Remove the page header, navigation blocks, columns. Position text at the bottom left of the cover. 
- Ensure all text is white. 

[Gist of new markup for example.](https://gist.github.com/kjellr/53f8032aced47fe8ee5623a4370ee60c)
 
_Currently:_
Screen Shot 2021-11-22 at 3 42 49 PM

_Target:_
Screen Shot 2021-11-22 at 3 50 36 PM

## Film Poster

Missing image.

_Currently:_
Screen Shot 2021-11-22 at 3 43 21 PM

## Header, Description, and Grid of Projects

Can we adjust this one so that this last one never falls out of alignment because of the line break below? Might need to rebuild using multiple, stacked columns blocks.  

Screen Shot 2021-11-22 at 3 54 39 PM

## Grid of Projects

Can we swap "I love you" for another video that has the same aspect ratio as the rest? Currently, this doesn't line up correctly: 

Screen Shot 2021-11-22 at 3 55 03 PM

Alternatively, we could just remove that last row of videos for now, to match what's on the homepage. 
(Even if we do that though, this problem will be noticeable in the "Header, Description, and Grid of Projects" pattern above.)

---

cc @beafialho

kjellr:
Closes #5070.

- Updates the "I Love You" video to the new version. 
- Updates the "Grid" and "Header, Description, and Grid of Projects" patterns so that handle wrapping text more nicely. 

## Screenshots: 

Before|After
---|---
142934380-a07071f5-0fae-4f1d-bd41-31397e02cc98|Screen Shot 2021-11-24 at 9 31 29 AM

Before|After
---|---
|Screen Shot 2021-11-24 at 9 34 48 AM

### DIFF
--- a/videomaker/inc/patterns/header-description-and-grid-of-projects.php
+++ b/videomaker/inc/patterns/header-description-and-grid-of-projects.php
@@ -89,9 +89,9 @@ return array(
 		<h4>' . esc_html__( 'Plenty of Fish In The Sea', 'videomaker' ) . '</h4>
 		<!-- /wp:heading -->
 
-		<!-- wp:embed {"url":"https://videopress.com/v/e7QDnuzH","type":"video","providerNameSlug":"videopress","responsive":true,"className":"wp-embed-aspect-16-9 wp-has-aspect-ratio"} -->
+		<!-- wp:embed {"url":"https://videopress.com/v/TKDIpWEl","type":"video","providerNameSlug":"videopress","responsive":true,"className":"wp-embed-aspect-16-9 wp-has-aspect-ratio"} -->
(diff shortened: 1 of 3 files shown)


scruffian (approved):
LGTM

PR #5091, opened 2021-11-24
Videomaker: adjusting active states

### LINKED ISSUE #5069: Videomaker: Adjust active states for pagination and tags
I'm opening these as a single issue since they're pretty small. 

1. The active state for pagination is too prominent. If we can't make the white background happen _only_ behind the text, let's lose it entirely: 
active-state

2. The active state for tags leads to unreadable text. Can we turn the background of the pill shape to white of its active state to fix this? 
active-state-2

MaggieCabrera:

This PR fixes the active state in a few places for the theme.

The Post navigation blocks before:

Screenshot 2021-11-24 at 16 43 08

And after:

Screenshot 2021-11-24 at 16 42 17

The query pagination block was affected too but the markup is different and I couldn't conserve the active state without breaking the look of the block on its normal state so I added an exception for it on the active links rule. It now looks like this (we are inheriting the outline from Blockbase, I didn't want to touch that, is that ok?)

Screenshot 2021-11-24 at 16 47 46

The tag pills never had any issues for me, couldn't replicate the behavior on Safari or Firefox:

Screenshot 2021-11-24 at 16 49 36

#### Related issue(s):

Closes https://github.com/Automattic/themes/issues/5069

### DIFF
--- a/videomaker/sass/blocks/_post-navigation-link.scss
+++ b/videomaker/sass/blocks/_post-navigation-link.scss
@@ -1,6 +1,9 @@
 .wp-block-post-navigation-link {
 	flex-grow: 1;
-	border-top: 1px solid var(--wp--custom--color--primary);
+
+	&:not(:empty){
+		border-top: 1px solid var(--wp--custom--color--primary);
+	}
 
 	a {
 		display: inline-block;
--- a/videomaker/sass/blocks/_post-navigation-link.scss
+++ b/videomaker/sass/blocks/_post-navigation-link.scss
@@ -1,10 +1,10 @@
 .wp-block-post-navigation-link {
 	flex-grow: 1;
+	border-top: 1px solid var(--wp--custom--color--primary);
 
 	a {
-		border-top: 1px solid var(--wp--custom--color--primary);
-		display: block;
-		padding-top: 10px;
+		display: inline-block;
+		margin-top: 10px;
 	}
 }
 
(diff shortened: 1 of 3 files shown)


kjellr:
Screenshots look good! I can't reproduce the tag issue anymore either. 👍

videomaker/sass/blocks/_post-navigation-link.scss
```diff
@@ -1,10 +1,10 @@
 .wp-block-post-navigation-link {
 	flex-grow: 1;
+	border-top: 1px solid var(--wp--custom--color--primary);
```
mikachan: It looks like we get a border with no text (Next/Previous) on the first or last post when the border is on the container:

image

mikachan (commented):
This works great apart from the border on the first or last posts. I can't think of a good way around this, other than changing the markup of the pagination link block.

I also can't see the issue with the tags.

videomaker/sass/blocks/_post-navigation-link.scss
MaggieCabrera: aaaaah good catch, let me see how I can work around this

MaggieCabrera: `:empty` should catch it now!

mikachan: Yes awesome! Great idea, this works for me!

mikachan (approved):

PR #5096, opened 2021-11-24
Add remote theme

scruffian:
A skeleton for the Remote theme:
localhost_4759_ (1)


scruffian:
@kjellr how do you envisage the homepage content working? How would it be edited in the Classic dashboard? Also, the subscribe form will only work on dotcom - what should we do for dotorg?

kjellr:
> how do you envisage the homepage content working? How would it be edited in the Classic dashboard?

I'd say this should just be a `home.html` template, but the universal nature of this complicates things. 

I think our best bet for now would be to built this as a separate "Intro and Post List" page template, which folks can choose if they would like to use it (similar to how we treat the `page-large-header.html` template in Twenty Twenty-Two). Classic users can edit it using the Template Editor if they'd like. 

Either that, or we just bundle the intro + query as actual post content, and use headstart to assign it as a default static homepage on WP.com?

> Also, the subscribe form will only work on dotcom - what should we do for dotorg?

Yeah, I ran into this while creating the first version. The post terms block allows you to create a tag cloud of the categories, so let's try using that there instead: 

Block Pattern_ Categories

scruffian:
> The post terms block allows you to create a tag cloud of the categories, so let's try using that there instead:

I think this will only show the categories of the current page, not all categories for the site.

kjellr:
> I think this will only show the categories of the current page, not all categories for the site.

Oh sorry, not the Post Terms block. The Tag Cloud block has this functionality inside of the "Taxonomy" dropdown here:
 
Screen Shot 2021-11-30 at 10 12 57 AM

kjellr:
This is looking good for an initial commit. A few small notes: 

1. Can we make [the footer wide-aligned instead of full](https://cloudup.com/cKuEbZsVmW3)? 

2. The header [looks weird when there's a site logo](https://cloudup.com/c1l931SAvmu). We should match the Quadrat treatment here. 

3. We should add an `archive.html` template, and remove the Archive Title block from the `index.html` template to sync up with Twenty Twenty-Two. 

4. We should stop using "Post Content" inside of the `home.html` template. Hierarchy-wise, that template is used when there is no static homepage. It's not tied to a specific post or page, so there is no obvious post content to use there. In my testing it's currently pulling in some (seemingly) random post content from the main query. 

Let's do this instead: 

- Rename `home.html` to a custom template instead. We can name it something generic, like "Posts with Introduction" (`index-with-introduction.html`) maybe? If we do that, and then also adopt the `archive.html` and `index.html` treatment noted above, we don't really need a template called `home.html`. 
- Later on, when we create headstart content, we can use headstart to auto-assign a static homepage and map it to this new "Posts with Introduction" template.

Does that make sense?

PR #5100, opened 2021-11-25
Blockbase: Remove layout inheritance from the footer template

scruffian:
This removes the layout inheritance from the footer template part block. This allows themes to create full width footers without defining new templates.

I also removed the russell footer since it's identical to the Blockbase one.


mikachan (approved):
This looks good to me 🚢

PR #5101, opened 2021-11-25
Blockbase: Update heading font size

scruffian:
The "huge" size in Blockbase is 32px, so we can use the variable for the heading here.

This change will mean that child themes don't need to redeclare the font size for h2 elements, they will just need to change the font size definition.


mikachan (commented):
The change to theme.json looks good to me.

I think the changes from #5100 are included here too!

PR #5103, opened 2021-11-26
Blockbase: add a footer only template

MaggieCabrera:

This PR adds a footer only custom template to Blockbase, so that child themes like Paxton 2 don't need to [define it themselves](https://github.com/Automattic/themes/pull/5054).

After talking (p1637901915219800-slack-C029FM1EH) to @alaczek I think it could be a good idea to include this template directly in Blockbase, since multiple themes will make use of it.

An alternative to this is to alter the dotcom footer credits script to include the footer credits automatically on every page that doesn't have it but this has two problems: 

- The markup for the footer may not be correct for every theme, some may decide to change some things like padding, background color, etc and this brings inconsistency with the rest of the pages that do include a footer
- It makes the "blank" template not blank anymore. I'd love @kjellr input on this but as I see it the blank template is there to allow for full customization of a page and being forced to have a footer kind of defeats that purpose.

We could of course not include this in Blockbase and let the child themes define the footer every time. Keen to hear your thoughts!

### DIFF
--- a/blockbase/block-templates/footer-only.html
+++ b/blockbase/block-templates/footer-only.html
@@ -1,0 +1,3 @@
+<!-- wp:post-content {"layout":{"inherit":true}} /-->
+
+<!-- wp:template-part {"slug":"footer","tagName":"footer","layout":{"inherit":true},"className":"site-footer-container"} /--> 
--- a/blockbase/theme.json
+++ b/blockbase/theme.json
@@ -27,7 +27,15 @@
 				"page",
 				"post"
 			]
-		}
+		},
+        {
+            "name": "footer-only",
+            "title": "Footer Only",
+            "postTypes": [
+                "page",
+                "post"
+            ]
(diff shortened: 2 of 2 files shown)


scruffian (approved):
This looks like a perfect solution to the problem

kjellr:
> It makes the "blank" template not blank anymore. I'd love @kjellr input on this but as I see it the blank template is there to allow for full customization of a page and being forced to have a footer kind of defeats that purpose.

Just to confirm, does this _replace_ the blank template? Or just add an additional, "Footer-only" template?

MaggieCabrera:
> Just to confirm, does this _replace_ the blank template? Or just add an additional, "Footer-only" template?

it's an extra template, I am of the opinion that the blank template shouldn't go away

kjellr:
Ah ok cool. That works for me. 👍

PR #5104, opened 2021-11-26
Blockbase: Update paragraph gap

scruffian:
The intention of this CSS was, I believe, to prevent paragraphs inside the post content from from being affected by blockGap. The problem with it is that it breaks flex layout:

Add this to your footer:

```
<!-- wp:group {"layout":{"type":"flex"}} -->
<div class="wp-block-group">
	<!-- wp:paragraph {"align":"center","style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} -->
	<p class="has-text-align-center" style="font-size:var(--wp--custom--font-sizes--tiny);">Proudly Powered by <a href="https://wordpress.org" rel="nofollow">WordPress</a></p>
	<!-- /wp:paragraph -->
	<!-- wp:paragraph {"align":"center","style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} -->
	<p class="has-text-align-center" style="font-size:var(--wp--custom--font-sizes--tiny);">Proudly Powered by <a href="https://wordpress.org" rel="nofollow">WordPress</a></p>
	<!-- /wp:paragraph -->

</div>
<!-- /wp:group -->
```

You will see a margin above the second paragraph.

This change restricts that margin reset to only target the post content. This is still not ideal, as flex items inside post content will still have this bug, but at least the scope is more limited.


mikachan (approved):
This works for me as you described and I think it's a good idea to limit the scope to just the post content for now.

PR #5107, opened 2021-11-26
Update template and parts directories

### LINKED ISSUE #5105: Blockbase: Update template directories
## Steps to replicate
The template directory names for block themes have changed: https://github.com/WordPress/gutenberg/pull/36647

block-templates -> templates
block-template-parts -> parts

We should make this change in Blockbase and children.

matiasbenedetto:
Renaming folders:
block-templates -> templates
block-template-parts -> parts
Following the new Gutenberg convention: https://github.com/WordPress/gutenberg/pull/36647

#### Related issue(s):
https://github.com/Automattic/themes/issues/5105


MaggieCabrera:
Just a heads up, I think [git mv](https://git-scm.com/docs/git-mv) is cleaner for this (in terms of files changed)

matiasbenedetto:
Ohh, I never used `git mv`. :sweat_smile:   It seems there are different opinions about if it is the same as `mv` or not.
[oreilly](https://www.oreilly.com/library/view/version-control-with/9780596158187/ch05s06.html), [gitFaq](https://git.wiki.kernel.org/index.php/GitFaq#Why_does_Git_not_.22track.22_renames.3F), [StackOverflow](https://stackoverflow.com/questions/1094269/whats-the-purpose-of-git-mv). I will create a new PR to test if we can see any difference. :thinking:

matiasbenedetto:
The [other PR](https://github.com/Automattic/themes/pull/5108) using `git mv` instead of `mv`.

MaggieCabrera:
> Ohh, I never used `git mv`. 😅 It seems there are different opinions about if it is the same as `mv` or not. [oreilly](https://www.oreilly.com/library/view/version-control-with/9780596158187/ch05s06.html), [gitFaq](https://git.wiki.kernel.org/index.php/GitFaq#Why_does_Git_not_.22track.22_renames.3F), [StackOverflow](https://stackoverflow.com/questions/1094269/whats-the-purpose-of-git-mv). I will create a new PR to test if we can see any difference. 🤔

yeah looks like they are basically the same" I was hoping we would maintain the history for the moved files while using github's history page, but looks like it doesn't work either way, sorry I made you do the same thing twice!

scruffian:
Closing in favour of https://github.com/Automattic/themes/pull/5108

PR #5114, opened 2021-11-29
Move calendar styles to blockbase

scruffian:
The calendar styles in core aren't very well adapted to themes. This code connects them to theme colors for Quadrat, so we can bring this to Blockbase and all themes can benefit.

Before:
Screenshot 2021-11-29 at 13 42 09

After:
Screenshot 2021-11-29 at 13 41 48


pbking:
Another case where - in my opinion - the tertiary color of Blockbase is too light:

image

The change is quite good, especially for child themes. I just suggest a darker default tertiary color for Blockbase as we are finding more uses for it.

pbking:
It seems Geologist and Zoologist still have the calendar CSS.  (Are there others I missed checking?) Once that's removed this change should be good to go.  It kind of looks like that was started but I think not removed all the way and also not recompiled.

scruffian:
@Automattic/flow-patrol-manage this will need to happen to any themes based on Quadrat

PR #5115, opened 2021-11-29
Blockbase: Move table stripes to Blockbase

scruffian:
Table stripes in core have default color which isn't controlled by the theme. This takes the changes from Quadrat and applies them to Blockbase so that all Blockbase children can benefit.


quadrat/sass/blocks/_table.scss
```diff
@@ -32,4 +32,4 @@
 			background-color: var(--wp--preset--color--tertiary);
```
pbking: Looks like the stripy bit wasn't removed from Quadrat (or I think Zoologist and Geologist as there are still references to `.wp-block-table.is-style-stripes` in all of those three).

pbking:
The tertiary color value (FAFAFA) for default blockbase is VERY slight.
Are we sure that's the best color choice? (Can you pick out the one with a different color?)

image

scruffian:
@Automattic/flow-patrol-manage This will need to happen in any other themes that were generated from Quadrat...

PR #5116, opened 2021-11-29
Remove unused search block code

scruffian:
This code is no longer needed as we zero the border in blockbase:
Screenshot 2021-11-29 at 14 36 19

### DIFF
--- a/geologist/assets/theme.css
+++ b/geologist/assets/theme.css
@@ -253,10 +253,6 @@ ul ul {
 	text-align: unset;
 }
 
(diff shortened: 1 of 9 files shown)


mikachan (commented):
Looks good! Only thing I noticed is that it looks like the three affected themes need their theme.css rebuilding to include the change.

scruffian:
@Automattic/flow-patrol-manage this one will also need to happen in any themes derived from Quadrat

PR #5102, opened 2021-11-25
Blockbase: header refactor

MaggieCabrera:
#### Changes proposed in this Pull Request:

WIP

Before:

before

After:

after

Issue #5113, opened 2021-11-29
Block themes: remove styles from template parts


MaggieCabrera:
After https://github.com/WordPress/gutenberg/pull/36571 was merged, template parts no longer support spacing, layout and color options. We should refactor our themes in the same way TT2 is doing https://github.com/WordPress/twentytwentytwo/pull/262

Issue #5087, opened 2021-11-24
Blockbase: Add alternative Header template parts which span full width


jordesign:
## Steps to replicate

When editing a site with Blockbase - set a background color on the Header template part

## Result

The Header is not full width within the page - so the background remains contained with the content width.

![image (2)](https://user-images.githubusercontent.com/1842363/143231365-b85e3b71-9001-4f3e-bb9b-475e66cf6c59.png)

## Expected

I hoped/expected to be able to place a background on the Header template part - and have it be applied full width.

Could we consider adding other Header parts that users could choose from - that would allow this?

jordesign:
Pinging @kjellr  and @beafialho based on discussion in Slack.

kjellr:
As noted on slack, we should aim to build these header template parts in such a way that if the user applied a background, it'll "just work" and be full width.

supernovia:
Can we do the same with cover blocks? I can't think of many cases where people want cover blocks (whether in the template or not) to leave margins when they're full width.

MaggieCabrera:
> Can we do the same with cover blocks? I can't think of many cases where people want cover blocks (whether in the template or not) to leave margins when they're full width.

Cover blocks that are set to full width are supposed to go all the way to the edges of the viewport: pcrugM-lB-p2

supernovia:
Oh, you're right -- it is just when the cover block is in the header or footer that it's stuck with funny margins.

supernovia:
Would we also consider allowing the header to be wide width, while we're at it? Just in case someone needs that for their design.

supernovia:
(I've tried adding items to columns or groups and setting wide width on that, but can't)

scruffian:
Closing in favour of https://github.com/Automattic/themes/issues/5119

PR #5120, opened 2021-11-30
Blockbase: refactor header alignments

### LINKED ISSUE #5066: Blockbase: Editor alignment styles do not apply to site header
## Steps to replicate

- Activate Blockbase
- Edit the header block in the Editor by aligning the items to the right or center
- The items do not change alignment

https://user-images.githubusercontent.com/1645628/142636423-30240657-eb7c-4e8e-b519-5c951cf07409.mov

It looks like this is because of two styles in _header.scss:

`justify-content: start;` ([here](https://github.com/Automattic/themes/blob/cdb5883c518ce7bc282998ffb6d5e4a2786841fb/blockbase/sass/base/_header.scss#L2)) & `flex-grow: 1;` ([here](https://github.com/Automattic/themes/blob/cdb5883c518ce7bc282998ffb6d5e4a2786841fb/blockbase/sass/base/_header.scss#L7))

### LINKED ISSUE #5078: Quadrat: changing justify settings in Header -> Row block has no effect

MaggieCabrera:

This PR is an alternative to https://github.com/Automattic/themes/pull/5102. Props to @matiasbenedetto for helping out with this (sorry, I didn't credit you on the commit, I forgot!).

This PR aims to eliminate as much CSS from the header as possible, so that we can allow the user to modify the header in the site editor as they wish.

Screenshot 2021-11-30 at 14 24 33

Screenshot 2021-11-30 at 14 24 27

Screen Capture on 2021-11-30 at 14-28-04

#### Related issue(s):

Closes #5066, #5078

### DIFF
--- a/blockbase/block-template-parts/header.html
+++ b/blockbase/block-template-parts/header.html
@@ -1,14 +1,19 @@
-<!-- wp:group {"className":"site-header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"80px"}}}} -->
-<div class="wp-block-group site-header" style="padding-bottom:80px">
-	<!-- wp:site-logo /-->
+<!-- wp:group {"className":"site-header","layout":{"type":"flex","justifyContent":"space-between"},"style":{"spacing":{"padding":{"bottom":"80px","top":"var(--wp--custom--gap--vertical)"}}}} -->
+<div class="wp-block-group site-header" style="padding-bottom:80px;padding-top:var(--wp--custom--gap--vertical)">
 
-	<!-- wp:group -->
-	<div class="wp-block-group">
-		<!-- wp:site-title /-->
-		<!-- wp:site-tagline {"style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
+	<!-- wp:group {"className":"site-brand","layout":{"type":"flex"}} -->
+	<div class="wp-block-group site-brand">
+		<!-- wp:site-logo /-->
+
+		<!-- wp:group -->
+		<div class="wp-block-group">
+			<!-- wp:site-title /-->
+			<!-- wp:site-tagline {"style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
+		</div>
+		<!-- /wp:group -->
 	</div>
 	<!-- /wp:group -->
-
-	<!-- wp:navigation {"__unstableLocation":"primary","__unstableSocialLinks":"social","className":"is-style-blockbase-navigation-improved-responsive","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"}} /-->
+	
+	<!-- wp:navigation {"__unstableLocation":"primary","__unstableSocialLinks":"social","className":"is-style-blockbase-navigation-improved-responsive","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /-->
 </div>
 <!-- /wp:group -->
--- a/blockbase/sass/blocks/_navigation.scss
+++ b/blockbase/sass/blocks/_navigation.scss
@@ -22,6 +22,7 @@
 }
 
 .wp-block-navigation.is-style-blockbase-navigation-improved-responsive {
+	flex-grow: 1;
 	&.is-responsive .is-menu-open {
 		font-size: var(--wp--preset--font-size--medium);
 
--- a/blockbase/sass/blocks/_navigation.scss
+++ b/blockbase/sass/blocks/_navigation.scss
@@ -22,7 +22,6 @@
 }
 
 .wp-block-navigation.is-style-blockbase-navigation-improved-responsive {
-	flex-grow: 1;
 	&.is-responsive .is-menu-open {
 		font-size: var(--wp--preset--font-size--medium);
 
--- a/quadrat/block-template-parts/header.html
+++ b/quadrat/block-template-parts/header.html
@@ -17,4 +17,3 @@
 	<!-- wp:navigation {"__unstableLocation":"primary","__unstableSocialLinks":"social","className":"is-style-blockbase-navigation-improved-responsive","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /-->
 </div>
 <!-- /wp:group -->
-
--- a/quadrat/block-template-parts/header.html
+++ b/quadrat/block-template-parts/header.html
@@ -1,5 +1,5 @@
-<!-- wp:group {"className":"site-header","layout":{"type":"flex","justifyContent":"space-between"},"style":{"spacing":{"padding":{"bottom":"80px","top":"var(--wp--custom--gap--vertical)"}}}} -->
-<div class="wp-block-group site-header" style="padding-bottom:80px;padding-top:var(--wp--custom--gap--vertical)">
+<!-- wp:group {"className":"site-header","layout":{"type":"flex","justifyContent":"space-between"},"style":{"spacing":{"padding":{"bottom":"170px","top":"var(--wp--custom--gap--vertical)"}}}} -->
+<div class="wp-block-group site-header" style="padding-bottom:170px;padding-top:var(--wp--custom--gap--vertical)">
 
 	<!-- wp:group {"className":"site-brand","layout":{"type":"flex"}} -->
 	<div class="wp-block-group site-brand">
--- a/quadrat/block-template-parts/header.html
+++ b/quadrat/block-template-parts/header.html
@@ -1,8 +1,20 @@
-<!-- wp:group {"tagName":"header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"170px"}}},"className":"site-header"} -->
-<header class="wp-block-group site-header" style="padding-bottom:170px">
-	<!-- wp:site-logo /-->
-	<!-- wp:site-title /-->
-	<!-- wp:site-tagline {"style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
-	<!-- wp:navigation {"__unstableLocation":"primary","__unstableSocialLinks":"social","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"}} /-->
-</header>
+<!-- wp:group {"className":"site-header","layout":{"type":"flex","justifyContent":"space-between"},"style":{"spacing":{"padding":{"bottom":"80px","top":"var(--wp--custom--gap--vertical)"}}}} -->
+<div class="wp-block-group site-header" style="padding-bottom:80px;padding-top:var(--wp--custom--gap--vertical)">
+
+	<!-- wp:group {"className":"site-brand","layout":{"type":"flex"}} -->
+	<div class="wp-block-group site-brand">
+		<!-- wp:site-logo /-->
+
+		<!-- wp:group {"className":"site-words","layout":{"type":"flex"}} -->
+		<div class="wp-block-group site-words">
+			<!-- wp:site-title /-->
+			<!-- wp:site-tagline {"style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
+		</div>
+		<!-- /wp:group -->
+	</div>
+	<!-- /wp:group -->
+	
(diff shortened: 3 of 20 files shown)


MaggieCabrera:
This PR needs to address child themes too, but I'm keen to hear your reviews for Blockbase

blockbase/block-template-parts/header.html
```diff
@@ -1,14 +1,19 @@
-<!-- wp:group {"className":"site-header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"80px"}}}} -->
-<div class="wp-block-group site-header" style="padding-bottom:80px">
-	<!-- wp:site-logo /-->
+<!-- wp:group {"className":"site-header","layout":{"type":"flex","justifyContent":"space-between"},"style":{"spacing":{"padding":{"bottom":"80px","top":"var(--wp--custom--gap--vertical)"}}}} -->
+<div class="wp-block-group site-header" style="padding-bottom:80px;padding-top:var(--wp--custom--gap--vertical)">
```
pbking, line 2: I'm hesitant to use variables in this spot.  It doesn't translate well into the FSE.  That was (kind of) one advantage to that 'padding top' being in CSS as opposed to the header (variables make a lot more sense there...)

pbking:
The pass for Blockbase seems solid (except for the variable comment I made).

I see that at least Quadrat/Zoologist/Geologist (and I haven't yet evaluated all the other themes) are impacted by the change and should also be refactored to support the removal of CSS.

(Skatepark, Videomaker and  seems fine as does Seedlet-blocks and Mayland-blocks.)

blockbase/block-template-parts/header.html
scruffian: What if we had a fallback value?

mikachan (commented):
This works really well! It fixes all the issues I was seeing and removes some code. Awesome!

pbking:
This is ready for another look-see.  Quadrat + children have been refactored to align as expected.

blockbase/sass/blocks/_navigation.scss
```diff
@@ -22,6 +22,7 @@
 }
 
 .wp-block-navigation.is-style-blockbase-navigation-improved-responsive {
+	flex-grow: 1;
```
mikachan: This seems to break the ability to change the header row alignment from the editor:

https://user-images.githubusercontent.com/1645628/144227713-7254cf61-5ead-4266-a710-b8667a86b690.mov

If I remove `flex-grow: 1` then these alignments work, but I'm guessing this breaks something else?!

mikachan (commented):
The Quadrat and children changes are looking really good, other than the `flow-grow` comment.

It looks like we need to add some `padding-top` back to the headers of Mayland Blocks, Videomaker, Kerr, and Russell.

image

image

blockbase/sass/blocks/_navigation.scss
MaggieCabrera: yeah, the flex grow breaks the editor, I didn't refactor the block style for the navigation, it would need re testing

quadrat/block-template-parts/header.html
```diff
@@ -1,8 +1,19 @@
-<!-- wp:group {"tagName":"header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"170px"}}},"className":"site-header"} -->
-<header class="wp-block-group site-header" style="padding-bottom:170px">
-	<!-- wp:site-logo /-->
-	<!-- wp:site-title /-->
-	<!-- wp:site-tagline {"style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--tiny)"}}} /-->
-	<!-- wp:navigation {"__unstableLocation":"primary","__unstableSocialLinks":"social","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"}} /-->
-</header>
-<!-- /wp:group -->
+<!-- wp:group {"className":"site-header","layout":{"type":"flex","justifyContent":"space-between"},"style":{"spacing":{"padding":{"bottom":"80px","top":"var(--wp--custom--gap--vertical)"}}}} -->
```
MaggieCabrera: I think the padding bottom for Quadrat is 170px, else we could just let the theme inherit the template part from the parent

MaggieCabrera:
I'm thinking about this and I have a few points I'd like to discuss with both the devs and @beafialho and @kjellr:

- What's the difference between themes in terms of headers? Is it just the spacing below or can we keep the markup and let the children inherit the parent's header while keeping what's unique for each child?
- We need to refactor the [template parts](https://github.com/Automattic/themes/issues/5113) soon to remove any kind of styling, that means we need group wrappers for those styles too, can we absorb those theme specific changes with those wrappers and leave the template part the same for all themes (just inherit the one from the parent)?
- Can we remove the block style? it's kind of confusing and as discussed over slack, it's not a great user experience. I think this PR leaves the header in a good state in terms of editor experience and overall design on the frontend.

blockbase/sass/blocks/_navigation.scss
scruffian: flex-grow is needed so that the navigation aligns right when it wraps

MaggieCabrera: it's not needed anymore with this PR's approach, we wrap logo, site title and tagline in a group, use the editor's "space between" rule to align correctly without using flex-grow, and then use display contents on the wrapper we added for everything to work fine in mobile. It can be refactored, I just forgot we had the block style altogether

MaggieCabrera:
I think maybe the block style can be a header variation like TT2 supplies

pbking:
The only difference between Blockbase and Quadrat-family headers is that in blockbase the Title & Tagline are stacked and Quadrat (and derivatives) they are not.  Were we able to consolidate those designs into one then the gap is the only thing that would be different across themes.

We've discussed elsewhere being able to represent the different gaps in a better way. The problem is that if it's in the header then the header **has** to be customized per-theme.  If instead it's customized in the pages that consume the header (as they are now... mostly) then each PAGE has to express that opinion.  It's gotta be somewhere and we can't just use a variable.

blockbase/sass/blocks/_navigation.scss
scruffian: Without this rule the navigation looks like this when it wraps:
Screenshot 2021-12-01 at 18 05 03

blockbase/block-template-parts/header.html
pbking: I do want to note that NOT using a variable here means we LOSE the dynamic nature of the current value: `min(30px, 5vw)` 

Which is a sad loss.  So I'm kinda on-the-fence about it.

* Using it as-is means that while the site editor shows a BLANK value a user can still edit it.
* Using a variable in CSS means a user can't edit it.
* Using a pixel value shows that value to the user, but it doesn't get smaller on smaller screens.

None of those options are great.

blockbase/sass/blocks/_navigation.scss
pbking: After thinking about this a bit...

If that's the behavior of blocks in a row block when alignment is set to "space between" then I don't know that we should mess with it with the `flex-grow` attribute.

It looks funny (wrong) EITHER way when the menu wraps.  (I think when wrapped right-aligned looks MORE wrong.)  But looks isn't why I think we should do away with the `flex-grow` but rather the behavior.  If we don't like it then we should open a Gutenberg issue regarding the wrap behavior of rows using "space between" as it seems it wouldn't be necessarily specific to the header design.

quadrat/block-template-parts/header.html
pbking: There are more differences than that.  In this change the <group> wrapping the title and tagline is a row in Quadrat & Co.  

Thanks for catching the gap size though.  I have corrected it.

scruffian:
Another option would to be have a template part specifically for the gap...

blockbase/sass/blocks/_navigation.scss
scruffian: I agree this is the best solution. I think its unlikely anyone will notice.

blockbase/block-template-parts/header.html
MaggieCabrera: In the future the value won't be blank in the editor, but it will let the user know that the value is dynamic. It's a bit obscure but it's the way the editor is meant to work, so I think it's the best option for us too, since the editor will assume that this is a thing that themes will want to do in some places.

blockbase/sass/blocks/_navigation.scss
MaggieCabrera: yeah, sorry @scruffian I didn't know you meant about the flex-grow when wrapping (we used it initially to make the layout actually work, else the menu would always move to the left no matter what). I think I agree with Jason on this too.

MaggieCabrera:
> Another option would to be have a template part specifically for the gap...

like a "header container" kind of thing?

> The only difference between Blockbase and Quadrat-family headers is that in blockbase the Title & Tagline are stacked and Quadrat (and derivatives) they are not. 

I think that's a design decision and the reason we are providing that option as a block style, yes? I think we could offer that option as an additional header template part, same with what was suggested about having a [wide aligned option](https://github.com/Automattic/themes/issues/5087#issuecomment-984196791) which is apparently something we can't do at the moment with the current header markup

scruffian:
We'll need to make adjustments to all the other Blockbase children too

blockbase/block-template-parts/header.html
kjellr: Yeah, we're using variables this way in Twenty Twenty-Two, and I think it's ok as long as there's a fallback for when/if users switch themes.

kjellr:
> The only difference between Blockbase and Quadrat-family headers is that in blockbase the Title & Tagline are stacked and Quadrat (and derivatives) they are not.

That's generally right. In Videomaker, the header is wide-aligned instead of full-aligned. Skatepark does some weirder things which we have the option of stripping out since it hasn't launched yet. 

Ideally, all of these variants are available to any Blockbase theme as alternate template parts/patterns.

> Can we remove the block style? it's kind of confusing and as discussed over slack, it's not a great user experience. I think this PR leaves the header in a good state in terms of editor experience and overall design on the frontend.

Can you clarify which block style we're talking about here? (Sorry — I'm coming in and out of involvement with this segment of Blockbase so it's hard to keep it straight. 😄)

MaggieCabrera:
Thanks @kjellr !

> That's generally right. In Videomaker, the header is wide-aligned instead of full-aligned. Skatepark does some weirder things which we have the option of stripping out since it hasn't launched yet.
> 
> Ideally, all of these variants are available to any Blockbase theme as alternate template parts/patterns.

This is not the case yet, but I agree with you. We should make all of these different template parts in Blockbase.
I think we should align Skatepark with the rest of these.

> Can you clarify which block style we're talking about here? (Sorry — I'm coming in and out of involvement with this segment of Blockbase so it's hard to keep it straight. 😄)

We used a block style to change the appearance of the responsive menu on mobile when it's open. I think we should revisit this and make it an alternative template part instead of a block style. We used block styles so we wouldn't disturb already existing websites.

MaggieCabrera:
I've fixed the header for the rest of the themes, this is ready for another review. I'll try to isolate all our header variations on a different PR so we can simplify the child themes, I don't want to hold off this any longer

pbking:
Mayland blocks has a [bit about navigation block alignment](https://github.com/Automattic/themes/blob/26b1837176d9ed5f9357dc340ee3f1ec3898379f/mayland-blocks/sass/theme.scss#L15) I think we should also remove.

(edit) Nevermind on that.  The header template in Mayland depends on that; A better time to remove would be in the PR you mention.

pbking:
Looks like Videomaker's mobile layout isn't quite working out.  It was off prior to the change, is off after the change but for different reasons.  Suggest addressing it in the next PR.

Before:
image

After:
image

pbking (dismissed):
Progress but not quite done.  Followups noted.
Not sure where it falls with other related changes but I think these are good.
🚢

kjellr:
Aside from the Videomaker issue above, does this result in visual changes to any launched themes?

pbking:
> Aside from the Videomaker issue above, does this result in visual changes to any launched themes?

Aside from that there are no visible changes that I noted in my travels.

kjellr:
Awesome, thank you!

scruffian:
I don't think this is ready to go, we need to update the other blockbase children. The full list is:

Arbutus
Blockbase
Geologist
Mayland Blocks
Quadrat
Russell
Seedlet Blocks
Skatepark
Videomaker
Zoologist

scruffian:
@Automattic/flow-patrol-manage - this will need to happen in other Blockbase children too.

scruffian:
Closing in favour of #5145

Issue #5119, opened 2021-11-30
Blockbase & Related universal themes:  Header is not full width / alignment unchangeable


Addison-Stavlo:
In Blockbase and related child themes the header does not fill the width of the screen in-editor nor in the rendered front end. 
 MORE IMPORTANTLY, this style seems hardcoded and can not be edited by the user in the block editor.  This can easily be seen if we customize our header with a background color (or similarly with the recent removal of block specific styles for the template part; the background color of a group, row, or other internal block wrapping the entire content):

![Screen Shot 2021-11-30 at 6 56 12 AM](https://user-images.githubusercontent.com/28742426/144043091-376b39ae-d46a-4017-a5cb-37a58f31fbac.png)
![Screen Shot 2021-11-30 at 6 57 32 AM](https://user-images.githubusercontent.com/28742426/144043276-5ac64745-8d02-4c0d-9ed5-19e3e26d704c.png)

### Expected Behavior

Header width should not be hardcoded or restricted by CSS and should be customizable by the user.  If the design requires space on either side of the header, then the header block should be inside a group block which can be used to control this alignment on the template.

### Example with tt1-blocks.

The default header with a background color set is full width:
![Screen Shot 2021-11-30 at 7 04 07 AM](https://user-images.githubusercontent.com/28742426/144044116-4e43f7b2-0c71-4b33-91a6-a1793e0fd74d.png)

Wrapping the header in a group block can achieve this non-full width design of Blockbase, while still giving the user the ability to change this setting:

![Screen Shot 2021-11-30 at 7 05 43 AM](https://user-images.githubusercontent.com/28742426/144044315-42e48e12-8eab-4fc7-a2e2-8d9adc697de0.png)

MaggieCabrera:
Related to https://github.com/Automattic/themes/issues/5087 and https://github.com/Automattic/themes/issues/5078

Addison-Stavlo:
> Related to #5087 and #5078

Thanks @MaggieCabrera !  
To elaborate a bit:  

>  #5087

I think this problem is less about adding more template parts, and more about using blocks and block settings to achieve the initial styles we want where appropriate over any hard coded CSS (I believe [this issue](https://github.com/Automattic/themes/issues/5118) may have a similar cause as well).  This alignment issue should be resolvable for everyones needs without relying on additional template parts.

> #5078

May be another symptom of this cause.

matiasbenedetto:
[...]

PR #5121, opened 2021-11-30
Header: making it full width

### LINKED ISSUE #5087: Blockbase: Add alternative Header template parts which span full width

matiasbenedetto:
- Making the header full width.

## Demo:
Before:
image

After:
image

#### Related issue(s):
https://github.com/Automattic/themes/issues/5087
https://github.com/Automattic/themes/issues/5119

### DIFF
--- a/blockbase/block-template-parts/header.html
+++ b/blockbase/block-template-parts/header.html
@@ -1,5 +1,5 @@
 <!-- wp:group {"className":"site-header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"80px"}}}} -->
-<div class="wp-block-group site-header" style="padding-bottom:80px">
+<div class="wp-block-group site-header alignfull">
 	<!-- wp:site-logo /-->
 
 	<!-- wp:group -->
--- a/blockbase/block-template-parts/header.html
+++ b/blockbase/block-template-parts/header.html
@@ -1,5 +1,5 @@
-<!-- wp:group {"className":"site-header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"80px"}}}} -->
-<div class="wp-block-group site-header alignfull">
+<!-- wp:group {"className":"site-header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"40px","top":"40px", "left":"30px", "right":"30px"},"margin":{"bottom":"40px", "left":"-30px", "right":"-30px"}}}} -->
+<div class="wp-block-group site-header" style="padding-bottom:40px; padding-top:40px; padding-left: 30px; padding-right: 30px; margin-bottom:40px; margin-left: -30px; margin-right: -30px;">
 	<!-- wp:site-logo /-->
 
 	<!-- wp:group -->
--- a/blockbase/sass/base/_header.scss
+++ b/blockbase/sass/base/_header.scss
@@ -1,10 +1,6 @@
 .wp-site-blocks .site-header {
-	display: flex;
 	justify-content: start;
 	overflow: inherit;
-	padding-top: var(--wp--custom--gap--vertical);
-	padding-bottom: var(--wp--custom--gap--vertical);
-	margin-bottom: var(--wp--custom--gap--vertical);
 
 	.wp-block-navigation {
 		flex-grow: 1;
--- a/blockbase/sass/base/_header.scss
+++ b/blockbase/sass/base/_header.scss
@@ -1,7 +1,10 @@
 .wp-site-blocks .site-header {
+	display: flex;
 	justify-content: start;
 	overflow: inherit;
 	padding-top: var(--wp--custom--gap--vertical);
+	padding-bottom: var(--wp--custom--gap--vertical);
+	margin-bottom: var(--wp--custom--gap--vertical);
 
 	.wp-block-navigation {
 		flex-grow: 1;
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -138,9 +138,12 @@ pre {
 }
 
 .wp-site-blocks .site-header {
+	display: flex;
 	justify-content: start;
 	overflow: inherit;
 	padding-top: var(--wp--custom--gap--vertical);
+	padding-bottom: var(--wp--custom--gap--vertical);
+	margin-bottom: var(--wp--custom--gap--vertical);
 }
 
 .wp-site-blocks .site-header .wp-block-navigation {
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -138,12 +138,8 @@ pre {
 }
 
 .wp-site-blocks .site-header {
-	display: flex;
 	justify-content: start;
 	overflow: inherit;
-	padding-top: var(--wp--custom--gap--vertical);
-	padding-bottom: var(--wp--custom--gap--vertical);
-	margin-bottom: var(--wp--custom--gap--vertical);
 }
 
 .wp-site-blocks .site-header .wp-block-navigation {


blockbase/block-template-parts/header.html
```diff
@@ -1,5 +1,5 @@
 <!-- wp:group {"className":"site-header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"80px"}}}} -->
-<div class="wp-block-group site-header" style="padding-bottom:80px">
+<div class="wp-block-group site-header alignfull">
```
pbking: The Block markup shows an 80px bottom padding but that was removed here.  It should either be removed in both places or kept in both places otherwise the block editor will complain.

We've been making a push elsewhere to keep the layout details in the templates (allowing the user to have more control of that) so I believe keeping the 80px here is more appropriate than moving it into the _header.scss file.  (I believe the padding-top is actually moved FROM the _header.scss to the template in another change I recently reviewed).

blockbase/sass/base/_header.scss
```diff
@@ -1,7 +1,10 @@
 .wp-site-blocks .site-header {
+	display: flex;
```
pbking: `display:flex` shouldn't be necessary here since the layout type 'flex' is assigned in the template.

blockbase/block-template-parts/header.html
MaggieCabrera: I don't think adding alignfull only as a class is the correct markup. In fact I don't think flex group blocks allow for alignments but this is working on my local site. It's weird.

+1 on Jason's comments about the padding. I think you did it because it looks bad when we have a colored background, make it a margin instead, so we maintain the same spacing as we originally had

blockbase/sass/base/_header.scss
```diff
@@ -1,7 +1,10 @@
...
 	justify-content: start;
 	overflow: inherit;
 	padding-top: var(--wp--custom--gap--vertical);
+	padding-bottom: var(--wp--custom--gap--vertical);
+	margin-bottom: var(--wp--custom--gap--vertical);
```
Addison-Stavlo: In general, shouldn't all of these styles be accomplished via blocks and their style attributes?

Hardcoding CSS into these sections results in a broken FSE experience, where block controls surfaced to the user in the editor are overridden and/or have no effect when selected as a result.

blockbase/block-template-parts/header.html
matiasbenedetto: > The Block markup shows an 80px bottom padding but that was removed here. It should either be removed in both places or kept in both places otherwise the block editor will complain.

@pbking I wasn't aware of that. I will fix it in the next commit.

> I don't think adding alignfull only as a class is the correct markup. In fact I don't think flex group blocks allow for alignments but this is working on my local site. It's weird.

@MaggieCabrera `alignfull` is just about margin and padding. It's not related to flex justification.

blockbase/sass/base/_header.scss
matiasbenedetto: :+1: Removing it in the next commit.

blockbase/sass/base/_header.scss
matiasbenedetto: :+1: Thanks for the input. I tried a different approach in the latest commit.

matiasbenedetto:
I refactored the code following your suggestions. The final output is the same but the `padding` dimensions are editable via site editor. `margin` dimensions are also working but don't appear in the site editor UI.


image

blockbase/block-template-parts/header.html
```diff
@@ -1,5 +1,5 @@
-<!-- wp:group {"className":"site-header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"80px"}}}} -->
-<div class="wp-block-group site-header" style="padding-bottom:80px">
```
pbking, line 2: The 'vertical' gap is defined as `"vertical": "min(30px, 5vw)"` (the dynamic-ness being something I'm sad to lose).  

But that means the top gap should probably default to 30px rahter than 40px as defined here.

blockbase/block-template-parts/header.html
```diff
@@ -1,5 +1,5 @@
...
+<!-- wp:group {"className":"site-header","layout":{"type":"flex"},"style":{"spacing":{"padding":{"bottom":"40px","top":"40px", "left":"30px", "right":"30px"},"margin":{"bottom":"40px", "left":"-30px", "right":"-30px"}}}} -->
+<div class="wp-block-group site-header" style="padding-bottom:40px; padding-top:40px; padding-left: 30px; padding-right: 30px; margin-bottom:40px; margin-left: -30px; margin-right: -30px;">
```
pbking, line 2: The 'vertical' gap is defined as "vertical": "min(30px, 5vw)" (the dynamic-ness being something I'm sad to lose).

But that means the top gap should probably default to 30px rahter than 40px as defined here.

pbking:
I'm not terribly comfortable with the negative margins used in the template in this change.  Better to leverage the negative margins defined by 'alignfull' which the entire system can know about.  I understand @MaggieCabrera 's hesitency to use that incorrectly though as that still puts a user in the same state of being unable to modify that property value.

I think we should consider yet another wrapper to achieve this.  Something along the lines of this:

```
<div class="wp-block-group alignfull">

	<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"bottom":"40px","top":"30px"}}},"className":"site-header","layout":{"type":"flex"}} -->
	<div class="wp-block-group alignfull site-header" style="padding-top:30px;padding-bottom:40px">
		
	<!-- wp:site-logo /-->
	
	<!-- wp:group -->
	<div class="wp-block-group">
		<!-- wp:site-title /-->
	</div>
	<!-- /wp:group -->
	
	<!-- wp:navigation {"__unstableLocation":"primary","className":"is-style-blockbase-navigation-improved-responsive","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"}} /--></div>
	<!-- /wp:group -->
</div>
<!-- /wp:group -->
```

image

This way using `alignwide` won't be wrong... it will actually be in a group where the usage of that is proper.

This does introduce other unfortunate side effects in the responsive menu, but it's more "the block way" than "the CSS way" and reflects a design a user could both create themselves using the FSE exclusively and make it more editable.

blockbase/sass/base/_header.scss
scruffian: @Addison-Stavlo ideally yes, but there are things we need to do that aren't possible with the current block controls and attributes. In this case we need the site logo to appear in a different position in smaller viewports - Gutenberg doesn't yet offer us any controls for responsive layouts, so this kind of thing is not yet possible. As soon as we have the tools we need we will remove all the CSS.

blockbase/block-template-parts/header.html
MaggieCabrera: `alignfull` is a GB class assigned to blocks that are aligned with the "Full width" option. The class works here because GB loads the related CSS but it's not the correct way of applying it.

MaggieCabrera:
> I'm not terribly comfortable with the negative margins used in the template in this change. Better to leverage the negative margins defined by 'alignfull' which the entire system can know about. I understand @MaggieCabrera 's hesitency to use that incorrectly though as that still puts a user in the same state of being unable to modify that property value.
> 

I agree with Jason's proposal, which also solves the "wide" problem which is very valid. I think that in combination with what's being done over at https://github.com/Automattic/themes/pull/5120 that fixes the responsive issues would be the way to go for this.

It's hard working on this on two separate PRs so I'm happy if you want to hijack what I did there and add it to this PR and we abandon that one or we fix the alignment first then follow up with the justification issues building on top of this.

matiasbenedetto:
> This way using alignwide won't be wrong... it will actually be in a group where the usage of that is proper.

> I agree with Jason's proposal, which also solves the "wide" problem which is very valid.

Ok, if you are ok with that I can revert the latest commit and work on the initial proposal of this PR.

> It's hard working on this on two separate PRs so I'm happy if you want to hijack what I did there and add it to this PR and we abandon that one or we fix the alignment first then follow up with the justification issues building on top of this.

Ok, we can put this on hold until #5120 is merged.

MaggieCabrera:
> Ok, we can put this on hold until #5120 is merged.

I think it makes more sense to land this one first, it's a bit simpler and I can't handle #5120 myself today, so keep at it and then I'll rebase the other PR

PR #5122, opened 2021-11-30
Blockbase: Add padding to footer

### LINKED ISSUE #5084: Blockbase: Footer has no margin
At some point the footer of Blockbase lost it's padding/margin rules and now is stuck at the bottom of the page:

Screenshot 2021-11-24 at 12 10 20

mikachan:
This PR adds padding to the group block around the footer. I've used fixed values of 80px on the top and 30px on the bottom, which matches the header spacing at large resolutions. I tried to keep it simple by using fixed values, but maybe this is too simple!

#### Related issue(s):
Closes #5084


danieldudzic (approved):
Works fine 👌. I have checked to see how other themes handle this, and they don't seem to be using anything fancy.

PR #5127, opened 2021-12-01
Blockbase: Add an archive template

### LINKED ISSUE #5111: Blockbase + family: Add an archive.html
## Steps to replicate

In TT2 we have an archive.html template, which means we don't need the archive title block. We should do the same thing in Blockbase and its children.

scruffian:
This adds an archive template to Blockbase so that we don't have the archive title block on the index template.

#### Related issue(s):
Fixes #5111

### DIFF
--- a/arbutus/block-templates/archive.html
+++ b/arbutus/block-templates/archive.html
@@ -2,6 +2,7 @@
 
 <!-- wp:query {"tagName":"main","layout":{"inherit":true}} -->
 <main class="wp-block-query">
+<!-- wp:query-title {"type":"archive","style":{"spacing":{"margin":{"bottom":"100px"}}}} /-->
 <!-- wp:post-template -->
 <!-- wp:group -->
 <div class="wp-block-group">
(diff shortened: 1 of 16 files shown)


scruffian:
@Automattic/flow-patrol-manage this will need to happen in any other themes you're working on.

pbking:
Is the reason an `archive.html` is added to every theme due to the difference in gap between header and content?  I don't see any other differences.

Were it not for that the `archive.html` template could be added exclusively to Blockbase and not the child themes right?

scruffian:
There are other difference - the index template in most themes is different to the one in Blockbase. We could consider creating a template part and sharing it between both the index and the archive.

pbking:
Compared to the 3 options of:
* Having a mostly copy of index.html
* building a template part and sharing that between two things
* using a block

I'm not sure I see the downside of using the query-title block.  Is the design result the same either way?

(edit). I guess that's the point though, if we remove it from the template a user doesn't see it in the FSE and seeing it in the FSE is confusing.  I take a while to get there but I get there sometimes.

This seems to work as advertised.  I'm good with it as-is without optimizing with template parts, etc.

pbking (approved):

PR #5138, opened 2021-12-03
Adding table styles to tables added inside a widget area

### LINKED ISSUE #5124: Textbook theme: table styling not applied to tables in widget area
### Quick summary

When adding a table to the widget area in this theme, the styling the theme uses for tables in the content area is not applied, even though the widget preview shows the table correctly styled.

### Steps to reproduce

1. Add a table with a few rows of content to a post or page using the Default style. Also toggle the header and footer row settings on. 
2. Add a table to the sidebar, using identical settings
3. View the post

Variation 1:

As above, but compare with and without a background color set for the table.

Variation 2:

1. As above, but select the Stripes style.
2. Compare with and without a background color set.

For the table in the post content and the table in the sidebar to be styled the same, and for the background color to be applied to the table.

#### Scenario 1: Default style and background color set

The table in the sidebar doesn't show any border or styling for alternating lines like the one in the post or the widget preview.

The table in the post and widget preview shows the background color only for the footer row, while the table in the sidebar shows the background for the entire table.

Customize__Shift_to_the_left_–_This_is_the_tagline

#### Scenario 2: Default style and no background color set

The table in the sidebar doesn't show any border or styling for alternating lines like the one in the post or widget preview

Customize__Shift_to_the_left_–_This_is_the_tagline

#### Scenario 3: Stripes style and background color set

The table in the sidebar shows the striped styling as expected, while the table in the post and widget preview doesn't show any stripes, and the background color only in the footer row.

Customize__Shift_to_the_left_–_This_is_the_tagline

#### Scenario 4: Stripes style and no background color set

The table in the sidebar shows no styling at all, while the table in the post and widget preview only has the footer row styled differently.

Customize__Shift_to_the_left_–_This_is_the_tagline

### Context

Reported in the support forums at https://wordpress.com/forums/topic/boundary-line-not-showing-in-table/?view=all

### Operating System

_No response_

### Browser

_No response_

### Simple, Atomic or both?

_No response_

### Theme-specific issue?

Textbook

### Other notes

It looks like the styling for the table in the post content is coming from this CSS:

`.entry-content table thead tr {
 font-size:12.8px;
 font-size:.8rem;
 font-weight:700;
 letter-spacing:.075em;
 line-height:1.2375em;
 text-transform:uppercase;
 background:#222;
 color:#fff
}`

It's specifically targeting `.entry-content`, so that CSS is not being applied to tables appearing outside the content area, for example, in widgets.

### Reproducibility

Consistent

### Severity

_No response_

### Available workarounds?

_No response_

### Workaround details

It should be possible to add the necessary styling to the sidebar table using CSS, but it's not an option for a free user like the one who reported the issue.

KokkieH:
Currently the theme's styling for tables are only applied to tables appearing in post/page content, with the `.entry-content` class applied. As a result, a table added to the widget area is unstyled.

This pull request modifies the `style.css` file so the styles available for tables get applied to tables with the `.widget-area` class as well as `.entry-content`

**Before:**

Textbook before

**After:**

Textbook after

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/5124

### DIFF
--- a/textbook/style.css
+++ b/textbook/style.css
@@ -662,15 +662,18 @@ table td {
   text-align: left;
 }
 
-.entry-content table {
+.entry-content table,
+.widget-area table {
   border-collapse: collapse;
 }
 
-.entry-content table tbody tr:nth-child(odd) {
+.entry-content table tbody tr:nth-child(odd),
+.widget-area table tbody tr:nth-child(odd) {
   background: #FFF;
 }
 
-.entry-content table tbody {
+.entry-content table tbody
+.widget-area table tbody {
   background: #FFF;
   border: 1px solid #ce4639;
   -webkit-border-bottom-left-radius: 0.5em;
@@ -682,16 +685,20 @@ table td {
   overflow: hidden;
 }
 
-.entry-content table tbody tr:nth-child(even) {
+.entry-content table tbody tr:nth-child(even),
+.widget-area table tbody tr:nth-child(even) {
   background: #f0eeec;
 }
 
 .entry-content table th,
-.entry-content table td {
+.entry-content table td,
+.widget-area table th,
+.widget-area table td {
   border: 1px solid #ce4639;
 }
 
-.entry-content table thead tr {
+.entry-content table thead tr,
+.widget-area table thead tr {
   font-size: 12.8px;
   font-size: 0.8rem;
   font-weight: 700;
--- a/textbook/style.css
+++ b/textbook/style.css
@@ -668,7 +668,8 @@ table td {
 }
 
 .entry-content table tbody tr:nth-child(odd),
-.widget-area table tbody tr:nth-child(odd) {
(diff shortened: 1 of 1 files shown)


KokkieH:
The table footer row doesn't have an explicit background set, but shows the background of the parent element. For `.entry-content` this is white, and for `.widget-area` this is grey. https://github.com/Automattic/themes/pull/5138/commits/bedfc4e879399c3bb702e5ecea9e16bc63189a60 fixes this.

Then I noticed the white background I added to the table bleeds through along the left and top edges of the table if the Stripes style is used. If explicitly setting a custom background color for the table, that likewise bleeds through, in both the widget and content areas. https://github.com/Automattic/themes/pull/5138/commits/567a53d00321f765fc80f73f3db7a33ddc37fbda fixes this.

Textbook

PR #5143, opened 2021-12-03
Geologist: Remove unused template

scruffian:
I don't think this template is used so we should remove it.


MaggieCabrera (approved):
Yesss, thank you

PR #5144, opened 2021-12-03
Add default layout to Geologist single posts

scruffian:
Somehow the post titles are being allowed to stretch full width. Before:
Screenshot 2021-12-03 at 16 54 12

After:
Screenshot 2021-12-03 at 16 53 30

@kjellr I assume this isn't intentional

### DIFF
--- a/geologist/block-templates/single.html
+++ b/geologist/block-templates/single.html
@@ -3,7 +3,7 @@
 <!-- wp:group {"tagName":"main"} -->
 <main class="wp-block-group">
 
-	<!-- wp:group -->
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
https://github.com/Automattic/themes/pull/4720 the change is coming from here, but it looks like it might be an error

kjellr (approved):
> @kjellr I assume this isn't intentional

Yeah, that doesn't seem right to me. It should align with the post content. 

I double checked, and this same part of the `page.html` template has layout enabled already. So I think this makes sense 👍

Issue #4965, opened 2021-11-02
Skatepark: Missing space between the social nav and the menu on mobile


kjellr:
~40px is fine. 

Current|Mockup
---|---
<img width="319" alt="Screen Shot 2021-11-02 at 3 35 46 PM" src="https://user-images.githubusercontent.com/1202812/139940741-6abab1ff-b922-42bd-b6a5-6247f6c3dd9e.png">|<img width="319" alt="Screen Shot 2021-11-02 at 3 37 25 PM" src="https://user-images.githubusercontent.com/1202812/139940743-8c4fe8b6-0223-431e-9a23-9951d25315ba.png">

Issue #4964, opened 2021-11-02
Skatepark: Vertical interior spacing in the header is incorrect


kjellr:
I feel like we've already fixed this once or twice. There should be more spacing between the site title and tagline, and also between the menu and the social nav on desktop. 

Please reference the design comps, as those have extensive mockups for how the header should appear. 

Current Dev|Mockup
---|---
<img width="1287" alt="Screen Shot 2021-11-02 at 3 33 12 PM" src="https://user-images.githubusercontent.com/1202812/139940090-942282fe-07db-44b6-b3ee-71359ce6beb3.png">|![Home](https://user-images.githubusercontent.com/1202812/139940021-45b4f24a-cf52-420c-9a5a-63c2dd1c3aa3.jpg)

Issue #4591, opened 2021-09-14
Skatepark: Mobile spacing around social menu is incorrect


kjellr:
Current|Intended
---|---
![current](https://user-images.githubusercontent.com/1202812/133313655-be11e739-f06f-4bc7-b759-4c93a8782255.png)|![iPhone 8 - 19](https://user-images.githubusercontent.com/1202812/133313665-9eb11ee1-b71c-4ee5-8388-74d11986ce5f.png)

Intended margins: 

![iPhone 8 - 19](https://user-images.githubusercontent.com/1202812/133313976-18cf17b4-3c50-4b8c-b198-8279292b6f0b.png)

danieldudzic:
I don't want to add a CSS fix, as this calls for a `theme.json` setting. Hoever, I can't find one. I'm not even sure how to target the Social Icons menu. It's not the `Social Links` block, right?

kjellr:
This is still an issue — the spacing below the tagline is incorrect even when the social menu is inactive:

Current|Design Comp
---|---
<img width="373" alt="Screen Shot 2021-11-02 at 3 24 37 PM" src="https://user-images.githubusercontent.com/1202812/139935371-f53a8e33-ac44-4966-826d-0c93a946d86e.png">|![iPhone 8 - 22](https://user-images.githubusercontent.com/1202812/139937498-2a4bc33f-9bcf-4a88-9c04-87a97fde7bfc.png)

PR #5145, opened 2021-12-03
Try/consolidate headers

### LINKED ISSUE #5087: Blockbase: Add alternative Header template parts which span full width

pbking:
Started from https://github.com/Automattic/themes/pull/5120 and encorporating changes and discussion from https://github.com/Automattic/themes/pull/5121.

The task of this change is to consolidate header styles designed for child themes into Blockbase which child themes can use as header/footer options.  (Or maybe introduce those options in the children, regardless they should be alternative template parts, not replacements).

Closes #5087 #5119 #5066 #5166


pbking:
Example of the goal with Quadrat's header style:

image

geologist/sass/templates/_header.scss
```diff
@@ -0,0 +1,9 @@
+@include break-small-only {
```
scruffian: should this live in blockbase?

blockbase/theme.json
```diff
@@ -247,7 +251,8 @@
 			"gap": {
 				"baseline": "15px",
 				"horizontal": "min(30px, 5vw)",
-				"vertical": "min(30px, 5vw)"
+				"vertical": "min(30px, 5vw)",
+				"header": "50px"
```
scruffian: I seem to remember proposing this some time ago!

pbking:
I THINK this is ready for a solid shake-down.

### Template Part Alternatives

All header options (with the exception of Skatepark's) have been rolled up into Blockbase as alternate header template parts.  The templates in the child themes have been changed to load those alternate templates by default.  'header-linear' is Quadrat's header used by Geologist and Zoologist.  'header-arbutus' is uh... creatively named and only used by Arbututs.  'header-centered' is Seedlet-Block's and 'header-minimal' is Videomaker's.

There is MINIMAL (but still some) CSS remaining for the header.  It's responsible for modifying the layout and position of the site logo when in 'mobile' view as well as stacking title/tag in mobile view for the 'header-linear'

image

Skatepark is the exception to this.  The Theme's header template and the CSS that went along with it were too large and too specific to be able to be refactored easily.  Instead the header is still offered as an alternate template (header-ollie), but it's provided by Skatepark rather than Blockbase.  This COULD be refactored to simplify the header and extract the CSS making it available for other themes but not for this pass.

image

### Header Spacing

All of the template header templates include a new "header-spacer" template.  The default template provided by Blockbase is simply a 50px spacer block.  Adjusting the spacer in that template part changes the gap across the site.  This has the benefits of a css variable in utility (define once, use many places) but is conveniently able to be modified by the user.

Where necessary themes that defined that gap in page templates were refactored to instead expect that space in the header.

image

### Full Width Background Options

Each of the headers were (if necessary) refactored to allow a full-width group to have a background color.  This was done with standard block layout techniques rather than CSS tricks to ensure that users of the FSE are able to customize the header completely.

image

image


### Site Logo Sizing

CSS dictating the size of the theme's logo has been removed and replaced with block attributes.  Unfortunately our themes have been dictating a HEIGHT while only a WIDTH option is available on the site-logo block.  

image

### Get Rid of Navigation Block Style

The alternate block style for navigation that cleans up the responsive menu was removed and is just styling provided by blockbase by default.

geologist/sass/templates/_header.scss
pbking: It does indeed.

blockbase/block-template-parts/header-arbutus.html
```diff
@@ -0,0 +1,27 @@
+<!-- wp:group {"className":"gapless-group","layout":{"inherit":"true"}} -->
```
scruffian: I think we can achieve the differences in this file using theme.json

scruffian:
I noticed a slight issue in the Videomaker header (it was on trunk too) so I've used the same classes in the header-minimal template to achieve the design intention.

scruffian:
I also fixed some spacing issues in the Seedlet header:
Screenshot 2021-12-08 at 09 56 14

scruffian:
I noticed that the navigation was off in Quadrat/zoologist/Geologist, so I pushed a fix for that too.

scruffian:
I made some simplifications to the arbutus header to make it more inline with the other themes. Unfortunately there's no way to specify a block style in theme.json so it will need to be a different template for now. I've renamed it to `header-rounded-logo` so that the difference is more obvious.

scruffian:
I think this looks good to merge now, but I'd like another review from someone...

blockbase/block-template-parts/header-arbutus.html
pbking: In general we want a gap between items, we only want SOME groups to be "gapless".  I'm not sure how that could be achieved in theme.json.  

Alternately we could just leave the gap and adjust the size of the spacer to be it's current size minus the gap size.  The only problem with that was that it looked a little strange as an adjusting mechanism with the gap there, and also there are a couple of themes where the "header gap" is smaller than the "block gap" and the design target couldn't be achieved.  If it's OK making those gaps slightly larger than they are now then we could do away with the concept entirely.

pbking:
> I noticed that the navigation was off in Quadrat/zoologist/Geologist, so I pushed a fix for that too.

The [changes](https://github.com/Automattic/themes/pull/5145/commits/2fba0189cad562ee484aabf4040e5f8b271002c8) were put into those themes rather than blockbase. I believe those should go in Blockbase so that anything that header option gets the navigation fixes.

Also the responsive menus in all themes seem a little wonkey, especially when there are social icons.

scruffian:
> I believe those should go in Blockbase so that anything that header option gets the navigation fixes.

The navigation for these themes is right aligned, not left, so there does need to be something different in the theme somehow. I think the navigation needs greatly simplifying, but I think that needs to be a followup PR as this one is so big already. My aim was just to fix the visual regression so we could ship this :)

pbking:
Adjusted the blockbase (default) navigation RE: Social icons:

Before: (also happening on trunk)

OR


After: 

OR

pbking:
Quradrat still has some visual problems and these are also happening on trunk.  Those didn't seem to get all-the-way resolved.

pbking:
After: 


Also those styles are now a part of Blockbase assigned to the `.blockbase-responsive-navigation-linear` class which is assigned to the navigation block in the `header-linear` header option.  Thus anything that leverages that header template part will also get the matching responsive styling.  

(Zoologist)


The same has been done for videomaker.

pbking:
> The navigation for these themes is right aligned, not left, so there does need to be something different in the theme somehow. I think the navigation needs greatly simplifying, but I think that needs to be a followup PR as this one is so big already.

This should be visually correct now and accessible to any Blockbase theme.  It could still definitely be simplified.  What's been refactored into Blockbase is pretty much whole-hog what came from Quadrat, but it's not too heavy.  I think it could be fine as-is until Gutenberg better offers some kind of responsive styling options.

I think this is ready for (one more?) look?

blockbase/sass/base/_layout.scss
```diff
@@ -21,3 +21,10 @@ body.admin-bar {
 		min-height: calc( 100vh - var(--wpadmin-bar--height) );
 	}
 }
+
+// Provided to allow a group to prevent the standard gap from being used between elements
+.gapless-group {
+	> * {
+		margin-top: 0 !important;
+	}
+}
```
MaggieCabrera: should this be in [_utility.scss](https://github.com/Automattic/themes/blob/trunk/blockbase/sass/base/_utility.scss) instead?

skatepark/theme.json
```diff
@@ -1,6 +1,12 @@
 {
 	"$schema": "https://json.schemastore.org/theme-v1.json",
 	"version": 1,
+	"templateParts": [
+		{
+			"name": "header-ollie",
+			"area": "header"
```
MaggieCabrera, line 7: this one is not on Blockbase? Since Skatepark has not been yet launched, can we simplify the header so it uses one of the others that we've already defined? Mainly to avoid the extra CSS to maintain if at all possible.

/cc @kjellr

scruffian: Or we could move it to Blockbase to give all themes another option. It would be nice to be able to achieve it with less CSS though.

scruffian (approved):
LGTM

skatepark/theme.json
pbking: I would really like to add it as another option to Blockbase too.  

At the same time I was also keen to evaluate/demonstrate what the template part option added in a child theme would behave.  I don't think it's a bad experience and works for this pass (I hope).  It just felt too daunting compared to the other headers to integrate into Blockbase now though.  I definitely don't want to trade that header in for one of the others; the design is solid and it's a really good option I think.  It just needs more "Native Block Layout" refactoring to bring it in.

MaggieCabrera: We can move it to Blockbase on a follow up PR. I still think it needs some work, there's few couple of issues opened against it because it's not looking like the design: https://github.com/Automattic/themes/issues/4965 https://github.com/Automattic/themes/issues/4591 https://github.com/Automattic/themes/issues/4964

MaggieCabrera (approved):
Bring it in and we can iterate on Skatepark later on

blockbase/sass/base/_layout.scss
pbking: I agree.  Changed.

scruffian:
@Automattic/flow-patrol-manage these changes will need to happen to other Blockbase children...

alaczek:
Could I please get some guidance on how to best make those changes in Kerr and Byrne?

In Byrne the issue is with both top padding and nav menu placement.
image

In Kerr I fixed the top padding, but the nav menu placement is off.
image

scruffian:
What I would try is removing the header template from those themes so it falls back to the default Blockbase one. Blockbase supplies a variety of different header templates now, so you could choose one that you need.

If these themes need a different header from the one supplied then you'll need to add the `justifyContent` property to the header group block.

alaczek:
Thanks for the tip! I tried the Blockbase headers and the linear one sounds like a good fit, but the spacer at the bottom really throws off the balance (in Byrne and Kerr at least). So I'm guessing I'm still going to need a custom header - unless there is another way?

scruffian:
If you copy the template part called `header-spacer.hmtl`  from Blockbase then you should be able to adjust the amount of space using that template.

alaczek:
Yeah, I've thought about that too, but I do want zero, and it feels slightly hacky? But happy to do it if it's the right way :)

scruffian:
You can do it either way, I think both are fine :)

pbking:
If you want it zero you could just include an empty header-spacer.html template (like [Arbutus](https://github.com/Automattic/themes/blob/trunk/arbutus/block-template-parts/header-spacer.html)).  That way no spacer will be added there.  

Adjusting the space by way of the header-spacer template ensures that the same amount of space (none) will be used should the user switch to another header option.  If you make a custom header instead then if the user switches to another style then the default amount of space introduced by Blockbase will be used instead.

alaczek:
Thanks for the tip @pbking - this is what I'm going to do!

simison:
Empty header spacer causes UX issue, where "plus button" is _always_ visible in the editor: https://github.com/Automattic/themes/issues/5236

mtias:
Please, let's remove this concept of a "header spacer template" altogether and look for a different solution. This is super confusing as a user.

PR #5147, opened 2021-12-04
Blockbase: replace user key with custom

### LINKED ISSUE #5077: Blockbase: rename user and core to custom and default
For 5.9 (12.1) there's been a change that updates the core and user instances of theme.json data to custom and default. We need to update our scripts accordingly.

https://github.com/WordPress/gutenberg/pull/36748
https://github.com/WordPress/gutenberg/pull/36645

mikachan:
This PR replaces the `user` origin key with `custom`, from this GB PR: https://github.com/WordPress/gutenberg/pull/36748.

To test, make sure you're using at least GB 12.1, and change the color and fonts of Blockbase in the Customizer. Without this change, these settings will be overwritten/reverted when leaving or refreshing the Customizer. With this change, these settings should save correctly.

I couldn't find any references to the `core` key, which we should change to `default` (from this PR: https://github.com/WordPress/gutenberg/pull/36645), and as the color and font customizations seem to be working with the above, I'm hoping it's only the `user` key we need to change.

#### Related issue(s):
Closes #5077

### DIFF
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -121,6 +121,16 @@ function blockbase_fonts_url() {
 				$font_families[] = $font['google'];
 			}
 		}
+
+	// NOTE: This should be removed once Gutenberg 12.1 lands stably in all environments
+	} else if ( ! empty( $theme_data['typography']['fontFamilies']['user'] ) ) {
+		foreach( $theme_data['typography']['fontFamilies']['user'] as $font ) {
+			if ( ! empty( $font['google'] ) ) {
+				$font_families[] = $font['google'];
+			}
+		}
+	// End Gutenberg < 12.1 compatibility patch
+
 	} else {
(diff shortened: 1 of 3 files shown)


scruffian (approved):
I also renamed a key in functions.php in Blockbase

scruffian:
When 12.1 launches on wpcom we need to launch this ASAP

scruffian:
We might want to change the approach here so that it can work before 12.1 launches.

pbking:
I pushed a change that will allow this to work with both 12.0 and 12.1.  Tested with first 12.0 and then upgraded Gutenberg to 12.1 and evaluated the same setup.

mikachan (commented):
> I pushed a change that will allow this to work with both 12.0 and 12.1. Tested with first 12.0 and then upgraded Gutenberg to 12.1 and evaluated the same setup.

I've just tested in the same way (on 12.0 and then 12.1) and it works well for me too 🚢✨

PR #5149, opened 2021-12-05
Blockbase Fonts: Only unset properties that are set

### LINKED ISSUE #58724: Russell: Adding custom CSS causes fonts selected in Customizer to no longer be reflected on the front end
### Quick summary

Adding any custom CSS to a site running Russell results in the site's default (or other fonts) fonts to be wiped out.

### Steps to reproduce

1. Activate Russell 
2. Under Appearance > Customize > Additional CSS, add any custom CSS at all, such as:
```
.entry-content {
  font-size: 14px;
}
```

3. Click "Save Changes"

Custom CSS should be applied, but **nothing else should happen**. 

Fonts should remain what they are set to in the Customizer, which in Russell's case are DM Sans (Headings) and Libre Baskerville (Body)

### What actually happened

Fonts selected in the Customizer are no longer reflected on the front end:

Global_Styles_Demo_–_An_awesome_tagline

### Context

Chat 33005377-hc
Follow-up ticket: 4561965-zen 

### Simple, Atomic or both?

_No response_

### Theme-specific issue?

Russell (may affect other themes as well, have not tested)

_No response_

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

_No response_

### Workaround details

Did not find any workaround

scruffian:
This is another approach to https://github.com/Automattic/themes/pull/5148.

The problem with this "unset" code is that if there are no properties set in the object, calling unset actually creates properties at that level, for example if we have an object of the form:

```
{ "version":"2" }
```

and we call `unset( $object->styles->typography->fontFamily)` on that object, the object is mutated to look like this:

```
{ "version":"2","styles":{} }
```

This will force the user version of theme.json to be empty and thus break all the styling on the frontend of the site.

The fix proposed here checks whether the properties exist in the object before unsetting them, which avoids the problem.

#### Related issue(s):
Fixes https://github.com/Automattic/wp-calypso/issues/58724

### DIFF
--- a/blockbase/inc/customizer/wp-customize-fonts.php
+++ b/blockbase/inc/customizer/wp-customize-fonts.php
@@ -437,6 +437,13 @@ class GlobalStylesFontsCustomizer {
 		}
 	}
 
+	function unset_property_if_it_exists( $property ) {
+		if ( isset( $property ) ) {
+			unset( $property );
+		}
+
+	}
+
 	function handle_customize_save_after( $wp_customize ) {
 		$body_value    = $wp_customize->get_setting( $this->section_key . 'body' )->value();
 		$heading_value = $wp_customize->get_setting( $this->section_key . 'heading' )->value();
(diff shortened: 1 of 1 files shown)


jeffikus:
I like this approach better than mine. Nice work - does this handle existing installations or only fresh sites like my initial approach in #5148

pbking (approved):
Good change.
I couldn't break it.
Tested on Gutenberg 12.0.1

PR #5151, opened 2021-12-07
Kingsley: Initial commit.

alaczek:
This is the initial commit of Kingsley - a super simple Blockbase child theme featuring a "Coming soon" homepage design. Everything else will fall back to the parent theme.

This is an updated version of Laporte, which will be retired when Kinglsey launches.


pbking (approved):
Nice and minimal.
Approved, but due to a lack of index.html template file requires [this fix](https://github.com/WordPress/gutenberg/pull/36910) to be in place before it can ship.

kingsley/block-template-parts/footer.html
```diff
@@ -0,0 +1,5 @@
+<!-- wp:group {"style":{"spacing":{"padding":{"top":"20px";"bottom":"20px"}}}} -->
```
scruffian, line 1: This footer looks very similar to the one that ships in Blockbase. Why do we need a different one?

alaczek: I like the smaller padding in this case. Here's how it looks like with Blockbase footer:
image

And how it looks like with a custom footer:
image

It's not a huge deal though and if you think it's going to save us headache with individual updates down the road, then I can remove it.

scruffian: I think this approach is fine for now. Once we have the concept of site padding in Gutenberg we can switch to that approach: https://github.com/WordPress/gutenberg/issues/35607

pbking: I'm not sure I see how site padding will allow us to control the gap **above** the footer.

scruffian: I hope there will be a variable we can reuse in other places.

pbking:
`index.html` template fixes what ailes us.

Ready to ship.

PR #5152, opened 2021-12-07
Ames: Initial commit.

alaczek:
This is the initial commit of Ames - a super simple Blockbase child theme featuring a gallery homepage design. Everything else will fall back to the parent theme.

This is an updated version of Antin, which will be retired when Ames launches.


pbking:
Because only two colors are defined in the palette the missing colors (primary, secondary, tertiary) need to be mapped to the colors that are available.  Using Skatepark as an example I've[ updated those palette mappings](https://github.com/Automattic/themes/pull/5152/commits/0f577d3f065ac956761d75b9db587f5969fcf0d7).

pbking (approved):
Approved, but due to a lack of index.html template file requires [this fix ](https://github.com/WordPress/gutenberg/pull/36910)to be in place before it can ship.

ames/block-template-parts/footer.html
```diff
@@ -0,0 +1,5 @@
+<!-- wp:group {"style":{"spacing":{"padding":{"top":"20px";"bottom":"20px"}}}} -->
```
scruffian: This footer looks very similar to the one that ships in Blockbase. Why do we need a different one?

alaczek: Yeah, looks like it's not needed. I'll remove it.

pbking:
presence of `index.html` template fixes the Gutenberg problem.  Approving for release.

PR #5153, opened 2021-12-07
Jackson: Initial commit.

alaczek:
This is the initial commit of Jackson - a super simple Blockbase child theme featuring a list of links on the homepage. Everything else will fall back to the parent theme.

This is an updated version of Jones, which will be retired when Jackson launches.


pbking (approved):
Made a couple of minor adjustments (the same as the other similar themes).  Looks good except the same requirement of [this fix](https://github.com/WordPress/gutenberg/pull/36910).

jackson/block-template-parts/footer.html
```diff
@@ -0,0 +1,5 @@
+<!-- wp:group {"style":{"spacing":{"padding":{"top":"20px";"bottom":"20px"}}}} -->
```
scruffian: This footer looks very similar to the one that ships in Blockbase. Why do we need a different one?

alaczek: I had issues with spacing - it used to be that the credit in Blockbase sat uncomfortably close to the bottom of the page, but I'm not seeing it anymore, so it can be removed.

alaczek:
> Looks good except the same requirement of this fix.
@pbking is this a blocker for launch? I need to give a ballpark launch date to Ian.

scruffian:
Without that fix the theme won't work correctly. There are work arounds we can do, but I'd prefer to get the issue fixed.

pbking:
The index.html file eliminates the immediate need for the Gutenberg fix.  Moving this to "approved".

PR #5158, opened 2021-12-07
Button 2: add jquery to deps array

### LINKED ISSUE #3201: Button 2: Mobile menu stopped working when a page is set as static homepage
## Summary
It looks like the latest update to 5.6 has broken the jQuery on the site. This is causing an error with the navigation js.

## Steps to replicate

1. Activate Button 2 theme
2. Setup primary menu
3. View site from a mobile device (dev console viewer does not replicate this, you need to use BrowserStack or actual mobile phone)
4. Try to open mobile menu.
5. View console

## Result

Nothing opens. Error in console for no jQuery defined.
[Screenshot](https://d.pr/i/MrCuup)

## Expected

Menu should open : )
[Screenshot](https://d.pr/i/Djicbk)

* User report: #27552996-hc
* Follow-up ticket: #3717531-zd

arcangelini:
There is an error that is appearing in some mobile browsers. The error that is showing is: `Uncaught ReferenceError: jQuery is not defined`, it seems that some mobile browsers are not loading this which is causing issues. As a solution, I have added `jquery` to the dependency array for the `main.js` file that is throwing the error.

#### Related issue(s):
Fixes #3201 

#### Screenshot:
**before**
Markup 2021-12-07 at 11 51 16

**after**
Markup 2021-12-07 at 11 57 27

### DIFF
--- a/button-2/functions.php
+++ b/button-2/functions.php
@@ -191,7 +191,7 @@ function button_2_scripts() {
 
 	wp_enqueue_style( 'button-2-blocks', get_template_directory_uri() . '/blocks.css' );
 
-	wp_enqueue_script( 'button-2-scripts', get_template_directory_uri() . '/assets/js/main.js', array(), '20170303', true );
+	wp_enqueue_script( 'button-2-scripts', get_template_directory_uri() . '/assets/js/main.js', array( 'jquery' ), '20211207', true );
 
 	wp_enqueue_script( 'button-2-navigation', get_template_directory_uri() . '/assets/js/navigation.js', array(), '20151215', true );
 


arcangelini:
Also, not sure if this is the right place for this? Let me know if this needs to be done from SVN and I can make a phab patch instead.

MaggieCabrera:
This took some time to test!

The PR fixes the error but I had some trouble reproducing the error and then checking if the fix solved it.

I couldn't reproduce locally so I sandboxed one of the sites with the problem and checked with browserstack following the fieldguide: PCYsg-iPs-p2

And this fixes the issue for me on one of the sites that couldn't open their menu! So thanks for the fix :D

MaggieCabrera (approved):

arcangelini:
Thanks @MaggieCabrera for your time, haha it was not fun to test or diagnose.

Is there any documentation on the deploy process to WPCOM? or do I just merge and let the rest handle itself?

MaggieCabrera:
> Thanks @MaggieCabrera for your time, haha it was not fun to test or diagnose.
> 
> Is there any documentation on the deploy process to WPCOM? or do I just merge and let the rest handle itself?

This is The Way: PNEWy-eqL-p2

PR #5163, opened 2021-12-08
TT2: Add color variations

### LINKED ISSUE #5160: Twenty Twenty-Two: Create color variation themes
Twenty Twenty-Two has had 5 color and font variants created for it [over in this PR](https://href.li?https://github.com/WordPress/theme-experiments/pull/292). These can be brought over to WordPress.com and launched as color variation themes. 
 
144467775-6cee1d9f-de8a-4358-a6d9-3bcea44ec9c0

scruffian:
This just copies the variations from https://github.com/WordPress/theme-experiments/pull/292

#### Related issue(s):
Closes #5160


scruffian:
Relies on https://github.com/WordPress/twentytwentytwo/pull/283

mikachan:
I can see the new themes but when I activate one, I can't see the header or footer template parts and the global styles don't seem to change. I'm guessing this is something in my setup but I can't figure it out!

This is from the mint variation, but the colours and fonts haven't changed and the header isn't loading in:

image

mikachan:
I'm on GB 12.1.0-rc.1

Edit: It's working on 12.0.0!

twentytwentytwo-blue/theme.json
```diff
@@ -0,0 +1,90 @@
+{
+	"version": 2,
+	"settings": {
+		"color": {
+			"palette": [
+				{
+					"slug": "foreground",
+					"color": "3F67C6",
```
mikachan: Looks like this is missing a # in front of the hex.

scruffian:
That's probably this issue: https://github.com/WordPress/gutenberg/pull/36910

mikachan (commented):
I had a couple of environment issues but I've got this working now 😅

All looking good to me, other than in the Swiss variation, where a couple of patterns use a background and text color combination that doesn't have good contrast. I can see this in:

- Video with header and details
- Video trailer
- Divider with image and color (light)
- Footer with text, title, and logo
- Text-only header with salmon background
- Text-only header with tagline and black background
- Logo and navigation header with gray background

Here's an example:
image

This uses foreground (black) as the background and secondary (#232521/dark gray) as the text color. I'm guessing the other patterns probably use these colors too but I haven't checked yet. Can we fix this with variation-specific patterns?

It'd be great if someone else could review as well, I don't trust my setup 100%..

scruffian:
I see the same problem. I noticed that "Swiss" defined a "quaternary" color which I don't think is used, so I replaced secondary with this, which looks like this:
Screenshot 2021-12-09 at 10 38 53

@kjellr what do you think?

kjellr:
> I see the same problem. I noticed that "Swiss" defined a "quaternary" color which I don't think is used

Weird, I don't remember doing that. 😅 

Let's use these colors instead: 

```
"palette": [
	{
		"slug": "foreground",
		"color": "#000000",
		"name": "Foreground"
	},
	{
		"slug": "background",
		"color": "#FFFFFF",
		"name": "Background"
	},
	{
		"slug": "primary",
		"color": "#D42731",
		"name": "Primary"
	},
	{
		"slug": "secondary",
		"color": "#F4F4F2",
		"name": "Secondary"
	},
	{
		"slug": "tertiary",
		"color": "#FFFFFF",
		"name": "Tertiary"
	}
]
```

(The repeated use of white for the tertiary color is intentional).

scruffian:
> Let's use these colors instead:

Done!

mikachan (approved):
Awesome! Those new colors look good and they've fixed the patterns mentioned above.

PR #5178, opened 2021-12-09
Fix/videomaker heading spacing

pbking:

Also removed the spacer in the videomaker header-spacer gap to achieve the design target.  Introduced the spacer in MOST templates (except for header-footer-only).

Adjusted the top-margin for header-footer-only to 0 to achieve the design target.


kjellr (commented):
This looks good! I pushed one small commit to re-adjust the size of the spacers so that it looks right visually. But this should be ready to go. 🙌

kjellr (approved):

PR #5179, opened 2021-12-09
Geologist Variations + Variation Creation Script

### LINKED ISSUE #5156: Geologist: Create color variation themes
We should create four color & font variations of the Geologist Theme. (We already have one White color variant, which is named Zoologist)

## Yellow

Color:

- Primary: `#000000`
- Secondary: `#333333`
- Background: `#FFF975`
- Tertiary: `#FFFFFF`

Font: 

- "Helvetica Neue", "Helvetica", "Arial", Sans-serif

Other Settings

- Heading, Post Title blocks:
    - Font Weight: Bold (700)

Screenshot: 

yellow

## Blue 

Color:

- Primary: `#F5F0E5`
- Secondary: `#FFFFFF`
- Background: `#424CBF`
- Tertiary: `#383E89`

Font: 

- "IBM Plex Mono", "Courier New", "Courier", monospace

Other Settings: 

- Site Title: 
    - Text transform: None
- Heading, Post Title blocks:
    - Font Style: Italic
    - Font Weight: Normal

Screenshot:

blue

## Cream 

Color:

- Primary: `#000000`
- Secondary: `#333333`
- Background: `#FFFAED`
- Tertiary: `#FFFFFF`

Font: 

- "Source Serif Pro", serif

Other Settings: 

- Site Title: 
    - Text transform: None
- Heading, Post Title blocks:
    - Font Style: Italic
    - Font Weight: Normal

Screenshot: 

cream

## Slate 

Color:

- Primary: `#FFFFFF`
- Secondary: `#FEF6B2`
- Background: `#192833`
- Tertiary: `#000000`

Font: 

- "Rubik", sans-serif

Other Settings: 

- Global Link Color: Secondary
- Heading, Post Title blocks:
    - Font Weight: Light (300)

Screenshot: 

slate

pbking:
This change adds all of the Geologist variations noted in #5156.

It does this by introducing a variation construction mechanism.

Variations are kept in the `/variations` folder.  The things that are unique about that theme are stored there and merged with the source.  All resources are copied with the exception of theme.json which is merged.

To run the script that builds all of the variations:
`npm run build:variations`

Also [changed](https://github.com/Automattic/themes/pull/5179/files#diff-1df64275b7f63668026745b96515d6e739b5335eac5b8aee0f1e54b6cf7fadbb) is a very small modification to Geologist, mapping the (previously unused) secondary palette color to the custom color mapper.  This will likely go unnoticed as Geologist doesn't use the secondary color anywhere.

Fixes #5156


kjellr:
These look good! Just a couple design notes (I may have left these out of my original instructions, so sorry about that!)

- **Cream, Blue**: Site title should be bold, not bold italic.
- **Slate**: Site title should be normal weight, not bold.

pbking:
>Cream, Blue: Site title should be bold, not bold italic.
Slate: Site title should be normal weight, not bold.

Adjusted.  Good for another look now.

variations/build-variations.mjs
```diff
@@ -0,0 +1,73 @@
...
+
+		// copy the readme
+		await fs.copy( localpath + '/variation-readme.md', destDir + '/variation-readme.md' );
+
+		// merge the theme.json files
+		const srcJsonFile = await fs.readFile( srcDir + '/theme.json', 'utf8' );
+		const srcVariationJsonFile = await fs.readFile( srcVariationDir + '/theme.json', 'utf8' );
+		const srcJson = JSON.parse( srcJsonFile );
+		const srcVariationJson = JSON.parse( srcVariationJsonFile );
+		const mergedJson = deepmerge(srcJson, srcVariationJson, {
```
scruffian, line 68: My concern with this is that this isn't how Gutenberg merges theme.json files, so we might get different results when we try to port this to a child theme.

pbking: What does it do differently?  I thought that's how it did it.  Merge objects, replace arrays.

scruffian: Oh you might be right. The way Gutenberg does it uses PHP and it's not very smart, so I know we need to be careful to copy it correctly. I'll investigate...

pbking: [This](https://github.com/WordPress/gutenberg/blob/12.0.01/lib/class-wp-theme-json-gutenberg.php#L1297) is where that logic was for v12.

But it seems to have gotten more complicated in its [current iteration](https://github.com/WordPress/gutenberg/blob/trunk/lib/compat/wordpress-5.9/class-wp-theme-json-gutenberg.php#L1404).

pbking: Do we need to note the limits to things that can be customized with our simple approach?  I think those basic rules (merge objects, replace arrays) get us the basic functionality we need for variations.

scruffian: Could we not reuse the `merge` function from Gutenberg itself?

scruffian: I can't see any way to do that. I think for now this approach is fine. If we notice issues we can iterate :)

scruffian:
I updated this so that the variations use the standard Blockbase. I think its ready to go...

scruffian:
The problem with this approach to the heading change is that now users will see two headers the same - header-linear will appear the same as header. I wonder if we should instead change the templates to point to the vanilla header?

pbking:
I'm not a fan of leveraging the header design in [that way,](https://github.com/Automattic/themes/pull/5179/commits/e1d212e5604778320599fde642dce4f7999d5ad0) it negates the act of putting it into Blockbase as a variation and does make for a confusing experience for users as you have noted.

If we want different headers to be the norm I think we should be writing that into the templates as you suggest.

You suggested that perhaps that aspect could also be scripted.  I'm keen on that.

kjellr:
The header in the child themes doesn't match the parent? The site title and tagline get stacked on the children. 🤔 

Parent|Child|Child
---|---|---
Screen Shot 2021-12-13 at 10 24 30 AM|Screen Shot 2021-12-13 at 10 24 43 AM|Screen Shot 2021-12-13 at 10 25 41 AM


Also, in the blue version, I don't think IBM Plex Mono is loading on my end? It's falling back to "Courier New" instead.

kjellr (approved):
I haven't tested in-depth, but in general the colors + headers look right on my end. 👍

geologist-blue/theme.json
```diff
@@ -176,13 +176,15 @@
 					"fontFamily": "'IBM Plex Mono', 'Courier New', 'Courier', monospace",
 					"fontSlug": "ibm-plex-mono",
 					"slug": "body-font",
-					"name": "Body (IBM Plex Mono)"
+					"name": "Body (IBM Plex Mono)",
+					"google": "family=IBM+Plex+Mono:ital,wght@0,100;0,200;0,300;0,400;0,500;0,600;0,700;1,100;1,200;1,300;1,400;1,500;1,600;1,700"
```
scruffian, line 180: I added this to the list of customizer fonts too.

PR #5194, opened 2021-12-14
Added site logo size restriction CSS

### LINKED ISSUE #2066: Varia & Child Themes: Site Logo is not being resized
## Steps to replicate

Go to the customizer and add a site logo. 
Then go to the homepage and see the Logo is being shown in full size. 
This is happening because there is no styling being applied for `site-logo` here https://github.com/Automattic/themes/blob/master/varia/sass/components/header/_site-branding.scss#L3

## Result

image

## Expected

The logo should be properly resized. Adding minimal CSS like following could solve the issue
```
.site-logo {
	position: absolute;
	right: calc(100% + (1.25 * #{$typescale-base}));
	z-index: 999;

	.custom-logo-link {
		display: inline-block;
		width: 64px;
		height: 64px;
	}
}
.site-branding {
	color: #{map-deep-get($config-header, "branding", "color", "text")};
	position: relative;
}
```
**After applying the above CSS**

image

pbking:

Adds CSS to the site logo in the header to restrict the width to 64px (allowing the height to scale to retain dimensions).

Confirmed that site header blocks used in the content are sized according to the user adjustments.

Rebuilt all Varia themes to include that change.

#### Related issue(s):

Fixes #2066 

Before:

After:

### DIFF
--- a/alves/style-rtl.css
+++ b/alves/style-rtl.css
@@ -2770,6 +2770,12 @@ table.is-style-stripes tbody tr:nth-child(odd),
  * Components
  * - Similar to Blocks but exist outside of the "current" editor context
  */
+.site-logo a {
+	display: inline-block;
+	width: 64px;
+	height: auto;
+}
+
 .site-branding {
 	color: #4d6974;
 }
--- a/alves/style.css
+++ b/alves/style.css
@@ -2789,6 +2789,12 @@ table.is-style-stripes tbody tr:nth-child(odd),
  * Components
  * - Similar to Blocks but exist outside of the "current" editor context
  */
+.site-logo a {
+	display: inline-block;
+	width: 64px;
+	height: auto;
+}
+
 .site-branding {
 	color: #4d6974;
 }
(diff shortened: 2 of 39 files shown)


alves/style-rtl.css
```diff
@@ -2770,6 +2770,12 @@ table.is-style-stripes tbody tr:nth-child(odd),
  * Components
  * - Similar to Blocks but exist outside of the "current" editor context
  */
+.site-logo a {
+	display: inline-block;
+	width: 64px;
```
MaggieCabrera, line 2775: Should we use max-width instead in case the user has a smaller logo? We don't want to stretch them, do we?

MaggieCabrera:
Maybe we need design eyes on this, since this change may affect users with a logo bigger than 64px, not sure about the protocol here. @kjellr @beafialho

kjellr:
Yeah, I don't think we should do this — there are far too many users of Varia themes at this point, and I'm sure many of them are happy with logos that are wider than `64px`. 

I know it's a little weird, but can we figure out a way that old themes could top-into this, and we could enable it by default for new installs?

pbking:
> Yeah, I don't think we should do this — there are far too many users of Varia themes at this point, and I'm sure many of them are happy with logos that are wider than 64px.

Good point.  Doing this removes what user control there is over sizing.

I would argue against this behavior for new installs too for the same reason.  My opinion is to just close this as "wontfix".

kjellr:
Yeah, I think that's ok for now.

PR #5195, opened 2021-12-14
[Seedlet] Replacing border for text-decoration to add underline to links

### LINKED ISSUE #4191: Blank Canvas: Link underlines break on multiple lines
## Steps to replicate

1. Create some text that goes onto a second line
2. Turn that text into a link

## Result

Because the CSS uses a border for link styling, the border only appears on the bottom row of text:

## Expected

I expect the link underline to work on both lines, not just the bottom line of text:

matiasbenedetto:
Replacing border for text-decoration to add underline to links

#### Related issue(s):
https://github.com/Automattic/themes/issues/4191

#### Screenshots

Before
image

After
image


MaggieCabrera:
This is affecting Blank Canvas and Spearhead too (both are Seedlet's children), we should check that they are not getting any visual regressions

pbking:
Spearhead links are double-underlined and the page title and site title shouldn't be underlined.

Original:
image

Current:
image

Blank Canvas site title shouldn't be underlined.

Original:



Current:

scruffian:
cc @kjellr to get some input on why we used border instead of underline

pbking:
My GUESS is that [text-underline-offset ](https://caniuse.com/?search=text-underline-offset)and wasn't available at the time and for ultimate control of position and style the bottom border was used.

I think this achieves the same design target as the original in a way that couldn't have been done a couple of years ago, and certainly not while IE11 was on our radar.

kjellr:
> My GUESS is that text-underline-offset and wasn't available at the time and for ultimate control of position and style the bottom border was used.
> 
> I think this achieves the same design target as the original in a way that couldn't have been done a couple of years ago, and certainly not while IE11 was on our radar.

Yep, that's correct. If we can be sure that the appearance matches the previous one, that's fine with me.

pbking:
Spearhead still looks unchanged.  Lines under Site title and page title persist, double underlines under a paragraph link.

The same is true of the Blank Canvas site title, an underline is still there.

Also Seedlet seems to have developed a double-underline site-title as well.  Or more likely I just didn't notice on my first pass. :P

image

matiasbenedetto:
@pbking I'm not able to reproduce the issue you mention. Example for blank canvas: 
image

matiasbenedetto:
Example from spearhead:
image

kjellr:
Yeah, we're doing something different for that underline. I can't remember why, but there was a good reason at the time. 😅

scruffian:
I believe the reason was so that that underline breaks around letters that drop below the line

kjellr:
Yeah, I _think_ that should happen automatically? I just tried some tests in the inspector using the demo site and was unable to get it working right in Safari. That underline is the most visible underline on the whole site (and it looks great today), so if we change this, we need to make sure that one looks identical.

pbking:
I think we can probably leave the Seedlet Site Title as-is (bottom border).  I don't know that we could style that the same and I don't see the downside to that exception.  Line-breaks on that element are unlikely.

kjellr:
👍 That sounds good. Plus, it's been that way for ages now.

matiasbenedetto:
In my latest commit, I disable the `text-decoration` for the site title to continue using the border.

seedlet/assets/sass/base/_base.scss
```diff
@@ -1,13 +1,13 @@
 // Links styles
 a {
-	border-bottom: 1px solid var(--global--color-secondary);
+	text-decoration: underline 1px var(--global--color-secondary);
+    text-underline-offset: 0.3em;
```
scruffian: Looks like you used spaces instead of tabs here.

seedlet/assets/sass/components/header/_header-branding.scss
```diff
@@ -17,6 +17,7 @@
 	margin-bottom: calc( var(--global--spacing-vertical) / 2 );
 
 	a {
+		text-decoration: none;
```
scruffian: I pushed a commit to move this lower so that the properties are alphabetical. Not important but I couldn't help myself!

scruffian (approved):
LGTM

PR #5197, opened 2021-12-15
Bennet: Initial commit.

alaczek:
This is the initial commit of Bennet - a super simple Blockbase child theme. 
This is an updated version of Bingley, which will be retired when Bennet launches. More info can be found here: pNEWy-exS-p2


bennett/block-template-parts/header-minimal.html
```diff
@@ -0,0 +1,31 @@
+<!-- wp:group {"className":"gapless-group","layout":{"inherit":"true"}} -->
```
pbking, line 1: The only changes I see introduced by this `header-minimal.html` template over the version in /blockbase are blank lines.  Can this be removed instead so the parent (Blockbase) instance is used instead?

pbking:
Because the header template part is only changed in index.html and header-footer-only.html templates all of the other templates still use the default blockbase header.

Index
image

A page
image

Unfortunately, in order to change the header template part used across the site the template part referenced has to be defined in each template:
404.html
archive.html
header-footer-only.html
index.html
page.html
search.html
single.html

alaczek:
Ugh, I totally forgot that I opened this PR. Let's close it in favor of #5283.

PR #5199, opened 2021-12-15
Seedlet: Fix navigation block styles in editor 

### LINKED ISSUE #3684: Seedlet: Navigation Block layout is different on editor and front-end
## Steps to replicate

1. Activate Seedlet
2. Add a Navigation block
3. Change the background color
4. See the result

## Result

1. In the editor, the background is only applied to the items not the whole menu
2. With a dark background, it's not possible to see/read the **Add a link** button
3. With the same background, hovering over also doesn't let you read the menu items

Editor | Front-end
--- | ---
Captura de Tela 2021-04-21 às 18 10 17 | Captura de Tela 2021-04-21 às 18 10 49

## Expected

* The same menu style in both editor and public site
* Readable links

<small>Part of the Empathy Flow testing p9b1hy-1Fq-p2</small>

mikachan:
This PR removes some background and margin properties from the navigation container in the editor for Seedlet. This fixes a bug where the navigation block looks different between the front end and the editor when a background is applied to the block. 

It also fixes the following:
- fixes an issue with the link hover color sometimes being the same color as the background
- removes a bottom border / underline from the default navigation block

Before:

| Editor | Front end |
| ----- | -------- |
| image |    image|

After:

| Editor | Front end |
| ----- | -------- |
|   image  |   image    |

#### Related issue(s):
Fixes #3684


seedlet/assets/sass/blocks/navigation/_editor.scss
```diff
@@ -16,23 +20,33 @@
         }
     }
 
+    .wp-block-navigation-item {
+        > a {
+            border: 0;
```
scruffian: Is this going to play nicely with https://github.com/Automattic/themes/pull/5195 ?

scruffian:
This works in my testing, but I'm not that keen on adding the extra CSS. What happens if we leave it out?

scruffian:
If I remove the extra CSS it looks pretty different to the frontend:
Screenshot 2021-12-20 at 10 43 15

I wonder if we can fix this in the block?

seedlet/assets/sass/blocks/navigation/_editor.scss
mikachan: Hmm.. possibly, although it looks like the navigation may need to override the `text-decoration` styles instead of the `border`.

mikachan:
> I wonder if we can fix this in the block?

Ah yeah, I was thinking this was very Seedlet-specific. I'll try playing around with the block now!

mikachan:
Opened a GB PR to hopefully fix the spacing issue here: https://github.com/WordPress/gutenberg/pull/37587

scruffian (approved):

PR #5200, opened 2021-12-15
Fixed deploy version bump commit problems

pbking:
This change fixes a couple of problems with auto-commit-and-pushing of Version Bump changes.

The deploy action no longer breaks if there aren't any version bump changes to push.

The deploy action prompts you before pushing the changes up to github.

Only the changed files are changed in the commit, other files (like scripts, especially this one...) are excluded from that change.

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -129,7 +129,7 @@ async function pushButtonDeploy(repoType) {
 			}
 
 			await executeCommand(`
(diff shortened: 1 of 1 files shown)


pbking:
Unfortunately, the `-a` flag was still in place when I tested this change and... it got picked up in the Version Bump push.

Hopefully for the last time.

Bringing this last (very important) part in to finish the job.

PR #5208, opened 2021-12-16
Hever: fix margin for alignfull child on single pages

### LINKED ISSUE #4978: Hever: Full-width layout block overlaps featured image
## Steps to replicate
Create a new post
Add a featured image
Add a Layout Grid block - keep it on full alignment
Put content in each column
Save and view the post on its own (not in a list of posts)

## Result
The layout block gets a -64px margin, only on .single, and only when the layout is full (which is default).

On pages, it looks awkward. On posts, the first line is unreadable if the post has a featured image.
Screen Shot 2021-11-03 at 11 09 17 AM

Screen Shot 2021-11-03 at 11 08 47 AM

## Expected
Here's how it looks if I set the alignment to none: the margin appears and the content looks more normal.

Screen Shot 2021-11-03 at 11 15 40 AM

Screen Shot 2021-11-03 at 11 08 15 AM

I suspect this is in place to deal with people using full-width cover images at the top of pages, since that looks more normal:

Screen Shot 2021-11-03 at 11 26 27 AM

* User report: https://wordpress.com/forums/topic/layout-grid-block-issues/
(note this has _multiple_ issues; I'm going to try to get those untangled)

* Follow-up ticket: https://wordpress.com/forums/topic/layout-grid-block-issues/

MaggieCabrera:

This PR reverts the change made by https://github.com/Automattic/themes/commit/c79bb5ed65d255cff2bd2bd8f1bd3a95d11651d9. I couldn't find a condition where we want this margin for the theme, I don't have enough context as to why this was introduced to make an exception to this rule but I've seen this breaks/looks bad on all the blocks that I tried giving full-width alignment on single pages (it has to be the first block on the post/page), so I opted to remove it entirely. 

#### Related issue(s):

Closes https://github.com/Automattic/themes/issues/4978

### DIFF
--- a/hever/sass/_extra-child-theme.scss
+++ b/hever/sass/_extra-child-theme.scss
@@ -334,19 +334,6 @@ article .entry-header .entry-title,
 }
 
 // Remove margin if alignfull is first element
-.singular {
-	.hentry .entry-content {
-		& > *:first-child {
-			&.alignfull {
-				margin-top: -#{0.666 * map-deep-get($config-global, "spacing", "vertical")};
-
-				@include media(mobile) {
-					margin-top: -#{2 * map-deep-get($config-global, "spacing", "vertical")};
-				}
-			}
-		}
-	}
-}
 
 .home.hide-homepage-title {
 	.hentry .entry-content {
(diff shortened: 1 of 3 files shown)


scruffian (approved):
LGTM

PR #5210, opened 2021-12-16
Exclude /sass from any premium themes pushed to premium repo

pbking:
This change excludes any SASS files pushed to the /premium repo.

This is because the projects may contain relative imports to projects outside of that premium theme (in the case of videomaker referencing blockbase .sass resources).


scruffian (commented):
LGTM

Issue #4993, opened 2021-11-04
Varia themes: Default text Color inside the Cover Block is not the same in the editor and the front-end


AtrumGeost:
## Steps to replicate

1. Activate the Hever theme and select the option to replace the Home page
2. Open the new Home page in the editor
3. Notice the color of the text inside the Cover block (white)
4. Notice the color on the actual site (black)

## Result

![Screen Shot on 2021-11-04 at 11-29-21](https://user-images.githubusercontent.com/3696121/140379726-8d0e405b-aed4-4bf5-b336-623847b2608b.png)

## Expected

To have the same color as in the editor:

![Screen Shot on 2021-11-04 at 11:30:11](https://user-images.githubusercontent.com/3696121/140379844-be99a981-da16-42d4-bbe8-2dd2a700dd18.png)

* User report: 4429491-zd-woothemes

dcoleonline:
Also in 4424948-zen

We should update them once this is resolved. For now, I've manually set the color option in the editor to override this issue.

AtrumGeost:
Another report here:

4452487-zd-woothemes

jordesign:
With some testing I've been able to narrow down this issue a little more. It appears to happen in the Varia parent theme and all child themes. (all I tested anyway).

The issue is that the default text color for content within Cover blocks (and others as well like Form blocks) isn't matching between editor and front-end preview.

Case (you can check Pdw1nU-d-p2 for testing):

Editor (default text appears white):

<img width="1368" alt="Screen Shot 2021-11-17 at 6 57 48 am" src="https://user-images.githubusercontent.com/1842363/142056730-8ff10058-1854-44ed-9359-2886d046692b.png">

Front End (default text appears dark):
<img width="1186" alt="Screen Shot 2021-11-17 at 6 58 49 am" src="https://user-images.githubusercontent.com/1842363/142056833-a90413bc-78dc-4cc5-be03-d9f77674fd46.png">

----

For bonus weirdness I made some changes to the colors in the customizer - and they had unexpected effects on the color pallet in the editor and the front end.

Customizer change (changing default text color to light green):
This changed text colors even when other colors had already been specifically applied (using color presets in the editor)

<img width="1474" alt="Screen Shot 2021-11-17 at 7 00 08 am" src="https://user-images.githubusercontent.com/1842363/142057219-f4e2a8ef-3111-4f6b-8a18-3592f9fe16d4.png">

In the Editor:

[...]

PR #5212, opened 2021-12-16
Varia: Remove code that changes the cover block colors

### LINKED ISSUE #5182: Varia Themes: Multuple issues with handling of Cover Block colors
This is an umbrella issue to gather multiple open issues related to the current handling of Cover blocks by Varia. 

1. At some point it appears that the default text color display of Varia themes became different between the editor (where it appears white) and the front end (where it appears dark grey). The symptoms of this are reported in 2 different issues:

o https://github.com/Automattic/themes/issues/5181
o https://github.com/Automattic/themes/issues/4993

2. It appears that Varia themes don't adjust the font color for contrast when the opacity of the cover overlay changes, or the darkness of the background color itself changes.. 

o https://github.com/Automattic/themes/issues/5051

scruffian:
Removes the code that deals with the colors in the cover blocks. It can be tested on this site: PdA1qf-f-p2

This seems to solve the initial report, but it will also change the colors of cover blocks on sites where users haven't customized them, so it feels quite risky. Not sure if there's another way...

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/5182 and #4993


pbking (approved):
Due to all the auto-formatting the change was a bit hard to follow but I think it just comes down to [this](https://github.com/Automattic/themes/pull/5212/files#diff-3168dce0e379e1cc5b0bd22433a9f997f3582dae7b03c7385c411b1710ce22ff) .scss file and [these](https://github.com/Automattic/themes/pull/5212/files#diff-2006c5daa354cd29b030f89bf94866d73c13040de08bd83bf30f2f81a9fb5f8a) PHP changes.

Those LGTM.  Removing those styles is the right call; that's Gutenberg business.

MaggieCabrera:
Yeah, I think this should work like this, I think it's fine if the color annotations don't affect the cover block anyway, let the editor and the user handle it.

PR #5213, opened 2021-12-16
Add "Livre" theme

kjellr:
This is a quick block theme inspired by a chat with @mtias and @pablohoneyhoney. It looks like this: 

Screen Shot 2021-12-16 at 9 32 09 AM

It's forked from Twenty Twenty-Two, so it should contain all the required templates and settings. It likely needs a little additional testing and `theme.json` pruning, but it's in a solid, workable state.


jeffikus:
This is pretty slick - I'm assuming the menu is intended to hold Chapters?

kjellr:
It's just a menu — it can hold whatever you want!

booklet/theme.json
```diff
@@ -0,0 +1,231 @@
...
+				"page",
+				"post"
+			]
+		}
+	],
+	"settings": {
+		"appearanceTools": true,
+		"color": {
+			"palette": [
+				{
```
scruffian: If I switch to this theme from a theme that has `primary`, `secondardy` and `tertiary` colors defined then my patterns will break because this theme doesn't define them. Maybe this doesn't matter....

scruffian (approved):
LGTM, just one minor comment.

booklet/theme.json
kjellr: Yeah, I'm not totally sure we _have_ to keep these. I guess it's part of the ongoing discussion over in [29568](http://href.li?https://github.com/WordPress/gutenberg/issues/29568).

I've [added in a few more colors for now](https://github.com/Automattic/themes/pull/5213/commits/d7e9c4e4fdac70227a29b908166a50f35f81884f). They're unused in the theme at the moment, but they'll come in handy for patterns in the future.

Screen Shot 2021-12-16 at 10 44 48 AM

kjellr:
I need to update the background color because I realized I basically landed on the same color Geologist uses. 😅

kjellr:
Colors are updated: 
Screen Shot 2021-12-16 at 11 59 29 AM

kjellr:
Let's hold off on merge for a moment, since we may change the name. 👍

kjellr:
Alright, this should be ready to merge for testing on dotcom. Changed the name to "Livre" (French for "book"), and made some subtle adjustments to the colors.

pbking:
I have pushed this theme to the pub/ repo and it should be available for activation now.  However doing so on my sandbox gives me a completely empty page error.  

The theme works as-expected locally.

pbking:
Found to be the change in template folders naming which happens in Gutenberg 12.2 (still in flight for simple sites).

scruffian:
Yeah I think we can just wait for 12.2 to ship...

PR #5220, opened 2021-12-17
Blockbase: Update alignment rules

scruffian:
This copies the alignment rules from TT2 to Blockbase. Its easier for us to maintain only one set of alignment CSS. In addition the Blockbase CSS causes this bug in the Site Editor:
Dec-17-2021 08-30-04

### DIFF
--- a/blockbase/sass/base/_alignment.scss
+++ b/blockbase/sass/base/_alignment.scss
@@ -37,14 +37,14 @@
  .edit-post-visual-editor__post-title-wrapper,
  .wp-block-group.alignfull,
  .is-root-container .wp-block[data-align="full"] > .wp-block-group {
-	padding-left: var(--wp--custom--spacing--outer);
-	padding-right: var(--wp--custom--spacing--outer);
+	padding-left: var(--wp--custom--gap--horizontal);
+	padding-right: var(--wp--custom--gap--horizontal);
  }
 
  .wp-site-blocks .alignfull,
  .is-root-container .wp-block[data-align="full"] {
-	margin-left: calc(-1 * var(--wp--custom--spacing--outer)) !important;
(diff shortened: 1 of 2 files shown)


blockbase/sass/base/_alignment.scss
```diff
@@ -54,3 +18,43 @@
...
+ .edit-post-visual-editor__post-title-wrapper,
+ .wp-block-group.alignfull,
+ .is-root-container .wp-block[data-align="full"] > .wp-block-group {
+	padding-left: var(--wp--custom--spacing--outer);
+	padding-right: var(--wp--custom--spacing--outer);
+ }
+
+ .wp-site-blocks .alignfull,
+ .is-root-container .wp-block[data-align="full"] {
+	margin-left: calc(-1 * var(--wp--custom--spacing--outer)) !important;
```
MaggieCabrera: this variable is not the same for blockbase

scruffian:
From what I can see from a quick review this does change our alignments so we might need extra code....

scruffian:
The only difference I can see is the behaviour of an image inside a full width group:

Screenshot 2021-12-17 at 09 12 23

I'm not sure what the expected behaviour is, and I'm not sure how Blockbase achieves this:
Screenshot 2021-12-17 at 09 13 30

MaggieCabrera:
> The only difference I can see is the behaviour of an image inside a full width group:
> 


What are you comparing here? Blockbase with TT2?

scruffian:
> What are you comparing here? Blockbase with TT2?

No, blockbase with pcrugM-lB-p2. It looks like the same issues aren't on trunk locally so I think this is probably ok...

MaggieCabrera (approved):
I don't see the issue from the gif and the alignments seem to be working fine for me after you fixed the wrong variable

PR #5221, opened 2021-12-17
Fixed submenu style of focused submenu to match hover styling

### LINKED ISSUE #1843: Varia: Menu items with anchor links and submenu items have persistent unstyled submenus
## Steps to replicate
1. Install and activate Varia or any child theme. 
2. Add a few items to your primary menu. One of them should be a Custom Link with either a single # or a complete anchor link, and at least one submenu item. Publish your changes.
3. Head to the site and click on that menu item. 

## Result

The drop-down menu persists even after you move your mouse away, unless you click on the item again.

On some of the child themes, it's particularly apparent that the persistent drop-down isn't styled the same as the one that appears on hover. Hever and Morden are examples. 

Tested on WordPress.com Premium (Hever, Morden, Stow); WordPress.com Business (Morden, see user report); self-hosted (Varia). 

No images/video because I'm having trouble with screen capturing at the moment - I'll edit with images once that resolves. 

## Expected

The drop-down appears when the parent item is clicked on, but does not persist (or, if this is intentional, persists styled the way the hover version is).

* User report: #17888781-hc
* Follow-up ticket: #2773404-zd

pbking:
Fixes: #1843

Varia supplies a "hover or focus" style for navigation submenus.

Some child themes modify this style, however they incorrectly selected the "focus" styles, thus when the submenu was "selected" (clicked and the menu stays open, for a11y reasons) the styles rolled back to varia's defaults.

This change just changes the selector to use `:focus-within` instead of (an absent) `.focus` class to address the problem.  The result is that the style now matches the "hover" state.

The following themes are affected:

Alves
Before:

After:

Balasana
Before:

After:

Barnsbury
Before:

After:

Brompton
Before:

After:

Dalston
Before:

After:

Hever
Before:

After:

Morden
Before:

After:

Redhill
Before:

After:

Rivington
Before:

After:

The rest of the Varia child themes either use the Varia default styling or properly style the focused state already.

### DIFF
--- a/alves/sass/_extra-child-theme.scss
+++ b/alves/sass/_extra-child-theme.scss
@@ -266,11 +266,8 @@ body:not(.fse-enabled) {
 			}
 
 			& > div > ul > li:hover,
-			& > div > ul > li.focus,
+			& > div > ul > li:focus-within,
 			& > div > ul > li.current-menu-item {
-
-				& > a {
-				}
 
 				& > ul {
 					box-shadow: none;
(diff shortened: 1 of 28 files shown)


mikachan (approved):
I've tested this using the instructions in #1843, and all these themes are looking good to me. I can see that the dropdowns now all look consistent with the hover states 👍

PR #5233, opened 2021-12-20
removed header template part from skatepark's json

MaggieCabrera:

This PR removes the header-ollie definition on the child theme's json, since it was moved upstream to the parent and we forgot to delete it from theme.json


mikachan (approved):
LGTM 🚢

PR #5234, opened 2021-12-20
Skatepark: use grid of posts for Search page

### LINKED ISSUE #5231: Skatepark: search page bugs
Looks like at some point we lost the separation between header and content in the search.html template:

Screenshot 2021-12-20 at 10 19 45

Looks like there's no design for this page, since the index.html of Skatepark shows the posts on a grid. I think we should alter the template to look like the index with proper spacing between the search block and the list of posts. What do you think @kjellr?

MaggieCabrera:

This PR encapsulates the grid of posts into a template part reusable between the index, archive and search pages. The search page now looks like this:

Screenshot 2021-12-20 at 10 46 29

#### Related issue(s):

Closes https://github.com/Automattic/themes/issues/5231


MaggieCabrera:
I just found the [original PR for the search page](https://github.com/Automattic/themes/pull/4410) (my bad for not checking beforehand...) but I still think it's good that we moved the query block to its own template part to avoid code duplication. Should we go back to the original design or keep the grid posts, @kjellr? The original issue was displaying visual regressions in any case from what was on #4410

beafialho:
> This PR encapsulates the grid of posts into a template part reusable between the index, archive and search pages.

@MaggieCabrera would it be a lot of extra work and time turning the Search template into [this](https://www.figma.com/file/CA8Gb16mt9yYNWOiVLAWak/Skatepark?node-id=2781%3A2551)?

Captura de ecrã 2021-12-20, às 11 03 27

I think it would align better with the design, but I also may lack context here. Let me know what you think.

MaggieCabrera:
> @MaggieCabrera would it be a lot of extra work and time turning the Search template into [this](CA8Gb16mt9yYNWOiVLAWak-fi-2781%3A2551)?
> 

Sadly there's no block available to us that gives us the search term :( that's why we are using the search block instead.

MaggieCabrera:
Ok so I came back to this, and I've recovered the old style original from https://github.com/Automattic/themes/pull/4410 while maintaining the template part I made for archive and index so this PR is a 2 for 1: I removed the duplicate query loop from archive and index and I improved the search template to look more closely like the design from Mel on the linked PR above. I don't see the design on the figma file so I had to eyeball it, please let me know if you think I should make any other changes @beafialho 

Mel's design:

image

Current state of this PR:

Screenshot 2021-12-21 at 16 55 11

skatepark/block-templates/archive.html
```diff
@@ -1,35 +1,16 @@
...
-	<!-- wp:query-pagination -->
-	<!-- wp:query-pagination-previous /-->
-
-	<!-- wp:query-pagination-numbers /-->
-
-	<!-- wp:query-pagination-next /-->
-	<!-- /wp:query-pagination --></div>
-	<!-- /wp:query -->
+	
+	<!-- wp:template-part {"slug":"posts-grid"} /-->
```
scruffian: This is the same duplication that we have in most Blockbase themes since we removed the archive title tag from the index.html template. Is moving the duplicated code to a template part another example of creating additional complexity in the UI, or is it acceptable in this case?

cc @mtias

scruffian:
Code looks good to me and works as advertised. I'll leave it to @beafialho to approve the design.

skatepark/block-templates/archive.html
MaggieCabrera: Maybe it should be a pattern instead!

MaggieCabrera: I made it a pattern instead! We already had it in place, so we were duplicating code again. But after my conversation with @mtias over slack I think this is the way to go. He said we have to be careful mixing patterns and template parts, the difference being that template parts are assumed to be used somewhere. 

In this case, I think it makes more sense for it to be a pattern that we can just reuse over the theme and let the user reuse it too if they want as well. The pattern block sure is handy to have for this kind of stuff!

MaggieCabrera:
This PR ended doing two completely different things at once, but I think it's simple enough to review.

@scruffian if you are ok with the blog posts refactor let me know and I'll be waiting on @beafialho thumbs up for the design of the search page and we can wrap this one up.

Thanks for keeping up with this!

kjellr:
Looks good from a design perspective. 👍

beafialho (approved):

scruffian (approved):
LGTM

PR #5235, opened 2021-12-20
Deploy: Use the requirement rather than the tested version

scruffian:
The deploy script looks for `Tested up to:` in the style.css file, which isn't provided by all themes (for example the Varia children). Instead this PR proposes that we use the `Requires at least:` text which is provided by more themes to determine the required WordPress version.


mikachan (approved):
This makes sense to me 👍

PR #5238, opened 2021-12-20
Blockbase: converted headers to patterns

MaggieCabrera:

This PR is born from [this discussion.](https://github.com/WordPress/gutenberg/issues/37529) While this is not a full solution to the issue, it's an improvement in terms of user experience while keeping things DRY.

I've added the header template parts as block patterns, this has two benefits: we keep the "designs" encapsulated and we also allow users to choose the headers using a carousel (I don't know why but the carousel only appears when using block patterns instead of having multiple template parts). This approach is also aligned to how TT2 is handling headers.

Screen Capture on 2021-12-20 at 15-53-08

The spacer now is inside the template part for the theme and is no longer an extra template part.

This is a WIP for discussion, I didn't replace the child themes headers in case we don't decide to go this way.


scruffian:
This looks good. Would it be better to update Gutenberg so that the carousel effect works for multiple template parts?

MaggieCabrera:
> This looks good. Would it be better to update Gutenberg so that the carousel effect works for multiple template parts?

In my opinion, yes! I don't know why is not like that for template parts, we should ask the designers about this.

MaggieCabrera:
The carousel also doesn't work on replace, only when creating/adding a new header template part

MaggieCabrera:
The problem with this method is that upon inserting a new header, there is no spacer at all, so the user won't know to add it by default. I don't think it's terrible, since the default header will always have the spacer and they can add their own if/when they notice that the spacing changed.

MaggieCabrera:
I went ahead and refactored all the headers, I think that's all the themes and variations

mikachan (commented):
I don't have much to add here, but I do like seeing the headers as patterns. They feel nicely organized for the user and it sounds like this provides a better experience for choosing different headers. I've also given this a test on my local site and everything worked well for me ✨

scruffian:
@mtias what do you think of this one? Is this a good option for now, or should we look into adding the carousel for template parts into Gutenberg?

MaggieCabrera:
Are there other themes not in the repo that would need to apply the same change?

scruffian:
>Are there other themes not in the repo that would need to apply the same change?

Yes, pinging @Automattic/flow-patrol-manage for them

davipontesblog:
This will be super helpful for users!!!

> The carousel also doesn't work on replace, only when creating/adding a new header template part

I'd like to heavily suggest we try to add some sort of "Change layout" button somewhere in the Header and Footer blocks that would essentially work around that product gap by automating the tasks of removing the existing header and adding a new block so that the user gets that carousel. I don't think most users will be able to figure out by themselves that they need to remove+readd to change the header layout, and even if they do, the process to actually do that today in Gutenberg is cumbersome, and it involves "selecting the right parent level of the block" which is a big point of friction for users already today when there are multiple nesting levels present in block hierarchy.

pbking:
From a development perspective I actually like this much better than providing multiple template parts.  It provides the theme a SINGLE point to define the design of the header, rather than requiring that design decision to be defined in every template while still providing the various header options to all of the themes.  Additionally the "spacing" happens independent to the header design and in the wheelhouse of the theme's header template part so that it can be easily adjusted across all templates.

MaggieCabrera:
I rebased this branch, do you all think it's good to go with this?

pbking:
I would rather combine this and #5289 giving us the best of both Patterns and Parts.

pbking:
I attempted what I meant on this branch (this one being much closer to 'done' than #5289 and it was easier than rolling another branch) though if that doesn't jive we can roll this back to [your rebase](https://github.com/Automattic/themes/pull/5238/commits/f3fa2b3fa4c719b68a73a7210466bd3050bbab7e).

scruffian:
I think we should revert 39b225fbd357441c846fd688eceec1ad3f2dd802 and merge

MaggieCabrera:
> I think we should revert [39b225f](https://github.com/Automattic/themes/commit/39b225fbd357441c846fd688eceec1ad3f2dd802) and merge

I think when we deploy this we need to time it with the rest of the changes from the themes that are not on this repo, else we are most probably going to break them

scruffian:
> I think when we deploy this we need to time it with the rest of the changes from the themes that are not on this repo, else we are most probably going to break them

Are the other themes launched yet?

pbking:
I did a grep in the pub themes folder for any blockbase themes that might be using a header template part and besides those represented here in this change found none.

However there are some themes in /premium that would be effected and @MaggieCabrera  is right, those need to be changed at the same time to avoid errors.

I did some experiments changing my header and upgrading with this change (in as many scenarios as I could think of) and didn't find any workflows that would be harmed.  If I had already changed a template part and then upgrade, then reverted those changes then everything seemed to be working as I expected.

I believe that this change is ready to come in once we also have changes to the following premium themes:

byrne, kerr, alonso 

I expected 'payton' to be in that list as well as it's a Blockbase theme in /premium, however it already uses the default Blockbase header template part so that is good to go without change.

[UPDATE}
I see that all of those are using 'blockbase-premium' which isn't the same and could actually be taken care of in a separate change.  Pushing this change to /pub won't break anything in /premium.  We just need to ensure we make that change WHEN we push blockbase to premium (whatever that workflow actually is...)

scruffian:
Also I don't believe the themes in /premium have been launched yet

creativecoder:
> The carousel also doesn't work on replace, only when creating/adding a new header template part

This is not good--I think more users will be looking to replace the initial headers/footers provided by the theme, since they're already there, compared to creating new template parts. When there's only one header or one footer provided by the theme, the replace button doesn't appear on the template part block at all (replacing the template part with a variation was something users looked for in both recent FSE user tests that I did). 

As a work around, is there any detriment to having both a template part and pattern for each variation (as suggested above), until Gutenberg adds the carousel to the template part replacement flow?

alaczek:
I'd like to echo the concerns raised by @creativecoder and @davipontesblog - removing the template part is hard to do, and I don't think customers will know it's required to access different designs. 

Could we have both to bridge the gap as @creativecoder suggested?

pbking:
Just re-tested a flow (I can't believe I hadn't tested before) that confirms a site can break with this change if it has already been modified to use one of the header template-parts.

To reproduce, using /trunk activate Blockbase.
Using [Replace] change the header template part to use something other than the default.  (I chose "header-linear").

See that the header is fine on the frontend.
Now checkout this branch and reload the site.
Note that the header is completely gone in the view and is unavailable in the site editor:


I'm not aware of how to mitigate this without keeping the template parts available.

scruffian (approved):
LGTM

PR #5241, opened 2021-12-21
Blockbase: fix variables for 2 colored themes

### LINKED ISSUE #5232: Skatepark: Navigation menus text color
When using one of our custom palettes, the color of the submenu isn't the primary one:

Screenshot 2021-12-20 at 09 59 06

MaggieCabrera:

This PR changes the variables used for the colors in the navigation block. We were using the preset values instead of the custom ones: those may not always be 5, so for themes like Skatepark `--wp--preset--color--primary` was empty while `--wp--preset--color--primary` actually had an alternative defined.

Screenshot 2021-12-21 at 17 30 30

#### Related issue(s):

Closes https://github.com/Automattic/themes/issues/5232

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -1023,7 +1023,7 @@ p.wp-block-site-tagline {
 
 .wp-block-table.is-style-stripes figcaption,
 .wp-block-table figcaption {
-	color: var(--wp--preset--color--primary);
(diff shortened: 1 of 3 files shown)


MaggieCabrera:
I applied the same fix for the tables CSS, this wasn't causing a visual bug, but the variable was undefined for two-colored themes like Skatepark.

scruffian (approved):

PR #5245, opened 2021-12-22
Blockbase: refactor footers to patterns

### LINKED ISSUE #5025: Consider adding a search icon to the Geologist + Zoologist footers
Let's try adding a search to the footer of Geologist + Zoologist. It should: 

- Use `var(--wp--custom--font-sizes--tiny)` text (We might need a block style to accomplish this, since we don't want that to apply to all search blocks).
- Have half of the normal `var(--wp--custom--form--padding)` around the input field. 
- Be the width of the content column.
- Only be available on the archive pages, on single post/pages. 

I think we should give this a try in a PR and see what we think. To avoid messing with folks' live sites, we may need to implement this in a way that doesn't add it to existing non-FSE sites. 

Mockup: 

Screen Shot 2021-11-12 at 7 54 33 AM

MaggieCabrera:

This is a WIP

This PR repeats the work done for headers in https://github.com/Automattic/themes/pull/5238 but for footers. We are removing duplicate code and giving the user options to use all possible footers designed between our themes no matter the theme you installed.

I have a couple of comments I'd like to discuss:

- I've not wrapped the "Proudly Powered by Wordpress" texts for translation because we have the plugin on dotcom to do that for us. I think it would be safe to do it anyway, let me know what you think.
- Should we make the pattern from https://github.com/Automattic/themes/issues/5025 be the default footer for Geologist and Zoologist even though those are already launched? Or should we just keep them as a pattern that users can swap to if they want?
- Skatepark has some text under the title: 

> Skatepark's coaches will work with you to develop and improve your skating abilities.

Should we change this text to something generic that works for all themes? Or maybe we should make this be the tagline block instead? @kjellr @beafialho  I'd love your input on this

Seedlet blocks footer was very different from the original, I tried to make it closer to what it should be. Also ended up opening https://github.com/WordPress/gutenberg/pull/37586 because it's related, I think it's fine if that doesn't get merged though.

Screen Capture on 2021-12-22 at 13-20-48

The new pattern on geologist with the small search block block style (I didn't apply it to the theme, just made the screenshot, this pattern is not used in any theme at the moment in this PR, pending discussion):

Screenshot 2021-12-22 at 13 42 05

#### Related issue(s):

I'm adding the extra pattern from https://github.com/Automattic/themes/issues/5025 

Closes https://github.com/Automattic/themes/issues/5025


kjellr:
> Should we change this text to something generic that works for all themes? Or maybe we should make this be the tagline block instead? @kjellr @beafialho I'd love your input on this

Yeah, the site tagline there is fine.

MaggieCabrera:
Rebased this one too. Changed the text on the skatepark pattern to the tagline.

@kjellr should we use the new pattern from #5025 on Geologist/Zoologist?

scruffian (approved):
LGTM

PR #5251, opened 2021-12-23
Blockbase: Update customizer PHP library to work with latest version of Gutenberg

scruffian:
Gutenberg now expects customizations to be explicitly stored under `custom` in the Global Styles. This updates the PHP files to store the customizations under `custom`. To test, try changing fonts and colors in the Customizer.


MaggieCabrera (approved):
This works for me on all my test environments locally and on dotcom

PR #5254, opened 2021-12-24
Generate blockbase package.json files

scruffian:
This adds a script to update the package.json files for all Blockbase children. To run:

1. cd blockbase
2. npm i
3. node generate-children.mjs

The package.json from Blockbase should be updated and added to the children.

The idea of this is two fold:
1. If we want to make changes to these package.json files, we only need to do it in one place and this script keeps the other themes in sync
2. In the future this can be expanded to generate other parts of the theme, and ultimately the whole theme.

### DIFF
--- a/blockbase/generate-children.mjs
+++ b/blockbase/generate-children.mjs
@@ -1,4 +1,4 @@
-import fs from 'fs';
+import { promises as fs } from 'fs';
 import deepmerge from 'deepmerge';
 import { dirname } from 'path';
 import { fileURLToPath } from 'url';
@@ -10,6 +10,11 @@ const localpath = dirname( fileURLToPath( import.meta.url ) );
 	return await generateChildren();
 })();
 
+async function getPackageJson( directory ) {
+    const packageJsonString = await fs.readFile( directory + '/package.json', 'utf8' );
+    return JSON.parse( packageJsonString );
+}
+
 async function generateChildren() {
 	const children = [ {
 		"name": 'Skatepark',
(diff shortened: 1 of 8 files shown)


blockbase/generate-children.mjs
```diff
@@ -0,0 +1,73 @@
...
+    return JSON.parse( packageJsonString );
+}
+
+async function generateChildren() {
+	const children = [
+		{
+			"name": 'Geologist',
+			"description": "Geologist is a streamlined theme for modern bloggers. It consists of a simple single column of posts, paired with a sophisticated color palette and beautiful sans-serif typography.",
+			"slug": 'geologist'
+		},
```
MaggieCabrera, line 24: we could maybe pick this up from the stylesheet of the themes? and just give an array of the themes that are the current children

scruffian: I did consider that, but I think ultimately this script would also generate the style sheets.

pbking:
I'm not sure what problem this solves...

> If we want to make changes to these package.json files, we only need to do it in one place and this script keeps the other themes in sync

I haven't noticed this as a thing we have had to do before.

> In the future this can be expanded to generate other parts of the theme, and ultimately the whole theme.

> Don't write code to solve problems you don't have yet.
-somebody somewhere

I think that the "generate theme" idea is covered well enough by the "variations" generation.  If there's an aspect to that which should be expanded then perhaps we should look to address that workflow.  Beyond "variations" I'm not sure that generation of Blockbase things is a thing we need to work towards.

scruffian:
> I haven't noticed this as a thing we have had to do before.

It's evident from that fact that these package.json files are all slightly different, that our current process for keeping them in sync isn't very effective :)

> Beyond "variations" I'm not sure that generation of Blockbase things is a thing we need to work towards.

At the moment we try to use the parent/child mechanism to solve our needs to have inheritance between themes. This causes a bunch of problems (see pNEWy-eAd-p2).

I think theme generation will help us solve these problems in a different way, allowing us to build themes that are optimised for the user rather than for us as developers.

pbking:
> It's evident from that fact that these package.json files are all slightly different, that our current process for keeping them in sync isn't very effective 

I would say that the slight differences in the package.json files shows that slight differences in package.json files doesn't effect much of anything.  We haven't a process for keeping package.json files in sync with Blockbase because we haven't yet had a need to. 

> I think theme generation will help us solve these problems in a different way, allowing us to build themes that are optimised for the user rather than for us as developers.

This doesn't change the result though.  Child themes are still child themes, but the package.json file is more similar to Blockbase's.

I see this as more like getting a haircut when you need to walk a mile.  It might make you look better walking but it doesn't get you there any faster.

I agree with how the problem is stated (as referenced above).  I just don't (yet) see how this work addresses any aspect of that problem.

scruffian:
I'm going to close this for now. I'll follow up with another idea. Thanks for the feedback!

PR #5265, opened 2022-01-04
Stewart: Tidy up navigation

kjellr:
This PR takes are of a few things: 

- It makes sure the navigation's overlay color matches the page background. Previously it was white. 
- It removes an unnecessary `1px` spacer block in the sidebar template. 
- It removes the sidebar template from the "sidebar" area to "uncategorized". Sidebar isn't a valid area at this point, and this was throwing an error message in the Site Editor.  

Aside from the overlay color change, things should basically look the same. Testing might look weird because of these two upstream bugs: 

- https://github.com/WordPress/gutenberg/issues/37695
- https://github.com/WordPress/gutenberg/issues/37694

### DIFF
--- a/stewart/parts/sidebar.html
+++ b/stewart/parts/sidebar.html
@@ -9,11 +9,7 @@
 <hr class="wp-block-separator is-style-wide"/>
 <!-- /wp:separator -->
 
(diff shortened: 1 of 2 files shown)


scruffian (approved):
LGTM

PR #5268, opened 2022-01-04
[Livre] Refine site title treatment

### LINKED ISSUE #5217: [Livre] Look into using small caps instead of italics
Look into opportunities to use small caps in Livre.

kjellr:
Closes #5217.

This PR adds small caps to the site title, and also introduces a little CSS to track out the letters when all-caps is active. It adds a new `x-small` font size to use in this context. 

Before|After
---|---
Screen Shot 2022-01-04 at 14 10 56|Screen Shot 2022-01-04 at 14 06 53


livre/parts/header.html
```diff
@@ -4,5 +4,5 @@
 <!-- wp:page-list {"isNavigationChild":true,"showSubmenuIcon":true,"openSubmenusOnClick":false} /-->
 <!-- /wp:navigation -->
 
-<!-- wp:site-title {"style":{"typography":{"fontStyle":"italic","fontWeight":"400"}}} /--></div>
-<!-- /wp:group -->
\ No newline at end of file
+<!-- wp:site-title {"style":{"spacing":{"padding":{"top":"0.35ch"}},"typography":{"textTransform":"uppercase"}}} /--></div>
```
kjellr, line 7: The padding here is a little weird, but is necessary to ensure that the text properly aligns vertically next to the menu button, since there are no descenders when the text is all caps. 

I figured putting it here is better than in CSS since the user can remove it if they don't want it anymore.

scruffian (approved):
LGTM

kjellr:
I pushed one more small change (d243a00) that updates the theme screenshot to match this new header treatment. Should be good to go. 👍

Issue #4580, opened 2021-09-14
Multiple Themes: Buttons in sidebar no longer apply correct colours since Block Widget Editor introduced.


poojashetty18:
## Steps to replicate

- Activate the Canard theme.
- Add the Follow blog widget.
- Update the theme colors from Colors and Background. 

![Screen Capture on 2021-09-14 at 15-08-29](https://user-images.githubusercontent.com/44837596/133253348-5a212283-1c99-4a29-b07c-ad7cfbf9dc5f.gif)

## Result
The follow blog widget color does not update. 

## Expected

The theme color should be reflected correctly on the follow blog widget and the user noticed this issue a month ago when they noticed the Gutenberg block widgets enabled.

* User report: 11041308-hc
* Follow-up ticket: 4297015-zen (the user wants to let them know once it's fixed)

mikeicode:
The issue appears to be that this button now has the class `wp-block-button__link`. This apparently unintentionally changes the style of the buttons on this block.

<img width="1328" alt="Screen Shot 2021-09-14 at 07 21 10" src="https://user-images.githubusercontent.com/10353785/133440377-0f95643c-416c-47c2-b61c-6384699e8377.png">


I've also run into this issue on another theme Maxwell (retired theme).

DavidSaladu:
I've put the ticket on hold for now, in this case, should this be corrected at the theme level? (the customer is under personal plan so we can't patch it with CSS)

Robertght:
I remember from @sixhours there's a core issue regarding widget updating, but I'm not sure if it's related to this.

jeffikus:
@danieldudzic @mikachan can either of you confirm this is theme related please?

mikachan:
I've just taken a look at this and I can see what you're seeing, @mikeicode, where the `wp-block-button__link` class is being applied and overriding the correct/current button color. I've tried a few different themes and I can see the same issue in each one, where the theme colors aren't applied to the Follow button.

I would guess this means it's not theme related, although I'm struggling to find a related issue in Gutenberg (there's a lot of issues related to buttons!). I'll report back if I find one.

jordesign:
Triaging this and it seems apparent this is effecting multiple themes and should be considered more holistically. At this stage reports are for:

* Canard
* Maxwell
* TwentySixteen

The bug effects any buttons which are included in a widget area (including those automatically included as part of widgets like the 'Follow' widget). This may have come into place when the Block Widget Editor was released.

[...]

PR #5272, opened 2022-01-05
Canard: fix button styles outside of entry-content

mikachan:
This applies the default button styles to all button blocks on Canard, instead of only targeting those in `entry-content`. This means that any buttons added outside of `entry-content`, such as in a widget area, will be styled the same as any other button blocks.

I'm not sure why these styles were specific to `entry-content`, so I don't know if this might break something else.

I've also added `!important` to the `background-color` and `color` properties of the default styles on the button link for the active, focus, and hover states. This means we should always have decent color contrast in these states when custom colors have been added via the editor.

To test, add a button to a widget area. Make sure this button is styled the same as a button in a page or post, and that custom styles can still be applied correctly.

#### Related issue(s):
Fixes part of #4580.

### DIFF
--- a/canard/blocks.css
+++ b/canard/blocks.css
@@ -236,6 +236,13 @@ p.has-drop-cap:not(:focus)::first-letter {
 	color: #fff;
 }
 
+.wp-block-button .wp-block-button__link:active,
+.wp-block-button .wp-block-button__link:focus,
+.wp-block-button .wp-block-button__link:hover {
+	color: #d11415;
+	background: #fff;
+}
+
 .is-style-outline .wp-block-button__link {
 	background: transparent;
 	border-color: currentColor;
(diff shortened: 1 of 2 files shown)


scruffian:
This doesn't seem to solve the problem for me:

Screenshot 2022-01-06 at 10 33 50

I think we need to update the color annotations which target the selectors we need to change when a user changes the colors.

mikachan:
> I think we need to update the color annotations which target the selectors we need to change when a user changes the colors.

Ahh I see, I think I misunderstood completely when I was working on this 🙈

I've updated the color annotations now so the colors of the buttons are changed with the color palette. I've also tidied up the CSS so the colors for outline buttons are more consistent.

scruffian (approved):
LGTM

PR #5273, opened 2022-01-05
[Livre] Clean up buttons and comment form

### LINKED ISSUE #5267: [Livre] Revise default button style
Buttons in Livre currently look like this:

Screen Shot 2022-01-04 at 2 15 27 PM
 
This doesn't match up well with the sophisticated quality of the rest of the theme. We should revise.

kjellr:
Fixes #5267.

This PR revises the default button style: 

Before|After
---|---
Screen Shot 2022-01-05 at 10 44 57 AM|Screen Shot 2022-01-05 at 10 42 42 AM

@beafialho what do you think of that style for the buttons? I'd kind of prefer to make the default button style an outline style, but I'm not sure if we can do that technically. So this seems like the next best solution. 

---

This PR also provides style + spacing fixes for the post comments block. I think it's appropriate and fine to do these changes in CSS, since it's not possible to modify the styles and spacing of anything within that block at the moment. 

Before|After
---|---
Screen Shot 2022-01-05 at 10 48 18|Screen Shot 2022-01-05 at 10 48 34

### DIFF
--- a/livre/style.css
+++ b/livre/style.css
@@ -84,6 +84,11 @@ a:active {
 	opacity: 0.90;
 } 
 
+.wp-block-button__link:hover, .wp-block-button__link:focus, 
+.wp-block-button__link:active, .wp-block-button__link:visited {
+	color: initial;
+}
+
 /*
  * Alignment styles.
  * These rules are temporary, and should not be relied on or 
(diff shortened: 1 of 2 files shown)


beafialho:
> @beafialho what do you think of that style for the buttons? I'd kind of prefer to make the default button style an outline style, but I'm not sure if we can do that technically. So this seems like the next best solution.

On top of that, in the comments section for instance, an outlined button would probably not jump out as well because the fields are also outlined. So I think this style works best.

livre/style.css
```diff
@@ -136,4 +141,40 @@ body > .is-root-container,
...
+/*
+ * Comment form cleanup.
+ */
+
+input {
+	font-family: inherit;
+}
+
+textarea,
+input:not([type="submit"]) {
```
scruffian: Should this also be :not([type="button"]) ?

scruffian (approved):
One minor comment. LGTM

livre/style.css
kjellr: Yes, it should. Thanks for catching that! Added in bf2ce20.

PR #5279, opened 2022-01-07
Livre: Update alignment rules

kjellr:
This PR updates the alignment rules to include Cover blocks as well. This syncs it back up with the alignment code in Twenty Twenty-Two. Cover block styles for these rules were left out of Twenty Twenty-Two by mistake, and added in https://github.com/WordPress/twentytwentytwo/pull/310.


scruffian (approved):
LGTM

PR #5280, opened 2022-01-07
[Livre] Update readme and copyright

kjellr:
This PR just cleans up the style.css and readme files so they're accurate. 🙂


pbking (approved):
👍

PR #5281, opened 2022-01-07
[Livre] Add header-footer-only template.

kjellr:
This PR adds a custom template to Livre that contains just a header and footer.

Screen Shot 2022-01-07 at 9 33 42 AM

_(Note that the footer won't actually show up yet since it relies on a block pattern that doesn't exist — I'm working on fixing that in a separate PR)._


pbking (approved):
Tested with #5282.
Approved when paired with that.

PR #5283, opened 2022-01-10
Bennett: Initial commit.

alaczek:
This is the initial commit of Bennet - a super simple Blockbase child theme featuring a portfolio on the homepage. 

I included a header template part - I copied the minimal header from Blockbase, because I wanted to add some custom styling to the site title. Please let me know if there's a better way to do it!

This is an updated version of Bingley, which will be retired when Bennett launches. More info in pNEWy-exS-p2


alaczek:
I just realized... Do I need to copy over all the templates from Blockbase to ensure the minimal header template part is used?

scruffian:
> I just realized... Do I need to copy over all the templates from Blockbase to ensure the minimal header template part is used?

If you want to use the one that's in Blockbase then there's no need to copy the files as this will inherit that template. If you want to make changes to it then you'll need to duplicate it and then make your changes...

alaczek:
Yeah, I figured it out. I ended ditching the customizations in the header and using one of the included ones. Could you take a look?

scruffian:
I added a commit which should simplify this somewhat! If you're happy with the change feel free to merge...

PR #5285, opened 2022-01-10
Deploy script: remove double quotes from commit message log

mikachan:
This removes any double quotes from the commit message logs in the deploy script.

This is to fix an error @MaggieCabrera and I saw where a revert commit from GH included double quotes in the commit message, and because they were not escaped they caused the script to error.


pbking (approved):
LGTM!
Thank you for fixing the source of the problem!

PR #5289, opened 2022-01-10
Refactored headers to be both patterns and parts

pbking:
[This bit](https://github.com/Automattic/themes/blob/e425b1bcf0f00c3fcd384a5c2a84bb865b45681c/blockbase/inc/block-patterns.php#L36-L56) is the crux of the matter, loading existing Block Template Parts as Patterns.

By loading the Template Parts as Patterns we now have the option to include the version of the header that we want in any child themes in one place ([/parts/header.html](https://github.com/Automattic/themes/pull/5289/files#diff-a1620d7b1f2a7a3f3d480e4570d03ab894b7413cee3fa6913a592fefe40b7f02)) which is important for our variations.

Additionally these headers are available to the user as patterns:

And by keeping those as Template Parts gives the user the option to EASILY REPLACE the header template part.

Inspired by #5238


pbking:
Oh, and as an added bonus the header patterns get to stay in HTML files!  A great boon to the developer!

pbking:
Closed in deference of #5238

PR #5290, opened 2022-01-10
Blockbase: update dependencies & change node-sass to dart sass

mikachan:
This updates the node dependencies for Blockbase & co. It also swaps node-sass for [dart sass](https://sass-lang.com/dart-sass), as node-sass has been deprecated.

I've also added postcss to the themes that didn't already have it, as I believe this is handling the browser prefixes, like -webkit- etc.


MaggieCabrera:
I'm using node v16.13.1 and I'm getting a different package-lock.json than what's on this PR. I'm also getting this when I do `npm i`. I'm on v16 because that's what our docs for the deploy script are recommending.

```
npm WARN old lockfile
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile
npm WARN old lockfile This is a one-time fix-up, please be patient...
npm WARN old lockfile
```

It seems to be building the CSS fine, I tried it on Blockbase and it's removing the -webkit prefixes from lines like:

```
a:hover, a:focus {
  -webkit-text-decoration-line: underline;
}
```

scruffian:
> I'm getting a different package-lock.json

Me too

>  it's removing the -webkit prefixes from lines like

Is this because these are deprecated?

MaggieCabrera:
> > it's removing the -webkit prefixes from lines like
> 
> Is this because these are deprecated?

According to [can I use](https://caniuse.com/?search=text-decoration-line), yes, but we should rebuild all the themes and update the CSS in this PR so we don't get surprise CSS when we build for other PRs

mikachan:
> I'm using node v16.13.1 and I'm getting a different package-lock.json than what's on this PR. 

Ah that's because I'm using node 14, and I think this was because I couldn't get node-sass working with node 16. I'll update the package-lock files for node 16.

Yeah, I think it's removing the -webkit prefixes because the `@wordpress/browserslist-config` has been updated.

mikachan:
I've run the build command in each theme so the CSS should all be up to date, but I'm not seeing the -webkit prefixes being removed.

I'm clearing cache and restarting - wondering if changing node versions is giving me different results..

The prefixes being removed sounds correct I think!

mikachan:
Hmm, it's postcss that's adding (or keeping) the browser prefixes. As far as I know, this uses browserslist, so I think this should be the version we go with as it's based on `@wordpress/browserslist-config`.

I've simplified the Blockbase build script so it's similar to the children, where `build` runs `build:scss`, and then `build:scss` always runs postcss.

pbking:
Change looks good.  Thanks for this change!  Only comment is same as @MaggieCabrera's, in that those CSS changes that this change will cause should come in now.

pbking:
In a similar vein, while tidying up package files I would like to remove them from the variations.  Their presence implies something that isn't true.

pbking:
Only found one small prefix change that still needed committing.  

(Once for blockbase and two other small places in a couple of children).

I built completely with both @14 and @16 versions of node and the minor changes I committed were all I found.

This change looks good to me!

🚢

pbking (approved):

scruffian (commented):
Is there a reason we switched from tabs to spaces as the indent format?

mikachan:
> Is there a reason we switched from tabs to spaces as the indent format?

I don't think dart sass supports tabs, I couldn't find an option for it in the [cli docs](https://sass-lang.com/documentation/cli/dart-sass).

scruffian (approved):
Makes sense. Let's merge then!

mikachan:
Sorry to pull this back from approved - I noticed some of the dependencies between the blockbase children weren't the same, so I've synced them up. I should've checked this before.

It's much the same as before but with updated package.json files, and all assets should be the same.

scruffian:
When I run this now I get a bunch of diffs still - the browser prefixes are showing again.

mikachan:
Hmm, so I think this is because the `autoprefixer` and `@wordpress/browserslist-config` weren't included in every theme, and now they are it means the browserlist config is being used consistently.

I've just found out you can run `npx autoprefixer --info`, which will list the browsers we're supporting (from the config) and which prefixes are needed, and this is what I got: https://pastebin.com/udhVChuy

The prefixes that have been added on the most recent run for me are text-decoration-line for webkit, which is in the above list as needing a prefix... so I'm pretty sure this is correct, and it looks like all themes are seeing the same changes now.

I've run a build again for all Blockbase themes and committed the CSS, hopefully we'll see the same results now!

scruffian (approved):
LGTM

pbking:
I fixed up the conflicts (in package.json files) and re-installed dependencies and rebuilt all of the themes to address the rest of the conflicts.

Added an npm script to easily rebuild everything with an existing utility: `npm run back:build` to go along with the `npm run batch:install`

This is looking really solid now.  Going to bring it in and continue the deploy.

PR #5291, opened 2022-01-10
Update dependencies for Dependabot alerts

mikachan:
This updates the node dependencies for the themes that have been flagged by [Dependabot](https://github.com/Automattic/themes/security/dependabot).

We should probably update everything else as well, but I thought I'd start with these. I'm not sure how we usually update things like this!

I've updated Blockbase dependencies in a separate PR: https://github.com/Automattic/themes/pull/5290


MaggieCabrera:
Thank you for doing this!

Same as #5290 I'm getting a different lock file when installing this with node 16. Pastebin [here](https://pastebin.com/pjNL6kfG) for reference with the warnings I get (it's too long to post here)

scruffian:
I think there's a script that does this, did you use it?

scruffian:
It's `npm run batch:audit:fix`

mikachan:
> Same as #5290 I'm getting a different lock file when installing this with node 16.

I've updated the package-lock files for node 16 now!

> I think there's a script that does this, did you use it?

Ah I did see this, but I didn't end up using it as I thought I may as well update all packages, rather than just those with known security vulnerabilities (which I think is what `npm audit fix` does). But then this meant I only updated specific themes manually, rather than running something like this over all the themes.

I've just tried running `npm audit fix` for Alves as a test, but I had to run `npm audit fix --force` to allow it to update packages to major versions. This updated all top-level dependencies apart from 2 ("@wordpress/browserslist-config" and "chokidar-cli") because they have no reported security problems. But we should probably be keeping these up to date regardless. Plus if our top-level dependencies are up to date, this would probably mean `npm audit fix` would work without the --force flag most of the time.

So, I think we also need a script that updates our top-level dependencies. I've added this in my last commit. The problem with this is that it will update major versions of packages without us knowing if it breaks something (like theme build scripts etc). But then I think it's important we generally keep things up to date, so I'm not sure what the best solution is! I could maybe move this new script to another PR and run it for all themes, and handle it separately?

scruffian:
>  I could maybe move this new script to another PR and run it for all themes, and handle it separately?

Sounds like a good idea

mikachan:
Ok thanks! I'll create another PR for that.

This PR should be ready for another review, and it should be node 16+ ready now.

scruffian (approved):
LGTM

PR #5293, opened 2022-01-10
Farrow: Initial commit.

alaczek:
This is the initial commit of Farrow - a super simple Blockbase child theme featuring an about me content on the homepage.

This is an updated version of Pollard, which will be retired when Farrow launches. More info in pNEWy-exS-p2


farrow/templates/index.html
```diff
@@ -0,0 +1,32 @@
+<!-- wp:template-part {"slug":"header","tagName":"header","layout":{"inherit":true}} /-->
```
scruffian: It looks like the only difference between this template and the one in Blockbase is that the header inherits the default layout. My preference would be to move that inside the header itself (creating a new header which inherits layout). This would give users more header template options, and limit unnecessary variations to our templates.

scruffian:
Actually looking at this some more I think we can just remove the extra index.html file...

alaczek:
I think it's ready to merge now?

scruffian (approved):
LGTM

PR #5294, opened 2022-01-10
Calvin: Initial commit.

alaczek:
This is the initial commit of Calvin - a super simple Blockbase child theme featuring a portfolio on the homepage.

This is an updated version of Glen, which will be retired when Calvin launches. More info in pNEWy-exS-p2


scruffian (approved):
LGTM

scruffian:
I added a commit to simplify the way headers work.

PR #5295, opened 2022-01-11
Revert "Revert Fix Customizer Font Editing causing Site Editor WSOD"

scruffian:
Reverting the revert: Automattic/themes#5284

### DIFF
--- a/blockbase/inc/customizer/wp-customize-fonts.php
+++ b/blockbase/inc/customizer/wp-customize-fonts.php
@@ -459,9 +459,9 @@ class GlobalStylesFontsCustomizer {
 		}
 	}
 
-	function unset_property_if_it_exists( $property ) {
-		if ( isset( $property ) ) {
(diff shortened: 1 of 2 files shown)


scruffian:
This was causing this error: 

```Fatal error: Uncaught Error: Cannot use object of type stdClass as array in .../wp-content/themes/pub/blockbase/inc/customizer/wp-customize-utils.php:18```

I have been unable to replicate this. I can only assume it happens when the data is saved with an old format. I have added a commit which will cast objects to arrays before we modify them, which I hope will address the issue.

pbking:
I went through all my version change dancing and didn't reproduce the Fatal nor did I see any problems with the code that was added.

That change seems like a good enough change though I wish we had a better way to reproduce the error to know it fixed it.

When it ships we should be sure it's when we can give the logs careful attention for a bit.

pbking (approved):

PR #5299, opened 2022-01-11
Kerr and Payton: Removed unused files

scruffian:
These themes were removed but these files are still hanging around. This PR removes them.


pbking (approved):
👍

PR #5300, opened 2022-01-11
Font size names update

### LINKED ISSUE #5209: Blockbase + family: Update font size names
Gutenberg has update the font size names: https://github.com/WordPress/gutenberg/pull/37381/files

We should follow suit in Blockbase

matiasbenedetto:

Font size name update following the Gutenberg change from this PR https://github.com/WordPress/gutenberg/pull/37381/files

Changes made for Blockbase children themes:
- 'huge' to 'x-large'
- 'tiny' to 'x-small'
- Move 'normal' from presets to custom

#### Related issue(s):
https://github.com/Automattic/themes/issues/5209


pbking:
That change also gets rid of 'normal'.  Should we follow suit here to match as well?  Should it move into the same 'custom' space as --tiny?

Since Gutenberg now uses "t-shirt sizes" (x-large instead of huge) does it make sense for Blockbase to also follow suit on that naming convention?  Changing our custom (but not surfaced as a user-select-able size) --tiny to --x-small?

I'm not saying we should do either of those, just asking.

scruffian:
I agree with both of those suggestions

jeffikus:
> @pbking does it make sense for Blockbase to also follow suit on that naming convention? 
It does to me. What are the implications for existing users?

matiasbenedetto:
I think this is another reason to get rid of the `custom` font size definitions and the need to move them to  `settings`:

Currently on **Blockbase** `theme.json` we have 5 font sizes at this path `settings.typography.fontSizes` and 1 font size (called `tiny`) defined at `settings.custom.fontSizes` .

The definitions at `settings.typography.fontSizes` are loaded by Gutenberg and added as options available to the user but that's not true for the option defined at `settings.custom.fontSizes `.

If we change the font size of an element using the font size defined in `custom` (`tiny` in this case) there is not a simple way to back to use `tiny` as font size  (see the screenshot) . I tried changing the font size of `Site Tagline` block which was using `tiny` and I don't see the way to go back to `tiny`.



image

scruffian:
The problem we have is that Gutenberg only defines 4 sizes and our designs use more than 4. If we add more sizes to the theme.json definition and then a user implements them on a page, their site becomes connected to those "non-standard" font size definitions. If they then switches to a theme that doesn't implement those sizes then their "non-standard" named sizes will default to normal. 

That's why we keep "tiny" inside custom - it allows us to use it in designs, without letting the user select it and having problems down the road.

IMO the best way to solve this would be to have a defined naming convention in Gutenberg that all themes can use, but there has been resistance to this in the community.

pbking:
> IMO the best way to solve this would be to have a defined naming convention in Gutenberg that all themes can use ...

And also to note that as of that Gutenberg change noted we also are down from FIVE to FOUR 'standard' size definitions that we MIGHT be able to count on and one of the remaining four now has a different slug.

matiasbenedetto:
I added the changes suggested and updated the PR description.

scruffian (approved):
LGTM

PR #5303, opened 2022-01-12
Remove customizer restoration as it is no longer being removed by Gutenberg.

### LINKED ISSUE #5131: WP Admin > Appearance shows Editor (Beta) link twice.
## Steps to replicate
Howdy, I just noticed on my test site, that when Geologist theme is installed and the Gutenberg plugin is active on a WordPress 5.9 beta site, the Appearance menu has the Editor (beta) link twice. 

## Result

Screen Shot 2021-12-01 at 5 36 52 PM

## Expected
to only see it once :-)

## Version: 
WordPress 5.9-beta1-52296 
Gutenberg Nightly (trunk from today) 
Geologist 1.0.15

pbking:

This change **removes** the **remove_action** to 'gutenberg_remove_legacy_pages'.  

Because this call was being REMOVED by Blockbase there were two links to the Editor being shown when running WordPress 5.9.

This was evaluated using WP 5.8.2 (and Gutenberg 12.3) and found that removing this remove_action didn't have negative side effects: The customizer is still available.  It seems that Gutenberg is no longer hiding that tool when a Block-based theme is detected.

#### Related issue(s):

Fixes #5131

### DIFF
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -150,14 +150,6 @@ function blockbase_fonts_url() {
 }
 
 /**
- * Restores the Customizer since we still rely on it.
- */
-// function blockbase_restore_customizer() {
-// 	remove_action( 'admin_menu', 'gutenberg_remove_legacy_pages' );
-// }
-// add_action( 'init', 'blockbase_restore_customizer' );
-
-/**
  * Customize Global Styles
  */
 if ( class_exists( 'WP_Theme_JSON_Resolver_Gutenberg' ) ) {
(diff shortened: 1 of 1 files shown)


mikachan (approved):
This works great for me. 

I can confirm that when using 5.9 I can only see one Editor link, but without this fix I can see two. And then when testing with 5.8.3 I can't see any errors, and the Customizer is still available.

scruffian (approved):
LGTM

PR #5304, opened 2022-01-12
Update all dependencies and add update script

mikachan:
This PR adds a script that handles updating all themes' dependencies. We already have a script that runs `npm audit fix` on all themes, but this usually doesn't work if a top-level dependency needs updating to a major version (it will say you need to add the --force flag). My thinking was that if we had an update script as well, it'd be easier to keep all top-level dependencies up to date, which should make running `npm audit fix` more successful as well.

I also ran the `build-all` script to update all the theme assets.


scruffian:
I ran this again and got a package_lock file change for some more themes. It looks ok to me now, but maybe you want to take another look?

mikachan:
Thanks @scruffian! Not sure how I managed to get the barnsbury package-lock in that state, looks like alves was similar so I've updated that as well.

I've also run the 'build all' script so we've got updated assets for all themes that need them.

scruffian (approved):
LGTM

pbking:
All of my package-lock.json files changed when I ran this, upgrading all of them to version 2 format.

Probably because I have a different version of NPM installed now?  I'm not of the opinion that the differences matter.  I usually just ignore package-lock.json changes.

Otherwise this looks good to go to me too.  Thank you very much for doing this @mikachan!

scruffian:
Is there a way to ensure we're all running the same node version?

mikachan:
> All of my package-lock.json files changed when I ran this, upgrading all of them to version 2 format.

Ah really? Hmm, when I ran the update script I must've still been using node 14. I meant to run everything on 16 so they all got an update! I'll commit the updated package-locks.

> Is there a way to ensure we're all running the same node version?

I think we can do this using a `.nvmrc` file.. I'll check! Although this assumes we're all using NVM.

mikachan:
I've added a `.nvmrc` file for review. Here's the [docs](https://github.com/nvm-sh/nvm/tree/02997b0753f66c9790c6016ed022ed2072c22603#nvmrc), but from my understanding, this is how we use it:

- Add a version number to `.nvmrc`. This can be a specific version, including the minor/patch numbers, or a major version (I've just gone with '16' for now)
- All of us should run `nvm use` locally to ensure we're all using the version specified in this file for this repo

This assumes we're all using NVM to manage node versions..

pbking:
> This assumes we're all using NVM to manage node versions..

I don't use NVM to manage node.  Perhaps I should but as a rule I lose half a day when I try to tinker with node versions so I change very rarely!

I think that node version expected can be expressed in [package.json](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines) and I'm not sure why nvm wouldn't refer to that instead...

I'm impartial and am fine with the expected version being expressed somewhere, and preferably ONE place.

mikachan:
> I don't use NVM to manage node. Perhaps I should but as a rule I lose half a day when I try to tinker with node versions so I change very rarely!

This is basically what's happened to me this week! But I feel all loyal to NVM because I've always used it..

Ah OK, maybe we should just specify it in the root package.json then? Then even if it's not enforced, we at least have a single reference for it.

scruffian (approved):
Let's do that and then bring this one in too..

mikachan:
Done! Thanks both.

scruffian (approved):
LGTM

PR #5309, opened 2022-01-13
AltoFocus: set correct value for `featured-content[show-all]` option.

david-binda:

The expected type of the `featured-content[show-all]` option is a string, not an integer.


scruffian (approved):
LGTM. Thanks for doing this.

PR #5108, opened 2021-11-26
Git moving templates and parts directories

matiasbenedetto:
#### Changes proposed in this Pull Request:
Git moving folders:
block-templates -> templates
block-template-parts -> parts
Following the new Gutenberg convention: WordPress/gutenberg#36647

#### Related issue(s):
https://github.com/Automattic/themes/issues/5105

PR #5311, opened 2022-01-13
Added a temp script to move block theme folder locations

### LINKED ISSUE #5105: Blockbase: Update template directories

pbking:
Opened this as an alternative to #5108.  Since #5108 needs to be rebased to come in it becomes more complicated of a transition the longer it stalls.  

Instead we can run this script right before we bring the change in which ensure that we're handing and moving the latest files.

The script should be deleted after it has been ran.

The script just copies any folders from old format to new.  It could probably be written more elegantly but this was the fastest path to done.

Run: `./move-block-template-folders.sh`
Validate that the changes are correct.
When merging this change the changes made by the script should be committed, the script deleted.

#### Related issue(s):

Resolves #5105 which depends on this Gutenberg change: WordPress/gutenberg#36647 which is in 12.1


scruffian (approved):
LGTM

PR #5314, opened 2022-01-13
Remove templates improperly nested in a folder due to script error in #5105

pbking:
Remove templates improperly nested in a folder due to script error in #5105


scruffian (approved):
LGTM

PR #5315, opened 2022-01-13
[Livro] Update alignment rules

kjellr:
This PR updates Livro's alignment rules to better handle Group blocks that have a background color as well as Cover blocks. 

When these blocks are added to the top level of the site, or directly within a Template Part at that level, they'll now go edge-to-edge by default. This feels natural, and allows folks to easily get full-width headers and footer backgrounds without any workarounds. 

The PR is a straight-up copy of [a recent update to Twenty Twenty-Two](https://href.li/?https://github.com/WordPress/twentytwentytwo/pull/336/), so you can read about the change there too if you'd like. All of these alignment styles are just copied/pasted from there. The same testing instructions apply here. 

Before|After
---|---
Screen Shot 2022-01-13 at 2 45 24 PM|Screen Shot 2022-01-13 at 2 44 53 PM

### DIFF
--- a/livro/style.css
+++ b/livro/style.css
@@ -105,6 +105,7 @@ a:active {
 body > .is-root-container,
 .edit-post-visual-editor__post-title-wrapper,
 .wp-block-group.alignfull,
+.wp-block-group.has-background,
 .wp-block-cover.alignfull,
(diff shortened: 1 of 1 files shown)


scruffian (approved):
LGTM

PR #5318, opened 2022-01-13
[Stewart] Add alignment rules

kjellr:
This PR adds the Twenty Twenty-Two alignment rules to Stewart. Previously, there were no alignment rules here at all.

### DIFF
--- a/stewart/theme.json
+++ b/stewart/theme.json
@@ -105,6 +105,9 @@
 			"margin": {
 				"horizontal": "30px",
 				"vertical": "30px"
+			},
+			"spacing": {
+				"outer": "min(1em, 30px)"
 			}
 		},
 		"layout": {
@@ -341,9 +344,6 @@
 					"text": "var(--wp--preset--color--foreground)"
 				}
 			}
-		},
-		"spacing": {
-			"padding": "0 var(--wp--custom--margin--horizontal)"
 		},
 		"typography": {
 			"lineHeight": "var(--wp--custom--body--typography--line-height)",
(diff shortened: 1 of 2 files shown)


stewart/theme.json
```diff
@@ -105,6 +105,9 @@
 			"margin": {
 				"horizontal": "30px",
 				"vertical": "30px"
+			},
+			"spacing": {
+				"outer": "min(1em, 30px)"
```
scruffian: TIL

scruffian: Does this actually do anything?

kjellr: Ha, whoops. No, I meant to put a `vw` unit there instead of the `1em`. This should scale more nicely now.

scruffian: I still don't know what `spacing > outer` does though!

kjellr: Oh! That's the variable used in all the new layout CSS here: 

https://github.com/Automattic/themes/blob/fafc2c5119fabffa2cddf78ce7112f65a1e884c9/stewart/style.css#L141-L142

It defines the outer left and right padding for the site. It's the same variable name [used in Twenty Twenty-Two](https://github.com/WordPress/twentytwentytwo/blob/0411debe4f66661d878c171d98614d8cfd0d4a5e/theme.json#L144) for easy copy/paste if this code needs to update.

scruffian: Ah nice one.

scruffian (approved):
LGTM

PR #5319, opened 2022-01-13
[Stewart] Rename template parts

kjellr:
By default, the header template part in Stewart was called "Header Spacer". I think that's going to be confusing for users, so this PR renames it to just "Header". 

We had a more traditional header template part that already used that name, so I've renamed that one to "Header with Site Title and Navigation". 

Before|After
---|---
Screen Shot 2022-01-13 at 3 46 07 PM|Screen Shot 2022-01-13 at 3 45 15 PM


scruffian (approved):
LGTM

PR #5320, opened 2022-01-13
Barnett: Initial commit.

alaczek:
This is the initial commit of Barnett - a super simple Blockbase child theme featuring link in bio style buttons on the homepage.

This is an updated version of Miller, which will be retired when Barnett launches. More info in pNEWy-exS-p2


scruffian (approved):
LGTM

PR #5325, opened 2022-01-14
Update variations

scruffian:
Somehow the variations were out of sync. This brings them back into sync.

To test run this and ensure you see no changes:

- cd variations
- node build-variations.mjs


pbking:
Perhaps "building variations" should be a part of the deploy action?
Any time there are any changes to the 'source' theme the variations would be "out of sync".

pbking:
I re-ran this against trunk and noticed a failing of our logic.

Each time this is ran the style.css file is replaced with whatever is in the source folder.  This includes the version, which is handled and kept updated by our deploy script.  Therefore this downgrades the variations to the incorrect version.

I'm going to manually correct this for this branch but we'll need something in our automation pipeline to handle that situation.

pbking (approved):
With the versions reverted this LGTM.

PR #5326, opened 2022-01-14
Adjust the sync logic to be simpler

pbking:
It was clear when trying to deploy the change to move templates to another folder that the "syncing" process of our deploy script wasn't working as intended.  "Moved" files weren't being tracked as "deleted".

This change relies 100% on rsync to bring the two locations into alignment.  (Thanks to @jffng for questioning why were weren't doing that already... )

First it determines which themes have changed (using existing logic) and rsyncs that theme's folder, INCLUDING removing deleted (or moved) files and empty folders from the destination (sandbox).

This can be tested in isolation by running `npm run deploy:push:changes` and also runs during the deploy action with `run deploy:svn`

### DIFF
--- a/.sandbox-ignore
+++ b/.sandbox-ignore
@@ -20,3 +20,5 @@ deploy-dotorg.sh
 **/*.zip
 variations
 variations/**
+inc/headstart
+languages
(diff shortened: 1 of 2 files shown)


jffng (approved):
This is working really well for me, and it greatly simplifies the utility code for this function.

I also tested renaming a theme directory + committing it (e.g. `block-template-parts` to `template-parts`, and the result from pushing the change was as expected.

PR #5331, opened 2022-01-14
Remove header-footer-only template from Blockbase children

scruffian:
Now that Blockbase children all use a "header" template, there's no need for a custom header-footer-only.html template since the one inherited from Blockbase is the same. We can remove these...

To test, create a page with the header-footer-only template and check that the header and footer look the same as on the index templates.


pbking:
videomaker[-white] has a special design for "header-footer-only".  Removing that custom template changes the design.

pbking (approved):
GTG for Arbutus and Russell

PR #5332, opened 2022-01-14
Blockbase: Remove custom 404 templates

scruffian:
Now that all Blockbase children use the same header template name, we can remove custom 404 templates as they all look the same.

To test, check that the headers and footers in Videomaker and Skatepark look the same on the 404 page as on other pages.


pbking:
Because Videomaker's heading spacing likely can't change I reverted those.  Skatepark's is fine to remove though.

pbking (approved):
🚢

PR #5333, opened 2022-01-14
Blockbase: Remove blank template

scruffian:
Blockbase children inherit the blank template from Blockbase, so they don't need to implement their own. To test check that you still see a blank template option in Russell.


pbking (approved):
Works as described

PR #5334, opened 2022-01-14
Blockbase children: Remove search templates

scruffian:
Now that Blockbase children all have a header template called `header` we don't need custom search templates in each theme. This removes all custom search templates so they will inherit from Blockbase.

To test, do a search in each affected theme and check that the template looks the same as before.

In the case of Quadrat and its variants, there will be a difference - page numbers won't show in search. Perhaps we should add them in to Blockbase too? Either way I think its worth the change to keep our code simpler.


pbking (approved):
Reverted the changes to videomaker templates.  Good to go.

PR #5335, opened 2022-01-14
Update videomaker header code

scruffian:
This updates the videomaker header for all the templates so that it implements the space inside the template with the pattern.

To test, check that all pages still have a header template.

### DIFF
--- a/videomaker/block-template-parts/header.html
+++ b/videomaker/block-template-parts/header.html
@@ -1,1 +1,5 @@
 <!-- wp:pattern {"slug":"blockbase/header-minimal"} /-->
+
+<!-- wp:spacer {"height":110} -->
+<div style="height:110px" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- /wp:spacer -->
--- a/videomaker/block-template-parts/header.html
+++ b/videomaker/block-template-parts/header.html
@@ -1,5 +1,5 @@
 <!-- wp:pattern {"slug":"blockbase/header-minimal"} /-->
 
-<!-- wp:spacer {"height":110} -->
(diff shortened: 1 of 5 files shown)


mikachan (commented):
This looks good to me. I tested each page and they all still correctly display the header.

One thing I did notice is that the spacer block currently has a top margin from the `.wp-site-blocks > * + *` rule, but this doesn't apply when it's in the header. So this PR reduces the vertical spacing by 20px. I'm not sure how much this matters! If it matters we could just add it to the size of the spacer.

pbking:
Makes sense to add it on to the spacer; good call.  Added that change.  Seems good to 🚢  to me.

pbking (dismissed):

pbking:
The design of videomaker has a special "header-footer-only" design that does NOT have that gap.  That situation should be considered differently so that we can use a standard gap and still maintain that gapless situation.

At first I thought we might use the same pattern block as the header.html but then changes by the user to the header template part wouldn't happen there. So I don't think that's an option.

Unfortunately I don't believe we can make this change for Videomaker.  The only way I can think of to achieve the design is to have that spacing element in all templates EXCEPT for "header-footer-only".

scruffian:
I'm not 100% sure that the lack of gap in the header-footer-only template is deliberate, and if it is, shouldn't we have that in all themese?

cc @kjellr

pbking:
I understand it's deliberate for Videomaker, not expected for all themes.

kjellr:
Yeah, the lack of that space is deliberate for Videomaker. But all other Blockbase themes (so far) should have it.

PR #5350, opened 2022-01-17
Fix/deploy versioning

pbking:
Due to a bug in the deploy process version bumping was lagging behind where it should have been; "has it been version bumped" was looking at the incorrect hash and determining that a theme HAD been version bumped, while it actually hadn't and should have been.

This change changes the order of "committing the change" (thus creating the hash) and updating the hash in the sandbox as well as adds a step to allow the user to verify (and change if necessary) the themes being automatically version bumped.

Additionally *.map files are now ignored when pushing to the sandbox.

### DIFF
--- a/.sandbox-ignore
+++ b/.sandbox-ignore
@@ -22,3 +22,4 @@ variations
 variations/**
 inc/headstart
 languages
+*.map
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -109,11 +109,24 @@ async function pushButtonDeploy(repoType) {
 
 		let thingsWentBump = await versionBumpThemes();
 
+		if( thingsWentBump ){
+			prompt = await inquirer.prompt([{
+				type: 'confirm',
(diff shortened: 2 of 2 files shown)


scruffian (approved):
LGTM

jffng (commented):
The change makes sense to me — we should commit (update the hash) _after_ the version bump changes have been committed.

The one thing I wasn't sure about is ignoring the .map files. Do we want those on dotcom to make it easier to track / debug CSS there?

pbking:
>The one thing I wasn't sure about is ignoring the .map files. Do we want those on dotcom to make it easier to track / debug CSS there?

We made an effort to NOT ship .map files in the .org builds, and that's what prompted me to remove 'em.  I don't know/expect my local map files to always be available and in-sync.  They are ignored by our .gitignore so there's no guarantee that the .map files I have on my machine would match the map files that would match the .css that is there.  

I see how they could be useful on our sandboxes, but I would prefer it not be an artifact in the deploy process.

PR #5357, opened 2022-01-18
Update screenshots from wpcom

scruffian:
This updates our screenshots with the latest ones from dotcom, which does its own image optimization


pbking:
Thanks for this.

PR #5361, opened 2022-01-18
Automate/dotorgdeploy

pbking:
Adding an action to deploy themes to dotorg automatically.

Any change pushed to /trunk will trigger this new action.  Then for every theme the version represented in style.css is compared to the version folders in SVN.  If that version doesn't exist it is created and committed to SVN.

Thus changing and committing an update to the version in style.css of any theme (in the list contained in the script) will trigger the release of a new version to dotorg.  This works with our dotcom deploy process as the updating of versions is the last action committed in the process.

The first successful run of this action is [here](https://github.com/Automattic/themes/actions/runs/1718715762).


mikachan (approved):
Thanks for doing this! It's looking great to me.

I haven't tested this, but I can see the successful run you've done, and the .yml file makes sense to me 👍

PR #5363, opened 2022-01-18
Blockbase: Load Block Patterns for children automatically

scruffian:
Rather than having custom code to load block patterns, we should just load them if they are there. This adds code to Blockbase to check for patterns in the child and load them.

To test check that you can still see the patterns in these themes:

Videomaker (and variants)
Quadrat (and variants)
Geologist (and variants)
Skatepark
Seedlet Blocks
Zoologist

### DIFF
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -183,3 +183,8 @@ require get_template_directory() . '/inc/block-styles.php';
  * Block Patterns.
  */
 require get_template_directory() . '/inc/block-patterns.php';
+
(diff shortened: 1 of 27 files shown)


mikachan (approved):
This is looking good! I've tested all the themes you mentioned and I could see all the patterns loading fine, both the theme-specific ones and the ones included with Blockbase.

jeffikus:
I like this :) decisions over options. Nice work getting rid of code in the child themes! Pinging @Automattic/flow-patrol-manage to update the other child themes.

PR #5366, opened 2022-01-19
Blockbase: Add a footer nav location

scruffian:
The `footer-vertical-menu` pattern references a `footer` nav location. This is only registered in Videomaker, not other Blockbase themes, so if other themes use that pattern it won't work.

In most cases people will probably only use this pattern if its already included in the theme (like it is with Videomaker) so this is probably not generally an issue. If people are add the pattern using the Site Editor, then they probably don't need to use the location anyway as they can just edit the nav directly.

The code here moves the registration of the footer nav location to Blockbase, so that its available to all children. This does have the side effect that the location will appear for all themes, even those that don't implement the pattern.

Another option is to remove this footer pattern and just implement a custom footer for videomaker


MaggieCabrera:
I think it's ok to have this on Blockbase. Blockbase has a pattern that implements a footer nav, it makes sense that a footer nav location is defined. Of the top of my mind I think Mayland uses a menu on its footer. Initially it uses the same menu as the header because that's what classic mayland does, but it could maybe use the footer nav as well.

pbking:
(edit)
> Blockbase has a pattern that implements a footer nav,

I didn't realize this until now, but if Blockbase is in the "navs in footers" business it makes sense to do it all the same way.  However, all of those target "primary" rather than a separate footer menu location.  

Is changing that to be the "footer" we define what we want?  If we make this change then I think we should standardize on that and make all of those footers use the footer menu target.

Otherwise this feels like a special situation where we really only want a 3rd option for Videomaker and that should be a custom footer implementation.

scruffian:
Closing in favour of https://github.com/Automattic/themes/pull/5374

PR #5368, opened 2022-01-19
Remove block style registration

scruffian:
We shouldn't be implementing custom block styles in themes as users will lose them when they switch themes. This removes the custom block styles but doesn't remove the CSS, so existing implementations, and the patterns, will still work.

To test, insert patterns that use these styles (e.g. Quadrat > Query Diamond) and check that they still show correctly.


blockbase/inc/block-styles.php
```diff
@@ -1,64 +0,0 @@
...
-
-	function blockbase_register_block_styles() {
-
-		if ( function_exists( 'register_block_style' ) ) {
-
-			/**
-			 * Register block styles
-			 */
-
-			//Post meta styles with icons
```
jffng, line 19: Just noting that adding the ability to toggle icons to post meta blocks is tracked in GB already:

https://github.com/WordPress/gutenberg/issues/30348
https://github.com/WordPress/gutenberg/issues/24957
https://github.com/WordPress/gutenberg/issues/24956
https://github.com/WordPress/gutenberg/issues/24952

seedlet-blocks/inc/block-styles.php
```diff
@@ -1,38 +0,0 @@
...
-		if ( function_exists( 'register_block_style' ) ) {
-
-			/**
-			 * Register block styles
-			 */
-			register_block_style(
-				'core/latest-posts',
-				array(
-					'name'         => 'seedlet-alternating-grid',
-					'label'        => __( 'Alternating Grid', 'seedlet-blocks' ),
```
jffng: Just confirming we want to remove these, based on @kjellr's reply in slack: p1642598004001100/1642590521.493900-slack-C029FM1EH

scruffian:
I've reverted the change to Seedlet Blocks as we should wait for that to be in Gutenberg before we remove it. I think this is ready for another review.

seedlet-blocks/inc/block-styles.php
pbking: Resolved.  Seedlet's stays for now.

pbking (approved):
This LGTM

🚢

PR #5370, opened 2022-01-19
Videomaker: Add pattern sources to their respective files

scruffian:
This adds the sources for each pattern to the top of each file. This has the benefit of making it clear which pattern uses which sources. It will also help us to generate the readme.txt file containing the sources for each pattern used in the theme.

There are some sources referenced which aren't used in the patterns anymore, so these have been removed.

### DIFF
--- a/videomaker/inc/patterns/about.php
+++ b/videomaker/inc/patterns/about.php
@@ -2,6 +2,11 @@
 /**
  * About.
  *
+ * Sources:
+ * https://stocksnap.io/photo/people-man-6O7JXC5DC5
+ * https://stocksnap.io/photo/people-man-NSV3ZCAJXT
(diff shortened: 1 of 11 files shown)


kjellr:
Interesting approach. I kind of like that the pattern source is more closely related to the file, but I worry that this overcomplicates things in favor of build tooling? I'm not sure this is simpler than just placing all the credit lines in one place like we're already doing.

pbking:
I agree that it might be overcomplication.  I'm not sure how often we evaluate and update the sources, but I'm doubtful that the location in which we store it will change the timing of that action.  If/when we were augmenting an already otherwise automated updating of readme.txt I would say go for it.  But until then there's two locations to update and that's twice as much friction.

PR #5367, opened 2022-01-19
Update Twenty Twenty-Two Variations

mikachan:
#### Changes proposed in this Pull Request:
This updates the TT2 variations in line with any recent changes from the [TT2 repo](https://github.com/WordPress/twentytwentytwo). It mainly involved updating the header-large-dark and header-small-dark template parts.

The Swiss variation was missing these parts entirely, so I've copied them into this variation as well so it's consistent with the others.

PR #5371, opened 2022-01-19
Update colors for header parts in Twenty Twenty-Two variations

mikachan:
I pushed [a change](https://github.com/Automattic/themes/pull/5367) earlier that updated the TT2 color variations to the latest versions of the header parts. In the process, I managed to mess up the color combinations for these parts. This PR fixes that - tested locally and on sandbox!

To test, make sure all the color variation colors are correct for the two header parts (the pink, red, and mint background colors are the easiest to spot).


scruffian (approved):
LGTM

PR #5374, opened 2022-01-19
Videomaker: Consolidate footer patterns in Videomaker

scruffian:
This removes a pattern from Blockbase and updates the corresponding pattern in Videomaker, so that themes aren't offered a pattern which won't work with their nav menu locations.

To test, check that the Videomaker footer continues to load as before, using the menu items in the footer nav location.


pbking (approved):
LGTM

PR #5376, opened 2022-01-19
Judith: initial commit

onuro:
This is the initial commit of Judith - a super simple Blockbase child theme featuring link in bio style buttons on the homepage.

This is an updated version of Baker, which will be retired when Judith launches. More info in pNEWy-exS-p2


scruffian (approved):
LGTM. I think Screenshot.png should be named screenshot.png

MaggieCabrera:
Are we keeping this one?

PR #5381, opened 2022-01-20
[Stewart] Readme Cleanup

kjellr:
Just some minor cleanup for the Stewart readme in preparation for submission to WP.org.

### DIFF
--- a/stewart/readme.txt
+++ b/stewart/readme.txt
@@ -1,7 +1,7 @@
 === Stewart ===
 Contributors: automattic
 Requires at least: 5.8
-Tested up to: 5.9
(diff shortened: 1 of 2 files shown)


stewart/readme.txt
```diff
@@ -1,7 +1,7 @@
 === Stewart ===
 Contributors: automattic
 Requires at least: 5.8
-Tested up to: 5.9
+Tested up to: 5.8
```
kjellr, line 4: This can't be 5.9 if we're submitting the theme before 5.9 is actually released. So I'm putting 5.8 for now.

pbking (approved):
👍

PR #5383, opened 2022-01-20
[Stewart] Fix alternate header template markup

kjellr:
Fixes some weird typos.


scruffian (approved):
Looks like a search and replace got greedy

PR #5386, opened 2022-01-21
Barnett: Update image credits.

alaczek:


barnett/style.css
```diff
@@ -29,8 +29,7 @@ GNU General Public License for more details.
 
 Barnett uses the following third-party resources:
 
-Green tropical leaves background. Original public domain image from Wikimedia Commons.
-Source: https://www.rawpixel.com/image/3286299/free-photo-image-flower-tropical-green
+Space Travel Poster (2015). Original from Official SpaceX Photos. Digitally enhanced by rawpixel.
```
scruffian: We don't need this in the style.css, only in the readme, so I suggest we remove this.

scruffian (approved):
LGTM

barnett/style.css
alaczek: Hmm... I was following the example of Blockbase there. I'm happy to remove credits from style.css though.

PR #5389, opened 2022-01-21
Blockbase: Fix social navigation warning

scruffian:
If the social navigation is empty this throws a warning. To test try setting no menu at the "Social" location


pbking:
I've not been able to reproduce this to validate that the change fixes it.  Everything continues to operate as expected with the change, I just can't confirm what it's fixing.

Issue #5211, opened 2021-12-16
Handle `Template:` discrepancy for premium themes


pbking:
Any theme in this repository that will be deployed as a 'premium' theme needs to have an absolute reference to their parent theme.  However while in THIS repository they have to be references relative.

For example Videomaker in this repo should be:
`Template: blockbase`
while in the /premium repo should be:
`Template: pub/blockbase`

How should this be handled?  Can/should it be automated?  Should that just be a manual thing to check every time those themes are deployed?

pbking:
Videomake is the only theme in this situation.  It may some day be transitioned out but for now this larger issue can be closed.

PR #5196, opened 2021-12-14
Transition Videomaker from /pub to /premium

pbking:
This change adds a command to push "premium themes" to the sandbox.

`npm run deploy:push:premium`

This will rsync the entire contents of the themes defined in the script to the premium location of the sandbox.  No other deployment actions are provided by the script.

This change begins the deployment of Videomaker to premium themes and should be accompanied by:

This change which removes Videomaker from /pub
D71549-code

This change which adds Videomaker and Videomaker-white to /premium
D71550-code

PR #5396, opened 2022-01-25
Premium: Update blockbase references when we push to the premium repo

scruffian:
When we push Videomaker and Videomaker White to the premium repo we need to update the `blockbase` slug with `blockbase-premium`. This script takes care of that change and then reverts the slug once the sync is complete.

#### Related issue(s):
https://github.com/Automattic/themes/pull/5196
https://github.com/Automattic/themes/issues/5211

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,8 @@
 		"sandbox:watch": "chokidar '**/*' -i '*/node_modules' -i '.git' -c './sandbox.sh push --ignore' --initial",
 		"sandbox:clean:git": "node ./theme-utils.mjs clean-sandbox-git",
 		"sandbox:clean:svn": "node ./theme-utils.mjs clean-sandbox-svn",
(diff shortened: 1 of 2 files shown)


pbking:
I've updated this change to accomplish the following:

The files reverted after the change are limited to only those files targeted to change (no more blanket `git reset` which might change things unintended to change).

The rsync command also deletes files that are absent from the source (local).

/sass folder is excluded from being pushed to the sandbox

added a new command to CLEAN the premium sandbox: `npm run sandbox:clean-premium:svn`

refactored the "premium themes" folder to replace the "directoresToIgnore" value so that there is only one collection (non-theme folders are now ignored for syncing purposes now, so there's no need for "variations" to be included in that list).

scruffian (commented):
This looks great, thanks for the improvements!

PR #5406, opened 2022-01-27
Fix/variation versioning

### LINKED ISSUE #5401: Deploy: Bump version numbers for variations
When running the variations script, the old version numbers are used. We should find a way to use the latest version number for each variation.

pbking:
This change does a few things:

First the 'template-mods' feature of the variations script has been removed.  It isn't used and its presence is confusing considering there's no examples.  The need was filled by using patterns instead of template parts so this has been removed.

Modifies the build variation script to use the version in the DESTINATION directory (if there is one) when the variation is built.  This addresses #5401.  To test run the script `npm run build:variations` and note that the versions in the variation style.css files is unchanged.

Additionally the variation script can now STAGE changes in git when ran.  Run it with `node ./variations/build-variations.mjs git-add-changes`

Lastly the deploy script (`theme-utils.mjs`) now leverages the build-variations.mjs script (with `git-add-changes` option) during the deployment process.  This now happens just before version-bumping themes (so that any variation themes that were changed are included in the version bump).

### DIFF
--- a/variations/build-variations.mjs
+++ b/variations/build-variations.mjs
@@ -6,9 +6,9 @@ import { fileURLToPath } from 'url';
 import { executeCommand, getThemeMetadata } from '../theme-utils.mjs';
 
 const localpath = dirname( fileURLToPath( import.meta.url ) );
+const args = process.argv.slice(2);
 
 async function start() {
-	let args = process.argv.slice(2);
 	let source = args?.[0];
 	let variation = args?.[1];
 	if ( source && variation ) {
@@ -75,6 +75,10 @@ async function buildVariation(source, variation) {
 		// replace the with current version
 		if ( currentVersion != null ) {
 			await executeCommand(`perl -pi -e 's/Version: (.*)$/"Version: '${currentVersion}'"/ge' ${destDir}/style.css`);
+		}
+
+		if ( args[0] == 'git-add-changes') {
+			await executeCommand(`git add ${destDir}`, true);
 		}
 	
 		console.log('Finished sucessfully.\n\n');
--- a/variations/build-variations.mjs
+++ b/variations/build-variations.mjs
@@ -7,7 +7,7 @@ import { executeCommand, getThemeMetadata } from '../theme-utils.mjs';
 
 const localpath = dirname( fileURLToPath( import.meta.url ) );
 
-(async function start() {
+async function start() {
 	let args = process.argv.slice(2);
 	let source = args?.[0];
 	let variation = args?.[1];
@@ -15,7 +15,9 @@ const localpath = dirname( fileURLToPath( import.meta.url ) );
 		return await buildVariation(source, variation);
 	}
 	return await buildAllVariations();
-})();
+}
+
+start();
 
 async function buildAllVariations(){
 	for (let source of getDirectories( localpath )){
@@ -34,7 +36,7 @@ async function buildVariation(source, variation) {
 
 	try {
 		// First grab the existing version if the variation exists already
-		let variationExists = true; //TODO
+		let variationExists = fs.existsSync(`${destDir}/style.css`);
 		let currentVersion = null;
 
 		if( variationExists ) {
@@ -51,7 +53,6 @@ async function buildVariation(source, variation) {
 		// remove unneeded resources
 		await fs.remove( destDir + '/sass');
 		await fs.remove( destDir + '/node_modules' );
-		await fs.remove( destDir + '/template-mods.json');
 		await fs.remove( destDir + '/package.json');
 		await fs.remove( destDir + '/package-lock.json');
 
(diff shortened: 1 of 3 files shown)


variations/build-variations.mjs
```diff
@@ -42,7 +53,6 @@ async function buildVariation(source, variation) {
 		// remove unneeded resources
 		await fs.remove( destDir + '/sass');
 		await fs.remove( destDir + '/node_modules' );
```
jffng, line 55: This is unrelated to the PR but it's the first time I'm trying out the `build:variations` script and I'm noticing the copy process feels a bit slow. I have a hunch the length of the copy process is impact by having to delete the node_modules folder for each variation, I wonder if there's an `fs` way of copying that can ignore certain files/folders to save some time.

jffng (commented):
I gave both commands a test and they seemed to check out according to your instructions — style.css was not bumped. 

The result I got from running `node ./variations/build-variations.mjs git-add-changes` was the following set of staged changes: 

Screen Shot 2022-01-27 at 4 12 50 PM

Is that expected?

pbking:
> The result I got from running node ./variations/build-variations.mjs git-add-changes was the following set of staged changes:

Yes, the changes you noted are expected.  Because the variation-readme.md file changed (removing reference to the unused template-mods feature) that file changed for all the variation themes.  

The videomaker-wite changes seem right as well, though @scruffian has been working on videomaker-white.  So it may only be right in that the script is working as expected, but perhaps not right in that maybe videomaker-white isn't a simple variation anymore.

scruffian:
It looks like vidoemaker white wasn't built the last time I made changes to videomaker, so those changes are expected

variations/build-variations.mjs
```diff
@@ -32,8 +35,16 @@ async function buildVariation(source, variation) {
...
 
 	try {
+		// First grab the existing version if the variation exists already
+		let variationExists = fs.existsSync(`${destDir}/style.css`);
+		let currentVersion = null;
 
-		// First empty the old directory.
+		if( variationExists ) {
+			let styleCss = fs.readFileSync(`${destDir}/style.css`, 'utf8');
+			currentVersion = getThemeMetadata(styleCss, 'Version');
```
jffng, line 44: Why do we pull the version number from the generated version instead of the source? 

For example, I'm looking at `variations/quadrat/quadrat-black/style.css` — the [version number in the variation source](https://github.com/Automattic/themes/blob/trunk/variations/quadrat/quadrat-black/style.css#L10) is lower than the [built version](https://github.com/Automattic/themes/blob/trunk/variations/quadrat/quadrat-black/style.css#L10). Is is that expected or a different issue?

jffng:
> Yes, the changes you noted are expected. Because the variation-readme.md file changed (removing reference to the unused template-mods feature) that file changed for all the variation themes.

> It looks like vidoemaker white wasn't built the last time I made changes to videomaker, so those changes are expected

👍 So those changes would be committed on the next deploy?

scruffian:
I don't think this is working as expected. Quadrat is on 1.1.35 but the variations are getting a version of 1.1.37. I think its copying the version from the variation not from the parent.

pbking:
> I don't think this is working as expected. Quadrat is on 1.1.35 but the variations are getting a version of 1.1.37. I think its copying the version from the variation not from the parent.

> Why do we pull the version number from the generated version instead of the source?

It may be that the variation changes and the parent does not.  In that case the variation will need to be version bumped while the parent stays the same.  

However any time the parent has changes that require a version bump those changes will get distributed to the variations also causing an independent version bump.  Thus any change to the SOURCE also BUMPS the version of the variation, however those versions might not always be in sync.

While the variations are spawns of another theme they are still a whole and complete theme in-and-of themselves just like the source theme.  Our version bumping logic doesn't treat the variations any differently.  Ensuring that those variations are kept strictly in sync with their source would probably require a refactoring of the version bumping logic.

It may be that only one variation changes (Geologist-yellow changes to use a different color of yellow or whatever) but the sibling variations do not.  Those versions are independent of both each other and the parent.

The version represented in the /varations location were just the version of the SOURCE when the variation was created.  IMO this version should have started at 1.0.0 rather than whatever version the source was when it was created.  But perhaps that doesn't best represent the variation.

The above is my assumption of what is appropriate.  I can certainly see argument in keeping the versions of all the source and variations in sync.  Changing this script to do that would be pretty easy if there's consensus that that is the preferred behavior.

scruffian (approved):
Ok yeah I get you. I don't have strong feelings either way. Let's bring this in...

variations/build-variations.mjs
pbking: Yeah, something like that could be done to speed it up.  Prefer to do that in a separate PR as that can be completely under the hood.

variations/build-variations.mjs
pbking: Discussed below.

PR #5408, opened 2022-01-28
Fix dotorg deploy for seedlet-blocks and livro

pbking:
Removed commas that were preventing seedlet-blocks and levro from being properly deployed to dotorg

They were causing the path (both file path and SVN path) from being correct resulting in un-deployed themes.

See [this deployment action result](https://github.com/Automattic/themes/runs/4960239459?check_suite_focus=true) for example of a failed run.


mikachan (approved):
This LGTM.

Nice error handling in the failed run too!

pbking:
Success: https://github.com/Automattic/themes/runs/5008925934?check_suite_focus=true

PR #5415, opened 2022-01-31
WIP Skatepark: theme variations

MaggieCabrera:

I've opened this PR to consider the possibility of leveraging https://github.com/WordPress/gutenberg/pull/35619 instead of the customizer (since Skatepark has not been yet launched) for the color customizations + duotone. In this PR:

- I added a filter so that the child themes can disable color customization. This could be something we can extend to other Blockbase functions to ake some child themes "less universal" while still benefiting from the parent's CSS, patterns and templates.
- I added variations for all the colors, those seem to be working mostly fine, but the duotone filter is not showing on the editor. This is a known issue reported in https://github.com/WordPress/gutenberg/issues/35331
- I added a variation that disables duotone for the blocks and that works as intended but since style variations don't have labels, it's not recognizable as to what the variation does (the icon is the same as the default theme). Variations will have labels in the future, but they don't have one now.

### DIFF
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -153,8 +153,14 @@ function blockbase_fonts_url() {
  * Customize Global Styles
  */
 if ( class_exists( 'WP_Theme_JSON_Resolver_Gutenberg' ) ) {
-	require get_template_directory() . '/inc/customizer/wp-customize-colors.php';
-	require get_template_directory() . '/inc/customizer/wp-customize-color-palettes.php';
+
+	$customize_colors = apply_filters( 'blockbase_customize_colors', true );
+
+	if ( $customize_colors ) {
+		require get_template_directory() . '/inc/customizer/wp-customize-colors.php';
+		require get_template_directory() . '/inc/customizer/wp-customize-color-palettes.php';
+	}
+
 	require get_template_directory() . '/inc/customizer/wp-customize-fonts.php';
 }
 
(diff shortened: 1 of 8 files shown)


MaggieCabrera:
I wonder if we could have a way to generate dotcom variations using this instead of having to build the whole theme from scratch. that means users won't lose their changes if they decide they want to try another color customization directly from the FSE

scruffian:
> I wonder if we could have a way to generate dotcom variations using this instead of having to build the whole theme from scratch. that means users won't lose their changes if they decide they want to try another color customization directly from the FSE

Yeah this is my biggest concern. We want the action of changing the colors to be non-destructive, but with this approach you would lose all your customizations.

pbking:
I'm not seeing a downside to this method.  Variations really seems like the way to go for Skatepark.

<tangent>
I think we could refactor the color palette selection in Blockbase to leverage the variation JSON (instead of custom variables in theme.json), duplicating what's going on in Global Styles even more closely.  That would eliminate the need to remove the color customizer for Skatepark (since the behavior would be the same).  
</tangent>

scruffian:
> I'm not seeing a downside to this method.

The downside I can see is that when you switch variations you lose all your other customisations.

MaggieCabrera:
Closing in favor of https://github.com/Automattic/themes/pull/5789

PR #5418, opened 2022-01-31
Blockbase search template update

### LINKED ISSUE #5403: Block-Based Themes: Search Results Archive Page is limited to 10 results, no pagination
### Quick summary

User reached out in 33876389-hc, pointing out that the search results page was limited to 10 results, even though their search term yielded 16 results in their Posts area. I was able to confirm this issue on my test site, using their theme, Geologist (Cream), but I also tested with several other block-based themes. 

### Steps to reproduce

1. Test on site that has more than 10 posts
2. Add search block to home page, if there isn't a search function
3. Search for something that will yield more than 10 results. The letter _e_ is a good bet.
4. Count the number of search entries, it will max out at 10 results.

If all the search results aren't shown on one page, then you would expect there to be pagination with additional search results on additional pages. 

### What actually happened

There are only 10 results, and there is no option to find more (either through loading more via a button or using pagination).

### Context

User reported in chat, and I was able to confirm with testing.

### Simple, Atomic or both?

Simple

### Theme-specific issue?

Geologist (Cream) - user theme
Calvin
Quadrat
Arbutus 
Zoologist

_No response_

### Reproducibility

Consistent

### Severity

Most (> 50%)

### Available workarounds?

_No response_

### Workaround details

With the launch of FSE, a user will be able to edit the Search page template, however it seems good to have a better solution baked into the themes themselves.

matiasbenedetto:
Blockbase search template update.

Changes:
- Adding search block
- Updating posts query template
- Adding missing query pagination block

#### Related issue(s):
https://github.com/Automattic/themes/issues/5403

### DIFF
--- a/blockbase/block-templates/search.html
+++ b/blockbase/block-templates/search.html
@@ -1,20 +1,47 @@
 <!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
-<!-- wp:group {"layout":{"inherit":true},"tagName":"main"} -->
-<main class="wp-block-group">
+<!-- wp:group {"layout":{"inherit":true}} -->
+<div class="wp-block-group">
+	<!-- wp:search /-->
 
-<!-- wp:query -->
-<div class="wp-block-query">
-<!-- wp:post-template -->
-
-	<!-- wp:post-title {"level":5,"isLink":true} /-->
-	<!-- wp:post-excerpt /-->
-
-<!-- /wp:post-template -->
+	<!-- wp:spacer {"height":40} -->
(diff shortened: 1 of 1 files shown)


blockbase/block-templates/search.html
```diff
@@ -1,20 +1,47 @@
 <!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
-<!-- wp:group {"layout":{"inherit":true},"tagName":"main"} -->
-<main class="wp-block-group">
```
MaggieCabrera, line 4: the main tag is there for a11y reasons

scruffian (approved):
LGTM

PR #5419, opened 2022-02-01
Blockbase: Remove the fallback for the navigation block

### LINKED ISSUE #5337: Arbutus menu locations not functioning as expected
### Quick summary

A user reported not being able to hide their primary menu which was being displayed in the footer (social menu location) of their simple site using Arbutus despite not being set to do so. 

When the no menu is set to display in the primary menu location, the menu still displays there as well. 

I could replicate on my own site. 

### Steps to reproduce

1. Set a primary menu and to display in the primary menu location. 
2. Toggle display of the menu off for that location.

The primary menu would only display in the primary location if set to do so.

The menu would not display at all if no location was set. 

### What actually happened

The menu set to display in the primary location also displays in the social menu location.

When no display location is set for a menu, the primary menu still appears in both the primary and social menu locations.

Markup on 2022-01-06 at 15:47:36
 

### Context

Customer report: 4654288-zd-woothemes (provided CSS to hide the menu for now)

Though the [live demo](PdvP9O-2-p2) does currently show the primary menu in the social menu location, the [WordPress.com/themes/arbutus](https://wordpress.com/theme/arbutus/support/testblogsajmes.wordpress.com) page has an image showing primary and social menus displayed only in their respective locations.

### Simple, Atomic or both?

Simple

### Theme-specific issue?

Arbutus

_No response_

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

_No response_

### Workaround details

_No response_

scruffian:
By default the navigation block outputs a list of pages when it doesn't have any other content. This is causing some confusion in https://github.com/Automattic/themes/issues/5337 and makes it impossible for universal themes to have an empty navigation block.

This PR removes that fallback so that if no menu is specified at a location then nothing displays. We might want to only do this for Arbutus.

To test:
- switch to arbutus
- set your primary menu location to not have any menu
- check that the block doesn't output anything
- set your primary menu location to a defined menu
- check that the block outputs the menu defined at the primary menu location

### DIFF
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -175,6 +175,14 @@ add_action(
 );
 
 /**
+ * Disable the fallback for the core/navigation block.
+ */
+function blockbase_core_navigation_render_fallback() {
+	return null;
+}
+add_filter( 'block_core_navigation_render_fallback', 'blockbase_core_navigation_render_fallback' );
+
+/**
  * Block Styles.
  */
 require get_template_directory() . '/inc/block-styles.php';


pbking (approved):
Works as expected.  I believe this should be the behavior of any Blockbase theme.

PR #5435, opened 2022-02-01
POC: remove customizer options from Blockbase children via filters

MaggieCabrera:

This PR removes the customizer options on Skatepark via filters on Blockbase. To test this check that everything is working as intended (fonts are showing correctly, colors can change via the FSE only) and there's no customizer option under the Appearance menu

context: pNEWy-eGS-p2#comment-54234

### DIFF
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -153,9 +153,17 @@ function blockbase_fonts_url() {
  * Customize Global Styles
  */
 if ( class_exists( 'WP_Theme_JSON_Resolver_Gutenberg' ) ) {
-	require get_template_directory() . '/inc/customizer/wp-customize-colors.php';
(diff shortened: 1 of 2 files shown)


mikachan (approved):
This looks good to me. 

In Skatepark, I can't see the Customizer under Appearance, and I can change font colors via the FSE. If I change to another Blockbase theme, I can still see the Customizer and change fonts and colors from there.

scruffian:
I think we should hold off on this until https://github.com/WordPress/gutenberg/pull/38442 merges

MaggieCabrera:
I agree, but I think we should add the filters to Blockbase for the future.

pbking:
Since Skatepark is no longer a Blockbase child Imma gonna close this one.

PR #5094, opened 2021-11-24
Add archeo (as Blockbase child)

scruffian:
#### Changes proposed in this Pull Request:
This is a skeleton version of Archeo:
localhost_4759_

PR #5436, opened 2022-02-01
Add Archeo (as standalone theme)

mikachan:
This is based on the original initial [Archeo PR](https://github.com/Automattic/themes/pull/5094), but I've made it a standalone theme rather than a Blockbase child.

I've used the `functions.php` file from Livro, which is working great so far but I don't know if anything needs to be changed here. I've also based the `style.css` file on the one from Stewart.

This should also cover the spacing and layout settings I think (#5421).


scruffian (approved):
This looks like a great start. I think we should probably just merge and iterate.

mikachan:
Thanks @scruffian! Merging and iterating sounds good to me.

PR #5439, opened 2022-02-01
Deployment: Removed the check for uncomitted changes.  

pbking:
During the build process determining which themes had changed included checking for uncomitted changes.  

Those shouldn't be considered for version bumping or any other deploy action.  

Additionally that was returning false-positives for the variations as illustrated in [this commit](https://github.com/Automattic/themes/commit/7013fc330a0565a33268bf18634c75ef87784b19)  which automatically version bumped all of the variations.


scruffian (approved):
LGTM

PR #5441, opened 2022-02-01
Blockbase - register standard header and footer in theme json.

Addison-Stavlo:

Currently, only the extra variations of headers are registered in the theme.json for blockbase.

This means that when loading blockbase for the first time, the header and footer in place do not have any area classification and are set to general. Child themes end up with the same issue, unless they explicitly register these in their theme.json as well (as is the case with quadrat, but not geologist).

#### Testing the changes.

* Load the site editor using the blockbase theme with no customized template parts.
* Open the template dropdown in the top bar, and notice that the variations of the template parts on the page are listed as 'General':

Screen Shot 2022-02-01 at 8 00 18 AM

* Apply this diff and reload the editor
* Verify that the header and footer classifications are as expected:

Screen Shot 2022-02-01 at 7 55 23 AM

NOTE - if testing on wpcom, an additional fix to the theme json resolver is required to see the expected end result (must patch D74195-code to your sandbox).  However on self hosted there are no additional steps.

#### Related issue(s):

### DIFF
--- a/blockbase/theme.json
+++ b/blockbase/theme.json
@@ -25,6 +25,14 @@
 		{
 			"name": "header-wide",
 			"area": "header"
+		},
+		{
+			"name": "header",
+			"area": "header"
+		},
+		{
+			"name": "footer",
+			"area": "footer"
 		}
 	],
 	"customTemplates": [


scruffian (approved):
LGTM. Thanks for fixing this

PR #5442, opened 2022-02-01
Videomaker: use locally-hosted version of Inter instead of Google Fonts

### LINKED ISSUE #5388: [Videomaker] Use locally-hosted version of Inter instead of Google Fonts
The Google Fonts version of Inter doesn't include italics: 

https://fonts.google.com/specimen/Inter

As a result, italic text will be faked by the browser by slanting the letters. It's not an elegant solution. Instead of sticking with this older version, we should switch to using the variable version from here, which we would bundle with the theme:

https://github.com/rsms/inter/

matiasbenedetto:
- Use locally-hosted variable version of Inter instead of Google Fonts
- Use sans-serif as a fallback for Inter 

#### Related issue(s):
https://github.com/Automattic/themes/issues/5388


scruffian (approved):
LGTM. The font URL was redirecting to the same URL but with a trailing slash, so I just added the trailing slash. Not sure what's going on there...

PR #5443, opened 2022-02-01
Include blockbase-premium in pushed themes

pbking:
Add 'blockbase-premium' to the collection of themes that are (modifed and then) pushed to /themes/premium

This does NOT (yet) change 'namespace' changes from 'blockbase' to 'blockbase-premium'

An example of the diff created by a run from this script can be found here: 
D74204-code

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -679,20 +679,34 @@ async function pushPremiumToSandbox() {
 		'block-template-parts/footer.html'
 	];
 
+	// Include blockbase in the themes we make changes to
+	const actionableThemes = premiumThemes.concat(['blockbase']);
+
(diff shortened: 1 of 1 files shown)


pbking:
So it seems there is already a utility to PULL the changes from Blockbase (the one in wpcom-git /themes/pub) and rebuild blockbase-premium.  (As opposed to this method to PUSH the changes from the /trunk of GitHub).

I have some concerns that this different behavior will result in a version of Blockbase that lags behind what deployed in /pub as it won't be a part of our regular deployments. 

However the needs of the /premium are not the same as /pub and there are other themes not represented in this repo that should be considered for that upgrade process.

scruffian:
While the premium themes live in a different repo I think it makes sense to push blockbase-premium in a different process.

PR #5446, opened 2022-02-02
remove the default color check during color customizations save

### LINKED ISSUE #5387: Blockbase + co: Global styles colors are lost after changing settings in the customizer
Related conversation: p1642735935059000-slack-C01DF571R5Y

Let's check that any color changes done via GS are not wiped out when changing any other settings in the customizer

pbking:
Fixes: #5387

Removes the "check_if_colors_are_default" check that determines if the color customizations can be removed.

That state can no longer be determined with the current data and was causing colors to be reset unnecessarily.

To test:

Start with an un-customized site with a Blockbase theme activated.  (I used Blockbase)
Using the FSE customize the Primary color to a different color.
Save those changes.
(Note that the change is not represented in the editor due to this reported issue, however the site reflects the colors as does the customizer).
Using the customizer change a setting other than the colors.  (I used the changing of fonts for testing mostly, but any change at all seems to cause the issue)
Note that the colors stay the same color they were configured to be in the FSE Global Styles panel.

### DIFF
--- a/blockbase/inc/customizer/wp-customize-colors.php
+++ b/blockbase/inc/customizer/wp-customize-colors.php
@@ -181,54 +181,51 @@ class GlobalStylesColorCustomizer {
 			unset( $user_theme_json_post_content->settings->color->palette );
 		}
 
-		//Set the color palette if it is !== the default
-		if ( ! $this->check_if_colors_are_default() ) {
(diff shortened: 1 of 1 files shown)


scruffian (approved):
Works perfectly. Great job!

PR #5449, opened 2022-02-02
Fixed import to include __DIR__ instead of assuming relative path

pbking:
Deploying the latest build didn't mast muster due to this error:

```
remote: ***********************************
remote: Syntax Error in: blockbase/inc/customizer/wp-customize-colors.php:
remote: Relative path includes are not allowed. Please fix this relative include before committing the file:
remote:
remote: require_once 'wp-customize-global-styles-setting.php';
remote:
remote: You can use the __DIR__ or ABSPATH constants for this. For example:
remote:
remote: require( __DIR__ . '/myfile.php' );
remote: or
remote: require( ABSPATH . 'wp-content/mu-plugins/myfile.php' );
remote: ***********************************
```

This change uses the __DIR__ option.

### DIFF
--- a/blockbase/inc/customizer/wp-customize-color-palettes.php
+++ b/blockbase/inc/customizer/wp-customize-color-palettes.php
@@ -1,6 +1,6 @@
 <?php
 
-require_once 'wp-customize-color-palette-control.php';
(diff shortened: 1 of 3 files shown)


scruffian:
Thanks

PR #5452, opened 2022-02-03
Archeo: fixed font files and styles

MaggieCabrera:
#### Changes proposed in this Pull Request:

I found that the font weights of the font we are using is not looking like the figma files. I see that we are only using one weight in the designs (400 on figma) and that doesn't look like the weight of the font in the browser:

Figma | Google fonts
--- | ---
Screenshot 2022-02-03 at 12 07 49Screenshot 2022-02-03 at 12 07 38 | Screenshot 2022-02-03 at 12 08 03

To follow the design as closely as possible I made the following decisions:

- The body font-weight is 400
- The headings font-weight is 300 (not 400 like figma says)
- I added the bold font file and italics for all 3 weights (300,400,700) so when users change the blocks weights and style it works as intended and it's not the browser forcing the font to do ugly things
- I preloaded the 400 and 300 normal fonts.

This is the content I used for testing:
```

<!-- wp:paragraph -->
<p>This text is the default weight and style</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p><strong>This text has been bolded</strong></p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p><em>This text is in italics</em></p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p><strong><em>This text is bold and italic</em></strong></p>
<!-- /wp:paragraph -->

<!-- wp:heading {"level":1} -->
<h1 id="heading-2">HEADING 1</h1>
<!-- /wp:heading -->

<!-- wp:heading -->
<h2 id="heading-2">HEADING 2</h2>
<!-- /wp:heading -->

<!-- wp:heading {"level":3} -->
<h3 id="heading-3">HEADING 3</h3>
<!-- /wp:heading -->

<!-- wp:heading {"level":4} -->
<h4 id="heading-4">HEADING 4</h4>
<!-- /wp:heading -->

<!-- wp:heading {"level":5} -->
<h5 id="heading-4">HEADING 5</h5>
<!-- /wp:heading -->

<!-- wp:heading {"level":6} -->
<h6 id="heading-4">HEADING 6</h6>
<!-- /wp:heading -->

<!-- wp:paragraph {"fontSize":"small"} -->
<p class="has-small-font-size">Paragraph Small. WordPress is software designed for everyone, emphasizing accessibility, performance, security, and ease of use. We believe great software should work with minimum set up, so you can focus on sharing your story, product, or services freely.</p>
<!-- /wp:paragraph -->

[...]

PR #5453, opened 2022-02-03
Archeo: Simplify heading declarations

scruffian:
This moves a lot of the declarations for heading blocks to the core/heading section of global styles, since this also controls h1-h6 elements. This helps keep the code simpler. You should notice no difference in the frontend.

### DIFF
--- a/archeo/theme.json
+++ b/archeo/theme.json
@@ -43,12 +43,6 @@
 					"lineHeight": 1.5
 				}
 			},
-			"heading": {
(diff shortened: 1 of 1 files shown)


mikachan (approved):
This is much easier to read, thanks!

The only thing I'd note is that I think the font-weight for headings will probably change to 300 with [this PR](https://github.com/Automattic/themes/pull/5452), but I think that can be handled there instead of here.

PR #5459, opened 2022-02-04
Shoreditch: Prevent blocks inside the layout grid block going full width

### LINKED ISSUE #5384: Shoreditch: An 'About' Block Pattern using layout grid appears incorrectly
### Quick summary

The Block Pattern shown in the screenshot displays incorrectly when used with Shoreditch (and possibly other themes).

The content overlaps horizontally.

Screenshot on 2022-01-07 at 10-54-17
 

### Steps to reproduce

1. Use the Block Picker to add this specific block pattern to a page or post on a site using Shoreditch.

I expected the Block Pattern to display the content with proper horizontal spacing, respecting the Layout Grid settings. 

### What actually happened

The image and text overlap horizontally. The image also appears to display larger than intended. 

### Context

Customer report here. We resolved by switching to Mayfield. 30281278-hc

I tested on my own site and this occurs with Shoreditch active. I added a similar Layout Grid and tried both with and without settings that mimicked the Block Pattern, but my manually added Layout Grid works as expected. 

Post: pcXb56-AL-p2

### Simple, Atomic or both?

Simple

### Theme-specific issue?

Shoreditch.

_No response_

### Reproducibility

Consistent

### Severity

_No response_

### Available workarounds?

_No response_

### Workaround details

_No response_

scruffian:
The code to achieve full width has some strange side effects when used inside the jetpack layout grid block in Shoreditch. Since this is alignment code is an idiosyncrasy of the theme I think it's best to just add exceptions for this case.

To test
- Switch to Shoreditch
- Add the about us pattern with a yellow photo
- Check that the text isn't hidden by the photo

#### Related issue(s):
Fixes #5384

### DIFF
--- a/shoreditch/css/blocks.css
+++ b/shoreditch/css/blocks.css
@@ -30,7 +30,7 @@ Description: Used to style Gutenberg Blocks.
 	margin-right: 0;
 }
 
-.can-align-wide .alignfull,
+.can-align-wide :not(.wp-block-jetpack-layout-grid-column) > .alignfull,
 .page-template-full-width-page .alignfull,
 .page-template-panel-page .alignfull {
 	margin-left: calc(50% - 50vw);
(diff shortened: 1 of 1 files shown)


jffng (approved):
This fixes the overlapping text issue. 

Before | After
---- | ----
Screen Shot 2022-02-04 at 8 58 22 AM | Screen Shot 2022-02-04 at 8 58 01 AM

The layout grid is still not matching the editor and front-end, but I don't think we should address that here / now.

PR #5462, opened 2022-02-04
Archeo: Add navigation styles

### LINKED ISSUE #5431: Archeo: Navigation (if needed)

scruffian:
This updates the navigation block for Archeo:

Screenshot 2022-02-04 at 12 25 41
Screenshot 2022-02-04 at 12 25 52

@kjellr @beafialho I didn't see any mockups for this, so I just had a go at something myself. It would be nice if the text could left align but that's a limitation of the nav block at the moment. cc @getdave.

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/5431

### DIFF
--- a/archeo/parts/header.html
+++ b/archeo/parts/header.html
@@ -1,7 +1,7 @@
 <!-- wp:group {"layout":{"inherit":"true"}} -->
 <div class="wp-block-group">
-    <!-- wp:group {"align":"full","layout":{"type":"flex","justifyContent":"space-between"},"style":{"spacing":{"padding":{"bottom":"var(--wp--custom--spacing--medium)","top":"var(--wp--custom--spacing--medium)"}}}} -->
-    <div class="wp-block-group alignfull" style="padding-bottom:var(--wp--custom--spacing--medium);padding-top:var(--wp--custom--spacing--medium)">
+    <!-- wp:group {"align":"full","layout":{"type":"flex","justifyContent":"space-between"},"style":{"spacing":{"padding":{"bottom":"var(--wp--custom--spacing--medium)","top":"var(--wp--custom--spacing--outer)"}}}} -->
+    <div class="wp-block-group alignfull" style="padding-bottom:var(--wp--custom--spacing--medium);padding-top:var(--wp--custom--spacing--outer)">
         <!-- wp:site-title /-->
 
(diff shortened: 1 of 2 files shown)


getdave:
> the text could left align but that's a limitation of the nav block at the moment. cc @getdave.

Can you not use alignment on the Nav block to move the text over to the left?

https://user-images.githubusercontent.com/444434/152534198-1c8bbaf6-f4e1-4af7-b6af-bf4ec8134c12.mov

Or is it that you want different styles applied at different breakpoints?

scruffian:
We want the menu right aligned but the overlay left aligned!

kjellr:
Because we're using `space-between`, I believe it's actually fine to use left justification inside of the Nav block everywhere. The `space-between` rule will still push things over to the edge:

Screen Shot 2022-02-04 at 8 27 12 AM

nav


I pushed that change here, plus a couple other cleanup items:

- I removed blocks that don't appear in the comps. 
- I fixed the site top padding to match the sides, like in the comps. 
- I added the CSS snippet from Twenty Twenty-Two that helps the menu line up correctly with the close button by default. 

I'd like to see the mobile nav cleaned up a bit too if we can manage it: 

- Font size: `var(--wp--preset--font-size--large)`
- Font weight: 300
- Gap: ~20px

Current|Target
---|---
Screen Shot 2022-02-04 at 8 33 21 AM|Screen Shot 2022-02-04 at 8 33 12 AM

scruffian:
One thing to note, when we align the nav to the left like that, it does this when it wraps:
Screenshot 2022-02-04 at 13 41 45

This is probably nicer than the alternative though!

Do you think the items need less vertical space when they wrap?

kjellr:
That's totally fine for wrapping (I think i prefer it actually. 😄).

And yes, I think we should decrease the vertical gap there too.

scruffian:
How's this? I had to add custom CSS :(
Screenshot 2022-02-04 at 14 03 43

scruffian:
> And yes, I think we should decrease the vertical gap there too.

This is really hard to do because it's using that block gap rule. We can remove the blockgap, but then we'd need create space between the items in a different which I think also require custom CSS at the moment. Any ideas @getdave?

kjellr:
> How's this? I had to add custom CSS :(

That looks good for now. 👍

pbking (approved):

PR #5463, opened 2022-02-04
Speed up build variations script

jffng:
This PR makes a small change to the build variations script to try and speed it up.

To test, run `node ./variations/build-variations.mjs` on trunk and this PR. Verify the result is the same, but faster on this PR.


mikachan (approved):
Works great for me 👍  I can see the script runs faster on this branch, and I get the same results.

Issue #5429, opened 2022-02-01
Archeo: Block patterns


mikachan:
- [x] 1. Headline over dark image: https://github.com/Automattic/themes/pull/5464

![image](https://user-images.githubusercontent.com/1645628/152005776-f3af2644-4ff4-42cc-af9f-d1e05d24ca1b.png)

- [x] 2. Image with headline and description: https://github.com/Automattic/themes/pull/5467

![image](https://user-images.githubusercontent.com/1645628/152006096-347a0016-bb0c-4065-8e36-8b3446105535.png)

- [x] 3. Image with headline and separate description: https://github.com/Automattic/themes/pull/5470

![image](https://user-images.githubusercontent.com/1645628/152006275-777c8a4d-0682-42a2-9b9d-787af9797116.png)

- [x] 4. Image with description: https://github.com/Automattic/themes/pull/5479

![image](https://user-images.githubusercontent.com/1645628/152006467-c5e230fa-2325-48ea-b977-05512c708a4e.png)

- [x] 5. Simple list of posts: https://github.com/Automattic/themes/pull/5480

![image](https://user-images.githubusercontent.com/1645628/152006610-248cd256-2a9e-4bc2-b55b-521ff796f785.png)

- [x] 6. Image with description and right-aligned headline: https://github.com/Automattic/themes/pull/5487

![image](https://user-images.githubusercontent.com/1645628/152006669-3e78f5e2-c120-46eb-ba55-24cf764259bd.png)

- [x] 7. Image with headline on dark background: https://github.com/Automattic/themes/pull/5474

![image](https://user-images.githubusercontent.com/1645628/152006842-62513d11-5379-4ea1-8c51-b3b26198f0bc.png)

- [x] 8. Layout with two images and text: https://github.com/Automattic/themes/pull/5486

![image](https://user-images.githubusercontent.com/1645628/152007018-47cb581e-7903-4980-b51d-e31852374f7f.png)

- [x] 9. Layered images with headline: https://github.com/Automattic/themes/pull/5468

![image](https://user-images.githubusercontent.com/1645628/152007247-4ee6f386-6607-47d8-b59c-5b553475ce3f.png)

MaggieCabrera:
@beafialho @kjellr can we get some names for these? Also some alt texts and file names for the images will be needed too. Thank you!

MaggieCabrera:
[...]

PR #5464, opened 2022-02-04
Archeo: Headline over dark image pattern

MaggieCabrera:

I created this pattern as a cover block with 100vh and added it to "pages" category, I'm not sure if that's the correct way to go about this one. I also had trouble reproducing the image masking from the design. Figma shows that blend mode is "pass-through" and I'm not sure if we can even do that with CSS. @beafialho @kjellr what do you think about this? It looks bad with 70 opacity with the #242321 overlay. 

Screenshot 2022-02-04 at 16 11 59

#### Related issue(s):


beafialho:
Thanks @MaggieCabrera, I tried to do that in the editor and if you use ` #242321` with 80% opacity it should look as intended:

Captura de ecrã 2022-02-04, às 15 19 45

Here's the code from my attempt: https://gist.github.com/beafialho/7b98243a03918da02aa054ad83358a1e

MaggieCabrera:
> Here's the code from my attempt: https://gist.github.com/beafialho/7b98243a03918da02aa054ad83358a1e

That helps a lot! The main problem was the base image I was using, with yours it looks nice now. Thank you!!

beafialho:
Looks great 👍

MaggieCabrera:
Oh, @beafialho I forgot, do you have an alt text for this image?

mikachan (approved):
I guess we should merge when we've added the alt text, but this looks great to me!

scruffian:
If you need to merge so we can add more patterns we can always update the alt text later

kjellr:
Alt text: "Photo of the Governor's Palace at Uxmal"

PR #5466, opened 2022-02-04
Archeo: Center footer at small resolutions

mikachan:
This centers the footer at small resolutions (599px and below). I've used some CSS and a custom class as I'm not sure how else to do this. Is there a better way?

| Desktop | Mobile |
| ------- | ------- |
|  image   | image  |

I've also used `3vw` for the block gap in the navigation block, as this makes the spacing responsive. This is similar to the spacing on the mocks, as it's slightly bigger at desktop and smaller at mobile. Again, I'm not sure if this is the best way.


scruffian (approved):
I spent a while trying to work out if this could be done without the media query, but I couldn't find a way. LGTM

kjellr:
I think we can just leave this out, rather than use custom CSS.

mikachan:
Thanks @kjellr, we'll close this one then.

PR #5467, opened 2022-02-07
Archeo: Image with headline and description pattern

MaggieCabrera:

Screenshot 2022-02-07 at 11 09 50

I made this using the media text block inside a full-width group block (for padding). @beafialho the image is exported with a white background, do you have one without it or with the background color of the site (in that case we may want the whole group block to have the same color as background for consistency)?

#### Related issue(s):


archeo/inc/patterns/image-with-headline-description.php
```diff
@@ -0,0 +1,19 @@
+<?php
+/**
+ * Image with headline and description
+ */
+return array(
+	'title'      => __( 'Image with headline and description', 'archeo' ),
+	'categories' => array( 'pages' ),
+	'content'    => '<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"5vw","right":"5vw","bottom":"5vw","left":"5vw"}}}} -->
```
jffng: Should the bottom padding be 0, that way the image reaches all the way to the bottom of the group?

jffng (commented):
The code LGTM.

I have some minor comments / questions about the design:

- The spacing between the heading and text looks tighter in the comp, should this be tightened?
- The way it stacks on mobile is the image appears on top of the text, but the comp suggests it should be the other way around.

jffng:
I made some changes:

- Tightened the spacing between the heading and the text
- Used a new responsive value for the heading that matches the "Gigantic" font size in the comps
- Reduced the bottom padding of the group container.

> The way it stacks on mobile is the image appears on top of the text, but the comp suggests it should be the other way around.

I couldn't figure out how to change the order of its stacking without adding CSS.

archeo/inc/patterns/image-with-headline-description.php
```diff
@@ -0,0 +1,18 @@
...
+	'categories' => array( 'pages' ),
+	'content'    => '<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"5vw","right":"5vw","bottom":"0","left":"5vw"}}}} -->
+	<div class="wp-block-group alignfull" style="padding-top:5vw;padding-right:5vw;padding-bottom:0;padding-left:5vw"><!-- wp:media-text {"mediaPosition":"right","mediaLink":"' . esc_url( get_template_directory_uri() ) . '/assets/images/chahk.jpg","mediaType":"image","mediaWidth":64,"verticalAlignment":"top"} -->
+	<div class="wp-block-media-text alignwide has-media-on-the-right is-stacked-on-mobile is-vertically-aligned-top" style="grid-template-columns:auto 64%"><figure class="wp-block-media-text__media"><img src="' . esc_url( get_template_directory_uri() ) . '/assets/images/chahk.jpg" alt="' . esc_attr__( 'Chahk: rain deity', 'archeo' ) . '"/></figure><div class="wp-block-media-text__content"><!-- wp:heading {"style":{"typography":{"fontSize":"clamp(64px, 6vw, 100px)","lineHeight":"1"},"spacing":{"margin":{"bottom":"48px"}}}} --><h2 id="chahk-raindeity" style="font-size:clamp(64px, 6vw, 100px);line-height:1;margin-bottom:48px">' . wp_kses_post( __( 'CHAHK:<br>RAIN<br>DEITY', 'archeo' ) ) . '</h2>
```
jffng: I have a couple questions:

- Should the font-size value be a custom variable in theme.json too? 
- Semantically, is it okay that this is an h2 instead of a paragraph? I changed it to a heading so that I could get margin control...

mikachan: > Should the font-size value be a custom variable in theme.json too?

I have the same question, and I think it would be great if it was a variable, I just don't know where it should sit in theme.json (Maybe just an extra font size? But then I don't think it's selectable from the editor.)

> Semantically, is it okay that this is an h2 instead of a paragraph?

I think this is OK, as it's acting as a heading for this pattern.

mikachan:
> I couldn't figure out how to change the order of its stacking without adding CSS.

Hmm.. @kjellr adjusted the comps for us so this would work nicely, maybe we should change them back to the other way around? 😅

archeo/inc/patterns/image-with-headline-description.php
mikachan: On the wp:heading, we could use `"textTransform":"uppercase"` to automatically make the text uppercase here.

mikachan (commented):
This is looking great, thanks for picking it up!

The jpg is pretty big at 1.29MB, do we normally compress these before bundling them?

kjellr:
> The jpg is pretty big at 1.29MB, do we normally compress these before bundling them?

Yes, but let's do them all at once after all the patterns are in the theme. If you open an issue and assign it to me, I'll take care of it. 👍

kjellr:
Let's replace the image with this transparent one for now: 

https://cloudup.com/cyioESVKP7K

kjellr:
> Hmm.. @kjellr adjusted the comps for us so this would work nicely, maybe we should change them back to the other way around? 

Oh weird. Does the Media & Text block _always_ place the image on the top?

scruffian:
> Oh weird. Does the Media & Text block always place the image on the top?

Yes, unless you disable "Stack on mobile"

kjellr:
Ah, ok. I wouldn't change that with CSS then. But if that's the case, let's switch back to the version that used right-aligned text: 

Screen Shot 2022-02-08 at 9 12 02 AM

archeo/inc/patterns/image-with-headline-description.php
scruffian: Done in 2f4587c42

scruffian:
This is ready for another review

kjellr:
Just pushed a couple small updates: 

- I removed the transparent pixels around the image. 
- I set the group parent to inherit default layout (I think this was probably left out by mistake?).

I think this should be ready to go.

PR #5470, opened 2022-02-07
Archeo: Image with headline and separate description pattern

MaggieCabrera:

I couldn't get the alignments correctly without extra CSS. This is how it looks like right now @kjellr:

Screenshot 2022-02-07 at 16 18 42

#### Related issue(s):


kjellr:
There's actually a trick you can use to get this right! Check out the "About page with Media on the Left" pattern that comes with Twenty Twenty-Two: 

https://github.com/WordPress/twentytwentytwo/blob/e9f49804eab30b51b9cd9b7d6b16674f80483b2c/inc/patterns/page-about-media-left.php

It has a Site logo at the top, and then a bunch of space, and then some more text at the bottom. The key here was: 

- Set the image to fill the space. 
- Wrap the bottom text in a group block with some dynamic top padding. This allows you to push the two areas apart on large screens, but still decrease the space to something reasonable on smaller screens.

mikachan:
Thanks for the trick, @kjellr! I've tried it out on this pattern - I set the image to fill the space, and wrapped the bottom paragraph in a group block with dynamic top padding. I think it looks better, I'm just not sure about the mobile view still:

image

There's still quite a large space between the heading and the paragraph, but I'm not sure how to fix it without CSS (if it needs fixing).

kjellr:
> There's still quite a large space between the heading and the paragraph, but I'm not sure how to fix it without CSS (if it needs fixing).

I think that's good for now! Thanks.

scruffian (approved):
LGTM

Issue #5471, opened 2022-02-07
Skatepark: Final issues


MaggieCabrera:
After a call with @kjellr this is the list of what Skatepark needs to be ready to launch:

- Changing colors in the customizer on dotcom is not applying to duotone in the frontend. If we can't solve this easily, the consensus is that we just remove the customizer stuff from the theme altogether.
- Kjell is going to work on the home template and we'll include that to the theme instead of having a static page like the demo had until now. (Home: CA8Gb16mt9yYNWOiVLAWak-fi-2981%3A2284) https://github.com/Automattic/themes/pull/5491
- The spacing between menu items is very short, this might be caused by CSS https://github.com/Automattic/themes/pull/5472
- There is a block pattern for post navigation that shows errors on the editor and the pattern inserter. (I can't reproduce this anymore!)
- Remove as much CSS as possible from the theme and refactor what we possibly can such as:
   - Instead of using the caption text, use a columns block 
   - Use paragraphs and groups instead of blockquotes 
   - Remove the author avatar
  https://github.com/Automattic/themes/pull/5473
   
After removing the excess CSS we'd need another review from design to make sure all is acceptable for launch. The idea behind this is that all this extra CSS makes it harder on us when things change and users already have content in their pages. We can't go back and remove the CSS that will break that content. Instead, if everything is handled by the blocks, we can alter the block patterns and the change will only affect patterns that have not yet been placed on the site.

PR #5472, opened 2022-02-08
Blockbase: fixed spacing for header wide

MaggieCabrera:

Header wide is used exclusively by Skatepark by default. The spacing between the items was too tight, this was caused by the gap value on Skatepark being too small compared to other themes.

I removed an extra CSS rule on Blockbase that adds gap to the navigation, this is no longer necessary after the latest changes to the block. I tested on other themes and neither where affected by this change.

Before:

Screenshot 2022-02-08 at 10 38 31

After:
Screenshot 2022-02-08 at 10 32 55

#### Related issue(s):

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -164,9 +164,6 @@ body > .is-root-container > .wp-block-template-part > .wp-block-cover,
     text-align: center;
   }
 }
-.wp-site-blocks .site-header .wp-block-navigation__responsive-container-content {
-  gap: var(--wp--style--block-gap, 2em);
-}
 
 @media (max-width: 599px) {
   .wp-site-blocks .site-header-linear .site-words-stack-small {
--- a/blockbase/block-template-parts/header-wide.html
+++ b/blockbase/block-template-parts/header-wide.html
@@ -16,7 +16,7 @@
 		</div>
 		<!-- /wp:group -->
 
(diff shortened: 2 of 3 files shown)


scruffian (approved):
LGTM. Should the top of the navigation also align with the top of the site title?

MaggieCabrera:
> LGTM. Should the top of the navigation also align with the top of the site title?

I'm not sure if that would affect other stuff (the logo, the mobile behavior...) without adding extra CSS. In any case, @kjellr what do you think about the header as it stands on this PR?

scruffian:
I think we can merge this either way...

kjellr:
>  Should the top of the navigation also align with the top of the site title?

I would like it to, yes. If it's going to take CSS though, I'd rather we just removed the tagline to achieve that alignment. 

Aside from that, is it possible to extend custom blockGap to the social links too? They don't need the same giant spacing as the rest of the nav, but just a little bit more would be good. They look a little cramped currently.

MaggieCabrera:
For the social links CSS would be needed since we are using Blockbase, I'm going to merge all Skatepark PRs and then make a new one on top of them that removes Blockbase as a parent, then we can cover the header issues again.

scruffian:
Looking at this some more, I noticed that the block itself actually has a blockGap setting. I think we should try to get this fixed, so I am suggesting https://github.com/WordPress/wordpress-develop/pull/2307

PR #5473, opened 2022-02-08
Skatepark: refactor CSS into block markup

MaggieCabrera:

As per our discussions over at https://github.com/Automattic/themes/issues/5471 this PR is here to strip as much CSS as possible from Skatepark. This is a WIP, I'll be updating as I go with screenshots of the parts affected so it's easier to review when all is done.

- Paragraph with quote without additional CSS. I improved the spacing of the columns to fit the original design more closely.

Screenshot 2022-02-08 at 11 08 29

- Image with caption pattern:

Untitled

- Testimonial pattern:

Screenshot 2022-02-08 at 11 30 13

- Author block and post meta:

Screenshot 2022-02-08 at 17 27 28

- Post comments block:

Screenshot 2022-02-08 at 12 01 03

- Search block:

Screenshot 2022-02-08 at 12 03 04

- Text list with button pattern: this one required a double group block but that allowed us to remove all related CSS so I'm happy with it, since it looks like the design

Screenshot 2022-02-09 at 11 09 07

#### Related issue(s):


MaggieCabrera:
We can't control borders for just one side of a group, so we can't achieve something like this:

Screenshot 2022-02-08 at 17 18 33

MaggieCabrera:
We will be able to remove the separator CSS when https://github.com/WordPress/gutenberg/pull/38635 gets merged

MaggieCabrera:
So this is what I'm comfortable removing, there's still some CSS but I'm not too worried about it in terms of maintenance and I think it's fairly minor so we can keep it without it being a problem for users, but I'd defer to your expertise and opinion @kjellr

kjellr (approved):
Thanks, Maggie! This is great. I see that it's still a blockbase child, and pulls in some scss from there. I think we can still followup to divorce it from that, but this is a good start.

PR #5474, opened 2022-02-08
Archeo: Add Image with headline on dark background pattern

mikachan:
This adds the 'Image with headline on dark background' to Archeo.

image

#### Related issue(s):
#5429


archeo/inc/patterns/image-with-headline-on-dark-background.php
```diff
@@ -0,0 +1,17 @@
...
+	<div class="wp-block-media-text alignwide has-media-on-the-right is-stacked-on-mobile has-background-color has-foreground-background-color has-text-color has-background"><figure class="wp-block-media-text__media"><img src="' . esc_url( get_template_directory_uri() ) . '/assets/images/figure-gigantesque.jpg" alt="Photo of gigantic figure at Izamal" class="size-full"/></figure><div class="wp-block-media-text__content">
+	<!-- wp:group {"style":{"spacing":{"padding":{"top":"clamp(30px, 4vw, 80px)","right":"clamp(30px, 4vw, 80px)","bottom":"clamp(30px, 4vw, 80px)","left":"clamp(30px, 4vw, 80px)"}}}} -->
+	<div class="wp-block-group" style="padding-top:clamp(30px, 4vw, 80px);padding-right:clamp(30px, 4vw, 80px);padding-bottom:clamp(30px, 4vw, 80px);padding-left:clamp(30px, 4vw, 80px)">
+	<!-- wp:paragraph {"placeholder":"Content…","style":{"typography":{"fontStyle":"italic","fontWeight":"300","lineHeight":"1.5"}},"fontSize":"large"} -->
+	<p class="has-large-font-size" style="font-style:italic;font-weight:300;line-height:1.5">' . wp_kses_post( __( 'Figure gigantesque, <br>à Izamal; au bas de la seconde pyramide', 'archeo' ) ) . '</p>
```
scruffian: Do we need to do anything different given that this isn't in English? Should it even be translated? @Automattic/i18n

archeo/inc/patterns/image-with-headline-on-dark-background.php
```diff
@@ -0,0 +1,17 @@
...
+	<div class="wp-block-media-text alignwide has-media-on-the-right is-stacked-on-mobile has-background-color has-foreground-background-color has-text-color has-background"><figure class="wp-block-media-text__media"><img src="' . esc_url( get_template_directory_uri() ) . '/assets/images/figure-gigantesque.jpg" alt="Photo of gigantic figure at Izamal" class="size-full"/></figure><div class="wp-block-media-text__content">
+	<!-- wp:group {"style":{"spacing":{"padding":{"top":"clamp(30px, 4vw, 80px)","right":"clamp(30px, 4vw, 80px)","bottom":"clamp(30px, 4vw, 80px)","left":"clamp(30px, 4vw, 80px)"}}}} -->
+	<div class="wp-block-group" style="padding-top:clamp(30px, 4vw, 80px);padding-right:clamp(30px, 4vw, 80px);padding-bottom:clamp(30px, 4vw, 80px);padding-left:clamp(30px, 4vw, 80px)">
+	<!-- wp:paragraph {"placeholder":"Content…","style":{"typography":{"fontStyle":"italic","fontWeight":"300","lineHeight":"1.5"}},"fontSize":"large"} -->
```
kjellr: Can we change the line height to 1.4 instead? It's looking a little too spaced out for my liking. I've updated the comps just now to use that value.

kjellr:
I'm getting an error for this one: 
Screen Shot 2022-02-08 at 8 49 41 AM

scruffian:
> I'm getting an error for this one:
Fixed!

archeo/inc/patterns/image-with-headline-on-dark-background.php
scruffian: Fixed

archeo/inc/patterns/image-with-headline-on-dark-background.php
dlind1: The translators expect English text in originals, so it'd be a problem to get it translated. On the other hand, leaving it untranslated would mean that non-English users would see the same French text in any language. 
Would it be possible to use an English original here instead?

kjellr:
This looks good for now! This might just be some weirdness in my test environment, but I can't change the font size: 

recording

🤔 Other than that, this is good to go.

mikachan:
Great, thanks!

I think I've fixed the font size issue by updating the line height in the block markup. I could see it not working too, and this seemed to fix it on my side.

archeo/inc/patterns/image-with-headline-on-dark-background.php
kjellr: I personally think it's fine to keep this in the original language. It's just a placeholder anyway.

PR #5475, opened 2022-02-08
Update alignment styles

scruffian:
This updates our alignment styles to be the same as those in https://github.com/Automattic/themes/pull/5385

To test, make sure the alignments in [this post](https://theamdemo.wordpress.com/2021/08/23/gutenberg-alignment/) work as expected in:
- Livro
- Blockbase (and children)
- Stewart
- Archeo

### DIFF
--- a/archeo/style.css
+++ b/archeo/style.css
@@ -126,10 +126,7 @@ body > .is-root-container,
 .wp-site-blocks > .wp-block-template-part > .wp-block-group.has-background,
 .wp-site-blocks > .wp-block-template-part > .wp-block-cover,
 body > .is-root-container > .wp-block-cover,
-body
-	> .is-root-container
-	> .wp-block-template-part
-	> .wp-block-group.has-background,
+body > .is-root-container > .wp-block-template-part > .wp-block-group.has-background,
 body > .is-root-container > .wp-block-template-part > .wp-block-cover,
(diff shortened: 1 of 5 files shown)


archeo/style.css
```diff
@@ -125,7 +125,6 @@ body > .is-root-container,
 .wp-site-blocks > .wp-block-cover,
 .wp-site-blocks > .wp-block-template-part > .wp-block-group.has-background,
 .wp-site-blocks > .wp-block-template-part > .wp-block-cover,
-body > .is-root-container > .wp-block-group.has-background,
 body > .is-root-container > .wp-block-cover,
 body
```
kjellr: This formatting is really weird — not sure why?

scruffian: no I'm not sure what's happened there...

jffng (approved):
This LGTM. 

To clarify though, do these rules affect any of the default templates in Stewart? As far as I can tell alignments aren't respected in the single or page templates since it has a sidebar:

Screen Shot 2022-02-08 at 10 00 09 AM

I guess the rules would apply to the Blank template... Sorry just thinking out loud here.

kjellr:
Yeah, by default they just apply to the blank template. Or, if folks went ahead and added/deleted things from any of the default templates (which they should totally be doing anyway).

PR #5476, opened 2022-02-08
Update dependencies

scruffian:
This is the output of running `npm run audit:fix`. Everything should work as before.


jffng (commented):
I get a slightly different diff, the same number of files:

> 37 files changed, 2342 insertions(+), 143512 deletions(-)

I think it's probably fine? Maybe the result of different npm versions (I ran the audit with npm v6.14.15.)

scruffian:
Probably. I ran it again with node 16 and got different results again! I'll merge the node 16 one as that's what we say in package.json...

PR #5479, opened 2022-02-08
Archeo: Add Image with description pattern

mikachan:
This adds the 'Image with description' pattern to Archeo.

image

I don't think we can adjust the size of the separator until there's a fix for this: https://github.com/WordPress/gutenberg/issues/34638

#### Related issue(s):


archeo/inc/patterns/image-with-description.php
```diff
@@ -0,0 +1,31 @@
...
+	<div class="wp-block-column" style="flex-basis:33.33%"></div>
+	<!-- /wp:column -->
+	
+	<!-- wp:column {"width":"66.66%"} -->
+	<div class="wp-block-column" style="flex-basis:66.66%"><!-- wp:separator {"color":"primary","className":"is-style-wide"} -->
+	<hr class="wp-block-separator has-text-color has-background has-primary-background-color has-primary-color is-style-wide"/>
+	<!-- /wp:separator -->
+	
+	<!-- wp:paragraph -->
+	<p>' . __( 'Ancien Temple, à Chichen-Itza, applé le chateau', 'archeo' ) ) . '</p>
```
scruffian: @kjellr can we replace this with an English string?

scruffian:
I fixed an issue with an extra )

archeo/inc/patterns/image-with-description.php
kjellr: We've got French Strings throughout actually. They're pulled directly from the image descriptions here: 

https://digitalcollections.nypl.org/items/510d47db-11e9-a3d9-e040-e00a18064a99

At first I was into keeping them, but now that I'm thinking about it more I'm not totally sure. It seems a little weird to keep it in french. For now, I think we should just use this as is. But once all the patterns are implemented I can do a sweep through them all and replace the strings. We use the same images in multiple patterns, and that way we can make sure the strings are all consistent.

scruffian: Created an issue for this here: https://github.com/Automattic/themes/issues/5485

mikachan:
> I fixed an issue with an extra )

Thanks.. bad copying and pasting from me!

I've created an issue for the string replacement, for when all the patterns are finished: https://github.com/Automattic/themes/issues/5484

scruffian (approved):

PR #5480, opened 2022-02-08
Archeo: Add Simple list of posts pattern

mikachan:
This adds the 'Simple list of posts' pattern to Archeo.

image

I've used one line of CSS for the baseline alignment for the row block, and then referenced Maggie's [GB issue](https://github.com/WordPress/gutenberg/issues/38498).

#### Related issue(s):


archeo/inc/patterns/simple-list-of-posts.php
```diff
@@ -0,0 +1,19 @@
+<?php
+/**
+ * Simple list of posts
+ */
+return array(
+	'title'      => __( 'Simple list of posts', 'archeo' ),
+	'categories' => array( 'pages' ),
+	'content'    => '<!-- wp:group {"align":"full"} -->
```
scruffian: Should this be alignfull?

mikachan: I went with alignfull because of how this pattern is used on the homepage:

image

But maybe it should be alignwide?

kjellr:
The pattern should be alignwide (here and for the homepage). On the homepage, it'll be sitting inside of a full-width group block. Other than that this looks good!

kjellr (approved):
Looks great, thank you!

PR #5481, opened 2022-02-08
Archeo: add single template

### LINKED ISSUE #5423: Archeo: Templates: Post templates - Single
With featured image:

image

Without featured image:

image

jffng:

This PR revises the single template to match the design. 

Screen Shot 2022-02-08 at 2 25 11 PM

**To test**
- Check a post with and without a featured image
- Look on mobile 

**Known issues**
- Can't vertically align the post date + title to the bottom without CSS until https://github.com/WordPress/gutenberg/issues/38498
- Can't control when to stack the featured image on mobile since it's not a media & text block, and we don't want to use the "allow items to wrap" feature of the Row block

#### Related issue(s):

Addresses #5423


kjellr:
This is cool! Is there a way we can set a minimum height so that the header are remains the same height even if there's no featured image? Currently it collapses down to this: 

Screen Shot 2022-02-08 at 2 34 57 PM

jffng:
> Is there a way we can set a minimum height so that the header are remains the same height even if there's no featured image? 

I believe the only "container-style" block that allows you to set the height is the Cover block. We could open an enhancement on the Row block upstream to request this though?

I also tried making it with the Cover block to do just this (set the vertical height even if there's no featured image) but got stuck somewhere... I could try that again.

jffng:
> This is cool! 

Also, yes! I was pretty blown away using Gutenberg to make this template and just appreciating how far it's come.

kjellr:
> I also tried making it with the Cover block to do just this (set the vertical height even if there's no featured image) but got stuck somewhere... I could try that again.

Yeah, you might be able to place a cover w/a background color + min height in the left half of that layout, and then place the site title inside of there. It's a workaround, but I think it'll work.

scruffian:
I have some questions about this one:

1. Do we not want it to stack on mobile?
Screenshot 2022-02-09 at 09 53 45
 
2. Could we use a Media & Text block?
3. Should the text be aligned to the bottom of the block?
4. Can we use a spacer to create the extra height?

scruffian:
I added a cover block for the post title:
Screenshot 2022-02-09 at 11 54 58

jffng:
Nice @scruffian, I pushed an update to the cover block approach to use the "outer spacing" value so it lines up with other elements better.

Screen Shot 2022-02-09 at 9 49 29 AM

But now that the title is bottom aligned, I'm not sure it's an improvement because most titles will get cut off (you can't read it without scrolling). Curious what @kjellr thinks:

Before (1080x720) | After (1080x720)
---|---
Screen Shot 2022-02-09 at 9 58 04 AM | Screen Shot 2022-02-09 at 9 59 42 AM

---

> Do we not want it to stack on mobile?

I think we do but I'm not sure how to achieve this without CSS.

> Could we use a Media & Text block?

AFAIK it's not possible to create with the media & text block because you can't currently assign the media portion of that block to the featured image.

> Should the text be aligned to the bottom of the block?

That's what the current design suggests, and also why I brought up that Gutenberg issue in this PR description.

> Can we use a spacer to create the extra height?

Maybe, but the cover block approach seems to do the trick.

scruffian:
> I think we do but I'm not sure how to achieve this without CSS.

What about if we used a columns block?

jffng:
> What about if we used a columns block?

That could work. We wouldn't need the cover block at that point either... Let me try it.

jffng:
Okay I tried it and I think it's a good solution since it allows us to stack without CSS, and looks good with or without a featured image. I still ended up using the cover block for the post date + title, since it gives us the option to control the height.

Some questions for @kjellr:
- We're setting the height of this section to `100vh` — is that correct?
- Should the post title be bottom aligned like the design suggests, even if it gets cut off?

scruffian:
I'm going to merge this since we need it to progress on other tasks. We can circle back to it if there are changes needed.

kjellr:
> We're setting the height of this section to 100vh — is that correct?
> Should the post title be bottom aligned like the design suggests, even if it gets cut off?

I'd set it to something like 80vh, so it still has a similar effect, but it is less likely that the post title will be cut off.

PR #5482, opened 2022-02-08
[Archeo] Update Chivo's 300 weight to 100 weight globally

### LINKED ISSUE #5478: Archeo: Can we improve the Chivo font weights?
There are several differences between the Chivo font weights in the Archeo comps compared to the browser. We did a lot of work on this [here](https://github.com/Automattic/themes/pull/5452), but we're still seeing some differences.

An example can be seen in [this pattern](https://github.com/Automattic/themes/pull/5468), but we should run through each pattern and font style to see if we can improve them.

Chivo font: https://fonts.google.com/specimen/Chivo

kjellr:
Closes #5478. 

For some reason, the 300 weight in the comps doesn't match the 300 weight used on the live site. 🤷 

This PR replaces it with the 100 weight instead. 

Before|After
---|---
Screen Shot 2022-02-08 at 2 24 58 PM|Screen Shot 2022-02-08 at 2 24 39 PM


kjellr:
I also deleted `archeo/assets/fonts/Chivo.woff2`. As far as I can tell this wasn't actually being pulled in to the theme. Not sure where it came from, but I could use another gut check on that.

jffng (approved):
Do we want to keep 300 / Chivo-Light available since the Appearance controls allow you to select Medium vs Regular?

If we're fine not supplying that, this one LGTM.

kjellr:
> Do we want to keep 300 / Chivo-Light available since the Appearance controls allow you to select Medium vs Regular?

I'm not sure. We already don't include other weights like Medium, Semibold, etc. And it seems heavy-handed to enqueue _all_ the weights by default.

jffng:
We can add it back later if we long for it.

sstabrizi:
I just had a user report that the Chivo font weights are not working at all when used on the Archeo theme. Any time they update the font-weight no change is reflected in the editor, the font weight isn't affected at all. The user stated the following:

> I can select the weight I want, but it simply doesn't appear that way on the page either in the preview or on the live site, on any computer or phone. This problem started suddenly a few days ago -- before that font weights worked fine.

35215415-hc

mikachan:
> I just had a user report that the Chivo font weights are not working at all when used on the Archeo theme. 

Hi @sstabrizi. I've tested the Chivo font-weights on a local install and I can see them working correctly. Chivo is quite a thin font, although it is possible to visually see the difference in font weights:

image

Here's the markup I used to test this:

```
<!-- wp:paragraph -->
<p>This text is the default weight and style</p>
<!-- /wp:paragraph -->
<!-- wp:paragraph -->
<p><strong>This text has been bolded</strong></p>
<!-- /wp:paragraph -->
<!-- wp:paragraph -->
<p><em>This text is in italics</em></p>
<!-- /wp:paragraph -->
<!-- wp:paragraph -->
<p><strong><em>This text is bold and italic</em></strong></p>
<!-- /wp:paragraph -->
```

On the users' site, it looks like the default font-weight for all text has been set to 'Extra bold' in the Global Styles:

image

This means that this font-weight is overriding all other font-weights set in the content. I recommend the user sets the default 'Appearance' font-weight to 'Regular' in Global Styles, and then they should see the font-weight variation correctly in the site content.

PR #5486, opened 2022-02-09
Archeo: Add Layout with two images and text pattern

mikachan:
This adds the 'Layout with two images and text' pattern to Archeo. 

image

I've used dynamic spacers to space out the columns on the left, and then made the image in the media + text block fill the entire column.

#### Related issue(s):


scruffian (approved):
LGTM. We should make sure we use the same images in the other patterns that use these ones.

PR #5487, opened 2022-02-09
Archeo: Add pattern 6

scruffian:
Adds pattern 6:
Screenshot 2022-02-09 at 11 29 18

I used the min height fix on the cover block to get the text to align to the bottom of the image.


archeo/inc/patterns/image-with-description-and-right-aligned-headline.php
```diff
@@ -0,0 +1,33 @@
...
+ * Image with description and right-aligned headline
+ */
+return array(
+	'title'      => __( 'Image with description and right-aligned headline', 'archeo' ),
+	'categories' => array( 'pages' ),
+	'content'    => '<!-- wp:columns {"align":"full"} -->
+	<div class="wp-block-columns alignfull"><!-- wp:column {"width":"33.33%"} -->
+	<div class="wp-block-column" style="flex-basis:33.33%"><!-- wp:cover {"overlayColor":"background","minHeight":670,"contentPosition":"bottom center","isDark":false} -->
+	<div class="wp-block-cover is-light has-custom-content-position is-position-bottom-center" style="min-height:670px"><span aria-hidden="true" class="wp-block-cover__background has-background-background-color has-background-dim-100 has-background-dim"></span><div class="wp-block-cover__inner-container"><!-- wp:heading {"level":3,"fontSize":"large"} -->
+	<h3 class="has-large-font-size" id="palais-du-gouverneur-a-uxmal-detail-de-la-porte-principale-1">' . esc_html__( 'Palais Du Gouverneur, à Uxmal, detail de la Porte Principale', 'archeo' ). '</h3>
```
jffng: I think there should be an uppercase text transform on this H3, as well as being right aligned:

Current | Design
---|---
Screen Shot 2022-02-09 at 9 24 52 AM | Screen Shot 2022-02-09 at 9 25 08 AM

The spacing may need to be adjusted after those changes as well.

archeo/inc/patterns/image-with-description-and-right-aligned-headline.php
```diff
@@ -0,0 +1,33 @@
...
+
+	<!-- wp:column {"width":"66.66%"} -->
+	<div class="wp-block-column" style="flex-basis:66.66%"><!-- wp:cover {"url":"' . esc_url( get_template_directory_uri() ) . '/assets/images/palais-du-gouverneur.png","id":1692,"dimRatio":0,"minHeight":600,"isDark":false} -->
+	<div class="wp-block-cover is-light" style="min-height:600px"><span aria-hidden="true" class="wp-block-cover__background has-background-dim-0 has-background-dim"></span><img class="wp-block-cover__image-background wp-image-1692" alt="' . esc_attr__( 'Palais Du Gouverneur, à Uxmal, detail de la Porte Principale', 'archeo' ). '" src="' . esc_url( get_template_directory_uri() ) . '/assets/images/palais-du-gouverneur.png" data-object-fit="cover"/><div class="wp-block-cover__inner-container"><!-- wp:paragraph {"align":"center","placeholder":"Write title…","fontSize":"large"} -->
+	<p class="has-text-align-center has-large-font-size"></p>
+	<!-- /wp:paragraph --></div></div>
+	<!-- /wp:cover -->
+
+	<!-- wp:paragraph {"textColor":"primary","fontSize":"small"} -->
+	<p class="has-primary-color has-text-color has-small-font-size">' . esc_html__( 'Photograph by Désiré Charnay, 1862 – 1863', 'archeo' ) . '</p>
```
jffng, line 25: This text and the paragraph below it lack padding on smaller viewports:

Screen Shot 2022-02-09 at 9 20 45 AM

jffng (commented):
I am getting a block validation error when I insert this pattern: 

Editor | Error message 1 | Error message 2
---|---|---
Screen Shot 2022-02-09 at 9 18 35 AM|Screen Shot 2022-02-09 at 9 22 49 AM|Screen Shot 2022-02-09 at 9 23 25 AM

scruffian:
Thanks for the review @jffng I've addressed your feedback, so ready for another look :)

jffng:
Looking good, I still encounter the block validation error when inserting, maybe someone else can verify.

archeo/style.css
```diff
@@ -113,6 +113,7 @@ body > .is-root-container,
 .edit-post-visual-editor__post-title-wrapper,
 .wp-block-group.alignfull,
 .wp-block-group.has-background,
+.wp-block-columns.alignfull.has-background,
```
jffng: I added this rule because without it, Gutenberg's default padding for columns with a background color is applying and messing up the alignment:

Before | After
---|---
Screen Shot 2022-02-09 at 12 50 55 PM | Screen Shot 2022-02-09 at 12 50 40 PM

mikachan:
I think I've managed to fix the block validation error. I'm not 100% sure exactly what was wrong, but I think it was maybe the hard-coded image IDs and some of the classes? I just re-copied the code from GB until it was happy.

archeo/style.css
scruffian: We should probably bubble this change up to TT2, Blockbase, Stewart and Livro.

jffng:
> I think I've managed to fix the block validation error. I'm not 100% sure exactly what was wrong, but I think it was maybe the hard-coded image IDs and some of the classes? I just re-copied the code from GB until it was happy.

Nice, it worked, thank you @mikachan !

archeo/style.css
kjellr: The padding in the editor is a little too intense — not sure why that doesn't match? 
Screen Shot 2022-02-10 at 11 22 12 AM

jffng (approved):

kjellr:
Left [a small comment above](https://github.com/Automattic/themes/pull/5487#discussion_r803855138), but otherwise this looks good on my end. Thanks!

archeo/style.css
jffng: I missed the equivalent rule for targeting the editor. I added it here: https://github.com/Automattic/themes/pull/5487/commits/74bfa700042a05b36f84ce74f474bc566fe0ec20#diff-2769b329cf3d32aab10e3f345c0da915cd1f6d5ffba70ebdeb7c5786597dca3dR119

Do you think these additions make sense? We are saying, "apply the site wide padding and negative margins for columns that are aligned full with a background color".

jffng:
Still not sure about this: https://github.com/Automattic/themes/pull/5487#discussion_r803914063

archeo/style.css
mikachan: > Do you think these additions make sense?

Yeah I think this makes sense, especially with the background color applied. Interested to see what anyone else thinks though, in case we're missing something.

kjellr: I think it makes sense. But maybe break this out into a separate PR and more thoroughly test it against other themes?

scruffian:
I took the alignment rule out of this before I merged it, and moved it to https://github.com/Automattic/themes/pull/5526

PR #5491, opened 2022-02-09
Skatepark: new Home template

MaggieCabrera:

This PR includes the new homepage following this design CA8Gb16mt9yYNWOiVLAWak-fi-2981%3A2284

Screenshot 2022-02-09 at 17-41-22 Skatepark – Tagline

#### Related issue(s):

### DIFF
--- a/skatepark/inc/patterns/home-patterns.php
+++ b/skatepark/inc/patterns/home-patterns.php
@@ -1,0 +1,48 @@
+<?php
+/**
+ * Block patterns for home page.
+ *
+ * @package Skatepark
+ * @since 1.0.0
+ * 
+ */
+
+return array(
+	'title'      => __( 'Block patterns for home page', 'skatepark' ),
+	'categories' => array( 'skatepark' ),
+	'inserter'   => false,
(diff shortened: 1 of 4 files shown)


scruffian (approved):
LGTM but we probably want a design review from @kjellr

skatepark/inc/patterns/home-patterns.php
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @package Skatepark
+ * @since 1.0.0
+ * 
+ */
+
+return array(
+	'title'      => __( 'Block patterns for home page', 'skatepark' ),
+	'categories' => array( 'skatepark' ),
+	'inserter'   => false,
```
kjellr, line 13: It might make sense to prefix this filename with `hidden` like we did in Twenty Twenty-Two. That way it's super-clear that this one won't be shown in the inserter.

kjellr (approved):
This looks great! I pushed one minor change, decreasing the wide width value by ~30px to push "through" onto the second line. May be a slight over-reaction, but it eliminates the "skateboarding" widow, it's simple, and it doesn't negatively effect anything else. 😄

Issue #5448, opened 2022-02-02
Blockbase Color customizations show as new "custom palette"


pbking:
## Steps to replicate

Using WP 5.9 and Gutenberg /trunk
Activate a Blockbase theme
Customize the colors using the Customizer
Load the FSE and view the GS color palette.
Note that the colors edited in the customizer are represented as CUSTOM.

<img src="https://user-images.githubusercontent.com/146530/152217906-abdaa115-4d0b-4efa-86bf-a403812ac708.png" width=300>

Originally they changed the colors shown under "Theme"

This is due to the differences as to how the colors are being persisted between Customizer and FSE GS as noted [here](https://github.com/Automattic/themes/issues/5387).

scruffian:
I pushed a fix for this to https://github.com/Automattic/themes/pull/5492 - we now save the colors on the theme key.

pbking:
That fix was [removed](https://github.com/Automattic/themes/pull/5492#issuecomment-1039440192) from #5492.  Moving this back to "TODO" for another solution.

PR #5492, opened 2022-02-09
Blockbase: Use the Global Styles rest API in the customizer

### LINKED ISSUE #5308: Blockbase: Use the Global Styles rest API in the customizer
In  https://github.com/WordPress/gutenberg/pull/35801 a Global Styles API was introduced. We should reconsider the implementation of our Customizer/Global Styles bridge, to take advantage of this new API.

Alternatively we could just use the new style switcher in Gutenberg.

From https://github.com/Automattic/themes/pull/5307

matiasbenedetto:
- Blockbase: Use the Global Styles rest API in the customizer instead of our current approach.
- Removed Duotone settings saving because there seem to be not available in the customizer
- Revert this PR: https://github.com/Automattic/themes/pull/4740

#### Related issue(s):
https://github.com/Automattic/themes/issues/5308


scruffian:
It's bit tricky to review this with the coding standards changes combined. Would you be able to split that into two different commit, or two different PRs? Thansk!

blockbase/inc/customizer/wp-customize-fonts.php
```diff
@@ -504,51 +497,41 @@ function handle_customize_save_after( $wp_customize ) {
...
-			$this->unset_property_if_it_exists( $user_theme_json_post_content->styles->elements->h1->typography, 'fontFamily' );
-			$this->unset_property_if_it_exists( $user_theme_json_post_content->styles->elements->h2->typography, 'fontFamily' );
-			$this->unset_property_if_it_exists( $user_theme_json_post_content->styles->elements->h3->typography, 'fontFamily' );
-			$this->unset_property_if_it_exists( $user_theme_json_post_content->styles->elements->h4->typography, 'fontFamily' );
-			$this->unset_property_if_it_exists( $user_theme_json_post_content->styles->elements->h5->typography, 'fontFamily' );
-			$this->unset_property_if_it_exists( $user_theme_json_post_content->styles->elements->h6->typography, 'fontFamily' );
-			$this->unset_property_if_it_exists( $user_theme_json_post_content->styles->blocks->{'core/button'}->typography, 'fontFamily' );
-			$this->unset_property_if_it_exists( $user_theme_json_post_content->styles->blocks->{'core/post-title'}->typography, 'fontFamily' );
-			$this->unset_property_if_it_exists( $user_theme_json_post_content->styles->blocks->{'core/pullquote'}->typography, 'fontFamily' );
+			unset( $new_settings['typography'] );
```
scruffian: I think this should probably only unset $new_settings['typography']['fontFamilies']['custom'], since the user might make other customizations using Global Styles.

skatepark/block-templates/page.html
```diff
@@ -8,7 +8,8 @@
 <main class="wp-block-group" style="padding-bottom:5em">
 
 <!-- wp:group {"align":"full"} -->
-<div class="wp-block-group alignfull"><!-- wp:post-featured-image {"align":"full"} /--></div>
+<div class="wp-block-group">
+<!-- wp:post-featured-image {"align":"full","style":{"color":{"duotone":["#000","#B9FB9C"]}}} /--></div>
```
scruffian, line 12: I assume we don't want this change?

skatepark/inc/patterns/blog-posts.php
```diff
@@ -15,7 +15,7 @@
 	<hr class="wp-block-separator is-style-wide"/>
 	<!-- /wp:separator -->
 
-	<!-- wp:post-featured-image {"isLink":true} /-->
+	<!-- wp:post-featured-image {"isLink":true,"style":{"color":{"duotone":["#000","#B9FB9C"]}}} /-->
 
 	<!-- wp:post-title {"isLink":true,"style":{"typography":{"fontSize":"var(--wp--custom--font-sizes--normal)"}}} /-->
```
scruffian: I assume we don't want this change?

skatepark/inc/patterns/columns-in-container.php
```diff
@@ -39,8 +39,8 @@
 	<div style="height:40px" aria-hidden="true" class="wp-block-spacer"></div>
 	<!-- /wp:spacer -->
 
-	<!-- wp:image {"align":"wide","sizeSlug":"large","linkDestination":"none"} -->
-	<figure class="wp-block-image alignwide size-large"><img src="' . get_stylesheet_directory_uri() . '/assets/images/riding-skateboard.jpeg" alt="' . esc_html__( 'Close-up of a person riding a skateboard, focusing on their feet and the board. One foot is on the board, while the other foot is up, in motion. A skatepark is blurred in the background.', 'skatepark' ) . '"/></figure>
+	<!-- wp:image {"align":"wide","id":26,"sizeSlug":"large","linkDestination":"none","style":{"color":{"duotone":["#000","#B9FB9C"]}}} -->
```
scruffian, line 42: I assume we don't want this change?

skatepark/block-templates/page.html
MaggieCabrera: +1, we can ignore all the skatepark changes, that's done differently now

blockbase/inc/customizer/wp-customize-colors.php
```diff
@@ -162,85 +146,33 @@ function handle_customize_save_after( $wp_customize ) {
...
-						);
-					}
-				}
-			}
-		}
+		// Add the updated global styles to the update request
+		$update_request = new WP_REST_Request( 'PUT', '/wp/v2/global-styles/' );
+		$update_request->set_param( 'id', $user_custom_post_type_id );
+		$update_request->set_param( 'settings', $new_settings );
+		$update_request->set_param( 'styles', $new_styles );
```
scruffian, line 167: Since we don't modify styles, do we need this line (or line 158)?

blockbase/inc/customizer/wp-customize-fonts.php
```diff
@@ -504,51 +497,41 @@ function handle_customize_save_after( $wp_customize ) {
...
+		$update_request->set_param( 'styles', $new_styles );
+
 		// Update the theme.json with the new settings.
-		$user_theme_json_post->post_content = json_encode( $user_theme_json_post_content );
-		wp_update_post( $user_theme_json_post );
+		$new_styles = $global_styles_controller->update_item( $update_request );
 		delete_transient( 'global_styles' );
 		delete_transient( 'global_styles_' . get_stylesheet() );
 		delete_transient( 'gutenberg_global_styles' );
-		delete_transient( 'gutenberg_global_styles_' . get_stylesheet() );
```
scruffian: If we don't need this line then we probably don't need the line above.

blockbase/inc/customizer/wp-customize-fonts.php
```diff
@@ -504,51 +497,41 @@ function handle_customize_save_after( $wp_customize ) {
...
-			$this->unset_property_if_it_exists( $user_theme_json_post_content->styles->blocks->{'core/post-title'}->typography, 'fontFamily' );
-			$this->unset_property_if_it_exists( $user_theme_json_post_content->styles->blocks->{'core/pullquote'}->typography, 'fontFamily' );
+			unset( $new_settings['typography'] );
 		}
 
+		// Add the updated global styles to the update request
+		$update_request = new WP_REST_Request( 'PUT', '/wp/v2/global-styles/' );
+		$update_request->set_param( 'id', $user_custom_post_type_id );
+		$update_request->set_param( 'settings', $new_settings );
+		$update_request->set_param( 'styles', $new_styles );
```
scruffian, line 542: I don't think we modify styles, so do we need this line, or line 512?

blockbase/inc/customizer/wp-customize-fonts.php
scruffian: From what I can see, this is still needed for backwards compatibility: https://github.com/WordPress/gutenberg/blob/af2d4df1a0c62206c6c6eb24cd760715a45c20e2/lib/compat/wordpress-5.9/get-global-styles-and-settings.php#L97

scruffian:
This looks great. Just a few suggestion and I think we can merge

blockbase/inc/customizer/wp-customize-fonts.php
matiasbenedetto: Yep, I think we need this line because otherwise we would be deleting the previous "styles" content if it exists.

blockbase/inc/customizer/wp-customize-colors.php
matiasbenedetto: My rationale behind this is: If the user has something stored in the styles key and we don't add it to $update_request, that content will be wiped from the database. This line avoids that.

blockbase/inc/customizer/wp-customize-fonts.php
matiasbenedetto: Great, thanks for the link. I wasn't so sure about how transient works.

pbking:
This seems to blow away font customizations.  I believe that portion will need another look.  Steps (with Blockbase):

Using the Customizer:
* Change the fonts and publish changes.  (Reload the customizer to show that they stuck)
* Change the colors and publish the changes.
* Reloading the Customizer will show that the color choices remain but the font choices are back to the default values.

I tested the same steps against /trunk to verify that the behavior was expected before this change.

matiasbenedetto:
The last problem described by @pbking seems to be caused by this commit https://github.com/Automattic/themes/commit/dded23b133a2602386638e5f9a83a9979cc312c2

The issue is related to the existence in the database of styles saved to "custom" key, example: `settings.color.pallette.custom`. With the linked commit we are saving the customized options to: `settings.color.pallette.theme` but they are not applied because it seems that `custom` has higher priority.

Knowing this: should we add a few lines to handle this issue (move existing user custom styles from `custom` to the `theme` key)? or just revert the linked commit and continue using the `custom` key? What do you think @scruffian @pbking ?

Personally, I think it would be simpler and make sense to continue using the `custom` path because these are settings set using the `customizer` but maybe there are reasons to prefer the `theme` path instead.

I reverted the commit so you can test again @pbking.

scruffian:
I think its fine to merge using the `custom` key and then address that issue separately if we want to.

pbking:
I concur with @scruffian.  This fixes the issue at hand and we can address the "custom" bit in another PR.  I removed the reference to this fixing #5448 in the description and I believe this is ready to come in.

blockbase/inc/customizer/wp-customize-fonts.php
pbking: Doing so makes the customizer behave poorly.  (I wasn't able to get to the bottom of why.  Though I agree with the sentiment @scruffian.

Much like the "remove customizations when settings are default" was causing more problems then they were solving for colors I think the same is true here.  I argue that the whole if block should be removed.  Testing of that shows things continuing to work as expected without the possible side effect of removing customizations made elsewhere.

blockbase/inc/customizer/wp-customize-fonts.php
pbking: Is 'this' the deleted line?  Or the line that isn't deleted?  My understanding is that this line should be returned before we merge.

matiasbenedetto: I added that line again

blockbase/inc/customizer/wp-customize-fonts.php
matiasbenedetto: I removed the `if` block.

pbking (approved):
This LGTM!

PR #5493, opened 2022-02-09
Try fixing the husky actions.

jffng:

Brought up by @matiasbenedetto here: p1644432879528019-slack-C029FM1EH, it looks like our pre-commit hook stopped working at some point. This PR updates it to use husky's v7 configuration (it looks like we were using v4).

**Instructions**
- `npm i`
- `composer update`

**To test**
- Make a change on this branch to some php file and commit it.
- Verify the linter runs and identifies and attempts to fix any errors.

#### Related issue(s):


scruffian:
I think husky is for JS linting, we have composer tools for PHP formatting.

mikachan:
I think we should be able to use husky to run the composer tools.

When I tried this out, I was seeing this error:

`/themes/vendor/bin/phpcbf: No such file or directory`

So I updated `phpcodesniffer-composer-installer`, then ran `composer update` and it seemed to start working. I'll commit what I did to fix it my side, but if it looks weird we can just remove it.

blockbase/inc/social-navigation.php
```diff
@@ -4,9 +4,7 @@
 // to social links in the following conditions.
 function blockbase_condition_to_render_social_menu( $block_content, $block ) {
 	// The block should be a navigation block.
-	if ( 'core/navigation' !== $block['blockName'] ) {
-		return false;
-	}
+	if ( 'core/navigation' !== $block['blockName']){  return false;}
```
scruffian: I tried lots of things but I can't get it to fix this!

mikachan: Oh it doesn't fix that for me either! I was testing in an Archeo pattern, it seemed to pick up on me removing spaces between brackets in those files. I wonder what's different about this file..

jffng:
> So I updated phpcodesniffer-composer-installer, then ran composer update and it seemed to start working. I'll commit what I did to fix it my side, but if it looks weird we can just remove it.

Oh nice, I had to update `phpcodesniffer-composer-installer` too, but not sure why that change didn't get picked up...

blockbase/inc/social-navigation.php
jffng: Weird, I just made some changes in this file to test and phpcs picked them up + fixed them: https://github.com/Automattic/themes/pull/5493/commits/caa8a5edcc67fedd47b86b697b0d0ffced0e866c

jffng:
@scruffian @mikachan any objections to bringing this one in?

mikachan:
No objections from me if it's working for you! Oh, I guess we should fix [this line](https://github.com/Automattic/themes/pull/5493/files#diff-2519f85fd55cec5b4bdd6914f6bf51cc86a4e63b697a9fa92b18b43aa058f6c6R8) before bringing it in?

jffng:
> I think husky is for JS linting, we have composer tools for PHP formatting.

Husky handles the git hooks, running the composer action for PHP formatting.

Also just noting that if we merge this, everyone will have to `npm i` and `composer update`.

scruffian:
Happy to bring it in, but I wish it worked!

jffng:
Can you be more specific about what is not working? Is it not running the hook / linter at all for you? I understand it doesn't catch or fix the bracket appearing on the same line, but I think that's just a function of the ruleset we're using.

scruffian:
I thought it would catch the spacing issues though?

scruffian (approved):
After a lot of tinkering with my setup I got this working. Thanks for fixing it!

PR #5519, opened 2022-02-10
Archeo: Add core block settings

### LINKED ISSUE #5434: Archeo: Core block settings
- [x] Pullquote
- [x] Comments
- [x] Separator

mikachan:
This adds some default block styles for the following blocks:

- Comments
- Separator
- Pullquote
- Quote

I tried to style the pullquote just with theme.json, but there were a few things I could only handle with CSS. I've added the CSS here for now.

Are there any other blocks we should add settings for?

#### Related issue(s):
Closes #5434

### DIFF
--- a/archeo/style.css
+++ b/archeo/style.css
@@ -180,36 +180,3 @@ body > .is-root-container > .wp-block-template-part > .wp-block-cover,
 	background-color: var(--wp--preset--color--background);
 	color: var(--wp--preset--color--foreground);
 }
-
-/*
- * Pullquote styles
- */
- .wp-block-pullquote blockquote {
-	 margin-top: 0;
-	 margin-bottom: 0;
- }
-
-.wp-block-pullquote p {
-	line-height: 1.2;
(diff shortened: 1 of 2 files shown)


scruffian:
IMO we should remove the custom CSS, but I'll defer to @kjellr

kjellr:
Not sure why, but the pullquote font size changes if I make it left/right aligned:

recording

Anyway, I think we can remove all the CSS we're using for the pullquote + citation and just rely on the built-in styles for now. They're not too far off. We'd also have to remove the `fontSize` rule in `theme.json`, but that's ok too. It still looks alright with those changes: 

Screen Shot 2022-02-10 at 11 37 28 AM

For the Quote block, I think we should just keep the border there, but make it 1px wide. 

Thanks!

mikachan:
Thanks both! I've made the above changes.

I'm not sure why the font size changes on the pullquote with the different alignments. I think it must be intentional, as I can see it happening in other themes as well.

PR #5521, opened 2022-02-10
Archeo: change the height of the featured image to 80vh

scruffian:
As raised in https://github.com/Automattic/themes/pull/5481#issuecomment-1034984153 let's change the height of the featured image section to 80vh.

Screenshot 2022-02-10 at 15 51 23


kjellr (approved):
LGTM. There's a bunch of extra spacing in the header too, but I removed that for testing and this looks generally correct paired with it.

jffng (approved):
LGTM.

PR #5522, opened 2022-02-10
Archeo: Tidy up responsive font sizes

kjellr:
This PR adds a more fluid mobile > desktop font scale, and switches to using `clamp()` which is a little more straightforward than the mix of `min()` and `max()` we were using before. 

## Mobile

Before|After
---|---
Screen Shot 2022-02-10 at 11 00 14 AM|Screen Shot 2022-02-10 at 11 02 00 AM

## Desktop

Before|After
---|---
Screen Shot 2022-02-10 at 10 59 49 AM|Screen Shot 2022-02-10 at 11 01 48 AM

### DIFF
--- a/archeo/theme.json
+++ b/archeo/theme.json
@@ -90,7 +90,7 @@
 				},
 				{
 					"name": "Huge",
-					"size": "clamp(42px, 7vw, 72px)",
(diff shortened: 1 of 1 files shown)


MaggieCabrera (approved):
LGTM, we should be using clamp over min/max

PR #5523, opened 2022-02-11
Try making skatepark standalone block theme

### LINKED ISSUE #5471: Skatepark: Final issues

MaggieCabrera:

This PR converts Skatepark from a Blockbase child to a standalone block theme. 
- Some of Blockbase's ponyfills are ported over such as Search block and button styles. 
- The alignments styles that we are using for Archeo, Livro, etc. 
- The header and footer are brought back from Blockbase to Skatepark.
- We now need a script to load the fonts too
- The 404 page was being inherited from Blockbase, it is now included on Skatepark
- Some of the theme.json options from Blockbase needed to be brought to Skatepark as well

#### Related issue(s):

Closes https://github.com/Automattic/themes/issues/5471


scruffian:
I noticed that in the screenshot the social links are on a different line:
screenshot


Now this is a block theme we could go for that layout...

scruffian:
> Some of Blockbase's ponyfills are ported over such as Search block and button styles.
> The alignments styles that we are using for Archeo, Livro, etc.
> The header and footer are brought back from Blockbase to Skatepark.
> We now need a script to load the fonts too
> The 404 page was being inherited from Blockbase, it is now included on Skatepark
> Some of the theme.json options from Blockbase needed to be brought to Skatepark as well

I think this is all fine, its the same approach we use for other block themes. . We will probably want to remove this extra CSS when that stuff is in Gutenberg, but we're not adding additional complexity. For the patterns I think we'll want to use the patterns from the directory instead when we can.

skatepark/theme.json
```diff
@@ -1,7 +1,28 @@
 {
 	"$schema": "https://json.schemastore.org/theme-v1.json",
 	"version": 1,
+	"customTemplates": [
+		{
+			"name": "blank",
```
scruffian, line 6: Should we also add the header + footer only template?

scruffian (approved):
LGTM

MaggieCabrera:
> I noticed that in the screenshot the social links are on a different line: 
> Now this is a block theme we could go for that layout...

I tried that but I don't think it's possible unless the social links are decoupled from the navigation block. Either everything is in the same line or everything is vertical, I didn't find a way to make the line break between the links and the social links without adding CSS.

MaggieCabrera:
@kjellr if you give a 👍 we can merge and deploy to .org

skatepark/theme.json
kjellr: Yeah, I think so.

kjellr:
Thanks @MaggieCabrera! Just a handful of notes on my end: 

---

I'm seeing a lack of left/right padding on small screens: 

Screen Shot 2022-02-15 at 2 52 39 PM

Screen Shot 2022-02-15 at 2 52 49 PM


---

Also, it seems weird that "Next Post" shows up here, even when there's no next post: 

Screen Shot 2022-02-15 at 2 55 33 PM

I think the best solution is just to hide those Previous/Next Post titles completely for now. We can probably also remove the block pattern for this as well. 

---

On the homepage, the previous & next buttons appear too close to the posts themselves, so they're easy to miss. Let's add some margin/padding/separator or something to space them out. The extra space should match the space between the "Latest Posts" header and the post below it. 

Current|Intended
---|--- 
Screen Shot 2022-02-15 at 2 56 38 PM|Screen Shot 2022-02-15 at 2 57 17 PM

---

For some reason, separators in the editor appear less opaque than they do on the front end. Additionally, if you apply a custom color, they lose the custom width we define: 

separator

---

I think we should remove the  "Two Columns of Text" pattern. I'm not a fan of the text hierarchy, and it also doesn't scale well.

MaggieCabrera:
Thanks for the review @kjellr!

> I'm seeing a lack of left/right padding on small screens:
> 
Yeah, we were missing the correct custom variables for that, it should work fine now

I fixed all the other issues as well, so this should be ready for another review

kjellr:
Thanks, @MaggieCabrera! I pushed a couple quick updates:

- I added the hover rules to the nav items too, in https://github.com/Automattic/themes/pull/5523/commits/2e44548981643022505748f7c6f4b5375ff88dd0. 
- As per @scruffian's note here, I was able to (easily!) implement the intended design for the nav in [b4aace2](https://github.com/Automattic/themes/pull/5523/commits/b4aace253e43871b452660326db2818370163d9a). 

nav

This should be ready to go!

kjellr (approved):

PR #5532, opened 2022-02-15
Changed source of pagination arrows from CSS to block attributes.

### LINKED ISSUE #5520: Zoologist: Next Page within Pagination Block has the overlapping arrow in the Editor
## Steps to replicate

Create a free plan site with Zoologist theme at https://wordpress.com/start and observe the

## Result

The right arrow that should appear after the words `Next Page` appears overlapped over the word `Next` instead. This happens only in the Editor, not on the frontend.

image

image

## Expected

See the right arrow to the right of the word `Page`.

pbking:
This change removes the :content attributes that supplied the arrows for the previous/next elements and instead assigns those via block attributes.

This fixes that issue in Zoologist, Geologist and all of Geologist's variations.

The design remains unchanged except for fixing the noted error.

Editor:
image

View:
image

Fixes #5520

### DIFF
--- a/geologist-blue/assets/theme.css
+++ b/geologist-blue/assets/theme.css
@@ -1,4 +1,3 @@
-@charset "UTF-8";
 /**
  * Breakpoints & Media Queries
  */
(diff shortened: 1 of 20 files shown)


mikachan (approved):
Yay for removing more CSS 🎉 This looks great to me and works in each theme. I can see the arrow is now in the correct position in the editor.

PR #5535, opened 2022-02-15
Remote: Remove hero from index template and make it a pattern

### LINKED ISSUE #5515: Remote: Templates: Page templates - Home
Figma: S382KokJxg0G043wsNoFQj-fi-326%3A160

The text is a block pattern. We need to add pagination to the query block (no numbers).

The Subscription block is replaced by a tags cloud block similar to the categories one.

Home

matiasbenedetto:
- Remote: Remove hero from index template
- Remote: make hero a pattern
- Remote: remove pagination numbers from home template

#### Related issue(s):
https://github.com/Automattic/themes/issues/5515

### DIFF
--- a/remote/functions.php
+++ b/remote/functions.php
@@ -159,6 +159,14 @@ endif;
 
 add_action( 'wp_head', 'remote_preload_webfonts' );
 
-// Add block patterns
-//require get_template_directory() . '/inc/block-patterns.php';
+/**
+ * Block Patterns.
+ */
+require get_template_directory() . '/inc/block-patterns.php';
+
+// Add the child theme patterns if they exist.
+if ( file_exists( get_stylesheet_directory() . '/inc/block-patterns.php' ) ) {
+	require_once get_stylesheet_directory() . '/inc/block-patterns.php';
+}
+
 
--- a/remote/functions.php
+++ b/remote/functions.php
@@ -164,9 +164,5 @@ add_action( 'wp_head', 'remote_preload_webfonts' );
  */
 require get_template_directory() . '/inc/block-patterns.php';
 
-// Add the child theme patterns if they exist.
-if ( file_exists( get_stylesheet_directory() . '/inc/block-patterns.php' ) ) {
-	require_once get_stylesheet_directory() . '/inc/block-patterns.php';
-}
 
 
--- a/remote/inc/patterns/hero-text.php
+++ b/remote/inc/patterns/hero-text.php
@@ -12,7 +12,7 @@ return array(
     
     <!-- wp:group {"align":"wide"} -->
     <div class="wp-block-group alignwide"><!-- wp:paragraph {"style":{"typography":{"lineHeight":"1.17"}},"fontSize":"huge"} -->
-    <p class="has-huge-font-size" style="line-height:1.17">Hi, I’m Tiffany! After years of long commutes and crowded offices, I’m now a happy advocate for remote work.</p>
+    <p class="has-huge-font-size" style="line-height:1.17">' . esc_attr__( 'Hi, I’m Tiffany! After years of long commutes and crowded offices, I’m now a happy advocate for remote work.', 'remote' ) . '</p>
     <!-- /wp:paragraph --></div>
     <!-- /wp:group -->
     
--- a/remote/inc/patterns/hero-text.php
+++ b/remote/inc/patterns/hero-text.php
@@ -5,16 +5,14 @@
 return array(
 	'title'      => __( 'Hero text', 'remote' ),
 	'categories' => array( 'pages' ),
-	'content'    => '<!-- wp:group {"layout":{"inherit":true}} -->
-    <div class="wp-block-group"><!-- wp:spacer {"height":"32px"} -->
+	'content'    => '<!-- wp:group {"align":"wide"} -->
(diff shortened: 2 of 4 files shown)


remote/functions.php
```diff
@@ -159,6 +159,14 @@ function remote_preload_webfonts() {
...
-//require get_template_directory() . '/inc/block-patterns.php';
+/**
+ * Block Patterns.
+ */
+require get_template_directory() . '/inc/block-patterns.php';
+
+// Add the child theme patterns if they exist.
+if ( file_exists( get_stylesheet_directory() . '/inc/block-patterns.php' ) ) {
+	require_once get_stylesheet_directory() . '/inc/block-patterns.php';
+}
```
MaggieCabrera: Remote is not a child theme, I think the require on line 165 is enough

remote/inc/patterns/hero-text.php
```diff
@@ -0,0 +1,23 @@
...
+	'title'      => __( 'Hero text', 'remote' ),
+	'categories' => array( 'pages' ),
+	'content'    => '<!-- wp:group {"layout":{"inherit":true}} -->
+    <div class="wp-block-group"><!-- wp:spacer {"height":"32px"} -->
+    <div style="height:32px" aria-hidden="true" class="wp-block-spacer"></div>
+    <!-- /wp:spacer -->
+    
+    <!-- wp:group {"align":"wide"} -->
+    <div class="wp-block-group alignwide"><!-- wp:paragraph {"style":{"typography":{"lineHeight":"1.17"}},"fontSize":"huge"} -->
+    <p class="has-huge-font-size" style="line-height:1.17">Hi, I’m Tiffany! After years of long commutes and crowded offices, I’m now a happy advocate for remote work.</p>
```
MaggieCabrera: we should make this string translateable

remote/templates/home.html
```diff
@@ -1,53 +1,32 @@
...
+	
 	<!-- wp:column {"verticalAlignment":"center","width":""} -->
 	<div class="wp-block-column is-vertically-aligned-center"><!-- wp:post-title {"isLink":true,"style":{"spacing":{"margin":{"top":"0px","bottom":"0px"}},"typography":{"fontSize":"clamp(3rem, 6vw, 4rem)"}}} /--></div>
 	<!-- /wp:column --></div>
 	<!-- /wp:columns -->
 	<!-- /wp:post-template -->
-
-	<!-- wp:spacer {"height":32} -->
+	
+	<!-- wp:spacer {"height":"32px"} -->
```
MaggieCabrera: good catch!

remote/inc/patterns/hero-text.php
```diff
@@ -0,0 +1,23 @@
...
+	'title'      => __( 'Hero text', 'remote' ),
+	'categories' => array( 'pages' ),
+	'content'    => '<!-- wp:group {"layout":{"inherit":true}} -->
+    <div class="wp-block-group"><!-- wp:spacer {"height":"32px"} -->
+    <div style="height:32px" aria-hidden="true" class="wp-block-spacer"></div>
+    <!-- /wp:spacer -->
+    
+    <!-- wp:group {"align":"wide"} -->
+    <div class="wp-block-group alignwide"><!-- wp:paragraph {"style":{"typography":{"lineHeight":"1.17"}},"fontSize":"huge"} -->
+    <p class="has-huge-font-size" style="line-height:1.17">' . esc_attr__( 'Hi, I’m Tiffany! After years of long commutes and crowded offices, I’m now a happy advocate for remote work.', 'remote' ) . '</p>
```
MaggieCabrera: I think in this case it's better to use `esc_html__`, `esc_attr__` is good when you are translating things like an alt text for example

remote/functions.php
matiasbenedetto: Removed!

remote/inc/patterns/hero-text.php
matiasbenedetto: Replaced!

MaggieCabrera:
I'm having trouble with the wide alignments when inserting this pattern into a post or page. I think the group with inherit layout could be outside of the pattern (keeping it in the template), but this issue surfaces a problem that's going to be widespread for this theme. Single and Page templates don't allow for wide or full width alignments inside their content, yet we are shipping wide aligned patterns for this theme. They do work on the site editor if we apply them in other places, but not on the content of the pages. What do you think we should do @kjellr ?

pbking:
I was having trouble with alignments too @MaggieCabrera and I pushed a change to move that "wide" assignment from the lower level group to the top level allowing us to remove that lower level group completely.  Simpler is betterer I hope.

That change causes the pattern to render fine in the VIEW but it's still having issues in the editor and I'm not sure why.

I think that we should bring this change in and iterate on it if necessary; especially as it relates to the wider issue of the templates handling of wide things.  

There are a couple of other pattern items that would be good to bring in that depend on this PR and I think this is fine to bring in.

pbking:
:/ It wasn't until I was looking at the next issue down the line until I realized what you means about moving the inherit layout group into the template.

I've now added that there, the patterns doesn't have it and other than the editor now showing things wide enough in that instance, this looks OK now.

pbking (approved):
Good for now. May revisit wide patterns in the future RE: [this comment](https://github.com/Automattic/themes/pull/5535#issuecomment-1046795697).

PR #5537, opened 2022-02-16
Try: Update and fix lint-style

mikachan:
It looks like our lint-style command is not always automatically fixing errors on the pre-commit hook. This is an attempt to fix it.

This updates all top-level `@wordpress` packages and the `lint-staged` package. It also removes the `=` from the following command, as it looks like this no longer works:

```
	"lint-staged": {
		"*.scss": [
			"wp-scripts lint-style --fix="
		]
	},
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,14 +13,14 @@
 			},
 			"devDependencies": {
 				"@octokit/rest": "^18.12.0",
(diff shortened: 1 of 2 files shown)


MaggieCabrera (approved):
Thank you for this! It's working fine now

PR #5539, opened 2022-02-16
Remote: Adding search template and moving posts list to a pattern

### LINKED ISSUE #5512: Remote: Templates: Page templates - Search

matiasbenedetto:
Remote theme:
 - Adding search template 
 - Moving posts list to a pattern
 
 :warning:  To get a cleaner commit history I pointed the base branch of this PR to `update/5515` branch.

**This PR depends on:**
https://github.com/Automattic/themes/pull/5535

 
 #### Screenshot
image

#### Related issue(s):
https://github.com/Automattic/themes/issues/5512


remote/templates/search.html
```diff
@@ -0,0 +1,15 @@
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
+
+<!-- wp:group {"layout":{"inherit":true}} -->
+<div class="wp-block-group"><!-- wp:spacer {"height":"32px"} -->
+<div style="height:32px" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- /wp:spacer -->
+<!-- wp:search {"label":"Search","width":100,"widthUnit":"%","buttonText":"Search","style":{"border":{"radius":"0px"}}} /-->
```
jffng: I wonder if this should be added as a pattern also, so that the label and text can be translated.

remote/templates/search.html
```diff
@@ -0,0 +1,15 @@
...
+<!-- wp:spacer -->
+<div style="height:100px" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- /wp:spacer --></div>
+<!-- /wp:group -->
+
+<!-- wp:pattern {"slug":"remote/posts-list"} /-->
+
+<!-- wp:template-part {"slug":"footer","tagName":"footer","className":"site-footer-container","layout":{"inherit":true}} /-->
```
jffng: Do we need to add the `site-footer-container` className here?

matiasbenedetto: Nope, it seems like it is doing nothing so I'm removing it.

remote/templates/search.html
matiasbenedetto: Good idea, I moved that code to a pattern a made those strings translatable.

kjellr:
The search page looks fine. 

Is the search form moved into a pattern so that its text string can be translated? If so, I think we should hide that pattern from the inserter, since it's basically just a Search block.

remote/inc/block-patterns.php
```diff
@@ -42,6 +42,8 @@ function remote_register_block_patterns() {
 
 	$block_patterns = array(
 		'hero-text',
+		'posts-list',
+		'search-form',
```
kjellr: Just a quick note: I recommend when we hide a pattern that we also prefix the filename with `hidden-` to make it 100% clear without needing to look in the file. It's easy to miss that little `'inserter'` line. 🙂

pbking:
I find it easier to review and progress PRs when their focus is narrow. The search templates and post list pattern seem relatively unrelated; better to have had two PRs. 

(edit): Ah, I see why you combined the two (now that I'm looking at the search template).  The tricky branch dependency was tripping me up I think.

pbking:
Rebased and resolved conflicts.

pbking:
Had to change the template, but I may have broken that when I rebased.

LGTM, bringing it in.

pbking (approved):
🚢

PR #5541, opened 2022-02-16
Archeo: Update readme.txt and style.css

### LINKED ISSUE #5428: Archeo: Pre-launch: readme.txt

mikachan:
This updates the readme.txt and style.css files for Archeo. There's lots that I'm not sure of, but I thought I'd make a start.

The theme's description is currently: "A clean portfolio theme". Is this right, or do we have a longer description?

I did originally start adding some image credits to style.css, but @kjellr has sorted this in https://github.com/Automattic/themes/pull/5544 in readme.txt. I'm guessing we don't need to repeat the credits in both files?

#### Related issue(s):
Closes #5428.


kjellr:
> I did originally start adding some image credits to style.css, but @kjellr has sorted this in https://github.com/Automattic/themes/pull/5544 in readme.txt. I'm guessing we don't need to repeat the credits in both files?

Yeah, just readme.txt is fine. Do we need to indicate any derivatives in here? I know it uses the Twenty Twenty-Two alignment code, so it might be worthwhile to do that. Something like this maybe? 

https://github.com/Automattic/themes/blob/835ac632a30da559183cbf9ecd11fa5d0aa6a071/livro/readme.txt#L37

mikachan:
Ah yeah, good idea. I've added the same line about Twenty Twenty-Two to the readme.

kjellr (approved):

Issue #5504, opened 2022-02-10
Remote: Block patterns (if needed)


MaggieCabrera:
- [x] Introduction https://github.com/Automattic/themes/pull/5553

![Block Pattern_ Introduction](https://user-images.githubusercontent.com/3593343/153411039-27174290-c9c7-4f12-8dda-74ada10d1f9f.jpg)

- [x] Posts list

![Block Pattern_ Posts List](https://user-images.githubusercontent.com/3593343/153411038-dd2f5d8f-66c4-498f-9344-d5b1e7af9d19.jpg)

- [x] Posts Grid https://github.com/Automattic/themes/pull/5533

![Block Pattern_ Posts Grid](https://user-images.githubusercontent.com/3593343/153411035-685ef9b7-6020-4390-9aee-63b566f7a1ef.jpg)

- [x] ??? https://github.com/Automattic/themes/pull/5665

![Block Pattern_ Posts Grid(1)](https://user-images.githubusercontent.com/3593343/153411032-11230296-d558-4af9-b639-39033da9fed6.jpg)

- [x] Tags

![Block Pattern_ Tags](https://user-images.githubusercontent.com/3593343/153411030-332a8221-14b7-4319-996e-e0266a401c6e.jpg)

- [x] Large Quote https://github.com/Automattic/themes/pull/5696

![Block Pattern_ Large Quote](https://user-images.githubusercontent.com/3593343/153411028-2300800f-2694-445c-95b4-e2aa4498f019.jpg)

- [x] Small Quote https://github.com/Automattic/themes/pull/5703

![Block Pattern_ Small Quote](https://user-images.githubusercontent.com/3593343/153411025-44be9406-dac3-4644-a166-07dc13ebdc46.jpg)

- [x] Categories

![Block Pattern_ Categories](https://user-images.githubusercontent.com/3593343/153411023-6732f942-10e9-4848-a079-9472c04a0f02.jpg)

- [x] Subscribe (Jetpack block) https://github.com/Automattic/themes/pull/5714

![Block Pattern_ Subscribe](https://user-images.githubusercontent.com/3593343/153411019-aa99abba-2d22-474b-8821-a0527ea2743b.jpg)

MaggieCabrera:
@kjellr @beafialho I think Remote is in a good place for a design review now. All templates and patterns are done and merged and the next steps are launch tasks only.

kjellr:
Wonderful, thank you! We'll review alongside the demo site.

[...]

PR #5543, opened 2022-02-16
Remote: add footer and tags pattern

### LINKED ISSUE #5498: Remote: Templates: Footer template part (including mobile)
The subscribe block is replaced by the tags one. @kjellr do we want the tags and categories to be part of the footer template part or just on the home page and single templates?

Screenshot 2022-02-10 at 13 51 11

jffng:

This PR revises the footer template and adds the Tags pattern. 

I did not include the tags in the footer template, since it didn't make sense to have this section on every page. Not sure if folks have an opinion about this.

Screen Shot 2022-02-16 at 3 28 35 PM

To test:
- Go to the site editor and insert / verify the footer appears
- Insert the tags pattern

#### Related issue(s):

https://github.com/Automattic/themes/issues/5498
https://github.com/Automattic/themes/issues/5504 (Tags pattern)

### DIFF
--- a/remote/style.css
+++ b/remote/style.css
@@ -196,22 +196,17 @@ a:not(
 	outline-offset: 3px;
 }
 
-/*
- * Needed until https://github.com/WordPress/gutenberg/issues/30472
- */
-.wp-block-post-terms .wp-block-post-terms__separator {
-	display: none;
+.wp-block-tag-cloud.is-style-outline {
+	display: flex;
+	flex-wrap: wrap;
 }
 
 /*
  * Needed until https://github.com/WordPress/gutenberg/issues/34196 or something like it.
  */
-.wp-block-post-terms > a {
-	border: 1px solid var(--wp--preset--color--foreground);
-	border-radius: 6px;
-	display: inline-block;
-	padding: .8rem 1.25rem;
-	margin-right: var(--wp--style--block-gap);
-	margin-bottom: var(--wp--style--block-gap);
-	text-decoration: none;
+.wp-block-tag-cloud.is-style-outline a {
+	border-radius: 10px;
+	padding: .5rem 1.25rem;
+	margin-right: calc( 0.5 * var(--wp--style--block-gap) );
+	margin-bottom: calc( 0.75 * var(--wp--style--block-gap) );
 }
--- a/remote/style.css
+++ b/remote/style.css
@@ -196,17 +196,10 @@ a:not(
 	outline-offset: 3px;
 }
 
-.wp-block-tag-cloud.is-style-outline {
-	display: flex;
-	flex-wrap: wrap;
-}
-
 /*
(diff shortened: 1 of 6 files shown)


MaggieCabrera:
I think the tags are not the post tags but the tag cloud block (for the whole site). The design is already showing the tags cloud on the home page footer so I think it's safe to assume that is more or less global. I'm going to work on that assumption and then we can move it if it needs be.

Also, by using `<!-- wp:post-terms {"term":"post_tag"} /-->` we were altering the single post tags:

Screenshot 2022-02-21 at 17 52 59

I made some adjustments to the CSS so the pills don't look so big:

Screenshot 2022-02-21 at 18 03 38

pbking (approved):
This LGTM.

scruffian:
Should the tags be upper case?

remote/style.css
```diff
@@ -106,4 +106,19 @@ body > .is-root-container,
 	border: 1px solid var(--wp--preset--color--foreground);
 	background-color: transparent;
 	color: var(--wp--preset--color--foreground);
+}
+
+.wp-block-tag-cloud.is-style-outline {
+	display: flex;
```
scruffian: What does this solve? Could we add it to the block?

MaggieCabrera: currently, they have inline-block instead of flex and that creates a 4px ish gap between items that you can't remove via margin/padding, so flex is better. It can be added to the block but it would be a very minor change that would make more sense in context with adding options to alter that spacing between the pills

scruffian: Done here: https://github.com/WordPress/gutenberg/pull/38995

scruffian: These lines are in core now, so we can remove them from here.

jffng:
I made the tags uppercase and removed the CSS that's in Gutenberg now.

remote/theme.json
```diff
@@ -102,6 +102,11 @@
 				"border": {
 					"color": "var(--wp--preset--color--primary)"
 				}
+			},
+			"core/tag-cloud": {
+				"typography": {
+					"textTransform": "uppercase"
```
jffng: Not sure if this should be done here or in the template.

scruffian: I prefer it here, unless there's a reason not to have it apply to every block.

MaggieCabrera: I don't understand, they are not uppercase in the design:
Screenshot 2022-02-25 at 09 26 55

The tags/terms block is the one that is uppercase:

Screenshot 2022-02-25 at 09 27 51

scruffian: Oh sorry my bad

scruffian: Updated

scruffian (approved):
Sorry about the confusion over the text transform. I should have checked the comps instead of misremembering them. I think we can bring this in.

PR #5553, opened 2022-02-17
Remote: Add Introduction pattern

mikachan:
Small PR to add the Introduction pattern to Remote (well, technically moves it from the home template to a pattern!)

#### Related issue(s):

### DIFF
--- a/remote/functions.php
+++ b/remote/functions.php
@@ -160,5 +160,5 @@ endif;
 add_action( 'wp_head', 'remote_preload_webfonts' );
 
 // Add block patterns
-//require get_template_directory() . '/inc/block-patterns.php';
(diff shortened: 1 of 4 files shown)


MaggieCabrera:
I think this was already done in https://github.com/Automattic/themes/pull/5535/files but I think I prefer the pattern to not have all the group blocks and spacers included on it

mikachan:
Arg sorry I didn't see that! Happy to close this.

I think I prefer it without the group block at least, as then it's one less thing for users to change if they're inserting it inside another group block/any other block. I left the bottom spacer as I thought it was more likely that users would tend to use that one, but again it's easy for them to add/remove their own spacers.

pbking:
Done in #5535, closing this one.

PR #5557, opened 2022-02-17
Varia themes: fix cover block text color in editor

### LINKED ISSUE #5181: Varia Themes: Cover block -  text changes color upon editing anything on the same page
### Quick summary

Multiple users have reported that on live sites using the Cover block, the text has previously been white, and has since changed to grey or black (an unreadable color) after another unrelated element is edited. 

- No change in the HTML of the block in Gutenberg seems to exist. 
- The block looks the same with white text in the editor, it's the live view that has a different color.
- In the CSS, the only thing that is different is that  the `has-background-dim` class is moved down one level from `.wp-block-cover` to `.wp-block-cover__gradient-background`. It seems like `color: var(--wp--preset--color--background);` is set for both, though, so I have no idea what's happening there. All blocks inside still use inherit to get text color. 

### Steps to reproduce

1. On a site with existing Cover block templates with text in them (I used the Maywood default homepage in my test), activate Hever.
2. Make any change at all to anything on the page with the cover blocks in the editor. 

The text should stay white, as it does in the editor. 

### What actually happened

The text goes dark grey or black and is unreadable on a dark-overlay image. 

### Context

Most recently, 33065416-hc (AT) Also in 28882174-hc (Simple). Test site is available at hgshdfhasdfwerfas.wordpress.co ⁠— all I did to get that color change to trigger was change the _Welcome to Maywood_ H3 to a H4. 

### Simple, Atomic or both?

Simple, Atomic

### Theme-specific issue?

Both these sites are on Hever, but it might be broader than that! I haven't tested other themes. 

_No response_

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

Yes, easy to implement

### Workaround details

Manually setting text color works, but it's not an easy alternative for sites with lots of cover images.

jffng:

This PR removes some CSS that is causing a discrepancy in the editor vs front-end of the cover block's text color in Varia themes.

To Test:
- Follow the instructions in [this comment](https://github.com/Automattic/themes/issues/5181#issuecomment-988773746)
- Verify the text color in the front-end and editor match

#### Related issue(s):

Closes https://github.com/Automattic/themes/issues/5181


varia/sass/blocks/cover/_editor.scss
```diff
@@ -12,16 +12,15 @@
 		color: #{map-deep-get($config-cover, "color", "foreground")};
 	}
 
-    .wp-block-cover__inner-container,
```
jffng, line 15: The only meaningful change is removing this selector. Everything else is linting + the result of rebuilding the varia themes.

pbking: Thank you for pointing that out. :)

pbking (approved):
🚢

PR #5558, opened 2022-02-17
Add utilities to sync core themes to wpcom

pbking:
This adds a utility (and a push-button-process) to sync changes from the wporg repository to wpcom.

First is the command to sync those changes: `npm run core:sync THEMESLUG [REVISION]` 
(Until the theme has a stored .pub-svn-revision file (created by the utility) the REVISION must be given on the command line.  Based on comments in the repository 52047 seems to be the best revision to start with.)

This command creates a DIFF from the revision given to HEAD and applies it to the GIT repository on your sandbox.  Any conflicts that cannot be resolved are stored in .rej files.

Additionally a "push button" command has been created to assist with the entire process of cleaning, syncing, branching, committing, building, landing deploying and building zip files.  Use `npm run deploy:core:sync THEMESLUG [REVISION]` to use a wizard to guide you through that process (much like `npm run deploy` does for the free themes).

That process would look like this:
```
npm run deploy:core:sync twentytwenty 52047
... 
patch applied
fix any conflicts, remove any .rej files  
ensure theme works as expected
hit y to continue
...
Phabricator diff is created
review the diff
hit y to continue
...
changes are landed
theme is deployed
zip is built
```

Sample diff: D75201-code

The conflict resolution process has to happen on the sandbox and leveraging the [Remote - Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) plugin for VSCode is very helpful with that!

Additionally two other commands have been added to help debug and evaluate the changes made locally.  `npm run core:pull` and `npm run core:push` will rsync the core themes from your sandbox.

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -9,6 +9,7 @@ const sandboxPremiumThemesFolder = '/home/wpdev/public_html/wp-content/themes/pr
 const sandboxRootFolder = '/home/wpdev/public_html/';
 const isWin = process.platform === 'win32';
 const premiumThemes = [ 'videomaker', 'videomaker-white' ];
+const coreThemes = ['twentyten', 'twentyeleven', 'twentytwelve', 'twentythirteen', 'twentyfourteen', 'twentyfifteen', 'twentysixteen', 'twentyseventeen', 'twentynineteen', 'twentytwenty', 'twentytwentyone', 'twentytwentytwo'];
 
 (async function start() {
 	let args = process.argv.slice(2);
@@ -26,6 +27,10 @@ const premiumThemes = [ 'videomaker', 'videomaker-white' ];
 		case "deploy-preview": return deployPreview();
 		case "deploy-theme": return deployThemes([args?.[1]]);
 		case "build-com-zip": return buildComZip([args?.[1]]);
+		case "pull-core-themes": return pullCoreThemes();
+		case "push-core-themes": return pushCoreThemes();
+		case "sync-core-theme": return syncCoreTheme(args?.[1], args?.[2]);
+		case "deploy-sync-core-theme": return deploySyncCoreTheme(args?.[1], args?.[2]);
 	}
 	return showHelp();
 })();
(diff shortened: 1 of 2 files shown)


scruffian:
When I tried this for twentynineteen I got a lot of .rej files for changes to the version number 2.1-wpcom instead of 2.1. I guess this will be the case for all core themes. I wonder if there's another way to handle this so that we don't have to fix them manually every time?

pbking:
Handling the rejected merges for all fo the -wpcom changes is the hardest part of the process and yes, likely for every merge for every theme every time.  I agree that scripting that might be the best avenue.  It may be good to parse those .rej for that particular change as part of the automated process.  

Does that sound like a good follow up to this PR?

theme-utils.mjs
```diff
@@ -646,6 +706,50 @@ async function pushChangesToSandbox() {
...
+
+async function syncCoreTheme(theme, sinceRevision) {
+	if(!theme){
+		console.log('Must supply theme to sync and revision to start from');
+		return;
+	}
+	if(!sinceRevision) {
+		sinceRevision = await executeOnSandbox(`cat ${sandboxPublicThemesFolder}/${theme}/.pub-svn-revision`);
+	}
+	let latestRevision = await executeCommand(`svn info -r HEAD https://core.svn.wordpress.org/trunk | grep Revision | egrep -o "[0-9]+"`);
```
jffng, line 735: Thank you for working on this!!

To be sure, we want to use `core.svn` instead of `develop.svn`? I'm just wondering if there would be any benefit to syncing to the development version of Core, since it is often ahead of a release.  

For example, right now there are some changes to TT2 on trunk right now that would be good to sync back to wpcom, but would not be reflected if we point to `core.svn`.

pbking: That's a decision I'm happy for !me to make. :)

I think that syncing from develop is likely more the exception.  I see it as a tool mostly to sync "unknown changes that have happened" than "specific work that we need in wpcom sooner-than-later".

Perhaps it should be an argument?

jffng: Argument makes sense to me and could be addressed in a follow up :)

scruffian (approved):
LGTM

PR #5559, opened 2022-02-18
Hari: Initial commit.

alaczek:
This is the initial commit of Hari - a minimal, product-oriented Blockbase child theme (generated via the Create Blockbase Theme plugin). More info here: p9Jlb4-3N0-p2

Frame 24


hari/readme.txt
```diff
@@ -0,0 +1,43 @@
+=== Hari ===
+Contributors: Automattic
+Requires at least: 5.8
+Tested up to: 5.8
```
MaggieCabrera: this is 5.9, we'll update the plugin so new themes get that by default

hari/templates/404.html
```diff
@@ -0,0 +1,10 @@
+<!-- wp:template-part {"slug":"header-minimal","tagName":"header"} /-->
```
MaggieCabrera: to avoid duplicating templates just to add the different header, we changed Blockbase so all templates include the header.html template part, and inside that template part is where you insert the pattern you prefer, like `<!-- wp:pattern {"slug":"blockbase/header-minimal"} /-->`. Here's an example: the [index.html template](https://github.com/Automattic/themes/blob/f554716f193228fdbb8d7ee46c92a75280109b3f/videomaker/block-templates/index.html#L1) calls the header template part and the [header.html part](https://github.com/Automattic/themes/blob/trunk/videomaker/block-template-parts/header.html) includes the correct block pattern.

That way we only include templates that actually have changes on them, you don't need to copy over the 404, search, blank template, etc unless you want to modify further how they look.

pbking: Updated.  New Blockbase variations and derivatives should follow that design.

hari/readme.txt
pbking: updated here.  will add it to the list of things to update for the plugin.

pbking:
I've updated this PR per @MaggieCabrera 's comments so y'all can continue efforts on this theme.  I believe it's ready to bring in this initial commit.

hari/readme.txt
```diff
@@ -0,0 +1,43 @@
+=== Hari ===
+Contributors: Automattic
+Requires at least: 5.9
```
scruffian:
```suggestion
Requires at least: 5.8
```

scruffian (approved):
LGTM

PR #5561, opened 2022-02-18
Marl: Initial commit.

alaczek:
This is the initial commit of Marl - a minimal, product-oriented Blockbase child theme (generated via the Create Blockbase Theme plugin). More info here: [p9Jlb4-3N0-p2](https://href.li/?p9Jlb4-3N0-p2)

Frame 25


pbking:
As @MaggieCabrera mentioned in the Hari PR the header customizations can be made in just the header.html template part.  

I made that change and believe the design remains consistent.  I think this is ready to come in.

~~I found that, because Blockbase currently uses `block-template-part` folder that this child must also in order for those customizations to be picked up (`/template-parts/header.html` didn't work as expected while `/block-template-parts/header.html` did).~~

MaggieCabrera:
I believe the folder is called just `parts`

pbking:
> I believe the folder is called just `parts`

Thank you @MaggieCabrera!  Fixed.

marl/readme.txt
```diff
@@ -0,0 +1,41 @@
+=== Marl ===
+Contributors: Automattic
+Requires at least: 5.9
```
scruffian:
```suggestion
Requires at least: 5.8
```

scruffian (approved):
LGTM

alaczek:
Thank you both!

PR #5562, opened 2022-02-18
Archeo: Move footer part markup into pattern

mikachan:
This moves the Archeo default footer part markup into a pattern. 

I was seeing the 'invalid content' block error in the editor with the old one. Moving it into a pattern seems to fix this and allows the 'proudly powered by' text to be translatable.

This also removes the `site-footer-container` class, as I don't believe this is needed as there are no custom styles.

To test, make sure you don't see the 'attempt block recovery' error in the editor for the footer.


jffng (approved):
LGTM, block validation error goes away 👍

PR #5563, opened 2022-02-18
Archeo: Remove outer group from header

mikachan:
This removes the outer container group block from Archeo's header. It looks like we don't need it, and the `"layout":{"inherit":"true"}` was causing it to not fill the entire site width.

This also removes the `site-header` class. I don't think this is needed as we don't have any custom styles for this class.

To test, make sure that the header fills the full content width, like in the comps.


jffng (approved):
👍 

Before | After
---|---
Screen Shot 2022-02-18 at 8 52 31 AM | Screen Shot 2022-02-18 at 8 52 19 AM

PR #5568, opened 2022-02-21
Blockbase: Update Gutenberg compatibility

scruffian:
When the Gutenberg plugin isn't active Blockbase fonts don't work. Now that Global Styles are in core we can check for the presence of the core class rather than the Gutenberg one, which means that Blockbase no longer needs Gutenberg to work.

### DIFF
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -34,7 +34,7 @@ if ( ! function_exists( 'blockbase_support' ) ) :
 		register_nav_menus(
 			array(
 				'primary' => __( 'Primary Navigation', 'blockbase' ),
-				'social' => __( 'Social Navigation', 'blockbase' )
+				'social'  => __( 'Social Navigation', 'blockbase' ),
 			)
 		);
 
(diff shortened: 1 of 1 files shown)


scruffian:
Thinking about it some more I'm not sure about this approach. Do we still want Blockbase to work with 5.8 as well?

pbking:
If it's a choice of working with 5.8 (with Gutenberg) or 5.9 (without Gutenberg) I think that working with 5.8 sounds more important to me.  I have no qualms with Gutenberg always being a requirement for Blockbase.

scruffian:
Maybe we can do both...

scruffian:
I added a commit which adds compatibility for 5.8 and 5.9.

blockbase/functions.php
```diff
@@ -152,7 +153,7 @@ function blockbase_fonts_url() {
 /**
  * Customize Global Styles
  */
-if ( class_exists( 'WP_Theme_JSON_Resolver_Gutenberg' ) ) {
+if ( class_exists( 'WP_Theme_JSON_Resolver' ) || class_exists( 'WP_Theme_JSON_Resolver_Gutenberg' ) ) {
```
jffng, line 156: We call this class [in a few places](https://github.com/Automattic/themes/search?q=WP_Theme_JSON_Resolver_Gutenberg), so I think we need to determine which class is being called (plugin or Core) and reference that one? Otherwise when I try to use 5.8 + Gutenberg, I get:

`Fatal error: Uncaught Error: Class 'WP_Theme_JSON_Resolver_Gutenberg' not found in /Users/jeffreyong/Documents/THEMES/a8c-projects/themes/blockbase/inc/customizer/wp-customize-colors.php on line 56`

pbking: I would rather have Blockbase require Gutenberg than try to ensure that what we does can work with-or-without.  If we have to check for the gutenberg function anyway to operate on 5.8 I suggest we just check for that one.

scruffian: Yeah not sure this is worth the effort, lets close.

PR #5569, opened 2022-02-21
update parent theme for marl and attar

scruffian:
The parent theme slug should be lowercase


pbking (approved):
👍

PR #5570, opened 2022-02-22
Dorna: Initial commit.

alaczek:
This is the initial commit of Dorna - a minimal, product-oriented Blockbase child theme (generated via the Create Blockbase Theme plugin). More info here: [p9Jlb4-3N0-p2](https://href.li/?p9Jlb4-3N0-p2)

Here's a preview:
Dorna


MaggieCabrera:
I fixed the paragraph issue for you, the styles key needs to be at the same level as settings, not inside it. It should work as intended now, let us know if you need more help!

alaczek:
Thank you! Is it good to merge now?

MaggieCabrera (approved):

PR #5571, opened 2022-02-22
Winkel: Initial commit.

alaczek:
This is the initial commit of Winkel - a minimal, product-oriented Blockbase child theme (generated via the Create Blockbase Theme plugin). More info here: [p9Jlb4-3N0-p2](https://href.li/?p9Jlb4-3N0-p2)

Here's a preview:
Frame 4265


winkel/style.css
```diff
@@ -0,0 +1,16 @@
+/*
+Theme Name: Winkel
+Theme URI: https://wordpress.com/theme/winkel
+Author: Automattic
+Author URI: https://automattic.com/
+Description: Winkel is a minimal, product-oriented theme.
+Requires at least: 5.9
```
scruffian:
```suggestion
Requires at least: 5.8
```

scruffian:
In this case the only difference between the templates in this theme and in Blockbase is the header, so I renamed the template part to `header` and then simply removed all the other templates.

I noticed that the header template is a bit different to the header-centered pattern in Blockbase, but I'm not sure if that's deliberate. For example the site logo block looks like this:

`<!-- wp:site-logo {"width":64,"shouldSyncIcon":false,"align":"center"} /-->`

Is there a reason to disable syncing the site logo in this theme?

The navigation also uses a small font size, but we might want to set that in theme.json.

alaczek:
> Is there a reason to disable syncing the site logo in this theme?

Nope, I just missed this. Should be fixed now.

> The navigation also uses a small font size, but we might want to set that in theme.json.

Good point, I'll take care of it. 

Is it good to go now?

scruffian:
I think with those changes to the header, you should be able to replace the contents of the header template part with a reference to the pattern...

scruffian:
I added a commit to refactor the header-centered pattern so that it can be used more widely. Does the header still look right in Winkel?

@Automattic/theam this is a really good example PR - In the original PR the code changes were massive, but in the final version they are very minimal. This really shows the power of theme.json when combined with patterns.

pbking:
Seedlet-blocks (the only theme using that pattern) checks out.  This LGTM.

🚢

pbking (approved):

PR #5572, opened 2022-02-22
Skatepark: updated changelog

MaggieCabrera:
I believe we needed to update the changelog before resubmitting Skatepark to the directory. This PR adds that and a version bump


mikachan (approved):
LGTM 👍

PR #5573, opened 2022-02-22
Archeo: Remove text decoration from post titles

scruffian:
This removes the underline from the post title links on the home page:

Screenshot 2022-02-22 at 13 54 05

Needs https://github.com/WordPress/gutenberg/pull/38986 to work. I'm hesitant to add more CSS for this when the fix in Gutenberg is simple.


MaggieCabrera (approved):
I agree, and this LGTM, let's move this to approved but not mergeable then and if the GB PR stalls then we can think of an alternative

MaggieCabrera:
mmh, looks like this is the only issue missing before launch, and the home page looks a bit strange with the underline. Should we patch this so we can submit and then introduce this fix when it lands on GB?

beafialho:
I'm still seeing the underlines by default on my local site but the screenshot looks good. If the post titles are underlined on hover, we can merge this.

scruffian:
Let's just merge with a patch

mikachan:
I've added a CSS patch for the text-decoration styles, and I've made sure the post titles are underlined on hover. I think this should be ready to come in now 🤞

archeo/style.css
```diff
@@ -198,8 +199,19 @@ body > .is-root-container > .wp-block-template-part > .wp-block-cover,
...
+ * Needed until https://github.com/WordPress/gutenberg/issues/38498 is fixed.
  */
 .wp-block-group.simple-list-of-posts {
 	align-items: baseline;
 }
+
+/*
+ * Needed until https://github.com/WordPress/gutenberg/pull/38986 is merged.
+ */
+.wp-block-group.simple-list-of-posts a {
```
scruffian: Do we want to tie this to the custom class name, or should it apply to all instances of the block?

mikachan: Hmm, I guess I was thinking of this as being a specific pattern style, but maybe we do want this to apply to all post title blocks. Now you've mentioned it, I'm thinking this would probably be the better default style for users.

@kjellr @beafialho, what do you think?

kjellr: Making it the default for all Post Titles sounds good to me.

mikachan: Thanks @kjellr, I've applied these styles to all Post Titles now.

MaggieCabrera (approved):

PR #5577, opened 2022-02-23
Blockbase: fix alignments

scruffian:

Blockbase was missing a declaration of inheriting layout on the home template, which causes some alignment issues:

Before:
Screenshot 2022-02-23 at 09 52 09

After:
Screenshot 2022-02-23 at 09 51 57

### DIFF
--- a/blockbase/block-templates/index.html
+++ b/blockbase/block-templates/index.html
@@ -3,7 +3,7 @@
 <!-- wp:query {"tagName":"main","layout":{"inherit":true}} -->
 <main class="wp-block-query">
 <!-- wp:post-template -->
-<!-- wp:group -->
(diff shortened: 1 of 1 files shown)


blockbase/block-templates/index.html
```diff
@@ -3,7 +3,7 @@
 <!-- wp:query {"tagName":"main","layout":{"inherit":true}} -->
 <main class="wp-block-query">
 <!-- wp:post-template -->
-<!-- wp:group -->
+<!-- wp:group {"layout":{"inherit":true}} -->
```
MaggieCabrera, line 6: this is already inside a layout inherit, even though this solves the bug, the original markup should have been working in any case

scruffian:
This is a bandaid not a fix.

PR #5580, opened 2022-02-23
Remote: Comment form styles

### LINKED ISSUE #5501: Remote: Comment form styles (dotcom and dotorg) (if needed)
@kjellr @beafialho the form is currently looking like this:
image

matiasbenedetto:
Remote: Comment form styles

#### Related issue(s):
https://github.com/Automattic/themes/issues/5501


jffng:
Here is a corresponding wpcom patch D75679-code required to not break the comment form there.

scruffian:
@kjellr do you think it's worth the custom CSS for this?

kjellr:
> @kjellr do you think it's worth the custom CSS for this?

I think the CSS that styles the form fields is fine, but I would leave out the `grid` stuff, since it's not strictly necessary to re-arrange items.

matiasbenedetto:
@kjellr this is the mockup:
image


This is the styles implementation with grid:
image


This is the syles implementation without grid:
image

Should we remove the grid css?

kjellr:
That's fine. We should fix the padding on the submit button too while we're at it.

matiasbenedetto:
> That's fine. We should fix the padding on the submit button too while we're at it.

@kjellr sorry but I didn't get what is fine. Should we remove the grid CSS or not?

kjellr:
Yes, please remove the grid.

pbking (approved):
LGTM
🚢

PR #5586, opened 2022-02-24
Blockbase: Remove list ponyfill

scruffian:
Now https://github.com/WordPress/gutenberg/pull/27511 is closed we don't need the Blockbase ponyfill. This removes it for Blockbase. No child themes implemented it. @alaczek could you check whether any premium themes use this custom property?


mikachan (approved):
This LGTM 👍

Tentatively approving until @alaczek has a chance to test as well.

scruffian:
I think it's ok to merge this as the premium themes use a different version of Blockbase, so they can manage their upgrade process independently.

PR #5609, opened 2022-03-02
Archeo: Add fonts to theme.json

scruffian:
This moves the font definitions to the theme.json file. This was made possible because  https://github.com/WordPress/gutenberg/pull/37140 was merged.

To test check that the font loads as before.

I'm not sure if we can also remove the pre-loading script.

#### Related issue(s):
https://github.com/WordPress/gutenberg/pull/37140


archeo/functions.php
```diff
@@ -63,119 +60,5 @@ function archeo_styles() {
...
-	 * Only the main web font (font-weight: 100,400, font-style: normal) is preloaded here since that font is always relevant. 
-	 * The other fonts are only needed if the user changed style or weight of the fonts,
-	 * and therefore preloading it would in most cases regress performance when that font would otherwise not be loaded
-	 * at all.
-	 *
-	 * @since Archeo 1.0
-	 *
-	 * @return void
-	 */
-	function archeo_preload_webfonts() {
```
scruffian, line 168: We might still need this.

mikachan: Yeah it looks like we still need this to preload the fonts.

mikachan (approved):
Thanks for doing this! This seems to work great for me, all the fonts load as they did before. The only difference I could see was that the main font files are no longer preloaded, so I think we should keep that function as you mentioned.

MaggieCabrera:
We should hold off on merging this until it makes it to dotcom, right?

mikachan:
Yes I think so, it looks like it'll be included in GB 12.8.

MaggieCabrera:
We should do this to Stewart, Livro and Remote too

scruffian:
> We should do this to Stewart, Livro and Remote too

Done

scruffian:
Do you think we should be preloading fonts? [This comment](https://github.com/WordPress/gutenberg/pull/39391#issuecomment-1067674628) is very interesting. Maybe we just need to change font-display to 'block'? See https://css-tricks.com/almanac/properties/f/font-display/

MaggieCabrera:
> Do you think we should be preloading fonts? [This comment](https://github.com/WordPress/gutenberg/pull/39391#issuecomment-1067674628) is very interesting. Maybe we just need to change font-display to 'block'? See https://css-tricks.com/almanac/properties/f/font-display/

Makes sense to me, if CSS already has that option

scruffian:
I added a commit to remove the code that preloads fonts. Instead we can use `font-display: block` which means that if the font is available we'll use it but if not then we'll fallback to a default so that users can at least read the content.

MaggieCabrera:
I think we can bring this in now. I tested on my sandbox and it's working, someone else have a look and then we can merge, please?

jffng:
> I think we can bring this in now. I tested on my sandbox and it's working, someone else have a look and then we can merge, please?

Tested across all affected themes and the fonts are loading as expected.

jffng (approved):

Issue #222, opened 2018-08-07
Dara: Turning on AMP page in Jetpack hides the mobile menu from the theme


mapk:
## Steps to replicate

Install Dara, and go to the Jetpack screen. Turn on AMP and view on a mobile device. The menu is not available.

Reported in Zendesk: Incident #1329550

laurelfulford:
This doesn't seem to be theme-specific -- I was able to reproduce with Twenty Sixteen -- but needs more investigation to see if it's something that needs to be addressed in the themes, or in the AMP plugin.

xpurichan:
The user is checking status of this, I let them know we're still working on things.

It does seem to be a high priority type of issue if we're able to test and validate this is present on AMP pages across multiple themes.

alaczek:
I'm seeing the same behaviour using Baskerville 2, but I'm wondering if this is not how AMP is supposed to work. 
I search the AMP plugin support forums on .org, and it sounds like AMP does not support mobile menu display: https://wordpress.org/support/topic/no-amp-menu-with-you-plugin/

The thread mentions this should be possible in the [1.0-beta2 version](https://github.com/Automattic/amp-wp/releases/tag/1.0-beta2) of the plugin, which I don't think is running on WordPress.com (but integration on theme side is likely required too). See: https://github.com/Automattic/amp-wp/wiki/Implementing-Interactivity#hamburger-menu-toggle

To sum up, this sound like an all-themes issue, and might require more resources.

laurelfulford:
Switching from `needs patch` to `needs testing`, because this sounds like it needs more digging (and WP.com updates outside of themes) before this can be moved forward.

scruffian:
This sounds like a Jetpack issue.

Robertght:
While this is not the same behavior, we have another theme affected by AMP here: https://github.com/Automattic/wp-calypso/issues/56740

Another one: https://github.com/Automattic/themes/issues/4637

Should we gather all related cases in one report or keep them separate? cc @scruffian @jeffikus

marianearchimbaud:
I run into this same issue with a user using Dara theme: 33568199-hc
I asked them to please deactivate the AMP plugin, and the menu started working on mobile devices.

PR #5611, opened 2022-03-02
Remote: match design background color

matiasbenedetto:
Remote: match design background-color

Mockup:
image

Before:
image

After:
image

### DIFF
--- a/remote/theme.json
+++ b/remote/theme.json
@@ -20,7 +20,7 @@
 			"palette": [
 				{
 					"slug": "background",
-					"color": "#222222",
+					"color": "#000000",
 					"name": "Background"
 				},
 				{


madhusudhand (approved):
LGTM.

remote/theme.json
```diff
@@ -20,7 +20,7 @@
 			"palette": [
 				{
 					"slug": "background",
-					"color": "#222222",
+					"color": "#000000",
```
MaggieCabrera, line 23: Looking at the figma files, I see #222 is what's being used but the site does look lighter than the designs. @kjellr can you give us your expert eye?

kjellr: `#000000` is the correct background color here.

PR #5612, opened 2022-03-02
Blockbase: Add responsive fonts

scruffian:
This came up in a chat with @annezazu. It would be good to have responsive font sizes in Blockbase. The problem is that this will change user's sites, so we probably can't go ahead with it, but I thought I'd open it for discussion.

### DIFF
--- a/blockbase/theme.json
+++ b/blockbase/theme.json
@@ -380,17 +380,17 @@
 				},
 				{
 					"name": "Medium",
-					"size": "24px",
+					"size": "min(max(24px, 5vw), 30px)",
 					"slug": "medium"
 				},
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
@kjellr what do you think about this?

kjellr:
In general, I don't think we need responsive font sizes for the small(ish) sizes noted here. I don't really find scaling text down to be that necessary unless text is ~48px or larger. 

I also think it's fine to leave this out for now, since the hope is that [we'll implement responsive font sizes in core directly](http://href.li?https://github.com/WordPress/gutenberg/pull/33543).

PR #5623, opened 2022-03-04
Archeo: home page and header wide aligned

### LINKED ISSUE #5599: Archeo: Make header and footer wide-aligned
Currently, the header and footer are not in alignment with the rest of the default page content. They should both be wide aligned. 

Currently: 

Screen Shot 2022-03-01 at 2 12 24 PM

Screen Shot 2022-03-01 at 2 13 07 PM

MaggieCabrera:

This PR makes all the patterns in the home page and the header and footer wide aligned.

The first pattern is a media + text pattern and that block adds an 8% padding to the text bit that is making it not align with the header and footer. That can't be undone with block markup. 

Screenshot 2022-03-04 at 11 43 14

The demo site looks like this now:

Screenshot 2022-03-04 at 11-59-22 Archeo

#### Related issue(s):

https://github.com/Automattic/themes/issues/5599


kjellr:
Hmm. I'm not sure — this kind of works, but it's still not quite right. In general, what I'm hoping to do is get everything on the edges lined up on the same vertical axes. That's not the case here: 

alignment

Can you think of any other ways to get that to work? I'm probably ok with everything going 100% full-width if it needs to, as long as we can get equal outer margins like that.

kjellr:
Oh also, I don't think this is related to this PR, but full alignments appear broken in the post/page editor: 
Screen Shot 2022-03-07 at 3 18 33 PM

MaggieCabrera:
The problem here is the media + text blocks adding the extra padding we can't remove without CSS. Is it fine with you if we make a utility class to do that, @kjellr ?

Screenshot 2022-03-04 at 11 43 14

mikachan:
I've gone ahead and added a class to remove this padding, as I noticed there was already an existing GB issue for having control over it: https://github.com/WordPress/gutenberg/issues/28556

With this change, all the content on the home page should line up on both sides.

kjellr:
Can we re-create that using a column block instead of Media & Text? That might work, and eliminate the need for a utility class.

MaggieCabrera:
Done, I think this is looking fine now.

Screenshot 2022-03-14 at 12-21-17 Skatepark – Tagline

kjellr:
This is looking much better. 🎉 

The only remaining issue I'm noticing is that this "PALACE OF 
THE CIRCUS..." text bumps up against the screen edges on mobile. Is that something we can fix?

Screen Shot 2022-03-14 at 8 42 18 AM

jffng:
> text bumps up against the screen edges on mobile. Is that something we can fix?

I think so, I added some padding in the pattern: https://github.com/Automattic/themes/pull/5623/commits/4f7211d55d06e0e6123101914551a7f2a933c404

jffng (approved):

jffng:
Ah oops that doesn't work because it messes up the alignment on wider viewports. I pushed another idea here: [20d0a44](https://github.com/Automattic/themes/pull/5623/commits/20d0a44f6a13d3e106d2ccd75a7f98e0b6eca8b6)

The downside to this approach is that the image is no longer flush with the edge on smaller viewports:

Screen Shot 2022-03-14 at 3 23 42 PM

mikachan (approved):
> I pushed another idea here: [20d0a44](https://github.com/Automattic/themes/pull/5623/commits/20d0a44f6a13d3e106d2ccd75a7f98e0b6eca8b6)

This looks good to me, it lines up well with the other elements on both sides when I test on my local site. I think we should prioritize the alignment over the image being flush on smaller screens, but would be great to get @kjellr's sign off too.

kjellr:
> The downside to this approach is that the image is no longer flush with the edge on smaller viewports:

I think that's a reasonable compromise. 👍

PR #5629, opened 2022-03-07
Moved color customizations from custom to theme

### LINKED ISSUE #5448: Blockbase Color customizations show as new "custom palette"

pbking:
This fixes: #5448

Blockbase customizer now stores details in 'theme' space instead of 'custom'.  We now aren't able to determine the "default" color (since previously the 'theme' value was used to determine that) so this option has been removed.

Any colors ADDED by the user via FSE will be lost as this change deletes any color customizations in the 'custom' space.  Otherwise the changes persisted by the customizer wouldn't be realized.

A [previous attempt](https://github.com/Automattic/themes/commit/dded23b133a2602386638e5f9a83a9979cc312c2) at this changed both fonts and colors, however changing fonts doesn't seem to be necessary for appropriate user experience.  That is not attempted in this change, only colors.

To test save colors with the customizer on any Blockbase theme.  Then load the Global Styles panel in FSE and observe the colors.  You should see the customized colors under the Theme heading and no colors under custom:

### DIFF
--- a/blockbase/inc/customizer/wp-customize-colors.php
+++ b/blockbase/inc/customizer/wp-customize-colors.php
@@ -61,16 +61,8 @@ class GlobalStylesColorCustomizer {
 			$user_color_palette = $theme_json['settings']['color']['palette']['custom'];
 		}
 
-		// NOTE: This should be removed once Gutenberg 12.1 lands stably in all environments
-		elseif ( isset( $theme_json['settings']['color']['palette']['user'] ) ) {
-			$user_color_palette = $theme_json['settings']['color']['palette']['user'];
-		}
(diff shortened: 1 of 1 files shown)


matiasbenedetto (approved):
Tested on my local env. Seems to be working great :rocket: 
LGTM

PR #5634, opened 2022-03-08
Archeo: fix image block widths

mikachan:
This adds the `wp-image-` class and ID attribute to any image blocks in Archeo's patterns.

Since https://github.com/WordPress/gutenberg/pull/39045 was merged, only images that have the `wp-image-` class are targeted with the responsive styles. Without this class, the images are not resized to 100%:

| Without class | With class |
| ------------- | ---------- |
| image   | image     |

It looks like images in cover blocks and media + text blocks are not affected.

Do we need to run through all block themes and add these classes?

### DIFF
--- a/archeo/inc/patterns/image-with-description.php
+++ b/archeo/inc/patterns/image-with-description.php
@@ -24,7 +24,7 @@ return array(
 	<!-- /wp:paragraph -->
 
 	<!-- wp:image {"id":840,"sizeSlug":"full","linkDestination":"none"} -->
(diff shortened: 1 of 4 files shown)


MaggieCabrera:
those IDs won't work on all sites if they already have images installed with that ID. Joen said that he was going to make a PR to fix the issue, I think we can patch with CSS instead of changing the markup while the fix makes it to GB

mikachan:
Ah OK, I was wondering about the IDs. Thanks! I'll change this PR to a patch then.

mikachan:
I've added a CSS patch for all Image Block images, using this selector: `.wp-block-image img`

MaggieCabrera (approved):
LGTM

MaggieCabrera:
https://github.com/WordPress/gutenberg/pull/39340 will fix the image issue and when that makes it to dotcom we can remove the CSS

Issue #5595, opened 2022-02-25
Archeo: Submit to dotorg checklist


MaggieCabrera:
The only warnings for Archeo from the theme check plugin are:

```
WARNING: screenshot.jpg is 726.1 KB in size. Large file sizes have a negative impact on website performance and loading time. Compress images before using them.
WARNING: assets/images/chahk.png is 1.4 MB in size. Large file sizes have a negative impact on website performance and loading time. Compress images before using them.
```

- [x] Fix warnings
- [x] Bump version to 1.0.0 https://github.com/Automattic/themes/pull/5668
- [x] Check changelog
- [x] Patch https://github.com/Automattic/themes/pull/5573 
- [x] Ship it! https://themes.trac.wordpress.org/ticket/115279

kjellr:
> WARNING: screenshot.jpg is 726.1 KB in size. Large file sizes have a negative impact on website performance and loading time. Compress images before using them.
> WARNING: assets/images/chahk.png is 1.4 MB in size. Large file sizes have a negative impact on website performance and loading time. Compress images before using them.

The top warning should be addressed in #5598. Unfortunately, `chalk.png` is about as small as it can get while retaining its usage at that large size. So we'll need to just ignore that second warning for now.

MaggieCabrera:
I want to remember that we had trouble in the past when we had too many images or too big ones on dotcom. Maybe we could make the background of `chahk.png` be the same color as the background of the site and make it a jpg?

kjellr:
I considered that (as well as using PNG-8 with a background-color-hinted border), but if folks change the background of the site it looks very poor immediately. That'll be frustrating for folks, and will also be weird if/when we implement alternate Styles. In any case, I'll take one more pass at exporting it next week just in case I missed an option last time around.

kjellr:
Update: I got a little closer towards reducing the file size by refining the cutout edges. I'm going to continue working on it tomorrow. 👍

MaggieCabrera:
Awesome @kjellr ! Just a heads up because I'm building the headstart for skatepark and I encountered this. The limit for theme images on dotcom:

> [ ] Please Fix: https://skateparkdemo.files.wordpress.com/2021/09/13900064573_0b24c944f5_k.jpg is too large 700 kb (greater than 350 kb). Resize the image.

kjellr:
[...]

PR #5635, opened 2022-03-08
Replace Chahk image

kjellr:
Replaces the Chahk image in Archeo with a restrictive GIF version. This trims down the file size by about half, down to ~592kb. 

Addresses part of #5595


pbking (approved):
👍

PR #5637, opened 2022-03-08
Blockbase: avoid margin-top on the first paragraph child of a group

### LINKED ISSUE #5617: Blockbase: Unwanted top margin for paragraph
## Steps to replicate

* Add a Group block
* Add a Paragraph block inside the Group block
* Add paddings to the Group block
* See it on the front-end

For example:
```
<!-- wp:group {"style":{"spacing":{"padding":{"top":"8px","right":"8px","bottom":"8px","left":"8px"}}},"backgroundColor":"primary","textColor":"background"} -->
<div class="wp-block-group has-background-color has-primary-background-color has-text-color has-background" style="padding-top:8px;padding-right:8px;padding-bottom:8px;padding-left:8px"><!-- wp:paragraph -->
<!-- /wp:paragraph --></div>
<!-- /wp:group -->
```
## Result

There is an unwanted top margin for the Paragraph block on the front-end. There is no margin in the editor as expected.
This is similar to #5616. The CSS for a Quote block is messing this up.

**Front-end**
Screenshot 2022-03-03 at 19 30 10

**Editor**
Screenshot 2022-03-03 at 19 30 21

## Expected

No margin for the paragraph block.

matiasbenedetto:
Blockbase: add an exception in a too general CSS selector to avoid the unwanted margin described on the issue.

#### Code used:
```
<!-- wp:group {"style":{"spacing":{"padding":{"top":"25px","right":"25px","bottom":"25px","left":"25px"}}},"borderColor":"luminous-vivid-orange"} -->
<div class="wp-block-group has-border-color has-luminous-vivid-orange-border-color" style="padding-top:25px;padding-right:25px;padding-bottom:25px;padding-left:25px"><!-- wp:paragraph -->
<p>Hello, I'm a paragraph inside a group block.</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>Hello, I'm a paragraph inside a group block.</p>
<!-- /wp:paragraph -->

<!-- wp:group -->
<div class="wp-block-group"><!-- wp:paragraph -->
<p>Hello, I'm a paragraph inside a group, inside a group.</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group --></div>
<!-- /wp:group -->
```

#### Screenshots of Blockbase with this fix:

image
image

#### Screenshots of Blockbase without this fix:

image

#### Related issue(s):
https://github.com/Automattic/themes/issues/5617

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -514,7 +514,7 @@ input[type=checkbox] + label {
   flex-wrap: nowrap;
 }
 .wp-block-navigation.blockbase-responsive-navigation-minimal.is-responsive .is-menu-open.wp-block-navigation__responsive-container .wp-block-navigation__responsive-container-content .wp-block-navigation-item {
-  row-gap: 0rem;
+  row-gap: 0;
 }
 .wp-block-navigation.blockbase-responsive-navigation-minimal.is-responsive .is-menu-open.wp-block-navigation__responsive-container .wp-block-navigation__responsive-container-content .wp-block-navigation-item > a:hover {
   -webkit-text-decoration-line: underline;
@@ -593,7 +593,7 @@ input[type=checkbox] + label {
   flex-direction: column;
 }
 
-p.has-drop-cap:not(:focus):first-letter {
+p.has-drop-cap:not(:focus)::first-letter {
   font-size: var(--wp--custom--paragraph--dropcap--typography--font-size);
   font-weight: var(--wp--custom--paragraph--dropcap--typography--font-weight);
   margin: var(--wp--custom--paragraph--dropcap--margin);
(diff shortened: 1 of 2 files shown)


mikachan (approved):
LGTM!

MaggieCabrera:
Let's revisit this, after @iamtakashi [comment](https://github.com/Automattic/themes/issues/5617#issuecomment-1062865645) and looking at the [original problem](https://github.com/Automattic/themes/pull/5104) that this is trying to solve, I think the styles are too general. Ideally, we would target just flex groups (rows) but I don't think there's a way to do that. Can we fix this in the editor instead?

iamtakashi:
Hi, I've noticed this was closed. I assume this issue has been looked into at a different place therefore the closure.

Could you point me where that is? Thanks in advance!

matiasbenedetto:
@iamtakashi I closed this PR because I think the solution implemented in this PR is not good enough to solve the problem: It doesn't take care of other potential combinations of elements that can cause [the original issue](https://github.com/Automattic/themes/issues/5617). We would need to think of a better solution for this.

iamtakashi:
@matiasbenedetto That makes sense. Thanks for clarification.

PR #5640, opened 2022-03-08
Archeo: fix footer spacing at mobile

mikachan:
#### Changes proposed in this Pull Request:
This tidies up the footer spacing and alignment, following the suggestions from https://github.com/Automattic/themes/issues/5631.

Before|After
---|----
image|image

#### Related issue(s):
https://github.com/Automattic/themes/issues/5631

Issue #5631, opened 2022-03-07
[Archeo] Small screen fixes


kjellr:
Archeo has a few areas of its mobile templates that need adjustment: 

---

- [x] `home.html` top area https://github.com/Automattic/themes/pull/5645: 

The two areas indicated in pink should use the equivalent of 48px of padding.  

<img width="372" alt="Screen Shot 2022-03-07 at 3 32 36 PM" src="https://user-images.githubusercontent.com/1202812/157113858-56fdfbfc-1404-445b-a632-31dda07193d7.png">

---

- [x] Post lists: https://github.com/Automattic/themes/pull/5638

The spacing above each author name should be moved below it, so that the authors appear associated with the correct post titles on small screens: 

Before|After
---|---
<img width="371" alt="Screen Shot 2022-03-07 at 3 38 27 PM" src="https://user-images.githubusercontent.com/1202812/157114168-891bef37-72f7-4bee-a34f-ef369612edf6.png">|<img width="371" alt="Screen Shot 2022-03-07 at 3 38 19 PM" src="https://user-images.githubusercontent.com/1202812/157114169-a7305e43-2356-4f0a-bf5d-34061f146eb8.png">

---

- [x] Footer menu https://github.com/Automattic/themes/pull/5640:

The spacing and alignment of this footer menu looks like a bug. Can this be left aligned? Can we remove the vertical `blockGap`? 

<img width="374" alt="Screen Shot 2022-03-07 at 3 39 57 PM" src="https://user-images.githubusercontent.com/1202812/157114353-edc779d6-1019-4d90-82b5-2c19996524be.png">

---

- [x] Mobile menu (expanded) needs a little work too. Would it be possible to use the `large` preset size? And to decrease the `blockGap` value? https://github.com/Automattic/themes/pull/5644

Before|After
---|---
<img width="371" alt="Screen Shot 2022-03-07 at 3 42 09 PM" src="https://user-images.githubusercontent.com/1202812/157115076-dcfea691-40a1-4d9f-b5b8-76fd6d99d574.png">|<img width="371" alt="Screen Shot 2022-03-07 at 3 44 59 PM" src="https://user-images.githubusercontent.com/1202812/157115082-de8c69a9-3cef-48ef-a77e-f3654b9bde92.png">

kjellr:
I'm re-opening this one because that second one appears to be broken again: 

Current|Recommended
---|---
[...]

PR #5644, opened 2022-03-09
Archeo: add CSS for mobile menu font size

mikachan:
This increases the font size of the main navigation on small screens, as suggested in https://github.com/Automattic/themes/issues/5631:

Before|After
----|-----
image| image

The `blockGap` has already been reduced to `var(--wp--custom--spacing--small)` in https://github.com/Automattic/themes/pull/5640. This is what I've used in the above screenshot. Should this be reduced any further?

#### Related issue(s):
https://github.com/Automattic/themes/issues/5631

### DIFF
--- a/archeo/style.css
+++ b/archeo/style.css
@@ -224,3 +224,13 @@ body > .is-root-container > .wp-block-template-part > .wp-block-cover,
 	height: auto;
 	max-width: 100%;
 }
+
+/*
+ * Custom font size at small screens for navigation.
+ * Possible future fix: https://github.com/WordPress/gutenberg/pull/33543
+ */
+@media (max-width: 599px) {
+	.wp-block-navigation__responsive-container-content li {
+		font-size: var(--wp--preset--font-size--large) !important;
+	}
+}
--- a/archeo/style.css
+++ b/archeo/style.css
@@ -224,13 +224,3 @@ body > .is-root-container > .wp-block-template-part > .wp-block-cover,
 	height: auto;
 	max-width: 100%;
 }
-
-/*
- * Custom font size at small screens for navigation.
- * Possible future fix: https://github.com/WordPress/gutenberg/pull/33543
- */
-@media (max-width: 599px) {
-	.wp-block-navigation__responsive-container-content li {
-		font-size: var(--wp--preset--font-size--large) !important;
-	}
-}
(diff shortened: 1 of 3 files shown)


archeo/style.css
```diff
@@ -224,3 +224,13 @@ body > .is-root-container > .wp-block-template-part > .wp-block-cover,
 	height: auto;
 	max-width: 100%;
 }
+
+/*
+ * Custom font size at small screens for navigation.
+ * Possible future fix: https://github.com/WordPress/gutenberg/pull/33543
+ */
+@media (max-width: 599px) {
+	.wp-block-navigation__responsive-container-content li {
```
jffng: I noticed this selector doesn't work in the editor for non-standard links. For example, I have Submenu and a Home link, and the large font size isn't applying: 
Screen Shot 2022-03-09 at 2 12 32 PM

It makes me wonder if adding this CSS is worth it, or if it's going to cause more confusion if a user changes this navigation's font size in the editor and previews it. 

Another thought — could we create a custom CSS variable that uses a responsive type size and apply it to the navigation block [here](https://github.com/Automattic/themes/blob/trunk/archeo/theme.json#L140). That way any user modifications to the font-size would still take precedence.

pbking: > could we create a custom CSS variable that uses a responsive type size and apply it to the navigation block [here](https://github.com/Automattic/themes/blob/trunk/archeo/theme.json#L140).

Oh that's a fine idea.

kjellr:
If this doesn't already exist, we should file an issue to ask for the ability to specify an overlay-specific font size. I imagine this will be a common use case, and we already have the ability to set overlay-specific values for color via `overlayBackgroundColor` and `"overlayTextColor`.

pbking:
I don't see the responsive menu "small" as posted in the before shots, I looks correct as of [this change](https://github.com/Automattic/themes/pull/5462).  The only change I can see from this PR is that the font size configuration has been moved from the BLOCK to THEME.JSON which looks good to me.

archeo/style.css
mikachan: I think the CSS variable with a responsive type size is a great idea, however, I'm not sure how to do this where the font size is bigger on smaller screens and smaller on larger screens.

For now, I've gone with setting the larger font size for the overlay in CSS (we were already doing this, I missed it when I was working on this PR!) and setting the default font size in theme.json.

mikachan:
> The only change I can see from this PR is that the font size configuration has been moved from the BLOCK to THEME.JSON

Yeah this sounds correct, I missed that we'd already set the large font size for the responsive menu via CSS, but then it looks like it was being overwritten by the font size set in the header markup. So, I've removed the font size from the `header.html` and moved it to theme.json.

Going to look for / create an issue for setting an `overlayFontSize` (or similar) setting in GB now..

pbking:
I couldn't find a related issue so have opened this: https://github.com/WordPress/gutenberg/issues/39352

The change looks good, 🚢

pbking (approved):

jffng:
> Going to look for / create an issue for setting an overlayFontSize (or similar) setting in GB now..

Looks like some folks are starting to think about this holistically here: https://github.com/WordPress/gutenberg/issues/39142

PR #5645, opened 2022-03-09
Archeo: remove bottom margin from columns

mikachan:
This removes the bottom margin from the columns in the 'Image with headline and description' pattern. This allows the space at mobile between this pattern and the list of posts to be `48px`, as suggested in https://github.com/Automattic/themes/issues/5631.

Before|After
----|----
image | image

#### Related issue(s):
https://github.com/Automattic/themes/issues/5631

### DIFF
--- a/archeo/inc/patterns/image-with-headline-description.php
+++ b/archeo/inc/patterns/image-with-headline-description.php
@@ -7,8 +7,8 @@ return array(
 	'categories' => array( 'featured', 'images' ),
 	'content'    => '<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"0","right":"5vw","bottom":"0","left":"5vw"}}}} -->
 	<div class="wp-block-group alignfull" style="padding-top:0;padding-right:5vw;padding-bottom:0;padding-left:5vw"><!-- wp:media-text {"mediaLink":"' . esc_url( get_template_directory_uri() ) . '/assets/images/chahk.gif","mediaType":"image","mediaWidth":64,"verticalAlignment":"top"} -->
-	<div class="wp-block-media-text alignwide is-stacked-on-mobile is-vertically-aligned-top" style="grid-template-columns:64% auto"><figure class="wp-block-media-text__media"><img src="' . esc_url( get_template_directory_uri() ) . '/assets/images/chahk.gif" alt="' . esc_attr__( 'Chahk: rain deity', 'archeo' ) . '"/></figure><div class="wp-block-media-text__content"><!-- wp:heading {"textAlign":"right","style":{"typography":{"fontSize":"clamp(64px, 6vw, 100px)","lineHeight":"1","textTransform":"uppercase"},"spacing":{"margin":{"bottom":"48px"}}}} -->
-	<h2 class="has-text-align-right" id="chahk-raindeity" style="font-size:clamp(64px, 6vw, 100px);line-height:1;margin-bottom:48px;text-transform:uppercase">' . wp_kses_post( __( 'Chahk:<br>Rain<br>deity', 'archeo' ) ) . '</h2>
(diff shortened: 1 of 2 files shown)


archeo/inc/patterns/image-with-headline-description.php
```diff
@@ -11,8 +11,8 @@
 	<h2 class="has-text-align-right" id="chahk-raindeity" style="font-size:clamp(64px, 6vw, 100px);line-height:1;margin-bottom:48px;text-transform:uppercase">' . wp_kses_post( __( 'Chahk:<br>Rain<br>deity', 'archeo' ) ) . '</h2>
 	<!-- /wp:heading -->
 
-	<!-- wp:columns -->
-	<div class="wp-block-columns"><!-- wp:column {"width":"15%"} -->
+	<!-- wp:columns {"style":{"spacing":{"margin":{"bottom":"0px"}}}} -->
```
jffng: I am not sure this quite gets it there. I think the spacing is now the paragraph's default bottom margin, which is 1em: 

Screen Shot 2022-03-09 at 2 40 35 PM

We'd either have to overwrite the paragraph's bottom margin there, or add some arbitrary margin value to the columns to get the sum to be `48px`.

kjellr: I pushed a couple small changes that I think get this closer. I adjusted the value of the `medium` custom spacing value, but it didn't appear to have any negative effects elsewhere — I like the extra space it adds. 

I also messed with the bottom margin on the homepage main title there... it's not 100% how I'd like it, but it's pretty close.

mikachan: > I think the spacing is now the paragraph's default bottom margin, which is 1em

I was adding up the spacing above the 'Selected Writings' heading and the bottom of this paragraph to `48px`. I must've misunderstood!

> I adjusted the value of the medium custom spacing value

I think this looks good and doesn't seem to negatively affect anything else. Thanks!

archeo/inc/patterns/image-with-headline-description.php
```diff
@@ -7,8 +7,8 @@
...
-	<h2 class="has-text-align-right" id="chahk-raindeity" style="font-size:clamp(64px, 6vw, 100px);line-height:1;margin-bottom:48px;text-transform:uppercase">' . wp_kses_post( __( 'Chahk:<br>Rain<br>deity', 'archeo' ) ) . '</h2>
+	<div class="wp-block-media-text alignwide is-stacked-on-mobile is-vertically-aligned-top" style="grid-template-columns:64% auto"><figure class="wp-block-media-text__media"><img src="' . esc_url( get_template_directory_uri() ) . '/assets/images/chahk.gif" alt="' . esc_attr__( 'Chahk: rain deity', 'archeo' ) . '"/></figure><div class="wp-block-media-text__content"><!-- wp:heading {"textAlign":"right","style":{"typography":{"fontSize":"clamp(64px, 6vw, 100px)","lineHeight":"1","textTransform":"uppercase"},"spacing":{"margin":{"bottom":"clamp(5px, 2vw, 20px)"}}}} -->
+	<h2 class="has-text-align-right" id="chahk-raindeity" style="font-size:clamp(64px, 6vw, 100px);line-height:1;margin-bottom:clamp(10px, 2vw, 20px);text-transform:uppercase">' . wp_kses_post( __( 'Chahk:<br>Rain<br>deity', 'archeo' ) ) . '</h2>
```
jffng: The margin-bottom clamp values supplied for the attribute and the style don't match, so I'm getting a block validation error. 

Should the lower value be `5px` or `10px`? cc @kjellr

kjellr: Whoops — 5px should be fine. Thanks.

jffng (approved):

PR #5648, opened 2022-03-10
Archeo: remove responsive image fix

mikachan:
This removes the CSS patch for responsive images in patterns, as this has been fixed in GB here: https://github.com/WordPress/gutenberg/pull/39340 🎉

#### Related issue(s):
https://github.com/Automattic/themes/pull/5634


scruffian:
Great! I think this fix is in a bunch of themes so we should remove it from all of them...

pbking:
LGTM.
Approving and moving to the "not yet" column.

pbking (approved):

scruffian:
> Great! I think this fix is in a bunch of themes so we should remove it from all of them...

I checked and it's not in any block themes, so I think we're good.

Issue #5585, opened 2022-02-23
Multiple Themes: Tertiary color backgrounds are not visible on the site.


devNigel:
### Quick summary

On theme Stow, when Tertiary color is set as the background color, the background color is not working. I was able to reproduce this on Simple and AT sites.

I shared a workaround and let the user know that we will update them once it is fixed.

### Steps to reproduce

1. Activate Stow theme. 
2. On the page editor, add a paragraph block.
3. Apply the same color set as Tertiary color in Customize > Colors > Tertiary color as the background color of the block.
4. Preview it.
5. You will see the background color is something else.

<img width="1447" alt="image" src="https://user-images.githubusercontent.com/33957570/155267329-ad37fd33-b936-40bc-a9a9-7f9dccf8ecb4.png">

<img width="889" alt="image" src="https://user-images.githubusercontent.com/33957570/155267355-1f9c6ee3-0a4d-48dc-b663-801d6806140f.png">

### What you expected to happen

The background color seen in editor should be visible on live page as well.the 

### What actually happened

The background color on the live page was different.

### Context

User report: 4803117-zen

### Simple, Atomic or both?

Simple, Atomic

### Theme-specific issue?

Stow

I could not find other themes with tertiary color setting.

### Reproducibility

Consistent

### Severity

All

### Available workarounds?

Yes, easy to implement

### Workaround details

Use a slightly different color code that looks similar to Tertiary color.

jordesign:
I've found this to be the case on Hever as well. but not for Varia itself - so it seems specific to some child themes.
One thing I noted is that Varia does not include a Tertiary color itself - but the child themes do.

When the color is applied - a class is added to the element:

* `has-tertiary-background-color has-background`, or 
* `has-tertiary-color has-text-color`

But then no color declaration comes in the CSS.
Maybe related to this PR: https://github.com/Automattic/themes/pull/3223

Ineraction: #4803951-zen

Nic-Sevic:
another instance here: 4805667-zd-woothemes
this time with the columns block

AMcGrady:
Another instance here: 4809339-zd

Paragraph block background color not working on the Hever theme.

AMcGrady:
Reported here: 4809827-zd
Blank Canvas theme

AMcGrady:
Possibly related:
https://github.com/Automattic/themes/issues/4639
?

poeticabdul:
Reported here 33966058-hc for Rockfield theme

[...]

PR #5656, opened 2022-03-11
Stow: add tertiary color and background styling

madhusudhand:

* theme has all css spec for all the four background variants but not for tertiary
* add styles for tertiary text

#### screenshots

image

#### Related issue(s):

#5585


stow/style-editor.css
```diff
@@ -1182,6 +1182,17 @@ pre.wp-block-verse {
...
+.has-tertiary-background-color.has-background-dim {
+	background-color: var(--wp--preset--color--tertiary);
+	color: var(--wp--preset--color--background);
+}
+
+.has-tertiary-background-color:not(.has-text-color),
+.has-tertiary-background-color.has-background-dim:not(.has-text-color) {
+	color: var(--wp--preset--color--foreground);
+}
+
```
MaggieCabrera: Stow is a Varia child theme, we make the css changes in the sass files and build using `npm run start` or `npm run build`

MaggieCabrera: in this case I think this should be changed directly in varia, since the child theme is importing the partial from the parent, but it's weird that we need this only for Stow then, if none of the other Varia themes are using it.

MaggieCabrera: To help you test, you can use the markup on this post: pcrugM-jc-p2l give you access to that demo site

madhusudhand: Varia doesn't have a tertiary definition, so the missing styles in child themes.
What shall be the spec for the colors if we need to introduce `tertiary`. I couldn't find in figma.

image

@kjellr can you help here.

MaggieCabrera: this would be the tertiary color for stow: https://github.com/Automattic/themes/blob/15bb10aeb42cd3aa477ce755fe35048f7cdbbfce/stow/functions.php#L19 I think we are using different naming conventions on Varia where `tertiary` == `border` https://github.com/Automattic/themes/blob/15bb10aeb42cd3aa477ce755fe35048f7cdbbfce/stow/sass/_global-variables.scss#L14

MaggieCabrera: We probably need to add the new CSS rules here: https://github.com/Automattic/themes/blob/15bb10aeb42cd3aa477ce755fe35048f7cdbbfce/varia/sass/blocks/utilities/_style.scss#L117 and make sure that the name of the class is the same as the one the editor is outputting. It's very strange that this is happening now (Varia themes have been around for a few years and this feels like a big oversight). I'm thinking this bug might have been created by a refactor we did to move from sass variables to CSS variables for some of the varia children and something got lost in the process. This is the PR that did that for Stow: https://github.com/Automattic/themes/pull/3241

kjellr: Be sure to test any changes here with the Customizer's ability to change colors. The color annotations in Varia + its children are very complicated and prone to breaking.

madhusudhand: Updated Varia theme here: https://github.com/Automattic/themes/pull/5666

madhusudhand:
Following child themes of Varia are effected by this issue. 
* hever
* rockfield
* stow
* stratford

Updated styles conditionally in Varia theme (PR: #5666) and included color overrides for tertiary for the above four themes.
All other changes are because of the parent theme change.

Verified the changes in Varia and all child themes, they seems to be working as expected.

Hever:
image

Rockfield:
image

Stow:
image

Stratford:
image

MaggieCabrera:
Ok, I see that you've included the changes on the children on this separate PR. It's cleaner in terms of separating the amount of changes but it makes it harder to test, since you need both PRs at the same time.

alves/style-editor.css
```diff
@@ -647,7 +647,7 @@ object {
 
 .wp-block-cover h2,
 .wp-block-cover-image h2 {
-	font-size: 2.592em;
+	font-size: 2.592;
```
MaggieCabrera, line 650: I see where this is coming from but we should double check that it doesn't break existing websites even though it's the correct syntax.

hever/sass/_config-child-theme-deep.scss
```diff
@@ -18,25 +18,25 @@ $typescale-ratio: 1.15; // Run ratio math on 1em == $typescale-base * $typescale
 
 $config-global: (
 
-	/* Fonts */
+
```
MaggieCabrera, line 21: We shouldn't remove the comments from sass files

madhusudhand: This seems to be side effect of lint fixes. reverted the change

alves/style-editor.css
madhusudhand: Verified the changes against classes like `.has-larger-font-size` `.has-huge-font-size` and headers in Varia and one of child themes and it seems to be working fine.

madhusudhand:
Moved the changes to #5666 and closing this PR.

PR #5660, opened 2022-03-14
Try: Blockbase standalone theme

MaggieCabrera:

Following on our conversations about what kind of theme could be used as a starter for other teams that want to build Block themes but want a solid foundation instead of doing it from scratch, I quickly built this version of Blockbase that only has the basics as a standalone theme. 

**Who is this theme for?**

Someone who doesn't want to start from 0 and wants the basics in terms of visual structure to tweak and move without a massive boilerplate and relying 100% on FSE. Most probably someone who isn't 100% familiar with all that theme.json and FSE can do and doesn't want to miss something along the way.

**How should one use this theme?**

I think this theme is best used if you fork it and build on top of it. I don't think it makes sense to use it as a parent theme, it adds no functionality or a heavy stylistic foundation that you'd want to keep.

This theme includes:

- No customizer stuff or any other "Universal" functionality.
- No CSS ponyfills besides the alignments styles and I think I added the search + file blocks fix.
- No block patterns besides the 404 one for i18n purposes. I added it to the patterns folder and tried to implement it using https://github.com/WordPress/gutenberg/pull/36751 but looks like it's not working with the block pattern block! we can wait that one out or implement the pattern the old way as we see fit.
- Same design as Blockbase (typography, spacing, colors, block settings...)
- Same templates as Blockbase.

Please do forgive the horrible naming. It feels like a running joke already with the starter themes.

It looks like this:

Screenshot 2022-03-14 at 18-32-23 Skatepark – Tagline


pbking:
I think it's a great idea, it seems like a good starting place for a Block Theme.

~~I think that we should consider adding more things from ponyfill.  Doing so may make this a decent parent theme allowing for those ponyfills to ultimately go away in one place instead of all the themes that might have cloned it. Things like navigation styling and button hover states that still needs a ponyfill of some kind to work; standardizing the "Blockbase" way seems a good way to do that.~~

Edit: Yea I get carried away thinking about child themes sometimes.  I agree with comments below, this is a better tool to be used instead of "empty theme" to get started with rather than a parent theme to derive from.

block-canvas/parts/footer.html
```diff
@@ -0,0 +1,11 @@
+<!-- wp:group {"layout":{"inherit":true}} -->
```
scruffian, line 1: It would be great if these came from the pattern directory

block-canvas/theme.json
```diff
@@ -0,0 +1,418 @@
...
+			}
+		},
+		"color": {
+			"background": "var(--wp--preset--color--background)",
+			"text": "var(--wp--preset--color--foreground)"
+		},
+		"elements": {
+			"h1": {
+				"typography": {
+					"fontFamily": "var(--wp--preset--font-family--heading-font)",
```
scruffian, line 319: I think all the shared headings definitions can live under a headings property.

block-canvas/functions.php
```diff
@@ -0,0 +1,61 @@
...
+	 * Sets up theme defaults and registers support for various WordPress features.
+	 *
+	 * @since Block Canvas 1.0
+	 *
+	 * @return void
+	 */
+	function block_canvas_support() {
+
+		// Add support for block styles.
+		add_theme_support( 'wp-block-styles' );
```
scruffian: I'm not sure we want this

MaggieCabrera:
> I think that we should consider adding more things from ponyfill. Doing so may make this a decent parent theme allowing for those ponyfills to ultimately go away in one place instead of all the themes that might have cloned it. Things like navigation styling and button hover states that still needs a ponyfill of some kind to work; standardizing the "Blockbase" way seems a good way to do that.

I think then why not use Blockbase then? I didn't want to add any of the extras because I wanted this to be as close to a clean slate as possible. CSS in any case is bringing a maintenance burden and this one is not meant to be a parent theme that we will update and other themers can just build children and always get the latest code, but as a starting point for someone to build something new. It's meant to bring a base of what a block theme can include, nothing more (without looking terrible like emptytheme).

In fact, now I think I just need to remove everything that's not CSS alignments! And I'm keeping those because else it looks really bad.

block-canvas/style.css
```diff
@@ -0,0 +1,162 @@
...
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU General Public License for more details.
+
+*/
+
+/*
+Theme Name: Archeo
```
mikachan: I think this section has appeared by mistake.

mikachan (commented):
This is a great idea; I love that it's another option for themers to use as a starting point, and it's a great introduction to block themes.

I think we should add a readme file similar to the one we have for Blockbase, to help people understand when/why they could use either theme.

I also think the name is great 😁

block-canvas/theme.json
```diff
@@ -0,0 +1,418 @@
...
+					"fontSlug": "system-font",
+					"slug": "body-font",
+					"name": "Body (System Font)"
+				},
+				{
+					"fontFamily": "-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif",
+					"fontSlug": "system-font",
+					"slug": "heading-font",
+					"name": "Headings (System Font)"
+				}
```
pbking: I don't believe we need to use the 'fontSlug' attribute here.  I think that's only necessary for the Blockbase font customizations tooling.

It could be super to include a [webfont](https://github.com/WordPress/gutenberg/pull/37140) here as an example, though I also like that it's styled just like stock Blockbase which doesn't need one.

MaggieCabrera:
I'm trying to make this theme work with https://github.com/WordPress/gutenberg/pull/36751 for the 404 pattern but I'm having no luck, can any of you 👀 to see if the pattern works with the GB PR? maybe it's my env?

block-canvas/theme.json
MaggieCabrera: I thought about it but wanted to stay with the blockbase design, I'll remove de fontSlug

block-canvas/theme.json
MaggieCabrera: looks like we have on remote under styles/elements/headings but it doesn't seem to be working! do you remember how to do that?

block-canvas/parts/footer.html
MaggieCabrera: we can refactor the theme once we push them to the directory, but I agree

MaggieCabrera:
I think I have done all the suggested changes, should we bring this in to use as a starter with the create plugin ourselves? I think development of Pendant will inform any adjustments that we might want to make to this theme and we can iterate as we go.

I'd add it to gitignore on dotcom if we merge, though!

mikachan:
> should we bring this in to use as a starter with the create plugin ourselves? 

Sounds good to me!

mikachan (approved):

PR #5664, opened 2022-03-15
Remote: reassigned block patterns to default categories

MaggieCabrera:

Instead of using the pages category (blame copy paste when setting up the theme for that) I've assigned all patters to default categories for now. @kjellr @beafialho feel free to correct any that you think it's not the right one.

### DIFF
--- a/remote/inc/block-patterns.php
+++ b/remote/inc/block-patterns.php
@@ -14,6 +14,36 @@
  */
 function remote_register_block_patterns() {
 
+	$block_pattern_categories = array(
+		'featured' => array( 'label' => __( 'Featured', 'remote' ) ),
+		'columns'  => array( 'label' => __( 'Columns', 'remote' ) ),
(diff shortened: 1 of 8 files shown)


kjellr:
This seems fine! I _think_ we still need to define those categories just in case the external call to the pattern directory doesn't work? Or at least, that's what we did in Twenty Twenty-Two. 

In addition to their more specific categories, we should also add these categories to the "featured" category too.

MaggieCabrera:
We need to re register the categories so the theme patterns show up first, but they work fine without doing so. I did it anyway, since it's what we did on archeo too. I wish there was a simpler way to reorder patterns though, feels like a lot of PHP just for that!

I added all the patterns to the featured category too.

MaggieCabrera:
I opened a GB issue about this: https://github.com/WordPress/gutenberg/issues/39500

kjellr (approved):
Good thought about the Gutenberg issue! In the meantime, this looks good to go. 🚢

PR #5665, opened 2022-03-15
Remote: Image with text pattern

MaggieCabrera:

I didn't know what the proper name for this pattern is so I improvised. @kjellr we still have the problem that these patterns won't work as designed on a post or page, since alignments don't work with the offset column and they will only show properly aligned if used within a template.

Screenshot 2022-03-15 at 11 41 31

#### Related issue(s):


MaggieCabrera:
I couldn't find credits for the image on the figma file, @kjellr if you can provide them I'll add them to the readme file in this PR too

kjellr:
> we still have the problem that these patterns won't work as designed on a post or page, since alignments don't work with the offset column and they will only show properly aligned if used within a template.

I think that's generally ok. But maybe we remove that empty column on the left for this one? It'll appear little nicer in-post with that gone. 

As for the image credit: 

**Laptop Photo from Piqsels, CC0**
https://www.piqsels.com/en/public-domain-photo-svvhe

And in case we need this one later on: 

**Desk photo from Rawpixel, CC0**
https://www.rawpixel.com/image/5966102/macbook-modern-desk

MaggieCabrera:
Done! ready for ✅

kjellr (approved):
LGTM! Thanks.

remote/inc/patterns/image-with-text.php
```diff
@@ -0,0 +1,25 @@
+<?php
+/**
+ * Image with text
+ */
+return array(
+	'title'      => __( 'Image with text', 'remote' ),
+	'categories' => array( 'text', 'columns' ),
+	'content'    => '<!-- wp:image {"align":"wide","sizeSlug":"large","linkDestination":"none","className":"is-style-skatepark-aside-caption"} -->
```
jffng, line 8: Is the skatepark classname needed here?

MaggieCabrera: ugh, how copy paste betrays me, how did I manage that? I'll fix it in a follow-up, thanks for catching that!

PR #5666, opened 2022-03-15
Varia: add tertiary styles and update SASS variables to CSS variables

### LINKED ISSUE #5585: Multiple Themes: Tertiary color backgrounds are not visible on the site.

madhusudhand:

Varia is a parent theme and the tertiary styles are not defined.
It effects all of its child theme tertiary styles of block background and text colors.

* created styles for tertiary blocks
* updated sass variables to CSS variables to inherit changes in child-themes.

#### Screenshots

Verified the Varia and one of the child theme Stow and it is working as expected.

**Varia:**

image

**Stow:**

image

#### Related PRs:

#5656 

#### Related issue(s):

Fixes:  #5585


varia/sass/abstracts/_config-global.scss
```diff
@@ -50,59 +52,23 @@ $config-global: (
...
-		"primary": (
-			"default": blue,
-			"hover": indigo,
-		),
-		"secondary": (
-			"default": red,
-			"hover": darkred,
-		),
-		"foreground": (
-			"default": #444444,
```
MaggieCabrera: I think this will break other child themes, not all of them are using CSS variables. A good example of one Varia child that doesn't is Maywood.

MaggieCabrera: To test how this change affects all the varia children you should rebuild all of their CSS, running `sh rebuild-child-themes.sh` from inside the varia theme folder.

madhusudhand: Right. Many child themes are breaking. I will go ahead and convert to CSS variables.
How do you suggest PRs. 
* include everything here
* create a single PR for all child themes
* or individual PR for each child theme

MaggieCabrera: I believe there's a reason why we didn't use CSS variables for all child themes but I can't remember why! I think both methods should live in varia and each child is using one or the other.

varia/sass/blocks/utilities/_editor.scss
```diff
@@ -25,6 +26,11 @@
 	color: #{map-deep-get($config-global, "color", "secondary", "default")};
 }
 
+.wp-block .has-tertiary-color,
+.has-tertiary-color {
+	color: #{map-deep-get($config-global, "color", "tertiary", "default")};
```
madhusudhand: @MaggieCabrera For this rule to get compiled tertiary color must be defined under `global-config` even though the theme doesn't have a tertiary option.

Child themes like Rivinton doesn't have a tertiary either in the palette. 
By using CSS variables, Varia theme can give it a css-variable and inherit it in child theme without worrying about a color and child themes may or may not define these CSS variables.
However, we need to introduce CSS variables in all the child themes.
```
"tertiary": (
  "default": var(--wp--preset--color--tertiary),
  "hover": var(--wp--preset--color--tertiary-hover),
),
```

If we have to maintain both, Varia must define tertiary and give it a value `none` and child themes need to set a value or none.

MaggieCabrera:
I've been looking more into this issue and this is what I've found:

- For context: Varia themes implement the plugin "Color annotations" in dotcom (in your local install it works with a different UI, but it should work similarly I believe). Check for it under the customizer/colors. The original implementation was buggy and very messy (check barnsbury, on inc/wpcom-colors.php) and it wasn't done for all the child themes. We then decided that the themes that didn't have color annotations yet would implement them differently, hence the CSS variables are only working for those.
- Some of the themes that don't have the CSS implementation (ie: Exford) have more than 4 colors, but none are named tertiary. 
- I think the tertiary naming came at some point with the CSS variable implementation and that's why there are no classes for them in Varia. It must have been an oversight at some point.

I think to solve this we must audit exactly which themes are having this issue besides Stow, then add a class for them so it fixes the issue for users who have already this class applied somewhere in their sites. I'd prefer to rename the colors to fit the naming convention of the rest of the child themes but that means the users with this existing problem won't get a solution.

If a theme doesn't have a tertiary color, we are not adding one. If the class exists and the background color variable is not returning a color then that class is simply redundant and never used for that theme, so we don't worry about it.

The themes with tertiary color, which I'm pretty sure are all the themes that have the CSS variable implementation, are: hever, morden, rockfield, shawburn, stow, stratford. Check their functions.php file to see how the palettes are setup for the editor with the `varia_default_colors` function

varia/sass/blocks/blog-posts/_style.scss
```diff
@@ -141,16 +148,16 @@
...
-						color: currentColor;
+						color: currentcolor;
 					}
 				}
 			}
 		}
 	}
 }
 
-@import "../../base/extends";
```
madhusudhand, line 152: Removed `extends` import, because it creates duplicate styles.
It is already included here.
https://github.com/Automattic/themes/blob/18482fab16e744f723227652524bafa0ca8d2513/varia/sass/base/_imports.scss#L6-L8

varia/style.scss
```diff
@@ -28,11 +28,6 @@ Nicolas Gallagher and Jonathan Neal https://necolas.github.io/normalize.css/
  */
 @import "sass/abstracts/imports";
 
-/**
- * Extends
- */
-@import "sass/base/extends";
-
```
madhusudhand, line 35: Another duplicate entry. 
extends is already part of next line `@import "sass/base/imports";`

varia/sass/blocks/blog-posts/_style.scss
MaggieCabrera: great catch!

varia/sass/blocks/blog-posts/_style.scss
```diff
@@ -141,16 +148,16 @@
 					.has-background:not(.has-background-background-color) &,
 					[class*="background-color"]:not(.has-background-background-color) &,
 					[style*="background-color"] & {
-						color: currentColor;
+						color: currentcolor;
```
MaggieCabrera: I think this is being changed by the linter and I'm afraid this will break the rule in certain browsers since I believe the camelcase is correct for the keyword.

MaggieCabrera:
I rebuilt the css for all the children to check how they are affected by the varia changes and everything worked as expected. I'm mostly seeing the linting changes on the themes that don't define a tertiary color (though the `currentColor` issue I mentioned needs to be reverted). 

I think we should include the built CSS for all affected themes on this PR too, since that's what's really loading on the frontend, we want the CSS to be in line with what the sass is declaring.

varia/sass/blocks/blog-posts/_style.scss
madhusudhand: I did a quick check here https://caniuse.com/?search=currentColor. Both seems to be supported by all browsers.
However, I reverted them to be safe on not breaking the sites.

I would have to add a style-lint rule to avoid this in future changes.

MaggieCabrera: yeah, after further googling, it appears most modern browsers will accept both, and we are no longer supporting IE so maybe we can ignore this

MaggieCabrera:
I've tested this locally and on dotcom and it works as intended and color annotations seem to be working. They are taking their sweet time to load though (probably cache) but that's not directly a problem with this PR. Great job, thanks for fighting through this mess!

MaggieCabrera (approved):

Issue #5632, opened 2022-03-08
Archeo: Navigation block blockgap not working


MaggieCabrera:
While working on remote I was going over the Archeo block navigation markup and I think that it is wrong and it's not being applied.

mikachan:
I think `blockGap` is being applied correctly, although the value is about to be changed in this PR: https://github.com/Automattic/themes/pull/5640

Was it just `blockGap` that didn't seem to be working, or was there anything else?

MaggieCabrera:
just blockgap, I changed the value while pairing with Matias and nothing seemed to be happening

mikachan:
Oh! It seems to be working now, or at least it is for me 😅

pbking:
Blockgap is working as expected for me.  Adjusting the blockGap value on navigation block in the `header.html` template part changes the rendering of the desktop menu items.

Closing.  Please reopen if this persists after #5640 lands.

PR #5671, opened 2022-03-16
Archeo: Fix nav block blockGaps values

mikachan:
This PR changes the blockGap values in the header and footer templates from `var(--wp--custom--spacing--small)` to`2.5rem`.

We've only recently changed to the variable in this PR https://github.com/Automattic/themes/pull/5640, however, I can't see this working locally anymore and I don't think this is working on dotcom. The `rem` value seems to work consistently both locally and on dotcom. Here's what it looks like:

Footer nav at 400px:
image

Footer nav at 1200px:
image

Header nav at 1200px:
image

#### Related issues:
Probably related to https://github.com/Automattic/themes/issues/5632

### DIFF
--- a/archeo/inc/patterns/footer.php
+++ b/archeo/inc/patterns/footer.php
@@ -9,7 +9,7 @@ return array(
 	'content'    => '<!-- wp:group {"layout":{"inherit":"true"}} -->
 	<div class="wp-block-group"><!-- wp:group {"align":"wide","layout":{"type":"flex","justifyContent":"space-between"},"style":{"spacing":{"padding":{"bottom":"var(--wp--custom--spacing--medium)","top":"var(--wp--custom--spacing--medium)"}}}} -->
 	<div class="wp-block-group alignwide" style="padding-top: var(--wp--custom--spacing--medium); padding-bottom: var(--wp--custom--spacing--medium);">
-		<!-- wp:navigation {"layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"left"},"overlayMenu":"never","className":"site-footer","style":{"typography":{"fontStyle":"normal"},"spacing":{"blockGap":"var(--wp--custom--spacing--small)"}},"fontSize":"small"} /-->
+		<!-- wp:navigation {"layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"left"},"overlayMenu":"never","className":"site-footer","style":{"typography":{"fontStyle":"normal"},"spacing":{"blockGap":"2.5rem"}},"fontSize":"small"} /-->
 	
 		<!-- wp:paragraph {"align":"left","fontSize":"small","style":{"spacing":{"margin":{"top":0}}}} -->
 		<p class="has-small-font-size" style="margin-top: 0;">' .
--- a/archeo/parts/header.html
+++ b/archeo/parts/header.html
@@ -2,7 +2,7 @@
 <div class="wp-block-group"><!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"bottom":"var(--wp--custom--spacing--outer)","top":"var(--wp--custom--spacing--outer)"}}},"layout":{"type":"flex","justifyContent":"space-between"}} -->
 <div class="wp-block-group alignwide" style="padding-top:var(--wp--custom--spacing--outer);padding-bottom:var(--wp--custom--spacing--outer)"><!-- wp:site-title /-->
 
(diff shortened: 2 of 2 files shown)


archeo/inc/patterns/footer.php
```diff
@@ -9,7 +9,7 @@
 	'content'    => '<!-- wp:group {"layout":{"inherit":"true"}} -->
 	<div class="wp-block-group"><!-- wp:group {"align":"wide","layout":{"type":"flex","justifyContent":"space-between"},"style":{"spacing":{"padding":{"bottom":"var(--wp--custom--spacing--medium)","top":"var(--wp--custom--spacing--medium)"}}}} -->
 	<div class="wp-block-group alignwide" style="padding-top: var(--wp--custom--spacing--medium); padding-bottom: var(--wp--custom--spacing--medium);">
-		<!-- wp:navigation {"layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"left"},"overlayMenu":"never","className":"site-footer","style":{"typography":{"fontStyle":"normal"},"spacing":{"blockGap":"var(--wp--custom--spacing--small)"}},"fontSize":"small"} /-->
+		<!-- wp:navigation {"layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"left"},"overlayMenu":"never","className":"site-footer","style":{"typography":{"fontStyle":"normal"},"spacing":{"blockGap":"2.5rem"}},"fontSize":"small"} /-->
```
jffng, line 12: I also tested with the `clamp(20px, 4vw, 40px)` value and it doesn't work either.

jffng (commented):
This is strange. So at one point a variable value worked for `blockGap`?

I've been thinking about variables in patterns recently and more generally. I think whenever we provide a custom value in a pattern or template, we should provide a fallback value as well. I think that will help down the line when we want to add some of these patterns to the directory, and when custom templates and parts are preserved on theme switch.

That said, adding a fallback value (e.g. `var(--wp--custom--spacing--small, 2.5rem)`) doesn't work here either. Maybe worth reporting a bug upstream to investigate why?

mikachan:
Thanks for taking a look, I was worried it was just my environment! 

>So at one point a variable value worked for blockGap?

Yes, I'm sure this was working last week when we were working on this PR https://github.com/Automattic/themes/pull/5640, but now I'm wondering if I'd messed up my GB environment whilst working on it.. (and it was incorrectly working for me)

I really like the idea of adding a fallback value when we use variables, that makes a lot of sense.

I did take a look at GB issues/changes, it looks like support for blockGap had been removed for the Navigation block and then more recently added back in, but I think variable support was removed completely. I'm not sure I'm understanding the changes correctly though, so spending some time with them now. Here are some relevant links (I think):

- https://github.com/WordPress/gutenberg/pull/37543
- https://github.com/WordPress/gutenberg/pull/37360

jffng:
I see, thanks for the context! 

I wonder if it has to do with the sanitization for the blockGap attribute, since both clamp and var do not work.

scruffian:
I think it doesn't work at the moment and you have to use CSS :(

I played with it a bit here: https://github.com/WordPress/wordpress-develop/pull/2307

mikachan:
> I wonder if it has to do with the sanitization for the blockGap attribute, since both clamp and var do not work.

Yeah I think it could be something like this.

> I think it doesn't work at the moment and you have to use CSS :(

Ah OK, thanks for confirming. 

Hmm, should we pick a value that works in the markup (like the `2.5rem`) or add some CSS so we can still use `var(--wp--custom--spacing--small)`?

jffng:
> I think it doesn't work at the moment and you have to use CSS :(

But why does it work when we supply a non-dynamic value like rem and px for the blockGap attribute? 

> Hmm, should we pick a value that works in the markup (like the 2.5rem) or add some CSS so we can still use var(--wp--custom--spacing--small)?

I think the baked value approach is preferable. So this PR LGTM as long as @kjellr is okay with the `2.5rem` value used.

scruffian:
Oh I wasn't aware of that. Would you mind opening an issue for it?

jffng:
> Oh I wasn't aware of that. Would you mind opening an issue for it?

Done here: https://github.com/WordPress/gutenberg/issues/39583

kjellr:
This looks good. It appears that the spacing I implemented in https://github.com/Automattic/themes/pull/5645/commits/63f02df78bf53d526d923a9145adc646894b16da has been lost at some point, so I'm going to open a quick followup PR to fix that.

kjellr (approved):

PR #5696, opened 2022-03-17
Remote: add large quote pattern

madhusudhand:

Add large quote pattern for Remote theme.

I could not derive top and bottom padding from figma. I considered `20px` padding with reference to other patterns.

#### Screenshots

image

#### Related issue(s):

#5504

### DIFF
--- a/remote/inc/block-patterns.php
+++ b/remote/inc/block-patterns.php
@@ -42,6 +42,7 @@ function remote_register_block_patterns() {
 
 	$block_patterns = array(
 		'hero-text',
+		'large-quote',
 		'hidden-404-content',
 		'hidden-search-form',
 		'image-with-text',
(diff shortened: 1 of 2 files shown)


remote/inc/block-patterns.php
```diff
@@ -14,7 +14,7 @@
  */
 function remote_register_block_patterns() {
 	$block_pattern_categories = array(
-		'pages' => array( 'label' => __( 'Pages', 'remote' ) ),
+		'remote' => array( 'label' => __( 'Remote', 'remote' ) ),
```
MaggieCabrera: This will be affected by https://github.com/Automattic/themes/pull/5664
In this particular case, I think the categories we want for the quote pattern are text and featured. We are not adding a theme specific category

madhusudhand: Cool. Let me revert this change.

MaggieCabrera (approved):
LGTM!

PR #5697, opened 2022-03-17
Update Create Theme Checklist items

mikachan:
@pbking and I had a quick chat about the create-a-theme-checklist items, and we thought a couple needed updating.

The linting makes it look like there are more changes - this is currently just updating the issues list in the `createIssues` function.

- Removed 'from boilerplate' from the first item, as there are many ways to create a theme, not just from a boilerplate
- Added 'Templates: Page templates - Default', to cover creating `page.html`
- Added 'Templates: Post templates - Single', to cover creating `single.html`
- Removed mentions of '(including mobile)', as everything should include mobile
- Removed mentions of '(if needed)', as issues can just be deleted (or not created) if they're not needed

Is there anything else we should update here?


create-a-theme-checklist.js
```diff
@@ -30,69 +30,71 @@ async function createMilestone() {
...
 	const issues = [
-		'Create base theme from boilerplate',
+		'Create base theme',
 		'Create a demo site on dotcom',
 		'theme.json: Typography settings',
 		'theme.json: Color palette',
 		'theme.json: Alternative color palettes',
 		'theme.json: Margin/spacing settings + content layout',
 		'Templates: Page templates - Index',
+		'Templates: Page templates - Default',
```
MaggieCabrera: Why default and not just "page template" or "page"?

create-a-theme-checklist.js
```diff
@@ -30,69 +30,71 @@ async function createMilestone() {
...
 
-async function createIssues( milestoneNumber ) {
+async function createIssues(milestoneNumber) {
 	const issues = [
-		'Create base theme from boilerplate',
+		'Create base theme',
 		'Create a demo site on dotcom',
 		'theme.json: Typography settings',
 		'theme.json: Color palette',
 		'theme.json: Alternative color palettes',
```
MaggieCabrera: we probably don't need this at this point, since we are not including them, when we include theme.json variations we can revisit this

create-a-theme-checklist.js
```diff
@@ -30,69 +30,71 @@ async function createMilestone() {
...
-		'Comment form styles (dotcom and dotorg) (if needed)',
-		'Navigation (if needed)',
-		'Core block settings (if needed)',
+		'Templates: Post templates - Single',
+		'Templates: Header template part',
+		'Templates: Footer template part',
+		'Block patterns',
+		'Comment form styles (dotcom and dotorg)',
+		'Navigation',
+		'Core block settings',
```
MaggieCabrera: I'd rename this to "theme.json: Core block settings"

create-a-theme-checklist.js
```diff
@@ -30,69 +30,71 @@ async function createMilestone() {
...
-		'Block patterns (if needed)',
-		'Comment form styles (dotcom and dotorg) (if needed)',
-		'Navigation (if needed)',
-		'Core block settings (if needed)',
+		'Templates: Post templates - Single',
+		'Templates: Header template part',
+		'Templates: Footer template part',
+		'Block patterns',
+		'Comment form styles (dotcom and dotorg)',
+		'Navigation',
```
MaggieCabrera: We probably don't need this, since it's included in the header template part issue

create-a-theme-checklist.js
MaggieCabrera: Ok I see the other templates are called Page templates - 404 but those are not really pages, they are just 404 template or Search template instead.

mikachan: Yeah I know what you mean, I wasn't sure what to call this! I thought Page templates - Page sounded a little strange, but Default is maybe confusing. Maybe just 'Templates: page.html'?

MaggieCabrera: yeah, probably just naming templates what they are is the most straightforward you can get 🤣

pbking:
This seems like a good change from the default to me. :)
I'm keen to bring this in.  I think it's fine and good to tweak that list when running the utility to suit whoever is running and the unique needs of the project and these are a good upgrade to the base to consider.

pbking (approved):

PR #5703, opened 2022-03-18
Remote: add small quote pattern

madhusudhand:

Added small quote pattern for Remote theme
Updated image license.

image

#### Related issue(s):

#5504


madhusudhand:
@kjellr Can you provide image for the user icon.

MaggieCabrera:
The alignment issue is a known one for this theme. My suggestion is to remove the third empty column and instead of forcing the width of the text column, force the width of the one with the avatar image.

The image you can export it from figma directly like this:
Screenshot 2022-03-18 at 10 42 46

Check the code of the image-with-text pattern to see how we handle images in patterns.

kjellr:
Here's an image credit we can add to `readme.txt`: 

**Tattooed woman with laptop**
License: CC0
Source https://commons.wikimedia.org/wiki/File:Tattooed_woman_with_laptop_(Unsplash).jpg

MaggieCabrera:
Something is happening on the frontend but the image is loading like this:

Screenshot 2022-03-21 at 14 19 32

madhusudhand:
Right alignment seems to be shrinking the width of the image by adding extra margin.
Removing the alignment seems to make the image appear original size.

MaggieCabrera:
Since we are already restricting the size of the column, we can align the image center so that it looks better on mobile:

Not aligned | Align center
--- | ---
Screenshot 2022-03-22 at 13 14 35 | Screenshot 2022-03-22 at 13 14 42

And it would be interesting to see why the right alignment was adding the margin on the frontend but not on the site editor, maybe that's a bug that needs to be reported in GB.

kjellr:
> Since we are already restricting the size of the column, we can align the image center so that it looks better on mobile:

I'd prefer we keep it left aligned actually, since the text and attribution is also left aligned. 👍

madhusudhand:
Created an issue for GB WordPress/gutenberg#39673

MaggieCabrera (approved):

PR #5707, opened 2022-03-18
Removed reference to absent function that was previously removed.

### LINKED ISSUE #5704: Livro: FATAL error in WP-Admin
## Steps to replicate

Line 63 in functions.php causes a fatal error because the function `livro_editor_styles` isn't defined

## Result

Fatal error when trying to load WP admin.

## Expected

No fatal error :)

pbking:

Does what it says on the tin.

#### Related issue(s):

Fixes: #5704

### DIFF
--- a/livro/functions.php
+++ b/livro/functions.php
@@ -59,8 +59,5 @@ endif;
 
 add_action( 'wp_enqueue_scripts', 'livro_styles' );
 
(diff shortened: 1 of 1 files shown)


MaggieCabrera (approved):
LGTM

PR #5708, opened 2022-03-18
Update Archeo theme.json from export

scruffian:
This is an export of Archeo from the create block theme plugin. It brings the theme.json file inline with the standards.


MaggieCabrera (approved):
LGTM!

PR #5711, opened 2022-03-18
Videomaker: fixing fonts path

### LINKED ISSUE #5661: Videomaker: locally-hosted font path is wrong
## Steps to replicate

Activate Videomarker (on a WP.org / self-hosted version of WordPress). Check Dev Tools. Failure to load font error will display. :)

## Result

You should see this error:

_"Downloadable font: download failed (font-family: "Inter" style:normal weight:100..900 stretch:100 src index:0): status=2147746065 source: https://devsite.xyz/assets/fonts/Inter-roman.var.woff2/"_

## Expected

Unless I'm missing something super obvious since I generally only have worked with hybrid / classic themes, the path to the font should not be absolute. As a result the path is missing the necessary `/wp-content/themes/videomarker` bit when loading on the front end.

To get it to work for me I have to hack the CSS file as such, changing the path and removing the slash at the end that was recently added as well.

`src: url("./fonts/Inter-roman.var.woff2") format("woff2");`

matiasbenedetto:
Videomaker fixes for:
- Font assets paths.
- Font assets filenames change to avoid Apache server problem described here: https://github.com/philipbelesky/inter-ui/issues/17

#### Related issue(s):
https://github.com/Automattic/themes/issues/5661


MaggieCabrera:
Should we change Videomaker to use fonts [like this](https://github.com/Automattic/themes/pull/5609)?

mikachan:
> Should we change Videomaker to use fonts [like this](https://github.com/Automattic/themes/pull/5609)?

Yeah I think that's a good idea.

MaggieCabrera:
Mmmh, we might break the customizer options if we do that though! maybe it's not such a good idea until the Jetpack font selector is ready for users on dotcom.

mikachan:
Oh good point! Maybe we should wait then.

pbking:
I agree that Blockbase themes should wait for some font management solution before we upgrade this.

Change looks good.  I think we should ship it.

pbking (approved):

PR #5714, opened 2022-03-21
Remote: subscribe pattern

MaggieCabrera:

This PR includes the Jetpack subscribe block pattern. To test this, checkout this branch and run `npm run deploy:push:changes` to see the pattern on your sandbox.

Screenshot 2022-03-21 at 13 59 00

The Jetpack block doesn't allow for changing the background color of the input or the input text. This is as close as I could get to the design.

#### Related issue(s):

### DIFF
--- a/remote/inc/patterns/subscribe.php
+++ b/remote/inc/patterns/subscribe.php
@@ -1,0 +1,19 @@
+<?php
+/**
+ * Subscription form
+ */
+return array(
+	'title'      => __( 'Subscription form', 'remote' ),
+	'categories' => array( 'text', 'featured' ),
+	'content'    => '<!-- wp:paragraph {"style":{"typography":{"textTransform":"uppercase"}}} -->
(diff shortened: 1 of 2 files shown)


remote/inc/patterns/subscribe.php
```diff
@@ -0,0 +1,19 @@
+<?php
+/**
+ * Subscription form
+ */
+return array(
+	'title'      => __( 'Subscription form', 'remote' ),
+	'categories' => array( 'text', 'featured' ),
+	'content'    => '<!-- wp:paragraph {"style":{"typography":{"textTransform":"uppercase"}}} -->
```
jffng: I think this should also have the Small font size preset, if I'm interpreting the design correctly.

MaggieCabrera: you are right!

jffng (approved):

PR #5715, opened 2022-03-21
Archeo: Load patterns automatically

scruffian:
Loads the patterns for Archeo automatically based on https://github.com/WordPress/gutenberg/pull/36751/files

#### Related issue(s):
https://github.com/WordPress/gutenberg/pull/36751/files


MaggieCabrera:
[This](https://github.com/WordPress/gutenberg/pull/36751) will land in .com in 12.9

archeo/inc/block-patterns.php
```diff
@@ -43,37 +43,5 @@ function archeo_register_block_patterns() {
...
-	foreach ( $block_patterns as $block_pattern ) {
-		$pattern_file = get_theme_file_path( '/inc/patterns/' . $block_pattern . '.php' );
-
-		register_block_pattern(
-			'archeo/' . $block_pattern,
-			require $pattern_file
-		);
-	}
 }
 add_action( 'init', 'archeo_register_block_patterns', 9 );
```
MaggieCabrera: Maybe we should include this in functions.php. I don't think we need a file in inc/ just for this.

archeo/patterns/simple-list-of-posts-with-background.php
```diff
@@ -0,0 +1,35 @@
+<?php
+/**
+ * Title: Simple list of posts with title and background
+ * Slug: archeo/simple-list-of-posts-with-background
+ * Categories: featured, query
```
MaggieCabrera, line 5: should we add the query block type here and in the other query pattern?

MaggieCabrera: looks like we didn't have that on the old pattern, but I think we could use that here

MaggieCabrera:
This is looking good! I just have minor comments, thanks for doing this

mikachan:
I've removed inc/block-patterns.php and moved the code into functions.php, and then I've updated two patterns which have since been updated on trunk. I think this is ready for another review.

jffng (commented):
Very cool, working as advertised.

I did notice that some of the patterns load in a random order in the inserter, check out the Images category and notice how a couple Archeo patterns come first, followed by Gutenberg default patterns, then more Archeo patterns. Is there a way to have the theme's always appear first?

archeo/functions.php
```diff
@@ -60,5 +60,43 @@ function archeo_styles() {
...
-// Add block patterns
-require get_template_directory() . '/inc/block-patterns.php';
+/**
+ * Registers block patterns and categories.
+ *
+ * @since Archeo 1.0
+ *
+ * @return void
+ */
+function archeo_register_block_patterns() {
```
jffng: Nit but would this be a more accurate name for this function?

```suggestion
function archeo_register_block_pattern_categories() {
```

mikachan:
> I did notice that some of the patterns load in a random order in the inserter, check out the Images category and notice how a couple Archeo patterns come first, followed by Gutenberg default patterns, then more Archeo patterns. Is there a way to have the theme's always appear first?

Oh yeah, I can see the same. It doesn't look like they're listed alphabetically, plus I tried adding numbers to the start of the Archeo pattern titles but it didn't seem to make a difference to the order.

mikachan:
Just spotted @MaggieCabrera has already opened up an issue to address the pattern order: https://github.com/WordPress/gutenberg/issues/39500 

I'm going to move this into Approved but not mergeable, as I think it's ready to go when 12.9 lands.

mikachan (approved):

Issue #5705, opened 2022-03-18
Livro: Newsreader font is no longer loaded


markhowellsmead:
## Steps to replicate

The Newsreader font is still defined in CSS but the font file/s are no longer loaded.

## Result

The website typography appears in the standard browser fallback serif font.

## Expected

The Newsreader fonts in the assets folder should be loaded by the theme.

jffng:
Thanks for the report, we just made the switch to [load the fonts via theme.json](https://github.com/Automattic/themes/pull/5609) since Gutenberg was upgraded, my guess is this is an artifact of some caching.

Newsreader is loading on the demo site, so I'm going to close this:

<img width="1904" alt="Screen Shot 2022-03-18 at 11 44 57 AM" src="https://user-images.githubusercontent.com/5375500/159035940-8af11be2-995b-4172-ae21-7aa58033f1e1.png">

markhowellsmead:
@jffng Font still not loading even with caching disabled on the site and it's running the latest version of the theme (1.0.9). There is no reference in the source code to the WOFF files. If I look at the documentation for theme.json, there is no reference that the src attribute is supported for `fontFamilies`.

jffng:
The Webfonts API was recently added https://github.com/WordPress/gutenberg/pull/37140, and the theme.json documentation needs to be updated accordingly. This feature is not in core yet, are you running the Gutenberg plugin?

markhowellsmead:
No, I'm using the Theme directly from https://wordpress.org/themes/livro/. The Gutenberg plugin is highly unstable and often breaks functionality, so it's not possible to use it on a production site. I'm guessing that this issue is happening for the 200+ other non-WordPress.com users too, unless they're using the plugin.

PR #5718, opened 2022-03-21
Added a dependency check for Gutenberg 

pbking:
Added a dependency check for Gutenberg based on metadata stored in st…yle.css

Adds a (reuse-able) `/inc/gutenberg-dependency-check.php` which checks style.css for a `Requires Gutenberg:` entry and compares that against the version of Gutenberg installed.

If unactivated or a version is installed that is lower than expressed in the theme metadata an error is shown in the WP admin.

image

Tested with 12.8.1 installed.  Livro requires 12.8 (where webfonts was introduced).  

#### Related issue(s):

A tinker after discussing #5705

### DIFF
--- a/livro/inc/gutenberg-dependency-check.php
+++ b/livro/inc/gutenberg-dependency-check.php
@@ -9,7 +9,11 @@ function show_admin_messages() {
 		return;
 	}
 	$version = $match[0];
-	if( defined( 'IS_GUTENBERG_PLUGIN' ) ) {
+	if( defined( 'IS_GUTENBERG_PLUGIN' ) && defined( 'GUTENBERG_VERSION' ) ) {
+		if ( version_compare( GUTENBERG_VERSION, trim( $version ) ) >= 0 ) {
+			return;
+		}
+	}
 		$plugins = get_plugins();
 		foreach ( $plugins as $plugin ) {
 			if ( 'Gutenberg' === $plugin['Name'] ) {
--- a/livro/inc/gutenberg-dependency-check.php
+++ b/livro/inc/gutenberg-dependency-check.php
@@ -9,7 +9,7 @@ function show_admin_messages() {
 		return;
 	}
 	$version = $match[0];
-	if ( is_plugin_active( 'gutenberg/gutenberg.php' ) ) {
+	if( defined( 'IS_GUTENBERG_PLUGIN' ) ) {
 		$plugins = get_plugins();
 		foreach ( $plugins as $plugin ) {
 			if ( 'Gutenberg' === $plugin['Name'] ) {
--- a/livro/inc/gutenberg-dependency-check.php
+++ b/livro/inc/gutenberg-dependency-check.php
@@ -1,0 +1,27 @@
+<?php
+
+add_action('admin_notices', 'showAdminMessages');
+
+function showAdminMessages() {
+	$metadata = file_get_contents( get_stylesheet_directory() . '/style.css' );
+	preg_match( '/(?<=Requires Gutenberg:).+/', $metadata, $match );
+	if( 0 === sizeof($match) ) {
+		return;
+	}
+	$version = $match[0];
+	if( is_plugin_active( 'gutenberg/gutenberg.php' ) ) {
+		$plugins = get_plugins();
+		foreach( $plugins as $plugin ) {
+			if( $plugin['Name'] === 'Gutenberg' ) {
+				if( version_compare(trim($plugin['Version']), trim($version) ) >= 0 ) {
+					return;
+				}
+				break;
+			}
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


jffng (approved):
I think this is a nice clarification, it allows us to make explicit when a theme depends on Gutenberg.

jffng:
Ah I maybe approved too soon. I tested on dotcom and the warning is showing up when I activate Livro and go to /wp-admin:

Screen Shot 2022-03-21 at 4 28 47 PM

livro/inc/gutenberg-dependency-check.php
```diff
@@ -0,0 +1,27 @@
...
+add_action( 'admin_notices', 'show_admin_messages' );
+
+function show_admin_messages() {
+	$metadata = file_get_contents( get_stylesheet_directory() . '/style.css' );
+	preg_match( '/(?<=Requires Gutenberg:).+/', $metadata, $match );
+	if ( 0 === sizeof( $match ) ) {
+		return;
+	}
+	$version = $match[0];
+	if ( is_plugin_active( 'gutenberg/gutenberg.php' ) ) {
```
MaggieCabrera: this naming convention will not be true for all environments, as it's certainly not the case over on dotcom

pbking: Yea, I was afraid of that.  I was hoping I could check for activation in the loop above but it just tells what's installed, not what's activated.  Is there a better way to tell if a plugin (generically) is activated or should we do something Gutenberg specific?

MaggieCabrera: I don't know! I think I tried something like this before for similar reasons and got stuck at this point too.

scruffian: https://developer.wordpress.org/reference/functions/is_plugin_active/

scruffian: Maybe we can use `class_exists`?

pbking: I found [this](https://github.com/WordPress/gutenberg/blob/trunk/lib/load.php#L12) which seems to be suited to the purpose.

Checking for that constant seems to work well.

livro/inc/gutenberg-dependency-check.php
```diff
@@ -0,0 +1,27 @@
...
+function show_admin_messages() {
+	$metadata = file_get_contents( get_stylesheet_directory() . '/style.css' );
+	preg_match( '/(?<=Requires Gutenberg:).+/', $metadata, $match );
+	if ( 0 === sizeof( $match ) ) {
+		return;
+	}
+	$version = $match[0];
+	if( defined( 'IS_GUTENBERG_PLUGIN' ) ) {
+		$plugins = get_plugins();
+		foreach ( $plugins as $plugin ) {
```
jffng: I gave this a test on wpcom and it was still throwing the warning, I guess because Gutenberg is not returned from `get_plugins`.

livro/inc/gutenberg-dependency-check.php
```diff
@@ -0,0 +1,27 @@
...
+add_action( 'admin_notices', 'show_admin_messages' );
+
+function show_admin_messages() {
+	$metadata = file_get_contents( get_stylesheet_directory() . '/style.css' );
+	preg_match( '/(?<=Requires Gutenberg:).+/', $metadata, $match );
+	if ( 0 === sizeof( $match ) ) {
+		return;
+	}
+	$version = $match[0];
+	if( defined( 'IS_GUTENBERG_PLUGIN' ) ) {
```
jffng: I _think_ we can simplify to something like:

```suggestion
	if( defined( 'IS_GUTENBERG_PLUGIN' ) && defined( 'GUTENBERG_VERSION' ) ) {
		if ( version_compare( GUTENBERG_VERSION, trim( $version ) ) >= 0 ) {
			return;
		}
	}
```

pbking: Or even just `GUTENBERG_VERSION`?

pbking: Shoot, I merged that suggestion too quickly.  GUTENBERG_VERSION doesn't seem to be defined when in development mode...

pbking:
Pushed a change that leverages all the version checking methods and just skips version checking if there is neither GUTENBERG_VERSION constant OR get_plugins() doesn't return any info about Gutenberg.

It's a bit heavier than I was hoping for but I think it works OK.  

Tested in wpcom and there was no error message if it wasn't expected and did show if it was.  That is leveraging GUTENBERG_VERSION.

Tested locally with development version and get_plugins() did the job there.

jffng (approved):
Working with development, plugin, and wpcom GB environments. Thanks @pbking !

PR #5719, opened 2022-03-22
[Archeo] Add "Page (No Title)" template

kjellr:
This PR adds a quick alternate version of our Page template. It's exactly the same, except it's missing the page title. This will come in handy for a couple of the theme demos. 

Page|Page (No Title)
---|---
Screen Shot 2022-03-22 at 9 41 06 AM|Screen Shot 2022-03-22 at 9 40 57 AM


mikachan (approved):
Looks good and works well!

PR #5725, opened 2022-03-22
Add/pendant header

### LINKED ISSUE #5679: Pendant: Template Parts: Header
image

image

pbking:

This change adds a header template to Pendant aiming for the design.

A smattering of CSS was added mostly to handle the mobile version of the header.

The template was created using the FSE and saved using the Create Block Theme 'save' functionality.  Only the menu locations needed to be removed from the template.

The custom.gap.horizontal value in theme.json had to be adjusted by hand.  That value locks in a horizontal gap of 100px at around 1200 pixel screensize.  It then scales down to 20px around 400px.

In order to apply theme-specific style to this header (the included CSS) it isn't offered (and won't be) offered as a Block Pattern.

When reviewing the design please ignore the sub-navigation menu items and the entire mobile menu.  

#### Related issue(s):

Fixes #5679

#### Screen Shots:

Header Desktop (wide):

image

Header Desktop (narrow):

image

Mobile:

image


pbking:
@beafialho I'm not sure what the design intend of the 'Search' label is.  At the moment it's a link out to the search page, nestled in a navigation block.  If you have something different in mind please LMK.

beafialho:
@pbking the original intent of the 'Search' label in the header was to click it and open the search field, as shown below:

Captura de ecrã 2022-03-23, às 17 10 44

However, because I don't think that's possible, I suggest we split the menu items on both sides. There shouldn't be four items on one side, as it's not a good solution visually when they wrap into another line. Do you think that's doable?

Captura de ecrã 2022-03-23, às 17 11 06

Captura de ecrã 2022-03-23, às 17 14 03

At 900px it can turn into the mobile menu:

Captura de ecrã 2022-03-23, às 17 14 11

pbking:
I think we're locked into the 600px media query from Gutenberg as to when the navigation goes from "desktop" to "mobile".  Considering that, and the fact that there is always potential for the menus to wrap we'll need to style that situation somehow.  Ideally a user's menus wouldn't be so plentiful that they have to wrap but it's always a possibility.

I can explore separating the menus as you describe but I'm not sure that I can do that either.  My first thought was "just make two menus" but then only half of the navigation would be in the responsive menu.  

I DO think that I can land the styled search block you note:
image
I don't think it could transform on that on click (it would just always look like that) but that might be a viable solution?

pbking:
image

What do you think about this?  Threw that together petty easily...

And another in case it does have to wrap:
image

beafialho:
Thank you, the Search looks really good!

> And another in case it does have to wrap:

I really want to avoid it wrapping up like that. After thinking about this and realising there's no way we can make it work (not even with the Row block because we'd have 2 navigations), can we accomplish something like this instead? 

Having the mobile menu always on the left and social icons on the right. That way, the original intention of having the centered Site Title works.

Captura de ecrã 2022-03-24, às 15 25 47

pbking:
> can we accomplish something like this instead?

Yes absolutely!

pbking:
I set the top/bottom margins to `3vm`, that seems to suit both wide and small dimensions.  The left/right margins use the same `clamp()` values noted above.

Do you think we're on to something here?

Very wide:
image

Kinda wide:
image

Not very wide really:
image

Handheld:
image

beafialho:
Looks good! Is that the smallest possible size for the social icon? There's a slight difference between the screenshot and the mockup.

And I guess we're locked on the default hamburger menu icon, is that correct?

Captura de ecrã 2022-03-24, às 16 41 59

Also, despite the `clamp` function the header elements will align with the template right?

Captura de ecrã 2022-03-24, às 16 43 58

pbking:
> despite the clamp function the header elements will align with the template right?

Right, the clamp() value is site-wide for anything that's "full width".

> And I guess we're locked on the default hamburger menu icon, is that correct?

Yes. :( Though I look forward to the day that we can adjust that in the FSE.  I'm not sure why that is the default icon anyway!?  The triple-stack is what I would expect...

> Looks good! Is that the smallest possible size for the social icon? There's a slight difference between the screenshot and the mockup.

That is the smallest (only) size that we can get from the FSE.  I added a line of CSS and got this:

image

image

Dig?

beafialho:
> Right, the clamp() value is site-wide for anything that's "full width".

👍 

> Yes. :( Though I look forward to the day that we can adjust that in the FSE. I'm not sure why that is the default icon anyway!? The triple-stack is what I would expect...

Two stack can work well in some occasions, but it would be great to be able to customize the icon in FSE! Would it be too much to add a line of css to define thinner hamburger lines? It would suit the theme and typeface better. It's ok if that's not viable though.

> I added a line of CSS and got this:

Yes, better!

pbking:
> Would it be too much to add a line of css to define thinner hamburger lines? It would suit the theme and typeface better. 

I don't believe I can even do that!  It's SVG so I have less control over its details. :( 

I'm going to bring this branch in now and we can cycle on additional tweaks as we go.

PR #5735, opened 2022-03-23
Added Gutenberg check to Blockbase themes

pbking:

This adds a slightly modified version of the Gutenberg Check that was added to Livro.

This version takes into account child theme-ness.  Any child of Blockbase shows the same Gutenberg Required message.  Additionally the child may express a DIFFERENT version of Gutenberg that is required and the highest version will be used for the check/message.

#### Testing

* Activate Blockbase
* See no message.
* Change version of `Requires Gutenberg:` to 13.12 or something higher than what is available: See the message.  Return the value to normal.
* Deactivate Gutenberg: See the message. (Reactivate Gutenberg)
* Activate a Blockbase theme (such as Quadrat).
* See no message.
* Deactivate Gutenberg: See the message. (Reactivate Gutenberg)
* Add a `Requires Gutenberg: 15.23` entry to Quadrat's (or whatever theme you activated)'s style.css metadata: See the message.

image

NOTE: We should make these same changes to Levro so that any child themes have the same Gutenberg requirement messaging.


blockbase/inc/gutenberg-dependency-check.php
```diff
@@ -0,0 +1,60 @@
...
+}
+
+function show_admin_messages() {
+
+	if ( is_child_theme() ) {
+		$metadata = file_get_contents( get_template_directory() . '/style.css' );
+		preg_match( '/(?<=Requires Gutenberg:).+/', $metadata, $match );
+	}
+
+	if ( sizeof( $match ) > 0 ) {
```
jffng: I get an error that `$match` is undefined when I activate Blockbase: 

Screen Shot 2022-03-24 at 2 31 02 PM

Could we define `$metadata` before the child theme conditional, and overwrite it the conditional is met? Then right after this conditional but before the `sizeof` check, run `preg_match` to ensure $match is defined.

pbking: Oh, odd that it didn't error for me...

I'm not sure how that `if( sizeof())` check snuck outside of the `is_child_theme()` check but I put that back in there where it belongs and you shouldn't error now.

We could indeed invert that logic as you suggest but that would be a larger change. It's otherwise 6 vs 1/2 dozen to me.

jffng (approved):
🚢

seanconklin:
With today's update to Blockbase v2.0.31 I'm getting the message "The installed theme requires Gutenberg version 12.8 or higher." on my site with just the Blockbase theme (no child theme) and WordPress core v5.9.2 without the Gutenberg plugin. 

WordPress v5.9.2 contains Gutenberg v10.8-11.9 according to [https://developer.wordpress.org/block-editor/contributors/versions-in-wordpress/](https://developer.wordpress.org/block-editor/contributors/versions-in-wordpress/)

My expectation is for this vanilla Block theme to work with with WordPress 5.9+ core, though it's certainly understandable that child themes may layer on Gutenberg plugin features and require less stable versions.

marceloaof:
I would just like to add to that decision, that I also wish I could not depend on the plugin for the stable Blockbase theme to work.
Is it not better instead to upgrade the minimum WordPress version required for Blockbase theme, to always be the latest available? (WP 5.9.2 today). As soon as WP 6.0 is released, Blockbase theme would require at least: WP 6.0, keeping abreast of the latest advances provided by the Gutenberg plugin. (Today, Blockbase theme is keeping support to WP 5.7, 5.8 and 5.9)
This strategy would reduce developers' maintenance work, and would make life easier for developers and users.
Could this verification check for an additional Gutenberg plugin be disabled for the Blockbase theme and be enabled for its children that need it?

pbking:
@seanconklin and @marceloaof thank you for your feedback.

We just pushed a [change](https://github.com/Automattic/themes/pull/5823) removing the 'gut check'.  The environment we expect to run this in for our own uses includes the Gutenberg plugin, usually the latest released version.  So that's the capabilities we expect any of the newer themes we build to have.  
We expect to keep pace with that release cadence for features we are able to use.  
Blockbase doesn't currently meet the criteria where something higher than the latest WordPress is needed but it has in the past and may again in the future.  

And if not this theme then others will.

What is the best way, do you think, to handle communicating and handling that?

seanconklin:
Since full site editing has the power to make the theme irrelevant by supporting blank canvas design and high performance - we need ultra-light themes available for FSE. That’s what I figured Blockbase was, but I understand it has other purposes as well. I’ve built a virgin FSE theme I’m considering putting up. The best way to communicate this I’d say is to have a viable example to point to for the use case I’m suggesting. There’s more to document too, such as writing Block Style functions to achieve editor style consistency with the front end. I’ll put up a video, and link here in a bit.

marceloaof:
Hello. I'm not able to write concisely today, sorry for the long message.

I understand your message. The intended strategy for the Blockbase theme makes sense for internal work.

In this scenario, I write as feedback (not a complain) that, as an average person, I miss an empty and official theme from WordPress/Automattic available on the Themes repository, similar to Blockbase, with the distinction of not needing the Gutenberg plugin as a dependency. A theme which keeps as little extra code and design decisions as possible, so that each person can be encouraged to fully build their theme within the Site Editor, but with a development cycle similar to TT2, in the sense of introducing new features according to WP development cycle.

I understand that the team also offers Empty Theme and Block Canvas theme on GitHub, which perhaps fill the slower pace without the need of Gutenberg, while also maintaining the intention of Blockbase, but currently they are not listed in the WordPress Themes repository. Currently, on the WordPress Themes repository, from the 60 FSE-themes, only Blockbase theme is positioned to fulfill needs like mine. I can also see agencies taking an interest in an official theme similar to Blockbase, that is updated according to the WordPress development cycle rather than the Gutenberg development cycle.

seanconklin:
Webdevstudios blog post here [https://webdevstudios.com/2021/09/14/wordpress-fse-and-block-based-themes/](https://webdevstudios.com/2021/09/14/wordpress-fse-and-block-based-themes/) contains simple instructions for initializing a practically empty FSE theme. It works!

For most client projects though I lean towards using Twenty Twenty-Two theme to get some testing and support, minimal editor styling, base templates, global style editor definitions, etc.

I agree let's get a Blockbase type theme that tracks the release branch. I dream of a default theme that is configurable to exclude any nonessential stuff being loaded. I know theme.json doesn't exactly work that way (it's defined or undefined in a file), but maybe there's a way to build it - Twenty Twenty-Three!

MaggieCabrera:
Thanks so much for this feedback! Sometimes we forget that we aren't the only ones using this theme, and it's paramount that we are remembered that is not the case.

Blockbase started as a very experimental theme that tried to stabilize things as much as possible to allow production sites to work with the new GB features. At the start, before even 5.7, that required GB a 100%. Now that's not the case so much, but we still want to include the latest improvements as soon as possible.

I wonder if we should release Block Canvas? It was never advertised but it seems like it's getting some attention. I recommend you all give it a try if you want a lightweight FSE starter theme.

Right now Blockbase maybe should wait to include breaking features until a core release. For example: we could be including the comments query block to Blockbase today and remove the styles related to comments today, but that will require the plugin. We can create the fix and simply wait until 6.0 to ship it. Blockbase's comments already work, so there's no need for the fix to be on the theme today. Blockbase is as stable as it gets, we don't need to patch it on the daily if that means we are breaking other people's child themes or requiring the plugin.

seanconklin:
I think releasing the Block Canvas theme is a good fix for our purpose of having a stable, supported, minimalistic theme to use on production grade projects. Beware the name is easily confused with the Blank Canvas theme. Then Blockbase could be described as a canvas theme that tracks the Gutenberg plugin as an upstream for official themes or others that utilize bleeding-edge Gutenberg features. Many Linux projects are labeled with -devel for these branches, so perhaps it's worth considering blockbase vs. blockbase-devel or for that matter blockbase-gutenberg to be more consolidated and on brand. I agree the naming is a real challenge here!

PR #5741, opened 2022-03-24
Blockbase: add style overrides for navigation block links

### LINKED ISSUE #5670: Blockbase: Navigation block shows underlines on links in Site Editor, but not on live site
## Steps to replicate

* Activate Blockbase or a child theme
* In the Site Editor - add a menu to the Navigation block in the header. Confirm it shows underlines on the links/items.
*  View the live site.
Screen Shot 2022-03-16 at 3 13 40 pm

## Result

Links show without underlines

Screen Shot 2022-03-16 at 3 13 25 pm

## Expected

I expected the display of the menu items to be consistent between Site Editor and live site

madhusudhand:

Gutenberg ads `text-decoration: none` to links in navigation block and CSS specificity in theme overrides is less specific.
* included parent class in style overrides.

#### Results

* Similar to other page links, navigation block links doesn't show underline in normal state.
* underline appear on hover state.

blockbase-nav-links

#### Related issue(s):

Fixes: #5670

### DIFF
--- a/blockbase/sass/elements/_links.scss
+++ b/blockbase/sass/elements/_links.scss
@@ -1,5 +1,6 @@
 // Default link styles
-a {
+a,
+.wp-block-navigation a {
 	cursor: pointer;
 	text-underline-offset: 0.15em;
 	text-decoration: none;
@@ -11,8 +12,8 @@ a {
 }
 
 //Links that appear in the main content area
-.block-editor-block-list__layout a, // Needed for the post area
-.wp-block-post-content a {
+main.block-editor-block-list__layout a, // Needed for the post area
+.wp-block-post a {
 	text-decoration-line: underline;
 
 	&:hover,
@@ -23,6 +24,7 @@ a {
 
 // Select the focus states of all non-wpadmin and screen reader links
 a:not(.ab-item):not(.screen-reader-shortcut) {
+
 	&:active,
 	&:focus {
 		outline: 1px dotted currentColor;
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -221,24 +221,27 @@ body.admin-bar .wp-site-blocks {
  * Elements
  * - Styles for basic HTML elemants
  */
-a {
+a,
+.wp-block-navigation a {
   cursor: pointer;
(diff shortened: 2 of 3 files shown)


mikachan (commented):
This works great for me on the front end, but I'm still seeing underlines on the links in the header in the editor.

image

We probably also need to adjust the classes used for the links in the main content area, it looks like the GB classes have changed since we added these.

In `sass/elements/_links.scss` on lines 14-16, it looks like we could change the selectors to:

```
//Links that appear in the main content area
main.block-editor-block-list__block a, // Needed for the post area
.wp-block-post a {
```

This means the `text-decoration-line: underline;` rule should only apply to links in the content area, and not the header (or footer), which I think is what is currently happening.

madhusudhand:
@mikachan in general it feels right to me to preserve **underline by default and remove on hover** everywhere in the editor, when it is a link, instead of just to content area. Are we considering links in header and footer differently?

mikachan:
> Are we considering links in header and footer differently?

Yes that's right, the links outside the main content area (including the header and footer) should have no underline by default, and underline on hover/focus. The links in the main content area should have underline by default, and no underline on hover/focus.

blockbase/sass/elements/_links.scss
```diff
@@ -11,7 +12,7 @@ a {
 }
 
 //Links that appear in the main content area
-.block-editor-block-list__layout a, // Needed for the post area
+main.block-editor-block-list__layout a, // Needed for the post area
 .wp-block-post-content a {
```
mikachan: This looks good in the editor for the main content where I can see the underlines, but I still can't see the underlines on the front end, e.g. on the homepage:

image

Looks like we just need to change this selector to `.wp-block-post a` instead of `.wp-block-post-content a`.

mikachan (commented):
I think there's just one small tweak and then this one is done!

blockbase/sass/elements/_links.scss
madhusudhand: Included suggested changes and following is the UI.

image

mikachan: Awesome, looks good to me! 🎉

mikachan (approved):

Issue #5681, opened 2022-03-16
Pendant: Templates: Index


pbking:
<img width="459" alt="image" src="https://user-images.githubusercontent.com/146530/159974644-cf4cc20a-8e42-44ad-ac84-c97751fe9fce.png">

No need to make any patterns yet, just adjust index template to match the above.

pbking:
Closing, will seek design review stage instead of approval.

PR #5751, opened 2022-03-25
Pendant: layout for index template

MaggieCabrera:

This PR introduces the changes to the index template to fit the design. I don't know what to do about the heading on top of the page, so for now I made it to be static text. If we go with that we should make it a hidden pattern for i18n.

There are a few GB PRs needed for some of the blocks to work like in the design. They have been recently merged or are about to:

https://github.com/WordPress/gutenberg/pull/39796
https://github.com/WordPress/gutenberg/pull/39835

Screenshot 2022-03-29 at 11-52-08 Skatepark – Tagline

Screenshot 2022-03-29 at 11 51 54

#### Related issue(s):

May address #5681

### DIFF
--- a/pendant/templates/index.html
+++ b/pendant/templates/index.html
@@ -12,8 +12,7 @@
 	<!-- wp:group {"style":{"spacing":{"padding":{"top":"2em","bottom":"2em"}}}} -->
 	<div class="wp-block-group" style="padding-top:2em;padding-bottom:2em"><!-- wp:post-featured-image {"isLink":true,"width":"100%","height":"16em"} /-->
 	
-	<!-- wp:post-title {"isLink":true} /-->
-	
+	<!-- wp:post-title {"level":3,"isLink":true} /-->
 	<!-- wp:post-excerpt /-->
 
 	<!-- wp:read-more {"style":{"typography":{"fontStyle":"normal","fontWeight":"500"}},"fontSize":"medium"} /--></div>
--- a/pendant/templates/index.html
+++ b/pendant/templates/index.html
@@ -7,8 +7,8 @@
 <!-- /wp:group -->
 
 <!-- wp:query {"queryId":1,"query":{"perPage":4,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"exclude","inherit":true},"tagName":"main","displayLayout":{"type":"flex","columns":2},"layout":{"inherit":true}} -->
-<main class="wp-block-query"><!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"1em"}}}} -->
-	<div class="wp-block-group alignwide" style="padding-top:1em"><!-- wp:post-template -->
+<main class="wp-block-query"><!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"1em","bottom":"1.5em"}}}} -->
+	<div class="wp-block-group alignwide" style="padding-top:1em;padding-bottom: 1.5em"><!-- wp:post-template -->
 	<!-- wp:group {"style":{"spacing":{"padding":{"top":"2em","bottom":"2em"}}}} -->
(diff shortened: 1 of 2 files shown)


pendant/theme.json
```diff
@@ -218,6 +218,56 @@
...
+                            "textDecoration": "none"
+                        }
+                    }
+                }
+            },
+            "core/query-pagination-next":{
+                "typography": {
+                    "fontWeight": "500",
+                    "textTransform": "uppercase",
+                    "fontSize": "var(--wp--preset--font-size--small)"
```
jffng: The comps suggest this is the "extra small" font size (15px), but it doesn't look like that variable is configured yet via custom theme.json settings. Should we add it?

pendant/templates/index.html
```diff
@@ -1,33 +1,43 @@
...
+<!-- wp:group {"style":{"spacing":{"margin":{"top":"0px"},"padding":{"top":"5em","bottom":"5em"}}},"backgroundColor":"foreground","textColor":"background","layout":{"inherit":true}} -->
+<div class="wp-block-group has-background-color has-foreground-background-color has-text-color has-background" style="margin-top:0px;padding-top:5em;padding-bottom:5em"><!-- wp:heading {"level":1,"align":"wide"} -->
+	<h1 class="alignwide">Articles</h1>
+	<!-- /wp:heading --></div>
+<!-- /wp:group -->
+
+<!-- wp:query {"queryId":1,"query":{"perPage":4,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"exclude","inherit":true},"tagName":"main","displayLayout":{"type":"flex","columns":2},"layout":{"inherit":true}} -->
+<main class="wp-block-query"><!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"1em","bottom":"1.5em"}}}} -->
+	<div class="wp-block-group alignwide" style="padding-top:1em;padding-bottom: 1.5em"><!-- wp:post-template -->
+	<!-- wp:group {"style":{"spacing":{"padding":{"top":"2em","bottom":"2em"}}}} -->
```
jffng, line 12: I think the comps suggest there is less spacing here but I think yours looks good so maybe we leave it to the design pass of the templates.

jffng (commented):
Couple minor comments but generally LGTM!

pendant/theme.json
MaggieCabrera: yeah I think you are right

MaggieCabrera:
What do you folks think about the heading up top?

jffng:
> What do you folks think about the heading up top?

Could we make it a `query-title` block instead? That way we wouldn't need a separate archive template, since they are the same from a design perspective, as far as I can tell.

If we can't, I think static that we replace with a hidden pattern for translation purposes works too.

MaggieCabrera:
> > What do you folks think about the heading up top?
> 
> Could we make it a `query-title` block instead? That way we wouldn't need a separate archive template, since they are the same from a design perspective, as far as I can tell.
> 
> If we can't, I think static that we replace with a hidden pattern for translation purposes works too.

This makes me think when index.html will ever appear if we have home.html and archive.html defined... According to https://developer.wordpress.org/themes/basics/template-hierarchy/ that's... never

jffng:
True — if we define home and archive, index will never be used. Since Pendant will have a home template, should we switch this one to archive?

MaggieCabrera:
> True — if we define home and archive, index will never be used. Since Pendant will have a home template, should we switch this one to archive?

Well, we need index.html because without it the theme isn't detected as a block theme and the editor doesn't work. We could add the query title to index and remove the archive template instead but I think I'm going to open a GB issue about this because this doesn't make much sense in this case.

https://github.com/WordPress/gutenberg/issues/39931

pbking:
Made a few minor tweaks but going to bring this in.  Leaving the attached issue open for final design sign off.

PR #5752, opened 2022-03-25
Blockbase: update font sizes to rem instead of px for a11y

### LINKED ISSUE #5659: Blockbase theme: body text does not scale to browser default settings
In Chrome, under Appearance in Settings, change the font size to large or very large. Body text (text in the main section of the webpage) does not scale when the default font size in the browser is changed, which is poor for accessibility. 

I can correct it by adding "body {font-size:100%;}" in the Blockbase's ponyfil.css file.  Alternatively, defining the body element as font-size:1.2rem, for example, also appears to work. Normally, one would put the 100% font size in the HTML tag; however, in this case, it doesn't do anything. Also, one has to make sure that the header and footer template parts are defining the font sizes in rem/em in the editor, not px., so that the text in this section also changes appropriately.

In addition, looking at Blockbase's theme.json font sizes, starting line 375, are being defined in px rather than rems. 

For example, 
         fontSizes": [
				{
					"name": "Small",
					"size": "16px",
					"slug": "small"
				}, 

etc..

In contrast, the Twenty Twenty Two Theme defines font-size with rem:

"fontSizes": [
				{
					"size": "1rem",
					"slug": "small"
				},

The site https://richtabor.com/standardizing-theme-json-font-sizes/ suggests standardizing font sizes in rems as in the following:

```
{
"settings": {
    "typography": {
	"fontSizes": [
		{
			"name": "Tiny",
			"size": "0.64rem",
			"slug": "tiny"
		},
		{
			"name": "Small",
			"size": "0.8rem",
			"slug": "small"
		},
		{
			"name": "Normal",
			"size": "1rem",
			"slug": "normal"
		},
		{
			"name": "Large",
			"size": "1.25rem",
			"slug": "large"
		},
		{
			"name": "Extra Large",
			"size": "1.563rem",
			"slug": "extra-large"
		},
		{
			"name": "Huge",
			"size": "1.953rem",
			"slug": "huge"
		},
		{
			"name": "Gigantic",
			"size": "2.441rem",
			"slug": "gigantic"
		}
	]
    }
}
}

### PR DESCRIPTION
**Pull request by madhusudhand**

* update font sizes from px to rem to support a11y
* introduce `normal` and `huge` preset sizes to support a11y for inserted patterns ( some patterns using variables such as `--wp-preset-font-size-normal` which are in px )

#### Screenshots:

Browser font setting: Medium (default)

image

Browser font setting: Very Large

image

#### Related issue(s):

Fixes: #5659

### DIFF
--- a/blockbase/theme.json
+++ b/blockbase/theme.json
@@ -173,8 +173,9 @@
 				}
 			],
 			"fontSizes": {
-				"x-small": "14px",
-				"normal": "18px"
+				"x-small": "0.875rem",
+				"normal": "1.125rem",
+				"huge": "3rem"
 			},
 			"form": {
 				"padding": "calc( 0.5 * var(--wp--custom--gap--horizontal) )",
@@ -375,22 +376,22 @@
 			"fontSizes": [
 				{
 					"name": "Small",
-					"size": "16px",
+					"size": "1rem",
 					"slug": "small"
 				},
 				{
 					"name": "Medium",
-					"size": "24px",
+					"size": "1.5rem",
 					"slug": "medium"
 				},
 				{
 					"name": "Large",
-					"size": "28px",
+					"size": "1.75rem",
 					"slug": "large"
 				},
 				{
 					"name": "Extra Large",
-					"size": "32px",
+					"size": "2rem",
 					"slug": "x-large"
 				}
 			]
@@ -545,7 +546,7 @@
 					"fontFamily": "var(--wp--preset--font-family--heading-font)",
 					"fontWeight": "var(--wp--custom--heading--typography--font-weight)",
 					"lineHeight": "var(--wp--custom--heading--typography--line-height)",
-					"fontSize": "48px"
+					"fontSize": "var(--wp--custom--font-sizes--huge)"
 				},
 				"spacing": {
 					"margin": {

### CODE REVIEW

File: blockbase/theme.json [hunk 1]
```diff
@@ -173,8 +173,9 @@
 				}
 			],
 			"fontSizes": {
-				"x-small": "14px",
-				"normal": "18px"
```
madhusudhand, line 177: `custom-normal` which is used as default font, is more than `preset-normal` and it sets different fonts when patterns inserted. Is it a good idea to use the preset-normal with 18px instead?

pbking: I think we only want to keep the --wp--preset--font-size's to only the four Gutenberg provides (small, medium, large, x-large), due to the problems when a user sets content to a font size that might be undefined in other themes.  Better to use an explicit size in those circumstances (such as the font-size for the H1).  Thus there should be no --wp--preset--normal.

The sizes look good, I notice no changes in the rendered page.

jffng: +1 to @pbking 's feedback. I think we want to drop `normal` and `huge` from the `fontSizes` array. That way the font size picker will look like this:

Screen Shot 2022-03-25 at 11 06 21 AM

On this PR it looks like this:

Screen Shot 2022-03-25 at 10 57 20 AM

madhusudhand: Agreed. Removed normal and huge fonts from preset sizes.

jffng (approved):

PR #5754, opened 2022-03-25
Change how 'old custom theme colors' are cleaned out to be more specific

### LINKED ISSUE #5740: FSE Beta: Custom palette colors go away when saving a change in the Customizer
### Quick summary

The custom palette disappears when saving a change in the Customizer

### Steps to reproduce

1. Set a custom color inside FSE using the **"Styles" > Colors > Palette > Custom** option
2. Assign the color
3. Save the changes
4. Preview the site. All good!
5. Go to the Customizer and make a change, like the font, and save changes
6. Notice the custom color is gone

https://user-images.githubusercontent.com/3696121/149391819-b0dd8897-caf5-49f3-9749-b9c39041fa07.mp4

For the custom colors to be preserved.

### What actually happened

The custom colors were lost.

### Context

Dogfooding

### Simple, Atomic or both?

Simple

### Theme-specific issue?

I tested using the Zoologist theme

macOS Monterey 12.1 (21C52)
Chrome Version 97.0.4692.71 (Official Build) (x86_64)

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

_No response_

### Workaround details

_No response_

pbking:

This change only deletes from the 'custom' color space those color objects which have the exact same slugs as those in 'theme'.  This removes any colors that might have been previously stored in this space by this tool, but leaves any other colors a user might have added in the FSE.

To test add a custom color using the FSE:

image

Customize something in the customizer.  (I edited colors)

Return to the FSE (or block editor) and see that the customized color is still available.

#### Related issue(s):

Fixes: #5740

### DIFF
--- a/blockbase/inc/customizer/wp-customize-colors.php
+++ b/blockbase/inc/customizer/wp-customize-colors.php
@@ -152,7 +152,18 @@ class GlobalStylesColorCustomizer {
 		$new_settings['color']['palette']['theme'] = $this->user_color_palette;
 		// We used to set the values in 'custom' but moved to 'theme' to mirror GS functionality.
 		// This ensures that new saves don't store the customizations in both places.
-		unset($new_settings['color']['palette']['custom']);
+		if($new_settings['color']['palette']['custom']) {
+			foreach ( $new_settings['color']['palette']['theme'] as $theme_color ) {
+				foreach ( $new_settings['color']['palette']['custom'] as $key => $custom_color ) {
+					if( $theme_color['slug'] === $custom_color['slug'] ) {
+						unset( $new_settings['color']['palette']['custom'][$key] );
+					}
+				}
+			}
+			if ( ! $new_settings['color']['palette']['custom'] ) {
+				unset ( $new_settings['color']['palette']['custom'] );
+			}
+		}
 
 		// Add the updated global styles to the update request
 		$update_request = new WP_REST_Request( 'PUT', '/wp/v2/global-styles/' );


jffng (approved):
I gave this a spin and it indeed fixes the issue, nice work there.

What is strange to me about the current state — a user can change their colors in the full site editor, and then go back to the customizer and make changes there too. If the color has been changed in the site editor, the customizer selection won't be respected. If it hasn't, then the customizer selection will be respected.

The customizer is still useful for font selection (for now), but I wonder — should we be hiding the colors panel if the user has access to the site editor?

In any case, I _think_ this one is good to bring in, though I may be missing other customization scenarios to test for.

pbking:
> What is strange to me about the current state — a user can change their colors in the full site editor, and then go back to the customizer and make changes there too. If the color has been changed in the site editor, the customizer selection won't be respected. If it hasn't, then the customizer selection will be respected.

I think I'll take another look at this.  If what you describe above is true then we should keep working on this I think.  While it's not ideal that the colors be editable in two places in two ways I would still think/hope/expect that we can allow users to do so cleanly.

pbking:
From what I can tell the only selection that isn't honored is the "featured" selection a user might have chosen.  Going back to the customizer it will always show whatever you selected.  This is an issue even if you change the color with the customizer too, so maybe that isn't what you were talking about.

image

image

If I change colors in the Customizer I see that color reflected in the site itself as well as the FSE.  If I change the same color using the FSE then I likewise see that color represented in the site and Customizer.  From what I can tell a user can edit the color of the semantic colors using either tool (Customizer, FSE) and the result is the same.  And (now) a user can create a custom color using the FSE and if they use the Customizer again it sticks around.

If there are steps that bring us to a state other than this please let me know.

jffng:
Here's what I tried:

1. Activate Blockbase, reset my site's colors to default from the site editor:
Screen Shot 2022-03-28 at 3 37 25 PM

2. Add a custom color to the palette, set the background color to use this custom color:
Screen Shot 2022-03-28 at 3 39 17 PM

3. Go to the customizer, see that the background color set in the site editor is respected but not displayed in the color picker: 
Screen Shot 2022-03-28 at 3 38 55 PM
Screen Shot 2022-03-28 at 3 39 50 PM

pbking:
Ah, I see.

The disconnect is in the linked relationship between our expected, variable colors (such as --wp--preset--background) and how they are used as the background color for body.

It's configured in blockbase themes that the block background color is set to be the --wp--preset--background variable. Changing the "Background" color as you describe changed it from using the variable (that we can share elsewhere i.e. as the text color for buttons).  That's akin to changing the background color of a group block from "primary" to "blue". 

To my understanding the system is working like it should, though it's certainly a bit confusing.  Better would be to have access to these value as variables, but that concept has yet to materialize.

jffng:
> The disconnect is in the linked relationship between our expected, variable colors (such as --wp--preset--background) and how they are used as the background color for body. It's configured in blockbase themes that the block background color is set to be the --wp--preset--background variable.

Ah right! I forgot that's how it worked. It is confusing, but I agree this is the best we can do for now if we aim to support color configuration in both contexts.

Issue #5745, opened 2022-03-24
Pendant: Templates: About Page


pbking:
This page will consist of four patterns: 

<img width="326" alt="image" src="https://user-images.githubusercontent.com/146530/159976951-969ed660-84f0-499e-9a9a-d86142954d7c.png">
 
* Media & Text
* Image and Quote
* Team 

Once the design of the page and contained blocks is approved we'll extract the patterns from it.

beafialho:
@pbking I also added a mobile [mockup](https://www.figma.com/file/JdntsOfLqo8Vy0d5uTWHTI/Pendant?node-id=1597%3A216) for the About page.

PR #5759, opened 2022-03-26
Added pendant/templates/page-about.html to meet the design suggested …

PeterNdomano:
Added and edited pendant/templates/page-about.html to meet the design suggested in #5745 

### Related issue(s):
#5745 

### About page look after applying above changes
Screenshot 2022-03-27 at 00-39-01 About My Sandbox – My WP-Sandbox

@pbking

### DIFF
--- a/pendant/templates/page-about.html
+++ b/pendant/templates/page-about.html
@@ -1,0 +1,181 @@
+
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
+
+<!-- wp:columns {"layout":{"inherit":true},"align":"full","backgroundColor":"foreground"} -->
+<div class="wp-block-columns alignwide has-foreground-background-color has-background">
+    <!-- wp:column {"width":"33.33%"} -->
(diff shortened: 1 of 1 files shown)


pbking:
A solid start.

Check out how the layout is structured in [this continuation](https://github.com/Automattic/themes/pull/5784/files#diff-54b414dec30b7ac3c112b5e9b8ea64595734b4441a50878ec3c82c2c419f8b62) of your other pull request.  We can achieve edge-less (when needed) layout elements to make sure the gaps are only where we want them to be.

Also, while I work this project I'm testing the workflow of editing the templates with the FSE and "saving" the work using the [Create Block Theme plugin](https://github.com/Automattic/create-block-theme/pull/51).

If you're able to evaluate that while working on this (or other) themes your input would be appreciated.

PeterNdomano:
Thanks @pbking  I had a look and understood what you meant.

Also previously I wasn't using 'Create Block Theme' plugin but I'll start using it as soon as possible so expect my contribution on it too.

PR #5781, opened 2022-03-28
Remote: fix image with text path

### LINKED ISSUE #5778: [Remote] Image with text pattern has a broken image
The image for the "Image with text" pattern appears to be broken. When inserted, it shows just an image block placeholder.

Screen Shot 2022-03-28 at 10 51 17 AM

(As seen while creating the demo site on WP.com)

jffng:

This PR changes the filename of the image reference, making the first letter uppercase. I didn't realize this would have any impact, but it fixes the issue on dotcom for me.

#### Related issue(s):

Fixes #5778


pbking (approved):
Silly case sensitivity.

👍

Issue #5674, opened 2022-03-16
Pendant: Templates: 404


pbking:
<img width="649" alt="image" src="https://user-images.githubusercontent.com/146530/159967648-9fb324fd-1468-4c82-bfff-d50278a44708.png">

Modify 404 template to match the above.

Make the "There's nothing here" a variable sized font (not a registered size) so that it scales to small screens well. 

This should also resolve any changes needed to the Search Block.

No need to make this a pattern (yet).

PR #5786, opened 2022-03-29
Pendant: add 404 template

madhusudhand:

Without a letter spacing, the letters `t` and `h` are getting merged.

image

With at letter spacing of `1px` content with `130px` font size doesn't fit inside 1240px container.

image

**Other issues:**

* Search box has padding between border and search button. seems like overrides aren't working. need suggestions here.
* Used H1 and H4 for the text. Are these semantics right?

#### Related issue(s):

May address: #5674

### DIFF
--- a/pendant/style.css
+++ b/pendant/style.css
@@ -157,3 +157,7 @@ body > .is-root-container > .wp-block-template-part > .wp-block-cover,
 	margin-right: auto !important;
 	width: inherit;
 }
+
+.custom-404-wp-search {
+	max-width: 500px;
+}
--- a/pendant/style.css
+++ b/pendant/style.css
@@ -72,7 +72,7 @@ a:focus {
 	background-color: var(--wp--preset--color--background);
 	color: var(--wp--preset--color--foreground);
 	border: 2px solid var(--wp--preset--color--foreground);
-    	padding: 0.667em 1.333em;
+    	padding: 0.667em 1.333em !important;
 }
 
 .wp-block-button.is-style-outline .wp-block-button__link:hover {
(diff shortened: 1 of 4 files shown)


MaggieCabrera:
The design doesn't have letter spacing, I believe these are ligatures that come from the font. Looks like there's a CSS rule to disable them: `font-variant-ligatures: none;` but I don't think they look bad. I'd defer to the designers on this, but we can go ahead with the ligatures as they show and merge this, they will tell us after all templates are done wether they prefer it with or without.

About the text, we've used h2 in the past (I'm checking Archeo). The smaller text can be a paragraph, if the font size needs to be bigger you can change it for that specific block.

For the search block you are definitely going to need CSS, that's a known limitation of the search block, but my suggestion is that we go without and add the CSS once the designers review everything should they deem that CSS necessary.

pendant/templates/404.html
```diff
@@ -1,12 +1,16 @@
 <!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
-<!-- wp:group {"tagName":"main","layout":{"inherit":true}} -->
-<main class="wp-block-group">
+<!-- wp:group {"tagName":"main","style":{"spacing":{"padding":{"top":"72px","bottom":"128px"}}},"layout":{"inherit":true}} -->
+<main class="wp-block-group" style="padding-top:72px;padding-bottom:128px"><!-- wp:heading {"textAlign":"left","level":1,"align":"wide","style":{"typography":{"fontSize":"130px"}}} -->
+<h1 class="alignwide has-text-align-left" id="oops-that-page-can-t-be-found" style="font-size:130px">There's nothing here.</h1>
```
MaggieCabrera: We need this translated. The way we do this is via block patterns. Check how Archeo does it. For pendant we already have a block pattern created that you can leverage

pendant/templates/404.html
```diff
@@ -1,12 +1,16 @@
...
+<!-- /wp:heading -->
 
-	<!-- wp:pattern {"slug":"pendant/404"} /-->
```
MaggieCabrera, line 6: This was the call to the block pattern FYI

madhusudhand:
Latest UI

image

image

pendant/style.css
```diff
@@ -112,3 +112,7 @@ body > .is-root-container > .wp-block-template-part > .wp-block-cover,
 	margin-right: auto !important;
 	width: inherit;
 }
+
+.custom-404-wp-search {
+	max-width: 500px;
+}
```
madhusudhand, line 173: is there a way to set max-width: 500px without a custom class?
small screens should have full width, and large screens upto 500px.

MaggieCabrera: Maybe you can put the search block inside a column, you can control the width of those

pbking: This: `"width":500,"widthUnit":"px"` on the navigation block seems to have done the trick to keep it at 500px.

pbking:
Made a few small changes and will bring this in now.

Returned the heading to H1 instead of H2 and copied the sizing from the template to the H1 configuration.  Another recent PR upgraded the sizing of headings and this change seems to fit will in that direction.

Also made a couple of minor tweaks to the search box CSS and block attributes to achieve the design.  Closer to the goal; design can review further.

image

PR #5792, opened 2022-03-30
Blockbase: add theme support for wp-block-styles

### LINKED ISSUE #5642: Blockbase: There is no default vertical padding for a Group block with a background colour
## Steps to replicate

* Add a Group block with a background colour
* Add an image and a paragraph

Example:
```
<!-- wp:group {"backgroundColor":"cyan-bluish-gray"} -->
<div class="wp-block-group has-cyan-bluish-gray-background-color has-background"><!-- wp:image {"id":12502,"sizeSlug":"large","linkDestination":"none"} -->
<figure class="wp-block-image size-large"><img src="https://iamtakashi.files.wordpress.com/2021/10/tom-crew-t9srlyizfiq-unsplash.jpg?w=1024" alt="" class="wp-image-12502"/></figure>
<!-- /wp:image -->

<!-- wp:paragraph -->
<!-- /wp:paragraph --></div>
<!-- /wp:group -->
```
## Result

There is no default vertical padding in the Group block, so the image and the paragraph bump the Group block.

**Editor:**
Screenshot 2022-03-09 at 15 08 44

**Front of the site:**
Screenshot 2022-03-09 at 15 08 36

## Expected

I expected there would have been equal default paddings for all the sides. Or no padding at all for all the sides—one could argue that there should be no padding unless it's specified in the padding control in the sidebar.

This is what TT2 does. It's still strange that the side paddings are so large, but I suppose that's the theme style.

 Screenshot 2022-03-09 at 15 15 57

mikachan:
This PR adds theme support for `wp-block-styles` to Blockbase.

This is mainly to address https://github.com/Automattic/themes/issues/5642, although I'm not sure why we're not already using this in Blockbase, so not sure if this introduces other problems! However, from looking at the [CSS it includes](https://developer.wordpress.org/block-editor/how-to-guides/themes/theme-support/#default-block-styles) when enabled, it sounds like something we'd want to support. We also use it in a couple of other block themes.

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/5642

### DIFF
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -23,6 +23,9 @@ if ( ! function_exists( 'blockbase_support' ) ) :
 		// Experimental support for adding blocks inside nav menus
 		add_theme_support( 'block-nav-menus' );
 
+		// Add support for block styles.
+		add_theme_support( 'wp-block-styles' );
+
 		// Enqueue editor styles.
 		add_editor_style(
 			array(


MaggieCabrera:
I tried to find a conversation about this but I couldn't, so I can't remember if there's any reason why we shouldn't add this. I think the related issue it solves is valid, but I don't know if all the styles that come from that stylesheet are 100% compatible with global styles. I remember the quote/pullquote blocks adding some borders and stuff that couldn't be removed via the editor, don't know if that's the case any longer. @scruffian do you recall?

pbking:
Same in that I also recall that there was discussion, though not long ago I thought that Blockbase DID enable that theme support.  I mis-remembered how we decided to move forward.  I think that the styles for the table were also a reason to use/not-use the "block-styles" optional styles.

I do think it was a matter of it adding selectors that weren't (easily? possible?) override-able.  I don't know if that is still the case.

It's not clear to me why we decided what we decided and what harm it would bring to use (or continue not to use) that flag.

mikachan:
Thanks both! I can spend some time doing some thorough testing with these styles enabled, and maybe also try fixing the vertical padding issue with Blockbase-specific CSS as an alternative.

scruffian:
I would rather not do this. The styles that this opts in to aren't great. IMO it would be better to have a more targeted fix.

pbking:
I think we can close this.  It adds quite a fair bit of CSS and is a decision I think best made at the beginning of a theme's development.

Adding the padding explicitly (#5811)  works better for Blockbase anyway.

mikachan:
Agreed! Thanks for your help with this.

PR #5793, opened 2022-03-30
Blockbase: add select boxes to form styles

### LINKED ISSUE #5651: Blockbase: Border for Select block is different from other blocks for Form block.
## Steps to replicate

* Add a Form block.
* Also add a Select block in there
* See the border style

## Result
The border style for the Select block is different from other blocks and it looked strange.

## Expected
The border style is expected to be the same as others

**Editor**
Screenshot 2022-03-10 at 17 28 55

**Front of the site**
Screenshot 2022-03-10 at 17 28 14

With Twenty Twenty Two, it appeared like this.
Screenshot 2022-03-10 at 17 20 08

mikachan:
This adds `select` boxes to the existing form input styles for Blockbase, which brings the styling for any select boxes in line with other form elements.

Before|After
------|------
image|image

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/5651

### DIFF
--- a/blockbase/sass/elements/_forms.scss
+++ b/blockbase/sass/elements/_forms.scss
@@ -14,7 +14,8 @@ input[type="time"],
 input[type="datetime"],
 input[type="datetime-local"],
 input[type="color"],
-textarea {
+textarea,
+select {
 	background: var(--wp--custom--form--color--background);
 	border: var(--wp--custom--form--border--width) var(--wp--custom--form--border--style) var(--wp--custom--form--border--color);
 	border-radius: var(--wp--custom--form--border--radius);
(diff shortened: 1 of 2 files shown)


blockbase/sass/elements/_forms.scss
```diff
@@ -26,16 +27,17 @@ textarea {
 	&:focus {
 		border-color: var(--custom--form--color--border);
 		color: var(--wp--custom--form--color--text);
-		outline: 1px dotted currentColor;
+		outline: 1px dotted currentcolor;
```
jffng, line 30: Is the `currentColor` variable case sensitive?

jffng (approved):
This LGTM, left one small comment:

Before | After
--- | ---
Screen Shot 2022-03-30 at 10 34 27 AM | Screen Shot 2022-03-30 at 10 34 42 AM

blockbase/sass/elements/_forms.scss
mikachan: I don't think it is based on the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#color_keywords). If we want to maintain our current casing we should probably add a new lint rule for this..

PR #5782, opened 2022-03-28
Remote: post title adjustments

jffng:
#### Changes proposed in this Pull Request:

- Removes links from post titles.
- Sets line height for post titles in the post lists pattern.
- Adjusts post title font sizes in the single, page, and 404 to match the hero text font size

#### Related issue(s):

Closes #5773 #5774 #5775

PR #5800, opened 2022-03-31
Styled navigation (open overlay)

pbking:
This adds styling for navigation overlay menu.

The overlayBackgroundColor and overlayTextColor only seem to be available for non-default menu.  The controls on the FSE UI aren't available for a menu that hasn't been set so were set on template blocks manually.

The rest of the styling is custom CSS to apply a different style to sub-menu items.

This is (at best) inspired from the figma design.  Unavailable to execute from the original design:

image

* The site title can't be added.
* The close X defaults to the right side.  (This could be moved to the left side with CSS).
* The search block is unavailable for the overlay menu.
* The top menu items (with sub-menu items) cannot be opened/closed thus there is no ^ type of indicator or underline.

image

The actual implementation looks like this:

image

Fixes #5782


beafialho:
This all looks like a good alternative. As noted live with @pbking, submenu items should be **Medium** weight and it was adjusted.

pbking:
Brought in these changes and I think this covers the MOBILE design, however the desktop still uses the overlay navigation and needs further design work.  Will address that in another item.

beafialho:
@pbking I added mockups of what the desktop menu should look like to the [Figma file](https://www.figma.com/file/JdntsOfLqo8Vy0d5uTWHTI/Pendant?node-id=1610%3A2).

pbking:
Thank you @beafialho I have opened an issue to address that [here](https://github.com/Automattic/themes/issues/5818).

Issue #5794, opened 2022-03-31
Pendant: use rem unit for fonts instead of px for better a11y


madhusudhand:
For better a11y, use `rem` units for fonts in `theme.json`.

Steps:
* In browser settings update font size to small or large
* observe that page content doesn't get effected with this setting when `px` units used.

MaggieCabrera:
We could probably do this for Archeo, Livro, Stewart and Remote too.

pbking:
Thank you @madhusudhand good call.

This was addressed with other changes to font sizes [here](https://github.com/Automattic/themes/commit/7aaeda4aafa2d326067548e9597529f0dbb4734b).

madhusudhand:
> We could probably do this for Archeo, Livro, Stewart and Remote too.

Created https://github.com/Automattic/themes/pull/5801 for other themes.

PR #5801, opened 2022-04-01
multiple themes: update font sizes from px to rem for better a11y

madhusudhand:

Updated font sizes from `px` to relative `rem`s for better a11y

#### Related issue(s):

#5794

### DIFF
--- a/geologist-blue/theme.json
+++ b/geologist-blue/theme.json
@@ -115,8 +115,8 @@
 				"padding": "20px"
 			},
 			"fontSizes": {
-				"x-small": "16px",
-				"normal": "20px"
+				"x-small": "1rem",
+				"normal": "1.25rem"
 			},
 			"body": {
 				"typography": {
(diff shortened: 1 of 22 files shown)


geologist-blue/theme.json
```diff
@@ -261,7 +261,7 @@
 					}
 				},
 				"typography": {
-					"fontSize": "25px",
+					"fontSize": "1.5625rem",
```
pbking, line 264: I think we can limit the decimal place a bit.  `1.56` is probably as far as we need to go with `25px`

pbking:
The variations you don't need to necessarily change in this PR (to save you some time and make the PR easier to review).  You can just change Geologist and Quadrat, etc and then the "variations" are built during the deployment process (or manually using `npm run build:variations`.  There aren't any font size definition overrides in those [variations](https://github.com/Automattic/themes/tree/trunk/variations) so you just need to change the parents.

[edit] That said a project/wide find/replace would change those too wouldn't they?  And since there's no harm I'm not sure why I even raised it as an issue.

This looks fine to ship to me.

geologist-blue/theme.json
pbking: Poking around the repo I found evidence of rem values every bit as specific as those.  I say never mind, I'm good shipping with numbers that specific. ¯\_(ツ)_/¯

pbking (approved):

PR #5806, opened 2022-04-01
Add/pendant about page patterns

### LINKED ISSUE #5745: Pendant: Templates: About Page

pbking:

* First I created a new "front-page" template using the FSE and constructed the elements in the design.
* Next I "saved" that template (with the Create Block Theme plugin) to export the "front-page.html" markup.
* Then I copy/pasted the markup into individual pattern.php files.
* Lastly I localized the (non-name) text strings and pointed to theme provided images.

Additional changes:

Usage of the Block Gap has been removed (via CSS) from all top-level items on the page.  This was to allow for the design where the 'About' pattern at the top of a page (or a post using the "Header and Footer Only" template) will have no gap between the header and the body.

The line-height of headings was increased slightly to achieve this design without customizing the block instance.

image

Body text (medium size) was configured to scale (slightly) between 1rem and 1.25rem (1.25 being the original static value).

The `core/quote` block had typography settings set in theme.json as well as the quote's `cite` element styled via style.css.

This resulted in three new patterns:

image

Then I created a new page and dropped all of these new patterns onto it for a finished result of this:

Desktop
localhost_4759__page_id=1768

Mobile

#### Related issue(s):

Fixes: #5745


pbking:
Looking for a technical review of these bits to merge.  Design approval can come later.

jffng (approved):
This LGTM. Pushed a commit to escape the remaining strings + linting fixes.

I assume we'll address the categorization of the patterns, along with other small visual tweaks, in the design review. Thanks!

pbking:
Since (some of them) were names I didn't think they need to (or should be) escaped.  I missed the job string for sure though.  I don't see a lick of harm in it though, thanks for cleaning that up sir.

jffng:
> Since (some of them) were names I didn't think they need to (or should be) escaped.

Yeah, it's a bit strange but I think the idea is that every theme-provided string needs to be able to be localized, at least for the dotorg review this has caught us in the past.

pbking:
Made a few design changes based on the updated figma.  Bringing this in and will tweak any remaining issues in a later pass.

PR #5811, opened 2022-04-04
Blockbase: add padding for group blocks with a background

### LINKED ISSUE #5642: Blockbase: There is no default vertical padding for a Group block with a background colour

mikachan:
This adds vertical and horizontal padding to group blocks with a background, similar to the styles provided in `wp-block-styles`.

This is an alternative to https://github.com/Automattic/themes/pull/5792.

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/5642

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -757,13 +757,6 @@ p.has-drop-cap:not(:focus)::first-letter {
   margin-bottom: var(--wp--custom--gap--baseline);
 }
 
-.wp-block-post-content p.wp-block.wp-block-paragraph,
-.wp-block-post-content *[class^=wp-container] > * + p,
-.wp-block-post-content *[class^=wp-container] > p + *,
-.wp-block-post-content p {
-  margin-top: 1em;
-}
-
 .wp-block-post-template {
   margin-top: 0;
   margin-bottom: 0;
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -117,6 +117,11 @@ body > .is-root-container > .wp-block-template-part > .wp-block-cover,
   width: inherit;
 }
 
+/* Spacing for group blocks with a background color. */
+.wp-block-group.has-background {
+  padding: var(--wp--custom--gap--vertical) var(--wp--custom--gap--horizontal);
+}
+
 .has-primary-background-color {
   background-color: var(--wp--custom--color--primary) !important;
 }
--- a/blockbase/sass/base/_alignment.scss
+++ b/blockbase/sass/base/_alignment.scss
@@ -39,9 +39,9 @@ body > .is-root-container,
 .wp-block-group.has-background,
 .wp-block-columns.alignfull.has-background,
 .wp-block-cover.alignfull,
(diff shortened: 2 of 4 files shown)


pbking:
This seems to introduce additional padding for paragraph elements.

image

The 1em; margin-top was always there, but with the padding introduced it's now taken into account.

(trunk)
image

Chrome is reporting that the padding is coming from _pullquote.scss however it instead seems to be coming from _post-content.scss.

Is this still how we want to be handling paragraph gaps in blockbase?

pbking:
I removed the styles responsible for that opinionated padding.  What do you think @mikachan?

image

It was introduced [here](https://github.com/Automattic/themes/pull/5104) but I think it's causing more [problems](https://github.com/Automattic/themes/issues/5617) that it fixes.

mikachan:
I really like the idea of removing those styles. I've noticed this causes the spacing to be different between a heading and paragraph inside and outside a group block:

With opinionated styles:
image

Without opinionated styles:
image

However, this seems similar to how TT2 behaves:

image

Also, it looks like the [original issue mentioned here](https://github.com/Automattic/themes/pull/5104) now doesn't need these styles anyway. I can't see a margin above the second paragraph with the following markup:

```
<!-- wp:group {"layout":{"type":"flex","justifyContent":"center"}} -->
<div class="wp-block-group"><!-- wp:paragraph {"align":"center"} -->
<p class="has-text-align-center">Paragraph 1</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"align":"center"} -->
<p class="has-text-align-center">Paragraph 2</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group -->
```

image

So my vote is to remove these styles, especially as this will hopefully fix other issues too!

pbking (approved):
👍 
We've had some discussion.
Removing the custom <p> margins and adding the custom group.has-background padding seems like the win.

iamtakashi:
Thank you both for fixing the issues around this. The paragraph block has been weird for some time with Blockbase. Thanks again!

PR #5812, opened 2022-04-04
Adjusted spacing on single per design review

pbking:
This changes the spacing on the Pendant 'single' template from VW units to pixels as it suits the design better.

Used different spacing for heading (as opposed to just using block gap value) because of font spacing funkyness.
Before
image
image
image
image

After
image
image
image
image


pbking:
This was changed live with @beafialho so will go ahead and bring this one in.

PR #5816, opened 2022-04-05
Blockbase: Account for additional content via css grid configuration

### LINKED ISSUE #5622: Quadrat: Post Comments Block not aligning and cutting off footer on AT sites
### Quick summary

Adding the Post Comments block to the Single Posts template in Full Site Editor when in use with the Quadrat (black) theme, the block does not stay in alignment with the rest of the content (margin missing on both sides regardless of block alignment or if the block is in a group or columns block) and it seems to cause the rest of the page below it to be cut off and you can view the footer of the site and it shows the comment field cut off:

### Steps to reproduce

1. On a site with the Quadrat (black) theme (and there may be other themes) go to the Full Site Editor
2. Edit the Single Post (or possibly any page) template
3. Add the Post Comments block to the bottom of the post template
4. View a blog post on the front end

Alignment of the block should not overlap the footer

### What actually happened

Cuts off the footer and does not align with the content on the rest of the page (left aligned)

### Context

#34427726-hc

### Simple, Atomic or both?

Atomic

### Theme-specific issue?

Quadrat (black)

_No response_

### Reproducibility

Consistent

### Severity

One

### Available workarounds?

Yes, easy to implement

### Workaround details

This was the workaround we came up with as a CSS rule:

```
.wp-block-post-comments {
    margin-bottom: 50px;
    margin-left: auto !important;
    margin-right: auto !important;
}

#commentform iframe {
    width: 664px !important;
}
```

pbking:
ensure that any immediate children that are unaccounted for in the grid area definitions span both columns.

This is necessary because wp simple and wp atomic implement the comment section override differently.

Fixes: #5622

Checked in wpcom (simple) wpcom (automic) wporg (authenticated) wporg (unauthenticated)

Atomic:
image

Simple:
image

wporg (authenticated)
image

wporg (unauthenticated)
image

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -659,25 +659,25 @@ p.has-drop-cap:not(:focus)::first-letter {
   margin-top: 0;
   margin-bottom: var(--wp--custom--gap--vertical);
 }
-.wp-block-post-comments form .comment-notes {
+.wp-block-post-comments form > .comment-notes {
   grid-area: notes;
 }
-.wp-block-post-comments form .comment-form-author {
+.wp-block-post-comments form > .comment-form-author {
   grid-area: author;
 }
-.wp-block-post-comments form .comment-form-email {
+.wp-block-post-comments form > .comment-form-email {
   grid-area: email;
 }
-.wp-block-post-comments form .comment-form-url {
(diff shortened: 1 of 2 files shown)


MaggieCabrera (approved):
LGTM!

PR #5820, opened 2022-04-05
Patternized the pendant header

pbking:
This moves the pendant header into a pattern.

Nothing more; nothing less.


mikachan (approved):
Looks good!

PR #5821, opened 2022-04-05
Moved index content to patterns

pbking:
Just takes the markup from index.html template and moved it to (2) patterns.


mikachan (approved):
LGTM

PR #5822, opened 2022-04-05
Add Front Page template (and patterns)

### LINKED ISSUE #5807: Pendant: Front Page Template
## Task

Create the design expressed in the figma file in the `front-page.html` template.  

image

image

image

If possible use the FSE and 'save' the templates and any block settings using the Create Block Theme plugin.

Break apart the page into block pattern files with instances referenced in the front page template.

The "latest Articles" can be made a pattern from the index.html page template.  There will be 5 additional patterns: "Featured", "Our Picks", "Latest Interviews", "Subscribe" and "Instagram" 

Localize the strings in the patterns and point to images shipping with the theme.

pbking:
WIP

Add the template and relevant patterns for the Front Page template.

#### Related issue(s):

Fixes: #5807 

Currently have used the FSE to make a first pass at most of the patterns.  Still a WIP.


pbking:
The design calls for a "get updates" pattern.  However I believe that that functionality is only available via a plugin (JetPack).  Can that pattern be included in the theme?

image

pbking:
For the content at the top of the page I'm not sure how to align the text content with regular content...  

Using a 'Media & Text' block I don't seem to have the option to control the padding to align with the site... The block adds some padding that I don't have control of (outside of assigning a custom CSS class).

image

kjellr:
I think we had a similar problem with Archeo, and solved it by using a wide Columns block instead of a Media & Text.

mikachan:
Yes, for Archeo we switched the [Image with headline and description](https://github.com/Automattic/themes/blob/trunk/archeo/patterns/image-with-headline-description.php) pattern to Columns instead of Media & Text.

There's also the [Layered images with headline](https://github.com/Automattic/themes/blob/trunk/archeo/patterns/layered-images-with-headline.php) pattern that might work similarly to this one.

pbking:
Thanks everyone who helped me get this pattern working!  I think what's committed fits the bill! I think I used a little bit of every suggestion... :)

jffng:
Looking great. I pushed a commit that escaped all the strings, made the image URL paths dynamic, and added a query to the front page template per the design.

MaggieCabrera:
Why are we using front-page.html and not home.html like we've been doing for the rest of our themes?

jffng:
> Why are we using front-page.html and not home.html like we've been doing for the rest of our themes?

I don't know for sure, but I have a couple ideas:

1. It's called the Front Page template in the figma so we're following that.
2. If a user sets a static page for their home page, we don't want this template used for the blog posts page. 

Though I just tried setting a different static home page via Settings > Readings, and this template is still used for the home page, which feels like a bug?

kjellr:
> Though I just tried setting a different static home page via Settings > Readings, and this template is still used for the home page, which feels like a bug?

Yeah, that's the problem with the `front-page.html` template: Its placement in the hierarchy is too aggressive. We should use `home.html` here instead.

pbking:
Cool.  These patterns (and thus this page template) is looking really good.  Going to bring this in and made any adjustments needed and outlined by design in further PRs.

Issue #5803, opened 2022-04-01
Multiple themes: update how patterns are loaded


mikachan:
We should update how all patterns are included in our block themes, to allow them to be automatically loaded from Gutenberg, as we did with Archeo here: https://github.com/Automattic/themes/pull/5715

- [x] Remote https://github.com/Automattic/themes/pull/5829
- [x] Livro https://github.com/Automattic/themes/pull/5879
- [x] Stewart https://github.com/Automattic/themes/pull/5854
- [x] Skatepark https://github.com/Automattic/themes/pull/5858

MaggieCabrera:
I added a list of themes so we can use this as an overview issue. Should we do this to Blockbase themes too?

mikachan:
I guess if we include Blockbase themes then we'd maybe need to bump the required WP version for Blockbase to 6.0, as it looks like this is being backported to 6.0 (https://github.com/WordPress/wordpress-develop/pull/2488). We've only just updated the required WP version here https://github.com/Automattic/themes/pull/5823 to 5.8.

Hmm, so I think we're better leaving Blockbase as it is for now, but it would be good if we did these changes to Blockbase themes once WP 6.0 has been released.

matiasbenedetto:
Adding a PR for Stewart https://github.com/Automattic/themes/pull/5854

PR #5829, opened 2022-04-07
Remote: refactor patterns to load directly from the editor

MaggieCabrera:

Solves https://github.com/Automattic/themes/issues/5803 for Remote. Check that all patterns load, both on the inserter and in the template where the pattern block is used.

#### Related issue(s):

https://github.com/Automattic/themes/issues/5803


mikachan (approved):
This looks good to me! I can see the patterns still loading in the inserter and in all the templates.

Issue #5728, opened 2022-03-23
Implement new separator options


MaggieCabrera:
Since https://github.com/WordPress/gutenberg/pull/38428 has been merged, we should review how we handle the separator block in themes like Remote and Skatepark and apply the new changes

matiasbenedetto:
Two PRs to address these changes: https://github.com/Automattic/themes/pull/5938 and https://github.com/Automattic/themes/pull/5940

Issue #5582, opened 2022-02-09
Blockbase/Byrne Theme: Separator Block with "Dots" style appears misaligned in editor


cuemarie:
### Quick summary

Dots style separator displays the 3 dots on 2 lines instead of all parallel in the editor:

<img width="1383" alt="Markup 2022-02-09 at 09 35 53" src="https://user-images.githubusercontent.com/27249804/153265419-7dacca07-d3a7-443b-9f49-aff4c715cff7.png">
 

Dots appear fine on live site.

### Steps to reproduce

1. Activate Theme Byrne on a simple site
2. Add Separator Block to page or post
3. Set "Style" to "Dots"

### What you expected to happen

Dots should appear all in 1 line & parallel, like they do in other themes or the live site.

**Live Site:**

<img width="1225" alt="Screen Shot 2022-02-09 at 11 28 26 AM" src="https://user-images.githubusercontent.com/27249804/153266234-3d5f4032-858d-4414-ac8e-9cf7225e3c2b.png">

### What actually happened

The dots in the editor are misaligned.

<img width="1272" alt="Screen Shot 2022-02-09 at 11 28 55 AM" src="https://user-images.githubusercontent.com/27249804/153266813-b807500a-95d1-4932-923c-09517a777e2b.png">

### Context

Found during team testing

### Simple, Atomic or both?

Simple

### Theme-specific issue?

Byrne

### Browser, operating system and other notes

Was able to reproduce on both FSE-enabled and FSE-disabled site, with Byrne theme

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

There is no user impact

### Workaround details

No problem on live site - just a quirk

Robertght:
I can confirm this happens on AT sites as well.

formosattic:
Was able to reproduce it with Geologist, Quadrat.

It's also happening on self-hosted, I tested with:
* Blockbase: https://pcm.wordpress.org/themes/blockbase/
* Quadrat: https://pcm.wordpress.org/themes/quadrat/

Other reports listed at https://github.com/Automattic/wp-calypso/issues/61853

cc @mikeicode  @metabreak

MaggieCabrera:
This is a blockbase issue with the separator width CSS

mikeicode:
It'd be nice to fix this since it apparently affects all Blockbase child themes. I run into it when I try to use Russell on DIFM-L sites as reported here:

[...]

PR #5834, opened 2022-04-08
Implement new Separator block options

matiasbenedetto:
This PR implements the latest features from the Gutenberg Separator block to the following themes:

- Archeo
- ~~Blockbase~~
- Remote
- Skatepark
- Stewart

#### Related issue(s):
https://github.com/Automattic/themes/issues/5728
https://github.com/Automattic/themes/issues/5582


pbking:
For the moment my 'druthers is to refrain from doing this in Blockbase yet...  I'm increasingly of the opinion that Blockbase should be 5.9 sans-Gutenberg compatible.

MaggieCabrera:
> For the moment my 'druthers is to refrain from doing this in Blockbase yet... I'm increasingly of the opinion that Blockbase should be 5.9 sans-Gutenberg compatible.

I agree that it shouldn't require GB, but why stay in 5.9 only? I think we can hold off this change until the 6.0 release, but we shouldn't stop improving Blockbase to remove things from the ponyfill like it was the original idea. We should just do it when everything is stable and in core.

pbking:
> I think we can hold off this change until the 6.0 release,

Yes, wholeheartedly agree.  I don't mean to stagnate Blockbase at 5.9 but rather to hold off on adding this feature to Blockbase until then.

The other themes I feel less strongly about.  I see benefit to bringing this in for the other themes today.

archeo/theme.json
```diff
@@ -231,7 +231,7 @@
 			},
 			"core/separator": {
 				"color": {
-					"background": "var(--wp--preset--color--foreground)"
+					"text": "var(--wp--preset--color--foreground)"
```
pbking, line 234: I don't believe this should change to 'text'.  I believe 'background' is still the correct attribute to style.

archeo/templates/single.html
```diff
@@ -51,17 +51,23 @@
...
 		</div>
 		<!-- /wp:column -->
 
 		<!-- wp:column -->
-		<div class="wp-block-column"><!-- wp:separator {"className":"is-style-wide"} -->
-		<hr class="wp-block-separator is-style-wide"/>
+		<div class="wp-block-column">
+		
+		<!-- wp:separator {"style":{"color":{"background":"#fcfbf766"}},"className":"is-style-wide"} -->
+		<hr class="wp-block-separator has-text-color has-alpha-channel-opacity has-background is-style-wide" style="background-color:#fcfbf766;color:#fcfbf766"/>
```
pbking, line 68: This isn't one of the palette colors.  Can it be?  I think a couple of these classes aren't needed (has-text-color has-alpha-channel-opacity) and aren't represented on the block attributes anyway.

blockbase/functions.php
```diff
@@ -14,6 +14,9 @@ function blockbase_support() {
 		// Add support for editor styles.
 		add_theme_support( 'editor-styles' );
 
+		// Enqueue the Gutenberg plugin theme.css file
+		add_theme_support( 'wp-block-styles' );
```
pbking: Is the addition of `wp-block-styles` necessary to achieve this change?  It adds a lot of CSS to the theme and I'm hesitant to do that late in the game.  I know we evaluated it elsewhere and I believe we decided not to bring that into Blockbase.

remote/theme.json
```diff
@@ -145,8 +145,8 @@
 				}
 			},
 			"core/separator": {
-				"border": {
-					"color": "var(--wp--preset--color--primary)"
+				"color": {
+					"text": "var(--wp--preset--color--primary)"
```
pbking, line 149: I think this needs to be 'background' not 'text'

pbking:
From what I can tell the block attribute we're shifting to is `color.background` and these changes seem to target something different.

I would much prefer to keep the changes to accomplish this in Blockbase in a separate PR as the changes are potentially broader in scope, addresses different (and more) issues and we may want to time it differently (i.e. WP 6.0 release).  That said it seems to fail gracefully and it might be fine to bring it in before then.  It does fix the "three dots" issue so this is definitely the right fix for that.

I'm not entirely comfortable that this is ready, moving back to in-progress to have another pass.

remote/theme.json
matiasbenedetto: I'm not sure about that, please, try it using the Remote theme as an example. If you use `background` instead of `text` you will not be able to preview the default separator color in the editor.

archeo/theme.json
matiasbenedetto: I don't think so, please see my comments here: 
https://github.com/Automattic/themes/pull/5834#discussion_r853280069
https://github.com/Automattic/themes/pull/5834#discussion_r853212650

archeo/templates/single.html
matiasbenedetto: That is the palette color named `background` (`#FCFBF8`) with the alpha channel set to 40% to preserve the same look. 

If we use `color.text` instead of `color.background` in `theme.json` we will need this change. The rationale:

We need to add that because otherwise, these separators that get their color using `currentColor` become invisible, because the block color set in `theme.json` is `foreground` (kind of black). See how `currentColor` is applied to the block in the Gutenberg codebase: https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/separator/style.scss

`currentColor` only works with `color` (`color.text` in `theme.json`) that's why is impacted by the change in theme.json and why we need to set a color in this separator.

The CSS classes are added by Gutenberg to handle the legacy separator block version. I didn't add markup manually it was just the output of the editor.

matiasbenedetto:
I reverted the Gutenberg styles inclusion in Blockbase to explore that change in a separated PR because that change can have a broader impact. 

The other comments from @pbking were answered in the code snippets and I think are OK. I added rationale to have a more detailed explanation of the changes. I'm moving this PR to "needs review" again.

Thanks, @MaggieCabrera @pbking for the review!

blockbase/functions.php
matiasbenedetto: Yep, it is necessary. But I reverted the Gutenberg styles inclusion in Blockbase to explore that change in a separated PR because, as you said, that change can have a broader impact.

skatepark/inc/patterns/mixed-media-in-container.php
```diff
@@ -69,9 +69,11 @@
 
 	<!-- wp:columns {"align":"wide"} -->
 	<div class="wp-block-columns alignwide"><!-- wp:column -->
-	<div class="wp-block-column"><!-- wp:separator {"color":"primary","className":"is-style-wide"} -->
-	<hr class="wp-block-separator has-text-color has-background has-primary-background-color has-primary-color is-style-wide"/>
-	<!-- /wp:separator --></div>
+	<div class="wp-block-column">
+	
+	<!-- wp:separator {"backgroundColor":"primary","className":"is-style-wide"} -->
```
jffng: I get a block validation error when I try to insert this pattern. The result of clicking "Attempt Block Recovery" is:

```
<!-- wp:separator {"opacity":"css","backgroundColor":"primary","className":"is-style-wide"} -->
<hr class="wp-block-separator has-text-color has-primary-color has-css-opacity has-primary-background-color has-background is-style-wide"/>
<!-- /wp:separator -->
```

jffng (commented):
Thank you for tackling this @matiasbenedetto. In general I think this would be easier to review as separate PRs per theme, especially since there are many changes to visual elements. Also there are conflicts for some reason in Skateparks patterns, any idea what's going on there?

I tested Skatepark and the results are a bit confusing.

For the Home, Search templates, all the separators become slightly thicker. I think it's okay:

Home Before | Home After
--- | ---
Screen Shot 2022-04-22 at 11 14 15 AM | Screen Shot 2022-04-22 at 10 45 00 AM

Search Before | Search After
--- | ---
Screen Shot 2022-04-22 at 10 08 18 AM | Screen Shot 2022-04-22 at 10 08 07 AM

With already existing separator content in patterns, the separators become thicker too:

Before | After
---  | ---
Screen Shot 2022-04-22 at 11 11 35 AM | Screen Shot 2022-04-22 at 11 11 23 AM

I encountered one error inserting the Mixed Media in Container pattern.

matiasbenedetto:
:wave: the conflicts are related to the merge of the PR about moving the templates to a different folder. That is not strictly related to this PR, but anyway, I will create smaller PRs for each theme as @jffng proposes.

matiasbenedetto:
Closed in favor of https://github.com/Automattic/themes/pull/5940

matiasbenedetto:
Follow up for Blockbase: https://github.com/Automattic/themes/pull/5964

PR #5838, opened 2022-04-11
Multiple themes: add term-description to archive template

### LINKED ISSUE #5753: [Theme: Archeo] Missing term descriptions in category archives
### Quick summary

In its category & tag archives, Theme Archeo **does not show term descriptions** by default:

Screen Shot on 2022-03-25 at 13:42:29

Screen Shot on 2022-03-25 at 13:42:08

### Steps to reproduce

1. Create a term and give it a description.
2. Assign this terms to at least one post.
3. Visit the archive page for that post.

I expected to see the term description at the top of the term archive page.

### What actually happened

The term description was not displayed.

### Context

Testing new themes.

### Simple, Atomic or both?

Simple, Atomic

### Theme-specific issue?

I am opening this issue regarding **Archeo**, but we should consider making a point of adding term descriptions to all our themes by default. We had the exact same issue with other themes in the past:

https://github.com/Automattic/themes/issues/2793
https://github.com/Automattic/themes/issues/2809

N/A

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

Yes, difficult to implement

### Workaround details

We can add a term description block manually, but this might be hard to discover for users:

Screen Shot on 2022-03-25 at 13:41:29

I would rather have this term description block added by default to term archives.

madhusudhand:

Add term-description block in archive template in the following themes

### Arbutus

image

### Archeo

image

### Blockbase

image

### Remote

image

### Skatepark

image

### Livro

image

### Pendant

image

### Quadrat

image

### Russel

image

### Stewart

image

### Geologist

image

### Zoologist

image

### Videomaker

image

### Block Canvas

image

### Mayland Blocks

image

### Seedlet Blocks

image

#### Related issue(s):

Fixes: #5753

### DIFF
--- a/arbutus/block-templates/archive.html
+++ b/arbutus/block-templates/archive.html
@@ -2,7 +2,12 @@
 
 <!-- wp:query {"tagName":"main","layout":{"inherit":true}} -->
 <main class="wp-block-query">
-<!-- wp:query-title {"type":"archive","style":{"spacing":{"margin":{"bottom":"100px"}}}} /-->
+<!-- wp:group {"style":{"spacing":{"padding":{"bottom":"50px"}}}} -->
+<div class="wp-block-group" style="padding-bottom:50px">
+<!-- wp:query-title {"type":"archive","style":{"spacing":{"margin":{"bottom":"20px"}}}} /-->
+<!-- wp:term-description /--></div>
+<!-- /wp:group -->
+
 <!-- wp:post-template -->
 <!-- wp:group -->
 <div class="wp-block-group">
--- a/archeo/templates/archive.html
+++ b/archeo/templates/archive.html
@@ -2,7 +2,13 @@
 
 <!-- wp:query {"tagName":"main","layout":{"inherit":true}} -->
 <main class="wp-block-query">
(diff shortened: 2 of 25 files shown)


MaggieCabrera:
I think we need design eyes on this one @kjellr @beafialho

beafialho:
This was not contemplated in the designs to be there by default and, in my opinion, it could be added by users instead if they want to.

MaggieCabrera:
We've added the category descriptions to our themes in the past because classic themes didn't allow the user the possibility for customization, now users can simply add the block themselves to FSE. In my opinion theme designs don't need to include everything all the time.

kjellr:
> This was not contemplated in the designs to be there by default and, in my opinion, it could be added by users instead if they want to.

Yeah, I think it's worth selectively including this block in themes moving forward. But I'd advise against putting it in existing themes. Both for the reason @beafialho shared here, but also because this change would potentially make descriptions appear where they didn't before on live sites. In most cases that would be fine, but I imagine it could be confusing for some users.

madhusudhand:
Closing this base on the above discussion.

PR #5849, opened 2022-04-12
Archeo: Try using button as an element

scruffian:
This is useful for testing https://github.com/WordPress/gutenberg/pull/40260/files

Things to note:
- The CSS for the search block button has been removed
- The styles for buttons have been moved from `styles > blocks > core/button` to `styles > elements > button`.

### DIFF
--- a/archeo/style.css
+++ b/archeo/style.css
@@ -98,6 +98,10 @@ a:active,
 	padding: calc(0.667em + 2px) calc(1.333em + 2px);
 }
 
+.wp-block-search__button {
+	padding: calc(0.667em + 2px) calc(1.333em + 2px);
+}
+
 /*
  * Comment Form Fields
  */
--- a/archeo/style.css
+++ b/archeo/style.css
@@ -99,6 +99,7 @@ a:active,
 
 /*
  * Outline block button needs a padding tweak so it's the same size of normal buttons
+ * https://github.com/WordPress/gutenberg/issues/27476
  */
 .is-style-outline > :where(.wp-block-button__link), 
 :where(.wp-block-button__link).is-style-outline {
--- a/archeo/style.css
+++ b/archeo/style.css
@@ -83,21 +83,6 @@ a:active,
 }
 
 /*
- * File Block button styles.
- * Necessary until the following issues are resolved in Gutenberg:
- * https://github.com/WordPress/gutenberg/issues/27760
- */
-
-.wp-block-file .wp-block-file__button {
-	background-color: var(--wp--preset--color--foreground);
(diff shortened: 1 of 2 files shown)


MaggieCabrera:
We can merge this when 13.4 is live

MaggieCabrera:
There's a bit of a change before and after this PR, I think we may want to adjust it a little bit

Before:

Editor | Frontend 
--- | ---
Screenshot 2022-05-26 at 09 28 29 | Screenshot 2022-05-26 at 09 28 24

After:

Editor | Frontend 
--- | ---
Screenshot 2022-05-26 at 09 26 50 | Screenshot 2022-05-26 at 09 26 56

MaggieCabrera:
@scruffian I added a few changes to make the buttons look like they should. The padding control for the search block button can only be done via the elements API or CSS, since the search block padding is not affecting the button, but the whole block. By assigning a padding to the button element, the outline variation is always bigger than the regular button is, so that required extra CSS to fix. What do you think? If you agree with that, I'll go ahead and replicate this on the rest of the themes.

scruffian:
I'd rather we try to find ways to remove the CSS altogether. Are there changes we can make to Gutenberg so that we don't need it?

MaggieCabrera:
> I'd rather we try to find ways to remove the CSS altogether. Are there changes we can make to Gutenberg so that we don't need it?

if we could target block variations from theme.json... Can't think of anything else

MaggieCabrera:
> > I'd rather we try to find ways to remove the CSS altogether. Are there changes we can make to Gutenberg so that we don't need it?
> 
> if we could target block variations from theme.json... Can't think of anything else

https://github.com/WordPress/gutenberg/issues/27476 I believe would be the issue for this

scruffian:
Ok that looks like it will be a while away yet then, but moving closer :)

For now this seems fine...

MaggieCabrera (approved):
Alright then! I'll approve this and bring this in once dotcom gets the GB change. I'll work on a PR for Blockbase next.

MaggieCabrera:
Ok this one is not done, it looks like we didn't remove the file button CSS since https://github.com/WordPress/gutenberg/pull/41239 was merged, but that has bugs too.

MaggieCabrera:
the bugs should be fixed in https://github.com/WordPress/gutenberg/pull/41393 That makes 13.5 our merging target for this PR

MaggieCabrera:
We should make a quick test on dotcom but this should be mergeable now

pbking:
Tested on wpcom.
Works like a charm. :)
Sending it out into the wild now.

PR #5853, opened 2022-04-12
Small tweak cleansing logs of double-quotes

pbking:
[This PR](https://github.com/Automattic/themes/pull/5285) removed the "'s from log messages.  But didn't save them to the variable. :P

This change saves it to the variable so the cleansed value is actually used.


mikachan (approved):
😅. Looks good!

PR #5858, opened 2022-04-13
Skatepark: refactor patterns to load directly from the editor

### LINKED ISSUE #5803: Multiple themes: update how patterns are loaded

madhusudhand:

Validate patterns both in inserted and in template parts.

#### Related issue(s):

Fixes #5803 for Skatepark.


skatepark/patterns/paragraph-with-quote.php
```diff
@@ -0,0 +1,45 @@
+<?php
+/**
+ * Title: Paragraph with quote
+ * Slug: skatepark/paragraph-with-quote
+ * Categories: featured, text
```
mikachan: I think we should remove 'featured' from here.

skatepark/patterns/full-width-image-with-aside-caption.php
```diff
@@ -0,0 +1,26 @@
+<?php
+/**
+ * Title: Full width image with aside caption
+ * Slug: skatepark/full-width-image-with-aside-caption
+ * Categories: featured, text
```
mikachan: Maybe we could have an 'image' category as well?

skatepark/patterns/mixed-media-in-container.php
```diff
@@ -0,0 +1,89 @@
+<?php
+/**
+ * Title: Mixed Media in container
+ * Slug: skatepark/mixed-media-in-container
+ * Categories: featured, text
```
mikachan: This could be added to an 'image' category too.

skatepark/patterns/text-list-with-button.php
```diff
@@ -0,0 +1,34 @@
+<?php
+/**
+ * Title: Text list with button
+ * Slug: skatepark/text-list-with-button
+ * Categories: featured, text
```
mikachan: I'm not sure this needs to be in 'featured' either.

skatepark/patterns/three-columns.php
```diff
@@ -0,0 +1,42 @@
+<?php
+/**
+ * Title: Three columns
+ * Slug: skatepark/three-columns
+ * Categories: featured, columns
```
mikachan: Possibly doesn't need to be in 'featured' either.

mikachan (commented):
This is looking good and the implementation works well. I've just left a few comments about the pattern categories. I'm not sure we should have so many listed in 'Featured', and I think there are a couple that use images that we could add to a new 'Images' category.

skatepark/patterns/text-list-with-button.php
madhusudhand: I guess it doesn't needs to be. Removed from featured.

pbking:
I see all of @mikachan 's comments have been resolved.  Looks good to me.

🚢

pbking (approved):

PR #5861, opened 2022-04-13
Tools: add action to check for readme.txt changes

mikachan:
This PR adds a GH workflow that checks if a readme.txt file has been updated whenever any theme files have been updated. If it detects that a readme.txt hasn't been updated at the same time as a theme change, it will fail and 'grey-out' the PR merge button (we can still merge, but the button won't be green.)

This is a light alternative to something like https://github.com/Automattic/themes/pull/5092, which would require us to change our workflow quite a lot. I thought this action could act as a reminder whilst we work on how we want to automate the changelogs and if we want to use conventional commit messages, etc.

The action works by checking the following conditions:
- If a file has been updated in any sub-directory, in order to catch updates to theme files
- If a file has been updated in the repository root, in order to ignore these changes
- If a file has been updated in any excluded sub-directories, in order to ignore changes to variations/, .github/, .husky/ and .phab/
- And if any readme.txt file has changed (if so, the check will pass)

Possible problems:
- The action doesn't check if a specific readme.txt has been updated, only if any readme.txt has changed. I thought this would be good enough to catch most theme updates, especially as an effective reminder, but maybe this needs more work..

#### Related issue(s):
https://github.com/Automattic/themes/pull/5092

### DIFF
--- a/.github/workflows/changelog-check.yml
+++ b/.github/workflows/changelog-check.yml
@@ -1,0 +1,41 @@
+# This workflow checks if a readme.txt file has been updated whenever any theme files have been updated. 
+# This is to prevent updating theme files without also updating the corresponding changelog in the readme.txt.
+# It ignores any changes made to root files (e.g. theme-utils.mjs) and any files in excluded sub-directories.
+
+name: "Changelog Check"
+
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+jobs:
+  main:
+    name: Ensure a readme.txt file has been updated if a theme has been changed
+    runs-on: ubuntu-latest
+    steps:
+      - name: Verify any readme files
+        uses: dorny/paths-filter@v2.2.1
+        id: filter
+        with:
+          filters: |
+            theme_change:
+              - '**/*/**'
+            theme_readme:
+              - '**/readme.txt'
+            root_files:
+              - '*'
+              - '!**/*/**'
+            excluded_subdirectories:
+              - 'variations/**'
+              - '.github/**'
+              - '.husky/**'
+              - '.phab/**'
+      - name: Reminder to update changelog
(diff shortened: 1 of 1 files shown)


pbking:
I'm keen to evaluate this but I'm not sure how to do it without bringing it in...

I'm also keen for a check like this to go along with the deploy/version-bump script.  It might be helpful for ALL of the commit logs that are being applied to the tag (everything since the last deploy) are also added to the readme file and the user doing the deploy given the option to edit those before committing (as a part of the version bump step).  

For deployments that only effect a single theme (the best kind) there would likely be no manual changes needed.  And for the rest it should just be a matter of deleting the lines from each theme that aren't relevant.

pbking:
I'm keen to evaluate this but I'm not sure how to do it without bringing it in...

I'm also keen for a check like this to go along with the deploy/version-bump script.  It might be helpful for ALL of the commit logs that are being applied to the tag (everything since the last deploy) are also added to the readme file and the user doing the deploy given the option to edit those before committing (as a part of the version bump step).  

For deployments that only effect a single theme (the best kind) there would likely be no manual changes needed.  And for the rest it should just be a matter of deleting the lines from each theme that aren't relevant.

mikachan:
Thanks for taking a look @pbking. I've been testing this out using a different repo, mainly so I could safely keep getting stuff wrong 😅

> I'm also keen for a check like this to go along with the deploy/version-bump script.

I thought about something similar to this, but I talked myself out of it because I thought we'd maybe want to include something slightly different to the PR title in the changelog, and we probably wouldn't want to include all the changes from one deploy in each changelog. But, this is easily fixed by giving the user doing the deploy the option to edit all this during deployment. I like this, I'll give this a go next!

> For deployments that only effect a single theme (the best kind) there would likely be no manual changes needed.

Definitely the best kind 😁

mikachan:
Closing this in favor of https://github.com/Automattic/themes/pull/5889.

PR #5862, opened 2022-04-14
Meraki: Work in progress

alaczek:
* Update theme screenshot and image credits.
* Add custom footer pattern.
* Make footer full width.
* Update header - apply vertical alignment to header elements.
* Update theme.json file - add tiny font size, update color palette, add styling for button block.


pbking:
One of the changes in this (I haven't figured out which yet) introduces horizontal scrollbars:

image

pbking:
For reasons I can't explain the screenshot that was added doesn't seem to be showing up for me:
image

meraki/theme.json
```diff
@@ -110,11 +115,26 @@
 	},
 	"styles": {
 		"blocks": {
+			"core/button": {
```
pbking, line 118: Since this is a Blockbase theme we leverage a lot of CUSTOM button values so that they can be taken advantage of in other areas (such as forms and buttons that aren't button blocks) and to achieve better control of hover states etc.  Thus, rather than changing the `styles.block.core/button` attributes directly it's better to change `custom.button` values instead.  See [Quadrat](https://github.com/Automattic/themes/blob/trunk/quadrat/theme.json#L62) for an example of overriding those button styles for a Blockbase theme.

meraki/theme.json
```diff
@@ -110,11 +115,26 @@
...
+				},
+				"color": {
+					"background": "var(--wp--preset--color--foreground)",
+					"text": "var(--wp--preset--color--background)"
+				},
+				"typography": {
+					"letterSpacing": "0.05em",
+					"fontSize": "var(--wp--preset--font-size--tiny)",
+					"textTransform": "uppercase",
+					"lineHeight": "1.2"
```
pbking, line 130: I think the lineHeight value chosen introduces an odd padding in the button (top vs bottom)
image

You might want to tweak that one direction or another to ensure that the padding in the button falls evenly.

pbking:
Made a few comments, going to move this PR to `in-progress` for some further work before bringing it in.

Issue #5692, opened 2022-03-16
Pendant: Variations


pbking:
<img width="574" alt="image" src="https://user-images.githubusercontent.com/146530/162965967-0d8708e2-7f25-44b1-af84-1034862b408f.png">

pbking:
@beafialho do you expect these Color Studies to resolve in alternate color palettes for Pendant?

<img width="623" alt="image" src="https://user-images.githubusercontent.com/146530/159981736-c3370a97-4241-4225-8f9c-5ff1e680f9d2.png">

beafialho:
@pbking no, everything under "Drafts" is old explorations. I'm going to work on some alternate color palettes for Pendant and will let you know.

pbking:
Coolio, thank you.

beafialho:
@pbking since we will be no longer limited to color variations in themes, I added a few Style Variations to [the Figma file](https://www.figma.com/file/JdntsOfLqo8Vy0d5uTWHTI/?node-id=1729%3A50) that are fewer but accomplish totally different design moods.

pbking:
I believe all of the variations have been completed.  Closing this.

Any further variation changes that are needed should be opened as new items.

PR #5877, opened 2022-04-15
Add/pendant variations

pbking:
Adds three new variations.  Each variation includes color palette and font changes.

Makes progress on (but doesn't close) #5692 

image

image

localhost_4759_ (1)

image

localhost_4759_ (2)

image

localhost_4759_ (3)


beafialho:
@pbking only the Pink variation fonts are loading for me, but based on the screenshot:

1. "Pink" looks good.
2. "Electric Blue" is looking a bit different from the Figma file. This should have `Bold` H1's, but `Regular` in other headings and post titles. The other font, IBM Plex Mono should use `Regular` instead of `Light` and it should use `Bold` on buttons.
Captura de ecrã 2022-04-18, às 11 52 21
Captura de ecrã 2022-04-18, às 11 52 35

3. "Dark Navy" also has some differences. Headings and post titles should be `Light` but paragraphs should be `Regular`. Also, there are some that should use `Bold`. These are H4 from what I see in the editor?
Captura de ecrã 2022-04-18, às 12 10 39
Captura de ecrã 2022-04-18, às 12 10 29

pbking:
I'm going to bring this in and open two additional items for the changes needed for 'Electric Blue' and 'Dark Navy.  It's definitely clear that those aren't ready yet.

If you find any additional changes needed for 'Pink' please open as a new item.

PR #5878, opened 2022-04-18
Tools: Stale issues action update

MaggieCabrera:
After a few runs of this action, I think there's room for improvement. It looks like there's a [limit of operations per run](https://github.com/actions/stale#operations-per-run) so we've only had a few issues labeled since we created the action.

Instead of increasing the amount of operations I think we can simply run the action every other hour instead and change the order so it starts from the older issues. That should cover most of them faster, then we can go back to once a day for general maintenance once the action is not encountering as many stale issues.

I also added the option to remove the `Stale` label if there's updates to the issue. With the amount of issues we handle, I'd rather have that automated than let the issues labeled as stale even if they are not.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -7,7 +7,7 @@ name: Mark stale issues and pull requests
 
 on:
   schedule:
-  - cron: '34 19 * * *'
+  - cron: '0 */2 * * *'
 
 jobs:
   stale:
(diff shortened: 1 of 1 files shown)


pbking:
Good call @MaggieCabrera.  These changes look good to me.  I agree on setting this to eat through the issues now and then re-adjust it to be daily once we've gotten through them all.

Also REALLY happy for the automated removal of stale once an issue gets some traction.

pbking (approved):

PR #5888, opened 2022-04-19
Pendant: Fix link color

### LINKED ISSUE #5885: Pendant: Link colors
When working on the demo site, I added a link to the "October Picks" heading in the banner and the link's default color is Black. When trying to set it to White, all other links in the page turned white, which just inverted the problem.

Captura de ecrã 2022-04-19, às 12 28 27

pbking:
Apply the 'foreground' color to links in the group in the 'hero' pattern to match the 'text' color (which is set to `background`).

To achieve this (with the FSE) you must "turn on" link color editing. 
image
image

Fixes #5885

Before:
image

After:
image

### DIFF
--- a/pendant/patterns/hero.php
+++ b/pendant/patterns/hero.php
@@ -5,8 +5,8 @@
  */
 ?>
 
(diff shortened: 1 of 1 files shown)


matiasbenedetto (approved):
I tested it and it LGTM :+1:

PR #5889, opened 2022-04-19
Tools: automatically handle changelog updates

mikachan:
This PR adds a function called `updateThemeChangelog` to the deploy script, which is called during `versionBumpThemes`. It accepts a theme slug as an argument, and then runs through the following process for the corresponding theme:

- Gets the current theme version from `style.css`. This is run after a version bump has already happened, so this should always be the latest version.
- Gets the list of updates from the commit logs, and adds a bullet point (*) to each line
- Builds a new changelog entry based on the commit logs and the latest version number
- Updates the theme's `readme.txt` with the new changelog entry
- The user is given an option to manually update any of these changes at the same time as verifying the version bump changes

The idea is that all commit logs will be added to each updated theme's changelog, and the user will need to delete the appropriate logs during the manual step. Ideally, only a small number of themes should be updated per deployment, as this reduces the overhead when running the deploy script.

I've also added a new function called `getCommitLogs` to reduce the number of times we seemed to be repeating this functionality.

Thanks to @pbking for the prompt! This is an alternative to #5861.

#### Related issue(s):
https://github.com/Automattic/themes/pull/5861
https://github.com/Automattic/themes/pull/5092

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -141,7 +141,7 @@ async function pushButtonDeploy() {
 		if( thingsWentBump ){
 			prompt = await inquirer.prompt([{
 				type: 'confirm',
-				message: 'Are you good with the version bump changes? Make any manual adjustments now if necessary.',
+				message: 'Are you good with the version bump and changelog updates? Make any manual adjustments now if necessary.',
 				name: "continue",
 				default: false
 			}]);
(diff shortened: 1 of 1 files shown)


mikachan:
I've just tested this out on a merge and deploy here: https://github.com/Automattic/themes/commit/70ee5b19cfa133b97af5b0fc473c60516f911d06

It seemed to work well! I needed to remove the logs that didn't apply to every theme, but I think this is a great improvement to our process to ensure our changelogs are kept up to date.

Here's what a Livro change looked like:

image

pbking (approved):
👍 👍 👍 

A great change to our workflow!  Thank you.

PR #5900, opened 2022-04-20
Heiwa: initial commit

iamtakashi:
This is the initial commit of the Heiwa (Blockbase child) theme. This theme is a part of the verticals project (more info: pbxlJb-18I-p2). 

This is the first time making a theme in this era :) I most likely have done something wrong. Please let me know, and thanks in advance!!

Homepage example:
heiwa

cc @alaczek  @onuro


jeffikus:
I like the look of this one @iamtakashi :) Thanks @mikachan for adding this to the board!

heiwa/templates/header-footer-only.html
```diff
@@ -0,0 +1,7 @@
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
+
+<!-- wp:group {"tagName":"main","style":{"spacing":{"margin":{"top":"0px","bottom":"0px"}}}} -->
+<main class="wp-block-group" style="margin-top:0px;margin-bottom:0px"><!-- wp:post-content {"layout":{"inherit":true}} /--></main>
+<!-- /wp:group -->
+
+<!-- wp:template-part {"slug":"footer","tagName":"footer","className":"site-footer-container"} /-->
```
jffng: I don't think you need this class:

```suggestion
<!-- wp:template-part {"slug":"footer","tagName":"footer"} /-->
```

jffng (commented):
Hey @iamtakashi! Very exciting :) 

I have two general questions but this is looking really nice already 👏 

- What is the reason for making this a Blockbase child, versus as a standalone block theme? Have you seen [Block Canvas](https://github.com/Automattic/themes/blob/trunk/block-canvas/theme.json)? It may offer you a simpler starting point if you want to go the route of standalone block theme.
- Will you add a home.html template eventually?

iamtakashi:
Hi @jffng! Thanks for taking the time to review. Much appreciated!

> What is the reason for making this a Blockbase child, versus as a standalone block theme?

I was just following the document @alaczek has put together, and the doc seemed to focus on creating a Blockbase child theme. I'll try the Block Canvas next time!

> Will you add a home.html template eventually?

I don't mind working on it if that's better. I didn't make it just because the doc didn't mention it :) Shall I work on it?

jffng:
> I don't mind working on it if that's better. I didn't make it just because the doc didn't mention it :) Shall I work on it?

That can be a follow up, this LGTM as a starting point. I was mainly asking because you designed a nice home page mockup, and it would be great for users to have this already built for them to edit in the site editor. 

Are you familiar with how to deploy the theme to wpcom after you merge? Basically you'll want to run `npm run deploy` as soon as you merge this, and follow the prompts accordingly.

iamtakashi:
> That can be a follow up, this LGTM as a starting point. I was mainly asking because you designed a nice home page mockup, and it would be great for users to have this already built for them to edit in the site editor.

Thanks for the kind words. The suggestion makes sense. However, I was told this theme is only available on WP.com. I think we use the headstart(?) to build the homepage for the users who wish to have that. Am I correct about that? Do we still need to add home.html?

> Are you familiar with how to deploy the theme to wpcom after you merge? Basically you'll want to run npm run deploy as soon as you merge this, and follow the prompts accordingly.

Thanks for the info. I'm not very familiar with the current method at all. I had a looong break from launching a theme :)  Where do I run that command from? Is it from my local copy of this repo (I needed to fork the repo to make this PR) or my WP.com sandbox? Sorry for asking noob questions!

iamtakashi:
One more question. I only tested the theme in my local environment. How can I test it on WP.com?

MaggieCabrera:
> Thanks for the kind words. The suggestion makes sense. However, I was told this theme is only available on WP.com. I think we use the headstart(?) to build the homepage for the users who wish to have that. Am I correct about that? Do we still need to add home.html?

We do still use headstart but home.html is better for users to customize a home page without having to move from the site editor


> Thanks for the info. I'm not very familiar with the current method at all. I had a looong break from launching a theme :) Where do I run that command from? Is it from my local copy of this repo (I needed to fork the repo to make this PR) or my WP.com sandbox? Sorry for asking noob questions!

You run the command from the root of the repo, the instructions are [on the readme of the repo.](https://github.com/Automattic/themes#sandbox-tools) Let us know if you have any problems with that!

> One more question. I only tested the theme in my local environment. How can I test it on WP.com?

After the first deploy you can activate the theme on a test site, but not before, sadly.

iamtakashi:
@MaggieCabrera Thanks for the info and the clarification. 

> We do still use headstart but home.html is better for users to customize a home page without having to move from the site editor

That makes sense. I'll make a home.html!

> Let us know if you have any problems with that!

Thanks for the support. Much appreciated!

alaczek:
> We do still use headstart but home.html is better for users to customize a home page without having to move from the site editor

@MaggieCabrera could you elaborate on this one? Headstart works on theme switch AFAIK, so the content should be there when customer opens site editor. (to clarify, I'm not saying we shouldn't create home.html, just trying to understand how everything works :) )

MaggieCabrera:
> 
> @MaggieCabrera could you elaborate on this one? Headstart works on theme switch AFAIK, so the content should be there when customer opens site editor. (to clarify, I'm not saying we shouldn't create home.html, just trying to understand how everything works :) )

The way headstart works for home page content is it creates a page and pastes the content to it, then if the user wants to change their site to show posts instead of that page, they need to change it in the customizer. Using templates would be the FSE way to handle the home page instead. 

When the user goes to the site editor on a theme without home.html what they'll see is the index.html template instead, which is not what the home page actually shows if we headstart it with a home page with content already.

Another drawback of this is that we can't do any adjustments to the home page content once the site is headstarted, the content is just there and we have no way of pushing updates to it. With a home template, we can update the theme and the users sites will get the updates. This is less likely to be an issue when blocks get more stable but some of them are still changing in markup a bit, and it would be handy to keep some control there too.

iamtakashi:
@MaggieCabrera Thanks for the detailed reply. Much appreciated!

I wasn't involved in the discussion about this approach, and I'm learning about all this as I go, and I'd like to hear the team's thoughts on a few things that I noticed.

1. For a theme whose homepage doesn't show any posts, I find it strange to see the static content from the home.html when I set my homepage to show the latest posts on the Reading Settings page.

Screenshot 2022-04-28 at 17 25 56

How do we address this?  The template editor does explain it, but I suspect it is confusing to many.

Screenshot 2022-04-28 at 17 35 56

2. In WP.com, while the homepage is set to show a static page, if I activate a new theme with the option replacing the homepage template, it doesn't seem to do anything...

wordpress com_themes_iamtakashi wordpress com

3. This is similar to 2. In WP.com, while the homepage is set to show the latest posts if I activate a new theme with the option preserving my homepage, it doesn't preserve the homepage, and instead, the page shows the content from home.html.

wordpress com_themes_iamtakashi wordpress com (1)

Have we heard about these problems from WP.com users? What are the team's thoughts about these issues? I'm interested in figuring out how to move these forward.

MaggieCabrera:
@iamtakashi I'm not sure I have answers to all your questions. Here are a few conversations about this same topic that we've had before:

p1643685532877739-slack-C029FM1EH

p1637071486306500-slack-C029FM1EH

As of today we still rely on headstart for dotcom, but block themes should be able to provide a proper starting point for themes without it, and templates are a good first step. This is all still in flux, so we still need the fallback that headstart provides.

About your points 2 and 3, that flow is designed thinking about classic themes, where you have two options: either a static page or a list of posts. With block themes, there's no such thing anymore, since the home template could be either or both (like TT2 includes a query loop of posts in their home page but still provides static content in the form of patterns). When a user of a classic theme changes to a block theme there's certainly a disconnect there as you have certainly pointed out. We need to reflect on how we want that experience to be.

I don't think we've had recent issues brought here to our repo about this from users, don't know if there's any over at calypso.

@Automattic/theam do you all have any insights on this or something I might have missed from previous discussions on this topic?

iamtakashi:
@MaggieCabrera Thank you so much for providing your insights and the links to the previous discussions. I have not been able to digest all that yet, but I should be able to do that next week. Thanks again!

iamtakashi:
A sticky point about home.html in this theme is that the homepage shows the static content in the template even if you set the homepage to show the latest posts in the reading setting.

Since this theme is only going to be available on WP.com where we can use Headstart, can we go ahead without the home.html? I assume we can add it later if we think that's definitely a better experience.

I've just made minor edits, but if everything looks ok for you all, could you merge it and run the command for me? I'm asking this because I'm working on this from my fork of the repo. (I couldn't even add and push a branch to this repo for some authenticating reason.)

mikachan:
> Since this theme is only going to be available on WP.com where we can use Headstart, can we go ahead without the home.html? I assume we can add it later if we think that's definitely a better experience.

Yes, I think this works for now, especially as we can add it later 👍

> I've just made minor edits, but if everything looks ok for you all, could you merge it and run the command for me?

Yes sure, I'll merge this today and run through the deploy process for it. If you want to do this yourself in the future and have questions, anyone from the team will be happy to help or pair!

mikachan (approved):
This is looking good to me. As discussed in the comments, we will probably add a `home.html` template in the future, but otherwise, I think this is good to go 🚢

iamtakashi:
> Yes sure, I'll merge this today and run through the deploy process for it. If you want to do this yourself in the future and have questions, anyone from the team will be happy to help or pair!

Thank you, @mikachan so much for the support and for taking the deploy task this time for me. Much appreciated 🙇🏻‍♂️

Issue #5847, opened 2022-04-12
Blockbase: audit ponyfill for 6.0


MaggieCabrera:
Now that 6.0 is near, we should review Blockbase's CSS and see if we can remove more stuff (and bump the WP required version as we do so).

- Comments block https://github.com/Automattic/themes/issues/5833
- Separator block changes https://github.com/Automattic/themes/pull/5834
- We could add variations that mirror the palettes we are providing in the customizer options. This will show theme developers how variations are handled via FSE.
- If https://github.com/WordPress/gutenberg/pull/39926 makes it to 6.0 we can remove the alignments styles too

Any other changes come to mind?

jeffikus:
Nice work creating this 'overview' issue.

MaggieCabrera:
tagging @beafialho for opinion on adding variations to some of our other themes (maybe livro, archeo, stewart?)

beafialho:
@MaggieCabrera this makes sense to me. I'm curious to know your thoughts on adding style variations to all themes from now on, similarly to what we did with patterns? Would like to know @mtias thoughts too.

MaggieCabrera:
> @MaggieCabrera this makes sense to me. I'm curious to know your thoughts on adding style variations to all themes from now on, similarly to what we did with patterns? Would like to know @mtias thoughts too.

yes, I think style variations are a great idea, we can add them to the theme as part of the release or afterward like we are doing for other themes like remote, I don't have a preference.

beafialho:
I'll add it to my list to create some style variations for Livro, Archeo and Stewart :+1:

mtias:
What are these variations about?

beafialho:
@mtias these are [style variations](https://github.com/WordPress/gutenberg/issues/39700), as in the TT2 theme.

mtias:
Ah, I misread it as block style variations!

PR #5901, opened 2022-04-21
Blockbase: move button padding styles from ponyfill to theme.json

### LINKED ISSUE #5856: Blockbase: Button block does not respect paddings defined in Global Styles
## Steps to replicate

* Add a button on a page.
* Go to Global Styles and open the layout panel for Button block
* Change the paddings
* See the button not changing its size

## Result

The button stays the default style. The border-radius setting work as expected, but not the paddings. If I edit the paddings on the button individually, it works but not with Global Styles.

Screenshot 2022-04-12 at 22 59 37

## Expected

The padding in the button changes. I also tested with Twenty Twenty Two, and the result was expected.

madhusudhand:

Moved core/button block padding styles from ponyfill to theme.json so that global styles can be configured.

#### Screenshots

**Editor:**

image

**Frontend:**

button-global-styles

#### Related issue(s):

Fixes: #5856 
Related: #5847


scruffian:
My main concern with this is that the search and file blocks also match the same button style. I'm trying to introduce a button element block to help deal with that: https://github.com/WordPress/gutenberg/pull/40260

pbking:
Funny, I thought I had left comments on this already.

I also wanted to note that the 'outline' styles of the button don't follow the padding rules, however leveraging styling buttons "The Blockbase Way" does.  Here is an example from this branch (with padding set to 35 px):
image

Also this behavior won't work as expected in any of the Blockbase children.  This is Quadrat in regular then hover:

image

image

That's in addition to the "other non-block buttons" that @scruffian mentions.  I do think that the button element could definitely help that out, but we also have the hover state management that needs to be dealt with, as well as rolling those details into block styles (such as the outline style) that needs to be figured out and dealt with as well before we can move on from this method of button styling in Blockbase.

madhusudhand:
@pbking addressed the issues in blockbase and in the child themes.
To verify the approach, committed changes related to single child theme `quadrat`.

All other themes needs to be compiled and verified.

**Blockbase**
blockbase-btn

**Quadrat**
quadrat-btn

madhusudhand:
Similar to other block themes, I moved the button hover styles to [blockbase/style.css](https://github.com/Automattic/themes/pull/5901/files#diff-bc4c9bd7c51196b5bf24994a03126fb138ebb6c208ec88bacd3a264b07b35a1f) in blockbase.

However, these styles are not getting applied to child themes such as `quadrat`
Updated `blockbase/functions.php`, but still it doesn't work.
What am I missing here?

pbking:
> Similar to other block themes, I moved the button hover styles to [blockbase/style.css](https://github.com/Automattic/themes/pull/5901/files#diff-bc4c9bd7c51196b5bf24994a03126fb138ebb6c208ec88bacd3a264b07b35a1f) in blockbase.

Anything that Blockbase needs to make available to child themes should still reside in `ponyfill.css`.  The only change we should be making to support this is **what** css is in ponyfill, not where it is.

pbking:
> All other themes needs to be compiled and verified.

Any changes that require child themes to change isn't a change we can make for Blockbase.  There are an unknown number of themes outside of Automattic (and even inside Automattic not under our management) that would need to change.  So any changes we make to Blockbase (at this point) should ***not*** require changes to child themes. 
That does limit what changes we can make, for sure, but that's the cost of being a "parent theme" I think.

pbking:
@madhusudhand I took a look at this and I'm not sure why I was opposed earlier.  The changes to the child themes are reasonable.  I'm knocking this branch about and it's the best fix.  Thank you for working on this.

I'll push the change rebuilding the rest of the child themes css.

pbking:
Closing: Work merged in #6167

PR #5904, opened 2022-04-21
Fix/pendant variation blue

### LINKED ISSUE #5881: Pendant Variation Improvements: Electric Blue
Continues Progress of #5692

This variation should have Bold H1's, but Regular in other headings and post titles. The other font, IBM Plex Mono should use Regular instead of Light and it should use Bold on buttons.

pbking:
Fixes: #5881
Depends: [This Gutenberg item ](https://github.com/WordPress/gutenberg/pull/40489)needs to be merged before this works as expected.

A couple of minor template tweaks were made based on discussion:

image

The example social icon color was changed from 'white' to 'currentColor' so the variations look good:
image

This changes the fonts as discussed with design.
The variable weight Rubik font was used.  (I was able to get it to work after all! Yay.)

Desktop:
image

Mobile:
image

IBMPlexMono Regular and Bold (with both italic typefaces) are used:
image


pendant/patterns/header.php
```diff
@@ -13,7 +13,7 @@
 
 <!-- wp:site-title {"textAlign":"center","style":{"elements":{"link":{"color":{"text":"var:preset|color|background"}}}}} /-->
 
-<!-- wp:social-links {"iconColor":"background","iconColorValue":"#ffffff","className":"is-style-logos-only"} -->
+<!-- wp:social-links {"iconColor":"background","iconColorValue":"currentColor","className":"is-style-logos-only"} -->
```
MaggieCabrera, line 16: I think primary would have worked here too, but I don't have a preference, it looks fine in the editor so 👍

MaggieCabrera (approved):
This LGTM from a technical standpoint, @beafialho if you think the design looks good, we can bring this in

beafialho:
It seems like the Rubik font in @pbking screenshots isn't in Bold:

| Screenshot      | Intended |
| ----------- | ----------- |
|  164513644-b798eda7-5631-4c66-ae67-acf5fc53030a    |   Captura de ecrã 2022-04-26, às 12 28 11    |

Is that due to that Gutenberg item?

pbking:
> It seems like the Rubik font in @pbking screenshots isn't in Bold: Is that due to that Gutenberg item?

I was actually able to use the 'variable' typeface for Rubik so I was hoping that the weight would be as-desired with that.  So that should work fine even with the Gutenberg bug.  However, it doesn't seem to be as "variable" as I was hoping.  Setting the weight to 900 renders the exact same as the 700.

Perhaps it would be better to go back to the **-Bold** version instead.

Changing to that, here's a screenshot.

image

Look right @beafialho?

beafialho:
Looks right @pbking, one last thing, could you please check that H4, H5 and H6 are using Inter Regular? They seem a bit lighter to me in the screenshot.

pbking:
Good eye @beafialho! 
They were using -Light.
Changing to use -Regular for the lighter weights.

image

beafialho:
Ah-ha, that's great!

PR #5905, opened 2022-04-21
Fix/pendant variation navy

### LINKED ISSUE #5882: Pendant Variation Improvements: Dark Navy
Continues Progress of https://github.com/Automattic/themes/issues/5692

Headings and post titles should be Light but paragraphs should be Regular. Also, there are some that should use Bold. These are H4 from what I see in the editor?

pbking:
This addresses the feedback from design regarding the 'Dark Navy' Variation.

Closes: #5882

Depends on [this Gutenberg item](https://github.com/WordPress/gutenberg/pull/40489) to render fonts as expected.

image

The variable font wasn't slimming down for the smaller weights so the -Light and -Bold options were used instead.

image

image

image

### DIFF
--- a/pendant/patterns/post-topics.php
+++ b/pendant/patterns/post-topics.php
@@ -5,7 +5,7 @@
  */
 ?>
 
-<!-- wp:heading {"style":{"spacing":{"margin":{"bottom":"20px"}}},"fontSize":"large"} -->
-<h2 class="has-large-font-size" style="margin-bottom:20px"><?php echo esc_html__( 'Popular Topics', 'pendant' ); ?></h2>
+<!-- wp:heading {"level":4,"style":{"spacing":{"margin":{"bottom":"20px"}}},"fontSize":"large"} -->
+<h4 class="has-large-font-size" style="margin-bottom:20px"><?php echo esc_html__( 'Popular Topics', 'pendant' ); ?></h4>
 <!-- /wp:heading -->
 <!-- wp:categories /-->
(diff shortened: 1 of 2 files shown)


mikachan (approved):
This looks like it fixes all the issues in https://github.com/Automattic/themes/issues/5882 to me. I can see the correct fonts being used for both headings and body content for the navy variation.

Approving but tagging @beafialho too in case you want to take a look.

beafialho:
Font weights look good to me but I see a link in red. Is that intentional? The link should be the same color as the text, just underlined.

MaggieCabrera:
> Font weights look good to me but I see a link in red. Is that intentional? The link should be the same color as the text, just underlined.

that's a link with a custom color selected in the editor, to prove that the user can change it if they like, it's probably something on the test content Jason uses for general testing.

beafialho:
> that's a link with a custom color selected in the editor, to prove that the user can change it if they like, it's probably something on the test content Jason uses for general testing.

Oh I see, thanks!

One other thing I've noticed is the duotone in the images is missing. But that's also blocked by Gutenberg, right? If that's the case, this looks good to me.

pbking:
> One other thing I've noticed is the duotone in the images is missing. But that's also blocked by Gutenberg, right? If that's the case, this looks good to me.

Considering that a user doesn't have the option to disable that feature I'm not keen to ship it in a variation until they can.  The Skatepark and TwentyTwentyTwo-Variations that have that get so many bugs filed against them; it seems to be really confusing for users.

While we can ship with the duotone set for variations I strongly urge us not to (until the user can also modify that feature themselves).

PR #5906, opened 2022-04-21
Update/pendant page template

pbking:
This change gives the page templates the design treatment outlined for "Index" in the original design resources:

image

Before:
image
image

After:
image
image

Also (since the page template by default now shows the page, and the 'About' page design does not) the `header-footer-only` template has been modified slightly to exclude all top padding so that the design can be achieved:

Before:
image
image

After:
image
image

Also: Did some housekeeping and removed an unused class from all of the footers.


pbking:
@beafialho this fixes the "Where is the page title and why doesn't it look like it should?" problem we had while pearing the other day.

Now, creating a new page (to link to from the `Read All` link) will look correct.  And the idea of using that new page with the 'grid-of-posts' pattern will still be good.  Though I am still looking into why the `grid-of-posts` pattern doesn't actually WORK when placed on a page...

This change will require the 'About Page' example page to use the `header-and-footer-only` page template to achieve the desired design.
image


That fix will come in another change.

jffng (approved):
Tested the following use cases and they LGTM:

✅ Page template
✅ Header & footer only template with the "About pattern"

beafialho:
Thank you. @pbking could you try it out on the demo site? I'm still seeing this in the "Articles" page:

Captura de ecrã 2022-04-22, às 17 11 56

pbking:
I can confirm that when this change is in place the demo site's 'Articles' page looks as expected:

image

About has too much going on with the change in place (as expected) and will need to be changed to the "header and footer only" page template:

image

Gonna bring this change in and update that page.

pbking:
@beafialho this has been merged and deployed.  I've updated the 'About' page to use the 'header-footer-only' template and the 'Articles' page looks as expected.

beafialho:
Thank you @pbking, it does look as expected to me as well!

PR #5908, opened 2022-04-22
Remote: add style variations

### LINKED ISSUE #5500: Remote: theme.json: Alternative color palettes

madhusudhand:

Add the following style variations

* Timid Pink
* Mono Bright

Also fixed font sizes of headers in default theme.

**Question: Do I need to update any WP or GB version?**

#### Screenshots:

**Timid Pink**

timidpink

**Mono Bright**

forestgreen

#### Related issue(s):

Fixes: #5500


madhusudhand:
@beafialho Can you review these style variations? 
I am not confidant on duotone for `Timid Pink`. I have used the primary color and it seems to be too much.

beafialho:
@madhusudhand in your screenshots only the color varies, and the typography styles should vary too. I haven't yet finished these style variations, but I am almost done and will let you know.

madhusudhand:
@MaggieCabrera  Updated min version to 5.9, is this the right version for these changes?

MaggieCabrera:
> @MaggieCabrera Updated min version to 5.9, is this the right version for these changes?

5.9 has the variations code, but 6.0 is going to bring the improvements. I think 5.9 is enough in this case. We should check before we deploy to dotcom that the current GB version has the latest changes to the styles interface (if you see the titles of the variations, then that's the one).

beafialho:
Thanks @madhusudhand, I've noticed some discrepancies from the mockups:

**Timid Pink**
- Post titles don't seem to be 72px, as they should:

Captura de ecrã 2022-04-29, às 12 07 34

- H6 also doesn't match to the mockup:

Captura de ecrã 2022-04-29, às 12 08 50

- Paragraph's sizes don't look correct. Please check the [Figma styles](https://www.figma.com/file/S382KokJxg0G043wsNoFQj/Remote?node-id=1623%3A595).

- Paragraphs, buttons and H6 should have `0.05em` letter-spacing:

https://user-images.githubusercontent.com/65220155/165933901-91d75ee5-e960-45c2-84c4-513029606db3.mp4

- Buttons don't have a hover effect:

https://user-images.githubusercontent.com/65220155/165933933-da75d5f5-060f-4bad-a035-e01020986d5c.mp4

**Mono Bright**

- Post titles don't seem to be 72px

- H1 through H3 should be IBM Plex Sans `200`, but from H4 on should be IBM Plex Sans `300`

- Paragraph sizes also aren't correct, letter-spacing should be `0.02em`

**Forest Green**

- Post titles don't seem to be 72px

- H6 and Buttons should have a `0.02em` letter spacing

- Buttons also don't have a hover effect

For the duotone effects, I think some of the lighter colors weren't the ones I used, here are the correct ones. I will make sure to include these icons in the specs next time:

Captura de ecrã 2022-04-29, às 12 33 38
Captura de ecrã 2022-04-29, às 12 33 34
Captura de ecrã 2022-04-29, às 12 33 29

All these variations have font properties changes to achieve a specific style, I didn't just replace colors and fonts but I took some time to leave it [all noted on Figma](https://www.figma.com/file/S382KokJxg0G043wsNoFQj/Remote?node-id=1623%3A595). Please take a closer look at these, it's all specified there.

madhusudhand:
@beafialho Thanks for the detailed review and notes.

All the mentioned issues are addressed, except that not all paragraph sizes are included. 
Only standard sizes (small, normal, medium, large) and an additional Huge sizes are included.

beafialho:
Thank you @madhusudhand, I'll take a look in the next days and will provide feedback.

beafialho:
Thank you once again for working on this @madhusudhand.

> Only standard sizes (small, normal, medium, large) and an additional Huge sizes are included.

That's ok 👍 

I have some more comments about the font sizes. They're still not right. So that I understand, are font sizes changing between variations a problem?

The intended sizes are on the right column in the screenshots below:

**Timid Pink & Forest Green**

- Large and Huge still don't have the intended sizes

Captura de ecrã 2022-05-06, às 10 30 51
Captura de ecrã 2022-05-06, às 10 24 24

**Mono Bright**

- Neither size on the left column is correct.

Captura de ecrã 2022-05-06, às 10 28 11

madhusudhand:
@beafialho I was able to set different font sizes for variations. Now they should match the design spec.

Couple of places I could not set exact font sizes because size:small is used in the template and it cannot be overridden between variations.

So in Mono Bright they are set to 12px (small) instead of 14 (given in figma).

image

beafialho:
Thank you @madhusudhand. All paragraph sizes look fine. Let's leave those other places you mentioned with "Small" size as it looks equally good 👍 

A few comments:
- After taking a look at all the variations, let's keep "Forest Green" out for now as it lacks uniqueness. It's better to have two very unique ones.

- Buttons have different sizes when they're Filled or Outlined in "Mono Bright", they should have the same dimensions:

Captura de ecrã 2022-05-19, às 12 38 41

- Can we apply each variation's tertiary colors for highlighted text?

Captura de ecrã 2022-05-19, às 12 41 27

Captura de ecrã 2022-05-19, às 12 41 09

So they'll look like this:
Captura de ecrã 2022-05-19, às 12 50 34

Captura de ecrã 2022-05-19, às 12 50 39

- Links default color in "Mono Bright" should be yellow, instead of white:

https://user-images.githubusercontent.com/65220155/169286925-09ccc54f-7c7f-4309-8650-4858b4776848.mp4

Captura de ecrã 2022-05-19, às 12 52 28

madhusudhand:
@beafialho updated the PR.

> let's keep "Forest Green" out for now as it lacks uniqueness.

Removed Forest green and related fonts.

> Buttons have different sizes when they're Filled or Outlined in "Mono Bright", they should have the same dimensions:

Fixed by adding a custom style. I will have to draft an issue in GB to avoid such issues in future themes.

> Can we apply each variation's tertiary colors for highlighted text?

Fixed. Remote doesn't have a tertiary until now. introduced the color and applied it to text selection.

> Links default color in "Mono Bright" should be yellow, instead of white:

This seems to be a bit anti pattern with respect to other variations. Other used `foreground` color for link text and `primary` color on `hover`. We will not be override the color choices for a specific sytle variation. 
(This is because these are not controlled using theme.json yet. Should be possible in future if GB allows link style overrides through theme.json).

MaggieCabrera:
> This seems to be a bit anti pattern with respect to other variations. Other used `foreground` color for link text and `primary` color on `hover`. We will not be override the color choices for a specific sytle variation. (This is because these are not controlled using theme.json yet. Should be possible in future if GB allows link style overrides through theme.json).

We should be able to control de color of the link via theme.json [like this](https://github.com/Automattic/themes/blob/cf7e17f1fdde6f7efad85d747ad9188ebf0b08d8/livro/theme.json#L241). The hover is another matter for now, that would need CSS.

madhusudhand:
> We should be able to control de color of the link via theme.json [like this](https://github.com/Automattic/themes/blob/cf7e17f1fdde6f7efad85d747ad9188ebf0b08d8/livro/theme.json#L241). The hover is another matter for now, that would need CSS.

Ah Right. Missed that. Thanks for the suggestion.
Link works now as intended.

beafialho:
Thanks @madhusudhand! I'm afraid I can't see the changes even though I've cleared cache and refetched the PR.

- I see "Forest Green" there still
- I can't see the tertiary color when I select text
- Button dimensions aren't identical
- Can't see updated links either

Could you post screenshots of these, please?

I also have some other comments:

- Comment buttons seem to be using a system font:
Captura de ecrã 2022-05-24, às 11 46 44

- Post titles should use IBM Plex Sans Extra Light, as H1. I believe they were showing up correctly before.
- Headings font settings (line heights and H5/H6 font families) aren't correct. Please re-check, according to Figma:
Captura de ecrã 2022-05-24, às 11 57 29

madhusudhand:
@beafialho I did a force push because I rebased changes with trunk. Not sure if this could be the issue. Can you try fetching to a fresh branch and see if the changes are reflecting.

I updated the requested typography styles and button font. Attaching the screenshots below.

> I see "Forest Green" there still
style-variations

> I can't see the tertiary color when I select text
image

> Button dimensions aren't identical
image
image

> Can't see updated links either
image

beafialho:
Thank you @madhusudhand! I see it now, seems like most of it is accurate now.

The only thing I see that needs fine tuning is:
- Image captions aren't readable, which also happens in other variations. Caption color should be the same as text of each variation.

Captura de ecrã 2022-06-01, às 11 07 46
Captura de ecrã 2022-06-01, às 11 15 29
Captura de ecrã 2022-06-01, às 11 16 16

madhusudhand:
@beafialho Thanks again for validating. Gutenberg seems to be setting the caption color to `#555` without inheriting from body. I have set the style with foreground color. 

However because the caption sits inside image filter, it takes filter for color. (latest GB image block) Following is the result.

image
image
image

Do consider this as expected behaviour? or we should create an issue for GB.

beafialho:
Thank you @madhusudhand. These look good.

> Do consider this as expected behaviour?

If the caption sits outside the image (and video), as in the screenshots above, I'd expect it to inherit the body's color. 

If it sits on top of the image, then I'd expect it to be `#FFFFFF` or `#000000` according to the image overlay (if it's light or dark), was that what you're referring to?

madhusudhand:
Thanks for the inputs. The styling is not in the control of the theme. I will create an issue in GB.

beafialho:
Thank you!

pbking:
Would it behoove us to put a commented CSS fix in until the Gutenberg fix is addressed?
```
.wp-block-image figcaption {
  color: var(--wp--preset--color--foreground);
}
```
or something like that?  That should work across all of the variations.

[EDIT]. Well I'm silly.  I see that that is exactly what you have already done @madhusudhand 

Everything noted above seems to be addressed.  I'm marking this as ready to 🚢

pbking (approved):

PR #5936, opened 2022-04-25
Update Create Theme Checklist

jffng:
This PR cleans up the create theme checklist a bit, by refreshing the list of issues that are created and introducing some logic to avoid [secondary rate limits](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#secondary-rate-limits). According to github's [best practices](https://docs.github.com/en/rest/guides/best-practices-for-integrators): 

> If you're making a large number of POST, PATCH, PUT, or DELETE requests for a single user or client ID, wait at least one second between each request.

I tailored the issues slightly after reviewing the designs for Cult, but I think it can still apply for most new block themes.

### DIFF
--- a/create-a-theme-checklist.js
+++ b/create-a-theme-checklist.js
@@ -18,6 +18,7 @@ function sleep( ms ) {
 
 async function createLabel() {
 	try {
+		await sleep( 1000 );
 		return await octokit.request( 'POST /repos/Automattic/themes/labels', {
 			name: '[Theme] ' + theme,
 			color: 'c1f4a1',
@@ -30,6 +31,7 @@ async function createLabel() {
 
 async function createMilestone() {
 	try {
+		await sleep( 1000 );
 		return await octokit.request(
 			'POST /repos/Automattic/themes/milestones',
 			{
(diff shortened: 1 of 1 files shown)


create-a-theme-checklist.js
```diff
@@ -30,69 +40,78 @@ async function createMilestone() {
...
 					milestone: milestoneNumber,
-				})
-				.then((issueData) => {
-					addIssueToProject(issueData);
-				});
-		} catch (error) {
-			console.log(error);
+				}
+			);
+			// Uncomment the following if you want to add the issues to the project by default.
```
jffng: I don't think the default behavior to add to the board makes sense anymore, so I commented this out...

jffng:
Closing

create-a-theme-checklist.js
```diff
@@ -30,69 +40,78 @@ async function createMilestone() {
...
-				})
-				.then((issueData) => {
-					addIssueToProject(issueData);
-				});
-		} catch (error) {
-			console.log(error);
+				}
+			);
+			// If you want to automatically add this to the Theme Dev Board, uncomment this.
+			// .then( ( issueData ) => {
```
jffng, line 86: I don't think the default behavior to add to the board makes sense anymore, so I commented this out...

pbking:
The changes to the tasks created make sense, I think it's a better starting point.

I would definitely urge anyone running the script to customize that aspect to their liking and the needs of the theme (even if it doesn't get pushed into a change) but I like how that list is evolving.

DEFINITELY keen to remove the 'add to the board' bit.  There's still a fair bit more to do to the issues (most of them anyway) before they are really ready for "ToDo".

These changes LGTM.  I can't validate by running the script but assuming these changes were used for [cult](https://github.com/Automattic/themes/labels/%5BTheme%5D%20Cult) I say bring it in.

pbking (approved):

jffng:
> I would definitely urge anyone running the script to customize that aspect to their liking and the needs of the theme (even if it doesn't get pushed into a change)

💯 

> DEFINITELY keen to remove the 'add to the board' bit. There's still a fair bit more to do to the issues (most of them anyway) before they are really ready for "ToDo".

Yeah I left it in there for now because it's a nice reference in case you do want to automate some bit of that, but doesn't really make sense for this purpose.

PR #5938, opened 2022-04-25
Stewart: removing no needed definitions in theme.json

### LINKED ISSUE #5728: Implement new separator options

matiasbenedetto:
Stewart: removing no needed definitions about separator block in theme.json 

#### Related issue(s):
https://github.com/Automattic/themes/issues/5728

### DIFF
--- a/stewart/theme.json
+++ b/stewart/theme.json
@@ -285,8 +285,6 @@
 			},
 			"core/separator": {
 				"border": {
-					"color": "currentColor",
-					"style": "solid",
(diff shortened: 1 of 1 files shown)


MaggieCabrera (approved):
LGTM! I think we should do this for block canvas too, since we are using it as a base for new themes, but that can be done in a different PR

PR #5941, opened 2022-04-26
Better handle updating a theme with no readme.txt file

pbking:
When a theme doesn't have a readme.txt file (many of the wpcom-only themes such as varia children don't) the deploy script crashed.

This change checks for the existence of the file before trying to write to it.  If it's not there it just moves on to the next one.


pbking:
Updated script was used to deploy this: https://github.com/Automattic/themes/releases/tag/v1.0.114

jffng (approved):
LGTM, tested and verified this fixes the issue with themes with no readme.txt.

PR #5942, opened 2022-04-26
Fix the 'grid-of-posts' pattern so that it can be dropped in and used

pbking:
The pattern originally had 'inherit:true' which I thought was necessary for it to be used in a page like "search".  

However, it turns out that the 'inherity: true' caused the pattern to not work as expected when dropped into a page.

Also, it turns out that 'inherit: true' also isn't necessary to be used in the 'search' page template either.  

This just makes it 'false' so that the pattern can be dropped in a user-created page.

### DIFF
--- a/pendant/patterns/grid-of-posts.php
+++ b/pendant/patterns/grid-of-posts.php
@@ -6,7 +6,7 @@
  */
 ?>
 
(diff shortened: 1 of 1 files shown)


pbking:
@beafialho this was what was preventing the pattern from being used in [Articles](https://pendantdemo.wordpress.com/articles/) page.  With this in place we should be able to finish that page.

beafialho:
Thank you, the page works as expected!

PR #5943, opened 2022-04-26
Fix/pendant pattern meta

### LINKED ISSUE #5860: Pendant: Correct Pattern Categories
None of the patterns seem to be put into the correct categories.  And there is currently no category creation logic to support that (such as [this](https://github.com/Automattic/themes/blob/trunk/archeo/functions.php#L70) in Archeo)

### LINKED ISSUE #5884: Pendant: Block Patterns
The patterns showing up for me aren't the same ones specified on [Figma](https://www.figma.com/file/JdntsOfLqo8Vy0d5uTWHTI/Pendant?node-id=661%3A149). Here are the ones showing up:

Captura de ecrã 2022-04-19, às 11 04 56

Captura de ecrã 2022-04-19, às 11 04 34

Missing are:
- Our Picks (still needs refinement #5863)
- Subscribe
- Gallery

Let's not include as patterns:
- Post navigation
- Grid of posts
- Title
- Header and footer (since we already have plenty of them from other themes and will soon migrate all to the pattern directory)

pbking:
Closes: #5884 
Closes: #5860

Since the header isn't expected to be a pattern and there is nothing in it to localize this has been moved to the template instead.  (The footer, however, does have text to be localized so that remains a pattern, however it is hidden from the inserter).

The header was also refactored to be WIDE width instead of FULL width.  (I believe that was the original intent after looking at the designs more closely.  Please correct me if I'm mistaken.)

It now looks like this on very wide screens:

image

The rest of the patterns available in the inserter are as follows:

<hr>

image

You [mentioned](https://github.com/Automattic/themes/issues/5884#issue-1208130435) not including this one, however if we don't make this available to users then the [Articles](PdS3dp-M-p2) page would be hard to build.

<hr>

image

This one wasn't mentioned to include or not.  It was a complex bit to layout so including it as a pattern seems helpful.  LMK if you would rather me remove it from the inserter though.

<hr>

image

<hr>

image

<hr>

image

<hr>

image

<hr>

image

<hr>

image

<hr>

image

<hr>

At the moment ALL of the above patterns are exclusively in the 'Pendant' category.  Do we want to keep all of these in that category?  Some of them?

I assume we'll want to put them all into other categories (too) though.

I don't really know where they should go though...  Ideas?

Grid of Posts: Query
Latest Articles: Query
Home Page Hero: ?
Media & Text: ?
Image & Quote: ?
Gallery: Images
Our Picks: Images
Posts and Topics: ?
The Team: ?


beafialho:
> The header was also refactored to be WIDE width instead of FULL width. (I believe that was the original intent after looking at the designs more closely. Please correct me if I'm mistaken.)

Apologies as this may have been unclear and I missed it earlier: both the header and the footer should be full-width. I've [mocked up](https://www.figma.com/file/JdntsOfLqo8Vy0d5uTWHTI/Pendant?node-id=1858%3A71) what it should look like in a 2000px screen. Does that make sense?

2000px

> You https://github.com/Automattic/themes/issues/5884#issue-1208130435 not including this one, however if we don't make this available to users then the [Articles](https://github.com/Automattic/themes/pull/PdS3dp-M-p2) page would be hard to build.

Let's then keep the grid of posts and homepage hero pattern 👍 

> At the moment ALL of the above patterns are exclusively in the 'Pendant' category. Do we want to keep all of these in that category? Some of them?

I think it makes sense to keep them all in the Pendant category, except for the header and footer.

> I don't really know where they should go though... Ideas?

What do you think of these?

Home Page Hero: Call to Action
Media & Text: Media
Image & Quote: Quotes
Posts and Topics: List
The Team: List

pbking:
> Apologies as this may have been unclear and I missed it earlier: both the header and the footer should be full-width. I've [mocked up](JdntsOfLqo8Vy0d5uTWHTI-fi-1858%3A71) what it should look like in a 2000px screen. Does that make sense?

No apologies needed.  I should have clarified that earlier when I first had doubts.  That makes sense and I'm happy to oblige.

pbking:
Re-aligned:
localhost_4759__p=9 (1)

pendant/patterns/hero.php
```diff
@@ -1,7 +1,8 @@
 <?php
 /**
- * Title: Home page hero
+ * Title: Home Page Hero
  * slug: pendant/home-hero
+ * categories: pendant
```
madhusudhand: I guess we are not adding theme based categories.
this can be `featured, text`

pendant/patterns/image-and-quote.php
```diff
@@ -6,8 +6,8 @@
  */
```
madhusudhand, line 6: featured, text, columns

pendant/patterns/image-gallery.php
```diff
@@ -1,7 +1,8 @@
 <?php
 /**
- * Title: Image Gallery
+ * Title: Gallery
  * slug: pendant/image-gallery
+ * categories: pendant
```
madhusudhand, line 5: featured, gallery

pendant/patterns/image-series.php
```diff
@@ -1,7 +1,8 @@
 <?php
 /**
- * Title: Image Series
+ * Title: Our Picks
  * slug: pendant/image-series
+ * categories: pendant
```
madhusudhand, line 5: featured

pendant/patterns/media-and-text.php
```diff
@@ -1,7 +1,7 @@
 <?php
 /**
- * Title: About
- * slug: pendant/about
+ * Title: Media & Text 
+ * slug: pendant/media-and-text
  * categories: pendant
```
madhusudhand, line 5: featured, text

pendant/patterns/post-topics.php
```diff
@@ -2,6 +2,7 @@
 /**
  * Title: Posts and Topics
  * slug: pendant/posts-topics
+ * categories: pendant
```
madhusudhand, line 5: featured, text

madhusudhand (changes requested):

madhusudhand:
Here is what I think

Grid of Posts: query
Latest Articles: featured, query
Team: featured, columns
Posts and Topics: featured, text
Media & Text: featured, text
Image Series: featured (probably images as well)
Image Gallery: featured, gallery
Image with Quote: featured, text, columns
Hero: featured, text

pendant/patterns/hero.php
madhusudhand: posted summary of all patterns in PR comment.

pbking: Thank you, I have adjusted these based on both recommendations.

pbking:
I have adjusted the pattern categories based on the recommendations above.  'Pendant' has been replaced with 'Featured' and the other changes noted have been made (with a few exceptions).  'Columns' wasn't a used category that I could find so I haven't leveraged that one.

PR #5944, opened 2022-04-26
Added 'block-patterns' to tags

### LINKED ISSUE #5689: Pendant: Pre-launch: style.css tags

pbking:
Closes: #5689

This change adds 'block-patterns' to the tags.

The list then matches Archeo's tags.  I'm not sure what else is needed?  We good here?

### DIFF
--- a/pendant/style.css
+++ b/pendant/style.css
@@ -12,7 +12,7 @@ License: GNU General Public License v2 or later
 License URI: https://raw.githubusercontent.com/Automattic/themes/trunk/LICENSE
 Template: 
 Text Domain: pendant
-Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, featured-images, full-site-editing, rtl-language-support, theme-options, threaded-comments, translation-ready, wide-blocks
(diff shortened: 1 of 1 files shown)


MaggieCabrera (approved):
I don't think we need anything else, this looks good

madhusudhand:
Just curious, is the tag `custom-colors` for the themes with style variations?

MaggieCabrera:
https://make.wordpress.org/themes/handbook/review/required/theme-tags/

> Ability to customize colors from theme options(customizer) Important, the color customization must be something different from custom background implementation

I guess we should probably remove it then! We've used it in all our other block themes though

pbking:
While we can't use the customizer to customize colors I would still argue the tag is relevant and appropriate.  The theme is built is such a way that the user can customize the colors site-wide (even without style variations).

It being a block theme (and having the ability to change/modify colors) doesn't necessarily make that tag true... I think it's important that we leverage these defined colors in a way that makes customizing the site's colors fit with the expectation of "custom-colors".  Which we do on all of our themes and makes that tag relevant IMO.

PR #5945, opened 2022-04-26
Vertically Centered navigation.  Cleaned up style comments.

### LINKED ISSUE #5865: Pendant: Adjust spacing and alignment in desktop menu
The desktop menu should be horizontally centered and the space between items should be tighter. 

| Currently      | Intended |
| ----------- | ----------- |
|  Captura de ecrã 2022-04-14, às 09 56 35   |    Captura de ecrã 2022-04-14, às 09 58 02   |

Also, when I test the theme locally, I see this menu white by default instead of black. Any idea why?
Captura de ecrã 2022-04-14, às 10 06 21

pbking:
This change centers vertically the responsive navigation.

Fixes: #5865
Before:
image
image

After:
image
image

I was afraid that navigation elements that exceed the height would start "scrolled offscreen" but things appear to be working as desired:

After:
image
image

### DIFF
--- a/pendant/style.css
+++ b/pendant/style.css
@@ -225,14 +225,24 @@ body > .is-root-container > .wp-block-template-part > .wp-block-cover,
 	margin: 0 auto;
 	max-width: 820px;
 	height: 100%;
-	padding-top: 0;
+	padding-top: 2rem;
+	padding-bottom: 2rem;
 }
+
 .wp-block-navigation .wp-block-navigation__responsive-dialog,
 .wp-block-navigation .wp-block-navigation__responsive-close {
 	height: 100%;
 }
 .wp-block-navigation.is-responsive .is-menu-open.wp-block-navigation__responsive-container .wp-block-navigation__container {
 	justify-content: center;
+}
+.wp-block-navigation .wp-block-navigation__responsive-container-close {
+	left: 0;
+}
+.wp-block-navigation .wp-block-navigation__responsive-container.is-menu-open {
+	padding-left: var(--wp--custom--gap--horizontal);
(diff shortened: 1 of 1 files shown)


madhusudhand:
Verified the layout. It works good except for a small observation.

Close icon (x) is on the left in figma designs and I guess it makes sense to keep it on left because in this theme hamburger menu is also on the left. Clicking on left corner to open the menu and click on right corner to close feels not right.

pbking:
@beafialho do you have an opinion on the above comment? RE: placement of the close icon?

The implementation is the default behavior of Gutenberg.  However I believe I could override that with CSS if that's an important design decision.

One difficulty doing that is that we would have to re-introduce a top-padding so that the X and menu items wouldn't overlap and that might shift the vertical centering down a bit.

beafialho:
@pbking yes, now that I look at it I do think it's confusing to have the hamburger menu on the left and the `X` on the right. If we could work that around for this to work it would be great.

> One difficulty doing that is that we would have to re-introduce a top-padding so that the X and menu items wouldn't overlap and that might shift the vertical centering down a bit.

That top padding would need to match the header when the menu isn't open:

Captura de ecrã 2022-04-27, às 15 59 05

pbking:
I've adjusted the X arrow to align, but it will only align when the user does not have the wp-header:
Apr-28-2022 10-04-13
Apr-28-2022 10-09-15

I think I was able to introduce the needed padding to keep the menu away from the close button and still keep things centered vertically.

image

image

pbking:
This checks all the boxes.  Going to bring this in.  Further issues with navigation can be opened as additional items.

PR #5946, opened 2022-04-27
Cult: Fonts and Weights

### LINKED ISSUE #5923: Cult: Styles: Typography — Fonts & Weights
Load all fonts and weights used in the theme:

- Work Sans
- Petrona

MaggieCabrera:

This PR includes the font files and assigns proper values to the headings and different font sizes of text for the site. I also fixed the medium preset variable on block canvas that was missing a dash.

Screenshot 2022-04-27 at 11 14 00

#### Related issue(s):

Closes #5923


cult/theme.json
```diff
@@ -61,36 +61,90 @@
...
+							"fontStretch": "normal",
+							"src": [
+								"file:./assets/fonts/WorkSans-Italic.ttf"
+							]
+						}
+					]
 				}
 			],
 			"fontSizes": [
+				{
```
jffng: It looks like we are pretty inconsistent in the number of font sizes and their slugs:

[Archeo](https://github.com/Automattic/themes/blob/trunk/archeo/theme.json#L127-L153)
[Livro](https://github.com/Automattic/themes/blob/trunk/livro/theme.json#L104-L125)
[Pendant](https://github.com/Automattic/themes/blob/trunk/pendant/theme.json#L160-L186)

Should we try to tighten that up or is it inconsequential? I guess it has an impact if a user switches themes.

jffng (commented):
Left one question, other than that LGTM.

cult/theme.json
pbking: Indeed. I had the same thought when adding "Extra Small" to Pendant and we've definitely discussed previously.

It feels wrong to limit designs to the core sizes.  It's definitely not enough.

It feels cumbersome to leverage sizes outside of defined sizes.  

The '[custom](https://github.com/Automattic/themes/blob/trunk/blockbase/theme.json#L175)' font sizes in Blockbase do this so we can use a variable in the theme.json or css but then not allow a user to select it.  We can't leverage those sizes in patterns though 'cause it's the same problem.  But that feels icky too.

Providing users with font sizes that might go away if they switch themes is ichy.

At a minimum we need to consolidate on our naming outside of the norm.  Is it "tiny" or "extra small".  At least then switching amongst our themes will yield the same results.

At a maximum we need to remove any font size outside of what's provided by core and just use hard-coded sizes instead of those variables.

jffng: > At a maximum we need to remove any font size outside of what's provided by core and just use hard-coded sizes instead of those variables.

I definitely flip flop and can see the value to providing more presets than what core provides. But generally my opinion falls with the above, since it keeps things more consistent, portable, and predictable. It provides for a nicer UI as well: 

Core supplied | Extra presets
--- | ---
Screen Shot 2022-04-27 at 10 50 53 AM | Screen Shot 2022-04-27 at 10 51 13 AM

jffng (approved):

PR #5948, opened 2022-04-27
Appleton: Initial Commit

onuro:
This is the initial commit of the Appleton (Blockbase child) theme. This theme is a part of the verticals project (more info: pbxlJb-18I-p2).

My first premium theme, so I might be doing some things wrong :) Please correct me where I need to!

Thanks!

Homepage Example and [Draft Demo →](https://appleton105180900.wordpress.com/)

Appleton - Photographer


appleton/parts/header.html
```diff
@@ -0,0 +1,17 @@
...
+<div class="wp-block-group gapless-group"><!-- wp:group {"align":"full","style":{"spacing":{"padding":{"bottom":"var(--wp--custom--gap--vertical)","top":"var(--wp--custom--gap--vertical)"}}},"className":"site-header","layout":{"type":"flex","justifyContent":"space-between"}} -->
+<div class="wp-block-group alignfull site-header" style="padding-top:var(--wp--custom--gap--vertical);padding-bottom:var(--wp--custom--gap--vertical)"><!-- wp:group {"className":"site-brand","layout":{"type":"flex"}} -->
+<div class="wp-block-group site-brand"><!-- wp:site-logo {"width":64} /-->
+
+<!-- wp:group -->
+<div class="wp-block-group"><!-- wp:site-title /--></div>
+<!-- /wp:group --></div>
+<!-- /wp:group -->
+
+<!-- wp:navigation {"ref":22,"overlayMenu":"always","__unstableLocation":"primary","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /--></div>
```
pbking: The navigation block template part shouldn't have a ref id.  Otherwise the `__unstableLocation` bits won't work.

pbking:
The demo site looks smashing!  This theme will be really good when it matches.

A few things to point out about making it so:

The font weights represented in your demo don't seem to be present in theme.json (or the templates)

(my site)
image

For instance that `site-title` defaults to 700 weight but in your example site is 300.  That should be defined either in the theme.json (`{style.blocks.core/site-title.typography.fontWeight}`) for all instances of the site-title block or in the header template to effect only that one.

The same is true for font weights which are defaulting to 400 but your demo is lighter.

The footer in the demo looks really good, but should be represented in a template-part or pattern.

onuro:
@pbking thank you for the feedback! I've updated the typography settings rather on the template compared to site-wide weights since the layout is mostly built via custom patterns made for this theme. 

But if you think I should still apply those weights site-wide, I can do that too. Please let me know.

Also updated header/footer templates.

pbking:
@onuro I would suggest changing as much as you can at the theme level (meaning, in the theme.json configuration) rather than the pattern level.  In this way the patterns you create for this theme will be more likely to fit in any site's design (and likewise, other patterns more likely to fit into this theme's design).

If I create content (even just a post with headers and paragraphs) the result is noticeably different than if I exclusively use the patterns included.

onuro:
Thanks @pbking! I've updated the values as much as I can globally. I wanted to add uppercase to all headings but looks like it's not working for a headings setting yet.

Let me know if this works.

Thanks!

appleton/parts/header.html
```diff
@@ -0,0 +1,17 @@
...
+<div class="wp-block-group gapless-group"><!-- wp:group {"align":"full","style":{"spacing":{"padding":{"bottom":"20px","top":"40px","right":"4vw","left":"4vw"}}},"className":"site-header","layout":{"type":"flex","justifyContent":"space-between"}} -->
+<div class="wp-block-group alignfull site-header" style="padding-top:40px;padding-right:4vw;padding-bottom:20px;padding-left:4vw"><!-- wp:group {"className":"site-brand","layout":{"type":"flex"}} -->
+<div class="wp-block-group site-brand"><!-- wp:site-logo {"width":64} /-->
+
+<!-- wp:group -->
+<div class="wp-block-group"><!-- wp:site-title {"style":{"typography":{"textTransform":"uppercase","fontStyle":"normal","fontWeight":"300","fontSize":"24px","letterSpacing":"1.4px"}},"textColor":"custom-superdark"} /--></div>
+<!-- /wp:group --></div>
+<!-- /wp:group -->
+
+<!-- wp:navigation {"ref":22,"overlayMenu":"always","__unstableLocation":"primary","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /--></div>
```
mikachan: It looks like the ref id still needs to be removed here:

```suggestion
<!-- wp:navigation {"overlayMenu":"always","__unstableLocation":"primary","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /--></div>
```

appleton/parts/footer.html
```diff
@@ -0,0 +1,39 @@
...
+<div class="wp-block-group has-text-color has-background" style="background-color:#160e0e;color:#fffffe;font-size:15px;padding-top:6vw;padding-right:30px;padding-bottom:6vw;padding-left:30px"><!-- wp:columns {"align":"wide"} -->
+<div class="wp-block-columns alignwide"><!-- wp:column {"style":{"spacing":{"padding":{"top":"20px","right":"20px","bottom":"20px","left":"20px"}}}} -->
+<div class="wp-block-column" style="padding-top:20px;padding-right:20px;padding-bottom:20px;padding-left:20px"><!-- wp:group {"style":{"spacing":{"blockGap":{"top":"20px","left":"20px"}}}} -->
+<div class="wp-block-group"><!-- wp:site-title {"style":{"elements":{"link":{"color":{"text":"var:preset|color|background"}}},"typography":{"fontStyle":"normal","fontWeight":"600","fontSize":"1.2rem","textTransform":"uppercase"}},"textColor":"background"} /-->
+
+<!-- wp:paragraph -->
+<p>Photography portfolio website of a nature photographer,&nbsp;<strong>John Appleton</strong>.</p>
```
mikachan: These strings should be translatable. You need to use PHP for this, so I'd recommend creating a pattern (PHP file) for this footer, and then you can wrap the strings with `esc_html_e`, like this: `<?php esc_html_e( 'Text goes here.', 'appleton' ); ?>`

Example pattern in Archeo: https://github.com/Automattic/themes/blob/trunk/archeo/patterns/hidden-404.php#L14
Here's the pattern being used: https://github.com/Automattic/themes/blob/trunk/archeo/templates/404.html#L6

This `footer.html` would then just contain something like this, similar to Archeo's [footer](https://github.com/Automattic/themes/blob/trunk/archeo/parts/footer.html):

```
<!-- wp:pattern {"slug":"appleton/footer"} /-->
```

mikachan (commented):
This is looking good! I've left a couple of comments below.

Also, the screenshot has a pretty large file size at 1.25MB. Screenshot files are usually around or under 100KB. I don't think there's a rule for this, but I think it would be good if you could compress this some more.

appleton/parts/footer.html
onuro: Thanks! updated according to reference theme.

appleton/parts/header.html
onuro: Yeah sorry I missed it after the second export. Fixed now.

mikachan (approved):
Thanks, those changes look great! I think this is ready to bring in 🎉

PR #5955, opened 2022-04-27
Update header pattern loading

colinduwe:
Changed from get_template_directory() to get_theme_file_path() so that headers in the child theme override Blockbase.

#### Related issue(s):
https://github.com/Automattic/themes/issues/5956

### DIFF
--- a/blockbase/inc/block-patterns.php
+++ b/blockbase/inc/block-patterns.php
@@ -52,7 +52,7 @@ if ( ! function_exists( 'blockbase_register_block_patterns' ) ) :
 						'title'      => __( 'Blockbase Header (' . $header_pattern . ')', 'blockbase' ),
 						'categories' => array( 'header' ),
 						'blockTypes' => array( 'core/template-part/header' ),
-						'content'    => file_get_contents (get_template_directory() . '/block-template-parts/header-' . $header_pattern . '.html'),
+						'content'    => file_get_contents (get_theme_file_path( '/block-template-parts/header-' . $header_pattern . '.html' )),
 					)
 				);
 			}


MaggieCabrera:
Sorry for the long wait! I tested this and this fails if the child is not registering a duplicate of the parent's pattern:

`Warning: file_get_contents(/Users/maggie/a8c/Local sites/maggie/app/public/wp-content/themes/a8c-themes/quadrat/block-template-parts/header-centered.html): failed to open stream: No such file or directory in /Users/maggie/a8c/repos/themes/blockbase/inc/block-patterns.php on line 55`

colinduwe:
@MaggieCabrera Thanks for taking a look. I can't seem to generate that warning. 
Here's the bit of interest (Line 55): 
`'content'    => file_get_contents (get_theme_file_path( '/block-template-parts/header-' . $header_pattern . '.html'))`
[get_theme_file_path()](https://developer.wordpress.org/reference/functions/get_theme_file_path/) should never return the path to a non-existent file in the child theme directory. If it doesn't exist in the child theme dir it should return the path to the file in the parent theme dir (without checking if the parent file exists)
I wonder what's different between your environment and mine that would cause it to throw that warning for you.
For you `file_exists( get_stylesheet_directory() . '/block-template-parts/header-centered.htm')` is returning true but file_get_contents() can't open the file.

I'll keep pondering this and see if I can think of a reason and solution.

MaggieCabrera:
> 'content' => file_get_contents (get_theme_file_path( '/block-template-parts/header-' . $header_pattern . '.html'))

you are actually right, it was me doing it wrong on my end, this seems to be working correctly for me on Quadrat so I think we can bring this in

MaggieCabrera (approved):

PR #5957, opened 2022-04-28
Pendant: Fix site title font style

### LINKED ISSUE #5951: Pendant: Site title & categories font weight
I'm still seeing the site title and the categories in a different font weight than the mockup. It might be just me but I wanted to check: can you replicate?

| Test site | Figma |
| ----- | ----- |
| Captura de ecrã 2022-04-27, às 16 45 05 | Captura de ecrã 2022-04-27, às 16 46 51 |

madhusudhand:

Add new fontface and set font family to site-title

image

image

#### Related issue(s):

Fixes: #5951


madhusudhand:
Included only "Bold" font variant because the fontFamily is used only for site-title

pendant/theme.json
```diff
@@ -298,8 +315,10 @@
 			},
 			"core/site-title": {
 				"typography": {
+					"fontFamily": "var(--wp--preset--font-family--title-font)",
 					"fontSize": "var(--wp--preset--font-size--medium)",
 					"fontWeight": "900",
+					"textTransform": "uppercase",
```
pbking, line 321: I would rather leave off the 'uppercase'.  If a user really wants an uppercase title they can control that with their <shift> key, however if they don't with this in place they wouldn't have an option.

pbking:
Hey @beafialho, @madhusudhand rightly points out that the figma design uses Playfair for the site title typeface.  We've just been using the heading font (Literata).  It's a 194k file for the font so my 'druthers is to leave it with Literata but the call is yours to make.

The original issue doesn't show the site title correctly embolden due to the Gutenberg bug.  We decided not to release Pendant until that code has been released.

Playfair:
image

Literata (correctly rendered):
image

beafialho:
Oh no, sorry about the confusion! Playfair was set by mistake. The only fonts we should use for Pendant are Literata 72pt and Jost.

PR #5958, opened 2022-04-28
Added a command to push an individual theme to sandbox

pbking:
run the command:

```
npm run deploy:push:theme THEMEDIRECTORY
```

It works the same as `npm run deploy:push:changes` but doesn't calculate which themes have changed, it just pushes the theme defined.


jffng (approved):
Nice, works as advertised!

PR #5963, opened 2022-04-29
Re-export pendant theme.json with a more organized export logic

pbking:
This should have no actual change to the functionality (or actual configuration) of Pendant.  This just re-exports the theme.json using the improved logic found in [this change](https://github.com/Automattic/create-block-theme/pull/55) to the Create Block Theme plugin which mirrors the resources exported directly from core.

### DIFF
--- a/pendant/theme.json
+++ b/pendant/theme.json
@@ -1,71 +1,64 @@
 {
-	"version": 2,
-	"templateParts": [
-		{
-			"name": "header",
-			"area": "header"
-		},
-		{
-			"name": "footer",
-			"area": "footer"
-		}
-	],
 	"customTemplates": [
 		{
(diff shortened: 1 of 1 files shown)


pendant/theme.json
```diff
@@ -1,72 +1,65 @@
...
 			"postTypes": [
 				"page",
 				"post"
-			]
+			],
+			"title": "Footer Only"
 		}
 	],
 	"settings": {
+		"appearanceTools": true,
```
MaggieCabrera, line 29: looks like we didn't have this before, it's good that you did this then!

MaggieCabrera (approved):
LGTM! thanks for doing this, a sorted theme.json makes my brain happy

PR #5967, opened 2022-05-03
Set Barnsbury to show footer widgets in 2 columns on smaller screens

### LINKED ISSUE #5132: Barnsbury: Widgets in footer appear in a single left-aligned column.
## Steps to replicate
1 - Activate Barnsbury
2 - Add a few widgets

## Result
The widgets are all lined up on the left side of the screen at 50% width

## Expected
Given that they are only 50% of the screen width, it looks like they were supposed to either sit in the center, or be arranged in some sort columns.
Screen Shot 2021-12-01 at 8 07 42 PM

A user in this thread tried to set up their own three-column widget using blocks, and that was squished to the left, too.
Screen Shot 2021-12-01 at 7 56 09 PM

* User report: https://wordpress.com/forums/topic/my-footer-isnt-full-width-and-im-not-sure-why/
* Follow-up ticket: ^ same

jordesign:

* Float widgets (which are already 50% in width) so that they arrange in 2 columns
* Add CSS rules for clearing floats and providing padding 

#### Screenshot:

Screen Shot 2022-05-03 at 9 13 48 am

#### Related issue(s):

Resolves https://github.com/Automattic/themes/issues/5132


jordesign:
Just a note on the actual CSS used...

I've taken the 50% width as a prompt to lay out the footer using floats...

But if using flex or grid layout would be better - I can rework the PR

.gitignore
```diff
@@ -369,3 +369,4 @@ wu-wei/
 yoko/
 zbench/
 zoren/
+package-lock.json
```
mikachan: We shouldn't ignore this file as we like to keep these in source control.

mikachan (commented):
Thanks for working on this!

Your solution doesn't seem to show the widgets in 2 columns for me. I'm using Chrome:

image

This seems to be because of the `clear: left` on this selector: `& > .widget:nth-child(2n+2)`, which is clearing the float on the second widget.

I understand why you used floats based on the 50% width setting. Based on this, you could also try setting each widget to `display: inline-block`, plus some padding rules for spacing:

```
.site-footer .widget-area {
  max-width: 100%;
  margin-left: -16px;
  margin-right: -16px;

  & > .widget {
	  width: 50%;
	  display: inline-block;
	  padding-left: 16px;
	  padding-right: 16px;
  }
}
```

Alternatively, I'd probably suggest using a grid layout to achieve this with minimal code. Something like this should work:

```
.site-footer .widget-area {
  max-width: 100%;
  display: grid;
  grid-template-columns: 1fr 1fr;
  column-gap: 32px;
  
  & > .widget {
	  margin-top: 0;
  }
}
```

This is how this is working: 
- `display: grid` on the container sets a grid layout
- `grid-template-columns: 1fr 1fr` sets a 2 column layout
- `column-gap: 32px` sets the gap between the columns
- I've also removed the top margin on each item, as this was creating uneven spacing, plus we don't need to set a `width` as this is covered by `grid-template-columns`

jordesign:
Thanks Mika - definitely worth me doing it with CSS grid - I'll rework it :)

jordesign:
Alrighty - thanks for the guidance @mikachan  - I *think* both those are resolved now

mikachan (commented):
These changes look great! Just one more thing - it looks like you probably need to run `npm run build` again in the Barnsbury directory. I can see a double `max-width` in the compiled CSS, and when I run the build locally it removes this (and adds a few other recent changes from the linting).

Or let me know if you want me to commit my compiled CSS from your changes - happy either way!

jordesign:
Ah yes - I did indeed.

That's done now. Thanks again @mikachan

mikachan (approved):
LGTM! 🚢

jordesign:
Huzzah - thank you. @MaggieCabrera @mikachan  - is this the point I need to leave things to you folks to merge and deploy?

mikachan:
> is this the point I need to leave things to you folks to merge and deploy?

Yes that's right!

PR #5973, opened 2022-05-05
Curator: Add button styles

### LINKED ISSUE #5919: Curator: Styles: Buttons
Add default button styles for the theme and ensure they are applied correctly:

- [x] Button block
- [x] Search (in this case, I don't think we want the default button styles to apply to the search button)

madhusudhand:

Button styles for the following

### Filled, Outline buttons and Comments button

**Normal**

image
image

**Hover**

image
image

**Active**

image
image

**Focus**

image
image

### Search and file buttons

**Normal**

image

**Hover**

image

**Active**

image

**Focus**

image

#### Related issue(s):

Fixes: #5919

### DIFF
--- a/curator/style.css
+++ b/curator/style.css
@@ -28,10 +28,51 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
  * https://github.com/WordPress/gutenberg/issues/27075
  */
 
-.wp-block-search__button:hover,
-.wp-block-file .wp-block-file__button:hover,
+.wp-block-button:not(.is-style-outline) .wp-block-button__link {
+	padding: 0;
+}
 .wp-block-button__link:hover {
-	background-color: var(--wp--preset--color--secondary);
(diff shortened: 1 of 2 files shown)


jffng (approved):
Gave this a pretty thorough test and matches the design as far as I can tell. Implementation looks good too. 

The design suggests all the buttons be in the uppercase, but I don't think we should enforce that via CSS. I realized that option is not available in the button block settings, and I think it should be so I'll open up a PR for it.

Thanks @madhusudhand !

jffng:
https://github.com/WordPress/gutenberg/pull/40858

beafialho:
> The design suggests all the buttons be in the uppercase, but I don't think we should enforce that via CSS. I realized that option is not available in the button block settings, and I think it should be so I'll open up a PR for it.

Thanks, that option should be available IMO

Regarding the button states, I gave it a look and will open up an issue about them. There are some changes I'd like to make.

PR #5980, opened 2022-05-06
Curator: revise header template part

### LINKED ISSUE #5925: Curator: Templates: Header template part
Revise the header template part, verify it matches the design.

jffng:

This PR makes adjustments to the header part and associated styles.

Before | After
---- | ----
Screen Shot 2022-05-06 at 1 57 40 PM | Screen Shot 2022-05-06 at 1 57 28 PM

I'm not sure why the color of the style title is not being correctly applied. 

#### Related issue(s):

Fixes #5925


madhusudhand:
**Design:**

image

**frontend:**

image

**FSE:**

image

Following are the observations:

* site title color should be "foreground" (black)
* text is not vertically centered. (this is because of the 9px margin of separator block)
* separator block is not full width in FSE

curator/parts/header.html
```diff
@@ -1,29 +1,13 @@
...
-	
-		<!-- wp:navigation {"__unstableLocation":"primary","__unstableSocialLinks":"social","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /-->
-	
-	</div>
-	<!-- /wp:group -->
+ <!-- wp:group {"align":"full","style":{"spacing":{"padding":{"bottom":"var(\u002d\u002dwp\u002d\u002dcustom\u002d\u002dgap\u002d\u002dvertical)","top":"var(\u002d\u002dwp\u002d\u002dcustom\u002d\u002dgap\u002d\u002dvertical)"}}},"layout":{"type":"flex","justifyContent":"space-between"}} -->
+<div class="wp-block-group alignfull" style="padding-top:var(--wp--custom--gap--vertical);padding-bottom:var(--wp--custom--gap--vertical)"><!-- wp:site-title {"style":{"typography":{"fontSize":"1.375rem"}},"textColor":"foreground"} /-->
 
-</div>
+<!-- wp:navigation {"__unstableLocation":"primary","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /--></div>
```
MaggieCabrera: Do we need unstable location for a non universal theme? I don't see we've used it on archeo or remote for example

jffng: Could it still be useful for someone switching to this theme from a non-block theme that has a primary navigation defined? 

If this is not a likely scenario or it doesn't work this way anymore, I think we can remove it.

jffng:
> site title color should be "foreground" (black)

Fixed in https://github.com/Automattic/themes/pull/5980/commits/cd887fc740f64a20e166fd705e8dd60fceacd5ce

> text is not vertically centered. (this is because of the 9px margin of separator block)

Fixed in https://github.com/Automattic/themes/pull/5980/commits/9714bea0c80485e59f5f360c8857b908127f622b.

> separator block is not full width in FSE

This is due to a bug upstream. I opened a fix here: https://github.com/WordPress/gutenberg/pull/40937

jffng:
This is ready for another review. I think I addressed everything except the separator full width issue, which needs to be resolved upstream.

I'm addressing the mobile menu styling in a separate PR: https://github.com/Automattic/themes/pull/5988

beafialho:
The color of the Site title shows up correctly for me 👍 A few comments:

- The Site Title is underlined, and it should be underlined only on hover.

- I'm not sure if it's related to the header template part, but the single post should literally stick to the separator, without the white spacing:

Captura de ecrã 2022-05-27, às 11 17 08

- This doesn't look ideal. The stroke should at least be the same color. We'd have to add CSS to customize this as this is GB's default right?
Captura de ecrã 2022-05-27, às 11 20 54

madhusudhand:
> The Site Title is underlined, and it should be underlined only on hover.

curator-header-hover

> I'm not sure if it's related to the header template part, but the single post should literally stick to the separator, without the white spacing:

Issue is with Single post template. Header is occupying right height.

image

> This doesn't look ideal. The stroke should at least be the same color. We'd have to add CSS to customize this as this is GB's default right?

Right, it is GB default. We would need custom styles to override the behaviour. Following is the result with border color set to foreground color.

image

Is this looking fine?

jffng:
> I'm not sure if it's related to the header template part, but the single post should literally stick to the separator, without the white spacing:

Fixed in https://github.com/Automattic/themes/pull/5980/commits/c1eedfd5bc3a473b289ca6a1203837422315e9a6

The rest of the feedback I think has been addressed, so I think this is in a good place to be brought in and we can iterate. If there's more feedback on the nav submenu we can address in a follow up. Thoughts?

jffng:
Going to bring this in since it will make other template PRs easier to review from a design perspective. Can address additional feedback in follow ups, thanks.

Issue #3080, opened 2021-01-21
Canard: Blog Posts block - "Show Author" settings doesn't work as expected


ivan-ottinger:
## Steps to replicate

1. Activate the _Canard_ theme.
2. Add a _Blog posts_ block.
3. Enable the _Show Author_ setting.

## Result

Author name is displayed in the editor, but not in the front-end.

Backend:

![Markup on 2021-01-21 at 16:07:28](https://user-images.githubusercontent.com/25105483/105369472-dfd68500-5c02-11eb-92e9-def88f52a241.png)

Front-end:

![Markup on 2021-01-21 at 16:07:08](https://user-images.githubusercontent.com/25105483/105369505-e95fed00-5c02-11eb-9f19-c90558430261.png)

## Expected

Author name should be displayed in the front-end as well.

* User report: 27165309-hc
* Follow-up ticket: 3669607-zd-woothemes

cecilearkay:
I think it goes beyond that, there must be some setting somewhere that's preventing the meta elements to be hidden. In my case, the user's front page is set to latest blog posts, like in the demo, as opposed to your case where you add the _Blog Posts_ block to a page.

If you look in the CSS any post is set to have `display:none` on several on the entry meta:

**Sticky posts:**
<img width="378" alt="Screenshot 2022-04-13 at 10 56 30" src="https://user-images.githubusercontent.com/16645645/163141442-94f6034a-e20f-443f-b1d0-f16457110e62.png">

**Regular posts:**
<img width="417" alt="Screenshot 2022-04-13 at 10 57 34" src="https://user-images.githubusercontent.com/16645645/163141530-91e6e938-2729-49bb-93d1-8ee32c6b44f2.png">

There's also this CSS that may be getting in the way:
<img width="406" alt="Screenshot 2022-04-13 at 10 58 12" src="https://user-images.githubusercontent.com/16645645/163141576-7441e404-a59b-476c-9a4f-81c839ecca79.png">

I checked the logs quickly for the theme and I couldn't find anything that would cause the issue - especially when it's working fine on the demo site for Canard (https://canarddemo.wordpress.com/):

<img width="963" alt="Screenshot 2022-04-13 at 11 10 46" src="https://user-images.githubusercontent.com/16645645/163146939-94462023-c789-4302-afdb-13d70d29ba02.png">

My thinking is maybe `display:none` was expected for Sticky posts? And that it got applied everywhere?

Edit: kudos to @kettultim for the detective work!

[...]

PR #5981, opened 2022-05-07
#3080: Canard: Fix byline in blog post block

bobmatyas:

Fix `.byline` display in blog post carousel block by adding exception to `display: none` for `.byline` for the blog carousel block. Displaying `.byline` by default doesn't work as it messes up the post listing and shows the author where it wasn't before.

##### Before

Screenshot on 2022-05-07 at 12-29-58

##### After

Screenshot on 2022-05-07 at 12-45-56

#### Related issue(s):

https://github.com/Automattic/themes/issues/3080

### DIFF
--- a/canard/style.css
+++ b/canard/style.css
@@ -1959,6 +1959,7 @@ a:visited {
 	display: none;
 }
 .group-blog .byline,
+.wpnbha .byline,
 .wpnbpc .byline {
 	display: inline;
 }
--- a/canard/style.css
+++ b/canard/style.css
@@ -1958,8 +1958,8 @@ a:visited {
 .updated:not(.published) {
 	display: none;
 }
-.wpnbpc .byline,
-.group-blog .byline {
+.group-blog .byline,
+.wpnbpc .byline {
 	display: inline;
 }
 .byline img {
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
This is not working for me, I think it's because the `wpnbpc` class does not exist on a single post (which is where I am testing)

bobmatyas:
Thanks for looking.

I was testing on a page, but I just created a new post to test with the blog posts block. It looks like it still works as the `wpnbpc` class is present for the carousel:

Screenshot on 2022-05-09 at 09-04-02

MaggieCabrera:
I'm using the Blog posts block (according to the testing instructions in https://github.com/Automattic/themes/issues/3080) and that uses the class `wpnbha` instead

You can see my test site here: pckFNO-mA-p2

bobmatyas:
Sorry about that, I had the wrong block. 🤦 I added code to fix it on the Blog Post block and also left the code to fix the Post Carousel block since they both had the issue.

Here's a screenshot of the change on the Blog Post block:

Screenshot on 2022-05-09 at 09-32-03

Thanks for your help here.

MaggieCabrera (approved):
This seems to be working fine now, thanks so much for contributing, @bobmatyas

PR #5982, opened 2022-05-09
Curator: update link and navigation styles

### LINKED ISSUE #5924: Curator: Styles: Links
Add styles for default, hover, focus, and active states for standard and navigation links.

### LINKED ISSUE #5926: Curator: Styles: Navigation

madhusudhand:

Link Normal:

image

Link Hover:

image

Link Focus:

image

Link Active:

image

Navigation Normal:

image

Navigation Hover:

image

Navigation Focus:

image

Navigation Active:

image

#### Related issue(s):

Fixes: #5924
Fixes: #5926

### DIFF
--- a/curator/style.css
+++ b/curator/style.css
@@ -149,7 +149,7 @@ a:not(
 	.wp-block-pages-list__item .wp-block-pages-list__item__link,
 	.wp-block-navigation-link__content
 ):active {
-	background-color: var(--wp--custom--color--link-background);
+	background-color: var(--wp--preset--color--tertiary);
 }
 
 
--- a/curator/style.css
+++ b/curator/style.css
@@ -93,6 +93,63 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
 }
 
 /*
+ * Link Styles
+ */
+
+a {
+	padding: .4rem 0;
+}
+
+a:not(
+	.wp-block-search__button,
+	.wp-block-button__link
+):hover {
+	text-decoration: none;
+}
+
+a:not(
+	.wp-block-search__button,
+	.wp-block-button__link
+):focus {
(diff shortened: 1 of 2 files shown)


curator/style.css
```diff
@@ -92,6 +92,63 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
 	padding: calc(0.667em + 2px) calc(1.333em + 2px);
 }
 
+/*
+ * Link Styles
+ */
+
+a {
+	padding: .4rem 0;
```
jffng, line 103: I think we should give links a bit of an offset: 

Current | Suggested
--- | ---
Screen Shot 2022-05-09 at 8 19 15 AM | Screen Shot 2022-05-09 at 8 19 05 AM

```suggestion
	padding: .4rem 0;
	text-underline-offset: .1rem;
```

curator/theme.json
```diff
@@ -58,6 +58,9 @@
 			]
 		},
 		"custom": {
+			"color": {
+				"link-background": "#F4ECEC"
```
jffng: Can we map this to the theme preset color variable instead? This way, if the user changes the theme's palette via global styles there, it will match.

```suggestion
				"link-background": "var(--wp--preset--color--tertiary)"
```

curator/style.css
madhusudhand: Updated the style.

curator/theme.json
madhusudhand: Makes sense.
Tertiary color in the figma seems to be using the color F4ECEC in the rectangle, but wrongly indicated a different lighter color in text (FCFAFA). Updated tertiary and used the same for link background.

jffng (approved):

PR #5984, opened 2022-05-09
Curator: adjust font styles

### LINKED ISSUE #5920: Curator: Styles: Typography — Sizes & Line Height
Update the font sizes and line heights for headings and paragraphs.

mikachan:
This PR adjusts a couple of font settings for Curator in alignment with the comps.

The main changes include:
- Setting individual line heights for the different headings
- Setting letter spacing for all headings
- Adjusting the responsive font sizes

I also checked the paragraph sizes, and these looked good to me.

#### Related issue(s):
Closes #5920

### DIFF
--- a/curator/theme.json
+++ b/curator/theme.json
@@ -322,28 +322,32 @@
 				"typography": {
 					"fontFamily": "var(--wp--preset--font-family--heading-font)",
 					"fontWeight": "900",
-					"fontSize": "clamp(5.875rem, 10vw, 3.375rem)"
+					"fontSize": "clamp(5.875rem, 10vw, 3.375rem)",
+					"letterSpacing": "-0.02em"
 				}
 			},
 			"h2": {
 				"typography": {
 					"fontFamily": "var(--wp--preset--font-family--heading-font)",
 					"fontWeight": "900",
-					"fontSize": "clamp(2.75rem, 10vw, 2rem)"
+					"fontSize": "clamp(2.75rem, 10vw, 2rem)",
+					"letterSpacing": "-0.02em"
 				}
 			},
 			"h3": {
 				"typography": {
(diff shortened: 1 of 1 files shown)


madhusudhand (commented):
LGTM!

For the paragraphs, the default font size is set to `1.125` and the same is also called `Medium`
Should we call it "Normal" instead. and rename other larger fonts. 

`Small -> Small`
`Medium -> Normal`
`Large -> Medium`
`Extra Large -> Large`

madhusudhand (approved):

mikachan:
> Should we call it "Normal" instead. and rename other larger fonts.

Yeah I think this is a good idea! It seems more in line with what we've done in other themes like Archeo and Remote.

jffng:
> Should we call it "Normal" instead. and rename other larger fonts.

Hmm I'm not sure I understand why we should rename it to Normal? There's been a lot of conversation about font size presets and I think we should stick to the Gutenberg supplied defaults — [small, medium, large, extra  large](https://github.com/WordPress/gutenberg/blob/trunk/lib/compat/wordpress-6.0/theme.json#L197-L218) — to maximize cross-theme compatibility, but maybe I'm misunderstanding something about your suggestion.

Some more conversation here: https://github.com/Automattic/themes/pull/5946#pullrequestreview-955069064 and p1651073265276749-slack-C029FM1EH

mikachan:
Ahh I thought Normal was part of the GB defaults, but only based on how much I saw it crop up on previous themes! Yeah I think sticking with the GB defaults makes sense, and should help when users switch themes.

Thanks for the extra context too, I'd missed both conversations!

jffng:
Oops I meant to leave this open so @beafialho could review. Anyways, we can address any feedback in a follow up :) Thanks all.

PR #5985, opened 2022-05-09
Utility to update patterns with escaped strings and image paths

madhusudhand:

This wraps text and image paths inside php esc functions.

Verified with text and image paths.
Need help in validating, improving this.

As the input for this utility is git updated files, this can become a pre-commit hook in future. 
(need to add additional step to stage the files after updates).

#### Related issue(s):

None.

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -1188,7 +1188,7 @@ async function escapePatterns() {
 		const rewriter = new RewritingStream();
 
 		rewriter.on('text', (_, raw) => {
-			rewriter.emitRaw(translateText(raw, themeSlug));
+			rewriter.emitRaw(escapeText(raw, themeSlug));
 		});
 
 		rewriter.on('startTag', (startTag) => {
@@ -1196,9 +1196,9 @@ async function escapePatterns() {
 				const attrs = startTag.attrs.filter(attr => ['src', 'alt'].includes(attr.name));
 				attrs.forEach(attr => {
 					if (attr.name === 'src') {
-						attr.value = updateImagePath(attr.value);
+						attr.value = escapeImagePath(attr.value);
 					} else if (attr.name === 'alt') {
-						attr.value = translateText(attr.value, themeSlug);
+						attr.value = escapeText(attr.value, themeSlug, true);
 					}
 				});
 			}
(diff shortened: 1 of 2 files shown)


MaggieCabrera:
Thanks for doing this! This will save so much time! 

There are two other scenarios that I've encountered that are not being considered by this script:

- [ ] When the string has HTML tags (such as a `<br>` or a `<strong>` tag). In those cases we do something like: `wp_kses_post( __( 'AUGUSTA SMITH<br>AMBER ANDREWS<br>MARY HAN', 'videomaker' ) )`
- [x] The image alt attribute needs to use `esc_attr__( 'Photograph of a man holding a camera', 'videomaker' )`

theme-utils.mjs
```diff
@@ -107,6 +109,10 @@ const commands = {
 		additionalArgs: '<theme-slug> <since>',
 		run: (args) => rebuildThemeChangelog(args?.[1], args?.[2])
 	},
+	"escape-patterns": {
```
jffng, line 112: Rebased this to test / use it on Curator, worked like a charm 🌟

jffng (approved):

theme-utils.mjs
```diff
@@ -1130,3 +1136,106 @@ export async function executeCommand(command, logResponse) {
...
+		return `<?php echo ${escFunction}( '${text.replace('\'', '\\\'')}', '${themeSlug}' ); ?>`;
+	}
+
+	function escapeImagePath(src) {
+		if (!src || src.trim().startsWith('<?php')) return src;
+		
+		const assetsDir = 'assets';
+		const parts = src.split('/');
+		const resultSrc = parts.slice(parts.indexOf(assetsDir)).join('/');
+		return `<?php echo esc_url( get_template_directory_uri() ); ?>/${resultSrc}`;
```
jffng, line 1231: It may be more useful to prepend the usual path where we place static image files: 
```suggestion
		return `<?php echo esc_url( get_template_directory_uri() ); ?>/assets/images/${resultSrc}`;
```

madhusudhand: `resultSrc` is calculated to include `assets` and the image path in the previous line.

madhusudhand:
Merging the current version. We can continue to improve when there are more scenarios to cover.

Issue #4730, opened 2021-09-28
Canard - Blog Posts Block background and height


silvio-meter:
## Steps to replicate

1. Activate the Canard theme
2. Create a new page and add a Blog Posts Block (grid view).

## Result

The post excerpt shows the black background and there's a height issue as well.

Screenshot: 

![Optional alternative text if image doesn't load](https://groovytestingsite.files.wordpress.com/2021/09/annotation-on-2021-09-28-at-19-28-55.png)
--

##Expected

There shouldn't be any issues with post thumbnails and excerpts while adding the Blog Posts Block.

[What you expected to happen.]

* User report: [#4332912-zen]
* Follow-up ticket: [#4332912-zen]

vinnykaur:
The user followed up about the progress again. For now, I have shared the below CSS with them after adding a Custom Design subscription as a one-time exception:

```
figure.post-thumbnail {
    height: auto;
}
```

AtrumGeost:
Got another report here:

33586519-hc

benchilcote:
Another case here: 34336354-hc - Gave CSS as temporary workaround.

thesacredpath:
Another report here: 34449061-zen Gave CSS as temp workaround.

Follow up: 4824166-zen

Greatdane:
4877496-zd-woothemes

jp-imagines:
Another report in 35151362-hc. Gave user the [CSS workaround](https://github.com/Automattic/themes/issues/4730#issuecomment-967772245).

mikachan:
Fixed by https://github.com/Automattic/themes/pull/5986

arunsathiya:
I have followed up with all customers.

PR #5986, opened 2022-05-09
#4730: Canard: Fix Blog Post block thumbnail background

bobmatyas:

Adjust background color and height in blog post block.

##### Before

Screenshot on 2022-05-09 at 13-03-02

##### After

Screenshot on 2022-05-09 at 13-13-02

#### Related issue(s):

https://github.com/Automattic/themes/issues/4730

### DIFF
--- a/canard/style.css
+++ b/canard/style.css
@@ -2136,9 +2136,7 @@ body:not(.group-blog) .sticky .entry-summary + .entry-meta > span:nth-of-type(3)
 
 /* Post Thumbnail */
 .post-thumbnail {
-	background: #000;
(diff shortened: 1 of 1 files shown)


mikachan (approved):
This looks good to me. The only potential issue I can see is if the featured image is a different aspect ratio to the images next to it, it may cause a difference in column heights, which it wouldn't have done before:

With this PR:

image

Without this PR:

image

But it looks like this doesn't change the alignment of multiple rows of posts, and I think this change is still a massive improvement, so approving based on this ✨

PR #5987, opened 2022-05-10
Curator: revise single template

### LINKED ISSUE #5931: Curator: Templates: Single
**With featured image**

Single Post

**Without a featured image**

image

PeterNdomano:
…issues #5931

Edited curator/templates/single.html to meet the design suggested in #5931  for single post view. Made the template file completely compatible with FSE.

#### Related issue(s):
#5931


curator/templates/single.html
```diff
@@ -1,28 +1,121 @@
...
+<div class="wp-block-column"><!-- wp:post-content {"layout":{"inherit":true}} /--></div>
+<!-- /wp:column --></div>
+<!-- /wp:columns --></div>
+<!-- /wp:group -->
+
+<!-- wp:spacer {"height":"60px"} -->
+<div style="height:60px" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- /wp:spacer -->
+
+<!-- wp:group {"style":{"spacing":{"padding":{"top":"30px","bottom":"30px"}},"border":{"top":{"color":"var:preset|color|foreground","width":"2px"},"bottom":{"color":"var:preset|color|foreground","width":"2px"}}},"layout":{"inherit":false}} -->
```
jffng: The width of the separator here should be 1px, and it should also extend full width (touching the edges of the screen).

curator/templates/single.html
```diff
@@ -1,28 +1,121 @@
...
+<div class="wp-block-group" style="border-top-color:var(--wp--preset--color--foreground);border-top-width:2px;border-bottom-color:var(--wp--preset--color--foreground);border-bottom-width:2px;padding-top:30px;padding-bottom:30px"><!-- wp:group {"layout":{"inherit":true}} -->
+<div class="wp-block-group"><!-- wp:group {"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
+<div class="wp-block-group"><!-- wp:post-navigation-link {"type":"previous","label":"\u003cstrong\u003ePREVIOUS POST\u003c/strong\u003e","fontSize":"large"} /-->
```
jffng: We'll need to escape all strings, including these labels as well as line 66 ("Comments"). Also uppercase should be done via text transform / CSS.

To do this, maybe we should create a "hidden" pattern for the comments, like the 404 pattern: https://github.com/Automattic/themes/blob/trunk/curator/patterns/404.php#L5

curator/templates/single.html
```diff
@@ -1,28 +1,121 @@
-<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
+<!-- wp:template-part {"slug":"header","theme":"curator","tagName":"header"} /-->
```
jffng: Let's get rid of the theme slug here, so the templates and parts always resolve from the currently active theme:

```suggestion
<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
```

Same for the footer part.

curator/templates/single.html
```diff
@@ -1,28 +1,121 @@
...
+<div class="wp-block-group"><!-- wp:paragraph {"fontSize":"x-large"} -->
+<p class="has-x-large-font-size">Comments</p>
+<!-- /wp:paragraph -->
+
+<!-- wp:paragraph {"fontSize":"x-large"} -->
+<p class="has-x-large-font-size">(</p>
+<!-- /wp:paragraph -->
 
-<!-- wp:post-content {"layout":{"inherit":true}} /-->
+<!-- wp:post-comments-count {"textAlign":"center","fontSize":"x-large"} /-->
```
jffng: There's quite a bit of space between the parenthesis and comments count here — any way we can reduce that? 

Screen Shot 2022-05-10 at 11 27 47 AM

curator/templates/single.html
```diff
@@ -1,28 +1,121 @@
...
+<!-- wp:spacer {"height":"60px"} -->
+<div style="height:60px" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- /wp:spacer -->
+
+<!-- wp:group {"style":{"spacing":{"padding":{"top":"30px","bottom":"30px"}},"border":{"top":{"color":"var:preset|color|foreground","width":"2px"},"bottom":{"color":"var:preset|color|foreground","width":"2px"}}},"layout":{"inherit":false}} -->
+<div class="wp-block-group" style="border-top-color:var(--wp--preset--color--foreground);border-top-width:2px;border-bottom-color:var(--wp--preset--color--foreground);border-bottom-width:2px;padding-top:30px;padding-bottom:30px"><!-- wp:group {"layout":{"inherit":true}} -->
+<div class="wp-block-group"><!-- wp:group {"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
+<div class="wp-block-group"><!-- wp:post-navigation-link {"type":"previous","label":"\u003cstrong\u003ePREVIOUS POST\u003c/strong\u003e","fontSize":"large"} /-->
+
+<!-- wp:post-navigation-link {"label":"\u003cstrong\u003eNEXT POST\u003c/strong\u003e","fontSize":"large"} /--></div>
```
jffng: The font family should be Work Sans for the pagination: 

Screen Shot 2022-05-10 at 11 30 30 AM

jffng (changes requested):
Thank you for this @PeterNdomano ! Very cool, also nice to see the new post comments blocks in action. 

I left some initial comments.

PeterNdomano:
Thanks @jffng for your review. All the requested changes were covered in the commits  but I would like to point out these 2 things:


### 1) I removed prev and next arrows in post navigation for mobile view to prevent the phrase "PREVIOUS POST" from wrapping over 2 lines. It's just my suggestion for mobile view, if it ain't right we can bring them back!.. Consider the images below.

on mobile
mobile-post-nav

on desktop
desktop-post-nav




### 2) I tried to detect if the post has a featured image inside the pattern curator/post-hero so that I can load specific cover block in order to match the targeted design, Consider images below.

without featured image
without-featured-image

with featured image
with-featured-image



So, to accomplish that I tried using the method below but it doesn't seem to work, would you please guide me on that.

```
<?php if ( has_post_thumbnail() ): ?>
<!-- wp:cover {"useFeaturedImage":true,"dimRatio":50,"overlayColor":"foreground","minHeight":100,"minHeightUnit":"vh"} -->
<!-- /wp:cover -->
<?php else: ?>
<!-- wp:cover {"overlayColor":"background","minHeight":50,"minHeightUnit":"vh","isDark":false} -->
<!-- /wp:cover -->
<?php endif; ?>
```

Thanks once more @jffng .

curator/templates/single.html
```diff
@@ -1,28 +1,45 @@
...
+<!-- wp:spacer {"height":"60px"} -->
+<div style="height:60px" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- /wp:spacer -->
 
-<!-- wp:template-part {"slug":"post-meta"} /-->
-</main>
+<!-- wp:group {"layout":{"inherit":true}} -->
+<div class="wp-block-group"><!-- wp:columns {"verticalAlignment":"center","align":"wide"} -->
+<div class="wp-block-columns alignwide are-vertically-aligned-center"><!-- wp:column {"verticalAlignment":"center","width":"100%","layout":{"inherit":false}} -->
+<div class="wp-block-column is-vertically-aligned-center" style="flex-basis:100%"><!-- wp:post-excerpt {"textAlign":"center","fontSize":"x-large"} /--></div>
```
madhusudhand: There is no need of an except block inside single post page. 
First paragraph in the designs is just another patterns which is wide and center aligned, which is just an indication that user can add such patterns/blocks.

curator/patterns/post-navigation.php
```diff
@@ -0,0 +1,34 @@
...
+<p class="curator-post-navigation-icon has-large-font-size">←</p>
+<!-- /wp:paragraph -->
+
+<!-- wp:post-navigation-link {"type":"previous","label":"<?php echo esc_html__( 'PREVIOUS POST', 'curator' ); ?>","style":{"typography":{"lineHeight":"1.5","textTransform":"uppercase"}},"fontSize":"large"} /--></div>
+<!-- /wp:group --></div>
+<!-- /wp:column -->
+
+<!-- wp:column -->
+<div class="wp-block-column"><!-- wp:group {"style":{"spacing":{"blockGap":"2px"}},"className":"curator-post-navigation","layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"right"}} -->
+<div class="wp-block-group curator-post-navigation"><!-- wp:post-navigation-link {"label":"<?php echo esc_html__( 'NEXT POST', 'curator' ); ?>","style":{"typography":{"textTransform":"uppercase"}},"fontSize":"large"} /-->
```
madhusudhand: icons ← and → font not matching. also the gap between icon and text.

image
image

Just icon appears when there isn't next/prev post.
image
May be it is safe to keep the arrow part of label instead of a separate block.

curator/patterns/post-hero.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+?>
+
+<?php if ( has_post_thumbnail() ): ?>
+<!-- wp:cover {"useFeaturedImage":true,"dimRatio":50,"overlayColor":"foreground","minHeight":100,"minHeightUnit":"vh"} -->
+<div class="wp-block-cover" style="min-height:100vh"><span aria-hidden="true" class="wp-block-cover__background has-foreground-background-color has-background-dim"></span><div class="wp-block-cover__inner-container"><!-- wp:group {"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"center"}} -->
+<div class="wp-block-group"><!-- wp:post-terms {"term":"post_tag","style":{"typography":{"fontStyle":"normal","fontWeight":"700"}},"textColor":"background","className":"curator-post-tags","fontSize":"medium"} /-->
+
+<!-- wp:post-date {"textColor":"background","fontSize":"medium"} /--></div>
```
madhusudhand, line 15: There is a pattern created for the terms and date [here](https://github.com/Automattic/themes/pull/5994/files#diff-73a0eafe38a2cdc775a3a9f5898b911ea762d3d1e36f1f2d3c32854a55e05109). Please use it once the PR gets merged.

madhusudhand (commented):
Hi @PeterNdomano, Thanks for splitting the page into multiple patterns. 
I posted few comments inline.

curator/patterns/post-comments.php
```diff
@@ -0,0 +1,70 @@
...
+
+?>
+
+<!-- wp:group {"layout":{"inherit":true}} -->
+<div class="wp-block-group"><!-- wp:post-comments-form /--></div>
+<!-- /wp:group -->
+
+<!-- wp:spacer {"height":"60px"} -->
+<div style="height:60px" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- /wp:spacer -->
```
madhusudhand, line 16: Adjust the spacing

image image

curator/patterns/post-comments.php
```diff
@@ -0,0 +1,70 @@
...
+<!-- wp:paragraph {"textColor":"foreground","fontSize":"x-large"} -->
+<p class="has-foreground-color has-text-color has-x-large-font-size">(</p>
+<!-- /wp:paragraph -->
+
+<!-- wp:post-comments-count {"textAlign":"center","textColor":"foreground","fontSize":"x-large"} /-->
+
+<!-- wp:paragraph {"textColor":"foreground","fontSize":"x-large"} -->
+<p class="has-foreground-color has-text-color has-x-large-font-size">)</p>
+<!-- /wp:paragraph --></div>
+<!-- /wp:group --></div>
```
madhusudhand: Adjust block gap to `0` and add right padding to comments block.


| Current  | Expected |
| image  | image  |

curator/patterns/post-hero.php
```diff
@@ -0,0 +1,32 @@
+<?php
+/**
+ * Title: Single Post Hero
+ * Slug: curator/post-hero
+ * Inserter: no
+ */
+
+?>
+
+<?php if ( has_post_thumbnail() ): ?>
```
jffng: I would avoid this logic inside the patterns, I think this should be handled by the cover block itself. In other words, the block should be able to detect if there is a featured image, and render itself accordingly. 

My recommendation for now is to remove it and use the "useFeaturedImage" version of the cover block you have here.

curator/patterns/post-navigation.php
jffng:
```suggestion
<div class="wp-block-group curator-post-navigation"><!-- wp:post-navigation-link {"label":"<?php echo esc_html__( 'Next Post', 'curator' ); ?>","style":{"typography":{"textTransform":"uppercase"}},"fontSize":"large"} /-->
```

curator/patterns/post-navigation.php
```diff
@@ -0,0 +1,34 @@
...
+<div class="wp-block-group alignfull" style="border-top-color:var(--wp--preset--color--foreground);border-top-width:1px;border-bottom-color:var(--wp--preset--color--foreground);border-bottom-width:1px;padding-top:30px;padding-bottom:30px"><!-- wp:group {"layout":{"inherit":true}} -->
+<div class="wp-block-group"><!-- wp:columns {"isStackedOnMobile":false} -->
+<div class="wp-block-columns is-not-stacked-on-mobile"><!-- wp:column {"width":"50%","layout":{"inherit":true}} -->
+<div class="wp-block-column" style="flex-basis:50%"><!-- wp:group {"style":{"spacing":{"blockGap":"2px"}},"className":"curator-post-navigation","layout":{"type":"flex","flexWrap":"nowrap"}} -->
+<div class="wp-block-group curator-post-navigation"><!-- wp:paragraph {"className":"curator-post-navigation-icon","fontSize":"large"} -->
+<p class="curator-post-navigation-icon has-large-font-size">←</p>
+<!-- /wp:paragraph -->
+
+<!-- wp:post-navigation-link {"type":"previous","label":"<?php echo esc_html__( 'PREVIOUS POST', 'curator' ); ?>","style":{"typography":{"lineHeight":"1.5","textTransform":"uppercase"}},"fontSize":"large"} /--></div>
```
jffng:
```suggestion
<!-- wp:post-navigation-link {"type":"previous","label":"<?php echo esc_html__( 'Previous Post', 'curator' ); ?>","style":{"typography":{"lineHeight":"1.5","textTransform":"uppercase"}},"fontSize":"large"} /--></div>
```

curator/patterns/post-navigation.php
jffng: > May be it is safe to keep the arrow part of label instead of a separate block

I agree with this.

curator/patterns/post-hero.php
jffng: Noting that the tradeoff with this recommendation, I think for now we have to be okay with the following state if there is no featured image set: 

Screen Shot 2022-05-23 at 11 22 53 AM

PeterNdomano: Got u @jffng , I'll adopt your recommendation.

PeterNdomano:
Thank you guys @madhusudhand @jffng 
I'll work on the changes ASAP

jffng:
@PeterNdomano thanks for your work here, I made a number of changes to simplify this one a bit and work within the parameters of what's possible right now. 

@beafialho I think this is in a decent spot to review. Here are a few scenarios to test the Single Post template:

- With and without a featured image
- Post containing wide and full width content
- Adding some comments

I do think be useful to review and hopefully merge #5980 first, so that we can get a better sense of this template in context. 

One question — is the meta supposed to display tags, categories, both, or just whether the post is featured or not?

Screen Shot 2022-05-26 at 12 57 40 PM

There are a couple of issues that I think we should address in follow ups before this PR gets too big:

- The previous / next link's color and font family depend on this Gutenberg PR: https://github.com/WordPress/gutenberg/pull/41378
- Hover/active states to the Previous / Next links, as well as the post terms in the featured image area, perhaps this should be addressed with https://github.com/Automattic/themes/pull/6028
- I've removed the arrows from these links for now since it's not clear to me how these should be implemented, and would like to address this in a follow up PR and / or Gutenberg issue / PR.

PeterNdomano:
Thanks @jffng I noticed that this PR still depends on some other unmerged PRs like [this one on post meta and date](https://github.com/Automattic/themes/pull/5994/files#diff-73a0eafe38a2cdc775a3a9f5898b911ea762d3d1e36f1f2d3c32854a55e05109) as mentioned by @madhusudhand in the previous review. So I thought I had to wait for those PRs to get merged first, that's why I didn't commit yet after your last reviews. However, I hope review & clarification from @beafialho will help us to keep going with this PR.

beafialho:
Thank you all for your work. @jffng I'll review the Single Post template and give feedback.

> One question — is the meta supposed to display tags, categories, both, or just whether the post is featured or not?

In the post template, just categories and post date.

beafialho:
I tried some of those options you mentioned:

- This is what a post with a long title looks like, quite tight. I tried adding 120px padding above and below (not on the sides) and it looks better.

| Currently | With top and bottom padding |
| ----- | ----- |
| Captura de ecrã 2022-05-27, às 10 36 51 |  Captura de ecrã 2022-05-27, às 10 38 28 |

- The post category is underlined by default. It shouldn't be, it should be underlined on hover.

Captura de ecrã 2022-05-27, às 11 06 31

- Comments look good in general! My only comment is: can the Reply and Edit buttons use a smaller size? They shouldn't be as highlighted as the text.

Captura de ecrã 2022-05-27, às 10 45 21

madhusudhand:
@PeterNdomano The design of the category, date pattern changed to different style in archive pages. You no longer need to depend on the other PR. Please style them with above design.

PeterNdomano:
Thanks @beafialho for your review. 

@jffng  I added 120px padding on top and bottom inside the inner cover and now it looks fine. I tried to modify fontSize using `{"fontSize":"medium"}` inside `comment-content` block and `{"fontSize":"small"}` inside both reply and edit links in order to differentiate the sizes as suggested, but unfortunately they don't seem to work. Perhaps it's an issue with these comment blocks and maybe we should use raw CSS. 

Regarding the underlining of post-category before hover I hope @jffng will have a better solution coz last time I tried to do it with raw CSS inside `styles.css` file

PeterNdomano:
> @PeterNdomano The design of the category, date pattern changed to different style in archive pages. You no longer need to depend on the other PR. Please style them with above design.

Thanks @madhusudhand

jffng:
Thanks everyone! 

I tried a long title and the top and bottom padding look better, but maybe we should add some to the sides as well since you can end up in situations like the following depending on the word length? 

Screen Shot 2022-05-27 at 10 43 21 AM

> The post category is underlined by default. It shouldn't be, it should be underlined on hover.

Yeah, I was hoping we can address that as a part of https://github.com/Automattic/themes/pull/6028, since that one is handling link styles and we will want to be concise with those rules.

> I tried to modify fontSize using {"fontSize":"medium"} inside comment-content block and {"fontSize":"small"}

@PeterNdomano these changes appear seem to be working for me!

Screen Shot 2022-05-27 at 10 50 02 AM

PeterNdomano:
>but maybe we should add some to the sides as well since you can end up in situations like the following depending on the word length?

Yeah, that's right. I just added 20px left and right padding. I chose 20px after some number of tests to make sure it displays well on both mobile and desktop. What do you think about it @jffng 

>these changes appear seem to be working for me!

Ohhh...! I think it was an issue with caching on my end.

jffng:
>  I just added 20px left and right padding. I chose 20px after some number of tests to make sure it displays well on both mobile and desktop. What do you think about it

Let's use the same outer spacing value as the rest of the site's padding here. I updated that in https://github.com/Automattic/themes/pull/5987/commits/d6529ca9d41155e26cbd57635c9d080113a37b0e

I've addressed the remaining feedback from @beafialho and rebased this one, I think it's in a good spot to bring in. There is one dependency on a GB PR ( https://github.com/WordPress/gutenberg/pull/41378 ) but I don't think that should hold this up. Thanks @PeterNdomano @madhusudhand @beafialho!

jffng (approved):

Issue #2220, opened 2020-07-03
Radcliffe 2: Blog Posts Block List View does not show Featured Images


otpidusprime:
## Steps to replicate

1. Activate Radcliffe 2 theme
2. Edit post/page using the Block Editor (you should have posts with Featured Images)
3. Add Blog Posts Block
4. Click "List View" and "Show Media on Left or Right".
    ![Screen Shot on 2020-07-03 at 21:00:27](https://user-images.githubusercontent.com/29238672/86482837-58ade680-bd70-11ea-8e94-b445079d330b.png)
5. Click "Preview" on the top right corner

## Result

It does not show Featured Image on the site, _tho on the Editor, it previews fine_.

#### On site:
![Screen Shot on 2020-07-03 at 21:03:45](https://user-images.githubusercontent.com/29238672/86483015-c5c17c00-bd70-11ea-97cc-4f20c6afbafd.png)

#### On Editor:
![Screen Shot on 2020-07-03 at 21:04:05](https://user-images.githubusercontent.com/29238672/86483043-d245d480-bd70-11ea-9d00-f697e15860b7.png)

## Expected

Render Featured image in List View on site:

![Screen Shot on 2020-07-03 at 21:05:55](https://user-images.githubusercontent.com/29238672/86483138-01f4dc80-bd71-11ea-9be2-be8a06d3abd5.png)

**Note:** The issue is only on the List View(Featured Image on Left or Right) in Radcliffe 2.

* User report: [hc-19920357](https://mc.a8c.com:443/support-stats/happychat/transcript.php?id=19920357)
* Follow-up ticket: [zd-3111537](https://woothemes.zendesk.com/agent/tickets/3110545)

otpidusprime:
As a temp. workaround for sites under Premium plan or above, we can use this CSS to fix this:

```
.wpnbha.image-alignleft .post-thumbnail,
.wpnbha.image-alignright .post-thumbnail {
    flex: 1 0 50%;
}

.wpnbha.image-alignleft .entry-wrapper,
.wpnbha.image-alignright .entry-wrapper {
    min-width: calc(100% - 50%);
}
```

For what its worth, I noticed that if we set the `flex-shrink` value `1`, the feature image disappears.

brookewp:
For a Personal plan site, I had them change the width to wide or full width which also helped the problem (zd-3336925)

PR #5993, opened 2022-05-10
#2220: Radcliffe 2: Fix Blog Posts block display list w/images

bobmatyas:
##### Before
Screenshot on 2022-05-10 at 14-16-25

##### After

Screenshot on 2022-05-10 at 14-18-03

#### Related issue(s):

#2220

### DIFF
--- a/radcliffe-2/style.css
+++ b/radcliffe-2/style.css
@@ -1507,6 +1507,11 @@ textarea {
 	color: #fff;
 }
 
+.wpnbha.image-alignleft .entry-wrapper,
+.wpnbha.image-alignright .entry-wrapper {
+    min-width: 50%;
+}
+
 .entry-header {
 	padding: 36px 0;
 	text-align: center;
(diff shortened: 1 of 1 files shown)


mikachan (commented):
Thanks for working on this!

This works well for me at resolutions above 600px in width, but at resolutions smaller than this, it looks like we need a similar fix:

image

It looks like the following selectors need a min-width property as well:

image

image

The `min-width: 50%` fix should probably sit in a media query for min-width 600px, e.g. `@media only screen and (min-width: 600px)`, and then the fix for the smaller resolutions can sit outside of this.

bobmatyas:
Thanks for the feedback!

I made some adjustments that should fix this on mobile and moved the `min-width: 50%` inside a media query. I also added the additional selectors.

#### Screenshot
Screenshot on 2022-05-11 at 12-21-15

mikachan (approved):
Brill, this LGTM! 🚢

PR #5994, opened 2022-05-11
add archive template and posts-grid pattern

### LINKED ISSUE #5932: Curator: Templates: Archive
Screen Shot 2022-05-02 at 9 41 50 AM

madhusudhand:

image

#### Related issue(s):

Fixes: #5932 

## Issues to resolve

This PR has the following problems that needs to be solved. Need suggestions.

1. When there are not tags, UI starts with the separator dot (.). Is there a way to conditionally add it?

image

2. Is there a block to add the title `Search: search term`? Archive title isn't working in search template

image

### DIFF
--- a/curator/patterns/hidden-post-terms-and-date.php
+++ b/curator/patterns/hidden-post-terms-and-date.php
@@ -10,6 +10,6 @@
 <!-- wp:group {"style":{"spacing":{"blockGap":"10px"}},"layout":{"type":"flex","flexWrap":"nowrap"}} -->
 <div class="wp-block-group">
 <!-- wp:post-date {"isLink":true} /-->
-<!-- wp:post-terms {"term":"post_tag"} /-->
+<!-- wp:post-terms {"term":"category"} /-->
 </div>
 <!-- /wp:group -->
--- a/curator/patterns/hidden-post-terms-and-date.php
+++ b/curator/patterns/hidden-post-terms-and-date.php
@@ -7,7 +7,7 @@
 
 ?>
 
-<!-- wp:group {"style":{"spacing":{"blockGap":"10px"}},"layout":{"type":"flex","orientation":"vertical"}} -->
+<!-- wp:group {"style":{"spacing":{"blockGap":"0px"}},"layout":{"type":"flex","orientation":"vertical"}} -->
 <div class="wp-block-group"><!-- wp:post-terms {"term":"category"} /-->
 <!-- wp:post-date {"isLink":true,"style":{"typography":{"fontStyle":"italic","fontWeight":"400","fontSize":"14px"}}} /--></div>
 <!-- /wp:group -->
--- a/curator/patterns/hidden-post-terms-and-date.php
+++ b/curator/patterns/hidden-post-terms-and-date.php
@@ -7,12 +7,9 @@
 
 ?>
 
-<!-- wp:group {"style":{"spacing":{"blockGap":"5px"}},"layout":{"type":"flex","flexWrap":"nowrap"}} -->
+<!-- wp:group {"style":{"spacing":{"blockGap":"10px"}},"layout":{"type":"flex","flexWrap":"nowrap"}} -->
 <div class="wp-block-group">
-<!-- wp:post-terms {"term":"post_tag"} /-->
-<!-- wp:paragraph -->
-<p>·</p>
-<!-- /wp:paragraph -->
 <!-- wp:post-date {"isLink":true} /-->
+<!-- wp:post-terms {"term":"post_tag"} /-->
 </div>
 <!-- /wp:group -->
--- a/curator/patterns/hidden-post-terms-and-date.php
+++ b/curator/patterns/hidden-post-terms-and-date.php
@@ -7,9 +7,7 @@
 
 ?>
 
-<!-- wp:group {"style":{"spacing":{"blockGap":"10px"}},"layout":{"type":"flex","flexWrap":"nowrap"}} -->
-<div class="wp-block-group">
-<!-- wp:post-date {"isLink":true} /-->
-<!-- wp:post-terms {"term":"category"} /-->
-</div>
+<!-- wp:group {"style":{"spacing":{"blockGap":"10px"}},"layout":{"type":"flex","orientation":"vertical"}} -->
+<div class="wp-block-group"><!-- wp:post-terms {"term":"category"} /-->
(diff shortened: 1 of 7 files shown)


curator/patterns/posts-grid.php
```diff
@@ -0,0 +1,29 @@
...
+ * Slug: curator/posts-grid
+ * Categories: featured, query
+ */
+?>
+
+<!-- wp:group {"align":"full","layout":{"type":"flex","flexWrap":"nowrap"}} -->
+<div class="wp-block-group alignfull"><!-- wp:query {"queryId":0,"query":{"perPage":6,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true},"displayLayout":{"type":"flex","columns":3},"align":"wide"} -->
+<div class="wp-block-query alignwide"><!-- wp:post-template -->
+<!-- wp:group {"style":{"spacing":{"padding":{"top":"20px","right":"20px","bottom":"20px","left":"20px"}}},"layout":{"inherit":false}} -->
+<div class="wp-block-group" style="padding-top:20px;padding-right:20px;padding-bottom:20px;padding-left:20px"><!-- wp:post-featured-image {"isLink":true} /-->
```
jffng: The spacing here seems a little off from the designs, maybe a little more space below the featured image, or reducing the space between the post terms and post title: 

Screen Shot 2022-05-11 at 10 53 38 AM

jffng (commented):
> When there are not tags, UI starts with the separator dot (.). Is there a way to conditionally add it?

Not with block templates, it would have to be in the block options somehow. I wonder if we should just remove the separator dot for now, unless we are okay with how it looks when there are no tags. 

> Is there a block to add the title Search: search term? Archive title isn't working in search template

I think it would be useful if the archive title block supported rendering the search term, or something along those lines. 

In the meantime though, maybe this template could match our other block themes' search templates where the input is rendered at the top of the results page, for example: 

Screen Shot 2022-05-11 at 11 01 48 AM

What do you think @beafialho ?

beafialho:
> Not with block templates, it would have to be in the block options somehow. I wonder if we should just remove the separator dot for now, unless we are okay with how it looks when there are no tags.

Yeah 😞 unfortunately removing the dot might be the only option. Will it look like this with the category then? This is also not ideal.

Captura de ecrã 2022-05-12, às 09 14 43

> I think it would be useful if the archive title block supported rendering the search term, or something along those lines.

Absolutely. I keep designing `Search: "term"` for every theme but it isn't possible... Would be great if it could be achieved. In the meantime, showing the search field like you suggested works @jffng.

madhusudhand:
Adjusted spacing, and used search block in the search page.

image

image

jffng:
> Will it look like this with the category then? This is also not ideal.

Yes. Is that acceptable? Also just noting what happens when there is more than one category: 

Screen Shot 2022-05-23 at 10 58 06 AM

curator/patterns/hidden-post-terms-and-date.php
```diff
@@ -0,0 +1,15 @@
...
+ * Slug: curator/post-terms-and-date
+ * Inserter: no
+ */
+
+?>
+
+<!-- wp:group {"style":{"spacing":{"blockGap":"10px"}},"layout":{"type":"flex","flexWrap":"nowrap"}} -->
+<div class="wp-block-group">
+<!-- wp:post-date {"isLink":true} /-->
+<!-- wp:post-terms {"term":"post_tag"} /-->
```
jffng: Neither the category nor tag is rendering for me right now, I think it should be the following: 

```suggestion
<!-- wp:post-terms {"term":"category"} /-->
```

beafialho:
> Yes. Is that acceptable? Also just noting what happens when there is more than one category:

I'm not sure this is a good alternative. If there are more categories it'll look quite full and the dot really helped separating the information.

Maybe whenever posts are showing up as a grid it works better if we use the Tiny paragraph size (14px) in post date with 10px spacing in the Stack block. Can you try it?

Captura de ecrã 2022-05-24, às 12 16 16

jffng:
> Maybe whenever posts are showing up as a grid it works better if we use the Tiny paragraph size (14px) in post date with 10px spacing in the Stack block. Can you try it?

Sure thing, I tried that here: https://github.com/Automattic/themes/pull/5994/commits/900d744cc93f43a9e59d486436bfe7f50ca93578

Screen Shot 2022-05-25 at 10 58 50 AM

What do you think @beafialho ? Using 10px as the block gap makes it appear bigger than the spacing suggested in the comps, but I left it there for your consideration.

beafialho:
Thanks @jffng, can you try 5px instead of 10px please?

jffng:
Sure thing:

0px | 5px | 10px
--- | --- | ---
Screen Shot 2022-05-25 at 11 35 49 AM | Screen Shot 2022-05-25 at 11 36 04 AM | Screen Shot 2022-05-25 at 11 36 13 AM

beafialho:
Thank you, that was just what I needed! 0px seems to look the best to me.

curator/theme.json
```diff
@@ -231,12 +231,43 @@
...
 			},
+			"core/post-terms": {
+				"elements": {
+					"link": {
+						"color": {
+							"text": "var(--wp--preset--color--foreground)"
+						},
+						"typography": {
+							"textDecoration": "none",
+							"textTransform": "capitalize",
```
jffng: I think this can be removed...

beafialho:
I love how this is looking! I have spotted a few things to be polished:

- Navigation should look like this (both in Archive as in Search templates):

Captura de ecrã 2022-05-27, às 12 24 04

Keep in mind that it shouldn't be underlined by default.

- Is there a way to avoid this? Right before it stacks for the mobile layout, the content, more specifically the post titles look quite tight.

Captura de ecrã 2022-05-27, às 12 25 43

madhusudhand:
@beafialho 

> Navigation should look like this (both in Archive as in Search templates):

Navigation styling is in progress as part of the other PR https://github.com/Automattic/themes/pull/5987
Once it gets merged we will be consuming the same component here.

> Is there a way to avoid this? Right before it stacks for the mobile layout, the content, more specifically the post titles look quite tight.

Gutenberg is supporting only mobile breakpoints and we might need custom styles to adjust to 2 column layout. However, because the grid is a pattern, it will break the page when theme switched.

Following is the UI for iPad Mini device size.

image

madhusudhand:
Created https://github.com/Automattic/themes/issues/6059 to add pagination once the single page PR is merged.

Merging this PR to reduce inter dependencies. Please add if these pages need more changes.

Issue #4934, opened 2021-10-26
Canard, Twenty Sixteen, Boardwalk: Padding issue on <mark> element when using inline colors 


mikeicode:
### Quick summary

Inline colors use the `<mark>` HTML element (a recent Gutenberg change). Some themes have CSS to add padding to this element but that results in spaces between text.

### Steps to reproduce

1. Create a line of text with no spaces
2. Add inline colors to the text
3. View published page

### What you expected to happen

The published page should have no spaces, like the editor view.

![Screen Shot on 2021-10-26 at 12:52:53](https://user-images.githubusercontent.com/10353785/138925977-9eaa79ec-e1b1-4a63-a26f-6181689f0154.png)

### What actually happened

Padding added by the theme gives the appearance of spaces.

![Screen Shot on 2021-10-26 at 12:56:14](https://user-images.githubusercontent.com/10353785/138925699-13f10d53-9cec-45a2-80bc-5ff099d64206.png)

### Context

32204955-hc

Used this to resolve:

```
/* Mark spacing fix */

mark.has-inline-color {
  padding: unset;
}
```

### Operating System

macOS

### Simple, Atomic or both?

Simple, Atomic

### Theme-specific issue?

Multiple themes have this. Of around 10 random themes I chose these had the issue:

Canard, Twenty Sixteen, Boardwalk

### Reproducibility

Consistent

### Available workarounds?

Yes, easy to implement

### Workaround details

If the site can use CSS:

```
/* Mark spacing fix */

mark.has-inline-color {
  padding: unset;
}
```

Robertght:
I've seen this in a few cases recently as well.  Can this be looked into in the near future?

cc @jeffikus / @scruffian

jeffikus:
@Robertght thanks for reporting. I'll triage this tomorrow and add to the board if we have capacity next week.

KokkieH:
Another report for Twenty-Sixteen: https://wordpress.com/forums/topic/changing-font-color-causes-text-to-shift-to-right/?view=all

mikachan:
Fixed by https://github.com/Automattic/themes/pull/5997

PR #5997, opened 2022-05-11
#4934: Canard: Fix highlight padding

bobmatyas:

Added CSS fix to removing padding when inline color is set using highlight feature in Gutenberg.

##### Before

Screenshot on 2022-05-11 at 12-34-09

##### After

Screenshot on 2022-05-11 at 14-33-15

#### Related issue(s):

https://github.com/Automattic/themes/issues/4934

### DIFF
--- a/canard/style.css
+++ b/canard/style.css
@@ -423,15 +423,12 @@ acronym {
 	border-bottom: 1px dotted #222;
 	cursor: help;
 }
-mark,
+mark:not(.has-inline-color),
 ins {
 	background: #d11415;
 	color: #fff;
 	padding: 0 0.2em;
 	text-decoration: none;
-}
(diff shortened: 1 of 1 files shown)


mikachan (commented):
This works well, but I'd probably tackle this slightly differently, and add a `:not` to the original selector adding the padding (on [line 426](https://github.com/Automattic/themes/blob/trunk/canard/style.css#L426)):

```
mark:not(.has-inline-color),
ins {
	background: #d11415;
	color: #fff;
	padding: 0 0.2em;
	text-decoration: none;
}
```

bobmatyas:
Good call. I made the adjustment. 👍

mikachan (approved):
Awesome!

PR #5998, opened 2022-05-12
Meraki: Update the footer pattern.

alaczek:
With this PR I'm moving some of the homepage design into the footer pattern. The intent is for this to show up in the footer on every page.

I wrapped all strings into the translation functions, but I wasn't sure how to handle the Jetpack's Subscribe block settings. 

Before:
image

After:
image


meraki/patterns/footer-default.php
```diff
@@ -6,6 +6,81 @@
...
+				?>
+			</p>
+			<!-- /wp:paragraph -->
+
+			<!-- wp:jetpack/subscriptions {"buttonOnNewLine":true,"buttonWidth":"100%","submitButtonText":"SUBSCRIBE","buttonBackgroundColor":"background","textColor":"black","fontSize":"14px","customFontSize":"14px","borderRadius":50,"borderColor":"background","spacing":10,"className":"is-style-split"} -->
+			<div class="wp-block-jetpack-subscriptions wp-block-jetpack-subscriptions__supports-newline wp-block-jetpack-subscriptions__use-newline is-style-split">[jetpack_subscription_form show_subscribers_total="false" button_on_newline="true" submit_button_text="SUBSCRIBE" custom_font_size="14px" custom_border_radius="50" custom_border_weight="1" custom_button_width="100%" custom_padding="15" custom_spacing="10" submit_button_classes="has-14-px-font-size has-background-border-color has-text-color has-black-color has-background has-background-background-color" email_field_classes="has-14-px-font-size has-background-border-color" show_only_email_and_button="true" success_message="Success! An email was just sent to confirm your subscription. Please find the email now and click 'Confirm Follow' to start subscribing."]</div>
```
MaggieCabrera: [Here](https://github.com/Automattic/themes/blob/trunk/remote/patterns/subscribe.php)'s how we handle that for Remote

MaggieCabrera (approved):
This LGTM pending the subscribe button translation strings

alaczek:
I experimented a bit and it turns out those block settings containing default strings can be removed without triggering block errors, so that's what I did. Let me know if that works.

MaggieCabrera:
> I experimented a bit and it turns out those block settings containing default strings can be removed without triggering block errors, so that's what I did. Let me know if that works.

I don't know much about the jetpack block, removing the strings doesn't seem to make any difference at first, but maybe they are used elsewhere or they are there in case the user wants to customize the success message. I'd rather leave them and translate them like the rest of the strings just in case but I don't have a strong opinion about it.

alaczek:
> I'd rather leave them and translate them like the rest of the strings just in case but I don't have a strong opinion about it.

Fair enough! I went back to the demo site, re-inserted the code, and copied it to the template so that they match. Turns out the only string hard coded is the success message, which I now included in the footer pattern.

PR #5999, opened 2022-05-12
Meraki: Style updates.

alaczek:
Minor styling changes to theme.json:
* update the secondary color definition.
* update default heading font-weight.

### DIFF
--- a/meraki/theme.json
+++ b/meraki/theme.json
@@ -4,18 +4,6 @@
 		"color": {
 			"palette": [
 				{
-					"slug": "primary",
-					"color": "#1a2530",
-					"name": "Primary",
-					"default": "#000000"
-				},
-				{
-					"slug": "secondary",
-					"color": "#f4e7f0",
-					"name": "Secondary",
-					"default": "#8d3f6e"
-				},
-				{
 					"slug": "foreground",
 					"color": "#1a2530",
 					"name": "Foreground",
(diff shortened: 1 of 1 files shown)


alaczek:
I wanted to update the button hover color as well - do I need to do it via style.css ([like in TT2](https://github.com/WordPress/twentytwentytwo/blob/trunk/style.css#L74))?

MaggieCabrera:
> I wanted to update the button hover color as well - do I need to do it via style.css ([like in TT2](https://github.com/WordPress/twentytwentytwo/blob/trunk/style.css#L74))?

Since this is a Blockbase child, you want to change the value of [this](https://github.com/Automattic/themes/blob/b4c9812b821819b76c4256a9b936ffb855e08cda/blockbase/theme.json#L113) variable.

alaczek:
Thanks @MaggieCabrera! What if I wanted to do the same thing as TT2 - don't change the color, but just add opacity - how would I do that in theme.json?

MaggieCabrera:
> Thanks @MaggieCabrera! What if I wanted to do the same thing as TT2 - don't change the color, but just add opacity - how would I do that in theme.json?

I don't think we contemplated that with Blockbase. You can either add an rgba value with opacity to the theme.json variable or simply add CSS. I think CSS is better in this case, so follow what TT2 does.

alaczek:
Yeah, ok. I don't think I'm going to do that in this case after all, so I'm assuming this is good to merge.

MaggieCabrera (approved):
LGTM

PR #6001, opened 2022-05-12
Heiwa: Add site-footer-container css class to remove the gap between main and footer

iamtakashi:
A minor edit that intends to remove the gap between the main and the footer. Thanks in advance for the review!

**Before:**
Screenshot 2022-05-12 at 15 13 20

**After:**
Screenshot 2022-05-12 at 15 14 18

### DIFF
--- a/heiwa/templates/header-footer-only.html
+++ b/heiwa/templates/header-footer-only.html
@@ -4,4 +4,4 @@
 <main class="wp-block-group" style="margin-top:0px;margin-bottom:0px"><!-- wp:post-content {"layout":{"inherit":true}} /--></main>
 <!-- /wp:group -->
 
-<!-- wp:template-part {"slug":"footer","tagName":"footer"} /-->
+<!-- wp:template-part {"slug":"footer","tagName":"footer","className":"site-footer-container"} /-->


heiwa/templates/header-footer-only.html
```diff
@@ -4,4 +4,4 @@
 <main class="wp-block-group" style="margin-top:0px;margin-bottom:0px"><!-- wp:post-content {"layout":{"inherit":true}} /--></main>
 <!-- /wp:group -->
 
-<!-- wp:template-part {"slug":"footer","tagName":"footer"} /-->
\ No newline at end of file
+<!-- wp:template-part {"slug":"footer","tagName":"footer","className":"site-footer-container"} /-->
```
jffng, line 7: I am wondering why this class + CSS rule in blockbase is necessary:

https://github.com/Automattic/themes/blob/b45eaa66b9e91ef52ff4c2db59739ac902a45007/blockbase/sass/base/_layout.scss#L12

MaggieCabrera: I was wondering the same and I don't think we can do that without a class right now, since template parts don't have any props anymore. In Blockbase using that we are doing a sort of Sticky footer situation for pages without content or content that doesn't span the whole height of the viewport. It pushes the footer to the bottom of the page. We could probably achieve this with an extra group wrapper, but I don't think we can remove the class anymore for existing sites not to break.

iamtakashi: So, I'm using the class that is meant for a different purpose. Is it still ok to use it to fix the issue above? Otherwise, what would you recommend to fix the issue?

iamtakashi:
Any further thoughts on the use of the CSS class or any alternative way to solve the issue?

MaggieCabrera (approved):
I don't think we have better means right now to fix this and I don't think this should be a blocker since the parent is using the same class. We can revisit this in the future if we come up with a better solution

PR #6002, opened 2022-05-12
Add a contributing readme

mikachan:
This PR adds a CONTRIBUTING.md file, which contains more info about how to contribute to our repo without adding too much content to our main readme.

I've also added a `.nvmrc` file which allows people who are using NVM to check the correct Node version for the repo by running `nvm use`. I'm happy to take this out as I know not everyone uses NVM, but I thought it was useful for those that do, and doesn't really impact those that don't.

I've also moved some of the content from the main readme into the new file, as it's specifically related to contributing, but happy to move this back as well (this includes the Coding Standards and Sandbox sections).

The main reason for doing this is to help any new contributors set up their environment and point people in the right direction for good issues to work on. I think we could also add short readmes for individual themes (like we have with Blockbase), but they could be separate PRs.


.nvmrc
```diff
@@ -0,0 +1 @@
+16
```
MaggieCabrera, line 1: +1 on having this, it's useful for me

CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Browser Support
+
+We support the same browsers as WordPress Core, [listed here](https://make.wordpress.org/core/handbook/best-practices/browser-support/).
+
+## Contributing CSS
+
+A great way to get started with contributing to our themes is to pick up a CSS fix or patch.
+
+A lot of our themes use SASS for styling, and then use Node to compile the final CSS. Usually, these themes will have a `sass` directory ([example](https://github.com/Automattic/themes/tree/trunk/varia/sass)), and there will also be a `package.json` file at the root ([example](https://github.com/Automattic/themes/blob/trunk/varia/package.json)).
```
MaggieCabrera: Thanks a million for this!!

CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
...
+
+In order to change the styles in these themes, you need to edit the SASS file first, and then run a build command to compile the CSS. Here is the process in most cases, using Varia as an example:
+
+1. Change to the theme's directory `cd varia`
+2. Find the `.scss` file you need to edit in the `sass` directory
+3. Make your changes
+4. Run the build command using `npm run build`
+5. You should see your updates in the newly built `.css` file/s
+
+Please refer to either the themes readme or the themes `package.json` file for theme specific instructions.
```
MaggieCabrera: I'd say, if you are unsure what to do, leave us a comment so that we can improve documentation for other themes that may work differently? Though I can't think of sass being used outside Varia + Seedlet, can you?

CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Best Practices
+
+We use the latest features from the CSS, HTML and JS specifications, where our browser support will allow it. [Caniuse.com](https://caniuse.com/) is a great way to check if a feature is supported.
+
+We use intrinsic web design in our themes where possible. [Every Layout](https://every-layout.dev/rudiments/boxes/) is a good resource for this.
+
+## Sandbox Tools
```
MaggieCabrera, line 102: I think we should add a section about deployment. Not all our contributors have a sandbox and not all of the ones who do have actual commit rights to the pub theme folder, so at least mentioning that and providing alternatives for them (basically telling them to ask us for deployment) would be good.

I'd also add a mention that merging to trunk doesn't equal your work gets deployed directly. Maybe we should have an action that comments on a PR from someone not on the team when it gets approved to let them now what the next steps are?

MaggieCabrera:
Thanks a lot for the initiative here! I left you a few comments. I've seen that so far we've had 3 types of contributors outside the team: devs from other teams, HEs and people from outside A8c that want to contribute, so my comments have all 3 in mind.

CONTRIBUTING.md
mikachan: Thanks, I think what you've said sounds better, plus I didn't like that line about looking at the package.json files..

Yeah I don't think it's likely to be anything outside of Varia and Seedlet!

CONTRIBUTING.md
mikachan: This is a great idea! I'll add a deployment section.

> Maybe we should have an action that comments on a PR from someone not on the team when it gets approved to let them now what the next steps are?

This sounds great too! ✨

mikachan:
Thanks so much for the feedback! I think I've addressed it all, apart from the action for commenting on an approved PR. Might put that in a separate PR.

MaggieCabrera (approved):
LGTM! We can iterate on this ofc, but we can bring this in!

PR #6007, opened 2022-05-13
#5737: Apostrophe 2: Fix VideoPress display if alignwide or alignfull

### LINKED ISSUE #5737: Apostrophe 2: VideoPress videos don't display if set to alignwide or alignfull
### Quick summary

When adding a video from the site's media library to any page or post, using the Apostrophe 2 theme, that video will work fine in the editor, but won't display on the live or preview version of the page if its alignment is set to either "Wide" or "Full"

### Steps to reproduce

1. Start with a site with the Apostrophe 2 theme active
2. create a new page or post, or edit an existing one
3. add a VideoPress block to the post
4. set the block's alignment to either "Full Width" or "Wide Width"
   - Annotation on 2022-03-23 at 14-07-40
5. either preview the page in a new tab, or save it and navigate to it
6.  The image will not be visible
    - screenshot

The image should be visible

### What actually happened

The image is not visible

### Context

reported by a user in 34835320-hc

I changed two of my testing sites to the Apostrophe 2 theme, and was able to recreate with multiple videos. 

Digging into the CSS in dev tools reveals that the videos _do_ load on the page, but are not visible, due to the following CSS rules, which position them outside of the viewable section of their parent element. The rule for `alignwide` is within a media query, so these do work properly on smaller screens. No such luck for the `alignfull` videos.
```
@media (min-width: 1024px) {
   .apostrophe-2-no-sidebar .wp-block-embed.is-type-video.alignwide iframe {
       position: absolute;
       top: 0;
       left: 0;
       bottom: 0;
       right: 0;
       width: 100%;
       height: 100%;
   }
}
```
```
.apostrophe-2-no-sidebar .wp-block-embed.is-type-video.alignfull iframe {
    position: absolute;
    top: 0;
    left: 0;
    bottom: 0;
    right: 0;
    width: 100% !important;
    height: 100% !important;
}
```

### Simple, Atomic or both?

Simple, Atomic

### Theme-specific issue?

Apostrophe 2

_No response_

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

Yes, difficult to implement

### Workaround details

Currently, a user running the apostrophe 2 theme can't set videos to Wide or Full Width. It might be possible to override this with some custom CSS, but I have not had any luck writing that so far. 

Otherwise, their options are to either compromise and use different alignment settings, or host the video elsewhere (YouTube or Vimeo) and embed it on their site.

bobmatyas:
##### Before

Screenshot on 2022-05-13 at 11-39-57

##### After

Screenshot on 2022-05-13 at 11-54-52

#### Related issue(s):

https://github.com/Automattic/themes/issues/5737

### DIFF
--- a/apostrophe-2/css/blocks.css
+++ b/apostrophe-2/css/blocks.css
@@ -93,7 +93,6 @@ body {
 
 .apostrophe-2-no-sidebar .wp-block-embed.is-type-video.alignfull iframe {
 	width: 100% !important;
-	height: 100% !important;
 }
 
 .apostrophe-2-no-sidebar .wp-block-embed.is-type-video.alignfull:before {
@@ -101,15 +100,6 @@ body {
 	display: block;
 }
 
-.apostrophe-2-no-sidebar .wp-block-embed.is-type-video.alignfull iframe {
-	position:  absolute;
-	top: 0;
-	left: 0;
-	bottom: 0;
-	right: 0;
-	width: 100% !important;
-	height: 100% !important;
-}
 
 /* Wide Width */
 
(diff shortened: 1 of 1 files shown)


pbking (approved):
Works as advertised.
No obvious side-effects.
🚢

PR #6008, opened 2022-05-17
Rebalance: Add page password protection for portfolio page

wojtekn:

Rebalance theme has two issues related to the password-protected page:
* it doesn't show the password form if the page is protected with a password. It happens because the Portfolio Page template doesn't display the page content using the `the_content()` function, which internally displays the password form.
* portfolio items are displayed even if the page is protected. It behaves that way because portfolio items are not wrapped with a password check. 

This PR fixes both issues: it wraps the code responsible for displaying the portfolio with the condition that checks for password protection. Then it displays `the_content()` if a password is required. I used a similar implementation to one from the Blask theme.

#### Testing steps

1. Log in to WP Admin
3. Ensure that Rebalance theme from this PR is installed and enabled
* add any content
* ensure that “Template" is “Portfolio Page.”
* set “Visibility” to “Password protected,” choose a password

The expected behavior is that portfolio items and page content are not displayed, and the password form is displayed instead. Then, when the user enters the password, they should see portfolio items. They shouldn't see the page content.

#### Related issue(s):

### DIFF
--- a/rebalance/portfolio-page.php
+++ b/rebalance/portfolio-page.php
@@ -15,74 +15,93 @@ get_header(); ?>
 	<div id="primary" class="content-area">
 
 		<?php
-			/*
-			 * Include the Featured Project loop.
-			 * - if featured posts exist, collect their IDs and exclude them from the main loop
-			 */
-			if ( rebalance_has_featured_projects( 1 ) ) {
-				$featured_ids = rebalance_get_featured_project_ids();
-				get_template_part( 'template-parts/section', 'featured' );
-			} else {
-				$featured_ids = null;
-			} ?>
+		if ( post_password_required() ) :
+				the_content();
 
-		<main id="main" class="site-main" role="main">
+			else :
 
-			<?php // Set Up New Query
+				/*
+				 * Include the Featured Project loop.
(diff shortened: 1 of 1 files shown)


p-jackson (commented):
This theme seems a bit weird in how it doesn't include any content from the portfolio template page in the final output. But I guess that must be how it's always worked. Everything works as expected during testing. Just added a note below.

rebalance/portfolio-page.php
```diff
@@ -15,74 +15,93 @@
...
-				get_template_part( 'template-parts/section', 'featured' );
-			} else {
-				$featured_ids = null;
-			} ?>
-
-		<main id="main" class="site-main" role="main">
-
-			<?php // Set Up New Query
+		if ( post_password_required() ) :
+				the_content();
```
p-jackson, line 20: Could we add a small explanatory comment for this line? I think this is a fine method for showing the "this is password protected" message and input box, but it's not that intuitive.

wojtekn: Thanks for the suggestion, I've added the comment.

p-jackson (approved):
LGTM

PR #6010, opened 2022-05-17
Move Blockbase Google Fonts to Locally Hosted ( DO NOT MERGE )

creativecoder:

- Migrate custom font settings to be more compatible with Global Styles (solves https://github.com/Automattic/wp-calypso/issues/63229)
- Add Google Fonts to theme.json to populate Font family options in the editor
- Include Jetpack package for loading Google fonts from the editor settings, so that Jetpack plugin is not required
- Remove Customizer font settings and add deprecation message that links to Global styles
- Disable Jetpack Google Fonts module, if active, so it doesn't conflict with Blockbase fonts

**Note:** This migrates the Heading font from the Customizer to the Global Styles settings for the Post Title and Heading blocks. The Heading block typography settings require Gutenberg 13.3, which has not yet been released, so this PR should wait to be merged until then.

#### Testing Instructions
- Activate a blockbase theme ( Quadrat, Bennett, Blockbase, etc. )
- Before applying this branch, set custom fonts from Customizer > Fonts

Screen Shot 2022-05-25 at 12 21 05 AM

- Navigate to Appearance > Editor ( beta )
- Verify that custom google fonts are not selectable in the global styles typography dropdown of the site editor

Screen Shot 2022-05-25 at 12 22 56 AM

- Apply this branch and reload a page to migrate the settings
- You should see the body font you selected in Global Styles > Typography

Screen Shot 2022-05-25 at 12 26 46 AM

- You should see the heading font you selected in Global Styles > Blocks > Heading > Typography (currently requires Gutenberg `trunk`) and Global Styles > Blocks > Post Title > Typography

Screen Shot 2022-05-25 at 12 26 58 AM

- You should see all the fonts that Blockbase provides in the list of font families provided by Global Styles and Block Settings

Screen Shot 2022-05-25 at 12 27 58 AM

- When you navigate back to the customizer, you should see a deprecation notice
Screen Shot 2022-05-24 at 11 49 09 PM

**Note:**
- To test repeatedly, delete the `blockbase_legacy_font_settings` from wp_options database table to trigger the settings migration (migration only runs when this option is not present).


MaggieCabrera:
Thanks for this! the team is in a meetup this week, I added this PR to our board so we'll review as soon as possible

jeyip:
## Testing
### Requirements
**Custom fonts in the site editor**
- [x] All the fonts that Blockbase provided in the customizer should now be available in Global Styles and Block Settings
- [x] Font settings that were previously set in the customizer should be applied automatically to global styles and the site editor
- [x] Font settings that were previously set should still be displayed in the frontend
- [x] The blockbase fonts customizer feature should display a deprecation notice once legacy data is migrated 

**Google Fonts Jetpack module**
- [x] If Blockbase or a Blockbase child theme is active, the google fonts jetpack module should not allow toggling. It should immediately be disabled to prevent duplicating fonts in the editor font family settings (blocks and Global Styles)
- [x] If a Blockbase related theme is not active,  the google fonts jetpack module should allow toggling.

### Browsers
- [x] Edge
- [x] Firefox
- [x] Chrome

creativecoder:
Noting that the "Use Site Editor" link in the Fonts section of the Customizer is meant to open the Site editor with the Global Styles panel open.

image

This doesn't currently work with the WP 6.0 release candidate... tracking that in https://core.trac.wordpress.org/ticket/55752.

jeyip:
🎉 🎉 🎉 Tested for dotcom simple sites. I can confirm that a site with legacy blockbase custom fonts was migrated successfully. One thing to note is that I couldn't confirm the heading font was mapped to global styles heading blocks because typography settings won't be exposed until Gutenberg v13.3. Fonts _did_ map successfully to the global styles post title block though.

To save others the trouble, I navigated to the themes/blockbase directory in my local machine from the terminal. Then I ran this command  `rsync -azP . sandbox_username@sandbox_address:~/public_html/wp-content/themes/pub/blockbase` where sandbox information should be replaced with your respective credentials.

Screen Shot 2022-05-19 at 2 59 34 AM

Screen Shot 2022-05-19 at 2 59 57 AM


I'll test this for atomic sites tomorrow.

MaggieCabrera:
I tested this locally, with Gutenberg installed but not Jetpack on a site running Quadrat with custom fonts selected via the customizer. As soon as I checked out this branch, the fonts stop working, they are not being loaded at all. I don't see the font-face rules either for them, but the CSS is requesting the correct font family name.

Screenshot 2022-05-23 at 11 06 00

jeyip:
> I tested this locally, with Gutenberg installed but not Jetpack on a site running Quadrat with custom fonts selected via the customizer. As soon as I checked out this branch, the fonts stop working, they are not being loaded at all.

I have a hunch that this is related to the testing environment and not the code itself -- things had to be run in a specific way when I was testing the PR alongside @creativecoder. Taking a look at your issues today.

jeyip:
@MaggieCabrera thanks for the patience in testing. It turns out that what you found was, indeed, a bug 😅 

I've been mistakenly thinking that the behavior we've been seeing has been a caching issue, because, when we select the font in global styles and refresh the page, the font _does_ load properly.

The issue was that @creativecoder and I didn't realize that font enqueueing for the editor and the front share the same url generation method `blockbase_fonts_url`. We added font filtering to the function to avoid enqueuing fonts that aren't used in the frontend, and were inadvertently filtering out enqueuing those fonts in the editor as well.

We now enqueue all google fonts options in the editor. The issue should be fixed now 🎉 


**Note:**
Before testing again, you'll want to remove the `blockbase_legacy_font_settings` from the `options` table to completely reset the environment

MaggieCabrera:
> @MaggieCabrera thanks for the patience in testing. It turns out that what you found was, indeed, a bug 😅
> 
> I've been mistakenly thinking that the behavior we've been seeing has been a caching issue, because, when we select the font in global styles and refresh the page, the font _does_ load properly.
> 
> The issue was that @creativecoder and I didn't realize that font enqueueing for the editor and the front share the same url generation method `blockbase_fonts_url`. We added font filtering to the function to avoid enqueuing fonts that aren't used in the frontend, and were inadvertently filtering out enqueuing those fonts in the editor as well.
> 
> We now enqueue all google fonts options in the editor. The issue should be fixed now 🎉
> 
> **Note:** Before testing again, you'll want to remove the `blockbase_legacy_font_settings` from the `options` table to completely reset the environment

I'm sorry but something is still broken for me. I tested on a fresh site, my setup is basically a fresh install using Local by Flywheel with a symlink to my themes and plugin folder (using Gutenberg trunk). I'm testing again with Quadrat, first selecting fonts in the customizer while on trunk, then switch to this branch and check the frontend. In this case I see that the CSS variable that should be applied is not defined:

Screenshot 2022-05-24 at 11 56 06

Debugging this a bit I see that when running `blockbase_fonts_url()` `$blockbase_block_font_families` is an empty array. `$fonts_to_load` is showing the fonts I selected in the customizer and `$font_families` are showing the original font families from the Quadrat theme theme.json file (not the user selected ones). Hope that helps narrow it down? Let me know if I can help with this more.

jeyip:
> I'm sorry but something is still broken for me. I tested on a fresh site, my setup is basically a fresh install using Local by Flywheel with a symlink to my themes and plugin folder (using Gutenberg trunk). I'm testing again with Quadrat, first selecting fonts in the customizer while on trunk, then switch to this branch and check the frontend

Thanks again for the patience @MaggieCabrera! I think we've got the issues sorted this time. I tested our changes for Quadrat in the editor + frontend and things should be working as expected.

We, ultimately, had to do a noticeable amount of refactoring behind the scenes. We're effectively copying the functionality for custom google fonts from the dotcom Jetpack custom fonts implementation. The end result, however, is still the same, and delivers the same functionality that we agreed upon earlier paYE8P-1B3-p2#comment-1526.

- No dependencies on the Jetpack plugin
- The user will update fonts in the site editor
- The Blockbase fonts feature in the customizer will be deprecated
- There will no longer be conflicts when Blockbase is activated on a WordPress installation where the Jetpack custom google fonts feature is enabled

This PR is ready for review again 👍

blockbase/functions.php
```diff
@@ -187,3 +134,99 @@ function blockbase_core_navigation_render_fallback() {
...
+	$fonts_to_register = apply_filters( 'blockbase_google_fonts_list', BLOCKBASE_GOOGLE_FONTS_LIST );
+
+	foreach ( $fonts_to_register as $font_family ) {
+		wp_register_webfonts(
+			array(
+				array(
+					'font-family'  => $font_family,
+					'font-weight'  => '100 900',
+					'font-style'   => 'normal',
+					'font-display' => 'fallback',
```
mikachan: It looks like we previously used `swap` for the font-display of Blockbase fonts, and it also looks like that's what [Google recommends](https://fonts.google.com/knowledge/using_type/using_web_fonts#optimizing-font-loading) for their fonts.

I don't think this is a big issue, but it would be nice to use the same setting we did previously, unless there's a specific reason for using `fallback`.

mikachan (commented):
I followed the test instructions and this works really well for me in Quadrat green and pink variations, and Geologist (and variations), but it doesn't seem to work in Blockbase itself or Quadrat (I haven't tested any other Blockbase themes yet).

- Tested with and without Jetpack activated
- Tested in Chrome, Safari and Firefox
- Tested using GB trunk and WordPress 6.0

For Quadrat variations, and Geologist and its variations:
- The correct font loads on the front end and in the editor
- I can change the font in the editor for both the body and the Post Title from the full selection of fonts
- I can see the new deprecation notice in the customizer

In Blockbase and Quadrat, the fonts don't seem to be registered, and the `webfonts-inline-css` style tag isn't added on the front end. I also don't see all the fonts loaded in the Editor:

image

I'm not sure what the difference is between the themes. If it helps, when switching between the above themes, the fonts both consistently work and don't work as described above. Let me know if you need any more details.

pbking:
The biggest problem I see at the moment is the critical error thrown if Gutenberg isn't activated.  Blockbase should be operable without Gutenberg (as of WP 5.9).  I'm currently testing on WP 6.0.

[edit]
Shoot.  I just realized that it's a problem elsewhere in Blockbase and 6.0. (!) Doesn't seem to be exclusive to this branch.

[edit(again)]
completely false alarm.  The error was in another experimental plugin I had loaded.

Continuing evaluation...

blockbase/functions.php
jeyip: Good catch! There's no specific reason -- the logic was carried over from our dotcom, site editor specific implementation of custom google fonts. I'll change to swap now 👍

jeyip: Fixed in https://github.com/Automattic/themes/pull/6010/commits/4d2dbb90c6859ec8fea83639d2df9082480ae3ed

jeyip:
> In Blockbase and Quadrat, the fonts don't seem to be registered, and the webfonts-inline-css style tag isn't added on the front end. I also don't see all the fonts loaded in the Editor:

Interesting 🤔 I'm unable to reproduce these problems on my local machine. I'm taking a closer look this morning -- I'll reach out again after I do some digging. Thanks for taking a look!

> If it helps, when switching between the above themes, the fonts both consistently work and don't work as described above.

@mikachan Would you mind elaborating on this? Do you mean that the you:

- Have Quadrat activated
- Walk through the testing steps
- See fonts not working properly
- Switch to Blockbase
- Switch back to Quadrat
- See fonts google fonts working

pbking:
I was able to reproduce (at least part) of this today. 

* With Blockbase /trunk Activated
* With Gutenberg /trunk Activated
* Use Customizer to assign a font to the Body (Arvo)
* Switch to this branch
* Render the view
* Note that Arvo is not being rendered in the view (due to the above notes from @mikachan)
image
* Note that Arvo IS being rendered in the FSE, though it is the only option.
image

pbking:
Where my testing diverges from @mikachan is that the above is also happening with Geologist for me.

I've also been switching back-and-forth between WP5.9 and WP6.0 and the above is also happening for me in both versions.

mikachan:
> Would you mind elaborating on this?

Yes, sure. These are the steps I followed when I was testing earlier:

- Activated Quadrat green variation (this was the first theme I tested with)
- Ran through the testing steps
- Saw the fonts working properly
- Switched to Quadrat (non-variation version)
- Saw the fonts NOT working properly
- Switched to Geologist
- Saw the fonts working properly
- Switched to Blockbase
- Saw the fonts NOT working properly
- Switched to each Quadrat variation
- Saw the fonts working properly in each variation
- Switched to Blockbase
- Saw the fonts NOT working properly
- Switched to each Geologist variation
- Saw the fonts working properly

Sorry, there are a lot of steps there... I think my main point was that I was seeing the fonts consistently working in Quadrat green and pink variations, Geologist, and Geologist variations, but when I switched to Blockbase or Quadrat, I could see them consistently not working. I thought it was worth highlighting as I can't understand what the difference between these themes would be that would cause them to interfere with this new functionality (i.e. I would have thought it would either consistently work OR not work across different Blockbase themes).

Now seeing @pbking's comment as well, I'm wondering if it's not the specific themes that are the issue, but rather the switching between Blockbase themes that causes an issue?

jeyip:
> Where my testing diverges from @mikachan is that the above is also happening with Geologist for me.
>
> I've also been switching back-and-forth between WP5.9 and WP6.0 and the above is also happening for me in both versions.

Just to make sure @pbking, are we removing the `blockbase_legacy_font_settings` site option each time we're testing a new theme? I left a note in the PR description above, but it might have been lost amongst all the instructions.

pbking:
I notice that as soon as I "Reset to Defaults" using any of the themes I'm testing with the list of available fonts is available and choosing from that list and saving per-instructions makes it work as expected.

It's only when the fonts are customized via the Customizer that the errors occur.

pbking:
> Just to make sure @pbking, are we removing the blockbase_legacy_font_settings site option each time we're testing a new theme?

No, I wasn't clearing that.  Though I'm not sure that seems like a great way to flag what needs to happen.  That might be the cause of what I'm seeing though.

jeyip:
> Though I'm not sure that seems like a great way to flag what needs to happen.

We can definitely chat more about this 👍 The intention is to ensure that the migration is only run a single time, but happy to make modifications if you have suggestions

pbking:
That 'migration' is definitely what I've been missing.  I completely passed over that change when I was reviewing.  Thanks for pointing that out.  

I think my biggest qualm is that it seems to be per-site rather than per-theme.  I think a better solution would be to store that transition as a `theme_mod` instead.  If I've customized Blockbase and Geologist then only the first that runs will be successful.  If I then switch to another Blockbase-based theme then I'm out of luck (unless I'm willing to remove ALL customizations to get that theme back into a font-editable state).

mikachan:
Ah, I wasn't clearing `blockbase_legacy_font_settings` either, this must be the cause. 🙈

I agree that this doesn't seem like the best way to flag this, for the same reason @pbking mentioned. +1 for a per-theme check 👍

pbking:
I also want to note that in some templates/patterns (in Blockbase [the footer-left template part] and other child themes theme.json [meraki: navigation-link] as examples) leverage the `heading-font` slug in places other than just the headings and post titles.  With this migration those designs will (I expect) just roll back to the the body font instead.  I'm not sure of the best solution to handle those (or the other child themes that we aren't aware of).

jeyip:
> I think my biggest qualm is that it seems to be per-site rather than per-theme. I think a better solution would be to store that transition as a theme_mod instead. If I've customized Blockbase and Geologist then only the first that runs will be successful. If I then switch to another Blockbase-based theme then I'm out of luck (unless I'm willing to remove ALL customizations to get that theme back into a font-editable state).

> I agree that this doesn't seem like the best way to flag this, for the same reason @pbking mentioned. +1 for a per-theme check 👍

This makes total sense -- I'll start updating the PR

jeyip:
@pbking @mikachan 

I ended up [switching the logic to use theme mods](https://github.com/Automattic/themes/pull/6010/commits/ab22d3a3b081ba3f8c8075e024cba0e5d8e45d56). Now, migration of the legacy custom font format for Blockbase will run once on a per theme basis, rather than a per site basis.

That is to say, things will behave in a more expected way if we do the following:

- Use the custom fonts feature while a blockbase theme like Quadrat is activated
- Switch to another blockbase theme like Bennett and use the custom fonts feature again
- Apply this PR
- Switch between the two themes

We should now see google fonts as selectable options in site editor global styles in all instances. I tested this behavior between Blockbase, Quadrat, and Bennett.

mikachan (commented):
Thanks @jeyip, this is working great for me! 🎉

I tested this following the same steps as before, and I can see all the available fonts both in the Editor and as registered fonts on the front end, in all the Blockbase themes I switched to (including Blockbase itself). Choosing different fonts in each theme worked well, too.

I'm happy to approve but if possible I'd like someone else from the themes team to double-check as well (@MaggieCabrera @pbking 🙏).

blockbase/functions.php
```diff
@@ -187,3 +134,99 @@ function blockbase_core_navigation_render_fallback() {
...
+					'provider'     => 'blockbase-google-fonts',
+				),
+				array(
+					'font-family'  => $font_family,
+					'font-weight'  => '100 900',
+					'font-style'   => 'italic',
+					'font-display' => 'swap',
+					'provider'     => 'blockbase-google-fonts',
+				),
+			)
```
creativecoder: I think we actually have the opportunity to be more selective here with the font weights by including only those needed by the themes, reducing the number of font-face rules that are output on the page.

What do you think of modifying `BLOCKBASE_GOOGLE_FONTS_LIST` to include the `google` information from `GlobalStylesFontsCustomizer->$fonts`, and load only the weights specified for each font?

blockbase/functions.php
```diff
@@ -187,3 +134,99 @@ function blockbase_core_navigation_render_fallback() {
...
+);
+
+/**
+ * Register a curated selection of Google Fonts.
+ *
+ * @return void
+ */
+function blockbase_register_google_fonts() {
+	// Use jetpack's implementation of custom google fonts if it is already active
+	if ( method_exists( 'Jetpack', 'is_module_active' ) && Jetpack::is_module_active( 'google-fonts' ) ) {
```
creativecoder: Would be a good idea to check when wpcomsh is auto-activating the Jetpack Google Fonts module, and move that to fire before this function, to make sure this check is working as expected. Otherwise this might run adding fonts, then the Jetpack module gets activated afterward, and we'd have duplicate fonts for a request.

blockbase/inc/customizer/wp-customize-fonts.php
```diff
@@ -3,552 +3,177 @@
...
-		$font_families[] = 'display=swap';
+	// converts data to array (in some cases settings and styles are objects insted of arrays)
+	$new_settings = (array) $global_styles->data['settings'];
+	$new_styles   = (array) $global_styles->data['styles'];
 
-		// Make a single request for the theme fonts.
-		return esc_url_raw( 'https://fonts.googleapis.com/css2?' . implode( '&', $font_families ) );
+	// Set new typography settings
+	if ( isset( $new_settings['typography']['fontFamilies'] ) ) {
+		unset( $new_settings['typography']['fontFamilies'] ); // TODO: Reconsider the depth of property we're unsetting
```
creativecoder: I believe we tested this and it needed to be `['typography']['fontFamilies']`, as `['typography']['fontFamilies']['custom']` didn't work.

creativecoder (commented):
@jeyip I left some thoughts looking over the latest version of the code.

blockbase/functions.php
creativecoder: (I'm wondering if we should do something similar with the Jetpack Google Fonts module, as well 🤔)

creativecoder:
>  think my biggest qualm is that it seems to be per-site rather than per-theme. I think a better solution would be to store that transition as a theme_mod instead.

That sounded good to me at first, but thinking through it, I'm not sure what's actually better about using a theme mod. As things are now (outside this PR), the Customizer font settings are stored as a Global Styles setting for the site (settings > typography > fontFamilies > custom), there's no separate setting for each theme that I'm aware of. So I don't see an advantage to running the migration code separately for each theme. The old settings will be gone after the first run.

jeyip:
> That sounded good to me at first, but thinking through it, I'm not sure what's actually better about using a theme mod. As things are now (outside this PR), the Customizer font settings are stored as a Global Styles setting for the site (settings > typography > fontFamilies > custom), there's no separate setting for each theme that I'm aware of

@creativecoder Global styles has separate custom post types for each theme ( e.g. `wp-global-styles-bennett`, `wp-global-styles-blockbase`, etc. ) Each custom post type might have its own set of stored customizer font settings. To elaborate, Arvo + Bodoni Moda might be set as body and heading fonts in Bennett while Space Mono + EB Garamond might be set in Blockbase. In that sense, isn't the blockbase custom fonts feature storing a separate setting for each theme?

blockbase/functions.php
jeyip: > What do you think of modifying BLOCKBASE_GOOGLE_FONTS_LIST to include the google information from GlobalStylesFontsCustomizer->$fonts, and load only the weights specified for each font?

I think this is fair, but would we mind doing it in a follow-up, seeing as how this is the same behavior that's already implemented in production for the Jetpack Google Fonts Module? That way we can tackle it in Blockbase 
_and_ Jetpack at the same time within the scope of another issue. Happy to create a ticket for us if you're okay with it. 👍 

> (I'm wondering if we should do something similar with the Jetpack Google Fonts module, as well 🤔)

I agree

blockbase/functions.php
jeyip: Good catch 👍 

Drafting a PR now

blockbase/inc/customizer/wp-customize-fonts.php
jeyip: > I believe we tested this and it needed to be ['typography']['fontFamilies'], as ['typography']['fontFamilies']['custom'] didn't work.

Yes! You're right. I think it's a case of us forgotting to remove the inline comment

jeyip: Removed in https://github.com/Automattic/themes/pull/6010/commits/1782be8707c829cd81d84732317c703abfd57d7c

blockbase/functions.php
jeyip: @creativecoder 959-gh-Automattic/wpcomsh

pbking:
> there's no separate setting for each theme that I'm aware of. 

There's definitely a separate collection of values for each theme.  Global Styles are per-theme; setting colors when Blockbase is activated has no effect when I activate Quadrat.  The same principal is here; modifications to font (or any 'user theme.json Global Styles value') is per-theme.  Thus migrating one theme has no effect on another, and if the flag is true site-wide then we have a "false positive" that a theme has been migrated when it has not.

blockbase/functions.php
pbking: From what I can tell the above is including only a single weight per-font family.  I'm not sure I see how more selective it can be.  Or perhaps I'm reading it wrong...

jeyip: > From what I can tell the above is including only a single weight per-font family. I'm not sure I see how more selective it can be. Or perhaps I'm reading it wrong...

Blockbase prescribes different font weight ranges for each font family, which is different from the 100 to 900 range we specify for _all_ fonts with the aforementioned code.

Screen Shot 2022-05-26 at 8 53 30 AM

I haven't verified anything yet, but I would've assumed a smaller range would optimize the font family requests?

blockbase/functions.php
```diff
@@ -187,3 +134,99 @@ function blockbase_core_navigation_render_fallback() {
...
+		return;
+	} 
+
+	if ( ! function_exists( 'wp_register_webfont_provider' ) || ! function_exists( 'wp_register_webfonts' ) ) {
+		return;
+	}
+
+	wp_register_webfont_provider( 'blockbase-google-fonts', '\Automattic\Jetpack\Fonts\Google_Fonts_Provider' );
```
zaguiini: Isn't the idea of the provider to be a singleton across all themes/sites? I don't see a reason for every theme to register one for itself 🤔

jeyip: @zaguiini We are currently delivering custom google fonts through the Jetpack google fonts module. One of the requests from the themes team was to avoid a dependency between Blockbase and the Jetpack plugin. You can read more about the decision in this P2 comment paYE8P-1B3-p2#comment-1526 and you see the alternative options we explored in this comment paYE8P-1B3-p2#comment-1522

This is why we end up using a lot of the same logic to migrate the Blockbase font customization feature to use global styles, and in this case, it's why we register a unique google fonts provider.

> I don't see a reason for every theme to register one for itself 🤔

Blockbase is an exception and an unfortunate result of how the legacy google font customization was implemented for these universal themes. A lot of this is covered in the P2 paYE8P-1B3-p2

zaguiini: Ah I see, makes total sense 👌

creativecoder:
> @creativecoder Global styles has separate custom post types for each theme

🤦‍♂️ Of course, thanks for reminding me of that. In that case using a theme mod for the settings migration is much better.

blockbase/functions.php
creativecoder: > I think this is fair, but would we mind doing it in a follow-up, seeing as how this is the same behavior that's already implemented in production for the Jetpack Google Fonts Module?

That's fine by me!

blockbase/inc/customizer/wp-customize-fonts.php
```diff
@@ -3,552 +3,177 @@
...
+		add_action( 'customize_register', array( $this, 'init_deprecation_notice' ) );
 	}
 
-	function get_font_family( $array, $configuration ) {
-		$variable = get_settings_array( $array, $configuration );
-		$slug     = preg_replace( '/var\(--wp--preset--font-family--(.*)\)/', '$1', $variable );
-		if ( ! isset( $this->fonts[ $slug ] ) ) {
-			$this->fonts[ $slug ] = $this->build_font_from_theme_data( $slug, $configuration );
+	function generate_deprecation_message() {
+		if ( ! defined( 'GUTENBERG_VERSION' ) || version_compare( GUTENBERG_VERSION, '13.3', '<=' ) ) {
```
creativecoder, line 137: Note that I updated this from `13.2` to `13.3`. WP.com currently has `13.2` and I'm not seeing the font family settings in Global Styles > Blocks > Heading > Typography

blockbase/inc/customizer/wp-customize-fonts.php
```diff
@@ -3,552 +3,178 @@
...
-		),
-		'work-sans'         => array(
-			'fontFamily' => '"Work Sans", sans-serif',
-			'slug'       => 'work-sans',
-			'name'       => 'Work Sans',
-			'google'     => 'family=Work+Sans:ital,wght@0,100..900;1,100..900',
-		),
-	);
+// Use an early priority to migrate legacy font settings before registering Google Fonts in `blockbase_register_google_fonts`.
+add_action( 'after_setup_theme', 'migrate_blockbase_custom_fonts', 0 );
```
creativecoder, line 7: Note that I moved this to an earlier priority so it runs before `blockbase_register_google_fonts`. I was seeing one request (when the migration was run) where the fonts didn't load on the front end, and this seems to fix it.

creativecoder (commented):
This is working well for me!

I tested my local dev site with and without Jetpack, as well as copying up the theme code to a Simple Site, and all is working as expected.

I'd approve this, but can't because I made the PR 😛

---

Also, for Simple Sites, we'll need to be sure to update both the `pub` and `premium` versions of Blockbase, so our premium Blockbase themes, like Videomaker, get the update as well.

---

One thing that could be improved (probably in a follow-up PR) is removing the `heading-font` ("Headings" in the screenshot below) and all it's uses in theme.json (and child theme.json). Without the Customizer setting, this doesn't really have any meaning, since it's the same font family as `body-font` (which maybe should just be called `system-font` now?)

image

pbking:
I think I was so focused on testing that this worked with user-modified fonts that I failed to test default fonts for child themes...

Considering that child themes express their font preferences in the same structure as user modifications I think that the portion that checks for "untransformed" font settings needs to evaluate not just 'custom' but also the value coming from the theme itself.

To replicate the issue:

* Starting with a clean slate (no theme_mods saved, no customized fonts, or Global Styles settings)
* Activate Quadrat
* Render a page
* Note that the font being rendered is the fallback 'Helvetica' rather than 'DM Sans' which is the typeface defined in Quadrat's theme.json

jeyip:
> Considering that child themes express their font preferences in the same structure as user modifications I think that the portion that checks for "untransformed" font settings needs to evaluate not just 'custom' but also the value coming from the theme itself.

Checking this out today 👀 

_Edit:_
Wow. I have no idea how you spotted this but great eyes @pbking 🤯 

| DM Sans | Helvetica |
| ----- | ----- |
| Screen Shot 2022-05-27 at 10 22 51 AM  | Screen Shot 2022-05-27 at 10 23 03 AM |

jeyip:
@pbking

The problem was twofold. We needed to migrate default theme settings, but there was also the fact that default theme fonts weren't working at all. This is because [theme defaults use slugs like "body-font" and "heading-font"](https://github.com/Automattic/themes/blob/bb8d20885067bd6a22bc1be9b7c0f49c299273eb/quadrat/theme.json#L239-L257), which create CSS custom properties that don't work properly with the new webfonts API.
```javascript
// Quadrat default theme.json font
{
    "fontFamily": "\"DM Sans\", sans-serif",
    "slug": "body-font",
    "name": "Body (DM Sans)",
   ...rest
}

// Theme default font CSS custom property
var:preset|font-family|body-font

// CSS custom property recognized by webfonts API
var:preset|font-family|dm-sans
```

Gonna need guidance here because I made an educated guess as to how we'd like to proceed. As an experiment to address the problem, I updated the theme.json typography setting of Quadrat extensively.

```javascript
// Before
{
    "fontFamily": "\"DM Sans\", sans-serif",
    "fontSlug": "dm-sans",
    "slug": "body-font",
    "name": "Body (DM Sans)",
    "google": "family=DM+Sans:ital,wght@0,400;0,500;0,700;1,400;1,500;1,700"
},
{
    "fontFamily": "\"DM Sans\", sans-serif",
    "fontSlug": "dm-sans",
    "slug": "heading-font",
    "name": "Headings (DM Sans)",
    "google": "family=DM+Sans:ital,wght@0,400;0,500;0,700;1,400;1,500;1,700"
}

// After
{
    "fontFamily": "\"DM Sans\", sans-serif",
    "slug": "dm-sans",
    "name": "Theme (DM Sans)"
}
```
- I swap the values for `fontSlug` and `slug` so that we create an appropriately named css custom property
- I remove the `google` and `fontSlug` properties
- I remove references to a heading font since it's not explicitly supported in global styles in the site editor
- I rename Body (DM Sans) to Theme (DM Sans)

With the correct CSS custom property being generated, the theme default works properly. This has the added benefit of aligning more closely with the traditional fonts API for theme.json settings.

The one hiccup is that we were using theme font defaults from theme.json settings to determine how to map body and heading fonts to the appropriate global styles. Because we consolidate body and heading fonts, we're no longer able to map the default fonts properly, so I had to create a dictionary in the migration script. https://github.com/Automattic/themes/pull/6010/commits/64ed40d15b20ff19b887555b2ff7fc8204441697#diff-1e8f5c8da22610cdd824ed551bb21abc65759f8961c8ea9d777a213a483c8284R111-R116

Does this seem reasonable? If so, we'd need to update each blockbase child's theme.json. I'm happy to take care of that when I'm back next week.

### Testing
- Starting with a clean slate (no theme_mods saved, no customized fonts, or Global Styles settings)
- Checkout this blockbase PR
- Activate Quadrat
- Load any `wp-admin` page to run the migration
- Navigate to the frontend
- Note that the font being rendered is `DM Sans`, and _not_ the `sans-serif` Helvetica fallback

| Before | After |
| ----- | ----- |
| Screen Shot 2022-05-27 at 3 55 44 PM |  Screen Shot 2022-05-27 at 4 47 05 PM

Any thoughts @MaggieCabrera @pbking @mikachan @creativecoder?

pbking:
Considering that there will be themes in the wild that we don't have control of that handle fonts in the same way that Quadrat does I'm extra keen to solve this in a way that doesn't require changes to the child theme.  I'm going to tinker with that today and see if I can come up with any suggestions.

The  'heading-font' and 'body-font' is the part I mentioned earlier where some themes were depending on those values in their patterns/settings.  It's not guaranteed to be consistent across themes how each of those values are used, only that they had been available to use.

jeyip:
@pbking 

After more consideration, @creativecoder and I took a closer look and realized that we could automatically enqueue default Blockbase theme google fonts. See our latest commit https://github.com/Automattic/themes/pull/6010/commits/a6bdb080f707cdf2544e42ebd0e5fe591f44298f. 

It appears as if the bug is fixed, and it seems like the solution satisfies all of the constraints you mentioned earlier:

- Doesn't require changes to child themes
- Leaves the `heading-font` and `body-font` values the same

Thoughts?

pbking:
> After more consideration, @creativecoder and I took a closer look and realized that we could automatically enqueue default Blockbase theme google fonts. See our latest commit https://github.com/Automattic/themes/commit/a6bdb080f707cdf2544e42ebd0e5fe591f44298f.

Brilliant!  Love it, it's the perfect fix for that situation IMO.  Now a child theme doesn't have to be updated to continue to work as expected.

That said a few child themes do have reference to heading-font and body-font outside of the defaults being set in the migration logic.  I think the best bet for those (few) cases are to go ahead and commit those changes to configuration.

[Meraki](https://github.com/Automattic/themes/blob/trunk/meraki/theme.json#L143) ([another](https://github.com/Automattic/themes/blob/trunk/meraki/theme.json#L169))
A Blockbase [Pattern](https://github.com/Automattic/themes/blob/trunk/blockbase/inc/patterns/footer-left.php#L19)

...and I think that's it for the Blockbase children.  So only Meraki really NEEDS to be updated with this change to maintain design.  (Note: I haven't looked for the same in any of the /premium blockbase children).


Just a few more (housekeeping) changes I suggest; 
* The color customizer files (aside from the code being added) have no use now and could be removed in this PR.  (There's a similar one for [colors](https://github.com/Automattic/themes/pull/6043)). 
* Can we move the font-related additions from ./functions.php to ./inc/customize-fonts.php?  (Renamed to be more appropriate if you feel that's needed since most of it isn't for the Customizer anymore).

jeyip:
@pbking 

Addressed the comments. I removed the unnecessary fonts files and moved fonts logic into the`inc` directory. I also updated Meraki's configuration. I can make a follow up for changes to the blockbase pattern.

Hopefully everything looks good now. As far as the update to premium themes, let me know how I can coordinate the deploy of this PR with your changes to premium logic.

pbking:
> let me know how I can coordinate the deploy of this PR with your changes to premium logic.

I have no intentions of updating any premium themes, I just wanted to point out that they should. :) 

I understand the changes to Blockbase will be pulled into the "premium" version at which point any changes like those for Meraki will need to happen then.  I'm not sure who will coordinate that.

As to the rest of the changes I think we're solid!  Thanks for working so hard on this!  I'mma approve these changes though I'm also going to stick a "do not merge" tag on this change as well.  I intend for this change and the color customization change to ship together in a major point release later this week.

pbking (dismissed):
A LOT OF WORK HERE!

A good change and improvement to Blockbase.

Approving, requesting that it wait to be merged.

jeyip:
> I have no intentions of updating any premium themes, I just wanted to point out that they should. :)

Gotcha 👍 Thanks for clarifying!

> A LOT OF WORK HERE!

> A good change and improvement to Blockbase.

> Approving, requesting that it wait to be merged.

Thanks for sticking with us through the reviews -- it's been much appreciated 🙏 

> As to the rest of the changes I think we're solid! Thanks for working so hard on this! I'mma approve these changes though I'm also going to stick a "do not merge" tag on this change as well. I intend for this change and the color customization change to ship together in a major point release later this week.

To make sure we're on the same page, does that mean the themes team will be handling deployment of this PR? To set expectations, I'm planning to head off on a vacation soon. Grant is also on a separate team now, so he'll be prioritizing separate work.

pbking:
> To make sure we're on the same page, does that mean the themes team will be handling deployment of this PR? 

Yes, I would like to coordinate the release with a few other things.  Since it's a fairly fundamental change to how Blockbase is managed I think it needs a major release.  I'll merge the change in the next few days.

blockbase/functions.php
jeyip: https://github.com/Automattic/wp-calypso/issues/64303

pbking:
I just noticed this, and I'm not sure why, but with this branch applied the site-title doesn't obey the font size values assigned via theme.json Global Styles:

image

(expected)
image

Rebasing to trunk didn't help.
The theme.json settings for `core/site-title` remain unchanged in the branch.
I am quite confused...

--- [UPDATE] ---

So far I've only been able to trace this happening upon the action in the `/inc/custom-google-fonts.php`:
```
	add_action( 'init', '\Automattic\Jetpack\Fonts\Introspectors\Global_Styles::enqueue_global_styles_fonts' );
}
```

With that bit removed the site-title global styles attributes are again applied.  I can't find any reason why the above logic should prevent those styles (or any others) from being applied but this is as far as I was able to trace the behavior.

Note that this is effecting child themes as well (as I would have expected):

image

Addison-Stavlo:
> I just noticed this, and I'm not sure why, but with this branch applied the site-title doesn't obey the font size values assigned via theme.json Global Styles:

Im trying to repro but am unable. If I have font size set by the block setting, that value is reflected. If i remove that and set a font size in GS -> Blocks -> Site Title -> Typography, that value is reflected.

Do you have more info on how to repro? I tested this by syncing this build to my sandbox and testing on a simple site.

jeyip:
> Im trying to repro but am unable. If I have font size set by the block setting, that value is reflected. If i remove that and set a font size in GS -> Blocks -> Site Title -> Typography, that value is reflected.

> Do you have more info on how to repro? I tested this by syncing this build to my sandbox and testing on a simple site.

After further testing, Addie and I were able to reproduce the issue.

One other bug to note is that saving block level global styles changes doesn't work either. The saved values appear to revert to the original body font ( e.g. global styles < blocks < post title < typography < font family modifications don't persist and revert to the body font )

We did some preliminary investigation, and the culprit has something to do with the newest Gutenberg version (v13.5). We tested this PR alongside Gutenberg v13.3 and the site title renders properly in that instance. The saving bug noted above is also fixed.

After running a `git bisect` on Gutenberg, the problem appears to be [this commit ](https://github.com/WordPress/gutenberg/commit/877668d19f934d39b25ac351751620207f727b18) from [this PR](https://github.com/WordPress/gutenberg/pull/41569
)
> Found an error to be addressed before this can ship.

@pbking hoping to clarify expectations here. Is the themes team going to handle the error and push a fix? This is looking like a rabbit hole, and Team Cylon has shifted focus to work on the help center.

Happy to discuss this further 👍 

cc @Addison-Stavlo @MaggieCabrera

jeyip:
Putting this PR on hold until we settle on a path forward with GDPR concerns

p1655712801370349-slack-C029FM1EH

pbking:
> Is the themes team going to handle the error and push a fix? This is looking like a rabbit hole, and Team Cylon has shifted focus to work on the help center.

@jeyip yes I'm currently pouring all of my time into the Blockbase font management issue, and while I'm particularly focused on the GDPR concerns mentioned I'm also planning to work through the `site-title` issue raised.  Tracking down the change to that commit looks like it will be particularly helpful so thank you very much!

jeyip:
Closing this in favor of https://github.com/Automattic/themes/pull/6123 for now

PR #6011, opened 2022-05-18
Lodestar: Add page password protection for portfolio page

wojtekn:

Lodestar theme has an issue related to the password-protected page: portfolio items are displayed even if the page is protected. It behaves that way because portfolio items are not wrapped with a password check.

#### Testing steps:

1. Log in to WP Admin
3. Ensure that the Lodestar theme from this PR is installed and enabled

- add any content
- ensure that “Template" is “Portfolio Template.”
- set “Visibility” to “Password protected.”

Screen Shot 2022-05-18 at 08 42 13

#### Related issue(s):

### DIFF
--- a/lodestar/components/features/portfolio/content-portfolio-template.php
+++ b/lodestar/components/features/portfolio/content-portfolio-template.php
@@ -7,8 +7,10 @@
 ?>
 
 <?php
-	if ( have_posts() ) :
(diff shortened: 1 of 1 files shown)


p-jackson (approved):
Nice, easy diff this time. Lodestar looks good 👍

PR #6012, opened 2022-05-18
Sketch: Add page password protection for portfolio page

wojtekn:

Sketch theme has an issue related to the password-protected page: portfolio items are displayed even if the page is protected. It behaves that way because portfolio items are not wrapped with a password check.

#### Testing steps:

1. Log in to WP Admin
3. Ensure that the Sketch theme from this PR is installed and enabled
- add any content
- ensure that “Template" is “Portfolio Page Template”
- set “Visibility” to “Password protected”

Screen Shot 2022-05-18 at 09 08 50

#### Related issue(s):

### DIFF
--- a/sketch/portfolio-page.php
+++ b/sketch/portfolio-page.php
@@ -11,36 +11,48 @@ get_header(); ?>
 		<main id="main" class="site-main" role="main">
 
 		<?php if ( sketch_has_featured_posts( 1 ) ) : ?>
-		    <div class="featured-content">
-		        <?php get_template_part( 'content-featured' ); ?>
-		    </div>
+			<div class="featured-content">
+				<?php get_template_part( 'content-featured' ); ?>
+			</div>
 		<?php endif; ?>
 
-		<?php if ( ! get_theme_mod( 'sketch_hide_portfolio_page_content' ) ) : ?>
-			<?php while ( have_posts() ) : the_post(); ?>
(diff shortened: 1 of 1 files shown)


sketch/portfolio-page.php
```diff
@@ -11,36 +11,48 @@
...
+			</div>
 		<?php endif; ?>
 
-		<?php if ( ! get_theme_mod( 'sketch_hide_portfolio_page_content' ) ) : ?>
-			<?php while ( have_posts() ) : the_post(); ?>
+		<?php if ( post_password_required() ) : ?>
 
-				<?php the_title( '<header class="page-header"><h1 class="page-title">', '</h1></header>' ); ?>
+			<?php the_title( '<header class="page-header"><h1 class="page-title">', '</h1></header>' ); ?>
+			<?php the_content(); ?>
```
p-jackson: We should wrap this in `.page-content` so that it matches how the revealed content is rendered.

```suggestion
			<div class="page-content">
				<?php the_content(); ?>
			</div>
```

p-jackson (changes requested):

sketch/portfolio-page.php
wojtekn: I've wrapped it with a similar container as used below.

p-jackson (approved):
Great! Sketch looking good 👍

mikachan (approved):

PR #6013, opened 2022-05-18
Illustratr: Add page password protection for portfolio page

wojtekn:

Illustratr theme has an issue related to the password-protected page: portfolio items are displayed even if the page is protected. It behaves that way because portfolio items are not wrapped with a password check.

#### Testing steps:

1. Log in to WP Admin
3. Ensure that the Illustratr theme from this PR is installed and enabled
* add any content
* ensure that “Template" is “Portfolio Page Template”
* set “Visibility” to “Password protected”

Screen Shot 2022-05-18 at 13 45 45

#### Related issue(s):

### DIFF
--- a/illustratr/page-templates/portfolio-page.php
+++ b/illustratr/page-templates/portfolio-page.php
@@ -11,7 +11,10 @@ get_header(); ?>
 		<main id="main" class="site-main" role="main">
 
 		<?php if ( ! get_theme_mod( 'illustratr_hide_portfolio_page_content' ) ) : ?>
-			<?php while ( have_posts() ) : the_post(); ?>
+			<?php
+			while ( have_posts() ) :
+				the_post();
+				?>
 
 				<?php if ( '' != get_the_post_thumbnail() ) : ?>
 					<div class="entry-thumbnail">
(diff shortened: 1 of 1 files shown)


vykes-mac:
Tested it according to the instructions and it works as expected. @wojtekn I'm trying to determine if step 

> Go to Portfolio -> All Projects and add one or more projects  

is necessary? It seem to work even when I deleted the projects.

wojtekn:
@vykes-mac portfolio items were displayed below the password protection form when the bug was in place. They were displayed regardless of whether the user entered a password or not. On the screenshot under the original bug ticket, we can see that: https://github.com/Automattic/wp-calypso/issues/60135.

The fix ensures that now password protection is checked for both the page content and portfolio items. Thus portfolio projects need to exist in the blog to let us test if they do not appear under the password protection form anymore.

Then when we enter the password, we need to validate if they are displayed correctly again.

p-jackson (approved):
Working as expected for me 👍 I see that the diff is mostly just an indent change due to the content being wrapped in a `post_password_required()` check. I see that, unlike the Orvis fix, this theme also requires special handling to render the page title when. The duplication is unfortunate but probably the best option. Refactoring it so that the title and content code can be shared would be a bigger change with more chance to mess things up. So I like the diff as it is.

@Automattic/theam do you prefer to do a review of your own before themes get deployed? Or are you happy for this to be deployed now.

PR #6019, opened 2022-05-19
Fix issue with duplicate title tags in blockbase

### LINKED ISSUE #5702: Blockbase: Duplicate Title tag on all blockbase & blockbase-premium themes
### Quick summary

Any new site created has duplicate `<title>` tags, regardless of plan. Only happens on child themes of blockbase or blockbase-premium.

### Steps to reproduce

1. Create a new Simple site (Free or Pro Plan)
2. Pick any theme (Free or Premium)
3. Check the page source. You will see two `<title>` tags.

I would expect the `<title>` to be shown once in the page source code - as showing twice could have an impact on SEO.

### What actually happened

The `<title>` is shown twice, which does not happen with themes like Hever or Twenty-TwentyTwo.

### Context

User is worried about SEO impact.
4848257-zen

### Simple, Atomic or both?

Both

### Theme-specific issue?

Any themes that are children of blockbase or blockbase-premium

markbiek:

All Blockbase (free & premium) child themes have two <title> tags rendering on the page.

The extra title tag comes from `wp-includes/block-template.php`. There's a `wp-head` action that gets triggered for themes with `title-tag` support and outputs a `<title>` tag.

See https://github.com/Automattic/themes/issues/5702 for the full
discussion but the gist is that `title-tag` support is redundant for block
themes.

#### Testing

1. Create a test wpcom site and pick any blockbase child theme (like Calvin, for example)
2. Go to the homepage of the site and view the page source in the browser.
3. You should see two <title> tags.

Now apply this branch.

1. Go to your test site and change to another blockbase child theme.
2. Go to the homepage of the site and view the page source in the browser.
3. You should see only one <title> tag.

#### Related issue(s):
Closes #5702

### DIFF
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -17,9 +17,6 @@ if ( ! function_exists( 'blockbase_support' ) ) :
 		// Add support for post thumbnails.
 		add_theme_support( 'post-thumbnails' );
 
-		// Declare that there are no <title> tags and allow WordPress to provide them
-		add_theme_support( 'title-tag' );
-
 		// Experimental support for adding blocks inside nav menus
 		add_theme_support( 'block-nav-menus' );
 


MaggieCabrera (approved):
LGTM

MaggieCabrera:
This needs to be replicated on the premium version of Blockbase @Automattic/serenity

PR #6029, opened 2022-05-23
Antonia - Initial Commit

onuro:
Initial commit of the Antonia, payments block based Blockbase child theme.

Homepage Example and [Draft Demo →](https://antonia29062342.wordpress.com/)

Antonia-Desktop


pbking:
I made one adjustment to the DM Sans font configuration.  The way the weights were defined in the 'google' attribute prevented them from being correctly leveraged (I can't say exactly why).  I copied the configuration from Quadrat which was also using DM-Sans.  Please adjust if I didn't get the intent right.

Otherwise this looks like a good starting point for this theme.

🚢

pbking (approved):

PR #6030, opened 2022-05-24
Meraki: Update footer pattern

alaczek:
* update background color; 
* update footer pattern structure to avoid horizontal scrollbars.


mikachan (approved):
This LGTM 👍

PR #6038, opened 2022-05-25
Updated theme-utils.mjs help text

markbiek:

Running `node theme-utils.mjs` (no arguments) or `node theme-utils.mjs
help` will now show a list of available commands.

Running `node theme-utils.mjs help <command>` will show the specific
help text for that command.

image

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -11,186 +11,135 @@ const isWin = process.platform === 'win32';
 const premiumThemes = ['videomaker', 'videomaker-white'];
 const coreThemes = ['twentyten', 'twentyeleven', 'twentytwelve', 'twentythirteen', 'twentyfourteen', 'twentyfifteen', 'twentysixteen', 'twentyseventeen', 'twentynineteen', 'twentytwenty', 'twentytwentyone', 'twentytwentytwo'];
 
+const commands = {
+	"push-button-deploy": {
+		helpText: `
+* Gets the last deployed hash from the sandbox
+* Version bump all themes that have changes since the last deployment
+* Commit the version bump change to github
+* Clean the sandbox and ensure it is up - to - date
+* Push all changed files(including removal of deleted files) since the last deployment
+* Update the 'last deployed' hash on the sandbox
+* Create a phabricator diff based on the changes since the last deployment.The description including the commit messages since the last deployment.
+* Open the Phabricator Diff in your browser
+* Create a tag in the github repository at this point of change which includes the phabricator link in the description
+* After pausing to allow testing, land and deploy the changes
+		`,
+		run: pushButtonDeploy
+	},
+	"clean-sandbox": {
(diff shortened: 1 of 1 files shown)


theme-utils.mjs
```diff
@@ -34,13 +34,166 @@ const coreThemes = ['twentyten', 'twentyeleven', 'twentytwelve', 'twentythirteen
...
+				additionalArgs = '<theme-slug> <add-changes, true/false>'
+				helpText = 'Use the commit log to build a list of recent changes and add them as a new changelog entry. If add-changes is true, the updated readme.txt will be staged.';
+				break;
+			}
+		case "rebuild-theme-changelog":
+			{
+				additionalArgs = '<theme-slug> <since>';
+				helpText = 'Rebuild the entire change long from the given starting hash.';
+				break;
+			}
```
madhusudhand: This is super helpful.

I have the following thoughts in the view of maintaining all commands from a single place.

Turn the information into an object. (also considering the switch case in line 17-38)
```js
const commands = {
  "push-button-deploy": { helpText: '',  additionalArgs: '', run: () =>  pushButtonDeploy() },
  "deploy-sync-core-theme": { helpText: '',  additionalArgs: '', run: () =>  deploySyncCoreTheme(args?.[1], args?.[2]) },
  ...
}
```

now the switch cases could be replaced by
```
// line 17-38
commands[command].run();

// line 75-190
const {helpText, additionalArgs} = commands[command];
```

Also `showHelp` method can be simplified to

```js
return `Available commands` + Object.keys(commands).join('\n')
```

markbiek: I like that! Much tidier looking!

Thanks for the suggestion, I should be able to get that taken care of today.

theme-utils.mjs
```diff
@@ -8,39 +8,141 @@ const sandboxPublicThemesFolder = '/home/wpdev/public_html/wp-content/themes/pub
...
 const isWin = process.platform === 'win32';
-const premiumThemes = [ 'videomaker', 'videomaker-white' ];
+const premiumThemes = ['videomaker', 'videomaker-white'];
 const coreThemes = ['twentyten', 'twentyeleven', 'twentytwelve', 'twentythirteen', 'twentyfourteen', 'twentyfifteen', 'twentysixteen', 'twentyseventeen', 'twentynineteen', 'twentytwenty', 'twentytwentyone', 'twentytwentytwo'];
 
+const commands = {
+	"push-button-deploy": {
+		helpText: `
+* Gets the last deployed hash from the sandbox
+* Version bump all themes have have changes since the last deployment
```
sixhours: "Version bump all themes *that* have changes..." I think?

markbiek: Good catch! That was in the original comment on the handler function and I didn't see it 😝

valterlorran (approved):
👍 Looks good.

madhusudhand (approved):
LGTM!

PR #6040, opened 2022-05-26
fix site-title alignment with multiple social icons

### LINKED ISSUE #5969: Pendant header menu pattern expected behaviour
I should have foreseen this, and it may not be a huge issue given that you can change everything in the editor. But when a user adds more social links to the header, the Site Title will move to the center of the row block, which is the natural behaviour. 

However, if the intention is for the Site Title to continue in the middle, that doesn't happen. I think that as a user I'd expect that if I added more items the Site Title wouldn't move.

Captura de ecrã 2022-05-03, às 10 37 08

Do you think this is a problem and if so what would be a good solution?

madhusudhand:

**Before**:
image

**After**:
image

#### Testing instructions:

1. Clear any previously done customisations to header template
2. Save and refresh the editor so that it picks up updated template.
3. Add two or more social icons and observe that site title is center aligned.

#### Related issue(s):

Fixes: #5969


pbking (approved):
Looks good to me.  Thanks for finding another way to do this without custom CSS.

PR #6041, opened 2022-05-26
Blockbase: Implement the Button elements API

MaggieCabrera:

This PR adds the button element to theme.json so it can be accessed by plugins. It shouldn't affect how the buttons look before or after this PR, on Blockbase or any of its children. I will create a followup PR that removes the CSS that will no longer be needed after this is merged, but I'll keep that separated since it will require that the theme has the GB installed for it to work. We can wait to merge that change after it lands in core.

Before | After
--- | ---
Screenshot 2022-05-26 at 12 38 39 | Screenshot 2022-05-26 at 12 38 44

### DIFF
--- a/blockbase/theme.json
+++ b/blockbase/theme.json
@@ -553,7 +553,8 @@
 					"fontFamily": "var(--wp--preset--font-family--body-font)",
 					"fontSize": "var(--wp--custom--button--typography--font-size)",
 					"fontWeight": "var(--wp--custom--button--typography--font-weight)",
-					"lineHeight": "var(--wp--custom--button--typography--line-height)"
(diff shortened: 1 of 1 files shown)


pbking:
This hasn't been moved out of 'in progress' but I think it should be safe to merge into Blockbase (whenever, though I tagged it for inclusion in the 3.0 release).

mikachan (approved):
This LGTM! I like the idea of this going live with Blockbase 3.0.

MaggieCabrera:
We should make a quick test on dotcom but this should be mergeable now

pbking:
Closing: Work merged in #6167

PR #6042, opened 2022-05-26
Blockbase: remove unused Button CSS after implementing elements API

MaggieCabrera:

This PR is built on top of https://github.com/Automattic/themes/pull/6041

We are removing the CSS that is now covered by the elements API implementation in #6041. The buttons on the search block, file block and button block should look the same as before.

I decided to leave the mixins as they are, in case they are used in other child themes, while removing the includes from Blockbase itself. I'm keeping all the hover stuff so we are not breaking anyone's themes either until that is supported by theme.json too.


pbking:
This button API doesn't allow us to handle hover states right?  So at a minimum we'll need to keep the hover state mixins to keep feature parity. 

Also border styles also still seems to still be lacking as something we can define so is that going to have to stay as well?

I think we can be more aggressive that what we've got going on in the PR at the moment though.  Typography, color and padding seems to be handled pretty well so I think all of that mixin CSS can go too.

Lastly, I had a little question as to how we might be able to change how we're targeting some non-button stuff.  We use some classes like `.wp-block-search__button` to target search buttons explicitly (and we'll need to continue to for said hover styles at least).  But do you think we'll be able to use the new `wp-element-button` in its stead?

MaggieCabrera:
> 
> I think we can be more aggressive that what we've got going on in the PR at the moment though. Typography, color and padding seems to be handled pretty well so I think all of that mixin CSS can go too.
> 

I'm not removing that from the mixin in case the mixin is being used by other elements that don't leverage the elements API, like a 3rd party plugin or block. It's a long stretch to assume, but it's a possibility and keeping the mixin intact doesn't produce extra CSS on Blockbase, only on the child that wants to make use of the mixin.

MaggieCabrera:
> Lastly, I had a little question as to how we might be able to change how we're targeting some non-button stuff. We use some classes like `.wp-block-search__button` to target search buttons explicitly (and we'll need to continue to for said hover styles at least). But do you think we'll be able to use the new `wp-element-button` in its stead?

The new class will not be present on blocks that are already present on sites, so I'd rather keep the current markup for the CSS that we have

pbking:
Tested and this works as expected (when Gutenberg is installed).  Since this requires a version of Gutenberg higher than what is currently in CORE I'm going to tag this item for Blockbase 3.1 (a release for WP 6.1).

madhusudhand:
closing this because it points to an old branch. created pr to trunk here #6714

PR #6044, opened 2022-05-26
#5875: Illustratr - Remove gallery from top of gallery posts

bobmatyas:

- Removed duplicate gallery from top of posts
- I removed the PHP code calling this rather than using CSS, but if this is used somewhere else that I'm missing, I'm happy to adjust the approach.
- It also looks like the PHP linter fixed some spacing issues. 

#### Individual Post
Screenshot on 2022-05-26 at 15-05-07

#### Category Page
Screenshot on 2022-05-26 at 15-06-55

#### Individual Post

Screenshot on 2022-05-26 at 15-09-27

#### Category Page

Screenshot on 2022-05-26 at 15-09-55

## Related issue(s):

https://github.com/Automattic/themes/issues/5875

### DIFF
--- a/illustratr/content-gallery.php
+++ b/illustratr/content-gallery.php
@@ -11,26 +11,20 @@ if ( isset( $GLOBALS['content_width'] ) ) {
 
 <article id="post-<?php the_ID(); ?>" <?php post_class(); ?>>
 
-	<?php if ( get_post_gallery() && ! post_password_required() ) : ?>
-		<div class="entry-gallery">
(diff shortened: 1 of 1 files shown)


pbking:
From what I can tell the removed code only duplicates a gallery, which I don't believe is very expected.  This tests out and looks like the right fix.

🚢

pbking:
From what I can tell the removed code only duplicates a gallery, which I don't believe is very expected.  This tests out and looks like the right fix.

🚢

pbking (approved):

PR #6045, opened 2022-05-27
Cultivate: Add base theme.

alaczek:
This PR adds Cultivate theme, generated from Block Canvas. The base theme has fonts and colors already updated, as well as header and footer template parts. Once this is merged I will create individual PRs for different templates, patterns, and fixes.

Here's a sneak preview of the theme for reference:
community_garden_preview


cultivate/parts/header.html
```diff
@@ -0,0 +1,34 @@
+<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"bottom":"0px","top":"0px","right":"0px","left":"0px"},"margin":{"top":"20px","bottom":"0px"}},"border":{"width":"2px"}},"borderColor":"primary","className":"site-header site-header-linear","layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
+<div class="wp-block-group alignfull site-header site-header-linear has-border-color has-primary-border-color" style="border-width:2px;margin-top:20px;margin-bottom:0px;padding-top:0px;padding-right:0px;padding-bottom:0px;padding-left:0px">
+
+	<!-- wp:group {"style":{"spacing":{"blockGap":"20px","padding":{"left":"2vw"}}},"className":"site-words-stack-small","layout":{"type":"flex"}} -->
+	<div class="wp-block-group site-words-stack-small" style="padding-left:2vw">
```
mikachan: I don't think the `site-words-stack-small` class is needed here.

cultivate/parts/header.html
```diff
@@ -0,0 +1,34 @@
...
+
+		<!-- wp:site-tagline {"fontSize":"normal"} /-->
+			
+	</div>
+	<!-- /wp:group -->
+
+	<!-- wp:group {"layout":{"type":"flex","flexWrap":"wrap","justifyContent":"right"}} -->
+	<div class="wp-block-group">
+
+		<!-- wp:navigation {"__unstableLocation":"primary","className":"blockbase-responsive-navigation-linear","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}},"fontSize":"normal"} /-->
```
mikachan: I don't think the `blockbase-responsive-navigation-linear` class is needed either.

cultivate/theme.json
```diff
@@ -0,0 +1,339 @@
...
+					"google": "family=Chivo:ital,wght@0,400;0,700;1,400;1,700",
+					"fontSlug": "chivo"
+				},
+				{
+					"fontFamily": "Rubik, sans-serif",
+					"slug": "heading-font",
+					"name": "Headings (Rubik)",
+					"google": "family=Rubik:ital,wght@0,400;0,700;1,400;1,700",
+					"fontSlug": "rubik"
+				}
```
mikachan: These fonts are both rendering as the browser default for sans-serif rather than the custom font:

image

This is probably because this style of font setting only works in Blockbase. Outside of Blockbase, you should set custom fonts [like this](https://github.com/Automattic/themes/blob/trunk/curator/theme.json#L93):

```
{
	"fontFamily": "'Work Sans', sans-serif",
	"name": "Headings (System Font)",
	"slug": "heading-font",
	"fontFace": [
		{
			"fontDisplay": "block",
			"fontFamily": "'Work Sans'",
			"fontWeight": "900",
			"fontStyle": "normal",
			"fontStretch": "normal",
			"src": [
				"file:./assets/fonts/WorkSans.ttf"
			]
		}
	]
}
```

You'll also need to bundle the font files [with the theme](https://github.com/Automattic/themes/tree/trunk/curator/assets/fonts).

mikachan (commented):
This is looking good! Just left some comments on a few things I noticed.

cultivate/parts/header.html
alaczek: Removed!

cultivate/parts/header.html
alaczek: Removed!

cultivate/theme.json
alaczek: Alright, I bundled the fonts with the theme and updated theme.json.

Isn't it a little weird though that we need to include fonts this way even though they are available on WordPress.com? AFAIK this theme will not be released on .org.

mikachan: > Isn't it a little weird though that we need to include fonts this way even though they are available on WordPress.com?

Oh yeah you're right! Maybe the fonts need to be done slightly differently for WordPress.com.. I guess they're handled by Jetpack?

I've just tested setting Chivo and Rubik as font families in theme.json on .com, without bundling the font files themselves, and it seems to work fine. It looks like you should be able to use something like the following:

```
	"fontFamilies": [
		{
			"fontFamily": "\"Chivo\", sans-serif",
			"slug": "chivo"
		},
		{
			"fontFamily": "\"Rubik\", sans-serif",
			"slug": "rubik"
		}
	]
```

And then there's no reason to include the font files either. Sorry! 🙈 This is much simpler.

alaczek: Alrighty, I unbundled the fonts and reverted theme.json file.

alaczek:
I just realized that this theme is supposed to be premium 😅 I still appreciate the review, but in the end, I won't be merging this theme, but removing from GH, and continuing development in private.

PR #6051, opened 2022-06-01
Appleton: Fix case on screenshot.png file.

alaczek:
WordPress.com is not picking up the screenshot in the showcase, so I'm changing it to lowercase.


roo2 (approved):
LGTM

PR #6056, opened 2022-06-02
#5788: Shawburn: Pullquote Block(full-Width): centered text isn't working on small screens

### LINKED ISSUE #5788: Shawburn: Pullquote Block(full-Width): centered text isn't working on small screens
### Quick summary

If we set a Pullquote block to be full-width with centered text, the content won't be centered on small screens.

### Steps to reproduce

1. Create a new post or page.
2. Add the Pullquote to the page and set it to be full-width. Also, make sure the text is set to be centered.
3. Preview the page. When reducing the window size to mobile sizes, the quote won't be centered.
4. You can also confirm that checking the page/post on a real device.

I'd expected the text to be correctly centered.

### What actually happened

The pullquote text isn't centered. There is a wrong margin causing this issue.

### Context

Customer report: 4887538-zen

### Simple, Atomic or both?

Simple, Atomic

### Theme-specific issue?

This issue seems to be specific to the `Shawburn` theme.

_No response_

### Reproducibility

Consistent

### Severity

One

### Available workarounds?

Yes, easy to implement

### Workaround details

This can be fixed with CSS code. Some variations of the code below can fix that:

```
.wp-block-pullquote.alignfull.has-text-align-center blockquote{
  margin: auto;
}
```

Gustavo-Hilario:

Centering parent blockquote HTML element when the pullquote block is set to be full width.

**Before**

Markup on 2022-06-01 at 19:10:39

**After** - centered paragraph

full width and text center

**After** - paragraph with a different alignment

full width only

#### Related issue(s):

Fixes #5788

### DIFF
--- a/libretto/style.css
+++ b/libretto/style.css
@@ -2834,6 +2834,7 @@ object {
     -webkit-transition: all 0.25s linear;
             transition: all 0.25s linear;
     width: 400px;
+    z-index: 999;
   }
 
   #site-navigation .search-form label::before {
--- a/libretto/style.css
+++ b/libretto/style.css
@@ -2834,7 +2834,6 @@ object {
     -webkit-transition: all 0.25s linear;
             transition: all 0.25s linear;
     width: 400px;
-    z-index: 999;
   }
 
   #site-navigation .search-form label::before {
--- a/shawburn/sass/_extra-child-theme.scss
+++ b/shawburn/sass/_extra-child-theme.scss
@@ -424,3 +424,8 @@ table,
 		margin-right: calc( .1 * #{map-deep-get($config-button, "padding", "horizontal")} );
 	}
 }
+
+/* center pullquote block when set to be full width */
+.wp-block-pullquote.alignfull blockquote {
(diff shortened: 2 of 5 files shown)


libretto/style.css
```diff
@@ -2834,6 +2834,7 @@ object {
     -webkit-transition: all 0.25s linear;
             transition: all 0.25s linear;
     width: 400px;
+    z-index: 999;
```
pbking: It looks like the commit that changed this wasn't intended to be in this PR.

pbking:
I removed the rouge commit and the fix works as expected.

🚢

pbking (approved):

pbking:
Oh...

I also realized (after approving) that this theme leverages SASS to construct the CSS.  So the change needed to be made there (and recompiled) rather than just making the CSS change.  Otherwise the next time the SASS is compiled the change would be lost.

I've made that change, but @Gustavo-Hilario if you could review that to know what to expect for other varia-child theme changes.

PR #6058, opened 2022-06-03
Curator: update page template

### LINKED ISSUE #5934: Curator: Templates: Page
image

madhusudhand:

image

image

**Note**: It has dependency on single page implementation for comments pattern

#### Related issue(s):

Fixes: #5934

### DIFF
--- a/curator/templates/page.html
+++ b/curator/templates/page.html
@@ -1,16 +1,13 @@
 
 <!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
-<!-- wp:cover {"overlayColor":"foreground","minHeight":525,"minHeightUnit":"px","style":{"spacing":{"padding":{"top":"0px","right":"0px","bottom":"0px","left":"0px"},"margin":{"top":"0px","right":"0px","bottom":"0px","left":"0px"}}}} -->
-<div class="wp-block-cover" style="padding-top:0px;padding-right:0px;padding-bottom:0px;padding-left:0px;margin-top:0px;margin-right:0px;margin-bottom:0px;margin-left:0px;min-height:525px">
-<span aria-hidden="true" class="wp-block-cover__background has-foreground-background-color has-background-dim-100 has-background-dim"></span>
-<div class="wp-block-cover__inner-container">
-<!-- wp:cover {"useFeaturedImage":true,"dimRatio":40,"overlayColor":"foreground","minHeight":525,"minHeightUnit":"px","style":{"spacing":{"padding":{"top":"120px","right":"clamp(20px, 4vw, 90px)","bottom":"120px","left":"clamp(20px, 4vw, 90px)"}}}} -->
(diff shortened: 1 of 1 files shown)


curator/templates/page.html
```diff
@@ -1,27 +1,18 @@
...
+<div class="wp-block-cover" style="padding-top:0px;padding-right:0px;padding-bottom:0px;padding-left:0px;margin-top:0px;margin-right:0px;margin-bottom:0px;margin-left:0px;min-height:525px">
+<span aria-hidden="true" class="wp-block-cover__background has-foreground-background-color has-background-dim-100 has-background-dim"></span>
+<div class="wp-block-cover__inner-container">
+<!-- wp:cover {"useFeaturedImage":true,"dimRatio":40,"overlayColor":"foreground","minHeight":525,"minHeightUnit":"px","style":{"spacing":{"padding":{"top":"120px","right":"20px","bottom":"120px","left":"20px"}}}} -->
+<div class="wp-block-cover" style="padding-top:120px;padding-right:20px;padding-bottom:120px;padding-left:20px;min-height:525px"><span aria-hidden="true" class="wp-block-cover__background has-foreground-background-color has-background-dim-40 has-background-dim"></span><div class="wp-block-cover__inner-container"><!-- wp:post-title {"textAlign":"center","level":1,"style":{"spacing":{"margin":{"top":"0px","right":"0px","bottom":"0px","left":"0px"}}},"textColor":"background"} /--></div></div>
```
jffng: I ended up using the [clamp value](https://github.com/Automattic/themes/blob/trunk/curator/theme.json#L63) for the left/right padding here in the single template, what do you think about using it here too?

jffng (commented):
I think @beafialho needs to give this one a review because I'm not sure about the height of the cover area, as well as the treatment for when there is no featured image.

> Note: It has dependency on single page implementation for comments pattern

I don't think the page template has comments?

curator/templates/page.html
madhusudhand: updated the values to clamp.

beafialho:
I was taking a look at this.

If I remember correctly it wasn't possible to remove the black cover block, right @jffng? This would be the ideal:

| Page with Featured Image | Page without Featured Image |
| ----- | ----- |
| Single Page   |   About-1  |

For a solution, since this is the page template, it might be better to stick with the simpler layout:

| Page with Featured Image | Page without Featured Image |
| ----- | ----- |
|  About |   About-1  |

jffng:
> If I remember correctly it wasn't possible to remove the black cover block, right @jffng?

That's right, because our solution was to include the post featured image block inside a cover block with black background. That way if the post had no featured image, the text would still be legible.

> For a solution, since this is the page template, it might be better to stick with the simpler layout:

👍

madhusudhand:
Update the template to suggested simpler designs.

| Page with Featured Image | Page without Featured Image |
| ----- | ----- |
| curator-page-template | curator-page-template-2 |

jffng (approved):
Reviewed this with @beafialho and it's ✅

PR #6061, opened 2022-06-03
curator: home page template

### LINKED ISSUE #6060: Curator: Templates: Home page
image

madhusudhand:

**Mobile:**

#### Related issue(s):

Fixes: #6060

### DIFF
--- a/curator/patterns/hero-post.php
+++ b/curator/patterns/hero-post.php
@@ -13,7 +13,7 @@
 <!-- wp:post-title {"textAlign":"center","level":1,"style":{"typography":{"lineHeight":"0.8"},"spacing":{"margin":{"top":"3rem"}}},"className":"hero-post-title"} /-->
 
 <!-- wp:group {"align":"wide","layout":{"wideSize":"900px"}} -->
-<div class="wp-block-group alignwide"><!-- wp:post-excerpt {"textAlign":"center","moreText":"<?php echo esc_html__( 'Read more', 'curator' ); ?>","style":{"spacing":{"padding":{"top":"1rem"}},"typography":{"lineHeight":"1.2"}},"className":"hero-post-excerpt"} /--></div>
+<div class="wp-block-group alignwide"><!-- wp:post-excerpt {"textAlign":"center","moreText":"<?php echo esc_html__( 'Read more', 'curator' ); ?>","style":{"spacing":{"padding":{"top":"1rem"}},"typography":{"lineHeight":"1.2","fontStyle":"normal","fontWeight":"300","letterSpacing":"-0.02em"}},"className":"hero-post-excerpt"} /--></div>
 <!-- /wp:group -->
 
 <!-- wp:separator {"style":{"spacing":{"margin":{"top":"4rem"}}},"className":"alignfull"} -->
--- a/curator/patterns/hero-post.php
+++ b/curator/patterns/hero-post.php
@@ -1,0 +1,22 @@
+<?php
+/**
+ * Title: Hero Post
+ * Slug: curator/hero-post
+ * Categories: text
+ */
+
+?>
+
+<!-- wp:group {"align":"full","style":{"spacing":{"margin":{"top":"5rem","bottom":"2rem"}}}} -->
+<div class="wp-block-group alignfull" style="margin-top:5rem;margin-bottom:2rem"><!-- wp:post-terms {"term":"category","textAlign":"center"} /-->
+
+<!-- wp:post-title {"textAlign":"center","level":1,"style":{"typography":{"lineHeight":"0.8"},"spacing":{"margin":{"top":"3rem"}}},"className":"hero-post-title"} /-->
+
+<!-- wp:group {"align":"wide","layout":{"wideSize":"900px"}} -->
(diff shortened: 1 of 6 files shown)


madhusudhand:
Q1: Can you suggest on, setting width of the separator line to 100%? It occupies parent width, because it is placed inside.

Q2: For Recent photos, if the post doesn't has a featured image, it leaves blank space. Should we consider this as a problem?

curator/style.css
```diff
@@ -171,6 +171,27 @@ a:not(
...
+ */
+.wp-block-post-template .wp-block-post {
+	margin-top: 0.5rem;
+}
+
+
+/*
+ * Post Excerpt
+ */
+.wp-block-post-excerpt .wp-block-post-excerpt__more-text a {
```
jffng, line 217: I wish it were easier to just make this look like the theme's button...

curator/patterns/hero-post.php
```diff
@@ -0,0 +1,20 @@
+<?php
+/**
+ * Title: Hero Post
```
jffng, line 3: If the latest post has no text (and thus no excerpt), the spacing becomes very tight:
 
Screen Shot 2022-06-14 at 11 34 10 AM

When there is an excerpt to display, I think the font size needs to be bigger (32px from the comps), and the width somehow limited. Also the heading size is 170px in the comps but maybe that is too big because post titles can be long?  

Current | Intended
--- | ---
Screen Shot 2022-06-14 at 11 37 33 AM | Screen Shot 2022-06-14 at 11 39 17 AM

jffng (commented):
Thanks @madhusudhand !

jffng:
> Q1: Can you suggest on, setting width of the separator line to 100%? It occupies parent width, because it is placed inside.

It think this is "breaking the rules" of the current alignment concept in Gutenberg, but if you add "alignfull" manually to the separator, it should work with our alignment CSS:

Screen Shot 2022-06-14 at 11 53 10 AM

The downside is you can't do this from the UI...

curator/patterns/hero-post.php
madhusudhand: Without excerpt content

image

With excerpt content

image

Post title is given font size of 8rem (128px). 170px appears to be too big and it breaks even the medium length post titles into multiple lines).
Excerpt is: 2rem (32px) and it takes a max width of 900px.

comps suggests them to be variable sizes. However, clamp values in the template doesn't work with GB enabled. 
But they work when GB disabled 😵‍💫 .
I will have to check if there is a related issue. Otherwise I will create one.

curator/style.css
```diff
@@ -183,6 +183,28 @@ a:not(
 	text-decoration: underline;
 }
 
+/*
+ * Post template block
+ */
+.wp-block-post-template .wp-block-post {
+	margin-top: 0.5rem;
```
MaggieCabrera: could we do this with a spacer block maybe?

beafialho:
From looking at the template and your screenshots, I think it's now quite close to the designs.

> Q2: For Recent photos, if the post doesn't has a featured image, it leaves blank space. Should we consider this as a problem?

I don't think so. That section was particularly thought to have images, regardless of whether it's a post or not. If the user wants, they could turn that into a gallery block or columns with photos instead. What do you think?

One question: is the screenshot below using the font size specified [here](https://www.figma.com/file/nW4lWMpFyCCjZcNOTZivQ5/Curator?node-id=318%3A256) on Figma?

172129359-1fe264ce-63f0-4e4a-af71-5d36a35bdd49

madhusudhand:
@beafialho 

> If the user wants, they could turn that into a gallery block or columns with photos instead. What do you think?

Right. we should leave that decision to user 👍 

> One question: is the screenshot below using the font size specified [here](nW4lWMpFyCCjZcNOTZivQ5-fi-318%3A256) on Figma?

I have used clamp in the template for the font sizes as follows

```
<!-- wp:post-title {"textAlign":"center","level":1,"style":{"typography":{"fontSize":"clamp(5.875rem, 10vw, 8rem)","lineHeight":"0.8"},"spacing":{"margin":{"top":"3rem"}}}} /-->
```
It's weird, the above template works with WP 6.0 without GB enabled. 
But the same template fails when GB is enabled.
Removed font-size from template and created custom classes to address this. 

Created: https://github.com/WordPress/gutenberg/issues/41859

curator/patterns/hero-post.php
madhusudhand: created: https://github.com/WordPress/gutenberg/issues/41859

jffng:
Reviewed this with @beafialho and made a few tweaks: 

- Adjusted the typography settings for the excerpt by decreasing the font weight and letter spacing slightly.
- Reduced the gallery to only show no more than two images  
- Reduced the lower boundary of the font size for the hero post title

@beafialho one change we spoke about but I did not implement — reducing the lower boundary (mobile size) of the Recent Posts titles. In the design comps, these are set to be font size H1 in desktop, but they map to the H2 font size on mobile. Would you like us to set a custom font size here ("clamp" value), or can we pick one heading level and rely on the default typography sizes for it? The downside to custom is that it makes it hard to replicate, and kind of makes the default sizes not as useful.

beafialho:
I see @jffng, feel free to use the H1 default size for Recent Posts on mobile too 👍

jffng (approved):

PR #6064, opened 2022-06-04
#6047: Seedlet: Fix no post message

### LINKED ISSUE #6047: Seedlet: Missing posts text incorrectly says posts page is homepage
### Quick summary

Seedlet shows a message on the posts page when there are no posts:

>Your site is set to show the the most recent posts on your homepage - but you don't have any Posts published.

This message displays on the posts page even when the page is an interior page, so this message is incorrect in those cases.

The message should be more general such as:

> Your site is set to show the the most recent posts on this page - but you don't have any Posts published.

### Steps to reproduce

1. Set the posts page as an interior page on a site with no posts
2. View error message

The message shouldn't say the site is set to displays posts on the homepage when it's not.

### What actually happened

Screen Shot on 2022-05-16 at 09:35:13

### Context

I ran into this creating a DIFM-L site for a customer.

It's possible this previous change was related:

https://github.com/Automattic/themes/pull/3875

### Simple, Atomic or both?

Simple, Atomic

### Theme-specific issue?

Seedlet

I was using Firefox 99.0.1, but this shouldn't be browser related.

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

_No response_

### Workaround details

This is only visible when there are no posts, so it's confusing information since it's incorrect but not a huge deal since it's gone after adding posts.

bobmatyas:
### Before

Annotation on 2022-06-04 at 15-55-32

### After

Annotation on 2022-06-04 at 15-56-52

#### Related issue(s):

Fixes https://github.com/Automattic/themes/issues/6047

### DIFF
--- a/seedlet/template-parts/content/content-none.php
+++ b/seedlet/template-parts/content/content-none.php
@@ -21,7 +21,7 @@
 			?>
 
 			<p>
-			<?php _e( "Your site is set to show the the most recent posts on your homepage - but you don't have any Posts published.", 'seedlet' ); ?></p>
+			<?php _e( "Your site is set to show the the most recent posts on your this page - but you don't have any Posts published.", 'seedlet' ); ?></p>
 			<p>
 				<a href="<?php echo esc_url( admin_url( 'edit.php' ) ); ?>" class="button">
 					<?php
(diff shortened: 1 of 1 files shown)


madhusudhand (approved):
LGTM!

PR #6070, opened 2022-06-06
Curator: update footer styles

### LINKED ISSUE #6069: Curator: Styles: Footer
**Large:**

image

**Small:**

image

madhusudhand:

**Large screens:**

image

**Small screens:**

image

**Known issues:**

There is a slight alignment issue with social icons with the right edge.
This is because of the padding of the social icon. We would need to add custom css to override this. Should we remove padding?

Also, icons come by default. we would need to hide them explicitly using custom css. Should we hide them?

image

#### Related issue(s):

Fixes: #6069

### DIFF
--- a/curator/parts/footer.html
+++ b/curator/parts/footer.html
@@ -1,11 +1,1 @@
-<!-- wp:group {"layout":{"inherit":true}} -->
-<div class="wp-block-group">
-	<!-- wp:group {"style":{"spacing":{"padding":{"top":"80px","bottom":"30px"}}}} -->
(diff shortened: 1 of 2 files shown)


beafialho:
> Also, icons come by default. we would need to hide them explicitly using custom css. Should we hide them?

I'd perhaps not use the Social Links block. I'd add a row with custom links.

To your point, it would be great though, if the editor allowed us to hide the social icons in that block, just showing the link name.

madhusudhand:
Updated to links

image

jffng (approved):
I escaped the links, this LGTM.

PR #6071, opened 2022-06-06
Curator: index template and styles

### LINKED ISSUE #5921: Curator: Templates: Index
image

madhusudhand:

image

image

**Small screens:**

image

image

Note: reduced the font-size to 18px in small screens to keep them in the same line. Changing the labels differently based on device isn't possible yet.

#### Related issue(s):

Fixes: #5921


madhusudhand:
image

**Arrow icon on hover**: I tried adding another block and to toggle its visibility on hover, parent needs a custom class and custom css. Is there any other approach to achieve this?

jffng:
Thank you @madhusudhand ! I made some changes to this in a pair session with @beafialho:

- Reworked the pagination layout, and tweaked the style of the active number to match the design. Also removed the pattern, since this was only working for me if the pagination lived _inside_ the query. 
- Reworked the meta layout since the single line is preferred. It makes me wonder though, should the archive template meta also match this layout, or is it okay that they're different now? 

Screen Shot 2022-06-23 at 3 55 21 PM

jffng (approved):

beafialho:
Thank you both.

> It makes me wonder though, should the archive template meta also match this layout, or is it okay that they're different now?

It's ok that they're different 👍

PR #6072, opened 2022-06-06
Block Canvas : Eliminated the naming of font families from 'body' and 'header' 

pbking:
Eliminated the naming of font families from 'body' and 'header' to a slug that matches the font.

Since the tooling that allows the customization of fonts included in themes leverage slugs that match the font, rather than the purpose this base theme has been changed to follow that intent.

Additionally, since this theme used the same font for both body and heading there is only one font-family defined now and the 'heading' customizations that were added to some blocks has been removed.

### DIFF
--- a/block-canvas/theme.json
+++ b/block-canvas/theme.json
@@ -94,15 +94,10 @@
             "fontFamilies": [
                 {
                     "fontFamily": "-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif",
-                    "name": "Body (System Font)",
-                    "slug": "body-font"
-                },
-                {
-                    "fontFamily": "-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif",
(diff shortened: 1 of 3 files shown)


MaggieCabrera:
Will this affect inserted patterns that already define the fonts using the old slug? I'm guessing there's not going to be many of those...

pbking:
I'm not sure that there are any.  Considering how theme-specific font families are, and how theme-agnostic patterns SHOULD be I don't suppose a pattern ever SHOULD be assigning a font family.  So if there are patterns this effect I would say the error is in the pattern.

This is the same problem as colors of course, using descriptive vs semantic naming.  Moving away from semantic font slugs, especially since there doesn't look to be a future of standardization, means to me that fonts can't/shouldn't be defined in patterns/template for the same reason that "emerald green" shouldn't be leveraged as colors in them.

mikachan (approved):
This looks good to me, and I think the reasoning makes sense.

It looks like this branch includes some tidying up in Blockbase too, but I think it's good to merge as well.

pbking:
> It looks like this branch includes some tidying up in Blockbase too

I'm glad you pointed that out @mikachan thank you.  They might be good changes and might come in elsewhere but I didn't intend for them to come in on this change.  I've added a commit returning them to the trunk state.

PR #6073, opened 2022-06-06
Moved templates from old folder location to new

### LINKED ISSUE #6062: Blockbase: Update name of template directories
Templates are expected to be in /templates (rather than /block-templates) and template parts should be in /parts (rather than /block-template-parts).

Not completely necessary to work as expected (both directories are checked) but the more current and expected location should be used.

pbking:
This changes the templates (and template parts) folder location from the old location to the new.  It also updates the "pattern registration" that registers Blockbase's template parts as patterns.

Intended for Blockbase 3


MaggieCabrera (approved):
The changes LGTM but I wonder if we should divide this into two PRs to avoid breakage during deploy. I've had that happened to me when renaming an included file, don't know if this would be the case for dotcom, you may get alerts even if for just a few minutes.

pbking:
Closing: Work merged in #6167

PR #6075, opened 2022-06-07
#5097: Maywood: Fix image alignment

### LINKED ISSUE #5097: Multiple themes: Image blocks are centered by default, not left aligned
### Quick summary

I would assume that Image blocks are supposed to be left aligned (not floated left however) when you don't set specific alignment. Testing multiple Varia and Blockbase themes they are centered by default. There is no way to set them to regular left alignment without the "Align Left" setting, which results in text wrapping.

### Steps to reproduce

* Activate Maywood
* Add an Image block, don't set a specific alignment

Published page should look like the editor with the image to the left without text wrapping. 

Screen Shot 2021-11-19 at 08 11 37

The caption probably shouldn't be centered but that's not the point of this issue.

### What actually happened

It's center aligned.

Screen Shot on 2021-11-19 at 08:12:43

The issue with setting it to Align Left is that the text wraps:

Screen Shot on 2021-11-19 at 08:13:38

On Maywood this CSS causes the issue:

Screen Shot on 2021-11-19 at 08:16:02

### Context

Building a DIFM-L site.

pdh1Xd-78-p2

### Operating System

macOS

### Browser

Mozilla Firefox

### Simple, Atomic or both?

Simple, Atomic

### Theme-specific issue?

Every Varia and Blockbase theme I tested. Maywood is the example above.

### Other notes

_No response_

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

Yes, easy to implement

### Workaround details

You could set the image to Align Left and add a Spacer block right under the image to break the float (so the text doesn't wrap the image).

bobmatyas:
Fixes image alignment in Maywood to bring frontend into parity with editor options

#### Before
----

Editor:

Screenshot on 2022-06-07 at 09-25-38

Frontend:

Screenshot on 2022-06-07 at 09-25-59

#### After
----

Editor:

Screenshot on 2022-06-07 at 09-25-38

Frontend:

Screenshot on 2022-06-07 at 09-26-37

### Related issue(s):

Fixes https://github.com/Automattic/themes/issues/5097

### DIFF
--- a/maywood/style.css
+++ b/maywood/style.css
@@ -1700,10 +1700,6 @@ h6, .h6 {
 	line-height: 1.125;
 }
 
-.wp-block-image {
-	text-align: center;
-}
-
 .wp-block-image figcaption {
 	color: #686868;
 	font-size: 0.69444rem;
--- a/maywood/style.css
+++ b/maywood/style.css
@@ -1700,6 +1700,10 @@ h6, .h6 {
 	line-height: 1.125;
 }
 
+.wp-block-image {
+	text-align: center;
+}
+
 .wp-block-image figcaption {
 	color: #686868;
 	font-size: 0.69444rem;
--- a/varia/sass/blocks/image/_style.scss
+++ b/varia/sass/blocks/image/_style.scss
@@ -1,5 +1,4 @@
 .wp-block-image {
-	text-align: center;
 
 	figcaption {
 		color: #{map-deep-get($config-global, "color", "foreground", "light")};


pbking (changes requested):
Maywood leverages SASS.  Thus this change would require that the source file be changed and then recompile, but that source is actually in [varia](https://github.com/Automattic/themes/blob/trunk/varia/sass/blocks/image/_style.scss#L2).

Thus deleting this CSS would effect ALL Varia themes.  Which perhaps is desirable based on the issue described.  But changing the CSS here would just get overwritten the next time the SASS was recompiled.

bobmatyas:
Thanks for the feedback here.

I made a change in Varia to reflect this (built and tested) and rebuilt the child themes (`npm run children`). Based on this comment it seems like the change should happen in Varia: https://github.com/Automattic/themes/issues/5097#issuecomment-985333064

I tested on Maywood and it fixed the issue. I also checked in Hever and Stratford and had no adverse effects. 

## Varia

### Before
Screenshot on 2022-06-10 at 14-55-52

### After
Screenshot on 2022-06-10 at 15-13-35

## Maywood

### Before

Screenshot on 2022-06-10 at 14-55-15


### After
Screenshot on 2022-06-10 at 14-59-51

pbking (changes requested):
This looks like the right change to the SASS file to make the change you want.  However it will still need to be compiled and the result of that compilation committed.

That's going to make for quite a few changed files, but that's how Varia and family works.

Run `npm run build` in the varia folder to build the CSS for Varia itself.  Then (also in the varia folder) run `./rebuild-child-themes.sh` (or `npm run children` which just calls that bash script).

The result with be a changed `style.css`, `style-rtl.css` and possibly a changed `package-lock.json` file for each varia theme which also needs to be committed to the PR.

Of course this means that the resulting behavior will be changed in all of these themes.  Which could be argued is correct, but it's also a change and has the potential to change the layout of a lot of people's sites.  Something to carefully consider when making the fix. (One person's 'fix' is another person's 'breakage'.)

bobmatyas:
I can build the child themes and commit, but maybe this needs more discussion by the Theme team as to whether or not this should be fixed? 

It seems like it is indeed not working properly (prior to the fix), but I am somewhat unsure whether or not it now makes sense to fix knowing that it will change so many themes. Theoretically, users shouldn't be relying on the auto-aligning since the editor was working properly and if they wanted center aligning, they would have had to manually do it.

PR #6076, opened 2022-06-07
Update theme url

onuro:
Fix the wrong theme Theme URI from the GH url to https://www.wordpress.com/theme/antonia

### DIFF
--- a/antonia/style.css
+++ b/antonia/style.css
@@ -1,6 +1,6 @@
 /*
 Theme Name: Antonia
-Theme URI: https://github.com/automattic/themes/tree/trunk/antonia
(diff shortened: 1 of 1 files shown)


antonia/style.css
```diff
@@ -1,6 +1,6 @@
 /*
 Theme Name: Antonia
-Theme URI: https://github.com/automattic/themes/tree/trunk/antonia
+Theme URI: https://www.wordpress.com/theme/antonia
```
alaczek: Let's drop the www from the address.

alaczek (changes requested):

alaczek (approved):

PR #6078, opened 2022-06-08
Removed the 'unstableSocialLinks' from themes that shouldn't have it …

pbking:
…(it's a Blockbase exclusive)

This attribute really isn't harming anything but it's only leveraged by Blockbase and its children so there's no need for it to be on these template parts.


madhusudhand (approved):
LGTM! :ship:

PR #6088, opened 2022-06-14
spearhead: fix site navigation position in small screens

### LINKED ISSUE #3880: Spearhead: Menu doesn't respond as expected to small viewport
image

Assumed the menu would be in the top right of the header. 

Note: the theme might be leaking CSS in the toolbar as well (note my avatar in the screenshot) or there could be something else going on.

madhusudhand:

This fixes the issues with menu position of spearhead theme on small screens.
Following are the four different scenarios.

**1. Without Admin bar, marketing bar**

without-both

**2. Admin bar only**

admin-bar

**3. Marketing bar only**

marketing-bar

**4. Both Admin and marketing bars**

both-bars

#### Related issue(s):

Fixes: #3880


pbking:
I see that the "jump up" behavior is present before this change, and that all of the situations replicate that.  But I wonder if that was intended or was a side-effect of something else.

Could/should we consider eliminating that "jumping behavior" while we're working on this menu issue?

scruffian:
I don't remember all the details here, but please make sure to test this when the site also has a tagline. Thanks!

madhusudhand:
@scruffian tagline is always hidden in small screens. It works fine with or without a tagline.

@pbking the "jump up" effect could be intentional. It is fine to leave the effect for the following reasons.

1. It maintains equal spacing around the title.
2. Also, it makes sense to jump up to save space when there is a site logo.

sh-with-logo

pbking:
> Also, it makes sense to jump up to save space when there is a site logo.

Ah, that's right.  I should have tested with a logo. :P 

> tagline is always hidden in small screens

👍 

This looks good to me.

pbking (approved):

PR #6096, opened 2022-06-17
#2044: Pique: Fix button block width on mobile

### LINKED ISSUE #2044: Pique: Text in Buttons block breaks unexpectedly on mobile
## Steps to replicate

1. Create a new page
2. Add a Buttons block with multiple words in the label (e.g. "I'm Interested")
3. Publish page and view in mobile-view

## Result

- Button label text wraps very narrowly and unexpectedly, breaking individual words even though there appears to be plenty of space

Screenshot on 2020-05-19 at 12-14-14

## Expected

- Words in button label would not be broken

* User report: #19442417-hc
* Follow-up ticket: #2985280-zen

bobmatyas:

Removed code that was setting a width of 75% for buttons on mobile.

#### Before
Screenshot on 2022-06-17 at 11-07-11

#### After
Screenshot on 2022-06-17 at 11-12-00

### Related issue(s):

Fixes https://github.com/Automattic/themes/issues/2044

### DIFF
--- a/pique/assets/css/blocks.css
+++ b/pique/assets/css/blocks.css
@@ -440,12 +440,6 @@ body.home .wp-block-table.alignwide .wp-block-table.alignfull {
 	color: #83b6cc;
 }
 
-@media (max-width: 767px) {
-	.wp-block-button .wp-block-button__link {
-		width: 75%;
-	}
-}
-
 /* Separator */
 
 .wp-block-separator {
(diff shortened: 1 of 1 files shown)


madhusudhand (changes requested):
Changes work for the button block. But it doesn't seem to fix the problem entirely.
For other button like blocks such as excerpt, a different style is applied which again limiting the width to 75%.

You may need to update or remove the css at `660` in `style.css`
image

image

bobmatyas:
I revisited this and ended up approaching this differently and set the width to `100%` on mobile. This will bring it in line with the other buttons as shown in this screenshot:

Screenshot on 2022-06-23 at 09-39-52

madhusudhand (approved):
LGTM!

PR #6109, opened 2022-06-22
Block canvas: comment block

MaggieCabrera:
This replicates the changes done on Blockbase in https://github.com/Automattic/themes/issues/5833 to Block Canvas as well so our new themes use the new block too

Edit: I rebased to pick up the Block canvas fixes and this is what it looks like

Before | After 
--- | ---
Screenshot 2022-06-23 at 16-36-58 Hello world! – maggie | Screenshot 2022-06-23 at 16-37-15 Hello world! – maggie

### DIFF
--- a/block-canvas/patterns/comments.php
+++ b/block-canvas/patterns/comments.php
@@ -1,0 +1,61 @@
+<?php
+/**
+ * Title: Comments
+ * slug: blockbase/comments
+ * inserter: no
+ */
+
+?>
+<!-- wp:comments-query-loop -->
+<div class="wp-block-comments-query-loop">
+	<!-- wp:comments-title /-->
+
+	<!-- wp:comment-template -->
+	<!-- wp:columns -->
+	<div class="wp-block-columns">
+		<!-- wp:column {"width":"40px"} -->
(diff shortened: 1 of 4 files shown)


MaggieCabrera:
I haven't added screenshots because I found some bugs in Block Canvas that I'll be fixing in a follow up PR

madhusudhand (commented):
Font size slightly increased for the `comment author`, `comment date` and `reply link` from `15.75px` => `18px` in comments block, where as other links such as `post category` and `post date` are at `14px`. 
We will have to bring both to same size.

Everything else LGTM!

MaggieCabrera:
> Font size slightly increased for the `comment author`, `comment date` and `reply link` from `15.75px` => `18px` in comments block, where as other links such as `post category` and `post date` are at `14px`. We will have to bring both to same size.
> 
> Everything else LGTM!

Great catch! I was using the same variables as Blockbase and those weren't working. Bad copy paste, bad!

MaggieCabrera:
This just needs a green check, should be good to go

mikachan (approved):
Looking good to me, and the latest changes fix @madhusudhand's comment ✅

PR #5700, opened 2022-03-17
Add and configure fonts for use in Pendant. 

pbking:
#### Changes proposed in this Pull Request:

Add necessary font files to /assets/fonts
Add entries for fonts in theme.json
Configure headings and body element attributes in theme.json
Ensure user-selectable font sizes are appropriate
Review for mobile/desktop size scaling

Note that most of this task doesn't seem to be do-able in the FSE.  Management of fonts is expected to be possible with 3rd party plugins or perhaps JetPack. 

#### Details:

In order to achieve the design the typefaces **Jost-Light** and **Literata-ExtraLight** have been used.

The original design notes a size difference between desktop/mobile for header font sizes.  This is achieved with `clamp()` values.

The original design notes that the typeface for Heading 4 (and presumeably 5) change from Literata to Jost from Desktop to Mobile.  This isn't possible with current tooling without CSS.  I've split the difference and made H5 and H6 uppercase Jost.  Please let me know if this is acceptable.

#### Related issue(s):

Fixes: #5676

#### Screenshots:

Headings (desktop):

image

Headings (mobile):

image

Body (desktop):

image

Body (mobile):

image

Used for evaluation:

```
<!-- wp:heading {"level":1} -->
<h1>Heading 1 <strong>Bold</strong> <em>Italic</em> <strong><em>BoldItalic</em></strong></h1>
<!-- /wp:heading -->

<!-- wp:heading -->
<h2>Heading 2 <strong>Bold</strong> <em>Italic</em> <strong><em>BoldItalic</em></strong></h2>
<!-- /wp:heading -->

<!-- wp:heading {"level":3} -->
<h3>Heading 3 <strong>Bold</strong> <em>Italic</em> <strong><em>BoldItalic</em></strong></h3>
<!-- /wp:heading -->

<!-- wp:heading {"level":4} -->
<h4>Heading 4 <strong>Bold</strong> <em>Italic</em> <strong><em>BoldItalic</em></strong></h4>
<!-- /wp:heading -->

<!-- wp:heading {"level":5} -->
<h5>Heading 5 <strong>Bold</strong> <em>Italic</em> <strong><em>BoldItalic</em></strong></h5>
<!-- /wp:heading -->

<!-- wp:heading {"level":6} -->
<h6>Heading 6 <strong>Bold</strong> <em>Italic</em> <strong><em>BoldItalic</em></strong></h6>
<!-- /wp:heading -->

<!-- wp:separator {"className":"is-style-default"} -->
[...]

PR #6110, opened 2022-06-22
Block canvas regressions

MaggieCabrera:
This PR fixes a few regressions and bugs I found while working on https://github.com/Automattic/themes/pull/6109

I found that after https://github.com/Automattic/themes/pull/5700 got merged, some of the pendant options like the colors were showing on Block Canvas, plus the font sizes were very big. Apparently at some point we removed the "normal" font size and that is ok, since it's not a default font size in core, but the previous blocks assigned to the normal font size were replaced by medium, which is too big for the original design (as much as we are not making this a 1:1 clone of Blockbase, it was looking very strange. 

What I did:

- Replace the Pendant color with the Blockbase ones
- Replaced the old instances of the normal custom font size with the actual value of the font size.
- Fixed the post meta that was looking left aligned

Before | After
--- | ---
Screenshot 2022-06-22 at 10-54-15 Hello world! – maggie | Screenshot 2022-06-22 at 10-54-01 Hello world! – maggie

### DIFF
--- a/block-canvas/parts/post-meta.html
+++ b/block-canvas/parts/post-meta.html
@@ -1,8 +1,12 @@
-<!-- wp:group {"layout":{"type":"flex"}} -->
+<!-- wp:group {"layout":{"inherit":true}} -->
 <div class="wp-block-group">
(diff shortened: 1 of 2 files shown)


jffng (approved):
LGTM, thanks for fixing these.

PR #6122, opened 2022-06-23
Themes: Remove site title text decoration from theme.json

scruffian:
This removes some unnecessary CSS from themes and moves it to theme.json- we can define the text decoration for site-title in JSON now.

To test, check that the following themes don't have text decoration on their site titles:
- Remote
- Curator
- Stewart
- Livro


mikachan (commented):
This is working well for me!

Should Curator also have the same theme.json changes as the other themes do (with the additional elements.link.typography.textDecoration setting)?

scruffian:
> Should Curator also have the same theme.json changes as the other themes do (with the additional elements.link.typography.textDecoration setting)?

Yes, good catch, added it.

mikachan (approved):
🚢

PR #6123, opened 2022-06-23
Fix/migrate blockbase font self hosted

pbking:
This branch continues on from the custom Jetpack Font Management implementation previously created in Blockbase and moves that to instead leverage fonts that are provided locally by the theme.

The goal is that this works both WITH and WITHOUT the Gutenberg plugin activated.

When Gutenberg is activated it provides a custom Font Provider that handles compiling and enqueuing the CSS.
When Gutenberg is NOT activated that action is done manually.

All of the fonts definitions are kept in Blockbase's theme.json file.  This is so font selection in the FSE can still work without Gutenberg.

The process attempts to migrate any font settings made via the customizer, or from an activated child theme that hasn't transitioned, to storing the font opinion as block styles.  Additionally the Blockbase font definitions are copied to the user space for child themes that haven't transitioned.

Only (and all) those fonts that are USED (assigned via theme.json files for via Global Styles) should have the font-face styles enqueued.  All font faces should be enqueued in the FSE.

Testing:
* This should be evaluated both WITH and WITHOUT the Gutenberg Plugin
* This will only work with WordPress 6+
* Test Blockbase instances that have font customizations made prior to this change (Gutenberg only, since customizations were only available in the customizer with Gutenberg activated)
* Test child themes with no customizations.
* Test child themes with customizations made prior to this change (Gutenberg Only)

Before:
image

After:
image


jeyip (commented):
If I understand things correctly, you've effectively replaced any references to the core Webfonts API and appended font face declarations to an inline style sheet instead in `custom-fonts.php` 

Looking at the font introspection logic and considering what you've done with it, I think your approach looks reasonable. 👍 

I'll test this out tomorrow

jeyip:
Apologies for not being able to test this out today! I'm blocking off time to make sure testing gets done tomorrow.

jeyip:
## Testing
### Requirements

**With Gutenberg**

- [x] No fonts are downloaded on initial load of frontend with Blockbase enabled
- [x] No font face values are enqueued in the base html
- [x] All of the fonts provided by Blockbase are available in the Font family dropdown.
- [x] All of the elements change in the site editor
- [x] Fonts selected in site editor are displayed in the page editor and the site frontend
- [x] Only the selected fonts are downloaded from the google fonts api 
- [x] Only the selected fonts have font face declarations added to inline styles
- [x] Post title global styles typography customizations behave as expected in the editor + frontend 

❌  I still see the bug you pointed out in the previous PR, whereby the site title font family change ( that's the result of the `Global Styles > Blocks > Heading > Typography` change ) isn't reflected on the frontend of the site.

| Site Editor      | Frontend |
| ----- | ----- |
| Screen Shot 2022-06-29 at 4 55 17 PM   | Screen Shot 2022-06-29 at 4 55 26 PM        |

**Without Gutenberg**
- [x] No fonts are downloaded on initial load of frontend with Blockbase enabled
- [x] No font face values are enqueued in the base html
- [x] All of the fonts provided by Blockbase are available in the Font family dropdown.
- [x] All of the elements change in the site editor
- [x] Fonts selected in site editor are displayed in the page editor and the site frontend
- [x] Only the selected fonts are downloaded from the google fonts api 
- [x] Only the selected fonts have font face declarations added to inline styles
- [x] Post title global styles typography customizations behave as expected in the editor + frontend

pbking:
Ok, I think this is finally ready to be put to bed.

@jeyip thank you for your testing so far!  If you are able to give this another whirl next week I would really appreciate it!  

The "big" site-title you noted above turned out to be caused by fetching Global Styles values too early, before all blocks had been registered.  This cached what blocks were allowed to be styled at the time the Global Styles were fetched and thus styles provided by the theme were ignored.  It was quite a journey to figure that out!  Thanks to @scruffian for pointing out the likely cause.

It was perhaps more work than it was worth to get this working sans Gutenberg; not just "not breaking" but for the user to actually be able to customize with all the fonts and selectively enqueue them.   But it felt like good work. :P

pbking:
I just realized that going along with this change it would be good to convert all of the child themes in this repository (and the premium themes when this goes there too).  The migration logic means that this doesn't HAVE to be done, but there's less to do when the blockbase font collection is allowed to bubble up, and a better experience when "resetting Global Styles to default"

jeyip:
> @jeyip thank you for your testing so far! If you are able to give this another whirl next week I would really appreciate it!

Hey my mistake. This totally fell off my radar. Taking a look later today

jeyip:
## Testing
### Requirements
**Gutenberg**
- [x] All of the fonts provided by Blockbase are available in the Font family dropdown.
- [x] Google font families are available in the site editor global styles typography dropdowns
- [x] Only the selected fonts are downloaded from the google fonts api
- [x] Only the selected fonts have font face declarations added to inline styles
- [ ] Post title global styles typography customizations behave as expected in the editor + frontend

I think I'm seeing another bug though.

On a completely fresh WordPress instance:
- with Gutenberg installed
- with the Blockbase theme activated
- with no customizer typography previously selected from the customizer
- with a global typography selected
- when I select a global styles "Heading" block font family, it's only respected in the editor -- not the frontend.

Screen Shot 2022-07-09 at 1 25 06 AM

Screen Shot 2022-07-09 at 1 25 15 AM

The global styles "Post Title" block font family, however, does work as expected.

pbking:
Hrm, I wasn't able to reproduce this bug. The expected typeface is showing for me in both editor and view.

image

I tried both is this branch and in the /blockbase-3 branch which puts all of the bits for this release together.  I wonder what might be different about your environment to make that happen.

pbking:
I'm pushing a change that includes refactoring all of the child themes in this repository to use the new font definition so that if there are no customizations the migration will not be necessary.  And also taken the opportunity to evaluate all of the child themes to ensure that the fonts look the same before and after this change.

jeyip:
> Hrm, I wasn't able to reproduce this bug. The expected typeface is showing for me in both editor and view.

I'll test this again tomorrow. I wonder if there's something specific to my environment 🤔 . 

To set expectations, I've been super busy onboarding a new hire and with new team priorities ( we just switched gears and are being redirected to a new project ). Would the themes team be able to help code review this PR? I won't be able to get back to this as often as I'd like, and I wouldn't want the momentum on these code changes to stall.

pbking:
Thanks for the heads up @jeyip and your willingness to help amongst your busy schedule!  We've had some eyes on this but it's so big it's hard to give a ✅ .  We've done some testing with this branch merged with the other Blockbase 3 issues all bundled together and we're looking at that pretty hard.  Hopefully anything will shake out there if nothing else.

jeyip:
> I'll test this again tomorrow. 

Couldn't get to testing today, but it's the first thing I'll do Monday morning

pbking:
Closing: Work merged in #6167

PR #6126, opened 2022-06-23
Curator: fix single template comments

jffng:

Fixes a bug where the comments were missing a closing group block markup, so the footer was not the correct width:

Before | After
--- | ---
Screen Shot 2022-06-23 at 12 26 50 PM | Screen Shot 2022-06-23 at 12 26 36 PM


madhusudhand (approved):
🚢

PR #6140, opened 2022-06-24
#5894: Blockbase: Fix comment subscription checkboxes

### LINKED ISSUE #5894: Dorna, Hari, etc: Comment form labels overlap on mobile
### Quick summary

The labels in the Leave a Reply area overlap on mobile devices.
Screen Shot 2022-04-17 at 8 14 58 PM

### Steps to reproduce

1. Start with a site that has posts.
2. Enable Dorna.
3. Use a mobile phone to find a post, then scroll down and look at the comments area. 

I want to be able to see the different fields of the reply box, like this does if I uncheck the offending CSS:

Screen Shot 2022-04-17 at 8 14 26 PM
 

### What actually happened

I get a jumble of text.
Screen Shot 2022-04-17 at 8 14 20 PM

### Context

Customer report here:
https://wordpress.com/forums/topic/overlapping-text-problem-with-leave-a-reply-on-apple-mobile-phone/

### Simple, Atomic or both?

_No response_

### Theme-specific issue?

Dorna, Hari, likely related themes

I've tested on my phone (iPhone SE) as well as with Chrome's inspect element.  The user who reported this also uses an iPhone.

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

No and the platform is unusable

### Workaround details

This CSS snippet would work, but as we're discouraged from handing out snippet fixes, we'll need to say there is not a workaround.

```
.wp-block-post-comments form>* {
    grid-area: unset;
}
```

bobmatyas:

Adjust the grid for post comments on Blockbase to fix overlap issue that happens on Blockbase and children. I tested change on Blockbase, Dorna, and Winkel.

#### Before 
Screenshot on 2022-06-24 at 11-23-18

#### After
Screenshot on 2022-06-24 at 12-47-22

### Related issue(s):

Fixes #5894

### DIFF
--- a/blockbase/sass/blocks/_post-comments.scss
+++ b/blockbase/sass/blocks/_post-comments.scss
@@ -1,5 +1,7 @@
 .wp-block-post-comments {
-	label, .comment-meta {
+
+	label,
+	.comment-meta {
 		font-size: var(--wp--custom--form--label--typography--font-size);
 	}
 
@@ -10,6 +12,7 @@
 		a {
 			display: inline-block;
 			text-decoration: underline;
+
 			&:hover {
 				text-decoration: none;
 			}
@@ -21,7 +24,8 @@
 		column-gap: 1em;
 		grid-template-rows: auto;
 		grid-template-areas:
-			"misc misc"
+			"loggedin loggedin"
+			"misc1 misc2"
 			"notes notes"
 			"author author"
 			"email url"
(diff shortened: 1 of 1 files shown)


madhusudhand (changes requested):
1. run `npm run build` include the generated css changes.

2. I guess it is best to keep the checkboxes in the bottom after the button.
Following is how other themes block themes show the UI

image

MaggieCabrera:
I think I'd rather go with https://github.com/Automattic/themes/pull/6109 as a solution instead, since it removes all the extra CSS from Blockbase and relies on the site editor instead.

bobmatyas:
I agree it seems like #6109 is a better way to go.

vcanales:
Closing, as the alternative PR (#6109) was merged.

PR #6146, opened 2022-06-28
Curator: add mobile menu styling

### LINKED ISSUE #6095: Curator: mobile menu styling
## Description

The mobile menu needs to be styled:

Current | Target
--- | ---
Screen Shot 2022-06-16 at 11 44 47 AM | Screen Shot 2022-06-16 at 11 45 35 AM

mikachan:
This PR adds styling for the mobile navigation in Curator. I worked off the styles in Figma but let me know if anything doesn't look right:

image

#### Related issue(s):
Closes #6095

### DIFF
--- a/curator/style.css
+++ b/curator/style.css
@@ -199,6 +199,14 @@ a:not(
 }
 
 /*
+ * Needed until https://github.com/WordPress/gutenberg/issues/39142 is fixed.
+ */
+ .wp-block-navigation__responsive-container.is-menu-open ul {
+	font-size: var(--wp--preset--font-size--huge);
+	font-weight: 300;
+}
+
+/*
  * Hero post
  * The following style are required because variable fontSize doen't work in templates
  * Feature request: https://github.com/WordPress/gutenberg/issues/41859
--- a/curator/style.css
+++ b/curator/style.css
@@ -202,7 +202,7 @@ a:not(
  * Needed until https://github.com/WordPress/gutenberg/issues/39142 is fixed.
  */
  .wp-block-navigation__responsive-container.is-menu-open ul {
-	font-size: var(--wp--preset--font-size--huge);
+	font-size: 2.75rem;
 	font-weight: 300;
 }
 
--- a/curator/theme.json
+++ b/curator/theme.json
@@ -187,11 +187,6 @@
 					"name": "Extra Large",
 					"size": "2rem",
 					"slug": "x-large"
-				},
-				{
-					"name": "Huge",
-					"size": "2.75rem",
-					"slug": "huge"
 				}
 			],
 			"lineHeight": true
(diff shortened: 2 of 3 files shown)


curator/style.css
```diff
@@ -198,6 +198,14 @@ a:not(
 	text-underline-offset: 0.25rem;
 }
 
+/*
+ * Needed until https://github.com/WordPress/gutenberg/issues/39142 is fixed.
+ */
+ .wp-block-navigation__responsive-container.is-menu-open ul {
+	font-size: var(--wp--preset--font-size--huge);
+	font-weight: 300;
```
jffng, line 206: The design suggests the items are also vertically centered, but I don't think we should add CSS for that here.

curator/theme.json
```diff
@@ -187,6 +187,11 @@
 					"name": "Extra Large",
 					"size": "2rem",
 					"slug": "x-large"
+				},
+				{
+					"name": "Huge",
+					"size": "2.75rem",
```
jffng: I _think_ we decided limit these font sizes to the sizes that are available in Gutenberg by default. That way switching themes is least likely to "break" a user's decision to map of preset font size to some text content. 

cc @pbking as I always gut check this question with you, and go back and forth every time 😅

jffng (commented):
Thanks for taking this on! I left two comments, but in general the mobile menu is looking good to me.

curator/theme.json
mikachan: Yeah I thought I remembered something like this, but I couldn't remember what we decided. Sorry for bringing it up again, now you've mentioned it I'm pretty sure it was to stick with the Gutenberg defaults! I'll fix this up.

curator/style.css
mikachan: Ah OK.

It would only be one line of CSS, but I know any additional CSS isn't great...

```
.wp-block-navigation__responsive-container.is-menu-open {
	justify-content: center;
}
```

This could be an option in GB, and it probably fits into the other conversations around adding styling controls to the mobile navigation (e.g. https://github.com/WordPress/gutenberg/issues/39142).

pbking: > It would only be one line of CSS ...

That would only work for menus that fit completely in the window.  Otherwise you can't scroll to get to what's "above" halfway.

There are ways to do it (I **think** we achieved it with Videomaker eventually) but it's not the simplest of solutions.

pbking (approved):
This looks good to me and seems to match the design.

🚢

curator/style.css
mikachan: > That would only work for menus that fit completely in the window. Otherwise you can't scroll to get to what's "above" halfway.

Ahh ok, that's not a great solution then!

If it becomes a problem we can add the CSS then, but it sounds like it's best to leave this for now.

Issue #6108, opened 2022-06-22
Link interactivity (e.g. `:hover`) styles


MaggieCabrera:
In order to test https://github.com/WordPress/gutenberg/pull/41786 and preparing for when that one gets merged we should implement link hover (and probably focus?) styles on our themes. I think we should start with:

- [x] Blockbase https://github.com/Automattic/themes/pull/6156
- [ ] Blank canvas
- [x] Block Canvas https://github.com/Automattic/themes/pull/6486
- [x] Curator https://github.com/Automattic/themes/pull/6150
- [ ] Livro
- [ ] Skatepark
- [ ] Stewart
- [x] Pendant https://github.com/Automattic/themes/pull/6179

And follow up to clean up CSS from any other themes that are using it.

There's a PR for [Archeo](https://github.com/Automattic/themes/pull/6101) already open.

getdave:
Hope you don't mind I updated the title and desc to include that you might also consider implementing `:focus` just for a11y reasons 🙇

madhusudhand:
In themes like curator, fill button and outline buttons are differently styled in `:hover` and other states. Current implementation takes care of `:hover` for both buttons together. Css styles are still required for outline button to style differently.

|  | Normal state | Hover state |
| --- | --- | --- |
| Fill button | <img width="156" alt="image" src="https://user-images.githubusercontent.com/1935113/176621481-35a79afe-2136-4d65-80bf-72c582b9cae7.png"> | <img width="165" alt="image" src="https://user-images.githubusercontent.com/1935113/176621366-5ffa690f-c1e9-4020-bd57-403f05b27cdb.png"> |
| Outline button | <img width="234" alt="image" src="https://user-images.githubusercontent.com/1935113/176623453-641dc9c4-3b76-4816-b6f4-ac449ceff0bb.png"> | <img width="234" alt="image" src="https://user-images.githubusercontent.com/1935113/176623548-51c1a2ab-9e1b-4404-aac3-fcc5d8379722.png"> |

jeffikus:
@madhusudhand @mikachan looks like only Blank Canvas is remaining on this list, any others we've missed?

mikachan:
Looking through our other themes, it looks like we may also need to update Livro, Skatepark and Stewart. I'll add them to the above list, they shouldn't take long to update.

matiasbenedetto:
[...]

PR #6150, opened 2022-06-30
Vivre: move link pseudo states to theme.json

madhusudhand:

Moved css to theme.json for `:hover`, `:focus` and `:active` states. There are few style left in css because the style properties aren't supported in theme.json. ex: `text-underline-offset`

#### Related issue(s):

#6108

### DIFF
--- a/curator/style.css
+++ b/curator/style.css
@@ -161,7 +161,6 @@ a:not(
 }
 
 .wp-block-post-title a:hover {
-	text-decoration-thickness: 0.25rem;
 	text-underline-offset: 0.25rem;
 }
 
--- a/curator/style.css
+++ b/curator/style.css
@@ -41,6 +41,7 @@ body {
 	background: var(--wp--preset--color--foreground);
 	border-color: var(--wp--preset--color--foreground);
 	color: var(--wp--preset--color--background);
+	text-decoration: none;
 }
 .wp-block-button__link.wp-block-button__link:active {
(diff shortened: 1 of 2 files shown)


pbking:
This is interesting and gave me a little pause.  In this instance we are trading a little CSS for a lot of theme.json configuration.  I don't think there's another way to do it, and I don't think it's wrong, but it's interesting to see how much less verbose CSS can be.

We did seem to lose something in translation with the post title links:

Before:
image

After:
image

And the read more link:

Before:
image

After:
image

If there are other differences they have escaped me.

madhusudhand:
Fixed both issues. Seems like they lost in the transition. 
Even though line thickness is set in css, with single value set to text-decoration, thickness is lost.
changing `textDecoration: underline` to `textDecoration: underline 0.25rem` works.

jffng (approved):
Thanks @madhusudhand this is looking good and very cool to see the pseudo states API in action.

In addition to rebasing, I made one change so the links match the design: https://github.com/Automattic/themes/pull/6150/commits/2b787d7c107d6ba3d9f51211532059fff269269b

Based on the figma, it looks like text links are underlined by default and not underlined on hover:

Screen Shot 2022-07-11 at 1 08 19 PM

Let me know if the way I implemented it makes sense.

pbking:
Previous comment deleted because I have no idea what I'm doing today.

This looks great. 👍

PR #6153, opened 2022-07-01
Blockbase children: update comments block

MaggieCabrera:

This is a follow up on https://github.com/Automattic/themes/pull/6080 and it's built on top of it. It should be merged right after #6080 to avoid much disruption when the CSS is removed. The loss of that CSS will not result on a broken site, it will simply show the default styling of the old Comment block, so I think it's not terrible for other users of the theme that may take longer to adopt this change if they are already overriding the templates that have the block on it.

@Automattic/serenity heads up on this change and Blockbase's too, for any premium themes that may need it.

Screenshot from Quadrat black:

Screenshot 2022-07-01 at 17 11 14

### DIFF
--- a/arbutus/block-templates/single.html
+++ b/arbutus/block-templates/single.html
@@ -18,7 +18,7 @@
 <div style="height:60px" aria-hidden="true" class="wp-block-spacer"></div>
 <!-- /wp:spacer -->
 
-<!-- wp:pattern {"slug":"blockbase/comments"} --></div>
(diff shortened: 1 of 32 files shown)


mikachan (approved):
Thanks so much for working on this! This all looks good to me following the changes from https://github.com/Automattic/themes/pull/6080.

Just quoting what you said above for anyone who merges this:

> It should be merged right after https://github.com/Automattic/themes/pull/6080 to avoid much disruption when the CSS is removed.

pbking:
Closing: Work merged in #6167

PR #6156, opened 2022-07-05
Blockbase: move link pseudo states to theme.json

madhusudhand:

Moved the CSS changes for link pseudo states `:hover`, `:active` and `:focus` to theme.json

#### Related issue(s):

#6108

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -239,31 +239,11 @@ See: https://github.com/Automattic/wp-calypso/issues/58958#issuecomment-10152637
  */
 a,
 .wp-block-navigation a {
-  cursor: pointer;
   text-underline-offset: 0.15em;
-  text-decoration: none;
-}
-a:hover, a:focus,
-.wp-block-navigation a:hover,
-.wp-block-navigation a:focus {
-  -webkit-text-decoration-line: underline;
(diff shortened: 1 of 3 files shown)


blockbase/sass/blocks/_post-comments.scss
```diff
@@ -1,5 +1,7 @@
 .wp-block-post-comments {
-	label, .comment-meta {
+
+	label,
```
pbking: I don't think there are any changes in this file other than formatting.  Can we remove this change to avoid confusion?

pbking:
This looks good and works as expected.   Tagged for inclusion in Blockbase 3.1 (aligning with WP 6.1) as the feature currently requires the Gutenberg plugin.

blockbase/sass/blocks/_post-comments.scss
pbking: Oh!  Nevermind.  I found the change.  The lots'a formatting changes made it hard to find.  Nevermind this comment!

pbking:
I noticed that Blockbase didn't by default show the underline in navigation elements prior to this change so I've taken that bit out of the configuration.  Perhaps it SHOULD but it's a change (and effects a lot of themes!).

That was the only thing that seems different and this should be fine to ship now.

PR #6159, opened 2022-07-05
Archeo: move CSS styles to theme.json

mikachan:
This PR moves some of the CSS from Archeo into the theme.json file.

I've moved styles for the following:
- search and file buttons
- post comments button
- site footer gap

~I've disabled `wp-block-styles` as these styles were overriding the search and file button theme.json styles, and I'm not sure we need this specifically for Archeo. However, I can't remember for sure, so we maybe need to keep this for now, which means probably keeping the search and file button CSS as well for the time being.~


MaggieCabrera:
Is this solving a regression on the block? My only concern with this is mainly missing styles like the [padding on blocks with a background](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/group/theme.scss) for existing sites. If it were a theme yet to be launched we could argue that the user can include the padding themselves, but on existing sites it could cause unexpected changes on existing content, so we'd have to weigh in if we are ok with that.

mikachan:
Ah yeah good point, I think the padding on blocks with a background is a good reason to keep it. This wasn't solving any recent bugs - I was just trying to remove as much CSS as possible! I'll re-enable the `wp-block-styles`.

MaggieCabrera:
> Ah yeah good point, I think the padding on blocks with a background is a good reason to keep it. This wasn't solving any recent bugs - I was just trying to remove as much CSS as possible! I'll re-enable the `wp-block-styles`.

I think it's a good initiative but maybe it's best suited for Curator rather than Archeo because of that

pbking:
Are there styles other than the background padding to account for?  I agree that since we started with the background color padding it should stay, but could we just add that in as custom CSS rather than include the entire `wp-block-styles` styles?

archeo/theme.json
```diff
@@ -166,11 +166,6 @@
 	},
 	"styles": {
 		"blocks": {
-			"core/button": {
-				"typography": {
-					"textTransform": "uppercase"
-				}
```
pbking, line 172: Why did you pull this out?  Was the UPPERCASE transform not in the original design?  It changes existing button styles.

pbking:
I guess this was already known but I just realized while testing this that any of the styles expressed in `styles>elements>button` are ignored for any content created prior to that being available (which is all the content I ever test with...).

I suppose that's not really a comment on this change, just a comment in general and I happened to have this item open when I discovered it. :P

archeo/theme.json
```diff
@@ -256,6 +257,29 @@
...
+				},
+				"color": {
+					"background": "var(--wp--preset--color--foreground)",
+					"text": "var(--wp--preset--color--background)"
+				},
+				"typography": {
+					"fontFamily": "var(--wp--preset--font-family--chivo)",
+					"fontSize": "var(--wp--preset--typography--font-size--normal)",
+					"textTransform": "uppercase"
+				},
```
pbking: This doesn't seem to be getting picked up.  Is textTransform something that can be assigned in this way?

image

mikachan: I think this was a bug with the elements API, and this has since been fixed. I've just had a play with moving it between `core/button` and the button element, and the button element seems to work consistently now 🤞

mikachan:
This should be ready for another review 🙇

archeo/theme.json
pbking: Yup, works now. :)

pbking:
This is working as expected now.  Just has some conflicts to resolve.

mikachan:
> This is working as expected now. Just has some conflicts to resolve.

Thanks @pbking! These should be resolved now.

pbking:
This looks good to go.
Plus I don't believe that it impacts how this is rendered without Gutenberg activated.  The button's are ALREADY a different style, so making this change doesn't effect them negatively.  I see no reason to wait for 6.1 to ship this change. 

Before (Without Gutenberg):
image

After (Without Gutenberg):
image

pbking (approved):

Issue #6142, opened 2022-06-27
Vivre: add support for Button elements


jffng:
As demonstrated in #6139, can we use the button element API on Curator to get rid of some CSS?

Specifically the search + file block button styles.

beafialho:
I'm not sure if I should open a new issue or if this is related, but the Search and File buttons aren't looking as they should:

- The black outline should be removed
- The hover state shouldn't have an underline in the text

https://user-images.githubusercontent.com/65220155/177313639-d5523e74-084f-4945-b2c9-c1c97e98b05d.mov

https://user-images.githubusercontent.com/65220155/177313629-89c49720-3483-4fbc-9357-6ef4b1a28c8a.mov

PR #6160, opened 2022-07-05
Curator: revise search template

### LINKED ISSUE #6157: Curator: Search template should match the Figma mockups
This is what it looks like now:
Captura de ecrã 2022-07-05, às 12 00 07

This is what it should look like:
Captura de ecrã 2022-07-05, às 12 08 11

jffng:

This PR makes some changes to the search template in Curator, to match the Figma design closer

Before | After
--- | ---
Screen Shot 2022-07-05 at 10 39 31 AM | Screen Shot 2022-07-05 at 10 39 18 AM

@beafialho in the Figma, the template suggests the search button is black by default, but in the button states section it is burgundy. Can you confirm which is correct? 

Search Template | Button States 
---- | ----
Captura de ecrã 2022-07-05, às 12 08 11 | Captura de ecrã 2022-05-11, às 11 55 07 

Also I didn't fix the issue with the button outlines because I think @matiasbenedetto is working on that as a part of #6142.

#### Related issue(s):

Closes #6157

### DIFF
--- a/curator/patterns/hidden-search-form.php
+++ b/curator/patterns/hidden-search-form.php
@@ -10,4 +10,4 @@
 <!-- wp:heading {"textAlign":"center","level":1,"style":{"spacing":{"margin":{"bottom":"4rem"}}}} -->
 <h1 class="has-text-align-center" style="margin-bottom:4rem"><?php echo esc_html__( 'Search', 'curator' ); ?></h1>
 <!-- /wp:heading -->
-<!-- wp:search {"label":"","showLabel":false,"placeholder":"<?php echo esc_html__( 'Search', 'curator' ); ?>","buttonUseIcon":true,"borderColor":"foreground"} /-->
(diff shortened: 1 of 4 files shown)


beafialho:
> @beafialho in the Figma, the template suggests the search button is black by default, but in the button states section it is burgundy. Can you confirm which is correct?

Yeah, sorry for the confusion. The correct is the states section: Burgundy by default and Black on hover.

Can we remove the outline from the updated button with the icon?

jffng:
> Can we remove the outline from the updated button with the icon?

Done in https://github.com/Automattic/themes/pull/6160/commits/a9b4fc61ec4c3878599c237be3b660e1511320c6, sorry @matiasbenedetto if there is some overlap with the work you're doing but wanted to fix it now since I think there still could be an issue once button elements are implemented.

PR #6164, opened 2022-07-06
Curator: rename to Vivre

jffng:

Renames the theme to Vivre since there is already an old theme named Curator.

We can rename the label and milestones, and maybe replace the screenshot also once the name change is approved.


jffng:
@beafialho do you think the screenshot should be updated to reflect the new theme name?

beafialho:
Yes, here it is:

Vivre Screenshot

And also the description:

```
Vivre is a bold, opinionated blogging theme, heavily inspired by fashion and lifestyle magazines and websites.
With large heavy sans-serif paired with a light elegant serif font, it sets the tone to a refined reading experience.
```

PR #6166, opened 2022-07-06
Curator: Implement support for Button elements in theme.json

### LINKED ISSUE #6142: Vivre: add support for Button elements

matiasbenedetto:
Adding settings for button elements in theme.json and removing the CSS no longer needed

#### Related issue(s):
https://github.com/Automattic/themes/issues/6142

### DIFF
--- a/curator/style.css
+++ b/curator/style.css
@@ -30,15 +30,14 @@ body {
  * https://github.com/WordPress/gutenberg/issues/27075
  */
 
-.wp-block-button:not(.is-style-outline) .wp-block-button__link:not(.has-background) {
-	padding: 0;
-}
 .wp-block-button__link:visited {
 	color: currentColor;
 }
-.wp-block-button:not(.is-style-outline) .wp-block-button__link:hover {
+.wp-block-button:not(.is-style-outline, .form-submit) .wp-block-button__link:hover {
 	text-decoration: underline;
-	color: var(--wp--preset--color--primary);
+	color: var(--wp--preset--color--background);
+	background: var(--wp--preset--color--foreground);
+	text-decoration: none;
 }
 .wp-block-button:is(.is-style-outline) .wp-block-button__link:hover {
 	background: var(--wp--preset--color--foreground);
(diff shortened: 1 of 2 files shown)


jffng:
I am having trouble testing this because when I use Gutenberg trunk I get the following fatal error (I think unrelated to this PR because I get the same error using emptytheme):

```( ! ) Fatal error: Uncaught Error: Call to undefined function gutenberg_style_engine_get_block_supports_styles() in /Users/jeffreyong/Documents/THEMES/a8c-projects/gutenberg/lib/block-supports/elements.php on line 108```

And when I switch to a stable release, e.g. 13.5.2, the buttons appear like this:

Screen Shot 2022-07-06 at 5 23 30 PM

matiasbenedetto:
@jffng is your dev env working fine? Can you reproduce this issue using the themes repo trunk branch?

I'm using the latest Gutenberg trunk version and it works fine for me. I'm getting this:
image

pbking:
This is what I'm picking up for reg'ler buttons on this branch (and Gutenberg trunk) which looks like what the theme.json is expressing but not what the design shows. (I didn't expect an outline around the button).  If that's expected then carry on!

image

Figma:
image

The search block looks right to me:

image

jffng:
I fixed my Gutenberg build and see the same thing as @pbking now. 

The default button states are incorrect now — there shouldn't be an outline or padding, and the hover state is picking up the incorrect colors.

**Expected**
Screen Shot 2022-07-08 at 10 13 33 AM

**This PR**
Screen Shot 2022-07-08 at 10 17 32 AM

I think this theme design is perhaps tricky because the default button blocks do not use the same style as other button elements (search, file, and comments).

pbking:
Pushed an updated that I think fixes the above and also makes the theme look pretty good with the Gutenberg plugin not activated.

Buttons:
image

Hover state:
image
image

Button elements (search block):
image

Buttons (sans-Gutenberg):
image

Button elements (search block) sans-gutenberg:
image

pbking:
I just realized that I changed the font family names in the change as well (from 'header and body' to the font family names).  I was going to open an issue for the same and shouldn't have made that change here.  Happy to back that out until we can discuss that further if need be.

jffng:
Nice, it's now looking correct for button elements but not button blocks. 

The design for default button blocks has no padding, with an underline on hover (not a background color). See [above](https://github.com/Automattic/themes/pull/6166#issuecomment-1179043547) for the design intent.

> I just realized that I changed the font family names in the change as well (from 'header and body' to the font family names).

Hmm I guess it raises the question, is there any benefit to having semantically named font slugs? Perhaps a question to discuss with design folks as well.

matiasbenedetto:
Thanks, @pbking and @jffng for your feedback and additions. I added a new commit around button and link styles.

jffng (approved):
I made one more [change](https://github.com/Automattic/themes/pull/6166/commits/d07155ae1e3c89e739703e87d395dc341ae9dd16) — remove the padding from default buttons via theme.json. 

LGTM!

PR #6043, opened 2022-05-26
Refactor/blockbase color admin

pbking:
#### Changes proposed in this Pull Request:

* Removes the Color Customization controls in the Customizer.
* Adds the "Palette Options" previously available in the Customizer Color Customization as "Theme Variations"

image

image

Note: Variations do NOT inherit from the parent.  This means that any child theme that does NOT provide theme variations will not have Blockbase's.  Child themes that did NOT provide palette collections would have been provided Blockbase's in the Customizer.

This change should not BREAK any themes, however any child themes should migrate their palettes to variations to continue to provide that functionality.

All of the public (and Videomaker) blockbase themes have been refactored (Geologist, Quadrat, Zoologist and Videomaker)

#### Related issue(s):

NOTE: This should be merged and released at the same time as the Font Customization refactor and released as a Major Point Release for the Blockbase theme.

PR #6167, opened 2022-07-06
Blockbase 3

pbking:
All of the branch from the Blockbase 3 milestone in one place!  (So they can all be tested before it ships to /trunk)

https://github.com/Automattic/themes/milestone/209

Refactor/blockbase color admin (https://github.com/Automattic/themes/pull/6043)
Moved templates from old folder location to new (https://github.com/Automattic/themes/pull/6073)
Blockbase: Implement the Button elements API (https://github.com/Automattic/themes/pull/6041)
Blockbase: Implement Comment Block and removed CSS (https://github.com/Automattic/themes/pull/6080)
Fix/migrate blockbase font self hosted (https://github.com/Automattic/themes/pull/6123)
Blockbase children: update comments block (https://github.com/Automattic/themes/pull/6153)
Blockbase: Changed the trigger to render social icons (https://github.com/Automattic/themes/pull/6079)
Blockbase: move button padding styles from ponyfill to theme.json (https://github.com/Automattic/themes/pull/5901)


jffng (commented):
Thanks for pulling this all together. I gave the following scenario a test and it worked without any issues as far as I could tell:

- Activated blockbase on trunk, made changes to colors, fonts, and navigation via the customizer
- Switched to this PR, verified my site appeared as expected (with no changes)
- Went to the customizer and verified I was being prompted to edit fonts and colors via the Site Editor
- Went to the site editor

I had two thoughts, not exactly bugs: 

1. The Customizer > Fonts and Colors prompts read a bit inconsistently — one has a button to visit the site editor and the other is very descriptive text:

Fonts | Colors
--- | ---
Screen Shot 2022-07-12 at 10 51 46 AM | Screen Shot 2022-07-12 at 10 51 51 AM

2. I don't know that anything can be done about this, but in the Customizer > Fonts, you have the option to edit your heading and body text. In the site editor, it's Text, Links, and Buttons, which isn't a 1:1 mapping... but I'm sure this is something you had to already consider.

If there are other scenarios that would be helpful to test, let me know!

pbking:
> The Customizer > Fonts and Colors prompts read a bit inconsistently 

Good point, I'll make the Colors match more closely the fonts.

> I don't know that anything can be done about this, but in the Customizer > Fonts, you have the option to edit your heading and body text. In the site editor, it's Text, Links, and Buttons, which isn't a 1:1 mapping... but I'm sure this is something you had to already consider.

This change actually migrates the font customizations from `body-font` in the customizer to the `text` font customization you mention while `header-font` is migrated to customizing two blocks: `core/heading` and `core/post-title` So to manage what was done in the customizer as a single dropdown you will need to change both of those blocks.  This is different from the "top level" places where you can change the fonts that you mention.  (Which I find very confusing!)
And there's no easy way to tell the user that!

pbking:
> If there are other scenarios that would be helpful to test, let me know!

It would also be good to test the same scenario you walked through but with child themes.

pbking:
Color instructions [updated](https://github.com/Automattic/themes/pull/6043/):
image

PR #6176, opened 2022-07-11
Remote: move link styles to theme.json

madhusudhand:

Move the link pseudo state styles to theme.json

### Testing instructions

1. Activate remote theme
2. Verify Link and navigation styles. They should work same as trunk on `:hover`, `:active` and `:focus`.
3. Also, verify the same for other style variations.

**Note:** This would be a breaking change before WP 6.1

#### Related issue(s):

#6108


remote/style.css
```diff
@@ -173,28 +173,8 @@ body > .is-root-container,
  * Link Details
  */
 
- a {
-	text-decoration-thickness: 0.075ex;
```
matiasbenedetto, line 177: Why we are removing the `text-decoration-thickness` property?

matiasbenedetto (approved):
This looks very good. I'm just adding a commit setting again the removed `text-decoration-thickness` but in `theme.json`.

matiasbenedetto:
Moving to "Approved but not mergeable" until theme.json link elements API makes to WordPress core.

PR #6178, opened 2022-07-11
Pendant: Removing default text decoration for link elements

### LINKED ISSUE #6170: Pendant: Implement latest Gutenberg link elements api for theme.json
## Steps to replicate

- Latest changes to Gutenberg links resulted in links being underlined by default. We need to remove the default underline for titles.
- Apart from that we can remove the CSS related to hover and focus state of links.

## Result

image

## Expected

image

matiasbenedetto:
Pendant theme:
The latest changes to Gutenberg links resulted in links being underlined by default. In this PR I'm removing default text-decoration for link elements

#### Related issue(s):
https://github.com/Automattic/themes/issues/6170

### DIFF
--- a/pendant/theme.json
+++ b/pendant/theme.json
@@ -347,6 +347,9 @@
 			"link": {
 				"color": {
 					"text": "var(--wp--preset--color--foreground)"
+				},
+				"typography": {
+					"textDecoration": "none"
 				}
 			}
 		},
--- a/pendant/theme.json
+++ b/pendant/theme.json
@@ -349,7 +349,7 @@
 					"text": "var(--wp--preset--color--foreground)"
 				},
 				"typography": {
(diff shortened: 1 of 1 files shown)


pbking:
This also impacts the `text-decoration-thickness` but unfortunately that doesn't seem to be fix-able via theme.json.

This is half of what's broken.  Do we need to fix the rest of that in CSS?

Expected:
image

Observed:
image

pbking:
Ah!  Actually I had forgotten that `text-decoration` is a shorthand property.  :) Setting the width there works smashingly. :D

`"textDecoration": "none 0.02em"`

With that change I think things look right again and this can ship.

Thanks for jumping on this @matiasbenedetto

pbking (approved):

PR #6179, opened 2022-07-11
Pendant: replacing CSS with theme.json settings for link elements

### LINKED ISSUE #6170: Pendant: Implement latest Gutenberg link elements api for theme.json

matiasbenedetto:
Pendant: replacing CSS with theme.json settings for link elements

:warning: The theme.json elements API for core/post-content settings works OK in the frontend but those styles are not applied in the editor. That is a bug potentially fixed by: https://github.com/WordPress/gutenberg/pull/42270

#### Related issue(s):
https://github.com/Automattic/themes/issues/6170


pbking:
This works with the plugin, however since this was released with hover states working without the plugin I believe we need to hang onto this until the feature has been released in core.

Sticking into the "yes but not yet" bucket unless there are objections.

pbking (approved):
Yes but not until pseudo-selectors are available in WP Core.

PR #6183, opened 2022-07-11
Vivre: Fixing post grid with one item

### LINKED ISSUE #6180: Vivre: Post grid breaks if it only has one post
## Steps to replicate
- Activate Vivre/Curator theme
- Enter a search term with only one result
- See how the grid looks

## Result

image

## Expected
image

matiasbenedetto:
Vivre: Post grid breaks if it only has one post. This PR tries to fix that.

#### Related issue(s):
https://github.com/Automattic/themes/issues/6180

Before:
image

After:
image

### DIFF
--- a/curator/patterns/posts-grid.php
+++ b/curator/patterns/posts-grid.php
@@ -6,29 +6,36 @@
  */
 ?>
 
-<!-- wp:group {"align":"full","layout":{"type":"flex","flexWrap":"nowrap"}} -->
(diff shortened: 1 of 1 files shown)


jffng (approved):
Before | After
--- | ---
Screen Shot 2022-07-11 at 12 34 57 PM | Screen Shot 2022-07-11 at 12 34 46 PM

It also still appears correctly when there is more than one post. 

Thank you!

PR #6186, opened 2022-07-11
Vivre: fix font slug references

jffng:

https://github.com/Automattic/themes/pull/6166 renamed the font slugs from their semantic names, and we missed a couple patterns where the old slugs were being referenced. This PR changes them to the correct names.


pbking:
Y'all I'm sorry.  I shouldn't have included that change in the other. :(

pbking (approved):
Tests out and looks right.  Kinda feels strange putting descriptive variables in a (potentially sharable) pattern but that's how it goes I suppose.

PR #6200, opened 2022-07-13
Migrate themes repo to npm workspaces

madhusudhand:

**Migrate to npm workspaces to solve/improve the following.**

* With workspaces, single `npm install` would do it for all themes.
* removes many lock files (from each individual theme)
* With every deployment, package.json version (of specific a theme) is bumped and lock file is left behind. Now that there is only single lock file, this not longer an issue.

Repo already mandates min node version to `16` (in .nvmrc and package.json). `Node 16` comes with min `npm v7` and workspaces are introduced in `npm v7`. Hence, there isn't any impact.

**Update the lock file version to 3**

Depending on the current npm version, when user does `npm install` it updates lock file version to npm default like from current v1 to 2 or 3. In order to control the behavior, introduced `.npmrc` which always keeps a specific version of lock file.

`npm v7` introduces [lock file version 3](https://docs.npmjs.com/cli/v8/using-npm/config#lockfile-version) which is smaller and works with max determinism of the package versions.

#### Related issue(s):

None.


madhusudhand:
With this change I hope to reduce too many `dependency bot` version bump PRs to minimal, because it has to update single lock file now, instead of looking at `30+` lock files.

mikachan:
Oh this is cool! I tried handling this with Lerna for this repo but it ended up being too heavy-handed. This seems like a great way to handle multiple projects with packages.

package.json
```diff
@@ -60,5 +60,43 @@
...
+		"shawburn",
+		"skatepark",
+		"sophisticated-business",
+		"spearhead",
+		"stow",
+		"stratford",
+		"varia",
+		"videomaker",
+		"zoologist"
+	]
```
vcanales: In order to avoid having to manually define new themes each time they're added, we could consider doing something like the following (not sure about the naming of the directory):

```json
  "workspaces": [
    "./packages/*",
  ]
```

That would also involve moving around some of the build, deployment, and publishing scripts to reflect the directory change though.

madhusudhand: Thats right. Defining packages directory is the best approach. 
Considering the fact that, we may not see many changes/additions to this because in the new block themes we avoid adding build dependencies, I left the directory structure as is.

mikachan: Could we use something like this instead, that covers all directories so we wouldn't need to change the directory structure?

```
"workspaces": ["**/*"],
```

I believe the directories that don't have a package.json file would be skipped.

madhusudhand: It works! 🎉 
```
"workspaces" : ["*"]
```

mikachan (approved):
LGTM!

PR #6205, opened 2022-07-13
Premium Theme: Script to sync from private repository

valterlorran:

* Add a new command to sync the private premium themes in the sandbox. This script will help with the deployment of the new premium themes that will be created on private repositories.

#### Testing instructions

* Apply this PR
* Run the following command to sync an existing theme: `yarn run deploy:pull:premium food-truck`

#### Related issue(s): 
https://github.com/Automattic/wp-calypso/issues/65110

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -970,7 +970,27 @@ async function syncPremiumTheme(theme) {
 	}
 }
 
+async function repositoryExists(repository) {
+	try {
+		await executeOnSandbox(`
+			git ls-remote ${repository}
+		`, true);
+		return true;
+	} catch(e) {
+		if (e.indexOf('ERROR: Repository not found') === -1) {
+			throw new Error('Unknow error verifing if the repository exists!');
+		}
+
+		return false;
+	}
+}
+
 async function pullPremiumTheme(theme) {
+	const repoExists = await repositoryExists(`git@github.com:Automattic/theme-${theme}.git`);
+	if (!repoExists) {
+		console.error(`Repository git@github.com:Automattic/theme-${theme}.git could not be found.`);
+		return;
+	}
 	if (!theme) {
 		console.error('Invalid theme!')
 		return null;
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -86,7 +86,7 @@ const commands = {
 		run: pullCoreThemes
 	},
 	"pull-premium-theme": {
-		helpText: 'Given a theme slug, sync it on the premium folder.',
+		helpText: 'Given a theme slug, sync it to the premium folder.',
 		additionalArgs: '<theme-slug>',
 		run: (args) => pullPremiumTheme(args?.[1])
 	},
@@ -978,7 +978,7 @@ async function repositoryExists(repository) {
 		return true;
 	} catch(e) {
 		if (e.indexOf('ERROR: Repository not found') === -1) {
-			throw new Error('Unknow error verifing if the repository exists!');
+			throw new Error('Unknown error verifying if the repository exists!');
 		}
 
 		return false;
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -85,6 +85,11 @@ const commands = {
 		helpText: 'Use rsync to copy any changed public CORE theme files from your sandbox to your local machine. CORE themes are any of the Twenty<whatever> themes.',
 		run: pullCoreThemes
 	},
+	"pull-premium-theme": {
+		helpText: 'Given a theme slug, sync it on the premium folder.',
+		additionalArgs: '<theme-slug>',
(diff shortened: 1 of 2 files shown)


theme-utils.mjs
```diff
@@ -937,6 +942,63 @@ async function pushChangesToSandbox() {
...
+
+async function repositoryExists(repository) {
+	try {
+		await executeOnSandbox(`
+			git ls-remote ${repository}
+		`, true);
+		return true;
+	} catch(e) {
+		if (e.indexOf('ERROR: Repository not found') === -1) {
+			throw new Error('Unknow error verifing if the repository exists!');
```
sixhours: Minor typos: "Unknown error verifying if the repository exists!"

theme-utils.mjs
```diff
@@ -85,6 +85,11 @@ const commands = {
 		helpText: 'Use rsync to copy any changed public CORE theme files from your sandbox to your local machine. CORE themes are any of the Twenty<whatever> themes.',
 		run: pullCoreThemes
 	},
+	"pull-premium-theme": {
+		helpText: 'Given a theme slug, sync it on the premium folder.',
```
markbiek: Minor word change for clarity 

```suggestion
		helpText: 'Given a theme slug, sync it to the premium folder.',
```

sixhours (approved):
Tested and this works well 🥳

madhusudhand (approved):
LGTM!

pbking:
I'm not sure that we want to put this logic here.  My understanding is that the premium work will be done in a separate (non wpcom) git repository (private repo on github?).  If that's the case then I suggest any scripting to pull/push/sync/deploy be kept separate in that repository instead.

The 'videomaker' theme is the only reason there is ANY "premium" stuff in the script already (and I'm not very comfortable with that, I would prefer it be removed and eliminate the "is it premium or not" logic completely.

I worry that merging the two is only going to create problems down the road.

If instead we AREN'T going to be keeping that work in a repository and instead operate completely on phabricator diffs (shudder) then I suppose having a mechanism to pull/push themes so that you can work on them locally makes sense.

andres-blanco:
> I'm not sure that we want to put this logic here. My understanding is that the premium work will be done in a separate (non wpcom) git repository (private repo on github?). If that's the case then I suggest any scripting to pull/push/sync/deploy be kept separate in that repository instead.
> 
> The 'videomaker' theme is the only reason there is ANY "premium" stuff in the script already (and I'm not very comfortable with that, I would prefer it be removed and eliminate the "is it premium or not" logic completely.
> 
> I worry that merging the two is only going to create problems down the road.
> 
> If instead we AREN'T going to be keeping that work in a repository and instead operate completely on phabricator diffs (shudder) then I suppose having a mechanism to pull/push themes so that you can work on them locally makes sense.

👋 @pbking We resolved putting this logic here since each premium theme is going to be handled in a separate private GH repo. This seems like a good-enough place to put the logic since we can reuse common methods. This is the P2 outlining the process: paYKcK-1Ru-p2

andres-blanco:
@blackjackkent do you want to chime in here?

blackjackkent:
Yep! @pbking we definitely want to have this logic in a centralized place so that future changes for multiple individual private theme repos aren't maintaining copies of the same code. With individual repos for every new premium theme being created, it would not be sustainable or practical to have a new copy of the script in each one.

If there's another centralized location for this script that you think would be more reasonable we could definitely look at moving it though.

pbking:
I definitely see the need to have a single place for the tools to manage the themes, but I don't think that THIS is that place.  There might be some overlap in functionality but I think that keeping the two together invites more complication than we have to. 

It was mentioned (I think in P2) that perhaps there could be an overarching project that brings in the individual repos via git submodules.  That seems like the more appropriate place for all of the tooling that you would need to work with the different theme repositories and the sandbox.

blackjackkent:
> I definitely see the need to have a single place for the tools to manage the themes, but I don't think that THIS is that place. There might be some overlap in functionality but I think that keeping the two together invites more complication than we have to.

Fair. Just to make sure I'm understanding correctly - your thought is that there should be one "public themes" syncing script and one "premium themes" syncing script?

Possibly we can create a GH repo purely for handling premium theme related tools and have that be our centralized point for at least all premium support work.

pbking:
> Fair. Just to make sure I'm understanding correctly - your thought is that there should be one "public themes" syncing script and one "premium themes" syncing script?

> Possibly we can create a GH repo purely for handling premium theme related tools and have that be our centralized point for at least all premium support work.

Yes, that's exactly what I was thinking. :)

valterlorran:
I've created the private repository following the consensus. The intention is to have a central place with all the tools related to premium themes.
[Automattic/premium-theme-tools](https://github.com/Automattic/premium-theme-tools)

Issue #2123, opened 2020-06-13
Varia + TwentyTwenty: Getty Images Embed Displayed Incorrectly


eduardozulian:
Getty Images embeds, just like Google Docs on in #1953, also don't have a wrapper div to make sure they are center-aligned. 
 
## Steps to replicate

1. Start a post or a page
2. Add a [Getty Images embed](https://wordpress.com/support/images/getty-images/)

## Result

<img width="1267" alt="Captura de Tela 2020-06-13 às 11 33 10" src="https://user-images.githubusercontent.com/1136160/84571500-e4cc7f80-ad69-11ea-8e4d-7ce1a5be342f.png">

## Expected

To see the embeds centered, which is only possible with some custom CSS:

```
/* Getty Images: center-align embeds -- XXXXX-hc/zen */
.getty.embed.image {
  display: block !important;
}
```
---
User report: #3060306-zen

mrfoxtalbot:
I am still able to reproduce this issue, @eduardozulian.

Aside from the CSS you shared above, another very simple way to work around this would be to group the HTML block:
![Screen Shot on 2022-06-23 at 16:12:04](https://user-images.githubusercontent.com/4452464/175321283-0cf552a7-b9cf-4c9d-aeef-cbef5364546d.png)

Having it sit inside a group block will fix the issue will prevent it from floating to one side:

![Screen Shot on 2022-06-23 at 16:12:30](https://user-images.githubusercontent.com/4452464/175321376-1ab490a4-5f58-4641-b55f-c9800eb7beb3.png)

vcanales:
As mentioned by @mrfoxtalbot, we should consider additionally documenting the process for making the alignment customizable through the editor: 

![Kapture 2022-07-22 at 18 52 21](https://user-images.githubusercontent.com/1157901/180578458-7324ecd3-0d9f-4b22-8f9d-3de9fdea7c03.gif)

bobmatyas:
Fixed via https://github.com/Automattic/themes/pull/6231

PR #6231, opened 2022-07-14
Varia - center Getty image embeds

gwensmithx:
Code is from this open issue: https://github.com/Automattic/themes/issues/2123

Display Getty embed images as blocks to center them. Currently when Getty images are embedded on a page they will float to the left outside of the page content.

#### Related issue(s):
https://github.com/Automattic/themes/issues/2123


vcanales (approved):
This looks good to me. Would be mergeable after resolving conflicts.

I left a comment over at #2123 suggesting possible next steps, which would mean documenting the process through which users could change the alignment.

https://github.com/Automattic/themes/issues/2123#issuecomment-1192991680

Issue #6221, opened 2022-07-13
club: Templates: Search, Category, Tag


matiasbenedetto:
Implement Search, Category, Tag pages following the mockups.
![image](https://user-images.githubusercontent.com/1310626/178961134-167bd1ba-c2aa-46dd-b1fb-7ec6286481a7.png)

PR #6236, opened 2022-07-14
Club: Templates: Search, Category, Tag

matiasbenedetto:
Implement archives template

#### Related issue(s):
https://github.com/Automattic/themes/issues/6221


club/patterns/post-list.php
```diff
@@ -0,0 +1,40 @@
+<?php
```
scruffian: Is there a reason this can't be an HTML file?

madhusudhand (commented):
It looks great. 
Following are few observations (I kept typography out of context because it would be taken care in different issues).

image

1. When post title is large, it breaks "read more" and date to next line. it should break the post title to next line keeping the date and button aligned with other posts.
2. post is not vertically center aligned. Also, it has different spacing bottom with and without featured image.
3. Uppercase the date.
4. Design suggests default button text as "READ IT"

club/patterns/post-list.php
matiasbenedetto: I think we need PHP to make the strings translatable.

matiasbenedetto:
The latest commits should cover these items:
> post is not vertically center aligned. Also, it has different spacing bottom with and without featured image.
> Uppercase the date.
> Design suggests default button text as "READ IT"

About the last remaining item:

> When post title is large, it breaks "read more" and date to next line. it should break the post title to next line keeping the date and button aligned with other posts.

I'm not sure if this is the expected behavior by watching the current design. Also it dosn't seem simple to achieve with CSS that would break the functionality of the editor.

I propose to approve this PR in its current state and try to follow up the last item as a refinement once the theme is more complete.

madhusudhand:
> I'm not sure if this is the expected behavior by watching the current design. Also it dosn't seem simple to achieve with CSS that would break the functionality of the editor.

Archive page design wraps the post title without bringing the other items to next line.
image

I guess may not be possible to avoid CSS to achieve the design. I tried the following to achieve the designs with minimal css. 

1. Setting the flex wrap to nowrap for all `three row blocks`.

image

This brings them to same line, but it still breaks the text of buttons and buttons to multiple lines. Also image needs fixed width shrinks.

```
// image block
flex-shrink: 0;

// post-date and read-more blocks
white-space: nowrap;
```

image

It seems to be working. But needs more testing in other areas of the theme as well.
We could reserve these changes to another PR and test it well before merging.

madhusudhand (approved):
Everything else LGTM!

matiasbenedetto:
Thanks for the testing @madhusudhand 
> Setting the flex wrap to nowrap for all three row blocks.


I think the problem with that approach is that it will work on large viewports but on mobile, it looks bad and it doesn't match with the wrap behavior present in the mockups. Example:

image

PR #6238, opened 2022-07-14
Club: adding missing pallette colors

matiasbenedetto:
Club: adding missing pallette colors

image


club/theme.json
```diff
@@ -46,6 +46,16 @@
 					"slug": "background",
 					"color": "#F6EB80",
 					"name": "Background"
+				},
+                {
```
jffng, line 40: Some awkward formatting going on here?

club/theme.json
```diff
@@ -46,6 +46,16 @@
...
+					"slug": "primary",
+					"color": "#000AFF",
```
jffng: What is the benefit to having this twice (both foreground and primary)? I guess if the user wants to change the button and link colors for example but not the foreground color?

club/theme.json
```diff
@@ -46,6 +46,16 @@
...
+					"name": "Primary"
+				},
+				{
+					"slug": "tetriary",
```
jffng: Kind of a nitpick, but `tertiary` is technically third order I believe, so should this actually be `secondary`?

club/theme.json
matiasbenedetto: I think the design is adding this option with the idea of making it customizable.

club/theme.json
matiasbenedetto: :thinking: I saw this in many of our latest themes. Tertiary color without a secondary one. I think the name can be related to the color theory and not to the number of colors in our palette. Primary is, of course, the primary color, and background would be the secondary.

club/theme.json
matiasbenedetto: Seems to be fixed now

club/theme.json
scruffian: I think that made more sense when we used the customizer for colors, but I don't know that it makes sense in block themes.

club/theme.json
```diff
@@ -37,17 +37,27 @@
...
+                {
+                    "slug": "primary",
+                    "color": "#000AFF",
+                    "name": "Primary"
+                },
+                {
+                    "slug": "tetriary",
+                    "color": "#FFF7A7",
+                    "name": "Tetriary"
+                }
```
vcanales: 👋  nothing to add other than this small typo:

```suggestion
                {
                    "slug": "tertiary",
                    "color": "#FFF7A7",
                    "name": "Tertiary"
                }
```

jffng (approved):

Issue #6248, opened 2022-07-15
Style variations missing in dotOrg published themes


madhusudhand:
## Steps to replicate

Style variations in the published themes [Pendant](https://wordpress.org/themes/pendant/) and [Remote](https://wordpress.org/themes/remote) are missing in dotOrg.
There exists styles directory, but it is empty.

Style variations are available and working fine in dotCom

## Result

Style variations aren't working local installs.

## Expected

dotOrg theme source should include style variations.

pbking:
The [assets excluded ](https://github.com/Automattic/themes/blob/blockbase-3/dotorg-exclude.txt#L6)from being packaged in the dotorg zip files include all json files while theme.json is [explicitly included](https://github.com/Automattic/themes/blob/blockbase-3/package-dotorg.sh#L14).

We need to either explicitly allow any *.json file in the /styles folder to be included or remove the *.json exclusion.  I'm not sure which is better.

jeffikus:
@pbking sounds like quickest path the fixing this is allowing the *.json files in the /styles folder.

madhusudhand:
Once the fix is merged, `Pendant` and `Remote` needs version bump.

pbking:
As mentioned those themes need a version bump.  Done here: https://github.com/Automattic/themes/pull/6259

madhusudhand:
Themes published to dotOrg now include style variations.

PR #6253, opened 2022-07-18
Update dotOrg publisher to include style variations

madhusudhand:

Include the `styles/*.json` in dorOrg publisher so that theme variations gets published.

#### Testing instructions

Validate the following to verify all the required theme files copied properly

```
cd pendant
rsync -avz --include-from='../dotorg-include.txt' --exclude 'pendant' --exclude-from '../dotorg-exclude.txt' ./ 'pendant'

# cleanup
rm -r pendant
```

#### Related issue(s):

#6248

### DIFF
--- a/dotorg-exclude.txt
+++ b/dotorg-exclude.txt
@@ -1,9 +1,9 @@
 inc/headstart
 node_modules
+package.json
 .git
 *.DS_Store
 *.sh
-*.json
 *.map
 *wpcom*
 *.zip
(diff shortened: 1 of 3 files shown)


pbking:
I don't see any *.json files that we would need to so aggressively exclude.  With the exception of package.json which we can just add to the dotorg-exclude.txt instead of *.json.

Would we be including anything we shouldn't if we make that change instead?

madhusudhand:
> Would we be including anything we shouldn't if we make that change instead?

Haven't excluded the lockfile, because it is no longer there with workspaces. excluding package.json alone making sense.
Updated the script.

pbking (approved):

PR #6199, opened 2022-07-12
#6004: Seedlet: Fix link underline colors in Safari

bobmatyas:
### Changes proposed in this Pull Request:

Safari doesn't seem to fully support the `text-decoration` shorthand. By separating out the values I was able to fix this in Safari

#### Testing:

1. Create a blog post with a link
2. View and note that the link now shows the underline in Safari

#### Before

Screenshot on 2022-07-12 at 14-54-35

#### After

Screenshot on 2022-07-12 at 15-12-18

### Related issue(s):

Fixes https://github.com/Automattic/themes/issues/6004

PR #6254, opened 2022-07-18
update deploy script to consider level 3 files for version bump

madhusudhand:

Themes such as seedlet has the `.scss` files under level 3 and the script would not version bump them.
These leaves `style.scss` and `style.css` at different versions and when the theme is rebuild it lowers the theme version.

### Question: Do we need to maintain `-wpcom` version suffix ?

The above script overrides the source file with a version without a `-wpcom` suffix and I left it because many themes doesn't have a suffix.

#### Related issue(s):

Resolves https://github.com/Automattic/themes/pull/6199#discussion_r921990603

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -728,7 +728,7 @@ async function versionBumpTheme(theme, addChanges) {
 	let styleCss = fs.readFileSync(`${theme}/style.css`, 'utf8');
 	let currentVersion = getThemeMetadata(styleCss, 'Version');
 
-	let filesToUpdate = await executeCommand(`find ${theme} -name package.json -o -name style.scss -o -name style-child-theme.scss -maxdepth 2`);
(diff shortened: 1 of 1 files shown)


pbking:
Thank you for finding and fixing.  All this time I thought those files were getting updated. !!

I was never clear about the purpose or need for the -wpcom suffix.  I believe that in situations where the themes may differ (such as core themes) I can see that being important to differentiate.  However I'm unclear if that is important or necessary for the themes in this repository effected by this deploy script.

That shouldn't hold up bringing this change in, however.

pbking (approved):
👍

jffng:
> I was never clear about the purpose or need for the -wpcom suffix.

I recall a (some would say dark) time when classic themes required patchwork code to support things like color palettes in the customizer on wordpress.com. The code required for that functionality did not make sense (or even work) in non-wpcom contexts, so it was placed  in separate files with the wpcom suffix and included conditionally.

PR #6259, opened 2022-07-20
Version Bump themes that have style variations so that they ship …

pbking:
Version Bump themes that have style variations so that they ship to wporg (via deploy action).

The change #6253 fixed the process, this triggers the change to be shipped.


madhusudhand (approved):
LGTM!

PR #6260, opened 2022-07-20
Antonia: Typography fixes

onuro:
Headings line height and weight updates


madhusudhand (approved):
This update also effects blocks such as post-title because they are also headers.
I believe thats the expectation.

PR #6268, opened 2022-07-21
Migrate antonia fonts to new Blockbase format

pbking:
When child fonts were migrated for Blockbase 3 this one was missed.  The result was that the fonts no longer loaded properly.

This change migrates the fonts configuration.  This should partially address [this Calypso issue](https://github.com/Automattic/wp-calypso/issues/65832).

To test:
Activate Antonia (with no customizations).  Load the view and ensure the body font is rendering Inter and the heading font is rendering DM Sans.

Navigate to the Full Site Editor and adjust the body font.  Ensure that all of the Blockbase fonts are available for selection there.  Ensure that choosing one of those fonts, and saving in the FSE "sticks".  (Reload the FSE, ensure the selection is still made.  Load the view, ensure the newly selected font is used.)

image

### DIFF
--- a/antonia/theme.json
+++ b/antonia/theme.json
@@ -11,6 +11,7 @@
 		"blocks": {
 			"core/heading": {
 				"typography": {
+                    			"fontFamily": "var:preset|font-family|dm-sans",
 					"fontStyle": "normal",
 					"lineHeight": "1.4",
 					"fontWeight": "300"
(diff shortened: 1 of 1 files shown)


jffng (approved):
Verified all of the expected Blockbase fonts load for Antonia and that setting them "sticks": 

Screen Shot 2022-07-21 at 10 18 43 AM

PR #6269, opened 2022-07-21
Filter out Jetpack fonts when Blockbase is activated

### LINKED ISSUE #65832: The font getting reset
### Quick summary

The user reached out to us to inform the fonts on their site changed. When tried they aren't able to change it back through Global Styles setting on Site Editor.

### Steps to reproduce

We don't know the exact steps here to replicate, but we're suspecting it is caused due to Blockbase themes since we've got multiple cases using different themes.

On my test site I'm able to replicate that changing the font doesn't save it. Here are the steps I followed:
1. Activate `Antonia` theme on my atomic site.
2. Checked the font from Global Style and it was set to default, showing Times New Roman font family when checked.
3. When changing the font from Global Styles, it doesn't save.

Here is a screencast video showing the saving of the font issue:

https://user-images.githubusercontent.com/7382505/180166193-78aef8d4-2888-4f3b-bc5b-978d4ab573a9.mov

The font family does not change randomly. And when the font is changed it gets saved correctly.

### What actually happened

The font was changed randomly.

### Context

Customer report:

Theme: `Antonia`
Chat: 20702780-hc
Follow up: 5412091-zd-woothemes
Site URL: https://roeartstudio.com/

Some other cases: 
Simple site: p1658376880935749-slack-CJDUN89L1
Atomic site: p1658354145149709-slack-C03TY6J1A

Another simple site: https://ltstrategicstories.com/

Simple, Atomic

### Theme-specific issue?

The issue is replicated on multiple themes:

Quadrat White, Antonia, Blank Canvas, etc.

_No response_

### Reproducibility

Intermittent

### Severity

Some (< 50%)

### Available workarounds?

Yes, difficult to implement

### Workaround details

In some cases of atomic sites, deactivating the Gutenberg plugin fixes the issue. On my user's site, deactivating the Gutenberg plugin didn't make any change.

pbking:
This fixes https://github.com/Automattic/wp-calypso/issues/65832

Blockbase and Jetpack are both trying to manage many of the same font faces.

Initially Blockbase was filtering out all of the font faces that it was managing from the list of font faces that Jetpack was handling via `blockbase_filter_jetpack_google_fonts_list`.  This only worked as expected when Blockbase filtered out ALL fonts; if some of them remained to be registered then ONLY those are made available to the user (the font faces configured in theme.json are excluded).  

Additionally when those fonts are added something different happens depending on if the activated theme is the Parent (Blockbase) or the Child.  All of the fonts expected to be used are in the parent.  When Jetpack adds additional fonts with the parent activated it just appends to that list.  When a child theme is activated it REPLACES that list. I suspect this is a bug in the Gutenberg webfont handling bits and expect to open an issue regarding that later.

In some situations (only on Atomic as far as I can tell) more font faces are provided via Jetpack's Google Font Provider than what are provided by Blockbase.  Thus in this situation ONLY those 'additional' typefaces are actually available.

To address this we're now filtering out ALL JETPACK FONTS from Blockbase themes and rely exclusively on Blockbase font handling.  Hopefully this can be re-evaluated with an update to WordPress 6.1 where we expect better webfont handling to be added to Blockbase.

The downside is that if Jetpack provides additional fonts (see the note about more fonts being available on Atomic sites) those will not be available in the theme.

When testing ensure that there are no "double registrations" of typefaces (check the Full Site Editor to ensure that when available type faces are available there are no duplicates).

### DIFF
--- a/blockbase/inc/fonts/custom-fonts.php
+++ b/blockbase/inc/fonts/custom-fonts.php
@@ -6,14 +6,10 @@ require get_template_directory() . '/inc/customizer/wp-customize-fonts.php';
 // Font Migration
 require get_template_directory() . '/inc/fonts/custom-font-migration.php';
 
-// If Jetpack is already providing these fonts we aren't going to fight it.
-// Just drop out of the running of trying to manage them. It provides all of the
-// fonts that we provide here, though it doesn't host them locally.
-if ( ! function_exists( 'jetpack_add_google_fonts_provider' ) ) {
-	add_action( 'init', 'enqueue_global_styles_fonts', 100 );
-	add_action( 'admin_init', 'enqueue_fse_font_styles' );
-	add_filter( 'pre_render_block', 'enqueue_block_fonts', 10, 2 );
-}
+add_action( 'init', 'enqueue_global_styles_fonts', 100 );
+add_action( 'admin_init', 'enqueue_fse_font_styles' );
+add_filter( 'pre_render_block', 'enqueue_block_fonts', 10, 2 );
+add_filter( 'jetpack_google_fonts_list', 'blockbase_filter_jetpack_google_fonts_list' );
 
 $blockbase_enqueued_font_slugs = array();
 
@@ -207,4 +203,14 @@ function enqueue_block_fonts( $content, $parsed_block ) {
 		}
 	}
 	return $content;
+}
+
+/**
+ * Jetpack may attempt to register fonts for the Google Font Provider.
+ * If that happens on a child theme then ONLY Jetpack fonts are registered.
+ * This 'filter' filters out all of the fonts Jetpack should register
+ * so that we depend exclusively on those provided by Blockbase.
+ */
+function blockbase_filter_jetpack_google_fonts_list( $list_to_filter ) {
+	return array();
 }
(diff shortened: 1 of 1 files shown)


jffng (commented):
I'm not quite sure of the testing instructions, but here's what I tried: 

1. Spun up a new jetpack site (https://dominant-fiji.jurassic.ninja/)
2. Activated latest Gutenberg
3. Uploaded this PR's version of blockbase and quadrat white, activated quadrat white
4. Verified no fonts were duplicately loaded

Screen Shot 2022-07-21 at 10 33 14 AM

pbking:
Hrm... I think that the "Jetpack Google Fonts" are only loaded when something specific is done in the installation.  (I haven't figured out what that is yet) It doesn't seem to be enabled by default.  They are enabled on wpcom simple sites and Atomic.

If I run this branch on wpcom simple site I do get duplicate fonts...
image

Though I really really really thought that I wasn't earlier today.

Update:  I have indeed discerned that this behaves differently with parent (Blockbase) and Child themes.  It works as expected in nearly every scenario/method with the parent but results are quite strange with child themes.  Still working on a fix.

jffng:
I re-tested, and I do not see duplicate fonts when I push up this change to a simple site: 

Screen Shot 2022-07-21 at 1 09 32 PM

Seems like a fair change to me in order to address the missing fonts from those sites.

PR #6273, opened 2022-07-22
Exclude node_modules from version bump

pbking:
A recent change to this script upped the maxdepth from 2 to 3.  However that causes the find command to delve into node_modules to adjust package.json files found there which results in a git error when those changes are attempted to be committed.

This change excludes `node_modules` folder from the find command so it will delve deep enough to adjust *.scss files but stay out of node_modules.

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -704,7 +704,7 @@ async function versionBumpTheme(theme, addChanges) {
 	let styleCss = fs.readFileSync(`${theme}/style.css`, 'utf8');
 	let currentVersion = getThemeMetadata(styleCss, 'Version');
 
-	let filesToUpdate = await executeCommand(`find ${theme} -path "*/node_modules/*" -prune -name package.json -o -name style.scss -o -name style-child-theme.scss -maxdepth 3 -print`);
+	let filesToUpdate = await executeCommand(`find ${theme} -not \\( -path "*/node_modules/*" -prune \\) -and \\( -name package.json -or -name style.scss -or -name style-child-theme.scss \\) -maxdepth 3`);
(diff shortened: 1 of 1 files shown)


vcanales:
Hi! I've been playing around with original command and I haven't run into any git errors yet; is there a specific way to reproduce those?

theme-utils.mjs
```diff
@@ -704,7 +704,7 @@ async function versionBumpTheme(theme, addChanges) {
 	let styleCss = fs.readFileSync(`${theme}/style.css`, 'utf8');
 	let currentVersion = getThemeMetadata(styleCss, 'Version');
 
-	let filesToUpdate = await executeCommand(`find ${theme} -name package.json -o -name style.scss -o -name style-child-theme.scss -maxdepth 3`);
+	let filesToUpdate = await executeCommand(`find ${theme} -path "*/node_modules/*" -prune -name package.json -o -name style.scss -o -name style-child-theme.scss -maxdepth 3 -print`);
```
madhusudhand: It is ignoring `package.json` because it makes `and` condition with it

```suggestion
	let filesToUpdate = await executeCommand(`find ${theme} -not \\( -path "*/node_modules/*" -prune \\) -and \\( -name package.json -or -name style.scss -or -name style-child-theme.scss \\) -maxdepth 3`);
```

madhusudhand:
> Hi! I've been playing around with original command and I haven't run into any git errors yet; is there a specific way to reproduce those?

It is when a theme has different version of package that gets installed in `theme/node_modules` which has package.json
Those files also considered under the given depth.

madhusudhand (approved):

PR #6277, opened 2022-07-23
#4607: Pique: Front panel H2 heading centering issue

Gustavo-Hilario:
Adjust `style.css` to ensure the customization for H2 elements on the home page panels is only applied to the H2 under the header parent element.

Applying customization on H2 HTML elements only on the H2 under the "header" area on the Pique panels. That also fixes the alignment issue for H2 elements.

**Before**

pique#4607-before

---

**After**

pique#4607-after

#### Related issue(s):

Pique: Front panel H2 heading centering issue #4607

If we want to preserve the previous style applied to `H2`, we can keep this change only:

```
@media (min-width: 768px) {
	.pique-panel-content header h2 {
		max-width: 700px;
	}
}
```

### DIFF
--- a/pique/style.css
+++ b/pique/style.css
@@ -2342,7 +2342,7 @@ body:not(.no-background-fixed) .site-footer {
 		padding: 40px;
 	}
 }
-.pique-panel-content h2 {
+.pique-panel-content header h2 {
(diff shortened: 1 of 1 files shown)


mikachan (approved):
Thanks! LGTM 👍

PR #6281, opened 2022-07-25
escape block attributes in patterns utility

madhusudhand:

Until now, the theme pattern escaping utility considered html string.
Because the blocks such as wp:search are represented as HTML comments they were ignored.

This PR extends the escaping to attributes of such blocks.

| Considered blocks | Considered Attrs |
| --- | --- |
| wp:search | label, placeholder, buttonText |
| wp:read-more | content |

More attributes, blocks can be included when a discovered during development of new patterns.

**Before:**
```
<!-- wp:search {"label":"","showLabel":false,"placeholder":"Search...","buttonText":"WHAT R U WAITING FOR?"} /-->
```

**After:**
```
<!-- wp:search {"label":"","showLabel":false,"placeholder":"<?php echo esc_html__( 'Search...', 'club' ); ?>","buttonText":"<?php echo esc_html__( 'WHAT R U WAITING FOR?', 'club' ); ?>"} /-->
```

#### Related issue(s):

Tools improvement.

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -1188,7 +1188,8 @@ async function escapePatterns() {
 				rewriter.emitRaw(rawHtml);
 				return;
 			}
-			// escape the strings in blocks without a html such as wp:search
+			// escape the strings in block config (blocks that are represented as comments)
+			// ex: <!-- wp:search {label: "Search"} /-->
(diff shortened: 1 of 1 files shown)


mikachan (approved):
Thanks for this! This works well in my testing, I think we can bring it in and iterate on it when we need to add more attributes, etc.

PR #6287, opened 2022-07-26
Allow multiple themes in 'deploy:theme' and 'deploy:zip' commands

pbking:
The most common use of `npm run deploy:theme THEMESLUG` and `deploy:zip THEMESLUG` is when the `npm run deploy` command fails for some reason during or before those steps have finished.

This results in the need to do those steps manually, one theme at a time.

This change allows the themes to be a comma delimited list.

before:
```
npm run deploy:theme apples
npm run deploy:theme bananas
npm run deploy:zip apples
npm run deploy:zip bananas
```

after:
```
npm run deploy:theme apples,bananas
npm run deploy:zip apples,bananas
```

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -74,12 +74,12 @@ const commands = {
 	"deploy-theme": {
 		helpText: 'This runs "deploy pub <theme>" on the provided list of themes.',
 		additionalArgs: '<array of theme slugs>',
-		run: (args) => deployThemes([args?.[1]])
+		run: (args) => deployThemes(args?.[1].split(','))
 	},
 	"build-com-zip": {
 		helpText: 'Build the production zip file for the specified theme.',
 		additionalArgs: '<theme-slug>',
-		run: (args) => buildComZip([args?.[1]])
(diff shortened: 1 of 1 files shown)


vcanales:
This looks good! How would you feel about also allowing spaces? 

```js
.split(/[ ,]+/)
```

This adds support for space separated slugs. It also has the added benefit of trimming the matching theme names.

```js
> "bananas, apples, other-theme, theme,another,  and another-one".split(/[ ,]+/)
< ["bananas", "apples", "other-theme", "theme", "another", "and", "another-one"] (7)
```

mikachan:
I agree with @vcanales, I think my instinct would be to add spaces between multiple themes when running this, so handling them is probably a good idea. Otherwise, this is ace!

pbking:
> This looks good! How would you feel about also allowing spaces?

Regex just makes it FANCY!  Works for me!

pbking:
Shoot!  I pushed but my machine hung and I didn't notice and I merged before it landed. :/

PR #6290, opened 2022-07-27
Update: Quadrat, Geologist & Zoologist quote block line height

mikachan:
This PR updates the quote block line-height for Quadrat and Geologist + variations. We were previously using a pixel value for the line-height, and this replaces that with a unitless value, as we do for other line-height values.

When I ran the `build:variations` script, the readme's for the variations were updated. This looks correct so I've left the changes in.

#### Related issue(s) / context:
p1658916922462169-slack-C029FM1EH


jffng (approved):
Makes sense and LGTM.

PR #6292, opened 2022-07-27
Refactor core sync utility to merge content locally before pushing to sandbox

### LINKED ISSUE #6288: Automate Core Theme Update Process
We need an automated process to update wpcom themes with changes to core themes.

We currently have a different repository with versioning that differentiates our core themes so that we have the option to make changes faster than the core theme pipeline allows, or changes specific to the platform.

We need to bring in the changes to core themes regularly. While this likely can't be fully automated it can hopefully be a push-button process.

We're operating between the wpcom git repository and the wporg github (or possibly wporg svn) repositories.

pbking:
This change refactors the core sync utility to do the syncing LOCALLY where the conflicts can be edited easily, then pushes the changes to the sandbox before creating a phabricator diff to review and land.

The ideal workflow is a single command: `npm run deploy:core:sync THEMESLUG`

This will pull the revision from the `.pub-svn-revision` file in each of the default theme directories to determine how far back to sync changes.  However at this moment none of the themes have a `.pub-svn-revision` file (it's created during this deploy process) so the revision needs to be supplied.  (I have been using `52221` as that seems to have been the last revision previously synced).

Thus: `npm run deploy:core:sync THEMESLUG 52221`

The `deploy:core:sync` command is a combination of the following commands:

```
node theme-utils.mjs clean-sandbox
node theme-utils.mjs checkout-core-theme THEMESLUG
node theme-utils.mjs sync-core-theme THEMESLUG REVISION
node theme-utils.mjs push-theme-to-sandbox THEMESLUG
node theme-utils.mjs create-core-phabricator-diff THEMESLUG REVISION
(TODO: land diff)
(TODO: deploy theme)
(TODO: build zip file)
```

The command needs to be ran for EACH core theme (there isn't a command that syncs ALL core themes to the latest since each theme should be carefully evaluated and deployed individually).  

Currently the script does NOT land the changes, deploy the changes or builds zip files; those steps will need to be completed manually for this stage.  That will make the workflow:
```
npm run deploy:core:sync THEMESLUG 52221
# (note the DIFF created from the previous step)
npm run deploy:land DIFFID
npm run deploy:theme THEMESLUG
npm run deploy:zip THEMESLUG
```

Using this process I have created the following diffs:

TwentyTwentyTwo: D84933-code
TwentyTwentyOne: D84937-code
TwentyTwenty: D84943-code
TwentyNineteen: D84944-code
TwentySeventeen: D84945-code
TwentySixteen: D84946-code
TwentyFifteen: (Currently there is an error processing this theme's logs that cause a failure.)
TwentyFourteen: D84948-code
TwentyThirteen: D84949-code
TwentyTwelve: D84950-code
TwentyEleven: D84951-code
TwentyTen: D84952-code

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -99,11 +99,6 @@ const commands = {
 		additionalArgs: '<theme-slug> <since-revision>',
 		run: (args) => syncCoreTheme(args?.[1], args?.[2])
 	},
-	"deploy-core-theme": {
-		helpText: 'Given a theme slug and SVN revision, deploy local changes to wpcom via Phabriactor',
-		additionalArgs: '<theme-slug> <since-revision>',
-		run: (args) => deployCoreTheme(args?.[1], args?.[2])
-	},
 	"deploy-sync-core-theme": {
 		helpText: 'Given a theme slug and SVN revision, sync the theme from the specified revision to the latest. This command contains additional prompts and error checking not provided by sync-core-theme.',
 		additionalArgs: '<theme-slug> <since-revision>',
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -81,18 +81,28 @@ const commands = {
 		additionalArgs: '<theme-slug>',
 		run: (args) => buildComZip([args?.[1]])
 	},
+	"checkout-core-theme": {
+		helpText: 'Use SVN to checkout the given core themes from the wpcom SVN repository.',
+		additionalArgs: '<theme-slug>',
+		run: (args) => checkoutCoreTheme(args?.[1])
+	},
 	"pull-core-themes": {
-		helpText: 'Use rsync to copy any changed public CORE theme files from your sandbox to your local machine. CORE themes are any of the Twenty<whatever> themes.',
+		helpText: 'Use rsync to copy all public CORE theme files from your sandbox to your local machine. CORE themes are any of the Twenty<whatever> themes.',
 		run: pullCoreThemes
 	},
 	"push-core-themes": {
-		helpText: 'Use rsync to copy any changed public CORE theme files from your local machine to your sandbox. CORE themes are any of the Twenty<whatever> themes.',
+		helpText: 'Use rsync to copy all public CORE theme files from your local machine to your sandbox. CORE themes are any of the Twenty<whatever> themes.',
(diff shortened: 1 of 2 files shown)


theme-utils.mjs
```diff
@@ -340,11 +355,12 @@ async function deploySyncCoreTheme(theme, sinceRevision) {
```
madhusudhand: Let's add the condition here as well to avoid running steps when theme is not set.
Even though each function return doing nothing, because there isn't a condition on the returned value, it is proceeding to next steps.

```suggestion
if (!theme) {
console.log('Must supply theme to sync and revision to start from');
return;
}
```

theme-utils.mjs
```diff
@@ -81,24 +81,39 @@ const commands = {
...
-		helpText: 'Given a theme slug and SVN revision, sync the theme from the specified revision to the latest. This command is generally run by deploy-sync-core-theme and not by itself.',
+		helpText: 'Given a theme slug and SVN revision, sync the theme from the specified revision to the latest. This requires the core theme to be currently checked out from the wpcom svn repository.',
 		additionalArgs: '<theme-slug> <since-revision>',
 		run: (args) => syncCoreTheme(args?.[1], args?.[2])
 	},
+	"deploy-core-theme": {
+		helpText: 'Given a theme slug and SVN revision, deploy local changes to wpcom via Phabriactor',
+		additionalArgs: '<theme-slug> <since-revision>',
+		run: (args) => deployCoreTheme(args?.[1], args?.[2])
+	},
```
madhusudhand: do we need they anymore?

madhusudhand (commented):
This is fab. Dropped couple of minor comments inline.

mikachan (commented):
Thanks so much for working on this!

I've just tested this by using it to sync and update TT2. Everything worked on my side, I didn't notice any errors and the steps made sense to me. Happy for this to be merged when everyone else is happy 👍

theme-utils.mjs
pbking: Good call.

theme-utils.mjs
pbking: Indeed we do not.  Thank you!

PR #6293, opened 2022-07-27
Rainfall: Try spacing presets

scruffian:
Now that the spacing presets have merged (https://github.com/WordPress/gutenberg/pull/41527) we should start using these in our themes. This is an example of how we could use them in Rainfall...

### DIFF
--- a/rainfall/patterns/contacts.php
+++ b/rainfall/patterns/contacts.php
@@ -5,8 +5,8 @@
  * Categories: featured
  */
 ?>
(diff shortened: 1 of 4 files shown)


mikachan (commented):
This looks like it's working great. Should we replace the rest of the spacing values in Rainfall (or at least the ones using `rem`) so they use a preset?

scruffian:
> Should we replace the rest of the spacing values in Rainfall (or at least the ones using rem) so they use a preset?

Go ahead :)

scruffian:
Thanks @mikachan

PR #6299, opened 2022-07-28
Blockbase: Locked post content blocks 

pbking:
Locked post content blocks (and their containers) so that they can't be easily removed by the user.

I have noticed that the users often edit content templates thinking it's something like the homepage Page and then not see their content on any of the pages or posts.

Locking the post content blocks into those templates may help mitigate that scenario by making it more difficult to remove those blocks.

This change will only effect those themes that don't provide page/post/etc templates.  If we decide this is a thing to do then we should also change those child theme templates. 

Is it helpful to apply to all of our themes?

Before:
image

After:
image

### DIFF
--- a/arbutus/block-templates/page.html
+++ b/arbutus/block-templates/page.html
@@ -3,7 +3,7 @@
 <!-- wp:group {"tagName":"main"} -->
 <main class="wp-block-group"><!-- wp:post-featured-image {"align":"full"} /-->
 
-<!-- wp:post-content {"layout":{"inherit":true}} /--></main>
+<!-- wp:post-content {"layout":{"inherit":true},"lock":{"move":false,"remove":true}} /--></main>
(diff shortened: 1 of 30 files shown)


mikachan (commented):
I really like this idea; I think it would help users to edit the type of templates you have mentioned, while still giving them the flexibility to remove this block if they really intend to.

I vote for adding this to all themes 👍

madhusudhand (commented):
It is useful and avoids accidental removal of the block. I think we can add the same to all child themes, block themes and we can make it a practice in future themes.

mikachan:
I've updated all the Blockbase child themes with the same change.

pbking:
> I've updated all the Blockbase child themes with the same change.

Nice!  Thanks!  I'mma ship this now!

PR #6262, opened 2022-07-21
Club: implementing fluid typography

matiasbenedetto:
#### Changes proposed in this Pull Request:
Club theme: Implement the [fluid type sizes API](https://github.com/WordPress/gutenberg/pull/39529) in theme.json.

:information_source:  Notes about how this new feature works in practice:
- `fluid.min` and `fluid.max`, obviously, works as max and min font sizes. `min` is applied in all the viewports below 768px wide. This value is enlarged progressively when the viewport grows. If the viewport is more that 1600px wide `max` is applied as font size.
- I've tested the API without Gutenberg activated and the `size` key is used as a fallback value. So leveraging the API won't imply a broken site for users not using the Gutenberg plugin. 

This PR is an alternative approach to this one https://github.com/Automattic/themes/pull/6255.
Thanks, @madhusudhand for working on it!

:warning: We still need to add the line height settings so I add this as a draft to discuss the fluid typography implementation.

#### Related issue(s):
https://github.com/Automattic/themes/issues/6213

PR #6302, opened 2022-07-29
Club: implementing footer template part

### LINKED ISSUE #6227: club: Templates: Footer template part
Implement the footer template part based on the designs.

Desktop:
image

Mobile:
image

matiasbenedetto:
Club: implementing footer template part

#### Related issue(s):
Fixes: https://github.com/Automattic/themes/issues/6227

### DIFF
--- a/club/parts/footer.html
+++ b/club/parts/footer.html
@@ -1,5 +1,6 @@
-<!-- wp:columns {"verticalAlignment":null} -->
-<div class="wp-block-columns"><!-- wp:column {"verticalAlignment":"bottom"} -->
+<!-- wp:group -->
+<div class="wp-block-group"><!-- wp:columns -->
+	<div class="wp-block-columns"><!-- wp:column {"verticalAlignment":"bottom"} -->
 	<div class="wp-block-column is-vertically-aligned-bottom"><!-- wp:group {"style":{"spacing":{"blockGap":"0.5rem"}},"layout":{"type":"flex","orientation":"vertical"}} -->
 	<div class="wp-block-group"><!-- wp:site-title {"level":2,"style":{"typography":{"lineHeight":"1.6"}},"fontSize":"default-title"} /-->
 	
@@ -21,5 +22,13 @@
 	<!-- /wp:group --></div>
 	<!-- /wp:column --></div>
 	<!-- /wp:columns -->
-
+	
+	<!-- wp:paragraph -->
+	<p>Powered by WordPress</p>
+	<!-- /wp:paragraph -->
+	
+	<!-- wp:spacer {"height":"1rem"} -->
+	<div style="height:1rem" aria-hidden="true" class="wp-block-spacer"></div>
+	<!-- /wp:spacer --></div>
+	<!-- /wp:group -->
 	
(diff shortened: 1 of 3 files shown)


madhusudhand:
Remove `page.html` from the commit. 
And in general it feels not right to review the template without right typography in place.

madhusudhand (commented):
Remove `page.html` from the commit. 
And in general it feels not right to review the template without right typography in place.

matiasbenedetto:
> And in general it feels not right to review the template without right typography in place.

You can just copy `theme.json` from `club/fluid-typography` branch and paste it into this branch. Then you will see both changes together.

club/parts/footer.html
```diff
@@ -1,11 +1,29 @@
...
+	<!-- /wp:group -->
+	
+	<!-- wp:paragraph -->
+	<p>Powered by WordPress</p>
+	<!-- /wp:paragraph -->
+	
+	<!-- wp:spacer {"height":"1rem"} -->
+	<div style="height:1rem" aria-hidden="true" class="wp-block-spacer"></div>
+	<!-- /wp:spacer --></div>
+	<!-- /wp:group -->
```
mikachan, line 29: The contents of this file should probably be a pattern, as we need to escape and translate all the text content.

mikachan (commented):
The site title and address paragraph aren't showing as flex columns for me:

image

They look fine in the editor, so I think this is possibly a GB bug:

image

If I force that content to stack on the front end, the layout is working great for me at both larger and smaller resolutions.

club/style.css
```diff
@@ -113,6 +113,14 @@ GNU General Public License for more details.
     flex-direction: column;
 }
 
+/*
+ * Footer
+ */ 
+
+ footer .wp-block-site-title a {
+	text-decoration: none;
```
mikachan, line 121: According to the design, it looks like this site title should be italic, which I think we'd need to set using CSS.

matiasbenedetto: We have already added this as a part of another PR: https://github.com/Automattic/themes/pull/6262/files#diff-ae97e73d61d965a9d30bf401f119e47437c0305c922c481000f7fd20aafbfa2eR447

club/parts/footer.html
matiasbenedetto: I think that each user will fill the footer with their own data and not translate this fake address data included in the theme, so adding the translatable strings seems pointless in this case,

matiasbenedetto:
> The site title and address paragraph aren't showing as flex columns for me:
> They look fine in the editor, so I think this is possibly a GB bug

Yep, it's a Gutenberg bug, I reported it here: https://github.com/WordPress/gutenberg/issues/42938 and submitted a potential fix: https://github.com/WordPress/gutenberg/pull/42939

club/parts/footer.html
mikachan: Ah yeah I see, that makes sense. I was thinking of the 'Powered by WordPress' text too but we don't tend to make this translatable anymore either.

mikachan (approved):
This looks good to me!

PR #6303, opened 2022-07-29
Fix/missing tags

andres-blanco:
Fixes missing tags in multiple themes. This will fix a Theme Showcase bug where selecting "Replace homepage" in the "Activate" modal does nothing.

More context paYKcK-1WT-p2 
Related issue https://github.com/Automattic/wp-calypso/issues/65900


madhusudhand (approved):
LGTM!

PR #6305, opened 2022-08-01
Rainfall: address header and footer improvements

### LINKED ISSUE #6245: Rainfall: Header and Footer improvements
Here are some comments related with the header and footer.

1. Both the header and footer template parts should be full width, aligning with the post featured images.

Captura de ecrã 2022-07-15, às 11 06 12
Captura de ecrã 2022-07-15, às 11 08 49

2. The footer text should be uppercase by default, like the header text.

Captura de ecrã 2022-07-15, às 11 10 59

3. You probably haven't yet addressed the mobile menu but wanted to mention it anyway. Is there any issue open in Gutenberg about the customisation of the mobile menu icon? It would be neat to have this be customised "Menu" and "Close" instead of the default hamburger and `X` icons.

Captura de ecrã 2022-07-15, às 11 13 04

4. Lastly, there's some weirdness in the footer on mobile.5. 

Captura de ecrã 2022-07-15, às 11 19 41

mikachan:
This addresses most of the feedback in https://github.com/Automattic/themes/issues/6245. 

I haven't been able to address all of it as I've tried to stick with what's currently available from GB, but let me know if we should manually add anything.

1. Header and footer are now full width
2. Footer text is now uppercase by default (I also made the links a navigation block)
image

3. Styled the mobile menu further:

We are currently able to swap the hamburger menu for text that says 'menu', but there is currently no option for replacing the 'X' with text (issue here: https://github.com/WordPress/gutenberg/issues/42848).

image

This is how the overlay looks now:
image

I haven't included the social menu in the overlay as this currently isn't possible in GB (I don't think?)

I've also styled the current menu item, but it looks like there's a bug with the `current-menu-item` class. Fix here: https://github.com/WordPress/gutenberg/pull/42849

5. Fixed how the footer menu wraps at mobile
image

#### Related issue(s):
https://github.com/Automattic/themes/issues/6245


scruffian:
I added a commit to remove the link color from the header group as it was breaking the overlay colors.

mikachan:
> I added a commit to remove the link color from the header group as it was breaking the overlay colors.

Nice, thanks!

beafialho:
Thank you! Header and footer alignments and letter case look good.

> I've also styled the current menu item, but it looks like there's a bug with the current-menu-item class.

I notice there's a tiny misplacement of the current menu item, is that the bug?

Captura de ecrã 2022-08-03, às 14 51 14

> We are currently able to swap the hamburger menu for text that says 'menu', but there is currently no option for replacing the 'X' with text (issue here: https://github.com/WordPress/gutenberg/issues/42848).

Thanks for opening that one!

> I haven't included the social menu in the overlay as this currently isn't possible in GB (I don't think?)

Ok, let's leave it like that for now.

I'm not sure this is the step to address this but the navigation links currently have no hover states. Those are specified [here](https://www.figma.com/file/oz0a2czNn0UNt9onrh8T5a/Rainfall?node-id=107%3A906).

beafialho (approved):

mikachan:
Thanks @beafialho!

> I notice there's a tiny misplacement of the current menu item, is that the bug?

I think this is due to the font-weight. We could probably fix it by specifying a different line-height for the current item.. I can take a proper look into it.

> I'm not sure this is the step to address this but the navigation links currently have no hover states.

Ah yeah, we can address this next!

PR #6312, opened 2022-08-03
Archeo: Try new root padding alignments

scruffian:
This tries implementing the root padding alignment rules proposed in https://github.com/WordPress/gutenberg/pull/42085. This means we can remove the alignment CSS.

## Comparison
| Before: | After: |
|-----------|-----------|
| before | after |

#### Related issue(s):
https://github.com/WordPress/gutenberg/pull/42085

### DIFF
--- a/archeo/patterns/layered-images-with-headline.php
+++ b/archeo/patterns/layered-images-with-headline.php
@@ -6,8 +6,8 @@
  */
 ?>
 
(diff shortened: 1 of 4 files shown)


mikachan (commented):
It looks like the new root padding alignments are working in most cases, apart from blocks nested in a full-width group block:

This PR:
image

Trunk (with alignment CSS & no root padding set):
image

scruffian:
I'm not sure that the behaviour in trunk is desirable anyway, I think this is a fair compromise.

mikachan:
Ah OK, I guess I was basing the intended gap on the wording in this test markup ('should have a gap between it and the edge'), and it seemed to be one of the biggest differences from the previous behaviour. 

If these current gaps aren't ideal (and I would assume 'full width' should have no gaps between the container and the edge anyway) then the root padding seems to be working well 👍

MaggieCabrera:
It looks like this will need some adjustments in Archeo, the text in the home page is not aligned anymore:

Before | After
--- | ---
Screenshot 2022-08-05 at 10 21 19 | Screenshot 2022-08-05 at 10 21 14

scruffian:
Good catch, thanks. I've zeroed some of the paddings here so that they match:
Screenshot 2022-08-05 at 15 08 32

vcanales (approved):
This LGTM 🚢

PR #6313, opened 2022-08-04
Remove nested page-list blocks

### LINKED ISSUE #6163: Remove any inner blocks from the `navigation` block
We should remove the `page-list` block (and any other inner blocks) as an inner block of the `navigation` block, prior to this GB PR being merged: https://github.com/WordPress/gutenberg/pull/42182.

> This change means that Themes should not be including inner blocks in their Navigation blocks in their templates/parts. The only exception to this would be if they wanted to opt out of some of the current / future menu picking optimisations and automations provided by the block (e.g. carrying across menus on Theme switch).

This only applies to block themes that have inner blocks inside the navigation block. These are the ones I can find, but it's worth double-checking:

- [ ] Archeo
- [ ] Livro
- [ ] Remote
- [ ] Skatepark
- [ ] Stewart

mikachan:
This PR removes any nested `page-list` blocks from existing `navigation` blocks. See https://github.com/Automattic/themes/issues/6163 for full context.

#### Related issue(s):
Closes https://github.com/Automattic/themes/issues/6163


pbking (approved):
Changes check out. 🚢

PR #6314, opened 2022-08-04
lock post-content block by default

### LINKED ISSUE #6311: Lock Postcontent Blocks
Just as [we have done for Blockbase and its kin](https://github.com/Automattic/themes/pull/6299) we should also do for all of our Block Themes.

`"lock":{"move":false,"remove":true}`

Add the block attributes to the post content block so that users cannot remove the block from the template (without first unlocking the block).

madhusudhand:

Locked post content blocks (and their containers) in all the block theme, so that they can't be easily removed by the user.

This is an extension of changes done here https://github.com/Automattic/themes/pull/6299 for blockbase and its children.

#### Related issue(s):

Fixes: #6311

### DIFF
--- a/arbutus/block-templates/page.html
+++ b/arbutus/block-templates/page.html
@@ -1,6 +1,6 @@
 <!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
-<!-- wp:group {"tagName":"main"} -->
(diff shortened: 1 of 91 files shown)


pbking:
Rebased and added just a few locks that were missing.  This checks out. 🚢

PR #6315, opened 2022-08-04
Removal of Varese from trunk

onuro:
There was a leftover file from the initial accediental commit of the Varese theme. This commit removes it to start over the Varese commit process


mikachan (approved):
Thanks!

PR #6321, opened 2022-08-05
Blockbase: fix font-size in Comments Title Block

ghost:


pbking (approved):
I kind of went back-and-forth on this one but I agree that it's a good change and see why there is the need for it.  🚢

PR #6324, opened 2022-08-05
add timeout to deployment script to kill the unresponsive process

madhusudhand:

**Change 1:**

When deploying multiple themes (more than 10) together, process sometimes stops at `deploy pub <theme>` because of the multiple connects and disconnects into sandbox.
Because of this, user has to kill the deployment process and manually find out what part of the process is remaining and retry again.

It happened multiple times when deploying large diffs such as D85430-code

This change sets a max timeout of `2 min` for any command to run on sandbox, and kills if it is not done within the time.
And it continues the deployment process with next set of themes and finally displays all the list of failed items, so that user can retry process only for the failed ones.

**Change 2:**

Introduced a progress bar to indicate the status.

image

**Change 3:**

Updated the version bump script to update package-lock.json (missed updating during workspace migration)

#### Testing:

Verified the changes by replacing `deploy pub <theme>` with `sleep 3`, and reducing the timeout to `2 sec`.
It is working great! Child process is killed and the theme listed at the end in the failed list.

#### Related issue(s):

Tools improvement.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,6 +28,7 @@
 				"octokit": "^2.0.3",
 				"open": "^8.4.0",
 				"parse5-html-rewriting-stream": "^7.0.0",
+				"string-progressbar": "^1.0.4",
 				"table": "^6.8.0"
 			},
 			"engines": {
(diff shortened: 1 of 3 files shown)


pbking:
This is so darn cool.  Love it and it looks good!  I can't test it until I see it deploy but the change looks good.  If you happened to use this on the last deploy I'm happy for this to go ahead and come in @madhusudhand.  Otherwise I'm keen to use this change the next time I deploy and if it's as successful as it looks bring it in then.

Oh, maybe I jumped the gun.  This is only set to 'in progress' and maybe you aren't done yet.  I'll wait patiently. :)

madhusudhand:
This is now Ready 🚀 

Verified the normal deployment workflow and is working.
For the timeout case, replaced `deploy pub <theme>` with `sleep 3`, and reduced the timeout to `2 sec`
It is working great! Child process is killed and the theme listed at the end in the failed list.

And since the last change, switched to a different progress bar because the earlier one overlapping with deploy stdout.
Current progress bar is simpler and cleaner.

Update:

Verified with D85617-code (24 themes). All good.

image

pbking (approved):
Good changes!  I see no reason not to bring this in now; The code looks solid.  I'm looking forward to using it on the next deploy.

PR #6327, opened 2022-08-05
#6323: Lodestar: Fix list alignment on homepage

### LINKED ISSUE #6323: Lodestar theme: List block on homepage appears center aligned by default
### Quick summary

On home page List block appears center aligned by default when using the Lodestar theme.

### Steps to reproduce

1. On a simple or AT site, activate the Lodestart theme.
2. On the home page, add a list block with a few list items.

The list should have appeared left-aligned.

### What actually happened

The list appears center aligned.

### Context

A user has reported this here: #5440049-zd. I was also able to replicate this on a test site. I was able to replicate this issue on both simple and AT sites and it happens only on the homepage. The list block appears left aligned on non-homepages. If you set an existing homepage with list block as a regular page, the list appears left-aligned. If you set a regular page with list block as the homepage, the list appears center-aligned.

Simple, Atomic

### Theme-specific issue?

Yes, this happens on Lodestar theme.

I have tested this on Google Chrome (Version 103.0.5060.134 (Official Build) (x86_64)), Safari (Version 15.6 (17613.3.9.1.5)) and Firefox (103.0.1 (64-bit)) and was able to replciate the issue on these browsers.

### Reproducibility

Consistent

### Severity

One

### Available workarounds?

Yes, easy to implement

### Workaround details

The following CSS helps.

.home .lodestar-intro .entry-content ul {
	text-align: left;
}

bobmatyas:
#### Editor 

Screenshot on 2022-08-05 at 12-54-40

### Before

Screenshot on 2022-08-05 at 12-35-18

#### After

Screenshot on 2022-08-05 at 12-35-48

### Related issue(s):

Fixes https://github.com/Automattic/themes/issues/6323

### DIFF
--- a/lodestar/style.css
+++ b/lodestar/style.css
@@ -1219,7 +1219,6 @@ body:not(.logged-in) .lodestar-panel .jetpack-testimonial .entry-header {
 	color: #555;
 	font-size: 20px;
 	font-size: 1.25rem;
-	text-align: center;
 }
 
 .lodestar-intro .entry-content > *:last-child {


lodestar/style.css
```diff
@@ -1219,7 +1219,6 @@ body:not(.logged-in) .lodestar-panel .jetpack-testimonial .entry-header {
 	color: #555;
 	font-size: 20px;
 	font-size: 1.25rem;
-	text-align: center;
```
mikachan, line 1222: I think it would be best to target just list elements here, rather than removing the center alignment for all homepage content, otherwise, we lose the center alignment for all content. e.g. something like this:

```
.lodestar-intro .entry-content ul,
.lodestar-intro .entry-content ol {
	text-align: left;
}
```

bobmatyas: @mikachan I removed the `text-align: center` because it made the editor view reflect what is happening on the frontend. If we just target the list, it looks like this in the editor if a paragraph is left-aligned:

Screenshot on 2022-08-08 at 10-02-16

but the frontend then shows:

Screenshot on 2022-08-08 at 10-01-57

If that is okay, I'm happy to change it.

pbking: If a page is NOT used as the homepage all of the content is left-aligned (like it shows in the editor).  When used as a homepage the alignment changes to center ALL the content, which is quite different and seems like a bug.

As far as I can tell this fix brings a page's alignment into uh... alignment (pun intended) with what it shows everywhere else.  Thus it doesn't seem to be an issue with just the list items, but with all content.  Therefore the change that's currently made seems right to me.

pbking (approved):

lodestar/style.css
mikachan: Ah OK, that sounds good to me and makes sense. I was thinking that un-centering the home page content would cause more issues, but aligning all the alignments 😁 sounds better.

PR #6328, opened 2022-08-05
Club: adding custom styles for navigation block

### LINKED ISSUE #6218: club: Styles: Navigation
Implement navigation following the mockups

image

image

matiasbenedetto:
Club: adding custom styles for the navigation block

#### Related issue(s):
Fixes: https://github.com/Automattic/themes/issues/6218

### DIFF
--- a/club/parts/header.html
+++ b/club/parts/header.html
@@ -8,6 +8,8 @@
 	<!-- /wp:group --></div>
 	<!-- /wp:group -->
 	
-	<!-- wp:navigation {"__unstableLocation":"primary","overlayBackgroundColor":"background","overlayTextColor":"foreground","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /--></div>
+	<!-- wp:navigation {__unstableLocation":"primary","overlayBackgroundColor":"background","overlayTextColor":"foreground","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"left","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /--></div>
 	<!-- /wp:group --></div>
 	<!-- /wp:group -->
+
+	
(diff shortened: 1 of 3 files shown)


mikachan (commented):
The responsive menu looks like this for me with these changes, where the items are being cut off on the left side:

image

I think this is partly related to the `width: 100%` on the individual nav items.

matiasbenedetto:
thanks, @mikachan for the review!

It seems to be working fine now:
image

club/style.css
```diff
@@ -172,6 +172,40 @@ a {
 	text-decoration: none;
  }
 
+/*
+ * Navigation
+ */
+ .wp-block-navigation__responsive-container.is-menu-open .wp-block-navigation__responsive-container-content .wp-block-navigation__container {
+	align-items: flex-start;
```
mikachan: Actually, now I've taken another look at this, I don't think we need any of these `flex-start`s (lines 179, 183 and 197). I think we can change the alignment on the navigation block itself (`justifyContent`) to left aligned, and as long as the row container block is set to space between (which it currently is), the menu will appear aligned to the right on larger resolutions. e.g. change the navigation block to something like this:

```
<!-- wp:navigation {"overlayBackgroundColor":"background","overlayTextColor":"foreground","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"left","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /-->
```

matiasbenedetto: Thanks for noticing this @mikachan. I wasn't aware of the ability of the navigation block to be configured in such a way. I removed that CSS and replaced it with the block configuration.

pbking:
🤔 The latest of this is showing a white background on the responsive navigation for me:

image

club/parts/header.html
```diff
@@ -8,6 +8,8 @@
 	<!-- /wp:group --></div>
 	<!-- /wp:group -->
 	
-	<!-- wp:navigation {"__unstableLocation":"primary","overlayBackgroundColor":"background","overlayTextColor":"foreground","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /--></div>
+	<!-- wp:navigation {__unstableLocation":"primary","overlayBackgroundColor":"background","overlayTextColor":"foreground","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"left","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /--></div>
```
mikachan: It looks like there's a missing `"` from the beginning of "__unstableLocation".

matiasbenedetto: Yep, thanks for notice that!

matiasbenedetto:
It should be working like this:

Desktop:
image


Mobile:
image
image

mikachan (approved):
LGTM! 🚢

PR #6331, opened 2022-08-06
Spearhead Blocks initial commit

onuro:
Initial commit for the Spearhead blocks, the FSE blocks version of the original Spearhead theme.

screenshot


spearheadblocks/style.scss
```diff
@@ -0,0 +1,34 @@
+/*
+Theme Name: Spearhead Blocks
```
pbking, line 2: You can remove this .scss file.  It doesn't look like there will be any CSS compilation for this theme so I don't think it will be needed at all.

spearheadblocks/style.css
```diff
@@ -0,0 +1,32 @@
...
+Author: Automattic
+Author URI: https://automattic.com
+Description: Spearhead Blocks is the Full Site Editing capable, block based version of the original Spearhead theme.
+Requires at least: 5.8
+Tested up to: 5.9
+Requires PHP: 5.7
+Version: 0.0.1
+License: GNU General Public License v2 or later
+License URI: https://raw.githubusercontent.com/Automattic/themes/trunk/LICENSE
+Template: blockbase
```
pbking, line 13: Consider using `blank-canvas` as a parent instead of Blockbase.  I gave it a quick test and it seems to have the same design.  Blank Canvas has less "baggage" than Blockbase.  

We haven't shipped Blank Canvas as a theme yet (thus it couldn't be used as a parent in production) but if there are viable usecases for it (such as this theme) it would be worth considering.

spearheadblocks/styles/midnight.json
```diff
@@ -0,0 +1,52 @@
...
+					"name": "Background"
+				}
+			]
+		}
+	},
+	"styles": {
+		"color": {
+			"background": "var(--wp--preset--color--background)",
+			"text": "var(--wp--preset--color--foreground)"
+		},
```
pbking, line 34: You could set this in theme.json and leave it out of the variation.

spearheadblocks/style.scss
```diff
@@ -0,0 +1,34 @@
...
+Author URI: https://automattic.com
+Description: Spearhead Blocks is the Full Site Editing capable, block based version of the original Spearhead theme.
+Requires at least: 5.8
+Tested up to: 5.9
+Requires PHP: 5.7
+Version: 0.0.1
+License: GNU General Public License v2 or later
+License URI: https://raw.githubusercontent.com/Automattic/themes/trunk/LICENSE
+Template: blockbase
+Text Domain: spearheadblocks
```
pbking, line 14: Convention tells us that this slug should be spearhead-blocks (as well as the folder name, etc)

pbking:
A really good version of Spearhead!

There are a few minor differences between Spearhead and Spearhead blocks I've noticed.

Margins for buttons and font size of buttons seems to be a bit different:

Spearhead:
image

Spearhead Blocks (current):
image

The page Title size is different and the default font size in the original I believe is a bit larger than what the block version has:

Spearhead:
image

Spearhead Blocks (current):
image

Top padding is higher:

Spearhead:
image

Spearhead Blocks:
image

pbking:
I thought the "dark mode" was one of Spearheads nicest design features. One thing you might want to consider; Instead of including a "dark mode" as an alternative style variation include the "dark colors" in the palette and mapping them with a media query:

image

```
  body {
    --wp--preset--color--primary: var(--wp--preset--color--primary-dark);
    --wp--preset--color--secondary: var(--wp--preset--color--secondary-dark);
    --wp--preset--color--foreground: var(--wp--preset--color--foreground-dark);
    --wp--preset--color--background: var(--wp--preset--color--background-dark);
  }
}
```

onuro:
Thank you so much for these inputs @pbking. Updating with fixes to those shortly

pbking:
Closing this one in deference to #6435.

Issue #3174, opened 2021-02-03
Dyad 2: WooCommerce Cart and Checkout Pages Display Incorrectly on Mobile Displays


edequalsawesome:
## Steps to replicate

* Activate Dyad 2 on a site that has WooCommerce active
* Add products to your cart on a mobile device
* View the Cart and Checkout pages

## Result

The styling of the Cart and Checkout pages is incorrect, making the elements appear poorly:

![Dyad 2 - WooCommerce Cart Bug - Before](https://user-images.githubusercontent.com/4000628/106816060-cee63480-6642-11eb-9a5e-549ded7ee50b.png)

![Dyad 2 - WooCommerce Checkout Bug - Before](https://user-images.githubusercontent.com/4000628/106816078-d4dc1580-6642-11eb-995b-b80b59281cc6.png)

## Expected

The elements would be resized to better accommodate mobile and tablet screen sizes

![Dyad 2 - WooCommerce Cart Bug - After](https://user-images.githubusercontent.com/4000628/106816124-eb826c80-6642-11eb-8fb2-a1445ae7f8e5.png)

![Dyad 2 - WooCommerce Checkout Bug - After](https://user-images.githubusercontent.com/4000628/106816139-efae8a00-6642-11eb-9fd4-63021c4b5b22.png)

---

The following CSS will help the WooCommerce Cart page appear laid out a bit better on mobile and tablet devices:

```
/* -- fix cart page layout for mobile and tablet devices | XXXXXX-hc/zen eD */
@media screen and (max-width: 960px) {
	/* remove stray : before product thumbnail */
	table.shop_table_responsive tr td.product-thumbnail:before {
		display: none;
	}
	
	/* make product images full-width */
	table.shop_table .product-thumbnail img {
		width: 100%;
	}
	
	/* make coupon section full-width, add margin */
	table.shop_table td.actions .coupon {
		width: 100%;
		margin-top: 10px;
		margin-bottom: 10px;
	}
	
	/* make coupon input field and button full-width */
	table.shop_table td.actions .coupon input[type="text"],
	table.shop_table td.actions .coupon .button {
		width: 100%;
	}
	
	/* add top and bottom margins to apply coupon button */
	table.shop_table td.actions .coupon .button {
		margin-top: 10px;
		margin-bottom: 10px;
	}
	
	/* make update cart button full-width, unset float, add bottom margin */
	table.shop_table td.actions .button {
		width: 100%;
		float: unset;
		margin-bottom: 10px;
[...]

PR #6333, opened 2022-08-07
Dyad 2/fix/cart and checkout page mobile layout

gwensmithx:

Make elements on cart and checkout pages on mobile/tablet full-width. They are currently spaced out very awkwardly.

Updated woocommerce.css under Cart and Checkout sections to add additional CSS to make these changes (CSS is in the original GH issue).

I didn't see a SASS directory for this theme which is why I added this CSS directly to the .css file.

Added lines 781 - 794 and 964 - 1007

#### Related issue(s):
https://github.com/Automattic/themes/issues/3174

### DIFF
--- a/club/style.css
+++ b/club/style.css
@@ -7,7 +7,7 @@ Description: Club is a funky, bright theme with vibrant colors and a quirky aest
 Requires at least: 5.8
 Tested up to: 6.0
 Requires PHP: 5.7
-Version: 0.0.16
+Version: 0.0.17
 License: GNU General Public License v2 or later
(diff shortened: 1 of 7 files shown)


matiasbenedetto (changes requested):
Hi @gwensmithx :wave: , In its current state, it's difficult to review this PR for 2 reasons:
- It seems to include changes to unexpected files. Please, rebase the PR to the latest trunk version.
- The file that apparently you want to modify in this PR (`dyad-2/woocommerce.css`) has thousands of changes because of the format/code-styling change. Please just commit the change you want to introduce and not the code styling change for the entire file.
To get a proper review I think you would need to address these problems.

PR #6335, opened 2022-08-08
Club: removing custom font size

matiasbenedetto:
In this PR I'm changing the Club theme font sizes in this way:

About custom custom `default-title` and `large`:
- Removing the custom font size called `default-title`.
- The former `default-title` font size is now the `large` font size in the font sizes array from `theme.json`.
- The former `large` font size (not used in any of the theme elements) is no longer available because it was replaced by the previous changes.

About ad-hoc `default` and `medium`:
- Removing the ad-hoc definition of the default font size using clamp.
- Replace the `medium` size definition (not used in any of the theme elements) for the default size.
- Set the `medium` font size as the default for the site.
- Set the `min` size of `medium` to `1rem` instead of `0.875rem` to avoid repeated font sizes and this Gutenberg bug: https://github.com/WordPress/gutenberg/issues/42683  

Change the footer paragraphs to make the theme use a font size definition defined in the default font sizes array instead of a custom font size.

#### Why?
Because if we use a custom font size for these elements and the user switches the theme those elements will lose the intended hierarchy in terms of font size and it will return to a default font size.


mikachan (approved):
This LGTM!

Issue #4628, opened 2021-09-16
Pique theme is not adding the default color palette css to the editor


glendaviesnz:
## Steps to replicate

- Add an image block with the Pique theme activated. 
- Try to set the font colour of the image caption using the theme palette - notice that the colour is not shown in the editor, but does show in the front end.

![theme-issue](https://user-images.githubusercontent.com/3629020/133694557-1b73464c-9b40-48e2-969a-7ecdab587b09.png)

## Result

Colours applied from theme default to image block captions do not show in the editor

## Expected

Colours applied from theme default to image block captions should show in the editor

PR #6342, opened 2022-08-08
#4628: Pique is not adding the default color palette css to the editor

Gustavo-Hilario:

Adding CSS code for default theme colors to be applied on the page/post editor: `editor-blocks.css`. Whenever we try to apply one of the theme's default colors to text using the `highlight` option, the color change is not getting applied to the elements. However, when checking the published page, we can see that the color change does work.

<table>
    <thead>
      <tr>
        <th>Before</th>
        <th>After</th>
    </tr>
    </thead>
    <tbody>
      <tr>
        <td></td>
        <td></td>
      </tr>
    </tbody>
  </table>

#### Related issue(s):

Pique theme is not adding the default color palette css to the editor #4628

### DIFF
--- a/pique/assets/css/editor-blocks.css
+++ b/pique/assets/css/editor-blocks.css
@@ -491,3 +491,63 @@ p.has-drop-cap:not(:focus)::first-letter {
 	letter-spacing: 0.05rem;
 	text-transform: uppercase;
 }
+
+/*--------------------------------------------------------------
+7.0 Blocks - Colors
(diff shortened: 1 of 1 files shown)


pbking (approved):
Yup, looks like oversight for the original build out.  I'm not sure why the `blocks.css` isn't just used for the editor which would include those colors but I suppose there are `reasons`.  This change fixes what ails.

PR #6344, opened 2022-08-08
#66356: Lodestar: Home page: Full width blocks are not full width

### LINKED ISSUE #6352: Lodestar: Home page: Full width blocks are not full width
### Quick summary

A page with full-width blocks won't show up as full-width when set as the home page.

### Steps to reproduce

1. Activate `Lodestar` theme.
2. Create a new page with full-width blocks. It can be a Cover, Group, Columns, or any other.
3. Set this page to be the home page in the site's Customizer.
4. Check the published page.

Home page content won't show up as full width.

I expected a similar full-width layout as we see in the page editor.

### What actually happened

The page content is within a narrower frame; therefore, not full width as set in the page editor.

Markup on 2022-08-08 at 14:24:00

The `max-width` property from this code is causing this issue.

```
.wrap {
    margin-left: auto;
    margin-right: auto;
    max-width: 1200px;
    max-width: calc(1200px - 1em);
    padding-left: 2em;
    padding-right: 2em;
}
```

### Context

Customer Report:

- [ ] 5458229-zen

Simple, Atomic

### Theme-specific issue?

Theme: Lodestar

_No response_

### Reproducibility

Consistent

### Severity

_No response_

### Available workarounds?

Yes, easy to implement

### Workaround details

CSS code should help here:

```
.home .site-main > .wrap {
  max-width: 100%;
  padding-left:0;
  padding-right: 0;
}

```

Gustavo-Hilario:

Removing a CSS selector that was preventing full-width blocks from being full width on the Lodestar home page.

---
<table>
    <thead>
      <tr>
        <th>Before</th>
        <th>After</th>
    </tr>
    </thead>
    <tbody>
      <tr>
        <td></td>
        <td></td>
      </tr>
    </tbody>
  </table>

#### Related issue(s):

Lodestar: Home page: Full width blocks are not full width Automattic/themes#6352

Fixes: https://github.com/Automattic/themes/issues/6352

### DIFF
--- a/lodestar/assets/css/blocks.css
+++ b/lodestar/assets/css/blocks.css
@@ -45,7 +45,8 @@ body:not(.has-sidebar) .site-content {
 }
 
 body:not(.has-sidebar) .alignfull,
-body:not(.has-sidebar).lodestar-front-page .lodestar-panel:not(.two-column) .alignfull {
+body:not(.has-sidebar).lodestar-front-page .lodestar-panel:not(.two-column) .alignfull, 
+body:not(.has-sidebar).lodestar-front-page .lodestar-intro:not(.two-column) .alignfull {
 	margin-left: calc(50% - 50vw);
 	margin-right: calc(50% - 50vw);
 	max-width: 1000%;
--- a/lodestar/assets/css/blocks.css
+++ b/lodestar/assets/css/blocks.css
@@ -44,7 +44,7 @@ body:not(.has-sidebar) .site-content {
 	}
 }
 
-body:not(.has-sidebar):not(.lodestar-front-page) .alignfull,
+body:not(.has-sidebar) .alignfull,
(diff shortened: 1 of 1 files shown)


pbking:
I see that the bit removed was originally added [here on purpose](https://github.com/Automattic/themes/pull/326).  Does the original issue (wide and full elements in two-column layouts) still apply?  Is it appropriately dealt with in another way?

pbking:
Yup, can confirm that the original reason that was applied is still valid:
image

It only seems to matter if it's using 2 columns so maybe something specific to the `.lodestar-panel two-column` classes (or something) is necessary to remain.

Gustavo-Hilario:
Thank you for the heads-up here, @pbking .

The code was allowing full-width content on all Lodestar panels, except the home page panel that has the `lodestar-intro` class, not `lodestar-panel`. I've added the code that will help with full-width content on the Lodestar home page.

pbking (approved):
👍 Works as intended with no side effects I could discern.

PR #6345, opened 2022-08-08
Club: Search page template

matiasbenedetto:
Implement Club theme Search page template

image


club/templates/search.html
```diff
@@ -1,47 +1,13 @@
 <!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
-<!-- wp:group {"layout":{"inherit":true}} -->
-<div class="wp-block-group">
-	<!-- wp:search /-->
+<!-- wp:group {"layout":{"inherit":false}} -->
+<div class="wp-block-group"><!-- wp:search /-->
```
madhusudhand: Design suggest to use a placeholder instead of label and a different text for the button. The search from 404 page could be re-used by extracting to a hidden pattern.

Design:
image

frontend:
image

madhusudhand (commented):
Dropped a minor inline comment.
Everything else LGTM!

club/templates/search.html
matiasbenedetto: Thanks for your input. Now it is looking like this:

image

madhusudhand (approved):
LGTM!

PR #6348, opened 2022-08-09
Club: implement button styles for interactive states in the read more block button

matiasbenedetto:
Club: implement button styles for interactive states in the read more block button

image

### DIFF
--- a/club/style.css
+++ b/club/style.css
@@ -47,6 +47,11 @@ a {
  * https://github.com/WordPress/gutenberg/issues/27075
  */
 
+ .wp-block-read-more:is(:hover, :focus) {
+	background-color: var(--wp--preset--color--foreground);
+	color: var(--wp--preset--color--background);
+ }
+
 :is(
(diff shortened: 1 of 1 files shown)


club/style.css
```diff
@@ -63,15 +68,7 @@ a {
...
-}
-
-:is(
-	.wp-block-search__button,
-	.wp-block-file .wp-block-file__button,
-	.wp-block-button__link,
-	.wp-elements-button,
+	.wp-block-read-more,
 ):where(:active) {
 	border-style: dotted;
```
mikachan, line 73: Only a small detail, but I think this should be `dashed` rather than `dotted`, when looking at the comps.

mikachan (commented):
This is looking great, I've just left one other comment.

club/style.css
matiasbenedetto: good catch! Change ready!

mikachan (approved):
Looks good to me!

Issue #6232, opened 2022-07-14
Classic themes: Padding issue on <mark> element when using inline colors(part 2)


Robertght:

cuemarie:
Encountered here: 6468992-zd-woothemes

This user was using a third-party classic theme, CoverNews by AF themes. Provided them with the CSS snippet above.

I don't think this is really a themes specific issue, though, so also transferring back to `/wp-calypso` and will take a look when I triage next - this may need to be reported in core.

PR #6350, opened 2022-08-09
#6232: Publication: Fix padding on inline text color

bobmatyas:
Fixes inline text color issue in Publication and Shoreditch

##### Before

Screenshot on 2022-08-09 at 11-30-23

##### After

Screenshot on 2022-08-09 at 11-33-15

##### Before

Screenshot on 2022-08-09 at 11-40-35

##### After

Screenshot on 2022-08-09 at 11-43-30

### Related issue(s):

Addresses https://github.com/Automattic/themes/issues/7164 for Publication and Shoreditch.

### DIFF
--- a/publication/style.css
+++ b/publication/style.css
@@ -439,7 +439,7 @@ acronym {
 	border-bottom: 1px dotted #222;
 	cursor: help;
 }
-mark,
+mark:not(.has-inline-color),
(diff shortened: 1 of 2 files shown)


madhusudhand (approved):
LGTM!

PR #6354, opened 2022-08-09
Removed group block wrapping query pagination

### LINKED ISSUE #6261: Blockbase: Can not duplicate query pagination in index template due to invalid nesting
## Steps to replicate

1. I activated the theme.
2. Opened the site editor, which presented the index template.
3. I expanded the list view and selected the query pagination block.
4. I selected the block toolbar. But I was unable to find an option to duplicate the block.

My goal was to have one query pagination at the top and one at the bottom of the post lists.

-I can add a new query pagination, but it would have been faster for me to duplicate the block
-I can also duplicate the group with the query pagination inside. The limitation is only for the query pagination block itself.

## Result
I was unable to find an option to duplicate the block.

## Expected

I expected to be able to duplicate the block.

* User report: [shortlink to chat/forum thread/ticket, ie 12345-hc]
* Follow-up ticket: [shortlink to ticket, ie 12345-zen]

------

I believe this is because the query pagination is inside the group:

```
	<div class="wp-block-group">
	<!-- wp:query-pagination -->
		<!-- wp:query-pagination-previous /-->

		<!-- wp:query-pagination-next /-->
	<!-- /wp:query-pagination -->
	</div>
	<!-- /wp:group -->
```

Placing the query pagination inside the group is not valid: Its only parent block is the query.
You can confirm this by trying to add a new group and adding a query pagination block inside it,
and by viewing [ block.json ](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/query-pagination/block.json#L7)

pbking:
Removed group block wrapping query pagination which was causing problems.

Fixes #6261

### DIFF
--- a/blockbase/templates/index.html
+++ b/blockbase/templates/index.html
@@ -15,8 +15,6 @@
 </div>
 <!-- /wp:group -->
 <!-- /wp:post-template -->
-<!-- wp:group {"layout":{"inherit":true}} -->
-	<div class="wp-block-group">
 	<!-- wp:query-pagination -->
 		<!-- wp:query-pagination-previous /-->
 
@@ -24,8 +22,6 @@
 
 		<!-- wp:query-pagination-next /-->
 	<!-- /wp:query-pagination -->
-	</div>
-	<!-- /wp:group -->
 </main>
 <!-- /wp:query -->
 


madhusudhand (approved):
LGTM!

PR #6358, opened 2022-08-09
Try using media query for Livro color schemes

pbking:
After thinking about the [same potential in Spearhead](https://github.com/Automattic/themes/pull/6331#issuecomment-1208412603) I gave using media queries a try for enabling dark/light mode on Livro.

The result is a palette that's 6 items (only 3 of the palette colors currently in the theme are actually useful, X 2).

image

There's a custom "routing variable" in theme.json (`--wp--custom--color--foreground`, etc) that we can point either towards the light or dark palette colors.  (`--wp--preset--color--foreground-dark`, etc) 

This then changes both the editor and view to use the dark or light colors depending on the user preferences.
image
image

Additionally, perhaps it would be possible to include style variations that permitted only light mode or dark modes.

### DIFF
--- a/livro/patterns/header-title-tagline-social.php
+++ b/livro/patterns/header-title-tagline-social.php
@@ -16,7 +16,7 @@
 <!-- wp:site-tagline {"textAlign":"center","fontSize":"x-small"} /--></div>
 <!-- /wp:group -->
 
-<!-- wp:social-links {"iconColor":"foreground","iconColorValue":"var(--wp--preset--color--foreground)","iconBackgroundColor":"background","iconBackgroundColorValue":"var(--wp--preset--color--background)","layout":{"type":"flex","justifyContent":"right"}} -->
+<!-- wp:social-links {"iconColor":"foreground","iconColorValue":"var(--wp--custom--color--foreground)","iconBackgroundColor":"background","iconBackgroundColorValue":"var(--wp--custom--color--background)","layout":{"type":"flex","justifyContent":"right"}} -->
 <ul class="wp-block-social-links has-icon-color has-icon-background-color"><!-- wp:social-link {"url":"#","service":"twitter"} /--></ul>
 <!-- /wp:social-links --></div>
 <!-- /wp:group -->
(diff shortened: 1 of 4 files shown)


scruffian:
This is cool. I'd love to see a PR for it in Gutenberg

pbking:
>  I'd love to see a PR for it in Gutenberg

How would that work?  Could palette color definitions be designated "dark" or "regular" and share the same slug?  How would that resolve in the UI?

It doesn't feel jankey to me (UI-wise or configuration-wise) doing it in the theme.  But if there were a standard mechanism provided in Gutenberg that could make the effect a lot more widespread.

madhusudhand:
This is working great for the default definitions.

About the color palette showing both light and dark colors, I am not sure if they are helpful or confusion.
When a paragraph is given a background color as `secondary-dark` it always stays the same irrespective of light or dark scheme, because it is given the preset variable.

```
.has-secondary-dark-background-color {
    background-color: var(--wp--preset--color--secondary-dark) !important;
}
```

Just to experiment, I tried a reverse approach to reset preset-variables instead.

```
"custom": {
	"color": {
		"secondary-dark": "#ffffff"
	},
}
```

```
  body.editor-styles-wrapper,
  body {
    --wp--preset--color--secondary: var(--wp--preset--color--secondary-dark);
  }
}
```

Theme works, but the theme/style.css media query cannot control color palette. It always shows light colors.

image

**For now current PR changes works best!**

### Further thoughts on Gutenberg 

It would become cool feature if the foundations are created in Gutenberg. In simple terms I am thinking of introducing a new property `colorDark` along side the `color` and let the `preset` variables given the right value based on the color scheme.

```
{
	"slug": "foreground",
	"color": "#1E1E1E",
	"colorDark": "#DBDBDB",
	"name": "Foreground"
},
```
Also, a light-dark toggle button in the panel would be much useful to indicate the support of dark mode for a theme and let user validate colors.

simison:
Wouldn't this break apart when people start customizing colors for their site?

Feels like this would need clearer UI in Gutenberg to:

- Make it clear such feature exists for their site, since it's not obvious in any way.
- Enable or disable such feature for the site
- Allow adjusting and previewing styles for dark/light separately

Neat to see experimentation tho!

Issue #6276, opened 2022-07-22
Livro: Add white/light style variation


jeffikus:
Livro does not have an option to switch to a "white/light" style. We should add a style variation for this, effectively inverting the colors that are the default.

## Expected

@simison @beafialho can you please help to clarify the design requirements here, if this is a simple matter of switching colors or if there was something else more specific in mind for this?

mtias:
We discussed this with @beafialho a few weeks ago. Just a a version that has dark text on a pure white background.

beafialho:
That's right, just a change in color. Besides the "White" one, I also created a "Cream" one. Since this theme is suitable for long reads, it could be a nice addition.

They can both be found [here](https://www.figma.com/file/ALdcQWdMu3wmdoNv59VwYl/Livro?node-id=508%3A11).

White:
![White](https://user-images.githubusercontent.com/65220155/180747433-31538ab5-b58b-44c1-9d5a-1a003466efd1.png)

Cream:
![Cream](https://user-images.githubusercontent.com/65220155/180747423-ba99667a-0392-4ed3-8930-403028efa9e4.png)

jeffikus:
#6359 has been merged and deployed. Closing this, if the button issue persists @vcanales please open up a separate issue. Thank you!

PR #6359, opened 2022-08-09
Livro: add white and cream variations

vcanales:
#### Related issue(s):
https://github.com/Automattic/themes/issues/6276

### Preview
Kapture 2022-08-10 at 12 04 50

### DIFF
--- a/livro/styles/cream.json
+++ b/livro/styles/cream.json
@@ -17,7 +17,7 @@
 				{
 					"slug": "secondary",
 					"color": "#000000",
-					"name": "Background"
+					"name": "Secondary"
 				}
 			]
 		}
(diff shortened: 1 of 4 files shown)


pbking:
It seems that the palette colors `primary` and `tertiary` aren't used in the theme at all.  Can we remove them with this change (both for the primary theme.json as well as the variations?)

livro/styles/cream.json
```diff
@@ -0,0 +1,36 @@
+{
+	"version": 2,
+	"title": "Cream",
+	"settings": {
+		"appearanceTools": true,
```
pbking: Since `appearanceTools` is true in the default styles I think this can be eliminated in both of these variations.

pbking:
I made a minor label change.  The 'Primary' colors were labeled "Background" in the original and variations.

I also removed the 'initial' color value for buttons.  I'm not sure why that was added but it didn't work with the hover state of buttons in the the variations.

This was otherwise good to 🚢

pbking (approved):

PR #6365, opened 2022-08-10
Build Error on Blockbase and Children

bobmatyas:
Creating this PR [per Slack thread](p1660160773198309-slack-C029FM1EH). It isn't actually to be merged.

I'm having trouble building Zoologist. When I `npm run build` I get:

```
Error: Can't find stylesheet to import.
  ╷
1 │ @import '../../node_modules/@wordpress/base-styles/breakpoints';
  │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ╵
  ../blockbase/sass/base/_breakpoints.scss 1:9  @import
  sass/theme.scss 1:9                           root stylesheet 
```
I did `npm install` on Zoologist and on Blockbase. It also looks like the `../../node_modules/@wordpress/base-styles/breakpoints` file is present: 

`1037 Aug 10 15:41 _breakpoints.scss`

I get the same error when I try to build Blockbase.

Any help would be appreciated :)


pbking:
I was able to reproduce with node 16.x
Things worked fine for me for 14.x

To reproduce I had to delete my node_modules folder I had in my existing `/blockbase` folder and re-`npm install`.  A subsequent `npm run build` resulted in the error noted above.

I believe this is because of our recent shift to a npm workspace (#6200) which resulted in all node modules being stored in the root project rather than individual project `/node_modules` folder.

madhusudhand:
@bobmatyas Thanks for the report. #6371 should fix the issue.

PR #6371, opened 2022-08-11
Update sass build scripts to import from node_modules

madhusudhand:
#### Changes proposed in this Pull Request:

* Remove relative path references to `@wordpress/base-styles`
* include `node_modules` to sass compilation load path
* Add `style-rts.css` to deployment script to avoid version mismatch
* update deployment script to not trim `-wpcom` from `style.scss`

#### Related issue(s):

#6365 #6200

PR #6372, opened 2022-08-11
fix theme version mismatch in style-rtl.css

madhusudhand:

This is the result of rebuilding all themes `npm run build --workspaces` after the fixes made to build and deployment scripts [here](https://github.com/Automattic/themes/pull/6371/files#diff-877b4284538ebb4cf1411cf4487c456bda52082b9e72e4a0b6770d5a80d7d5a9)

#### Related issue(s):

#6371

### DIFF
--- a/spearhead/style.css
+++ b/spearhead/style.css
@@ -12,7 +12,7 @@ License: GNU General Public License v2 or later
 License URI: LICENSE
 Template: seedlet
 Text Domain: spearhead
-Tags: one-column, flexible-header, accessibility-ready, custom-colors, custom-menu, custom-logo, editor-style, featured-images, footer-widgets, rtl-language-support, sticky-post, threaded-comments, translation-ready, auto-loading-homepage, jetpack-global-styles
+Tags: one-column, flexible-header, accessibility-ready, custom-colors, custom-menu, custom-logo, editor-style, featured-images, footer-widgets, rtl-language-support, sticky-post, threaded-comments, translation-ready, auto-loading-homepage, jetpack-global-styles, blog-homepage
 */
 /**
  * Required Variables
--- a/spearhead/style.css
+++ b/spearhead/style.css
@@ -12,7 +12,7 @@ License: GNU General Public License v2 or later
 License URI: LICENSE
 Template: seedlet
 Text Domain: spearhead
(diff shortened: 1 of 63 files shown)


spearhead/style.css
```diff
@@ -12,7 +12,7 @@ License: GNU General Public License v2 or later
 License URI: LICENSE
 Template: seedlet
 Text Domain: spearhead
-Tags: one-column, flexible-header, accessibility-ready, custom-colors, custom-menu, custom-logo, editor-style, featured-images, footer-widgets, rtl-language-support, sticky-post, threaded-comments, translation-ready, auto-loading-homepage, jetpack-global-styles, blog-homepage
+Tags: one-column, flexible-header, accessibility-ready, custom-colors, custom-menu, custom-logo, editor-style, featured-images, footer-widgets, rtl-language-support, sticky-post, threaded-comments, translation-ready, auto-loading-homepage, jetpack-global-styles
```
mikachan: I think we need to keep the `blog-homepage` tag, as this was recently added here: https://github.com/Automattic/themes/pull/6303

I saw this happen locally as well, and I believe it's happening because this tag is missing from Spearhead's style.scss file, so it's removed/not added on build.

mikachan (commented):
This looks good to me; I got the same result when I ran `npm run build --workspaces` with your update from https://github.com/Automattic/themes/pull/6371.

I've left one inline comment that isn't related to your change but can probably be bundled in with these updates.

pbking:
All of these changes seem to be either version alignment or formatting changes.  Nothing that warrants a version bump I think.

When we merge this can we make sure to not do any theme version bumping with the change?  I don't even see a need to deploy this (though I think that a weasel would pop out of somewhere and remind us that we need to if there's a discrepancy from what's in SVN and what's been deployed).

I'm not terribly convinced that this change is even needed.  It does bring everything "up-to-date" but I don't believe the version difference in the -rtl.css files matter and the rest of the formatting changes could really wait for another build of the theme.

I'm mostly impartial (though I wouldn't want to be the one to deploy a change to so many themes!)

madhusudhand:
@pbking Agreed, these changes neither need a version bump nor a deployment. Now that deployment script considers `-rtl` we could safely ignore these changes and when a theme gets deployed next time, `-rtl` will be in sync.

Closing this PR.

PR #6389, opened 2022-08-12
Bump node and npm versions to avoid updating lock file on dependency install

vcanales:

Running `npm i` after `nvm use` downgrades the saved `lockfileVersion` to 2. The lockfile requirement is set to [`lockfile-version = 3` on .npmrc.](https://github.com/Automattic/themes/blob/trunk/.npmrc#L2)

In order to avoid unnecessary updates to package-lock.json, I'm setting all automated installs and requirements to versions that are compatible between each other. 

##### Current
```sh
> nvm use
< Now using node v16.16.0 (npm v7.24.2)
```

##### After this change
```sh
> nvm use
< Now using node v17.9.1 (npm v8.11.0)
```


mikachan (commented):
Thanks for this, works for me! I noticed that my local package-lock.json also updated to include the same changes from the package.json `engines` property, so we may need to commit that as well (the node and npm version bumps).

pbking:
I haven't figure out what problem this solves yet or what changes to package-lock.json are happening.

From what I can tell lockFileVersion 3 is was introduced in npm 7 so AFAICT we don't need to bump that to 8.  I don't think I have unexpected/unnecessary changes to package-lock.json files when installing packages.

Is this an update to bring the versions of this project into alignment with others?

vcanales:
> Is this an update to bring the versions of this project into alignment with others?

These changes just make sure that the combination of `nvm install`, `nvm use` and `npm install` will not cause an unnecessary update to the lock file.

> I haven't figure out what problem this solves yet or what changes to package-lock.json are happening.

This is what I see on trunk vs this branch:


| trunk - node v16/npm v7 downgrade `lockfileVersion` to 2 | this branch - node v17/npm v8 don't change the `lockfileVersion` |
|--|--|
| Kapture 2022-08-22 at 11 03 23 |Kapture 2022-08-22 at 11 06 56|

mikachan (approved):
LGTM

madhusudhand (approved):

PR #6390, opened 2022-08-12
#5724: Gazette: Fix blog posts read more links

### LINKED ISSUE #5724: Gazette: Can't disable "Continue reading" link in the Blog Posts block
### Quick summary

When using the Gazette theme, the Blog Posts block shows the "Continue reading"  link even when "Read More" setting has been disabled.

The link doesn't show in the Editor but you can see it in the front end.

image

### Steps to reproduce

1. Activate the Gazette theme.
2. Ensure you have a couple of posts added with minimum 50 words (to trigger "Read More" link).
3. Add a Blog Posts block and disable the "Read More" link.
4. Save changes.

The posts should only show the excerpt with no "Continue reading" link.

### What actually happened

The "Continue reading" link still shows even though this setting has been disabled.

The link doesn't show on the editor, only the frontend.

### Context

Reported here: 4855115-zen

### Simple, Atomic or both?

_No response_

### Theme-specific issue?

Was not able to replicate on a theme other than `gazette` and `Barnsbury`, report here for `Barnsbury`:
5249-gh-Automattic/themes

 Firefox 98 on macOS (Catalina) 

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

Yes, easy to implement

### Workaround details

You could add this custom CSS:

```
/* Hide default "Read More" link // 4620858-zen (JJ) */
article .more-link:not([rel]) {
    display: none;
}
```

If the "Read More" link is enabled, the CSS code won't hide the link.

bobmatyas:

To test this one, you need to have the Newspack plugin installed with a Homepage Posts block added.

##### Editor with no "Read More" Link

Screenshot on 2022-08-12 at 14-16-39

##### Frontend with no "Read More" Link

Screenshot on 2022-08-12 at 14-17-28

##### Editor with "Read More" Enabled

Screenshot on 2022-08-12 at 14-17-02

##### Frontend with "Read More" Enabled

Screenshot on 2022-08-12 at 14-18-01

##### Editor
Screenshot on 2022-08-12 at 14-20-04

##### Frontend with "Read More" link set
Screenshot on 2022-08-12 at 14-19-37

##### Frontend with no "Read More" link set
Screenshot on 2022-08-12 at 14-20-42

 
#### Related issue(s):

Fixes https://github.com/Automattic/themes/issues/5724

### DIFF
--- a/gazette/css/blocks.css
+++ b/gazette/css/blocks.css
@@ -194,6 +194,12 @@ p.has-drop-cap:not(:focus)::first-letter {
 	margin-bottom: 30px;
 }
 
+/* Newspack Blog Posts */
+
+.wpnbha p .more-link {
+	display: none;
+}
+
 /*--------------------------------------------------------------
 5.0 Blocks - Widgets
 --------------------------------------------------------------*/
--- a/gazette/css/editor-blocks.css
+++ b/gazette/css/editor-blocks.css
@@ -475,6 +475,12 @@ p.has-drop-cap:not(:focus)::first-letter {
 	text-align: right;
 }
 
+/* Newspack Blog Posts */
+
+.wpnbha p .more-link {
+	display: none;
+}
+
 /*--------------------------------------------------------------
 5.0 Blocks - Layout Elements
 --------------------------------------------------------------*/


gazette/css/blocks.css
```diff
@@ -194,6 +194,12 @@ p.has-drop-cap:not(:focus)::first-letter {
 	margin-bottom: 30px;
 }
 
+/* Newspack Blog Posts */
+
+.wpnbha p .more-link {
```
pbking, line 199: I'm not sure what `.wpnbha` is but it seems you're targeting the newspack block perhaps better targeted via `.wp-block-newspack-blocks-homepage-articles` instead.

pbking:
This does limit the scope of change to that block which was my only concern about the change before reviewing.  Once the class noted above is either confirmed as appropriate or changed this is good to come in.

gazette/css/blocks.css
pbking: I guess `.wpnbha` is just shorthand for the longer class name.  I would argue that the full class name of the element should be used in the theme CSS to avoid confusion for somebody looking at the code in the future.

bobmatyas: Sorry, I haven't been able to jump into this due to other duties. `.wpnbha` is a shorthand class. I've seen in used in the past in other themes (and have used it on other PRs) but I agree it isn't the most descriptive class name.

pbking: I don't know why I didn't put that together...  That's why I don't like shorthand classes I guess; It's hard to make the connection!

In that case I suppose it's fine to leverage it.  I'll give this a green light and we can get it shipped.  Thanks for clarifying.

pbking (approved):

PR #6392, opened 2022-08-15
Lynx: Initial commit.

alaczek:
This is an initial commit of the Lynx theme. This theme will be used as a part of a tailored use cases project to store patterns and styles to create a "link in bio" style site. 

I generated it off of Block Canvas, but the templates and theme.json are copied from the Lettre theme. For the moment the theme includes only one pattern - others will be added later.

Here's the project thread and more info: pdqkMK-r1-p2


lynx/index.php
```diff
@@ -0,0 +1,2 @@
+<?php
```
jffng: I don't think this file is needed for a block theme.

lynx/patterns/link-in-bio-audio embed.php
```diff
@@ -0,0 +1,54 @@
+<?php
+/**
+ * Title: Link in Bio with audio embed
+ * Slug: lynx/link-in-bio-audio embed
```
jffng: Is this a valid slug? I think it's missing a hyphen.
```suggestion
 * Slug: lynx/link-in-bio-audio-embed
```

lynx/patterns/link-in-bio-audio embed.php
```diff
@@ -0,0 +1,54 @@
+<?php
+/**
+ * Title: Link in Bio with audio embed
```
jffng: More of a general comment, but should the reference to Link in Bio be changed to Lynx, globally, since Link in Bio is a specific (non-a8c) product?

lynx/patterns/link-in-bio-purple-background.php
```diff
@@ -0,0 +1,86 @@
+<?php
+/**
+ * Title: Link in Bio with purple background
+ * Slug: lynx/link-in-bio-purple-background
+ * Inserter: yes
+ * Categories: featured, link-in-bio
+ */
+?>
+
+<!-- wp:cover {"customOverlayColor":"#7749f9","minHeight":100,"minHeightUnit":"vh","align":"full","style":{"spacing":{"padding":{"top":"8vh","right":"5vh","bottom":"8vh","left":"5vh"}}}} -->
```
jffng: Inserting this pattern results in a strange layout:

Screen Shot 2022-08-18 at 2 50 33 PM

lynx/readme.txt
```diff
@@ -0,0 +1,77 @@
+=== Lynx ===
+Contributors: automattic
+Requires at least: 5.8
+Tested up to: 5.8.3
+Requires PHP: 5.6
+Stable tag: 1.0.1
```
jffng: I think this can be removed.

lynx/readme.txt
```diff
@@ -0,0 +1,77 @@
+=== Lynx ===
+Contributors: automattic
+Requires at least: 5.8
+Tested up to: 5.8.3
```
jffng:
```suggestion
Tested up to: 6.0
```

jffng (commented):
👋  @alaczek this generally is looking good to me. I am curious about the decision to make this a new theme, since it's basically just a blank canvas and adds a lot of new patterns.

lynx/patterns/link-in-bio-purple-background.php
alaczek: Yes, I noticed this too. I think it's an editor issue: https://github.com/WordPress/gutenberg/issues/34909

lynx/patterns/link-in-bio-audio embed.php
alaczek: You're right about the missing dash, but oddly enough the pattern still shows up for me in the inserter ¯\_(ツ)_/¯

Edit: It's because I named the file wrong too 🤦‍♀️

lynx/readme.txt
alaczek: I'm happy to remove this, but I still see this section in other themes we release. What's the difference? :)

lynx/patterns/link-in-bio-purple-background.php
```diff
@@ -0,0 +1,86 @@
...
+<!-- /wp:group -->
+
+<!-- wp:group {"style":{"color":{"background":"#ffffffe6","text":"#111111"},"border":{"radius":"100px"},"spacing":{"blockGap":"0px","padding":{"top":"10px","right":"10px","bottom":"10px","left":"10px"},"margin":{"top":"10px","bottom":"10px"}},"elements":{"link":{"color":{"text":"#111111"}}}}} -->
+<div class="wp-block-group has-text-color has-background has-link-color" style="border-radius:100px;color:#111111;background-color:#ffffffe6;margin-top:10px;margin-bottom:10px;padding-top:10px;padding-right:10px;padding-bottom:10px;padding-left:10px"><!-- wp:group {"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"left"}} -->
+<div class="wp-block-group"><!-- wp:social-links {"className":"is-style-logos-only"} -->
+<ul class="wp-block-social-links is-style-logos-only"><!-- wp:social-link {"url":"#","service":"twitter"} /--></ul>
+<!-- /wp:social-links -->
+
+<!-- wp:paragraph -->
+<p><a href="#"><?php echo  esc_html__( 'Twitter', 'lynx' ); ?></a></p>
```
simison: Do we really need to translate service names?

jffng: If the theme is to be submitted to the .org themes directory, I believe there is still the requirement that all strings be escaped + translated. 

If it's not going to be submitted there, then I don't think it's necessary to translate, but also not sure why it would be added to this repo instead of just living on dotcom.

lynx/readme.txt
jffng: The tag isn't required, and I don't think it's meaningful for block themes since they are not so complicated / don't bundle "functionality" like classic themes do, so the latest version should always be considered stable.

lynx/patterns/link-in-bio-purple-background.php
alaczek: We won't be submitting Lynx to .org. I added it here because I thought all free themes are worked on in this GH repo, not just the ones going to .org. Is this ok to keep it here?

I'll take care of removing the translation around social media services, but likely after this initial commit is merged.

lynx/patterns/link-in-bio-audio embed.php
alaczek: Link in Bio is the naming convention used in the onboarding flow this will feed into, so I'm following that for individual patterns.

lynx/patterns/link-in-bio-purple-background.php
alaczek: This particular pattern's setup is prooving too complex, so I'll likely be removing it, at least from the initial commit.

lynx/readme.txt
alaczek: That makes sense, thanks for taking the time to explain! I updated the readme.txt.

lynx/patterns/link-in-bio-purple-background.php
pbking: >  I thought all free themes are worked on in this GH repo, not just the ones going to .org

I don't think this would be the only theme on wpcom but not in themes directory.  I can't recall which but I think there are others.  For workflow I think all free themes being here (except those that are retired or near retirement) makes sense.

PR #6402, opened 2022-08-16
Disco: removing margin added when Gutenberg is disabled

matiasbenedetto:
Club: removing this margin added when Gutenberg is disabled

image

### DIFF
--- a/club/style.css
+++ b/club/style.css
@@ -161,6 +161,10 @@ a {
 	min-width: unset;
 }
 
+.club-post-list-pattern .wp-block-post {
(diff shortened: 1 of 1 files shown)


jffng (commented):
I'm not sure we should patch this, this is the results of my test on Core 6.0.1 without Gutenberg activated:

Trunk | This PR
--- | ---
Screen Shot 2022-08-16 at 11 49 57 AM | Screen Shot 2022-08-16 at 11 49 51 AM

pbking:
> I'm not sure we should patch this.

I agree.  The difference seems fine to me, it's temporary, and not adding CSS now means no changes to schedule.

PR #6403, opened 2022-08-16
Club: improve post editor by changing font size of title and width of layout

matiasbenedetto:
Club: improve post editor by changing font size of title and width of layout

Before:
image

After:
image


jffng (approved):
This LGTM and definitely feels like an improvement, so long as we are okay to change the theme's default and wide widths.

Before | After
--- | ---
Screen Shot 2022-08-16 at 11 41 14 AM | Screen Shot 2022-08-16 at 11 41 08 AM

PR #6405, opened 2022-08-16
Removed font families from the child so that Blockbases are used instead

pbking:
Fixes issues with Meraki theme noted here: https://github.com/Automattic/wp-calypso/issues/63229


mikachan (approved):
LGTM! Matches the behavior discussed [here](https://github.com/Automattic/wp-calypso/issues/63229).

PR #6413, opened 2022-08-18
Renaming theme "Club" to "Disco"

matiasbenedetto:
Renaming theme "Club" to "Disco" because "Club" is not available in dotOrg theme directory.


mikachan (approved):
LGTM 👍

PR #6417, opened 2022-08-18
Rainfall: Update navigation styles and tidy up templates

### LINKED ISSUE #6245: Rainfall: Header and Footer improvements

### LINKED ISSUE #6431: Rainfall: "Menu" term should be the same size has site title
Currently:

Captura de ecrã 2022-08-23, às 10 46 58

Intended:

Captura de ecrã 2022-08-23, às 10 47 47

mikachan:
This PR fixes the navigation styles on Rainfall to match the [designs](oz0a2czNn0UNt9onrh8T5a-fi-107%3A906) more closely. 

It also includes the following updates:

- adds the theme description from [here](https://github.com/Automattic/themes/issues/6421#issuecomment-1222095777)
- moves some CSS into theme.json
- adjusts the fluid typography settings to include some minimum values
- applies spacing presets across templates where possible
- updates all layout settings to new format (part of https://github.com/Automattic/themes/issues/6477)

#### Related issue(s):
https://github.com/Automattic/themes/pull/6305


rainfall/style.css
```diff
@@ -1,25 +1,38 @@
...
 Text Domain: rainfall
-Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, featured-images, full-site-editing, rtl-language-support, theme-options, threaded-comments, translation-ready, wide-blocks
+Tags: two-columns, custom-colors, custom-menu, custom-logo, editor-style, featured-images, full-site-editing, block-patterns, rtl-language-support, theme-options, threaded-comments, translation-ready, auto-loading-homepage, blog-homepage
 */
 
+/*
+ * Font smoothing
+ */
+body {
+	-moz-osx-font-smoothing: grayscale;
```
scruffian: I don't think we should be doing this.

scruffian:
With this change my header now has additional padding:

Screenshot 2022-09-05 at 10 38 15

mikachan:
I've just pushed a change that tweaks the fluid font sizes a bit more, I thought the range of sizes was too big as the small font size was looking quite big at larger resolutions.

> With this change my header now has additional padding:

I think this is because I've added left and right root padding values so we can use `useRootPaddingAwareAlignments`, and it looks like this is adding horizontal padding to the header. Mine doesn't look quite as bad/spaced out as yours, although I am seeing more padding on the align wide heading:

| Without this PR | With this PR |
| --------------- | ------------- |
| image | image | 
| image | image |

rainfall/style.css
mikachan: Yeah I'm not sure about this either! I feel like it's only there to improve font appearance for a small number of users.

scruffian: I removed it

scruffian (approved):
LGTM

PR #6429, opened 2022-08-22
Livro: fix button hover styles and button outline

vcanales:
#### Problem

After adding Cream and White variations for Livro, hover color on buttons started behaving differently among variations, making button text illegible on buttons depending on whether the background was clear or dark.

Also, editor styles setting `border-width: 0` were making button outlines invisible.
Kapture 2022-08-22 at 18 01 21
 

Set the correct hover color depending on the type of button (regular/outline).
Define the border-width for outline buttons.

### DIFF
--- a/livro/style.css
+++ b/livro/style.css
@@ -85,6 +85,21 @@ a:active {
 	opacity: 0.90;
 }
 
+
+.wp-block-button__link:not(.has-text-color) {
+	color: var(--wp--preset--color--background);
+}
+
+.is-style-outline > .wp-block-button__link,
+.wp-block-button__link.is-style-outline {
+	border-width: 2px;
+}
+
+.is-style-outline > .wp-block-button__link:not(.has-text-color),
+.wp-block-button__link.is-style-outline:not(.has-text-color) {
+	color: var(--wp--preset--color--foreground);
+}
+
 /*
  * Alignment styles.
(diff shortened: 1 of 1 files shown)


jffng:
Hey @vcanales ! 

I'm having trouble replicating the issue and the fix described. For example on trunk, I've added solid and outline style buttons to a post in each variation, and they all seem to appear correctly, granted there is some wonkiness in the padding but I think that's unrelated to this PR:

Screen Shot 2022-08-24 at 10 41 03 AM

Could you outline some testing instructions for this?

pbking:
This is also not reproducing for me on /trunk.  Will move it back to in-progress, but likely we can close this change.

mikachan:
I don't think any of these are Livro-specific issues.

I can reproduce the hover style issue across the new variations on GB 13.9.0, but not GB 14.0.2. I can also see the outline border issue, but again it seems to be a problem on 13.9.0, and not 14.0.2:

| GB 13.9.0 | GB 14.0.2 |
| --------- | --------- | 
| image | image |

On GB 14.0.2 (and trunk), it looks like the padding is different between a default and outline button on the front end, but it seems to match in the editor.

| Front end | Editor |
| --------- | ------ |
| image  | image |

So, I think I'd lean towards closing this issue for this theme specifically and helping to fix these other issues upstream.

PR #6434, opened 2022-08-23
#4914: Shoreditch: Fix padding on group block

### LINKED ISSUE #4914: Shoreditch: Group blocks don't align text with the right padding
### Quick summary

On Shoreditch, Group blocks with colored backgrounds have text go all the way to the edges of the block. The block itself is aligned with non-grouped text in the same width (full width or otherwise), so that adding padding would make it out of line with non-grouped text.

In the editor, there's padding (though the text is not aligned with non-grouped text, which is also fine): 

image

On the live site (both a user's and this demo), it looks like this: 

image

### Steps to reproduce

1. Activate Shoreditch on a site.
2. Add some paragraphs and columns, in and out of Group blocks with colored backgrounds, and including some set to full width.

Group blocks should have space between their edges and the start of text, and full-width blocks should go all the way to the edge of the page.

### What actually happened

The text starts at the very left side of the Group block, and full-width blocks have spacing on their left side. 

### Context

32251081-hc (Business site with plugins); replicated on test site (Free)

### Operating System

Windows, macOS, Linux

### Browser

Google Chrome/Chromium, Mozilla Firefox, Microsoft Edge, Microsoft Edge (legacy), Apple Safari

### Simple, Atomic or both?

Simple, Atomic

### Theme-specific issue?

_No response_

### Other notes

_No response_

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

Yes, difficult to implement

### Workaround details

Using another theme and using CSS or reusable blocks to mimic this theme's features, though that's naturally not ideal

bobmatyas:
To test, add group block and change background colors. Frontend now reflects what is shown in the editor.

##### Before

Screenshot on 2022-08-23 at 13-21-39

##### After

Screenshot on 2022-08-23 at 14-17-15

#### Related issue(s):

Fixes https://github.com/Automattic/themes/issues/4914

### DIFF
--- a/shoreditch/css/blocks.css
+++ b/shoreditch/css/blocks.css
@@ -320,6 +320,19 @@ p.has-drop-cap:not(:focus)::first-letter {
 	width: 100%;
 }
 
+/* Group */
+
+.wp-block-group {
+	margin-bottom: 1.5em;
+	padding: 1.5em;
+}
+
+.wp-block-group > * {
+	margin-left: auto;
+	margin-right: auto;
+	max-width: 900px;
+}
+
 /*--------------------------------------------------------------
 4.0 Blocks - Formatting
 --------------------------------------------------------------*/


pbking:
This change introduces a lot of padding changes.  If the goal is for the view and editor to match I believe that it's the editor that should change.

Before:
image

After:
image

pcrugM-lB-p2

PR #6435, opened 2022-08-24
Spearhead Blocks refactored Initial Commit

onuro:
Refactored commit of Spearhead Blocks, now being a clone of Block Canvas, instead of child of Blockbase. Also had to re-commit due to folder name of the theme has been renamed to match the naming conventions(maybe I shouldn't have created another PR, please let me know)

I tried the dark mode auto switch via style but I think I'm for now lacking the skill to reflect the dark mode changes to the editor. Ideas welcome :)

Thanks!


spearhead-blocks/parts/header.html
```diff
@@ -0,0 +1,19 @@
...
+<div class="wp-block-group"><!-- wp:site-logo {"width":64} /-->
+
+<!-- wp:group {"style":{"spacing":{"blockGap":"5px"}}} -->
+<div class="wp-block-group"><!-- wp:site-title /-->
+
+<!-- wp:site-tagline /--></div>
+<!-- /wp:group --></div>
+<!-- /wp:group -->
+
+<!-- wp:navigation {"ref":14,"__unstableLocation":"primary","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /--></div>
```
pbking: The navigation block has a reference.  This should be removed.

spearhead-blocks/style.css
```diff
@@ -0,0 +1,121 @@
...
+License URI: https://raw.githubusercontent.com/Automattic/themes/trunk/LICENSE
+Template:
+Text Domain: spearhead-blocks
+Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, featured-images, full-site-editing, rtl-language-support, theme-options, threaded-comments, translation-ready, wide-blocks
+*/
+/*
+ * Dark Mode
+ */
+@media (prefers-color-scheme: dark) {
+  body {
```
pbking: To get this to work in the editor you'll need to use:
```
  body.editor-styles-wrapper,
  body {
```

spearhead-blocks/theme.json
```diff
@@ -0,0 +1,455 @@
...
+				{
+					"color": "#ffffff",
+					"name": "Background",
+					"slug": "background"
+				},
+				{
+					"color": "#fafbf6",
+					"name": "Tertiary",
+					"slug": "tertiary"
+				},
```
pbking: I don't see the tertiary slug being used anywhere.  Is it necessary to include?

onuro: Ah, thought original Spearhead also had it. Apparently not :) Removing.

spearhead-blocks/style.css
onuro: Worked like charm ✨ 🙏🏿

spearhead-blocks/theme.json
```diff
@@ -0,0 +1,450 @@
...
+			"core/post-title": {
+				"elements": {
+					"link": {
+						":hover": {
+							"color": {
+								"text": "var(--wp--preset--color--primary)"
+							}
+						},
+						"color": {
+							"text": "var:preset|color|foreground"
```
mikachan, line 266: It looks like this was originally set to the primary color, based on the Seedlet base CSS [here](https://github.com/Automattic/themes/blob/trunk/seedlet/assets/sass/base/_base.scss#L3).

If we switched the text color for primary we could perhaps use foreground for the hover color.

spearhead-blocks/theme.json
```diff
@@ -0,0 +1,450 @@
...
+						":hover": {
+							"color": {
+								"text": "var(--wp--preset--color--secondary)"
+							}
+						},
+						"color": {
+							"text": "var:preset|color|primary"
+						},
+						"typography": {
+							"textDecoration": "none"
```
mikachan, line 358: Based on the original theme, it looks like `"textDecoration": "none"` needs to be applied to the :hover and :focus styles too.

spearhead-blocks/theme.json
```diff
@@ -0,0 +1,450 @@
...
+				},
+				"spacing": {
+					"margin": {
+						"bottom": "0"
+					}
+				},
+				"typography": {
+					"fontSize": "var:preset|font-size|medium",
+					"fontStyle": "normal",
+					"fontWeight": "700",
```
mikachan: The font-weight for post titles was set to 500 on the original theme.

spearhead-blocks/theme.json
```diff
@@ -0,0 +1,450 @@
...
+				"color": {
+					"text": "var:preset|color|primary"
+				},
+				"typography": {
+					"textDecoration": "none"
+				}
+			}
+		},
+		"spacing": {
+			"blockGap": "35px"
```
mikachan: I can see a few gaps in the original theme using 0.5em (I think it's defaulting to the GB default gap spacing), rather than 35px.

spearhead-blocks/parts/footer.html
```diff
@@ -0,0 +1,11 @@
+<!-- wp:group {"layout":{"inherit":true}} -->
+<div class="wp-block-group">
+	<!-- wp:group {"style":{"spacing":{"padding":{"top":"80px","bottom":"30px"}}}} -->
+	<div class="wp-block-group" style="padding-top:80px;padding-bottom:30px">
+		<!-- wp:paragraph {"align":"center"} -->
+		<p class="has-text-align-center">Proudly Powered by <a href="https://wordpress.org" rel="nofollow">WordPress</a></p>
```
mikachan: The original theme includes the site title in the footer, next to the 'powered by WordPress' line:

image

mikachan (commented):
This is looking really good! I've left some inline comments, which are mainly comparing this theme to the original Spearhead theme. I'm not sure we want to match all of the differences exactly - I just wanted to point out the ones I noticed.

We should probably close https://github.com/Automattic/themes/pull/6331 in favour of this PR, if you're happy moving ahead with this one.

spearhead-blocks/theme.json
onuro: Thanks but when I tried that, it wasn't matching with the demo, where all post titles were also 700 in weight:
https://spearheaddemo.wordpress.com/

PR #6437, opened 2022-08-24
Adding Light Color Scheme

onuro:
This PR adds light block style to the Videomaker, based on Videomaker Light.

Ref: https://wp.me/pbAok1-3a7#comment-5289

### DIFF
--- a/videomaker/styles/light.json
+++ b/videomaker/styles/light.json
@@ -1,0 +1,30 @@
+{
+	"version": 2,
+	"title": "Light",
+	"settings": {
+		"color": {
+			"palette": [
+				{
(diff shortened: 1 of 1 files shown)


videomaker/styles/light.json
```diff
@@ -0,0 +1,30 @@
...
+					"name": "Background"
+				},
+				{
+					"slug": "primary",
+					"color": "#000000",
+					"name": "Primary"
+				},
+				{
+					"slug": "tertiary",
+					"color": "var:preset|color|tertiary",
```
madhusudhand, line 24: I don't think colors are inherited. The variable isn't working.

image

PR #6438, opened 2022-08-24
Block Canvas: replace alignment CSS

jffng:

This PR removes the alignment CSS in favor of the Gutenberg enabled site wide padding. 

To test, activate block canvas and ensure the site has some global padding (i.e. content does not reach the edge of the screen).

#### Related issue(s):

None, but as more folks use block canvas as a starting point it would be good to not include this CSS if it's not needed!

### DIFF
--- a/block-canvas/theme.json
+++ b/block-canvas/theme.json
@@ -346,8 +346,8 @@
         "spacing": {
             "blockGap": "1.5rem",
             "padding": {
-                "right": "var(--wp--preset--spacing--30)",
-                "left": "var(--wp--preset--spacing--30)"
+                "right": "var(--wp--preset--spacing--60)",
+                "left": "var(--wp--preset--spacing--60)"
             }
         },
         "typography": {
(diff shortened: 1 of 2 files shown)


block-canvas/theme.json
```diff
@@ -343,7 +344,11 @@
             }
         },
         "spacing": {
-            "blockGap": "30px"
+            "blockGap": "1.5rem",
```
jffng, line 347: Thought it would be better to encourage `rem` usage also.

scruffian (approved):
Thanks for doing this.

block-canvas/theme.json
```diff
@@ -343,7 +344,11 @@
...
+            "padding": {
+                "right": "var(--wp--preset--spacing--30)",
```
scruffian: Do we need to opt in to spacing to get these?

pbking: I see that they are coming through and it's working as expected without opting into anything.

However, *--spacing--30 seems a fair bit smaller than the original 30px.

PR #6441, opened 2022-08-25
Lynx: Fix wrong button label in one of the patterns.

alaczek:
This is to fix a copy-paste error resulting in a wrong button label.


mikachan (approved):
LGTM 👍

PR #6450, opened 2022-08-29
Rainfall: add Home template

### LINKED ISSUE #6430: Rainfall: Add Front Page template
In addition to Index, Rainfall should have a "Home" or "Front Page" template that [looks like the one on Figma](https://www.figma.com/file/oz0a2czNn0UNt9onrh8T5a/Rainfall?node-id=29%3A153):

I've attempted to create it but ran into some bugs and difficulties:

**1. Post titles seem to be offset everywhere. This should only happen in post templates and in the header.**

Captura de ecrã 2022-08-23, às 10 35 13

**2. Sidebar template part should contain Categories and Latest Posts.**

Captura de ecrã 2022-08-23, às 10 38 22

**3. There have to be special footers and headers for the Front Page because both have content that "blends" with them.**

Here's the markup of what I created, though, for reference:

```
<!-- wp:group {"backgroundColor":"orange"} -->
<div class="wp-block-group has-orange-background-color has-background"><!-- wp:template-part {"slug":"header","theme":"themes/rainfall","tagName":"header"} /-->

<!-- wp:group {"layout":{"inherit":true,"type":"constrained"}} -->
<div class="wp-block-group"><!-- wp:spacer {"height":"12px"} -->
<div style="height:12px" aria-hidden="true" class="wp-block-spacer"></div>
<!-- /wp:spacer --></div>
<!-- /wp:group --></div>
<!-- /wp:group -->

<!-- wp:template-part {"slug":"query","theme":"themes/rainfall"} /-->

<!-- wp:group {"layout":{"type":"default"},"backgroundColor":"black"} -->
<div class="wp-block-group has-black-background-color has-background"><!-- wp:group {"layout":{"inherit":false},"align":"wide","style":{"spacing":{"padding":{"bottom":"8rem","top":"8rem"}}},"backgroundColor":"black"} -->
<div class="wp-block-group alignwide has-black-background-color has-background" style="padding-top:8rem;padding-bottom:8rem"><!-- wp:heading {"level":1,"textColor":"white"} -->
<h1 class="has-white-color has-text-color">Work With Us</h1>
<!-- /wp:heading -->

<!-- wp:columns -->
<div class="wp-block-columns"><!-- wp:column {"verticalAlignment":"center"} -->
<div class="wp-block-column is-vertically-aligned-center"><!-- wp:paragraph {"textColor":"white"} -->
<p class="has-white-color has-text-color">Rainfall is an online magazine featuring curated stories about real people. We are passionate about interviewing people from self-employed enterpreneurs to established CEO’s and represent the times.</p>
<!-- /wp:paragraph --></div>
<!-- /wp:column -->

<!-- wp:column {"verticalAlignment":"center","width":"var(\u002d\u002dwp\u002d\u002dpreset\u002d\u002dspacing\u002d\u002d70)"} -->
<div class="wp-block-column is-vertically-aligned-center" style="flex-basis:var(--wp--preset--spacing--70)"></div>
<!-- /wp:column -->

<!-- wp:column {"verticalAlignment":"center"} -->
<div class="wp-block-column is-vertically-aligned-center"><!-- wp:buttons {"layout":{"type":"flex","justifyContent":"right"}} -->
<div class="wp-block-buttons"><!-- wp:button {"backgroundColor":"white","textColor":"black","style":{"typography":{"textTransform":"uppercase"}},"fontSize":"small"} -->
<div class="wp-block-button has-custom-font-size has-small-font-size" style="text-transform:uppercase"><a class="wp-block-button__link has-black-color has-white-background-color has-text-color has-background wp-element-button">Browse Open Positions</a></div>
<!-- /wp:button --></div>
<!-- /wp:buttons --></div>
<!-- /wp:column --></div>
<!-- /wp:columns --></div>
<!-- /wp:group -->

<!-- wp:template-part {"slug":"dark-footer","theme":"themes/rainfall"} /--></div>
<!-- /wp:group -->
```

jffng:

This PR adds a Home template to rainfall.

#### Related issue(s):

Closes #6430


rainfall/style.css
```diff
@@ -15,11 +15,6 @@ Text Domain: rainfall
 Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, featured-images, full-site-editing, rtl-language-support, theme-options, threaded-comments, translation-ready, wide-blocks
 */
 
-.wp-block-post-title:not(.editor-post-title) {
```
jffng, line 20: This is a cool effect but I think the selector is too generic and will break anytime a user inserts a post title, without them being able to customize it.

jffng:
This is still somewhat WIP, since the elements should be probably be pattern-ized and escaped? 

I should move onto Pixel development though, so if someone else wants to pick up work on this PR that would be great.

pbking:
I'm working on localizing the pattern now and will look about the rest of it too.

mikachan:
I've updated this PR with the following:

- Patternized the sidebar and home templates
- Added the header image to the home template
- Removed some refs and ids from other templates
- Moved existing images to /assets directory and removed the /assets/images directory, as there are no other types of assets

This should be ready for another review 🙇

pbking:
This was a good chunk moving it in the right direction.  Should this be brought in (as it has important asset changes) and refine it further?

Going by the Figma these are the design differences I notice:

The home header is missing the uh, header part.
image

Can/should we offset the heading text like in the figma?  Would take CSS... is it the proper opportunity for a theme-specific block style?  It seems like a unique and important part of the theme's style.

image

The dimensions of the featured image are different and figma:
image

The sidebar pattern seems absent from the home template

There's a gap from something or another below the footer:
image

Even with the sidebar content perhaps this should still be constrained to 'wide' width.  It's difficult to discern the intent from the figma, but all of the text is contained under the image, which has a tall dimension which makes me think that as wide as the design is is as wide as it should get.

image

mikachan:
Thanks for taking a look! Yeah, we could bring it in now and refine it further.

Hmm, it looks like you're missing the header, footer, and sidebar template parts completely. Can you add them from within the editor? I've had this problem before too, and adding `"theme":"rainfall"` to the template part markup seemed to fix it. But that's already included here, so maybe that's not actually a fix...

> Can/should we offset the heading text like in the figma?

We were offsetting the titles with CSS, but this was [removed as part of this PR](https://github.com/Automattic/themes/pull/6450#discussion_r957569321) as it was being applied to too many elements. I'll look at this again and perhaps make it extremely specific (maybe in another PR after this).

> The dimensions of the featured image are different and figma:

Oh yeah! I've updated the image sizes in the query loop:

image

> Even with the sidebar content perhaps this should still be constrained to 'wide' width.

I was also thinking of making this wide width.. But then the edges of the content don't match the header/footer, and the large orange header in this case. I'm not sure which is best. Here's a wide width version:

local-test-site local_

rainfall/style.css
scruffian: I pushed a commit to do this in a different way

scruffian:
Don't we need to add the header and footer templates to home.html?

mikachan:
> Don't we need to add the header and footer templates to home.html?

They are there but they don't seem to show unless you insert them again in the editor! I saw the same problem when I started working on this PR, it seems that adding/removing the `"theme":"rainfall"` part of the template part tag makes them appear. I was seeing the 'template part missing' error in the editor until I added the theme name.

rainfall/style.css
mikachan: Thanks, this works well!

pbking:
There seems to be issue leveraging a template-part from a pattern.  If I move the content back to the template the template part header is loaded fine.  I'm not sure that's a situation I've tried before but it looks like it doesn't work out.

scruffian:
Why not just add the template parts to the home.html template?

scruffian:
> Why not just add the template parts to the home.html template?

I added a commit to do this.

mikachan:
> Why not just add the template parts to the home.html template?

~I think we've included them in the pattern because the header needs to inherit the orange background and the footer needs to inherit the black background. We should probably move them into the home.html template along with their background colours.~

Commented too soon - thanks! 😁

mikachan:
With those last template changes, I'm seeing the top and bottom parts of the home template with no left and right padding:

Top part:
image

Bottom part:
image

I don't know if it's just my environment. Setting `"layout":{"type":"constrained"}` seems to add too much horizontal padding compared to the query block (which I think is a known bug: https://github.com/WordPress/gutenberg/issues/43095).

pbking:
@mikachan would you take another look?  I did some hefty refactoring to this template.  Had a blast but probably spent too long on it. :P

I think in the process I've addressed the padding and spacing issues raised.

I applied the same class to the header to get the text offset.  That seemed safe to apply in a pattern.  
Also I noticed the min/max was inverted for the xx-large font size so I flipped those.

image
image

I brought in the same block gap for the columns as was in the content and also constrained the query content to "wide" size which seems to flow nicely and I think lays out as intended in the design.
I tinkered with the sizing of the images a bit.  Started with making them full width to hit the alignment in the design, but assigning the height was tricker.  Interested to know what you think of how it landed.

image

image

Finally I leveraged the existing CTA pattern and stuck the footer in a group with custom colors.
image

Which leaves the only remaining (I think) is the CTA pattern and it's not-very-mobile-friendly layout:

image

(which I think is a separate issue entirely)

mikachan (approved):
This is looking ace, thank you!

I really like the way you've handled the featured image sizes, and I think the height for these is working well.

I also think it's a great idea to split the template up into multiple patterns. It's much easier to read and understand, and it means users can more easily use parts of the design elsewhere.

I think we should bring this in. We can iterate if we need to! I've opened an issue for the CTA pattern here: https://github.com/Automattic/themes/issues/6526

PR #6478, opened 2022-08-30
Pixel: update color palette

jffng:
#### Changes proposed in this Pull Request:

Updates color palette to match the design.

#### Related issue(s):

Closes #6453

PR #6480, opened 2022-08-30
Pixel: set correct foreground color

vcanales:

Jumped the gun on merging https://github.com/Automattic/themes/pull/6478; the design actually uses `#040DE1` as Foreground color (same as Primary.)

### DIFF
--- a/pixel/theme.json
+++ b/pixel/theme.json
@@ -36,7 +36,7 @@
                     "slug": "primary"
                 },
                 {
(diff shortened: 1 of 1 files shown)


vcanales:
@jffng  I'm going to merge this, as colors looking OK depends on it, but I think we might want to discuss whether or not having Primary and Foreground be the same is useful.

jffng:
Thank you @vcanales !

PR #6486, opened 2022-08-31
Block canvas: replacing button :hover CSS by theme.json definition

matiasbenedetto:
Block canvas: replacing button pseudo selectors CSS by theme.json definitions

### DIFF
--- a/block-canvas/style.css
+++ b/block-canvas/style.css
@@ -38,18 +38,6 @@ GNU General Public License for more details.
 }
 
 /*
- * Button hover styles.
- * Necessary until the following issue is resolved in Gutenberg:
- * https://github.com/WordPress/gutenberg/issues/27075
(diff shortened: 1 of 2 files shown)


mikachan (commented):
I think you've accidentally added link pseudo-states, rather than replacing the button hover state CSS with theme.json settings, which you could do like this:

```
"elements": {
            "button": {
                      ":hover": {
                          "color": {
                              "background": "var(--wp--preset--color--secondary)"
                          }
                      }
```

matiasbenedetto:
> I think you've accidentally added link pseudo-states, rather than replacing the button hover state CSS with theme.json settings, which you could do like this:


that's true! my mistake. I updated the PR and description.

mikachan (approved):
Looks good!

Issue #6243, opened 2022-07-15
text-decoration shorthand with thickness has no support in WebKit browsers


madhusudhand:
## Steps to replicate

Text decoration short hand that has thickness included makes the entire css style invalid in safari.

```
"typography": {
	"textDecoration": "none 0.25rem"
}
```

<img width="789" alt="image" src="https://user-images.githubusercontent.com/1935113/179197230-1f8f5dac-76a9-4b7b-a7b1-a72e1d68da78.png">

## Browser support

<img width="789" alt="image" src="https://user-images.githubusercontent.com/1935113/179196995-54a14a26-d77c-46fa-b6ad-be73638d9fad.png">

## Result

This effects following block themes.

- [x] Vivre https://github.com/Automattic/themes/pull/6490
- [x] Pendant https://github.com/Automattic/themes/pull/6490
- [x] Disco

## Expected

Should separate `text-decoration-tickness` from `theme.json` and define in `style.css`

pbking:
"Safari, the new Internet Explorer"

pbking:
I rather think we should just close this one.  I don't think we have an option to NOT use shorthand in a few places, at least not without/until a Gutenberg change.  If anything this should be opened there I think (though I'm not sure exactly what the change should be... maybe including text-decoration-thickness as a Global Styles-able attribute?)

matiasbenedetto:
I updated the title because I can confirm that isn't working with other WebKit browsers either.

![Screenshot from 2022-08-12 12-43-32](https://user-images.githubusercontent.com/1310626/184338783-0efad0ef-9853-454c-bb79-6a694abcedd1.png)

matiasbenedetto:
I took this approach for Disco: https://github.com/Automattic/themes/pull/6384

mikachan:
Closing with https://github.com/Automattic/themes/pull/6490.

PR #6490, opened 2022-08-31
Removing thickness from text-decoration shorthand because it is not supported on WebKit

matiasbenedetto:
Vivre and Pendant: removing thickness from text-decoration shorthand because it is not supported on WebKit browsers as Safari

#### Related issue(s):
Part of: https://github.com/Automattic/themes/issues/6243


mikachan (approved):
LGTM 👍

PR #6502, opened 2022-09-02
Russell: adjust featured image in single template

### LINKED ISSUE #67300: Russell: Featured Image shouldn't be full width on Single Post template
### Quick summary

The featured image size on single posts is huge, and the user has to scroll through the image to read the content.

### Steps to reproduce

1. Add a Post Featured Image block on the post page.
2. Open the page preview. The featured image is very big.

It would be great if the featured image is not full width and if it could have a constrained height to avoid too much scrolling.

Intended image size:

russell-intended-size-featured-image

### What actually happened

The featured image size on single posts is very big on the Russell theme. 

russell-full-width-featured-image

### Context

Issue reported in forum- https://wordpress.com/forums/topic/how-can-i-make-the-featured-photo-in-russell-theme-smaller-size-i-upload/

Simple, Atomic

### Theme-specific issue?

Russell theme

_No response_

### Reproducibility

Consistent

### Severity

All

### Available workarounds?

Yes, easy to implement

### Workaround details

The users on the WordPress.com Premium plan and above can add this CSS- 

```
.wp-block-post-featured-image img {
width: unset;
}
```

yashitamittal11:
**Description**
The featured image size on single posts is huge, and the user has to scroll through the image to read the content.

This PR makes the featured image on the single template of Russell theme wide align, and gives it a set height of 30rem.

**Test instructions**

1. Add a Post Featured Image block on the post page.
2. Open the page preview. The featured image should be wide aligned with constrained height.

### Screenshots

| Before | After |
|---|---|
| russell-full-width-featured-image | russell-intended-size-featured-image |

#### Related issue(s):

Fixes https://github.com/Automattic/wp-calypso/issues/67300

### DIFF
--- a/russell/block-templates/single.html
+++ b/russell/block-templates/single.html
@@ -8,7 +8,7 @@
 
 <!-- wp:group {"tagName":"main","lock":{"move":false,"remove":true}} -->
 <main class="wp-block-group">
-<!-- wp:post-featured-image {"align":"full"} /-->
+<!-- wp:post-featured-image {"align":"wide","height":"30rem"} /-->
 
 <!-- wp:post-content {"layout":{"inherit":true},"lock":{"move":false,"remove":true}} /-->
 


mikachan (commented):
This solution addresses the issue of the image being too large on single posts, however, it does change the default design slightly. 

Users can change the alignment of the single post featured image using the Site Editor if they'd prefer the image to be smaller, so do we want to change the default design or allow users to adjust this themselves? 

@alaczek, I'm not sure who would make design decisions for Russell! Is that you?

alaczek:
> I'm not sure who would make design decisions for Russell! Is that you?

Yep, that would be me :) 

Russell was one of those single-page themes, and the design didn't extend far enough. Most things beyond the home page were inherited from the parent theme :/ 

I think this change makes a lot of sense! LGTM!

alaczek (approved):

hacchism:
Hi @mikachan @alaczek!

I got a user on 5540847-zen who complained about this change. The user is using portrait(vertical) featured images so now the images are cropped on single post pages, which they don't like. You can confirm that by going to https://russelldemo.wordpress.com/ and clicking on one of the posts under "LATEST STORIES".

Although I totally understand the intention behind the change, at the same time, I understand why the user dislikes the change.

Also, for users on Personal or lower, we can't provide a quick fix with CSS. Can we revert the change? Or at least, there should be a way for free/personal plan users to show their portrait(vertical) featured images without being cropped.

alaczek:
Ugh, sorry to hear this customer experience isn't good because of this change. I think it's for the better overall though, so I wouldn't necessarily revert it. The good news is it should be possible to adjust the single post template via the site editor without custom CSS.

To fix this, you'd need to open up the **Site Editor** and pick the **Single** template to edit, switch to code view and find this bit of code: 
`<!-- wp:post-featured-image {"height":"30rem","align":"wide"} /-->`
change it to
`<!-- wp:post-featured-image {"align":"full"} /-->`
Let me know if you're having trouble with this and I can help!

hacchism:
Oh, I've never thought of removing the parameter using the code view. This perfectly works for now👍 Thank you✨

PR #6504, opened 2022-09-05
Lynx: Update the mesh gradient pattern.

alaczek:
* Use white background color for buttons.
* Adjust font size.


mikachan (approved):
LGTM

Issue #6493, opened 2022-09-01
Update/test fluid typography in block themes 


madhusudhand:
This issue is to update and test the fluid typography in different block themes before the feature goes live in WP 6.1

Following are the list of themes uses clamp typography.

- [ ] Archeo https://github.com/Automattic/themes/pull/6046
- [x] Block-canvas https://github.com/Automattic/themes/pull/6507
- [x] Disco
- [ ] Heiwa
- [x] Lettre https://github.com/Automattic/themes/issues/6492
- [ ] Livro
- [x] Lynx https://github.com/Automattic/themes/pull/6572
- [ ] Meraki
- [ ] Pendant
- [x] Rainfall
- [x] Remote https://github.com/Automattic/themes/pull/6734
- [ ] Skatepark
- [ ] Vivre

pbking:
Considering that these themes already express dynamic font sizing for configured font sizes the only change I think is necessary is to **enable** fluid typography which would allow user-assigned font values to scale and leave the current configuration as-is until/unless changing it would bring further power to the user.

Does that sound like a reasonable compromise?

simison:
Could we have these on top of the list, unless any other reasons to have something else first?

> Lettre
> Lynx
> Livro

Lettre / Lynx are themes for the Newsletters & Link-in-Bio flows we're working on. We've observed issues related to typography that would get solved by this. (e.g. https://github.com/Automattic/themes/issues/6492)

Livro is our default theme for bloggers on WP.com so heaps of customers get it by default and it affects their their first impression, which might be on mobile often.

Shouldn't Blockbase be on the list since it affects most of the new themes on .com?

simison:
Marked Lettre done as per https://github.com/Automattic/themes/issues/6492#issuecomment-1241093228

Looking forward to Lynx theme. :-)

alaczek:
Lynx has a PR open here: https://github.com/Automattic/themes/pull/6572

madhusudhand:
Removing it from the board for now, as the fluid typography tested well with multiple themes and moving all the themes to fluid doesn't bring much value to the users.

PR #6507, opened 2022-09-05
Block-canvas: adding fluid type in theme.json

matiasbenedetto:
Block-canvas: adding fluid type in theme.json

#### Related issue(s):
Part of: https://github.com/Automattic/themes/issues/6493

### DIFF
--- a/block-canvas/patterns/404.php
+++ b/block-canvas/patterns/404.php
@@ -7,8 +7,8 @@
 
 ?>
 
-<!-- wp:heading {"textAlign":"center","level":1,"fontSize":"medium"} -->
-<h1 class="has-text-align-center has-medium-font-size" id="oops-that-page-can-t-be-found"><?php echo esc_html__( 'Oops! That page can&rsquo;t be found.', 'block-canvas' ); ?></h1>
+<!-- wp:heading {"textAlign":"center","level":1,"fontSize":"x-large"} -->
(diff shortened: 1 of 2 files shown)


mikachan (commented):
Looking good 👍

Do we want to add any minimum and maximum values to the existing font sizes? The small font size minimum defaults to 12px, which is quite small, so I'd maybe set a minimum for this at 0.813rem. I'm not too opinionated though!

Also, there are a few static font sizes set in theme.json for specific blocks, such as core/comment-content, core/navigation, core/quote, core/site-tagline, core/site-title, h1, h5, and the main body font size. Should we make these fluid as well?

pbking:
The `fontSize` configured that doesn't use one of the preset values seems to be consistently `1.125rem` including the actual default font size for the page (!). I suggest that it should exclusively be using preset (and probably `preset--medium` for those values).

Should preset values have opinionated max/min?  I don't feel strongly that they should, the value it would add is that it's already there for a cloned theme to further tweak.

It's only for that reason that I would also suggest configuring an opinionated global min/max; not because the default values are quite small, but so that the value is already there to change when this theme is cloned.

mikachan:
> It's only for that reason that I would also suggest configuring an opinionated global min/max; not because the default values are quite small, but so that the value is already there to change when this theme is cloned.

I think this is a good idea - provide the values so that they are more easily customizable when using this as a base theme. Maybe we could add them to just one/a few presets as an example of both opinionated min/max values and default calculations?

pbking:
I've added some more fluidity giving (good default I think) examples of how to configure fluidity and also leveraging the size presets instead of values.

matiasbenedetto:
I like the changes regarding the replacement of the custom font size for the CSS vars :+1: 
The examples can be helpful too.
LGTM

PR #6509, opened 2022-09-06
Pixel: update navigation menu

### LINKED ISSUE #6460: Pixel: Styles: Navigation
Match Navigation menu with designs: Cm51iCQSPu8Ar1qufDTeOQ-fi

* Large screens
image

* Small and medium screens
image

madhusudhand:

Updated site-title font family
image

Navigation on large screens
image

Navigation on small screens
image

#### Related issue(s):
Fixes: #6460


pixel/theme.json
```diff
@@ -251,7 +268,14 @@
             },
             "core/navigation": {
                 "typography": {
-                    "fontSize": "1.125rem"
+                    "fontSize": "0.875rem"
+                },
+                "spacing": {
+                    "blockGap": "1rem",
+                    "padding": {
+                        "left": "0.5rem !important",
```
jffng: Is the `important` is necessary because the root padding values end up overriding the spacing that is set on the block? Is that a bug, or expected?

jffng (commented):
I think the site title does not contain an underline by default, otherwise LGTM. 

This PR | Figma
--- | ---
Screen Shot 2022-09-06 at 11 15 13 AM | Screen Shot 2022-09-06 at 11 15 42 AM

pixel/theme.json
madhusudhand: Without it, navigation inherits root padding leaving more space on the right. adding `!important` to gain higher specificity.

image
image

Created an issue here: https://github.com/WordPress/gutenberg/issues/43931

madhusudhand:
> I think the site title does not contain an underline by default, otherwise LGTM.

Link styles are addressed here. #6517

jffng (approved):

PR #6510, opened 2022-09-06
Pixel: update block styles

madhusudhand:

This addresses the following.

Update `code` block styles
| Before | After |
| --- | --- |
| image | image |

Adds site logo to match the designs.
image

@henriqueiamarino @beafialho 
The above design is achievable and works good only with a square image. 
The UI breaks in the following scenarios.

- horizontal image
image

- vertical image
image

- When menu wraps into multiple lines
image

Suggest if its okay to keep this behaviour or change the alinement such as always align top.

#### Related issue(s):

Relates to: #6460

### DIFF
--- a/pixel/style.css
+++ b/pixel/style.css
@@ -85,3 +85,7 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
 	font-weight: 300;
 	line-height: 1.6;
 }
+
+.pixel-header {
+    flex-grow: 1;
+}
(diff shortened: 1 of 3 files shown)


beafialho:
Thanks @madhusudhand, can the site logo be cropped to always be a squared image? If not, it should be aligned by the top instead of the center.

pixel/style.css
```diff
@@ -85,3 +85,7 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
 	font-weight: 300;
 	line-height: 1.6;
 }
+
```
jffng, line 88:
```suggestion
// Necessary so the header containing the navigation fills the entire remaining horizontal space, since it is positioned in a flex group with the site logo.
```

jffng (approved):

PR #6517, opened 2022-09-07
Pixel: update link styles

### LINKED ISSUE #6456: Pixel: Styles: Links
Update link styles.

Blocks: 
* links
* site-title
* post-title 

States:
* default
* hover
* focus
* active

Design: Cm51iCQSPu8Ar1qufDTeOQ-fi

image

madhusudhand:

| block/state | default | hover | active | focus |
| --- | --- | --- | --- | --- |
| link | image | image | image | image |
| site-title | image | image | image | image |
| post-title | image | image | image | image |

Pagination:
image

#### Related issue(s):

Fixes: #6456

### DIFF
--- a/pixel/theme.json
+++ b/pixel/theme.json
@@ -276,6 +276,20 @@
                         "left": "0.25rem !important",
                         "right": "0.25rem !important"
                     }
+                },
+                "elements": {
+                    "link": {
+                        ":active": {
+                            "typography": {
+                                "textDecoration": "underline dotted"
+                            }
+                        },
+                        ":focus": {
+                            "typography": {
+                                "textDecoration": "underline solid"
+                            }
+                        }
+                    }
                 }
             },
             "core/post-date": {
@@ -300,11 +314,6 @@
                     "link": {
                         "typography": {
                             "textDecoration": "none"
-                        },
-                        ":hover": {
-                            "typography": {
-                                "textDecoration": "underline"
-                            }
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


pixel/theme.json
```diff
@@ -276,6 +276,41 @@
...
+                        },
+                        ":focus": {
+                            "typography": {
+                                "textDecoration": "underline solid"
+                            }
+                        }
+                    }
+                }
+            },
+            "core/post-content": {
```
jffng, line 295: Based on the Figma, links inside post content should be italicized in :active, :focus:, :hover states. I added a config for that here, but I noticed that it broke the block margins even though it should be a valid config: 

Screen Shot 2022-09-07 at 10 39 28 AM

Any idea why? I will investigate and open a bug report upstream if needed, but we may need to find a workaround in the meantime.

jffng (commented):
In the Figma, category links in the post list do not have underlines: 

Screen Shot 2022-09-07 at 10 35 53 AM

It seems inconsistent but wanted to confirm.

pixel/theme.json
jffng: Reported here: https://github.com/WordPress/gutenberg/issues/43955

madhusudhand: As a workaround added explicit `spacing -> blockGap` in `post-content` to fix the issue.

madhusudhand:
> In the Figma, category links in the post list do not have underlines:
> 
> Screen Shot 2022-09-07 at 10 35 53 AM
> 
> It seems inconsistent but wanted to confirm.

@jffng This behaviour is achieved using additional css. It is now matching figma.

@beafialho @henriqueiamarino can you confirm if it is expected to have different default styling for categories and tags?

henriqueiamarino:
Figma is updated @madhusudhand... Categories and Tags follow the same styling.

jffng (approved):

PR #6520, opened 2022-09-07
Pixl: refine the comments.

### LINKED ISSUE #6459: Pixl: Styles: Comments
Update comments block styles.

Design: Cm51iCQSPu8Ar1qufDTeOQ-fi

image

And update the font settings of the comments block.

Current state:
image

jffng:

Modifies the comments form and comment reply styles.

Screen Shot 2022-09-07 at 11 52 00 AM

@beafialho @henriqueiamarino could you advise on the above screenshot?  

I did not adjust the layout of the comments form since it would require CSS and is not editable by the user, do you think its layout is acceptable as is? 

#### Related issue(s):

Fixes #6459

### DIFF
--- a/pixel/patterns/comments.php
+++ b/pixel/patterns/comments.php
@@ -6,56 +6,39 @@
  */
 
 ?>
-<!-- wp:comments-query-loop -->
-<div class="wp-block-comments-query-loop">
-	<!-- wp:comments-title {"level":3} /-->
 
-	<!-- wp:comment-template -->
-	<!-- wp:columns -->
-	<div class="wp-block-columns">
-		<!-- wp:column {"width":"40px"} -->
-		<div class="wp-block-column" style="flex-basis:40px"></div>
-		<!-- /wp:column -->
+<!-- wp:comments {"className":"wp-block-comments-query-loop "} -->
+<div class="wp-block-comments wp-block-comments-query-loop"><!-- wp:comments-title {"level":3} /-->
 
-		<!-- wp:column -->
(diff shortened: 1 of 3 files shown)


henriqueiamarino:
Hey @jffng I assume the layout can go as it is but I'd like to learn about the spacing marked in red. Is that editable? If yes I can help you with that.
An extra question, just as I am still new to that: Is it possible to enlarge the **comment's section** width a little? This way we wouldn't have that single word suspended as a widow.
response-feedback

jffng:
@henriqueiamarino 

>  Is that editable?

Yes!

> Is it possible to enlarge the comment's section width a little?

We could make the comments section wider, but in that case it wouldn't align with the post content above it. What do you think?

beafialho:
> We could make the comments section wider, but in that case it wouldn't align with the post content above it. What do you think?

The comments section should align with the post content. I'd just make sure that text is 12px. About it being a widow, I believe that could be unavoidable if a screen is scaled up or down @henriqueiamarino.

One detail: the tick box should match the other boxes.

Captura de ecrã 2022-09-08, às 16 11 06

henriqueiamarino:
@jffng I believe this would make spacing more coherent.
response-feedback

madhusudhand:
Updated the spacing and fonts.

image

madhusudhand:
Changes looks good. Bringing it in. Any more changes will be addressed after the theme review by designer.

madhusudhand (approved):

PR #6523, opened 2022-09-08
Varia buttons: Fix custom text color when button background color is set to "foreground" or "background"

### LINKED ISSUE #3365: Varia: Issues with button colors
## Steps to replicate

If you activate Stratford and choose to use the theme's homepage, the issue seen in the screenshot below will be replicated. 

Some button colors aren't showing properly on Stratford. They look fine in the editor, but not on the live site. 

This issue can be seen on our current Stratford theme demo: https://stratforddemo.wordpress.com/
Screen Shot 2021-02-25 at 3 47 31 PM

As well, there seems to be an issue with different states. I couldn't replicate this but have seen this in multiple reports (listed below). 

Normal state: 
Screen Shot 2021-02-25 at 3 39 41 PM

Active state: 
Screen Shot 2021-02-25 at 3 39 52 PM

Visited state: 
Screen Shot 2021-02-25 at 3 40 05 PM

## Result

Button isn't reflecting editor, active/hover/visited states are set to colors that cause visibility issues. 

## Expected

That the active, hover, and visited should be visible or there should be an option to set those separately. 

* User report: zd-3761048, hc-28020210

yashitamittal11:
**Description**

If the user selects the custom text color and background color for the button, the custom text color does not show on the live page. The color looks fine in the editor though.

**Changes proposed in this Pull Request:**

This PR ensures the custom text color selected by the user shows on the live site.

**Test instructions**

1. Add a button on a page.
2. Select a custom text color for the button.
3. Set the button background to "foreground" or "background" color.
4. The custom text color shows correctly in the editor. However, the color on the live page is not correct.

Note: The custom text color shows correctly if the user selects another color from the color palette. As in this case, the color is added via inline CSS. The issue is only visible when the user chooses the default theme color from the theme color palette.

**Also, this PR addresses only one part of the issue with the buttons on the Varia theme.**

| Before | After |
|---|---|
|  | Varia colors fixed |

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/3365

### DIFF
--- a/alves/style-rtl.css
+++ b/alves/style-rtl.css
@@ -6,7 +6,7 @@ Author: Automattic
 Author URI: https://automattic.com/
 Description: Convincing design for your charity or organization’s online presence. Highlight your actions, causes and projects, Alves is versatile enough to be your personal site too.
 Requires at least: WordPress 4.9.6
-Version: 1.5.20
+Version: 1.5.21
 License: GNU General Public License v2 or later
 License URI: LICENSE
 Template: varia
@@ -2461,7 +2461,7 @@ table.is-style-stripes tbody tr:nth-child(odd),
 .has-background h4:not(.has-text-color),
 .has-background h5:not(.has-text-color),
 .has-background h6:not(.has-text-color) {
-	color: currentColor;
+	color: currentcolor;
 }
 
 .has-primary-background-color,
@@ -2491,8 +2491,8 @@ table.is-style-stripes tbody tr:nth-child(odd),
 }
 
 .has-background-dim,
-.has-foreground-background-color,
-.has-foreground-background-color.has-background-dim {
+.has-foreground-background-color:not(.has-text-color),
+.has-foreground-background-color.has-background-dim:not(.has-text-color) {
 	color: #ffffff;
 }
 
@@ -2539,7 +2539,6 @@ table.is-style-stripes tbody tr:nth-child(odd),
 .has-background-background-color,
 .has-background-background-color.has-background-dim {
 	background-color: #ffffff;
-	color: #394d55;
 }
 
 .has-background-background-color:not(.has-text-color),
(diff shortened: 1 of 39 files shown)


madhusudhand (changes requested):
Changes looks good.

@yashitamittal11 Can you run `npm run build` and `npm run children` inside `varia` directory and include those changes?

yashitamittal11:
@madhusudhand Should I commit the following files?
style-rtl.css
style.css

yashitamittal11:
I have committed the style-rtl.css and style.css files for all Varia's child themes except for Stow.
For Stow, style-rtl.css is not generated via this command "npm run children."

madhusudhand:
> For Stow, style-rtl.css is not generated via this command "npm run children."

Thanks for pointing. There seems to be a syntax issue that got introduced in a previous commit. 
Can you remove the quotes around the closing `}` [here](https://github.com/Automattic/themes/blob/f17c82bcdf79b0dda6af85afa6ee3e68884b20f9/stow/sass/style-child-theme-editor.scss#L177) and re-run the command to include the stow changes too.

```font-family: #{map-deep-get($config-global, "font", "family", "code")};```

yashitamittal11:
>Can you remove the quotes around the closing and re-run the command to include the stow changes too.

Thanks for the insight. I rerun the command after removing the quotes. I see three modified files now - 
1. ../stow/sass/style-child-theme-editor.scss
2. ../stow/style-rtl.css

The build job has added many formatting changes in the style-editor.css and style-rtl.css files. Is it okay to commit the files with these formatting changes?

style-rtl- stow

madhusudhand:
>  I see three modified files now 

Yes, that is expected.

> Is it okay to commit the files with these formatting changes?

Yes. you can commit these changes as well.

yashitamittal11:
Thanks. I have committed all the files.

madhusudhand (approved):
Thanks for working on this. Looks good to me.

yashitamittal11:
>Also, this PR addresses only one part of the issue with the buttons on the Varia theme.

This PR addresses only one part of the problem discussed in the original issue. Should we reopen the original issue or create a new GitHub issue for the unsolved problem?

madhusudhand:
I would suggest to keep the issue open and mention the parts that are fixed with this PR.

PR #6524, opened 2022-09-08
Pixel: revise page to account for featured image.

### LINKED ISSUE #6519: Pixel: featured image on pages
Should the page template include the featured image block? If so, what should its placement be? 

I assume same as the single template, but want to confirm, and if so we need to adjust the page.html.  

cc @beafialho @henriqueiamarino

jffng:

Before | After
--- | ---
screencapture-gutenbergtest-local-sample-page-2022-09-08-10_45_24 | screencapture-gutenbergtest-local-sample-page-2022-09-08-10_45_08

#### Related issue(s):
Fixes #6519

### DIFF
--- a/pixel/templates/page.html
+++ b/pixel/templates/page.html
@@ -1,18 +1,17 @@
 
 <!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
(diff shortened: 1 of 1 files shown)


vcanales (approved):
LGTM

PR #6531, opened 2022-09-09
Pixl: add duotone for images

madhusudhand:

image

Duotone applied to the image is also applied to the border, and the shadow stays as primary color thus creating a difference.

Created an upstream issue here: https://github.com/WordPress/gutenberg/issues/44184

@henriqueiamarino to confirm on this behavior.


madhusudhand:
@henriqueiamarino the created issue is closed because it might introduce additional complexity. https://github.com/WordPress/gutenberg/issues/44184#issuecomment-1248800288

Please suggest if we should add this duotone (which causes two different border colors) or not add the duotone.

PR #6542, opened 2022-09-13
Create/decibel

### LINKED ISSUE #6544: Decibel: Create Base Theme
Just make it exist and give us a place to do the good work.

Added here with some reasonable defaults to start with.  Nothing set it stone.

pbking:
This change is a story in two parts:

Block Canvas: 
* removed the `--custom--gap` values and replaced with `preset--spacing` values
* Removed redundant (and some opinionated) style configuration

These changes make the theme a slightly better starting point for modern themes.  (This is the part I'm most interested in for this change.  Is there anything else that could/should change to make this a better starting point?)

Ideally as the FIRST phase of development continues for Decibel this theme can continue to be modified to keep as many changes as possible in Block Canvas (benefitting future theme builds).

Decibel:
* Add Decibel as a CHILD theme of Block Canvas
* Add a (few) settings to make the theme unique and as a place to begin work.
* Add the base fonts

None of the opinions expressed in the theme are currently expected to be finished; they were just added as a starting point.

image


vcanales (commented):
👍 to the Decibel starting point.

> These changes make the theme a slightly better starting point for modern themes. 

Is this because it makes the theme a bit less opinionated? The difference I _think_ I see is that the gap would become static instead of fluid; what other advantages am I missing?

jffng (approved):
In general I appreciate the initiative to improve our starting point, and think the effort to use standard sizes in Block Canvas is definitely an improvement. 

This is my own personal bias, as well as recency bias working on Pixl and finishing it basically in two weeks, but I'm not so convinced of the value this approach will provide, versus potentially slowing down our development. I feel that every opinion (styles and templates) can now quickly be expressed starting from scratch, or a clone of a very barebones theme which Block Canvas already is. Having to manage changes in parent and child, and consider with each PR which should be an opinion of the parent vs child, feels like we could arrive at a different decision with every theme that's developed based on its particularities, making the effort a bit circular. 

That said, I'm open to trying this approach if you think there's value. I think one area that this PR does not touch that could use improvement in Block Canvas is the comments. Henrique had a nice [improvement](https://github.com/Automattic/themes/pull/6520) to the spacing and sizing overall in Pixl that I think could be worked in. Happy to bring that into block canvas in its respective decibel issue, and merge this as is.

pbking:
>  I feel that every opinion (styles and templates) can now quickly be expressed starting from scratch, or a clone of a very barebones theme which Block Canvas already is. Having to manage changes in parent and child, and consider with each PR which should be an opinion of the parent vs child, feels like we could arrive at a different decision with every theme that's developed based on its particularities, making the effort a bit circular.

Thank you very much for sharing that.

Ideally this part of the process won't actually take a lot of overhead and I hope to "break" from being a child early (in just a couple of days at most).  I noticed that Block Canvas had some cruft that I imagine would have (or should have) been removed on nearly any theme cloned and the gap stuff especially seemed like bad practice to continue using.  Considering that we aren't the only team likely to use Block Canvas (we've suggested it to nearly anyone who had been building Blockbase children) we should take regular opportunities to keep that theme as "standards bases" as possible.  Keeping it light and airy with few opinions is ideal and more than putting new things INTO Block Canvas I expect that if we change any more it will likely be to take things OUT of Block Canvas as we develop.

I like your idea of bringing in the Comments changes you mention; I'll open an issue to do just that.  Thank you!

PR #6565, opened 2022-09-13
Introduced some padding to Block Canvas' 404 template 

### LINKED ISSUE #6556: Decibel: Templates: 404
There were no comps in the figma for 404 page so I was planning to just roll the default and see how that shook out in the design review. However the default was looking a mite crowded so instead of making a new template I updated the Block Canvas template.

See #6565 for details

pbking:
(to be used in Decibel)

There were no comps in the figma for 404 page so I was planning to just roll the default and see how that shook out in the design review.  However the default was looking a mite crowded so instead of making a new template I updated the Block Canvas template.

Fixes #6556 

Before:
image

image

image
image

After:
image
image
image
image

### DIFF
--- a/block-canvas/parts/header.html
+++ b/block-canvas/parts/header.html
@@ -24,6 +24,6 @@
 </div>
 <!-- /wp:group -->
 
-<!-- wp:spacer {"height":var(--wp--preset--spacing--50)} -->
-<div style="height:var(--wp--preset--spacing--50)" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- wp:spacer {"height":"var(--wp--preset--spacing--60)"} -->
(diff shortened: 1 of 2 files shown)


jffng (commented):
This looks good in the view, but getting a couple block validation errors in the site editor: 

Screen Shot 2022-09-13 at 5 45 54 PM

Running Gutenberg trunk.

pbking:
Ah, I think those are the spacers in the headers and footer templates; recently transitioned from using values to using preset spacers.  I forgot that was an issue and didn't review that in the editor before it merged.  I don't recall if/how this can be done with a spacer.  If there's not a way to do it with the spacer perhaps we could configure the space with margin?

pbking:
Ok, so I was kinda right kinda wrong about the cause of the block validation errors.  Mostly just needed to fix the size variables.

All workin' now so gonna bring this in.

PR #6567, opened 2022-09-13
Decibel: add button styles

### LINKED ISSUE #6549: Decibel: Styles: Buttons
Style 'em buttons.

Just the `element`, not the block.

Default         |         Hover             |      Focus         |             Active
image
There are TWO buttons styles in the design.

Neither of them are "outline" style.  Should we include one anyway?  Should we just call the second style "outline"?  Add a 3rd block style option for an "inverted button style"?

vcanales:
Do not merge; waiting on confirmation of Outline Button design.

Kapture 2022-09-13 at 16 22 15

#### Related issue(s):
https://github.com/Automattic/themes/issues/6549

### DIFF
--- a/block-canvas/functions.php
+++ b/block-canvas/functions.php
@@ -43,7 +43,7 @@ if ( ! function_exists( 'block_canvas_styles' ) ) :
 		// Register theme stylesheet.
 		wp_register_style(
 			'block_canvas-style',
-			get_template_directory_uri() . '/style.css',
+			get_stylesheet_directory_uri() . '/style.css',
 			array(),
 			wp_get_theme()->get( 'Version' )
 		);
--- a/block-canvas/style.css
+++ b/block-canvas/style.css
@@ -50,3 +50,13 @@ GNU General Public License for more details.
 	padding-right: var(--wp--preset--spacing--50);
 	padding-left: var(--wp--preset--spacing--50);
 }
+
+/*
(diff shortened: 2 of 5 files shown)


pbking:
I swapped out the blue buttons for black since the blue isn't used in any of the designs.

image

Hover:
image
image

To get the outline button styles indeed we have to go to CSS.  I also added that CSS to Block Canvas as it's a common need with an eventual fix.

No matter what I do I can't get the outline style to render differently as it seems to show in your illustration.  Any idea what I might be doing differently?

vcanales:
> No matter what I do I can't get the outline style to render differently as it seems to show in your illustration. Any idea what I might be doing differently?

Do you mean the `outline` applied on focus? I've been using dev tools to activate :focus manually, since buttons within posts don't seem to be focusable through `tab`.

pbking:
Yea the outline applied on focus.  I'm not seeing it in either chrome or firefox neither when I focus with tab or devtools.  I don't see the properties being set.  I see other focus values (background color for example) but not outline.

pbking:
Bah, it was my outdated version of Gutenberg.  I didn't realize that had been added so recently.

I see it now. :)

pbking:
Now that I'm on the right version of Gutenberg this looks good now.  Gonna go ahead with the black button and not implement the blue.  If we find a need for it later we'll figure that out then.

PR #6571, opened 2022-09-14
Decibel: add typography styles

### LINKED ISSUE #6546: Decibel: Styles: Typography — Fonts & Weights & Sizes & Line heights
Bring in all the necessary font resources to style the theme.  (A very rough start was initally added but more work needs to be done for sure).

Configure the weights.
Configure the sizes (note the variable sizes for 3 headings and 2 text fonts, the rest should be static sizes)

Rubik is used for Headings (all sizes) 
DM Mono is used for text

Rubik is available as a variable-weight font but I'm not sure if we can achieve the design with that. 

Unsure if italic typefaces will be necessary (hopefully not).

Currently `.ttf` resources are in the project but we should probably replace those with `.woff2` if possible.

image

(variance values for mobile)
image

image

(variance values for mobile)
image

jffng:

Adds sizes, weights, and line heights for Decibels headings and paragraph elements.

#### Related issue(s):

Fixes #6546


jffng:
A few notes: 

- In the designs, the large and extra large font size presets have different weights. Is this something we should do via CSS? 
- The designs seem to suggest the line height changes for H2s from mobile to desktop. Is that something we want to do?
- I did not swap out the font files themselves in this PR 
- I could not get the UI for selecting the XL font size to work, but maybe that's something wrong on my end. The fontSizes config seems to be correct.

decibel/theme.json
```diff
@@ -145,7 +152,44 @@
...
+			},
+			"h2": {
+				"typography": {
+					"fontSize": "clamp(2rem, 0.15384615384615374rem + 3.8461538461538463vw, 4rem)",
+					"lineHeight": "1.1"
+				}
+			},
+			"h3": {
+				"typography": {
+					"fontSize": "clamp(1.75rem, 1.2884615384615383rem + 0.9615384615384616vw, 2.25rem)",
```
pbking: Can we use `preset--x-large` here?

jffng: Oh good call, done here: https://github.com/Automattic/themes/pull/6571/commits/659afa987b73334fae3c5164653647eb17b4b0fe

pbking:
> In the designs, the large and extra large font size presets have different weights. Is this something we should do via CSS?

I don't see a difference between wights in the design (perhaps it was changed? Or perhaps it's a matter how how the two systems render fonts differently?). The large headers show 500 and the large text presets show 300.  If we need to though, perhaps we can make the weight a dynamic calculated value like we do font size.  I think my 'druthers is to just stick with static values.

> The designs seem to suggest the line height changes for H2s from mobile to desktop. Is that something we want to do?

This difference I AM seeing in the designs and perhaps we should more strongly consider a dynamic value for the line height (at least for the headings which we can easily put into `theme.json`).

> I could not get the UI for selecting the XL font size to work

Took me a hot minute to figure out but I deduced that the cause was due to the `large` and `x-large` font sizes being exactly the same.  `x-large` was dynamic, `large` was not, so the resulting size was different, but the "starting size" was the same so it got confused.  I fixed it by tweaking the `large` font size and putting the `x-large` starting size between the min/max values.

About to drop in a couple of fixes.

pbking:
Ok, tinkered with the dynamic line height and realized that the sizes were inverted (line height gets large for smaller screen sizes) and that might be possible with some clamp math but it's outside if my ability.  Gonna shelve that idea for now.

pbking:
Last bit, I added the woff2 resources instead of the variable font (and it came out to half the size).  Changed the headings font weight from 700 (which looked right on variable font) to 500 (which was in the figma and looks right with the new resources).

Bringin' this in!

PR #6573, opened 2022-09-14
Heiwa: Update navigation width

iamtakashi:
This intends to override the width of the navigation that is currently forced to `wide` width in Gutenberg. Ref. https://github.com/WordPress/gutenberg/issues/44182

| Old | New |
| ----------- | ----------- |
| Screen Recording 9-20-2022 at 7 05 PM | Screen Recording 9-20-2022 at 7 03 PM |

### DIFF
--- a/heiwa/assets/theme.css
+++ b/heiwa/assets/theme.css
@@ -1,0 +1,28 @@
+/**
+ * It looks like the responsive navigation is forced to be the wide width. https://github.com/WordPress/gutenberg/pull/43576
+ */
+
+.wp-block-navigation__responsive-close {
+	max-width: none;
+}
+
+/**
+ * Override the default padding value from Gutenberg to avoid the icon jumping
+ */
+.wp-block-navigation__responsive-container.is-menu-open {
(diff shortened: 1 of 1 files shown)


heiwa/assets/theme.css
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Override the default padding value from Gutenberg to avoid the icon jumping
+ */
+.wp-block-navigation__responsive-container.is-menu-open {
+	padding-top: var(--wp--custom--gap--vertical);
+}
+
+/**
+ * Override Blockbase ponyfill style that makes navigation items left aligned.
```
madhusudhand: It seems to be an issue with all blockbase children, and having a fix in blockbase would be more appropriate.

The following should work without breaking child themes.

`_navigation.scss`, Lines 32 and 39
```
align-items: var(--navigation-layout-justification-setting, flex-start);
```

iamtakashi:
Hi @madhusudhand, 

I've adjusted this PR accordingly with the recent updates in Blockbase.

madhusudhand (approved):
LGTM!

PR #6577, opened 2022-09-14
Rainfall: fix navigation link colors

matiasbenedetto:
Rainfall: fix navigation link colors

Before:

https://user-images.githubusercontent.com/1310626/190247752-30d417aa-0933-40f9-a1cb-e6a88b8207b7.mp4

After:

https://user-images.githubusercontent.com/1310626/190248116-d0677b36-8b34-4556-bce2-15613eeceea4.mp4


madhusudhand (commented):
Top menu looks good.

image

But the menu from footer has couple of issues.

image

* Menu items from level 2 are not visible
* Menu dropdown appears below the menu and so it is not accessible. (something to address in Gutenberg)

matiasbenedetto:
I added the right color for the bottom submenus. If you are using a classic menu (for example if you are using the content from theamdemo)  you will not be able to see the right color because of a bug in the navigation block. This bug was reported here: https://github.com/WordPress/gutenberg/issues/44235

To see the right colors you need to create a new navigation menu and assign it to the navigation block.

PR #6579, opened 2022-09-15
Tools: add vscode settings to format files on save

madhusudhand:

* This formats the file contents on save.
* Also, replaces tabs with 4 spaces for consistency. (found few inconsistent spaces/tabs between developers while working on pixl)
* Removes extra spaces at the end, but leaves single new line

#### Testing

* Verified with html templates, css, scss, json files. It works.
* HTML inside php (patterns) are not formatted because html is ignored inside php.

We might see unwanted changes while working on few themes (all those new/deleted spaces), but over the time they all should be consistent.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,8 @@
+{
+    "editor.formatOnSave": true,
+    "files.insertFinalNewline": true,
+    "files.trimFinalNewlines": true,
+    "editor.detectIndentation": false,
+    "editor.tabSize": 4,
+    "editor.insertSpaces": true,
+}


vcanales (approved):
Thanks! LGTM.

pbking:
I'm not too big of a fan to be honest.

Tabs > Spaces; I thought that was the WP standard so would prefer we standardize on those.  I believe that theme.json generation uses spaces though, so there's likely a lot of that in there so maybe standardizing on spaces would be better for those resources?

While I like auto-formatter in principal it also makes small changes to files that don't conform a lot more difficult to review and the "auto-ness" of it ensures that those changes are coupled together (the fix and the formatting). That said we already have some auto-formatting on git commit hooks so we're already dealing with that.

matiasbenedetto:
:wave: I think we all agree that automatic code formatting is a useful thing to have and also in the fact that's not pleasant to integrate it into a large codebase. I'm agnostic about tabs/space or any other config, at the end of the day, the important objective is the style consistency and not the style itself. I believe that we should implement some kind of code formatting utility in this repo but not specifically this because of these reasons:

- It's only for one particular code editor (VS Code) and it will not work if someone uses other editors.
People using the editors will commit code formatted differently.

- You need to open the root project folder to make this work.
In this repo, I usually open my editor in the particular theme folder I'm working on. This config doesn't work because the config file simply is not there.

- I'm not sure about this and I consider it less important: It doesn't follow the WordPress code style. 

In my opinion, a good formatting solution should:
- Works independently of the code editor
- Can be implemented progressively
- Follows the WordPress style

I would prefer to revert this change until we land a better solution.
What do you think @Automattic/theam ?

pbking:
I agree with all of the above.  Auto-formatting is important and helpful but needs to be done very carefully.  

I will revert this and we can find another avenue.

pbking:
https://github.com/Automattic/themes/pull/6588

PR #6580, opened 2022-09-15
Blockbase: Update navigation justification

iamtakashi:
As per https://github.com/Automattic/themes/pull/6573/files#r971652726, this intends to update Blockbase to honour the justification setting.

| OLD | NEW |
| ----------- | ----------- |
| iamtakashi jurassic tube_wp-admin_site-editor php_postType=wp_template postId=blockbase%2F%2Findex (1) | iamtakashi jurassic tube_wp-admin_site-editor php_postType=wp_template postId=blockbase%2F%2Findex |

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -246,12 +246,14 @@ a,
 a:hover, a:focus,
 .wp-block-navigation a:hover,
 .wp-block-navigation a:focus {
-  text-decoration-line: underline;
+  -webkit-text-decoration-line: underline;
+          text-decoration-line: underline;
 }
 
 main.block-editor-block-list__layout a,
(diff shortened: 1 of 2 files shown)


madhusudhand (approved):
Included a small fix to handle the menus of level 2 and higher.
Verified the changes on child themes, the menu alignment works as expected.

However it has a side effect. The alignment of menus in the child themes such as `antonia` and `blockbase` itself changes to right because they have the default alignment as "right" in the template. So the menu shifts right otherwise it was left aligned.

`<!-- wp:navigation {"layout":{"type":"flex","justifyContent":"right"}} /-->`

But it comes to individual themes to have a fix to update the alignment in their header template, and in a separate fix.

@pbking Please share your thoughts.

PR #6584, opened 2022-09-16
Decibel: Add single template

### LINKED ISSUE #6557: Decibel: Templates: Single
There are currently no designs for this.  At the moment a featured image is full width with the following page heading:

image
image

We need some additional direction for pages.

jffng:

Adds a single template for Decibel.

#### Related issue(s):

Fixes #6557

### DIFF
--- a/block-canvas/templates/single.html
+++ b/block-canvas/templates/single.html
@@ -10,7 +10,7 @@
 <main class="wp-block-group">
 <!-- wp:post-featured-image {"align":"full"} /-->
 
-<!-- wp:post-content {"layout":{"inherit":true},"lock":{"move":false,"remove":true}} /-->
+<!-- wp:post-content {"layout":{"type":"constrained"},"lock":{"move":false,"remove":true}} /-->
 
 <!-- wp:template-part {"slug":"post-meta"} /-->
 </main>
--- a/block-canvas/theme.json
+++ b/block-canvas/theme.json
@@ -188,10 +188,6 @@
                     "margin": {
                         "bottom": "0"
                     }
-                },
-                "typography": {
-                    "fontSize": "var(--wp--preset--font-size--large)",
-                    "lineHeight": "1.125"
                 }
             },
             "core/pullquote": {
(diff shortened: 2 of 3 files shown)


block-canvas/templates/single.html
```diff
@@ -10,7 +10,7 @@
 <main class="wp-block-group">
 <!-- wp:post-featured-image {"align":"full"} /-->
 
-<!-- wp:post-content {"layout":{"inherit":true},"lock":{"move":false,"remove":true}} /-->
+<!-- wp:post-content {"layout":{"type":"constrained"},"lock":{"move":false,"remove":true}} /-->
```
jffng, line 13: Thought it would be good to update this to the latest syntax.

block-canvas/theme.json
```diff
@@ -188,10 +188,6 @@
                     "margin": {
                         "bottom": "0"
                     }
-                },
-                "typography": {
-                    "fontSize": "var(--wp--preset--font-size--large)",
```
jffng, line 193: This was overriding the post title default sizes for H2s in the template, so I removed it.

decibel/theme.json
```diff
@@ -32,7 +32,7 @@
 			]
 		},
 		"layout": {
-			"contentSize": "664px",
+			"contentSize": "1195px",
```
jffng: I'm not sure about this, it's what the design implies but is quite a wide width from a readability perspective. Also should there be any difference between this and the wideSize? 

cc @henriqueiamarino

jffng:
One other issue that I'm not sure if is intentional — the design currently contains no meta (author, date, categories, tags). Is this something we want to leave out or include? @henriqueiamarino.

decibel/theme.json
pbking: I'm gonna leave it at 664 for now.   I definitely think there should be a difference between the two.  I'll assume that content was just wide width.

pbking:
> One other issue that I'm not sure if is intentional — the design currently contains no meta (author, date, categories, tags). Is this something we want to leave out or include? @henriqueiamarino.

This theme doesn't feel very "bloggy".  The exclusion feels intentional and (personally) I like it better without.

Gonna bring this in now without that content.  We'll add it later if it's deemed necessary.

PR #6585, opened 2022-09-16
Added large illustration patterns

### LINKED ISSUE #6568: Decibel: Patterns: Illustrations
each of these illustrations should be available as a pattern (probably in a category 'Decibel Illustrations' and 'Decibel Illustrations (square)' or something like that:

Illustrations
image

Illustrations (square)
image

Our goal is for them to use Global Style Colors in the illustrations.  I notice that all of the assets are vector, so I'm hoping they can be stored as SVG (in the pattern markup rather than as media assets) and that SVG leverage CSS variables for the colors.

pbking:
image

This change adds the 16 SVG resources and 16 patterns as well as adding the category "Illustrations".

Closes #6568 

The small illustrations haven't been included as none of them are used in the designs.  We can add them later if found necessary.

NOTE: I tried VERY HARD to load these resources in a way that would allow us to style them with Global Styles.  
I got very close.
But not close enough.

If/when this theme gets variations we should revisit that.  It would be really cool if it worked...


jffng (commented):
These are all loading correctly, but for some reason they are showing up as Uncategorized for me, although from your screenshot it looks to be loading in the correct category ("Illustrations"). 

Post Editor | Site Editor
--- | ---
Screen Shot 2022-09-16 at 10 01 01 AM | Screen Shot 2022-09-16 at 10 04 18 AM

Other than that, I have more of a general question about about including all of these illustrations as block patterns. Is there a better way to distribute these lovely illustrations than bundling them with the theme? Should we be bundling them with the theme? 

A previous version of me would have said yes absolutely, this is what makes the theme unique. But more and more as we go towards global styles and themes are wholly configurable to the user's needs and taste, I don't think we should include assets in a theme that can't be customized via the editor (at least yet). Not to mention, it adds about 3mb to the theme. 

Curious what you and others think.

pbking:
I had the same questions, and to be honest probably feel about the same way that you do.  I love the illustrations, but I'm not sure that they offer enough to the theme for the size involved, and I have the same "portability" concerns.  Somewhat.  I'm still keen on themes offering something unique, and if these could have integrated better (color changing with global styles) I would have been all for including them.  

But rather I think we should include those found in the designs and none else.

pbking:
I reduced this down to the two actually used.  I had failed to commit the change that added the category so I've pushed that change as well.

image

PR #6587, opened 2022-09-16
defaulted the vscode settings to tabs over spaces

pbking:
I'm a fan of standards.  I'm not too keen on [IDE settings being in the project.](https://github.com/Automattic/themes/pull/6579)

At a minimum though we should be defaulting to tabs not spaces.

My (pretty strong) 'druthers is to remove the configuration completely and keep those settings in .gitignore but I can probably get over it.


pbking:
Closing as the origination of this change was reverted.

PR #6598, opened 2022-09-20
Decibel is no longer a Block Canvas child

pbking:
Assets copied from the (previous) parent Block Canvas into the decibel folder.  Namespaces changed where necessary.

The theme should behave exactly the same as prior to this change.


vcanales (approved):
It looks good to me!

Issue #6569, opened 2022-09-13
Loudness: Home Page Patterns


pbking:
Patterns to complete for the "Home" template:
- [ ] Pricing Table:
 
<img width="958" alt="image" src="https://user-images.githubusercontent.com/146530/189986934-1b876068-6b92-4926-aa30-7587b7048134.png">

- [ ] Guest Performers:
<img width="839" alt="image" src="https://user-images.githubusercontent.com/146530/189987045-cdf53f09-717b-41de-ab6f-462000e10ac0.png">

- [ ] Instructors:
<img width="530" alt="image" src="https://user-images.githubusercontent.com/146530/189987152-d3de5516-b94b-4afe-808b-8986d7bb960c.png">

PR #6603, opened 2022-09-20
Decibel: add pricing table

jffng:

Adds the pricing table pattern: 

Screen Shot 2022-09-20 at 2 41 22 PM

A few notes:

- The font weight of the dollar value is off because I couldn't affect the font-weight there, maybe we need to load more weights? 
- I tried this out with an h-stack / group block instead but it doesn't wrap well. 
- The button is kind of centered in the comps, but I am not sure that design scales so well so I just left aligned it. 

#### Related issue(s):

Part of #6569


vcanales (approved):
LGTM! 

Kapture 2022-09-20 at 21 07 53

PR #6606, opened 2022-09-21
Loudness: Add quote styles 

### LINKED ISSUE #6553: Loudness: Styles: Quote
No Quote/Pullquote styles in design.  I believe we need some input:

By default:

Pullquote:
image

Quote:
image

vcanales:

Add styles for the Quote and Pullquote blocks.

Also, I'm approaching the Pullquote design in two ways; either as a Pattern, or by styling the block with CSS. The reason for this is that I find the CSS to get the default Pullquote block to look like we want it to a bit too involved. They look identical, except for the paragraph spacing on the Pattern one:

image

I would like to get your opinion on which option we should adopt. I'm leaning towards Patterns, as it allows us to show off how simple it is to implement this, as well as being safer in terms of customization — I'm not sure how well the CSS solution survives user customization.

#### Related issue(s):
Fixes: #6553

### DIFF
--- a/loudness/style.css
+++ b/loudness/style.css
@@ -86,3 +86,49 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
 .wp-block-navigation.wp-block-navigation .wp-block-navigation__responsive-container.is-menu-open .wp-block-navigation__container {
 	width: 100%;
 }
+
+/* Quote styles */
+.wp-block-quote {
+	font-size: var(--wp--preset--font-size--large);
+}
+
+.wp-block-quote cite {
+	text-transform: uppercase;
+	font-style: normal;
+	font-size: var(--wp--preset--font-size--medium);
+}
+
+.wp-block-quote cite::before {
+	content: "— ";
+}
+
+
+/* Pullquote Styles */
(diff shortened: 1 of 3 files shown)


jffng (commented):
I agree with the way you're leaning — the pattern seems like the way to go here, considering how much CSS is needed to get the default pull quote to appear as designed.

vcanales:
We have 3 styles then! 
In the Figma file there's a Large Quote design, but this can be done by changing the font size of the Block in the editor, so I don't think it warrants any styling on this end. Going ahead like this when we get it ✅'d

image

loudness/style.css
```diff
@@ -86,3 +86,19 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
 .wp-block-navigation.wp-block-navigation .wp-block-navigation__responsive-container.is-menu-open .wp-block-navigation__container {
 	width: 100%;
 }
+
+/* Quote styles */
+.wp-block-quote {
+	font-size: var(--wp--preset--font-size--large);
```
jffng: Can this be done via theme.json?

vcanales: It can! I'll update it.

PR #6615, opened 2022-09-22
Added a front-page template

pbking:
This change adds a stylized front-page template to loudness.

The background of the header is colored with the primary color.
There is no space between the header and content so that the design can be achieved.

Sample:

image

### DIFF
--- a/loudness/patterns/front-page.php
+++ b/loudness/patterns/front-page.php
@@ -2,15 +2,17 @@
 /**
  * Title: Front Page
  * Slug: loudness/front-page
+ * Inserter: no
  */
 ?>
 
 <?php
 if ( 'posts' === get_option( 'show_on_front' ) ) {
 	// Show query
+	echo( '<!-- wp:spacer {"height":"var(--wp--preset--spacing--70)"} --><div style="height:var(--wp--preset--spacing--70)" aria-hidden="true" class="wp-block-spacer"></div><!-- /wp:spacer -->' );
 	echo( '<!-- wp:pattern {"slug":"loudness/posts-list"} /-->' );
 } else {
 	// show content
-	echo( '<!-- wp:post-content {"tagName":"main","layout":{"inherit":true},"lock":{"move":false,"remove":true}} /-->' );
+	echo( '<!-- wp:post-content {"tagName":"main","layout":{"inherit":true}} /-->' );
 }
 
--- a/loudness/patterns/front-page.php
+++ b/loudness/patterns/front-page.php
@@ -1,0 +1,16 @@
+<?php
+/**
+ * Title: Front Page
+ * Slug: loudness/front-page
+ */
+?>
+
+<?php
+if ( 'posts' === get_option( 'show_on_front' ) ) {
+	// Show query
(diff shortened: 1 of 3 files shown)


loudness/templates/front-page.html
```diff
@@ -0,0 +1,14 @@
+<!-- wp:group {"style":{"spacing":{"blockGap":"0px"}},"layout":{"type":"default"}} -->
+<div class="wp-block-group">
+
+	<!-- wp:group {"backgroundColor":"primary","layout":{"type":"default"}} -->
+	<div class="wp-block-group has-primary-background-color has-background">
+		<!-- wp:template-part {"slug":"header","theme":"themes/loudness","tagName":"header"} /-->
+	</div>
+	<!-- /wp:group -->
+
+	<!-- wp:post-content {"tagName":"main","layout":{"inherit":true},"lock":{"move":false,"remove":true}} /-->
```
jffng: I'm sorry! I realized after our conversation that this will not respect the Reading Settings. I.e. if the site is set to display latest posts, the home page will show the content of the most recent post. 

What do you think?

pbking: ... but if we include a query loop too?

pbking: No... then it shows both.

loudness/patterns/front-page.php
```diff
@@ -0,0 +1,16 @@
+<?php
+/**
+ * Title: Front Page
+ * Slug: loudness/front-page
+ */
+?>
+
+<?php
+if ( 'posts' === get_option( 'show_on_front' ) ) {
```
jffng, line 9: I think this is quite clever, and it works. But is it ultimately a workaround that should be addressed in core? cc @scruffian @MaggieCabrera for their thoughts.

The problem we are trying to solve is `home.html` isn't working well for home page setup, especially on dotcom. Lots of context here https://dotcom.wordpress.com/2022/09/01/template-vs-post-page-editing-the-confusion and https://themedevp2.wordpress.com/2022/09/20/unified-approach-to-home-page-setup-across-themes/

The idea with this PR is that we supply a front-page instead of home, so there is one template used for the home page. Depending on the site's reading settings, we would like it to render different blocks, by default.

jffng, line 10: This is clever and it works to boot! But is it working around something that should be supplied in core?

pbking: > But is it working around something that should be supplied in core?

That is absolutely the question and the answer (in my opinion) is absolutely.

The only right fix is for there to be a built-in mechanism to define one template for content and another for a query.  

HOWEVER this mechanism is exactly what has been happening in classic themes and has become the expectation.  From what I've read this behavior (there were a few mechanisms to achieve it) with front-page.php has been the standard way to achieve what we're achieving here for some time.

So while I absolutely believe that core SHOULD provide this, and we should get that on the roadmap, I also believe that until that lands this could be a pretty reasonable method to achieve the desired results.

pbking:
I'm not ready to throw my weight behind this option, but I think I would like to bring it in for now.  Either potentially as an example for a talking point and also to make sure something is in place for the design discussion later.

I'm sure we're not done with figuring this out.

PR #6623, opened 2022-09-26
pixl: remove footer and header-footer only templates

madhusudhand:

Remove footer and header-footer templates from pixl.


mikachan (approved):
LGTM

PR #6628, opened 2022-09-26
Refactored Loudness index templates.

pbking:
Removed extra templates. (footer only, header-footer-only),   Moved 'home' template to a landing page custom template called 'Landing Page'.

To use the "fancy" template for a page it must be assigned:

image


loudness/theme.json
```diff
@@ -9,20 +9,12 @@
 			"title": "Blank"
 		},
 		{
-			"name": "header-footer-only",
+			"name": "landing-page",
 			"postTypes": [
 				"page",
 				"post"
```
mikachan: Should we remove 'post' from this list? It's probably unlikely to be used on a post..

pbking: I was wondering the same. :)
I shall do.

mikachan (approved):
Looking good to me 👍

PR #6632, opened 2022-09-26
Rebalance: #679: Fix overlapping content cards w/Lazy Load enabled

### LINKED ISSUE #679: Rebalance: posts on front page are overlapping
## Steps to replicate
With the front page set to show recent blog posts, the posts intermittently overlap. Usually this doesn't happen on the first page, but only on pages that appear after you select "Older Posts".

## Result
See screenshot: https://cld.wthms.co/dd94ni

## Expected
Posts should be sorted properly.

## Interactions
* User report: #11256675-HC
* Follow-up ticket: #1882306-ZD

bobmatyas:

With Lazy Load enabled content areas overlap. Fixed by excluding Jetpack Lazy Load from processing images on `is_paged()` and `is_home()` so that the JavaScript can size the containers properly.

##### Before With Lazy Load Enabled

rebalance-01

#####  Before Without Lazy Load Enabled:

rebalance-02

##### After Lazy Load Disabled via PHP:

rebalance-03

#### Related issue(s):

Fixes #679

### DIFF
--- a/rebalance/inc/jetpack.php
+++ b/rebalance/inc/jetpack.php
@@ -278,3 +278,18 @@ function rebalance_remove_lazy_load_hooks() {
     add_action( 'wp_head', array( $instance, 'remove_filters' ), 10000 );
     remove_action( 'wp_enqueue_scripts', array( $instance, 'enqueue_assets' ) );
 }
+
+/**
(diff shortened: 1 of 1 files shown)


pbking (approved):
Works as expected.
🚢

PR #6647, opened 2022-09-30
Lynx: Remove button styles from the green style variation.

alaczek:
The custom button styles are not working well on the style switch, need more testing.

### DIFF
--- a/lynx/styles/green.json
+++ b/lynx/styles/green.json
@@ -26,21 +26,5 @@
 				}
 			]
 		}
-	},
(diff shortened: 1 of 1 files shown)


simison:
I just noticed this in testing. :-) Super weird how it's just when going from Default → Green, but swtiching via other styles works.

https://user-images.githubusercontent.com/87168/193211401-653541f7-78bf-4aeb-9e65-bb584d9fa23a.mov

Issue #1815, opened 2020-02-20
Intergalactic 2: Goodreads Widget Text is White When Placed in the Footer


zdenys:
## Steps to replicate

Add [Goodreads Widget](https://en.support.wordpress.com/widgets/goodreads-widget/) to the Footer Area and to the Sidebar

## Result

The text inside the Widget is white, hence not visible

![](https://d.pr/i/hHRbdN+)
_Link: https://d.pr/i/hHRbdN_

## Expected

See the text in Black just like in the Sidebar

![](https://d.pr/i/nkdL6q+)
_Link: https://d.pr/i/nkdL6q_

* User report: https://it.forums.wordpress.com/topic/bugged-widget-goodreads/

KokkieH:
Another report in https://wordpress.com/forums/topic/goodreads-footer-widget-not-showing-text/?view=all

PR #6650, opened 2022-09-30
#1815: Intergalactic 2: Fix Goodreads widget link colors

bobmatyas:
To test this, you can use Intergalactic 2 on a Jetpack site and add the Goodreads widget to the footer.

##### Before

Screenshot on 2022-09-30 at 11-35-50

##### Test
Screenshot on 2022-09-30 at 12-38-03

#### Related issue(s):

#1815


pbking (approved):
Does what's on the tin. :)
🚢

PR #6663, opened 2022-10-05
Videomaker: Fix primary color usage

### LINKED ISSUE #65609: Videomaker: Group Block Not Showing Correct Color in Editor When Set to Use Custom Primary Color
### Quick summary

Editor is not reflecting the proper color set as the background color for a group block when the color is set to use the "Primary" color. The editor shows the group block with a black background color. Live version of page shows correct color.

 

### Steps to reproduce

1. Activate `Videomaker White` theme.
2. On a page, add some paragraph blocks and group them.
3. Set the group block background to use the "Primary" color.
4. Observer the editor versus what is shown on the live site.

The editor to show the group block reflecting the "Primary" background color that was set.

### What actually happened

The group block looks like this:

Markup on 2022-07-06 at 12:45:29

### Context

Customer report: 35875720-HC

### Simple, Atomic or both?

Simple, Atomic

### Theme-specific issue?

Videomaker White

I tested a few other themes and was not able to reproduce the issue with them.

_No response_

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

Yes, easy to implement

### Workaround details

Set the group block background to use a gradient color. Set the gradient color to use the same color values for both ends of the gradient.

Markup on 2022-07-06 at 12:47:28

nelsonec87:

Currently, when setting a custom primary color, the foreground color is used instead, as described here: https://github.com/Automattic/wp-calypso/issues/65609

image

I've tracked the issue to the global variable `--wp--custom--color--primary` being set to `--wp--preset--color--foreground`.

image

as opposed to what happens in other themes:

image

This PR changes `theme.json` to fix that for both Videomaker and Videomaker White.

Closes https://github.com/Automattic/wp-calypso/issues/65609

### DIFF
--- a/videomaker-white/theme.json
+++ b/videomaker-white/theme.json
@@ -53,8 +53,8 @@
 			"color": {
 				"foreground": "var(--wp--preset--color--foreground)",
 				"background": "var(--wp--preset--color--background)",
-				"primary": "var(--wp--preset--color--foreground)",
-				"secondary": "var(--wp--preset--color--foreground)",
+				"primary": "var(--wp--preset--color--primary)",
+				"secondary": "var(--wp--preset--color--secondary)",
 				"tertiary": "var(--wp--preset--color--tertiary)"
 			},
 			"fontSizes": {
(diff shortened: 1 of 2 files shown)


daledupreez (commented):
This looks OK to me, but I have so little context with this that I would rather someone with deeper context to take a look.

scruffian (approved):
LGTM

PR #6664, opened 2022-10-06
Masu: Initial Commit

iamtakashi:
This adds a new blog theme, Masu.

#### More info:
p9Jlb4-5pi-p2

#### Example:
screencapture-iamtakashi-jurassic-tube-2022-10-06-20_39_02

Thanks in advance for the review!

cc @alaczek @onuro


alaczek:
I have a few visual comments after testing locally.

The post-meta can be taller than the post title, which makes the two misaligned. I felt that both should be aligned to the top. Maybe the post meta doesn't need the author since the author is displayed under the post content already? That should help with the misalignment.

image

Not sure if there's a way to fix this, but the mobile nav menu is left-aligned when pushed to a new line on mobile:
image

Overall I felt that the body font size is too big (I had to double-check that I didn't have zoom enabled in my browser) - maybe could be tweaked some more? 

The archive title typography could use some tweaking as it's quite hard to spot. I think it would also make sense to enable the "Show archive type in title" option.
image

On larger screens (above ~1500px browser width) the lines above paging in single post view are getting wider than the line below the post:
image

I also noticed on mobile the lines are doubled, but I guess there's no way around that?
image

masu/style.css
```diff
@@ -0,0 +1,307 @@
...
+Description: Masu is a blog theme inspired by a traditional square wooden box used to measure rice in Japan.
+Requires at least: 5.8
+Tested up to: 6.0
+Requires PHP: 5.7
+Version: 1.0.0
+License: GNU General Public License v2 or later
+License URI: http://www.gnu.org/licenses/gpl-2.0.html
+Template: 
+Text Domain: masu
+Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, featured-images, full-site-editing, block-patterns, rtl-language-support, sticky-post, threaded-comments, translation-ready, wide-blocks, auto-loading-homepage, blog-homepage
```
alaczek, line 15: Could you remind me what does `blog-homepage` tag do?

masu/parts/header.html
```diff
@@ -0,0 +1,11 @@
+<!-- wp:group {"style":{"spacing":{"padding":{"top":"max(4vw, 32px)","bottom":"max(2vw, 32px)"}}},"className":"gapless-group","layout":{"inherit":"true","type":"constrained"}} -->
+<div class="wp-block-group gapless-group" style="padding-top:max(4vw, 32px);padding-bottom:max(2vw, 32px)"><!-- wp:group {"align":"wide","style":{"spacing":{"blockGap":"2rem"}},"layout":{"type":"flex","justifyContent":"space-between"}} -->
+<div class="wp-block-group alignwide"><!-- wp:group {"style":{"spacing":{"blockGap":"2rem"}},"layout":{"type":"flex"}} -->
+<div class="wp-block-group"><!-- wp:site-logo /-->
+
+<!-- wp:site-title {"style":{"elements":{"link":{"color":{"text":"var:preset|color|black"}}}}} /--></div>
+<!-- /wp:group -->
+
+<!-- wp:navigation {"__unstableLocation":"primary","overlayBackgroundColor":"beige","overlayTextColor":"burgundy","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"},"blockGap":"1.41em"}}} /--></div>
```
alaczek: I think the `"__unstableLocation":"primary"` can be removed as per this comment: pNEWy-f9f-p2#comment-54757

masu/theme.json
```diff
@@ -0,0 +1,699 @@
+{
+	"version": 2,
+	"$schema": "https://schemas.wp.org/trunk/theme.json",
+	"settings": {
+		"appearanceTools": true,
+		"color": {
+			"palette": [
+				{
+					"color": "#501511",
```
alaczek, line 9: Was there a decision made somewhere about the color naming convention going forward?

masu/style.css
iamtakashi: I'm not sure :) But I've seen the tag recent launched themes that has its homepage with blog posts.

masu/parts/header.html
iamtakashi: 👍

masu/theme.json
onuro: I was going to add the same. I wonder why colors have custom slugs against primary/secondary etc.

onuro:
I don't know if reproducable but I had this issue in the site editor immediately after activating the theme locally on a fresh site with theme unit test posts:

Maybe it's a Gutenberg issue irrelevant of the theme.

image

Update: above issue is isolated to sites with theme unit test data only(probably because it has pre-Gutenberg posts from more than 5 years ago). I think this can be ignored.

masu/theme.json
iamtakashi: I don't know if there is a decision has been made. We saw the thread about the convention used in Rainfall, and I wanted to try it. But I think I'll return to the current convention, as many patterns will look awful in this theme.

iamtakashi:
Thanks for testing and the feedback, @alaczek !

> The post-meta can be taller than the post title, which makes the two misaligned. I felt that both should be aligned to the top. Maybe the post meta doesn't need the author since the author is displayed under the post content already? That should help with the misalignment.

They are actually designed to be aligned at the bottom. Since the post title size is large, the title looks pushed up for a few pixels. I've now pushed up the post meta to balance them. 

I wouldn't worry about the height of the post meta because of the alignment. And there always be posts with multiple categories and tags anyway.

iamtakashi jurassic tube__p=60

----

> Not sure if there's a way to fix this, but the mobile nav menu is left-aligned when pushed to a new line on mobile:

I don't know if there is, but I prefer the mobile nav to be on the left in the case.

----

> Overall I felt that the body font size is too big (I had to double-check that I didn't have zoom enabled in my browser) - maybe could be tweaked some more?

It does fall on the larger side, but I don't personally see it as a problem. With the generous column size, if we make the font size smaller, we would have too many characters on a line, which would also negatively affect readability. Now the characters per line is around the sweet spot.

Screenshot 2022-10-10 at 17 10 52

----

> The archive title typography could use some tweaking as it's quite hard to spot. I think it would also make sense to enable the "Show archive type in title" option.

Hmm...  Is it really hard to spot? And I don't like the prefix :) I think it's verbose and redundant.

iamtakashi jurassic tube__cat=1

----

> On larger screens (above ~1500px browser width) the lines above paging in single post view are getting wider than the line below the post:

:+1 I changed it to be constrained. I don't think we have a good way to avoid the double line on small screens. 

----

P.S. I'm curious why the headings on your screenshot aren't bold... Looks like it's not just heading. It seems like no bold is working at all on your test site.

iamtakashi:
Let me know if I can push this to WP.com 🙇

alaczek:
> They are actually designed to be aligned at the bottom.

Yeah, that makes sense :) 

> Hmm... Is it really hard to spot? And I don't like the prefix :) I think it's verbose and redundant.

It is hard to spot when none of the headings are in bold ;) I'll look into this on my local install - maybe it's just me? Let's see how it behaves on .com and go from there. Feel free to merge and let's keep testing.

masu/theme.json
```diff
@@ -0,0 +1,699 @@
...
+							"typography": {
+								"textDecoration": "underline solid"
+							}
+						}
+					}
+				}
+			},
+			"core/paragraph": {
+				"typography": {
+					"lineHeight": "calc( 1em + 0.875rem )"
```
geriux, line 319: Hey there @iamtakashi 👋  I wonder if there is a reason why this doesn't have a number value or a `var` name and it just includes a `calc()` function? Currently, we are updating the mobile editor code to take into account these cases, but until that code goes live and the adoption of that future update reaches a nice percentage, this currently breaks for users who use this theme. Is there a chance we could update it to not have a `calc` function and just a value/var with a value instead?

iamtakashi: Hi @geriux 👋 Thanks for the ping. The reason why I used that was to make sure paragraphs look good regardless of the size. When we use a single value, it gets hideous and separated too much as the size increases.

| Relative value (current) | static value (e.g.1.5) |
| ----------- | ----------- |
| nice | ugly |

If I'm not wrong, the paragraph block doesn't allow specifying different line height values for different sizes, and I use this solution in other themes coming up.

How about like this? Would this work with the mobile editor code, or would this be the same? Do you have any suggestions?

```
"settings": {
	"custom": {
		"typography": {
			"line-height": {
				"test": "calc( 1em + 0.875rem )"
			}
		}
	}
},
"styles": {
	"blocks": {
		"core/paragraph": {
			"typography": {
				"lineHeight": "var(--wp--custom--typography--line-height--test)"
			}
		}
	}
}
```

geriux: Hey there! Thanks for the quick reply and for also suggesting a workaround. After some investigation, we found we also had issues while parsing decimals in math expressions like in this one `calc( 1em + 0.875rem )` unfortunately this will need to be a fix for the app itself and not this actual theme because it could be a common use case in other themes configuration.

Let's keep it as it is now and we will support it correctly in an upcoming app release.

Thanks again!

iamtakashi: Got it. No worries!

PR #6665, opened 2022-10-07
Initial Commit of Clean Slate

onuro:
This PR introduces the Clean Slate theme, a blank/starter theme in replacement of Blank Canvas, started out with the purpose of using in Pattern Assembler project.

This theme is a stripped down version of TT3 theme, with some more custom theme.json and template parts updates.

I've kept the fonts from TT3, as the use of a fonts library for this theme TBD.

Please let me know what could be improved with this theme, as I'm still checking for things that may be needed to be updated.

Thank you!


scruffian:
I think if we want a truly clean slate we should remove a lot more - all the patterns for example can go, and quite a few of the templates as well.

scruffian:
What happens if we remove all templates except index?

onuro:
> What happens if we remove all templates except index?

Tbh makes sense. @mtias this is also something along the lines of what you've expressed for this right?

clean-slate/parts/post-meta.html
```diff
@@ -0,0 +1,12 @@
...
+	<!-- wp:group {"layout":{"type":"flex"}} -->
+	<div class="wp-block-group">
+		<!-- wp:post-author-name {"isLink":true,"fontSize":"small"} /-->
+		<!-- wp:post-date {"isLink":true,"fontSize":"small"} /-->
+		<!-- wp:post-terms {"term":"category","fontSize":"small"} /-->
+		<!-- wp:post-terms {"term": "post_tag","fontSize":"small"} /-->
+	</div>
+	<!-- /wp:group -->
+</div>
+<!-- /wp:group -->
```
bangank36: You need a new line here

clean-slate/style.css
```diff
@@ -0,0 +1,16 @@
...
+Requires at least: 5.8
+Tested up to: 5.9
+Requires PHP: 5.7
+Version: 0.0.1
+License: GNU General Public License v2 or later
+License URI: http://www.gnu.org/licenses/gpl-2.0.html
+Template: 
+Text Domain: clean-slate
+Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, featured-images, full-site-editing, rtl-language-support, theme-options, threaded-comments, translation-ready, wide-blocks
+*/
```
bangank36: New line needed

clean-slate/templates/single.html
```diff
@@ -0,0 +1,61 @@
...
+
+<!-- wp:comments-pagination-next /-->
+<!-- /wp:comments-pagination -->
+
+<!-- wp:post-comments-form /--></div>
+<!-- /wp:comments -->
+</div>
+<!-- /wp:group -->
+
+<!-- wp:template-part {"slug":"footer","tagName":"footer"} /-->
```
bangank36: New line needed

alaczek:
I'm also wondering about the inclusion of font assets. What do you think about removing them and reverting to using system fonts?

arthur791004:
Could we have the `home.html` file without any content and use it to display the home page of this theme? cc @mtias

clean-slate/style.css
```diff
@@ -10,7 +10,7 @@ Requires PHP: 5.7
 Version: 0.0.1
 License: GNU General Public License v2 or later
 License URI: http://www.gnu.org/licenses/gpl-2.0.html
-Template: 
+Template:
```
scruffian:
```suggestion
```

onuro: @scruffian should I be removing the Template line all together or just remove the space next to the colon?

jffng: Removing it entirely is best. This line should be included only if there is a Parent theme.

blank-canvas-3/theme.json
```diff
@@ -0,0 +1,360 @@
...
+			},
+			"heading": {
+				"typography": {
+					"fontWeight": "400",
+					"lineHeight": "1.4"
+				}
+			},
+			"link": {
+				"color": {
+					"text": "var(--wp--preset--color--primary)"
```
scruffian, line 304: It might be interesting to use `ref` properties, to reduce the duplication: https://github.com/WordPress/gutenberg/pull/41696

scruffian:
I'm not clear how "blank" we want this to be. The templates still contain a lot of opinion. Do we want to strip it back to just display page content?

blank-canvas-3/theme.json
```diff
@@ -0,0 +1,360 @@
+{
+	"customTemplates": [
```
jffng: I think the `customTemplates` config can be removed since there is no longer a blank template supplied by the theme.

blank-canvas-3/theme.json
onuro: Hi @scruffian could you elaborate on this? Is do I simply use the following? Because when I did, the theme.json and the theme style breaks. I also don't understand the reason fully even though I've read the PR.

```
"link": {
				"color": {
					"text": { ref: "styles.color.primary" }
				},
```

scruffian: Try using:

```
"color": {
	"text": { ref: "styles.color.text" }
},
```

The reason being that if a user changes their body text color then links will update automatically.

PR #6676, opened 2022-10-19
Rainfall: general improvements

mikachan:
This PR indents the block markup in many of the templates so the code is easier to read (sorry, this makes the diffs difficult to read!)

It also ensures the default and outline buttons are the same size.

| Before | After |
| ------ | ----- |
| image | image |

### DIFF
--- a/rainfall/theme.json
+++ b/rainfall/theme.json
@@ -360,7 +360,7 @@
 					"color": "var(--wp--preset--color--black)",
 					"radius": "0",
 					"style": "solid",
-					"width": "1px"
+					"width": "2px"
 				},
 				"color": {
 					"background": {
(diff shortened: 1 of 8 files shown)


rainfall/theme.json
```diff
@@ -357,8 +357,10 @@
 		"elements": {
 			"button": {
 				"border": {
+					"color": "var(--wp--preset--color--black)",
 					"radius": "0",
-					"width": "0"
+					"style": "solid",
+					"width": "1px"
```
madhusudhand: there is a 1px difference still. set it to `2px`, to match the outline button border.

image

mikachan: Thanks!! Done

madhusudhand (approved):
Looks good!

PR #6677, opened 2022-10-20
Aine: technical review

jffng:

This PR adds a new theme Aine designed and built by @beafialho 🎉


aine/index.php
```diff
@@ -0,0 +1,2 @@
+<?php
```
jffng: Do we need this anymore?

aine/parts/header.html
```diff
@@ -0,0 +1,7 @@
+<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|60","right":"var:preset|spacing|60","left":"var:preset|spacing|60"}}},"layout":{"type":"default"}} -->
+<div class="wp-block-group alignfull" style="padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60);padding-left:var(--wp--preset--spacing--60)"><!-- wp:group {"align":"full","layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
+<div class="wp-block-group alignfull"><!-- wp:site-title {"style":{"typography":{"fontStyle":"normal","fontWeight":"700","textTransform":"uppercase"}},"fontSize":"medium"} /-->
+
+<!-- wp:navigation {"ref":182,"overlayMenu":"never","layout":{"type":"flex","justifyContent":"space-between"},"style":{"spacing":{"blockGap":"30px"}},"fontSize":"medium"} /--></div>
```
jffng: Can we remove the `ref`?

aine/style.css
```diff
@@ -0,0 +1,76 @@
...
+	font-size: var(--wp--preset--font-size--small);
+	text-align: center;
+}
+
+/*
+ * Link styles
+ */
+a {
+	text-decoration-thickness: .0625em !important;
+	text-underline-offset: .15em
```
jffng:
```suggestion
	text-underline-offset: .15em;
```

aine/templates/wp-custom-template-landing-page.html
```diff
@@ -0,0 +1,15 @@
+<!-- wp:template-part {"slug":"header-landing-page"} /-->
```
jffng: I know it sounds like a broken record 😅 , but I'm not sure if we should ship a custom template to achieve a unique landing page design.

aine/templates/blank.html
```diff
@@ -0,0 +1 @@
+<!-- wp:post-content {"layout":{"inherit":true},"lock":{"move":false,"remove":true}} /-->
```
jffng: Do we need a blank template?

beafialho: I guess we don't.

aine/templates/wp-custom-template-landing-page.html
beafialho: Yes, I was not 100% sure about what to do here. Maybe I could build this as an index, leaving the posts and include the description at the bottom in the footer, but I wonder if that's even more confusing.

aine/index.php
mikachan: No, we don't need to include this file anymore.

aine/parts/footer.html
```diff
@@ -0,0 +1,9 @@
+<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"80px","bottom":"30px"}}}} -->
+<div class="wp-block-group alignfull" style="padding-top:80px;padding-bottom:30px"><!-- wp:spacer {"height":"80px"} -->
+<div style="height:80px" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- /wp:spacer -->
+
+<!-- wp:paragraph {"align":"center","style":{"typography":{"fontSize":"0.8rem","letterSpacing":"-0.2px"}}} -->
+<p class="has-text-align-center" style="font-size:0.8rem;letter-spacing:-0.2px">Designed with <a rel="nofollow" href="https://wordpress.org">WordPress</a></p>
```
mikachan: We should allow this string to be translatable, and this file should probably use the included footer pattern.

mikachan:
I've addressed a few of the comments made so far and then moved the footer and 404 contents into patterns, mainly so the strings are translatable.

I'm really liking the feel of this theme, especially the large header style 🤌

aine/templates/wp-custom-template-landing-page.html
jffng: As conversations progress, `front-page.html` that contains a post content block (like you've done with this custom template) could end up being the way to go here, when themes want to supply a unique template for their home page design.

mikachan: Maybe we should use index.html here so we can progress this theme?

jffng: Agreed, I moved it to index for now and got rid of the custom homepage template. @beafialho does this work for you?

jffng:
@beafialho could you provide the block markup for this content? I think we should insert it directly into the template:

Screenshot 2022-11-28 at 2 55 30 PM

Then I think this should be ready to go!

beafialho:
Do you mean adding the content in the Front Page template instead of a "Home" page? Should we do that in the following themes too?

Here's the markup:

```
<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"right":"0","left":"0","top":"0","bottom":"0"}}},"layout":{"type":"default"}} -->
<div class="wp-block-group alignfull" style="padding-top:0;padding-right:0;padding-bottom:0;padding-left:0"><!-- wp:query {"queryId":0,"query":{"perPage":3,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":false,"taxQuery":null,"parents":[]},"tagName":"main","displayLayout":{"type":"flex","columns":3},"layout":{"inherit":true}} -->
<main class="wp-block-query"><!-- wp:post-template {"align":"wide","fontSize":"medium"} -->
<!-- wp:group -->
<div class="wp-block-group"><!-- wp:post-featured-image {"isLink":true,"height":"360px"} /-->

<!-- wp:post-title {"isLink":true,"style":{"elements":{"link":{"color":{"text":"var:preset|color|foreground"}}}},"fontSize":"large"} /-->

<!-- wp:post-date /-->

<!-- wp:spacer {"height":"40px"} -->
<!-- /wp:spacer --></div>
<!-- /wp:group -->
<!-- /wp:post-template -->

<!-- wp:group {"align":"wide","layout":{"inherit":true,"type":"constrained"}} -->
<div class="wp-block-group alignwide"><!-- wp:query-pagination {"align":"wide","layout":{"type":"flex","justifyContent":"space-between"}} -->
<!-- wp:query-pagination-previous /-->

<!-- wp:query-pagination-next /-->
<!-- /wp:query-pagination --></div>
<!-- /wp:group --></main>
<!-- /wp:query --></div>
<!-- /wp:group -->

<!-- wp:spacer {"height":"80px"} -->
<div style="height:80px" aria-hidden="true" class="wp-block-spacer"></div>
<!-- /wp:spacer -->

<!-- wp:separator {"align":"full","backgroundColor":"secondary","className":"is-style-wide"} -->
<hr class="wp-block-separator alignfull has-text-color has-secondary-color has-alpha-channel-opacity has-secondary-background-color has-background is-style-wide"/>
<!-- /wp:separator -->

<!-- wp:heading {"level":3,"align":"wide","style":{"spacing":{"padding":{"right":"0","left":"0","top":"0","bottom":"0"}}},"fontSize":"x-large"} -->
<h3 class="alignwide has-x-large-font-size" style="padding-top:0;padding-right:0;padding-bottom:0;padding-left:0">Is this nice? is a weekly digital publication that features artists, craftmakers and artisans.</h3>
<!-- /wp:heading -->
```

jffng:
@beafialho thanks for the markup.

> Do you mean adding the content in the Front Page template instead of a "Home" page? 

I actually am opting to put this in the `index` template for now, and not supply a Home or a Front Page, for a couple reasons:

1. It seems we are moving towards a direction that allows content in templates to persist across theme switch.
2. We can always add templates, but we can't take them away once they are in use on someone's site.  Supplying fewer templates and leaning more on the template creation flows in the site editor for custom purposes should hopefully lead to less confusion around what themes are doing. Rainfall shipped with this approach too.

iamtakashi:
This theme has been called Muscat, and is already available in WP.org.

PR #6686, opened 2022-10-21
Archeo: Improve block markup and fix block errors

### LINKED ISSUE #6673: [Bug]: Archeo: The 'Layered images with headline' pattern has block errors on wpcom
### Quick summary

The pattern included in Archeo named "Layered images with headline" has a block error on wpcom.  

image

This pattern is included in the "Home" page so when a user replaces their homepage with the theme content this markup is included.

The pattern is rendering as expected locally.

image

From what I can tell the markup is the same and nothing has jumped out as being incorrect.

### Steps to reproduce

Activate Archeo.

Add the pattern "Layered images with headline"

There wouldn't be block markup errors

### What actually happened

There are block markup errors

### Browser

_No response_

### Context

_No response_

Simple

### Other notes

_No response_

### Reproducibility

Consistent

### Severity

_No response_

### Available workarounds?

_No response_

### Workaround details

_No response_

mikachan:
This PR improves the block markup for Archeo, which should fix a couple of block markup errors:

- Fixes the footer block markup error (I think it was caused by whitespace, also removed the margin as it doesn't look like this was needed)
- Fixes the block markup error in the 'Layered images with headline' pattern (the font size was inconsistent between the JSON and HTML)
- Changes the query loops to inherit: true for the 'simple list of posts' patterns
- Updates the comments block to the new block to fix a block notice/warning in the editor
- Replaces the post-author-name block for the post-author block, as this isn't included in WP 6.1 (we don't use it in any other theme currently)
- Indents block markup where possible so code is easier to read

Archeo's headstart has also been updated to reflect these changes: D90211-code

### DIFF
--- a/archeo/patterns/footer.php
+++ b/archeo/patterns/footer.php
@@ -13,15 +13,15 @@
 	<div class="wp-block-group alignwide" style="padding-top: var(--wp--custom--spacing--medium); padding-bottom: var(--wp--custom--spacing--medium);">
 		<!-- wp:navigation {"layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"left"},"overlayMenu":"never","className":"site-footer","style":{"typography":{"fontStyle":"normal"},"spacing":{"blockGap":{"top":"0.5em","left":"2.5em","bottom":"0.5rem","right":"2.5em"}}},"fontSize":"small"} /-->
 		
-		<!-- wp:paragraph {"align":"left","style":{"spacing":{"margin":{"top":0}}},"fontSize":"small"} -->
-		<p class="has-text-align-left has-small-font-size" style="margin-top: 0;">
-			<?php
-			printf(
-				/* Translators: WordPress link. */
-				esc_html__( 'Proudly powered by %s', 'archeo' ),
-				'<a href="' . esc_url( __( 'https://wordpress.org', 'archeo' ) ) . '" rel="nofollow">WordPress</a>'
-			);
-			?>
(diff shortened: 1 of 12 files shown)


pbking (approved):
These changes check out. 👍 
🚢

pbking:
The only thing about this I'm confused about:

> Fixes the block markup error in the 'Layered images with headline' pattern (the font size was inconsistent between the JSON and HTML)

I see the error and that it was fixed but what I don't understand is why this was erroring on wpcom but not wporg environment.

¯\_(ツ)_/¯

pbking:
The headstart annotation change also looks good (as far as I can tell) and that should also ship when this does.

mikachan:
Thanks @pbking!!

> I see the error and that it was fixed but what I don't understand is why this was erroring on wpcom but not wporg environment.

I was also seeing something very similar. I could see both the errors (in the footer and this pattern) consistently on wpcom, but I couldn't see them at all on wporg. Then, when I came back today fresh (so from changing themes a few times and spinning up my local environment again), I could see the errors on my local environment too. However, only the footer error was consistent. The error in the 'Layered images with headline' pattern only showed sometimes, although probably as often enough as 50% of the time.

I felt better about fixing these as soon as I saw the inconsistencies in the JSON comments and HTML, but this was definitely a struggle to work out initially when I couldn't replicate them on my local site. It would be good to investigate why these errors aren't showing up in the same way between the two environments.

PR #6688, opened 2022-10-24
Fix dotorg deploy action failure reporting.

### LINKED ISSUE #6678: Fail wporg deploy action when it doesn't deploy due to an old password
Consider recent [deploy runs](https://github.com/Automattic/themes/actions/workflows/dotorg-deploy.yml) to dotorg, and we don’t have a reporting condition for when deploy-dotorg.sh fails that I can see. 

[The job](https://github.com/Automattic/themes/actions/runs/3282931071/jobs/5406994716#step:3:928) runs “successfully” but the authentication fails for committing to SVN. So when committing to trunk, we will think that the themes are deployed to dotorg, but in reality they aren’t. 

The issue here was the SVN_PASSWORD secret, but we should update the job to “fail” correctly, and notify us when the job “fails”.

jffng:

Checks the output of the svn commit and exiting with a non-zero status code, if it finds an error.

#### Related issue(s):

#6678

### DIFF
--- a/deploy-dotorg.sh
+++ b/deploy-dotorg.sh
@@ -77,6 +77,9 @@ for THEME_SLUG in ${THEMES_TO_DEPLOY[@]} ; do
 	echo "➤ Copying theme '${THEME_SLUG}' version '${THEME_VERSION}' to svn repository... "
 	rsync -rc --delete --include=theme.json --exclude-from './dotorg-exclude.txt' ./$THEME_SLUG/ $SVN_DIR/$THEME_VERSION
 
+	# Remove the wpcom-specific tags used in some themes
+	find $SVN_DIR/$THEME_VERSION/style.css -type f -exec sed -i '' 's/, auto-loading-homepage//g' {} \; 
+
 	# Remove files from the previous version	
 	svn status $SVN_DIR/$THEME_VERSION | grep "^\!" | sed 's/^\! *//g' | xargs svn rm;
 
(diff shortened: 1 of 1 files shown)


deploy-dotorg.sh
```diff
@@ -77,13 +77,6 @@ for THEME_SLUG in ${THEMES_TO_DEPLOY[@]} ; do
 	echo "➤ Copying theme '${THEME_SLUG}' version '${THEME_VERSION}' to svn repository... "
 	rsync -rc --delete --include=theme.json --exclude-from './dotorg-exclude.txt' ./$THEME_SLUG/ $SVN_DIR/$THEME_VERSION
 
-	# Remove -wpcom from versoning
-	find $SVN_DIR/$THEME_VERSION/style.css -type f -exec sed -i '' 's/-wpcom//g' {} \; 
-
-	# Remove the wpcom-specific tags used in some themes
-	find $SVN_DIR/$THEME_VERSION/style.css -type f -exec sed -i '' 's/, auto-loading-homepage//g' {} \; 
-	find $SVN_DIR/$THEME_VERSION/style.css -type f -exec sed -i '' 's/, jetpack-global-styles//g' {} \; 
-
```
mikachan: Do we want to remove all these lines? I guess it's unlikely that the themes in our dotorg-deploy list will use a -wpcom version, but they may still use the 'auto-loading-homepage' tag, which probably needs to be removed for dotorg.

mikachan (commented):
This LGTM! Checking for 'svn: E' sounds like a good way to pick up on the error notices from SVN 👍

deploy-dotorg.sh
jffng: Ah good point, I will leave that line in!

mikachan (approved):
🚢 🚢

PR #6696, opened 2022-10-26
Header Update for matching with original Spearhead

onuro:
Header template update for home to match with the original Spearhead theme & screenshot update


spearhead-blocks/parts/header-home.html
```diff
@@ -0,0 +1,25 @@
...
+<div class="wp-block-group"><!-- wp:site-logo {"width":64} /-->
+
+<!-- wp:group {"style":{"spacing":{"blockGap":"5px"}}} -->
+<div class="wp-block-group"><!-- wp:site-title /-->
+
+<!-- wp:site-tagline /--></div>
+<!-- /wp:group --></div>
+<!-- /wp:group -->
+
+<!-- wp:navigation {"__unstableLocation":"primary","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /--></div>
```
pbking: I don't think this navigation block should have __unstableLocation.  It's a depreciated attribute and no navigation locations are defined.

pbking: I removed it. :)

pbking:
I think this looks good to ship.

I want to note that the home header template has un-localized strings in it.  Considering that that's "example" text my opinion is that that doesn't have to be in a pattern and localized; It's best the way it is.  I did think it important to note that however in case others feel differently.

Additionally there are a couple of uses of non-standard font sizes used.  Best to have those as configured font sizes so that users can more easily reproduce that.  Again that's opinion and no reason to block this change if you feel the size difference is important enough to include in the template but not important enough to include in the configuration.

pbking (approved):
Approved with notes.

spearhead-blocks/templates/single.html
```diff
@@ -3,11 +3,11 @@
...
 <div class="wp-block-group"><!-- wp:post-date {"textColor":"secondary","fontFamily":"ibm-plex-mono"} /-->
 
-<!-- wp:post-title {"style":{"typography":{"fontSize":"3rem","lineHeight":1.3999999999999999}}} /--></div>
+<!-- wp:post-title {"style":{"typography":{"fontSize":"3rem","lineHeight":1.2}}} /--></div>
 <!-- /wp:group -->
 
-<!-- wp:group {"tagName":"main","lock":{"move":false,"remove":true}} -->
-<main class="wp-block-group"><!-- wp:post-featured-image {"align":"full"} /-->
+<!-- wp:group {"layout":{"type":"constrained"},"style":{"spacing":{"margin":{"top":"var:preset|spacing|40"}}}} -->
+<div class="wp-block-group" style="margin-top:var(--wp--preset--spacing--40)"><!-- wp:post-featured-image {"align":"full"} /-->
```
mikachan, line 10: I think this element should still use a `main` tag. And should it still have the `lock` settings?

spearhead-blocks/parts/header-home.html
```diff
@@ -0,0 +1,25 @@
...
+
+<!-- wp:navigation {"layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /--></div>
+<!-- /wp:group -->
+
+<!-- wp:heading {"fontSize":"medium"} -->
+<h2 class="has-medium-font-size">Weekly news about tech, investing, and the digital age</h2>
+<!-- /wp:heading -->
+
+<!-- wp:paragraph {"style":{"typography":{"fontSize":"1.25rem"}}} -->
+<p style="font-size:1.25rem">Spearhead is a weekly tech podcast released every Wednesday. Join hosts Kjell, Mel, and Jeff as they dive deep into top events, the hottest tech news, and discover which startups received VC funding.</p>
```
mikachan, line 22: Should these strings be translatable? If so, they should be moved to a pattern.

PR #6697, opened 2022-10-26
Loudness: Separate PAGE and SINGLE templates

pbking:
Separate 'page' and 'single' templates.  

Removed comments from page template.  

Refined post title positing to match design.

Before:
image

After:
image


loudness/readme.txt
```diff
@@ -1,7 +1,7 @@
 === Loudness ===
 Contributors: Automattic
-Requires at least: 5.8
```
jffng, line 3: Just checking this change is desired?

jffng (approved):

loudness/readme.txt
pbking: I didn't intend to make that change in this PR but the change was intentional and desired.

PR #6698, opened 2022-10-26
Loudness: Add home template to the theme.

pbking:
Add home template with content from design and refine patterns used in home template.

image


jffng (approved):
Added some alt text, escaped all the text, and compressed the images. This one LGTM.

PR #6705, opened 2022-10-28
Block Canvas: small updates

mikachan:
This PR adds a few small updates to Block Canvas:

- Removes index.php, as this is no longer required
- Removes the blank.html template, as I believe we are no longer including this
- Adds links to Gutenberg issues for all the included CSS
- Removes the "Responsive menu container padding" CSS, as this is now handled via the root padding settings in theme.json
- Indents all block markup so it's easier to read
- Updates all layout settings to the new format (`"type":"constrained"` rather than `"inherit":"true"`)
- Swaps the post-author-name block for post-author, as post-author-name only works with Gutenberg active
- Updates the WordPress footer credit line to "Designed with WordPress" (based on the footer credit in [Aine](https://github.com/Automattic/themes/pull/6677))

### DIFF
--- a/block-canvas/parts/header.html
+++ b/block-canvas/parts/header.html
@@ -15,7 +15,7 @@
 		</div>
 		<!-- /wp:group -->
 	
-		<!-- wp:navigation {"__unstableLocation":"primary","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /-->
+		<!-- wp:navigation {"layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /-->
 	</div>
 	<!-- /wp:group -->
 </div>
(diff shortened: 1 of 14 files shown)


block-canvas/parts/header.html
```diff
@@ -1,19 +1,23 @@
...
+				<!-- wp:site-title /-->
+				<!-- wp:site-tagline /-->
+			</div>
+			<!-- /wp:group -->
+		</div>
+		<!-- /wp:group -->
 	
-	<!-- wp:site-tagline /--></div>
-	<!-- /wp:group --></div>
+		<!-- wp:navigation {"__unstableLocation":"primary","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /-->
```
jffng: Can we remove the `__unstableLocation` given the improved way the nav block fallbacks now?

jffng (commented):
Nice, thank you for doing this! Just left one small comment.

block-canvas/parts/header.html
mikachan: Ah yes, I think we can!

jffng (approved):

PR #6710, opened 2022-11-02
Adding license information for resources used in Loudness

pbking:
I have added the license information for the images found in the theme.  They are all from Stocksnap.

The illustrations were created by @kjellr and aren't currently hosted and licensed anywhere.  I have included a CC0 license reference for those images.

Is this an appropriate way to attribute and license this?


jeffikus:
@pbking can you ask on the trac ticket if we can submit the license with the theme please?

PR #6711, opened 2022-11-03
Blockbase: Ensure Blockbase migration script is only run once

jeyip:
WIP

#### Related issue(s):
See slack thread p1667491989163529-slack-CHN6J22MP

### DIFF
--- a/blockbase/inc/fonts/custom-font-migration.php
+++ b/blockbase/inc/fonts/custom-font-migration.php
@@ -3,6 +3,10 @@
 add_action( 'init', 'migrate_blockbase_custom_fonts', 99 );
 
 function migrate_blockbase_custom_fonts() {
+	// The data has already been transformed
+	if ( get_theme_mod( 'blockbase_legacy_font_settings' ) ) {
+		return;
+	}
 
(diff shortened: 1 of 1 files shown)


jeyip:
@pbking non-urgent ping. Just wanted to keep this on our radar

pbking:
To test this I first checked out the last version of Blockbase that allowed the user to customize colors via the customizer:
`tags/v1.0.169`

Then I used the customizer to customize said colors.

Then I checked out this branch and added a few debugging `die()` statements to easily step through what was happening.  I ensured the migration script ran (with an appropriately placed `die('migrating')`) and that the migration script wasn't running at all after that.

I ensured that the user selected colors were applied as expected and were configurable as expected in the FSE.  

I tested this on both Blockbase and Geologist.  There are many more child themes but I don't think it's necessary to test them.

I believe that this is good to go!

pbking (approved):
👍

pbking:
I made a small change to store a boolean instead of a value in that array.  The value is never used and the presence is what is checked so that's a more appropriate value.

Tests out the same as the above.

Shipping. :)

PR #6714, opened 2022-11-07
Try/cleanup buttons css

madhusudhand:
This PR is a copy of https://github.com/Automattic/themes/pull/6042 to trunk

We are removing the CSS that is now covered by the elements API implementation in https://github.com/Automattic/themes/pull/6041. The buttons on the search block, file block and button block should look the same as before.

I decided to leave the mixins as they are, in case they are used in other child themes, while removing the includes from Blockbase itself. I'm keeping all the hover stuff so we are not breaking anyone's themes either until that is supported by theme.json too.


mikachan (approved):
I noticed the file block button looked slightly different on hover with this PR - it was showing an underline. I've fixed this in the theme.json file here: https://github.com/Automattic/themes/pull/6714/commits/1d6b08fbf42c37755e164f91023782ac1d8a43fb

I also noticed the theme.json was using version 1, so I've updated this to version 2 (although I know that's not part of this update..)

Otherwise, looks good to me!

PR #6726, opened 2022-11-15
Blank Canvas 3 - Incorrect color value fix & increased widths

onuro:
- Assigned the correct color for text
- Wide and content widths are slightly increased for better support for Pattern Assembler


miksansegundo (approved):
👍

PR #6730, opened 2022-11-17
Remove  block attribute from any theme not defining nav menu locations

pbking:

A few themes had templates/patterns that were assigning `__unstableLocation` attributes to navigation blocks.  However, if navigation targets aren't defined by the theme (such as [here](https://github.com/Automattic/themes/blob/trunk/blockbase/functions.php#L38) in Blockbase) the attribute does nothing and is only confusing.

This change removes the attribute from any patterns and templates from themes that aren't defining those locations (which only Blockbase is doing).

### DIFF
--- a/disco/parts/header.html
+++ b/disco/parts/header.html
@@ -8,7 +8,7 @@
 	<!-- /wp:group --></div>
 	<!-- /wp:group -->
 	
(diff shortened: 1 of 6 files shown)


mikachan (approved):
LGTM!

jeffikus:
@pbking will this impact existing customer sites?

pbking:
> will this impact existing customer sites?

No this is just some housekeeping that removes the attribute from themes that weren't using the attribute in a way that helped the user anyway.  The themes this change removes the attribute from weren't defining navigation targets (in functions.php) so the attribute wasn't doing anything.

Removing this attribute from Blockbase (and children, or any other theme that assigns navigation location targets) would impact customer sites.

pbking:
Relatedly, I have opened #6738 to remove __unstableLocation from themes where it's properly managed.

PR #6731, opened 2022-11-17
Adjust the 'pricing' pattern to better fit in all screen dimensions

pbking:

At certain screen dimensions the pattern `pricing.html` had awkward line breaks.  These changes address those so that it should look OK in all screen dimensions:

Before:
image

After:
image


jffng (approved):
Collapses nicely across viewport widths 🚢

PR #6734, opened 2022-11-18
Remote: update fluid-typography and clean button styles

madhusudhand:

* This updates the typography settings to fluid (just migrated existing behavior without introducing fluid where it is not)
* Removes styles for button states. They are moved to theme.json

#### Related issue(s):

#6493

### DIFF
--- a/remote/style.css
+++ b/remote/style.css
@@ -7,7 +7,7 @@ Description: Remote is a dark, minimal block theme ideal for bloggers. Its defau
 Requires at least: 6.1
 Tested up to: 6.1
 Requires PHP: 5.7
-Version: 1.1.0
+Version: 1.0.12
 License: GNU General Public License v2 or later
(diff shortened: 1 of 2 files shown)


mikachan (approved):
LGTM 👍

remote/style.css
```diff
@@ -7,7 +7,7 @@ Description: Remote is a dark, minimal block theme ideal for bloggers. Its defau
 Requires at least: 6.1
 Tested up to: 6.1
 Requires PHP: 5.7
-Version: 1.0.12
+Version: 1.1.0
```
mikachan: I think this should still be `1.0.12`? This is the version listed in the changelog.

madhusudhand: updated minor version with respect to the changes. 
Reverted it from changes and it is best to bump the version at the time of deployment.

PR #6743, opened 2022-11-22
Blockbase: Enabled navigation fallback

pbking:
Removed the bits that disabled the navigation fallback so that navigation would work as expected.

This was causing an issue with sites with generated navigation (most specifically headstart-generated navigation).


scruffian (approved):
LGTM

PR #6745, opened 2022-11-23
Update Spacings

onuro:
Updated spacings to use core/editor default presets and removed TT3 based spacing presets.

### DIFF
--- a/blank-canvas-3/theme.json
+++ b/blank-canvas-3/theme.json
@@ -36,42 +36,9 @@
 			"contentSize": "700px",
 			"wideSize": "1160px"
 		},
+		"useRootPaddingAwareAlignments": true,
 		"spacing": {
-			"spacingScale": {
-				"steps": 0
-			},
-			"spacingSizes": [
-				{
-					"size": "clamp(1.5rem, 5vw, 2rem)",
-					"slug": "30",
-					"name": "1"
-				},
-				{
-					"size": "clamp(1.8rem, 1.8rem + ((1vw - 0.48rem) * 2.885), 3rem)",
-					"slug": "40",
-					"name": "2"
-				},
-				{
-					"size": "clamp(2.5rem, 8vw, 4.5rem)",
-					"slug": "50",
-					"name": "3"
-				},
-				{
-					"size": "clamp(3.75rem, 10vw, 7rem)",
-					"slug": "60",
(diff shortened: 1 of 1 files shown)


iamtakashi:
@onuro Thanks for working on the PR. As per the convo in Slack, I think it's a good idea to remove the custom `spacingSizes` from the theme, as the name suggests. It was too opinionated, in my opinion. The differences is huge.

| Old(custom) | New(default) |
| ----------- | ----------- |
| old | new |

Since the custom size is now removed with this, there are places that have different spacing from before, for example, the padding in the header. I don't think they need to be identical, but we might want to adjust them with the spacer block if the preset isn't enough.

Maybe, can you tackle that with another PR after we merge this?

onuro:
> @onuro Thanks for working on the PR. As per the convo in Slack, I think it's a good idea to remove the custom `spacingSizes` from the theme, as the name suggests. It was too opinionated, in my opinion. The differences is huge.

> Maybe, can you tackle that with another PR after we merge this?

@iamtakashi all the spacing sizes also removed in this PR. And there are only 8 declarations of spacings left assigned into some blocks(post title margins, comments etc).

Do you mean the global spacing could also be smaller?
image

iamtakashi:
> Do you mean the global spacing could also be smaller?

No :) What I'm trying to point out is,

The theme uses presets in a few places in the templates. I can see it's been used seven times in five templates, `footer.html`, `header.html`, `index.html`, `page.html`, and `single.html`.

Screenshot 2022-11-23 at 21 54 06

Screenshot 2022-11-23 at 21 54 15

With this PR, for example, `--wp--preset--spacing--60` is no longer the same value as before. Therefore the bottom padding in the header is quite different from before this PR. 

Frame 5

This is just one example. Again, I can see the preset was used in seven places in those five templates. Don't we want to adjust all of them? If we decide to adjust them, I'd suggest doing that in a separate PR after merging this.

I can also see the preset is used in four places in `theme.json`. That's also worth it for you to review.

onuro:
> This is just one example. Again, I can see the preset was used in seven places in those five templates. Don't we want to adjust all of them? If we decide to adjust them, I'd suggest doing that in a separate PR after merging this.

Ah got it thank you! I'll revisit those. Merging this pr for now and we can test the block level spacings in the next one.

PR #6752, opened 2022-11-29
Block canvas: remove unneeded CSS

jffng:

This PR removes some CSS that has since been added to Gutenberg (PR 45069)

### DIFF
--- a/block-canvas/style.css
+++ b/block-canvas/style.css
@@ -36,10 +36,10 @@ body {
 }
 
 /*
-* Control the hover stylings of outline block style.
-* Unnecessary once block styles are configurable via theme.json
-* https://github.com/WordPress/gutenberg/issues/42794 and https://github.com/WordPress/gutenberg/issues/44417
-*/
+ * Control the hover stylings of outline block style.
+ * Unnecessary once block styles are configurable via theme.json
+ * https://github.com/WordPress/gutenberg/issues/42794 and https://github.com/WordPress/gutenberg/issues/44417
+ */
 .wp-block-button.is-style-outline>.wp-block-button__link:not(.has-background):hover {
 	background-color: var(--wp--preset--color--secondary);
 	color: var(--wp--preset--color--background);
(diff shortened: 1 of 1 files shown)


block-canvas/style.css
```diff
@@ -39,37 +38,14 @@ body {
...
-}
-.wp-block-table td,
-.wp-block-table th {
-	padding: var(--wp--preset--spacing--30);
-	border: 1px solid;
-	word-break: normal;
-}
-.wp-block-table figcaption {
-	font-size: var(--wp--preset--font-size--small);
-	text-align: center;
```
mikachan, line 70: It looks like these lines are still needed, but I recently opened a GB PR to address this: https://github.com/WordPress/gutenberg/pull/46172 (I meant to look at this a while ago and this reminded me!)

block-canvas/style.css
```diff
@@ -39,37 +38,14 @@ body {
...
- * Currently table styles are only available with 'wp-block-styles' 
- * theme support (block css) thus the following needs to be included
- * since 'wp-block-styles' aren't used for this theme.
- * https://github.com/WordPress/gutenberg/issues/45065
- */
-.wp-block-table thead {
-	border-bottom: 3px solid;
-}
-.wp-block-table tfoot {
-	border-top: 3px solid;
```
mikachan, line 60: I think these are still needed too.

block-canvas/style.css
jffng: Thanks for following up — can we proceed with removing these or do you think some of the styles above are still needed?

mikachan: I guess it depends on how opinionated Block Canvas should be. I quite like the default styles of this caption being a smaller font size and centered, but perhaps we should remove these styles so we're relying on the defaults Core provides?

We could also use the `caption` element in theme.json to at least apply the small font size, rather than using CSS here.

block-canvas/style.css
mikachan: We can remove these lines now as Core includes them in the default styles: https://github.com/WordPress/gutenberg/pull/46567. 

We should probably wait for GB 14.9 to be released before merging this.

PR #6756, opened 2022-11-30
Muscat: add min height to post content

jffng:

This PR adds a 700px min height to post content block in the single and page templates.


pbking (approved):
Does what it says on the tin.

🚢

PR #6759, opened 2022-12-02
Blockbase + children: remove __unstableLocation

### LINKED ISSUE #6738: Eliminate __unstableLocation
Some themes (namely Blockbase and children) leverage the `__unstableLocation` attribute of the navigation block.

This was initially done when Blockbase was considered a "universal" theme which allowed people to leverage the customizer to edit navigation.  Blockbase has left its universal roots behind and needs to transition to be a simpler parent block theme.  This has already happened with colors and fonts and now needs to happen with navigation.

The difficulty is that any user who is leveraging their classically defined menus would lose them if the attribute were simply removed.  So some kind of migration will be necessary to transition users from using `__unstableLocation` to the native usage of the navigation block.

jffng:

https://github.com/WordPress/gutenberg/pull/45976 added a fallback for classic menus assigned to the `primary` location or named `primary`. 

Once that lands in the plugin (14.8), we should be able to safely remove this attribute without breaking user's navigations. 

**To test**

- Activate one of these themes and create a classic menu, assign it to the primary location
- Switch to this PR
- Verify the navigation appears as expected in the front end and site editor

#### Related issue(s):

Fixes https://github.com/Automattic/themes/issues/6738

### DIFF
--- a/blockbase/inc/social-navigation.php
+++ b/blockbase/inc/social-navigation.php
@@ -13,11 +13,6 @@ function blockbase_condition_to_render_social_menu( $block_content, $block ) {
 		return false;
 	}
 
-	// The block should have an unstable location attribute.
-	if ( empty( $block['attrs']['__unstableLocation'] ) ) {
-		return false;
-	}
-
 	// The block should be empty (no custom menu assigned)
 	if ( ! empty($block['attrs']['navigationMenuId']) || ! empty($block['attrs']['ref']) ) {
 		return false;
--- a/antonia/parts/header.html
+++ b/antonia/parts/header.html
@@ -8,7 +8,7 @@
 <!-- /wp:group --></div>
 <!-- /wp:group -->
 
-<!-- wp:navigation {"overlayMenu":"always","__unstableLocation":"primary","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal","flexWrap":"wrap"},"style":{"spacing":{"margin":{"top":"0"}}}} /--></div>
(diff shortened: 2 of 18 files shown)


blockbase/inc/social-navigation.php
```diff
@@ -13,11 +13,6 @@ function blockbase_condition_to_render_social_menu( $block_content, $block ) {
 		return false;
 	}
 
-	// The block should have an unstable location attribute.
```
pbking, line 16: Considering the social navigation render override only happens when an `__unstableLocation` is present none of this code will run.  I think we can completely remove the `social-navigation.php` file.

pbking, line 16: Is there still value in including this now that the __unstableLocation is gone?  
I don't see any themes that provide a `custom.navigation/social-links` value.

What other value does this provide?  Can we get rid of it completely?

jffng: > Can we get rid of it completely?

I think so, but is there potential for it to break how social menus are displayed for some sites running Blockbase or its children? There are five header parts that would appear to use this filter:

https://github.com/Automattic/themes/blob/trunk/blockbase/parts/header-default.html#L19
https://github.com/Automattic/themes/blob/trunk/blockbase/parts/header-linear.html#L19
https://github.com/Automattic/themes/blob/trunk/blockbase/parts/header-minimal.html#L21
https://github.com/Automattic/themes/blob/trunk/blockbase/parts/header-rounded-logo.html#L19
https://github.com/Automattic/themes/blob/trunk/blockbase/parts/header-wide.html#L19

pbking: Yes, I agree that for that reason it should stay.  The remainder of what's left is unconnected to `__unstableLocation` and removing it just removes functionality.

pbking (approved):
This is the change we need.  It's refreshing to get that attribute out.

This can ship once Gutenberg 14.8 is available on wpcom simple sites.

PR #6762, opened 2022-12-02
Loudness: Fix instructors pattern

pbking:
The pattern in question was complaining of missing attributes in wpcom environment.  Unfortunately adding those attributes caused block validation errors in a wporg environment.

To mitigate the problem I've removed the custom font assignment from the blocks in question.  The design is a little less great but there are no errors. 

Before:

image

After:

image


jffng (approved):
Works as expected, on initial insertion and reload in the editor, and view.

PR #6767, opened 2022-12-07
Tú: remove theme reference for the footer

iamtakashi:
I realised this after moving the theme to WP.com. The theme reference makes the footer not appearing on the homepage.

| Before | After |
| ----------- | ----------- |
| tuthemedemo wordpress com_ | tuthemedemo wordpress com_ (1) |


madhusudhand (approved):
LGTM!

PR #6775, opened 2022-12-12
Turn Pendant query loop inherit on by default

### LINKED ISSUE #6485: Pendant: Search template showing up results due to "Inherit query from template" being OFF by default
### Quick summary

_No response_

### Steps to reproduce

1. Activate Pendant
2. Check the Search template

When I search for something to receive the relevant results

### What actually happened

The "Inherit query from template" is OFF by default, so we ended up with the same Query Loop content

### Context

5504362-zen

Simple, Atomic

### Theme-specific issue?

Pendant

I couldn't replicate with Cultivate or Blockbase, but it appears there could be others as this was fixed in the past:

https://github.com/Automattic/wp-calypso/issues/62817

@jeffikus is there a way to target multiple themes and make sure this is turned on by default? 

_No response_

### Reproducibility

Intermittent

### Severity

Some (< 50%)

### Available workarounds?

Yes, easy to implement

### Workaround details

Turn on "Inherit query from template" in the Query Loop block

pbking:
Fixes: #6485 

I'm not sure why this was shipped off by default.  The Query loop has been configured to be on by default.  The pattern this loop is in is used in the Archive template as well as Index and Search (all of which I think it should be set to "true"

### DIFF
--- a/pendant/patterns/grid-of-posts.php
+++ b/pendant/patterns/grid-of-posts.php
@@ -6,7 +6,7 @@
  */
 ?>
 
-<!-- wp:query {"query":{"perPage":4,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"exclude","inherit":false},"tagName":"main","displayLayout":{"type":"flex","columns":2},"layout":{"inherit":true}} -->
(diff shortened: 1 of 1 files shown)


jffng (approved):
Good catch.

PR #6777, opened 2022-12-12
Allow Jetpack to register fonts that Blockbase has not.

### LINKED ISSUE #6765: Blockbase: Defer to Jetpack Fonts
Recently wpcom has started proxying fonts from google fonts API.  This may be enough to consider using Jetpack Fonts (which supplies fonts from there) to be used in Blockbase themes.

To do so Blockbase will need to be changed to check for Jetpack Fonts Module and, if active, not attempt to supply it's own font collection.

Currently the font collection is supplied in theme.json though and that is processed automatically so this may also require Blockbase to supply those fonts in a more sophisticated way (as a Font Provider as was originally intended).  It may be that the technical blockers that originally prevented that have since been dealt with (I can't recall offhand what exactly those where).

This task is mostly a prompt to dig into this to see if this is possible.

pbking:
I'm not sure that this works like it's supposed to but here's an idea for a fix to allow Jetpack curated fonts to work with blockbase.

This removes the "filter out all jetpack fonts" and instead just filters out those fonts that Blockbase is already providing.

I THINK that this was previously attempted but rejected because it was causing fonts from Blockbase not to work.  That doesn't seem to be the case any more (I think) so hopefully something has changed since then to allow this to work OK.

Fixes: #6765 

To test you'll need to have this running in an environment where Jetpack is providing fonts.  WPCOM is the easiest place to test this.  With this branch you should see additional Fonts that Jetpack is providing (at the time of my testing those were the following:

image

NOTE: Remember to update blockbase-premium as well!

### DIFF
--- a/blockbase/inc/fonts/custom-fonts.php
+++ b/blockbase/inc/fonts/custom-fonts.php
@@ -207,10 +207,20 @@ function enqueue_block_fonts( $content, $parsed_block ) {
 
 /**
  * Jetpack may attempt to register fonts for the Google Font Provider.
- * If that happens on a child theme then ONLY Jetpack fonts are registered.
- * This 'filter' filters out all of the fonts Jetpack should register
- * so that we depend exclusively on those provided by Blockbase.
+ * This filters out all of the fonts Blockbase has already registered.
  */
 function blockbase_filter_jetpack_google_fonts_list( $list_to_filter ) {
-	return array();
+	$font_families = array();
+	$filtered_list = array();
+	$fonts         = collect_fonts_from_blockbase();
+	foreach ( $fonts as $font ) {
+		$font_families[] = $font['name'];
+	}
+	foreach ( $list_to_filter as $jetpack_font_family ) {
+		if ( ! in_array( $jetpack_font_family, $font_families, true ) ) {
+			$filtered_list[] = $jetpack_font_family;
+		}
+	}
+	return $filtered_list;
 }
+


simison:
One of the problems I'm observing is that fonts from different sources end up in unexpected alphabetical order:

Screenshot 2022-12-15 at 12 44 49

blockbase/inc/fonts/custom-fonts.php
```diff
@@ -207,10 +207,20 @@ function enqueue_block_fonts( $content, $parsed_block ) {
...
- * so that we depend exclusively on those provided by Blockbase.
+ * This filters out all of the fonts Blockbase has already registered.
  */
 function blockbase_filter_jetpack_google_fonts_list( $list_to_filter ) {
-	return array();
+	$font_families = array();
+	$filtered_list = array();
+	$fonts         = collect_fonts_from_blockbase();
+	foreach ( $fonts as $font ) {
+		$font_families[] = $font['name'];
```
simison, line 217: Comparing by `name` seems to work, although I was expecting to have comparison by `slug`. No need to change tho.

simison (approved):
I'm approving this based on;

- ✅ Code looks fine
- ✅ Tests well in the editor
- 🟨 I'm not able to get most theme fonts work in front of the site with or without this PR; loading different fonts with themes/plugins is so messy that it's a bit hard to track where the problem is and it doesn't seem to be entirely specific to this PR. So please double check. Suspecting its due to localhost setup.
- 🟨   I'm not familiar with theme test process at all so I just copied the changed PHP lines to the theme pulled from .org repo, which might affect things.

Although I think we should still just disable Blockbase fonts entirely when Jetpack fonts module is active instead of doing this kind of "merge both sources" stuff, as Jetpack fonts:
    - works with all themes and experience is hence more consistent especially considering theme switches, 
    - is a single source that's easier to keep up to date. Blockbase seems to already have outdated fonts list. Current either/or setup is way too complex, and this PR is actually making it even more complicated (merged lists) although it does fix an issue. 
    - Jetpack fonts are now loaded from WP.com instead of Google

jffng (commented):
When testing I noticed a couple issues that gave me pause: 

1. The site editor did not appear to register any additional fonts from Jetpack. Which fonts should I expect to see? 

Screenshot 2022-12-19 at 10 38 59 AM

2. In the post editor, some additional fonts were registered but appended to the bottom of the list which feels weird, should this list be sorted somehow?

Screenshot 2022-12-19 at 10 39 38 AM

simison:
> The site editor did not appear to register any additional fonts from Jetpack. Which fonts should I expect to see?

[See list of additional fonts](https://github.com/Automattic/jetpack/pull/27441/files) (and I think there were some from previous additions)

Make also sure to enable Jetpack fonts module; 
1. Have latest Jetpack from dev or production. Dev has a few extra fonts (above PR)
1. connect jetpack
1. go to jetpack dashboard
1. at the footer, modules
1. Enable Fonts module

simison:
Those extra fonts seem to indeed be from Jetpack. [See the list.](https://github.com/Automattic/jetpack/blob/5f6b4ac5bd478019b5b1dade1d924fb4740cc790/projects/plugins/jetpack/modules/google-fonts.php#L22-L69)

jffng (approved):
This works as described, tested in both the editors and front-end. 

> Although I think we should still just disable Blockbase fonts entirely when Jetpack fonts module is active instead of doing this kind of "merge both sources" stuff, as Jetpack fonts:

I agree and looked into this a bit, at the moment there is no API for unregistering fonts AFAIK. A bit more context here: 

https://github.com/WordPress/gutenberg/issues/40288 

Ideally we would unregistering them programmatically if Jetpack fonts module is active, or not register them at all. We could rewrite the font registration + enqueuing in Blockbase, but I think that should be explored in a follow up. For now this PR allows the fonts list to be extended.

simison:
Regarding revert (https://github.com/Automattic/themes/pull/6788) I've been testing the changeset again on .com simple sites and can't replicate the "missing fonts" issue. Any hunch what I might be missing? Tested with Blockbase child theme "Appleton".

blockbase/inc/fonts/custom-fonts.php
```diff
@@ -207,10 +207,20 @@ function enqueue_block_fonts( $content, $parsed_block ) {
...
+	}
+	foreach ( $list_to_filter as $jetpack_font_family ) {
+		if ( ! in_array( $jetpack_font_family, $font_families, true ) ) {
+			$filtered_list[] = $jetpack_font_family;
+		}
+	}
+	return $filtered_list;
```
simison, line 224: Just to make this function easier to follow, something like this might make sense:

- Renamed vars, they weren't clear always where the data was coming from
- Condensed, hopefully without being too clever. Instead of `foreach` use `array_column`
- Considered using `array_filter` to remove second `foreach` but this felt still more readable.

```php
function blockbase_filter_jetpack_google_fonts_list( $jetpack_fonts ) {
	$theme_fonts         = collect_fonts_from_blockbase();
	$theme_font_families = array_column( $theme_fonts, 'name' );
	$filtered_list = array();

	// If the Jetpack font isn't in theme already, let Jetpack register it
	foreach ( $jetpack_fonts as $jetpack_font_family ) {
		if ( ! in_array( $jetpack_font_family, $theme_font_families, true ) ) {
			$filtered_list[] = $jetpack_font_family;
		}
	}

	return $filtered_list;
}
```

simison:
>I agree and looked into this a bit, at the moment there is no API for unregistering fonts AFAIK. A bit more context here:

@jffng BTW there [now](https://github.com/WordPress/gutenberg/pull/43492) is new `wp_deregister_font_family()` API

PR #6784, opened 2022-12-16
Block Canvas: Add home.html template

mikachan:
This adds a home.html template to Block Canvas, as it is currently not possible to add a home template via the Site Editor. This makes it possible to edit this template via the Site Editor, for example when using Block Canvas a base theme via Create Block Theme.

I used the same content as index.html.

Original discussion: p1670325836591219-slack-C029FM1EH


jffng (approved):
LGTM.

PR #6787, opened 2022-12-19
Upsidedown: Remove zero spacing from header and footer

mikachan:
This removes the margin and padding 0 styles from the header and footer of Upsidedown. This means that the global padding can work as expected at smaller resolutions, otherwise, the header and footer content have no padding at the edges of the page.

This is at 845px width:

| Before | After |
| ----- | ----- |
| image | image |

cc. @henriqueiamarino

### DIFF
--- a/upsidedown/patterns/footer.php
+++ b/upsidedown/patterns/footer.php
@@ -7,8 +7,8 @@
  */
 ?>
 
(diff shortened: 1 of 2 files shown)


jffng (approved):
LGTM!

PR #6788, opened 2022-12-20
Revert "Allow Jetpack to register fonts that Blockbase has not."

madhusudhand:
Reverts Automattic/themes#6777

Reverting this change as per the discussion p1670342089066579-slack-C029FM1EH


pbking:
I have discovered that the presence of `"provider": "blockbase-fonts"` is what has caused them to go missing.  I'm not sure WHY that has caused them to go missing, and to only go missing for CHILD fonts.  But I'm pretty sure that we can re-introduce the original change (#6777) to allow Blockbase and Jetpack to work together as expected.

PR #6790, opened 2022-12-20
Ignore tt3 screenshot

vcanales:
Doctom and Dotorg screenshots for this theme are different, so we need to avoid syncing when updating the theme, unless we explicitly intend to change the Dotcom image.

### DIFF
--- a/.sandbox-ignore
+++ b/.sandbox-ignore
@@ -25,3 +25,4 @@ inc/headstart
 languages
 *.map
 .DS_Store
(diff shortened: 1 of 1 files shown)


mikachan (approved):
This looks like it's working well for me! Here's an example diff it created: D96128-code

I ran `npm run deploy:core:sync twentytwentythree`. It does pick up that the screenshot.png needs updating, but when the diff is created, the screenshot file is skipped. I think this can come in 🚢

PR #6794, opened 2022-12-21
Blockbase: Register Fonts via Webfonts API instead of theme.json

jffng:

This PR changes the way Blockbase registers fonts, from theme.json to use the Webfonts API directly instead.  

This is done so that we can conditionally load them — only if Jetpack fonts are **not** active.

**To test**:

- Checkout this PR, activate Blockbase locally (no other plugins active), ensure custom fonts load as expected in the editors and front-end:

Screenshot 2022-12-21 at 11 39 16 AM

- Checkout this version of blockbase onto a wpcom simple site (you can use `npm run deploy:push:blockbase`, verify there are no duplicates and you see all the Jetpack fonts, e.g. with Zoologist:

Screenshot 2022-12-21 at 11 52 36 AM

#### Related issue(s):

Solves #6765, alternative to #6777 which failed for reasons I still can't quite identify.

#### More reasoning why

Quoting @simison [here](https://github.com/Automattic/themes/pull/6777#pullrequestreview-1219020200) for why I went with this approach instead:

> Although I think we should still just disable Blockbase fonts entirely when Jetpack fonts module is active instead of doing this kind of "merge both sources" stuff, as Jetpack fonts:
> - works with all themes and experience is hence more consistent especially considering theme switches,
> - is a single source that's easier to keep up to date. Blockbase seems to already have outdated fonts list. Current either/or setup is way too complex, and this PR is actually making it even more complicated (merged lists) although it does fix an issue.
>- Jetpack fonts are now loaded from WP.com instead of Google


blockbase/inc/fonts/custom-fonts.php
```diff
@@ -205,12 +199,37 @@ function enqueue_block_fonts( $content, $parsed_block ) {
...
+					'font-style'   => 'italic',
+					'font-display' => 'fallback',
+					'provider'     => $font_family['provider'],
+				),
+			)
+		);
+	}
 }
+
+add_action( 'init', 'enqueue_global_styles_fonts', 100 );
```
jffng, line 248: Note I just moved all action and filter registrations to the bottom since it seems to be more of a convention to do this way.

blockbase/functions.php
```diff
@@ -102,8 +102,13 @@ function blockbase_scripts() {
 	require get_template_directory() . '/inc/social-navigation.php';
 }
 
-require get_template_directory() . '/inc/fonts/custom-fonts.php';
 
+/**
+ * Do not use the custom fonts provided by Blockbase if Jetpack fonts module is active
+ */
+if ( ! class_exists( 'Jetpack' ) || ! Jetpack::is_module_active( 'google-fonts' ) ) {
```
jffng, line 109: @simison is this a safe way to check?

jffng:
Ah I just realized this is going to break Blockbase fonts for sites that are not running Gutenberg which exposes the public webfonts API currently.

jeffikus:
@jffng I like this approach. I especially like that you've separate out the json - that seems more efficient. 

About Blockbase breaking for sites not running Gutenberg, I remember discussing our assumption was that Blockbase was assuming Gutenberg was active, but gracefully degrades when it isn't, or am I remembering that incorrectly?

blockbase/functions.php
simison: I think so! but let's @Automattic/jetpack-crew for 2nd opinion

simison: Clever doing module check instead of "is dotcom" check 👍

blockbase/inc/fonts/custom-fonts.php
```diff
@@ -205,12 +199,37 @@ function enqueue_block_fonts( $content, $parsed_block ) {
...
+					'provider'     => $font_family['provider'],
+				),
+				array(
+					'font-family'  => $font_family['name'],
+					'font-weight'  => '100 900',
+					'font-style'   => 'italic',
+					'font-display' => 'fallback',
+					'provider'     => $font_family['provider'],
+				),
+			)
```
simison, line 243: FYI using `wp_register_webfonts` with this array signature is fine _today_, we'll just need to migrate to new format once the new API stabilizes, as this exact format got deprecated.

More in https://github.com/WordPress/gutenberg/pull/43492

jffng: Gotcha. If this approach is acceptable, we can open a follow-up PR to change the format and merge it after the new API lands in Gutenberg 14.9.

jffng:
@jeffikus 

> About Blockbase breaking for sites not running Gutenberg, I remember discussing our assumption was that Blockbase was assuming Gutenberg was active, but gracefully degrades when it isn't, or am I remembering that incorrectly?

That sounds right, by this assumption do you think it's acceptable to only have the system font available?

Note that this would only be until the public Webfonts API lands in Core.

jeffikus:
@jffng I am comfortable with that being the assumption in order to get the best out of the theme. System fonts seems a good fallback on .org - .com won't have that problem though.

@mikachan @ianstewart @matiasbenedetto @vcanales @madhusudhand @scruffian @MaggieCabrera any objections to this assumption?

@jffng we would also need to test this with the Blockbase premium themes and merge it across to the premium version of blockbase. But that's a separate implementation. Let's get the main Blockbase implementation complete first.

blockbase/functions.php
jeherve: Yes, it should work. The module will always be considered as "active" on WordPress.com simple, and is force-activated on WoA, so you'll end up using the module on both platforms at all times.

blockbase/inc/fonts/custom-fonts.php
simison: Yep exactly. We'll also need to migrate Jetpack https://github.com/Automattic/jetpack/pull/28054

simison:
>do you think it's acceptable to only have the system font available?

Sort of related to adding more safe web fonts in Gutenberg https://github.com/WordPress/gutenberg/issues/39552

mikachan:
> any objections to this assumption?

I like this approach, all sounds good to me 👍

jffng:
@jeffikus were you able to test this, locally or on dotcom?

blockbase/theme.json
```diff
@@ -347,197 +347,6 @@
 					"fontFamily": "-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif",
 					"slug": "system-font",
 					"name": "System Font"
-				},
-				{
-					"fontFamily": "Arvo, serif",
```
simison, line 352: Since these fonts were `Arvo, serif` previously here, will sites continue functioning if the font now is registered with just `Arvo` in global styles?

blockbase/theme.json
```diff
@@ -347,197 +347,6 @@
...
-					"provider": "blockbase-fonts"
-				},
-				{
-					"fontFamily": "'Fira Sans', sans-serif",
-					"slug": "fira-sans",
-					"name": "Fira Sans",
-					"provider": "blockbase-fonts"
-				},
-				{
-					"fontFamily": "'Helvetica Neue','Helvetica', 'Arial', sans-serif",
```
simison, line 406: Since these Helvetica+Arial fonts are from the OS, should you keep them here?

blockbase/functions.php
simison: I tested this PR in .com simple sites and the file didn't get loaded. 👍

blockbase/inc/fonts/custom-fonts.php
simison: More info here: https://github.com/Automattic/jetpack/issues/28063

simison:
> were you able to test this, locally or on dotcom?

I did test on .com simple sites — works fine (https://github.com/Automattic/themes/pull/6794#discussion_r1057757601). Just a bit concerned if old sites continue working (https://github.com/Automattic/themes/pull/6794/files#r1057756452).

blockbase/functions.php
anomiex: Is there any chance that the font module will become part of a standalone plugin at some point? If so, it might be better to do it like this:
```suggestion
if ( ! class_exists( \Automattic\Jetpack\Modules::class ) || ! ( new \Automattic\Jetpack\Modules() )->is_active( 'google-fonts' ) ) {
```
(feel free to put a `use Automattic\Jetpack\Modules;` at the top of the file instead of using a fully-qualified class name here, of course).

pbking:
>> About Blockbase breaking for sites not running Gutenberg, I remember discussing our assumption was that Blockbase was assuming Gutenberg was active, but gracefully degrades when it isn't, or am I remembering that incorrectly?

> That sounds right, by this assumption do you think it's acceptable to only have the system font available?

While that's the assumption I also have this scenario feels different in that the additional fonts ARE currently available with Blockbase (or any of the children) without Gutenberg activated.  So removing that functionality without waiting for an opportunity to refactor it seems like breakage rather than a degradation.

blockbase/theme.json
jffng: Yes, since the slug is just `Arvo` and thus generated CSS variable that applies the font is the same.

blockbase/theme.json
jffng: Oh yeah, we should keep this in theme.json. 

Done in https://github.com/Automattic/themes/pull/6794/commits/6bf79fface6926144b7457da703b8f3868a39656

jffng:
@pbking that's a fair point. This approach would break fonts for Blockbase + children that are not running the plugin, until the refactor webfonts API lands in core. And we don't really know what the surface area of sites that really affects is.

> without waiting for an opportunity to refactor it

Do you have any other ideas for how we might refactor this in the meantime? I do not know why https://github.com/Automattic/themes/pull/6777 did not work.

pbking:
@jffng the only alternative I can think of that doesn't negatively impact non-plugin-users is to add the fonts to Blockbase.  I have opened a change to do that here: https://github.com/Automattic/themes/pull/6820

hellofromtonya:
**Heads up:**
The Fonts API architectural work is now complete. 

- [The API renaming work merged today](https://github.com/WordPress/gutenberg/pull/46749) into Gutenberg `trunk`. For example, `wp_register_webfonts()` is now `wp_register_fonts()`.  

- Registering fonts through a theme's `theme.json` file is NOT impacted as the APIs automatically ingest, process, register, and enqueue defined fonts.

- Directly registering fonts via `wp_register_fonts()`: the data structure to register changed:
    - Variations are grouped to their font-family
    - The font-family is the key for this group of variations

For example:
```php
foreach ( $fonts_to_register as $font_family ) {
	wp_register_fonts(
		array(
			$font_family => array(
				array(
					'font-family'  => $font_family,
					'font-weight'  => '100 900',
					'font-style'   => 'normal',
					'font-display' => 'fallback',
					'provider'     => 'jetpack-google-fonts',
				),
				array(
					'font-family'  => $font_family,
					'font-weight'  => '100 900',
					'font-style'   => 'italic',
					'font-display' => 'fallback',
					'provider'     => 'jetpack-google-fonts',
				),
			),
		)
	);
}
```

Also added notes to Jetpack's Google Fonts Provider https://github.com/Automattic/jetpack/issues/28063#issuecomment-1387090575.

jffng:
Amazing, thank you for the update @hellofromtonya! The plan is for this updated API to make it into 6.2, right?

hellofromtonya:
>The plan is for this updated API to make it into 6.2, right?

Yes, hopeful for it to land in WP 6.2 🤞 @jffng

pbking:
I would like to ship this change (or some version of it), but not until 6.2 lands.  Until then #6831 should get wpcom where it needs to be.

PR #6806, opened 2022-12-29
Fix background color variable

onuro:
Theme was using --wp--preset--color--base variable for the background, while it should've been --wp--preset--color--background.

Corrected the assigned color for background.


madhusudhand (approved):
LGTM!

PR #6820, opened 2023-01-10
Added new collection of fonts to Blockbase

pbking:
This change adds all of the new fonts currently offered by Jetpack Webfont Provider so that Blockbase themes have feature parity since those fonts are currently provided by the theme.

The fonts added are the same as are added in [this change in Jetpack](https://github.com/Automattic/jetpack/pull/27441/files).

This is an alternative to attempts at allowing Blockbase to optionally load fonts (which requires Gutenberg to be activated) or allow (but filter) Jetpack's loaded fonts (which causes other problems).


pbking:
The following script was used to grab the fonts from Google Fonts API:

```
<?php

$family = $argv[1];
$family_slug = str_replace( ' ', '-', strtolower($family));


mkdir( './' . $family_slug );

ini_set('user_agent', 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36');
file_put_contents(  './' . $family_slug . '/font-face-source.css', fopen( 'https://fonts.googleapis.com/css2?format=woff2&family='.str_replace( ' ', '%20', $family).':ital,wght@0,100;0,200;0,300;0,400;0,500;0,600;0,700;0,800;0,900;1,100;1,200;1,300;1,400;1,500;1,600;1,700;1,800;1,900&display=swap', 'r' ) );

$contents = file_get_contents( './' . $family_slug . '/font-face-source.css' );
$modified_contents = $contents;
$matches;
$typefaces_to_download = Array();

$weight_conversion = array(
	'100' => 'Thin',
	'200' => 'ExtraLight',
	'300' => 'Light',
	'400' => 'Regular',
	'500' => 'Medium',
	'600' => 'SemiBold',
	'700' => 'Bold',
	'800' => 'ExtraBold',
	'900' => 'Black',
);

preg_match_all( '/\/\*.*}/sU', $contents, $matches );

foreach( $matches[0] as $match ) {

	preg_match('/(?<=\/\* ).*(?= \*\/)/', $match, $language );
	if( $language[0] == 'latin' ) {
		$language = '';
	} else {
		$language = '-' . $language[0];
	}

	preg_match('/(?<=font-style: ).*(?=;)/', $match, $style);
	preg_match('/(?<=font-weight: ).*(?=;)/', $match, $weight);
	preg_match('/(?<=src: url\().*?(?=\))/', $match, $url);

	$url = $url[0];

	$weight = '-' . $weight_conversion[$weight[0]];

	if($style[0] == 'normal' ) {
		$style = '';
	} else {
		$style = '-' . $style[0];
	}

	$new_filename = $family_slug . $language . $weight . $style . '.woff2';

	$modified_contents = str_replace( $url, './' . $new_filename, $modified_contents);

	if ( ! array_key_exists( $url, $typefaces_to_download ) ) {
		$typefaces_to_download[$url] = './' . $family_slug . '/' . $new_filename;
	}
	

}

file_put_contents( './' . $family_slug . '/font-face.css', $modified_contents );

foreach( $typefaces_to_download as $url => $new_typeface_filename ) {
	file_put_contents(  $new_typeface_filename, fopen($url, 'r' ) );
}
```

jffng (commented):
I am conflicted about this.

I agree we can't break fonts for existing sites that do not run Gutenberg. But this change feels directly targeted at dotcom, which runs Gutenberg, to reach feature parity with the fonts that Jetpack provides.

In general, I don't think Blockbase should increase the number of fonts it provides + manages, for a few reasons:

1. Managing a static font set is additional maintenance in Blockbase, so increasing the number of fonts would potentially increase the number of fonts we might want to deprecate down the line. 
2. We don't have any feedback from non-dotcom users that more fonts in the theme are desirable.
3. This increases the theme's size by about ~3mb which feels like a significant bump. Granted the fonts are only loaded if it's requested.

My preference would be to wait until the public webfonts API lands in core so we can dynamically register the fonts from Blockbase.

pbking:
Blockbase now works well with Jetpack fonts.  No need to add these fonts here; closing.

PR #6831, opened 2023-01-23
Fix/allow jetpack to register fonts (take 2)

pbking:
This is the same change originally made (and reverted) in #6777.  It was refactored [per suggestion](https://github.com/Automattic/themes/pull/6777#discussion_r1053380947).

This change requires [this gutenberg fix](https://github.com/WordPress/gutenberg/pull/47290) to be included before it works as expected with child themes.

To test ensure that you are in an environment where Jetpack is supplying additional Fonts (such as wpcom) and the Gutenberg change noted has been applied.  Activate a Blockbase Child theme and note the available Fonts.  All of the fonts offered by BOTH Blockbase and Jetpack should be available in the Site Editor and rendered as expected in the view.

### DIFF
--- a/blockbase/inc/fonts/custom-fonts.php
+++ b/blockbase/inc/fonts/custom-fonts.php
@@ -207,10 +207,20 @@ function enqueue_block_fonts( $content, $parsed_block ) {
 
 /**
  * Jetpack may attempt to register fonts for the Google Font Provider.
- * If that happens on a child theme then ONLY Jetpack fonts are registered.
- * This 'filter' filters out all of the fonts Jetpack should register
- * so that we depend exclusively on those provided by Blockbase.
+ * This filters out all of the fonts Blockbase has already registered.
  */
 function blockbase_filter_jetpack_google_fonts_list( $list_to_filter ) {
-	return array();
+	$font_families = array();
+	$filtered_list = array();
+	$fonts         = collect_fonts_from_blockbase();
+	foreach ( $fonts as $font ) {
+		$font_families[] = $font['name'];
+	}
+	foreach ( $list_to_filter as $jetpack_font_family ) {
+		if ( ! in_array( $jetpack_font_family, $font_families, true ) ) {
+			$filtered_list[] = $jetpack_font_family;
+		}
+	}
+	return $filtered_list;
 }
+
--- a/blockbase/inc/fonts/custom-fonts.php
+++ b/blockbase/inc/fonts/custom-fonts.php
@@ -209,18 +209,16 @@ function enqueue_block_fonts( $content, $parsed_block ) {
  * Jetpack may attempt to register fonts for the Google Font Provider.
  * This filters out all of the fonts Blockbase has already registered.
  */
-function blockbase_filter_jetpack_google_fonts_list( $list_to_filter ) {
-	$font_families = array();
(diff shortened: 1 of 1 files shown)


jffng:
Thanks for following up on this @pbking . Would it be okay if we waited until that required Gutenberg change lands on wpcom (15.1 I think is next week?) to test this change?

pbking:
> Would it be okay if we waited until that required Gutenberg change lands on wpcom (15.1 I think is next week?) to test this change?

Yea, that would definitely be the easiest way to verify and this feature couldn't ship until then anyway.

I've tested pretty well locally so I'm pretty confident in the change and am fine waiting until then.

jffng (commented):
Unfortunately this is still not working in my testing on wpcom: 

Blockbase | Pendant
--- | ---
Screenshot 2023-02-14 at 1 18 35 PM | Screenshot 2023-02-14 at 1 17 31 PM

mikachan (commented):
I've tested this out on a wpcom simple site with GB 15.1.1. I used [this file](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/modules/google-fonts.php) to determine which additional fonts I should see. I used Arbutus, Quadrat, and Zoologist.

This looks like it's working great in the Post Editor, as I can see 52 fonts here, and this list seems to correctly include the filtered Jetpack fonts. 

However, in the Site Editor, I'm only seeing 34 fonts, and I think the Jetpack fonts are missing. This is the same as what I see in the Post Editor without this PR.

| Site Editor (34 fonts) | Post Editor (52 fonts) |
| -------- | --------- |
| image | image |

In TT3 on wpcom, I see 50 fonts in both the Site and Post Editors.

I'm not sure if I've tested this correctly, but I'm guessing the font list should be the same in both editors. Maybe the Site Editor on wpcom is using the older version of GB still?

pbking:
That's interesting!  I've never seen a discrepancy before.  I believe I tested exclusively in the site editor so I'm surprised to see that this is ONLY working as expected in the post editor.

Indeed, all environments should include all 50-ish fonts.

I'll give this another test today as well and see if I can shake out what's going on.

pbking:
I tested in wpcom simple and found the same results as @mikachan.  

I used the same version of plugins (Gutenberg and Jetpack) locally and found all fonts to be loading as expected.  I'm not sure what the difference in environment would be causing the fonts to fail to load.

pbking:
Strangely: I took out ALL of the font customization bits from Blockbase (on my sandbox): I didn't call the font customization scripts and I even deleted the custom fonts defined in the theme.json file.  And I STILL saw all of the Fonts from Blockbase and none from Jetpack.  At that point I don't understand what would have been different about Blockbase and Twenty Twenty Three (where all of the Jetpack Fonts ARE available).

I'm very very confused right now...  It's almost as though the fonts being supplied are cached in some way some where some how.  

I will continue to dig.


...


[edit]

Bah.  I wasn't sandboxing the API.

It seems to work as expected when I do.

mikachan (approved):
> Bah. I wasn't sandboxing the API.
> It seems to work as expected when I do.

Sames! I can confirm I can see all the fonts, and the same number of fonts, in both editors when the API is sandboxed 🎉

I think this should be good to bring in now!

PR #6846, opened 2023-01-31
Bitacora: Initial commit

mikachan:
This adds Bitacora, a new theme designed and built by @iamtakashi ✨


mikachan:
I've tested this theme and its variations out, everything looks great to me. I think this is ready to bring in 🚢

PR #6848, opened 2023-02-01
Vetro: more design adjustments and fixes

jffng:

Before | After
--- | ---
Screenshot 2023-02-01 at 12 45 06 PM | Screenshot 2023-02-01 at 12 43 56 PM

Please also give the theme a once-over as this will serve as a sort of final review, thanks!

### DIFF
--- a/vetro/patterns/about.php
+++ b/vetro/patterns/about.php
@@ -6,17 +6,17 @@
  */
 ?>
 
-<!-- wp:group {"align":"full","layout":{"type":"constrained"}} -->
-<div class="wp-block-group alignfull"><!-- wp:heading {"textAlign":"left","level":1,"style":{"typography":{"lineHeight":"1.1","fontStyle":"normal","fontWeight":"600","letterSpacing":"-3px","fontSize":"4rem"},"spacing":{"padding":{"top":"0px","bottom":"0px"},"margin":{"top":"0px","right":"0px","bottom":"0px","left":"0px"}}}} -->
-<h1 class="wp-block-heading has-text-align-left" style="margin-top:0px;margin-right:0px;margin-bottom:0px;margin-left:0px;padding-top:0px;padding-bottom:0px;font-size:4rem;font-style:normal;font-weight:600;letter-spacing:-3px;line-height:1.1"><?php echo esc_html__( 'United Architects Collab is a studio focused on sustainable design, interiors and landscapes.', 'vetro' ); ?></h1>
+<!-- wp:group {"align":"wide","layout":{"type":"constrained"}} -->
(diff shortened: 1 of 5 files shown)


mikachan (commented):
The change in this PR is looking good.

I noticed the following when doing a final review of the theme:

1. The post excerpt in the single template sits very close to the border on the header:
image

2. If there is an image at the end of a post/page, there is no spacing between the image and the bottom border:
image

3. A couple of small issues with buttons:

- There are no hover, focus, or active states on the buttons.
- In the neutral variation, the button outline style doesn't have an outline/border.

jffng:
@mikachan thank you for the review!

> The post excerpt in the single template sits very close to the border on the header:

Hmm, this is what that gap looks like for me, would you able to able to share what's going on there in the inspector?

Screenshot 2023-02-02 at 10 19 52 AM

> If there is an image at the end of a post/page, there is no spacing between the image and the bottom border:

Ah this is actually by design, but the border should be full width. Fixed in [5bdc61b](https://github.com/Automattic/themes/pull/6848/commits/5bdc61b277215f95a43aa486f5a548343b39cf8b)

> There are no hover, focus, or active states on the buttons.

Fixed in [06dca00](https://github.com/Automattic/themes/pull/6848/commits/06dca00ba78ec87ec60405a54a532dba37159aad)

> In the neutral variation, the button outline style doesn't have an outline/border.

Fixed in [6c5df65](https://github.com/Automattic/themes/pull/6848/commits/6c5df6546491c146dce9efc395f4a346d6ca7e39)

mikachan:
> Hmm, this is what that gap looks like for me, would you able to able to share what's going on there in the inspector?

Oh! Ok, so the excerpt block in my local editor does have spacing, but the front end still doesn't. I'm running the latest GB trunk. Here's what I see with the inspector on the front end:

image

And in the editor, I see spacing, which is coming from the default margins on the paragraph element:

image

These default margins are overwritten on the front end by `.wp-block-post-excerpt__excerpt` setting them to 0.

(The other fixes look good 👍)

jffng:
Thank you again @mikachan ! I wasn't running trunk and there was a recent change to the post-excerpt block css that zero'd out those margins. 

I am not sure if it's the intended behavior, but I just made a fix in the theme anyways: [1fe9c4b](https://github.com/Automattic/themes/pull/6848/commits/1fe9c4b8cfedf9026548d68e1f116f734f4c4fba)

mikachan (approved):
> I am not sure if it's the intended behavior, but I just made a fix in the theme anyways: [1fe9c4b](https://github.com/Automattic/themes/pull/6848/commits/1fe9c4b8cfedf9026548d68e1f116f734f4c4fba)

Looks good to me!

PR #6860, opened 2023-02-06
Block Canvas: Add no results content to search template

iamtakashi:
The lack of no result contents in the search template was repeatedly mentioned in WP.com, and it'd be nice if Block Canvas resolves the issue, so that it's not necessary for theme designers to handle the case every time.

| Before | After |
| ----------- | ----------- |
| iamtakashi jurassic tube__s=Pneumonoultramicroscopicsilicovolcanoconiosis (1) | iamtakashi jurassic tube__s=Pneumonoultramicroscopicsilicovolcanoconiosis |

I also fixed a minor issue with the tag for `.wp-block-query` in the search template in this PR.

### DIFF
--- a/block-canvas/patterns/hidden-no-results-content.php
+++ b/block-canvas/patterns/hidden-no-results-content.php
@@ -1,0 +1,10 @@
+<?php
+/**
+ * Title: Hidden No Results Content
+ * Slug: block-canvas/hidden-no-results-content
+ * Inserter: no
+ */
+?>
+<!-- wp:paragraph -->
+<p><?php echo esc_html_x( 'Sorry, but nothing matched your search terms. Please try again with some different keywords.', 'Message explaining that there are no results returned from a search', 'block-canvas' ); ?></p>
+<!-- /wp:paragraph -->
--- a/block-canvas/templates/search.html
+++ b/block-canvas/templates/search.html
@@ -10,7 +10,7 @@
 </div>
 <!-- /wp:group -->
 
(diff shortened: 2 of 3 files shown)


mikachan (commented):
Thanks for doing this!

It looks like there's a package-lock.json included here by mistake, so this should be removed. Otherwise, this is looking good to me.

However, the no results block is not behaving as I expect. It seems that the no results message is always shown regardless of the search results. I've opened an issue in Gutenberg here: https://github.com/WordPress/gutenberg/issues/47793

mikachan (commented):
Ugh sorry @iamtakashi, just realised that this is the package-lock.json in the root of the repo. We do commit this, but there shouldn't be any changes to commit as part of this PR.

iamtakashi:
Sorry @mikachan, something went wrong in the initial commit. It should be good now.

iamtakashi:
> However, the no results block is not behaving as I expect. It seems that the no results message is always shown regardless of the search results.

Hmm, interesting, that does sound familiar, but I can't reproduce it now.

iamtakashi:
> but I can't reproduce it now.

I said it too soon 🙂 I can reproduce that now.

mikachan (approved):
🚢

iamtakashi:
@mikachan Since the theme isn't on WP.com, is it simply to merge it with the button above? Is there anything else we need to do other than merging? It's the first time for me to contribute to Block Canvas.

mikachan:
Yes that's right, with Block Canvas we just need to merge the PR.

PR #6862, opened 2023-02-10
Add an automated zip workflow

pbking:
#### Changes proposed in this Pull Request:

This action will automatically open a PR when a .zip file is attached to an issue BODY or COMMENT.

(NOTE: I think that this will need to be MERGED before it will actually run which makes it kinda hard to test...  You can see [many] [test runs here](https://github.com/pbking/zip-extraction-practice/actions/workflows/extrakt.yml) to see what this change is supposed to accomplish).

This action allows for the following workflow:

* A designer creates a theme locally (or on a demo site).
* They export the theme as a .zip file (using Create Block Theme) i.e. `banana.zip`.
* They open a [New Issue](https://github.com/Automattic/themes/issues/new) in this repository.  In the BODY of that issue they attach (just drag 'n drop) the theme .zip file that was exported.
* This script will then create a new branch (based on the name of the .zip file i.e. `auto/banana` ).
  * If a branch of that name already exists then it will be checked out instead of created
* The contents of that zip file will be extracted into a folder (also of the same name as the .zip file).  
* The new (or changed) files will be committed to that branch.
* A Pull Request will be opened to merge these changes to the trunk branch and associated with the originally opened Issue (noted in step 3).
* Anybody can then check out & review the changes; it's just a normal branch.  Comments can be made on the PR (or issue).
* To make changes to the PR the designer can simply repeat the steps; exporting the theme .zip file using Create Block Theme and either:
  * replacing the zip file on the issue
  * (recommended) adding a comment with a new zip file to the issue
* Each time a zip file is added to the issue a new commit is made to the branch
* Once the changes are approved the branch can be merged

This closes #6850

PR #6868, opened 2023-02-20
Add issue template to assist new theme submission

madhusudhand:

This is an extension to #6862 to help with issue creation.
It adds new option to new issue creation, and simplifies the required fields.

image

image


pbking:
I think that if we went with the .zip PR auto-creation this would be helpful, but I no longer think that that path is the most helpful to designers.

I'm gonna close this (as well as the .zip automation tasks)

PR #6871, opened 2023-02-21
Adventurer: Updates ready for dotorg

mikachan:
This PR removes the unused images from Adventurer and adds the license info for the included font and image assets. I swapped an image that was from Unsplash with an image from StockSnap, as StockSnap uses the CC0 license.

It also bumps the theme version ready for submission to dotorg.

With so many images not being used in the theme files (templates or patterns), it's made me wonder if this theme is missing some patterns. @beafialho, could you confirm if all the expected patterns are included in the theme files? The images are only used in the "About" and "Header with cover" patterns. Here's a list of the pattern files if it helps:

image


madhusudhand (commented):
Bumped min wp version to 6.1 because it uses fluid typography.
Compressed screenshot and images from assets

Even after 50% compression, `coconuts.png` is still `730 kb` and beyond allowed limits.
@beafialho can you suggest a different image with less size?

Everything else looks good. 🍾

pbking:
I converted the image from png to jpg and it went to 147k.  

I think this is ready to ship now.

Issue #6877, opened 2023-02-23
Launch new theme Startup


madhusudhand:
### Launch checklist

- [ ] Add theme to github [themes repo](https://github.com/Automattic/themes)
- [ ] Check theme name conflict (if any)
- [ ] Compress assets
- [ ] Update theme description
- [ ] Update readme.txt
- [ ] Add credits in readme
- [ ] Screenshot for showcase entry
- [ ] Create Showcase entry
- [ ] Create demo site
- [ ] Create headstart annotation
- [ ] Update dotOrg deployment list

PR #6878, opened 2023-02-23
Add new theme "Iotix" (formerly Startup)

madhusudhand:

Adds new theme `Startup`

image

#### Related issue(s):

#6877


madhusudhand:
Following are the initial observations of the theme.

* Missing screenshot
* Block styles mismatch for Site title, buttons, links etc..
* Different states such as hover, active needs fixes
* home.html for the default content

**Post**

image image

**Page**

image

**Archive**

image

jffng:
Perhaps we can have design weigh in on those issues? cc @beafialho

beafialho:
> Block styles mismatch for Site title, buttons, links etc..
> Different states such as hover, active needs fixes

Any idea why this is happening? 🤔 

Here's the theme screenshot: 

screenshot

jffng:
I made the following updates manually:

- Removed unneeded templates
- Escaped text and images, patternized as needed
- Added screenshot
- Removed nav refs

There are some outstanding styling issues but I think they'll be easier to notice and address when we have a demo site on wpcom.

The one thing we should do before merging is rename the theme since the name is taken.

PR #6884, opened 2023-02-28
Remove block-level duotone from skatepark patterns and templates

ajlende:

Remove unnecessary block-level duotone filters in skatepark—these filters are already applied globally.

Additionally, it will be easier to use these patterns in other themes since the custom duotone won't be attached.

### DIFF
--- a/skatepark/patterns/mixed-media-in-container.php
+++ b/skatepark/patterns/mixed-media-in-container.php
@@ -27,7 +27,7 @@
 <div style="height:40px" aria-hidden="true" class="wp-block-spacer"></div>
 <!-- /wp:spacer -->
 
-<!-- wp:image {"sizeSlug":"large","style":{"color":{"duotone":["#000","#BFF5A5"]}}} -->
+<!-- wp:image {"sizeSlug":"large"} -->
 <figure class="wp-block-image size-large"><img src="<?php echo get_stylesheet_directory_uri() . '/assets/images/skateboard-sideways.jpg'; ?>" alt="<?php echo esc_attr__( 'A skateboard laying on its side on top of concrete.', 'skatepark' ); ?>"/></figure>
 <!-- /wp:image -->
 
--- a/skatepark/block-templates/page.html
+++ b/skatepark/block-templates/page.html
@@ -9,7 +9,7 @@
 
 <!-- wp:group {"align":"full"} -->
 <div class="wp-block-group">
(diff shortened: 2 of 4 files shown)


skatepark/patterns/mixed-media-in-container.php
```diff
@@ -27,7 +27,7 @@
 <div style="height:40px" aria-hidden="true" class="wp-block-spacer"></div>
 <!-- /wp:spacer -->
 
-<!-- wp:image {"sizeSlug":"large","style":{"color":{"duotone":["#000","#BFF5A5"]}}} -->
+<!-- wp:image {"sizeSlug":"large"} -->
```
jeryj, line 30: I checked the `theme.json` for skatepark and see [`core/image` is using the default duotone filter](https://github.com/Automattic/themes/blob/trunk/skatepark/theme.json#L312) which is [`["#000","#B9FB9C"]`](https://github.com/Automattic/themes/blob/trunk/skatepark/theme.json#L37). This image uses a slightly different (more muted) green of `#BFF5A5`. I'm not sure if that was originally intentional or a mistake. I don't think anyone would notice the change, but I wanted to point it out here to make sure.

jeryj: I see a precedent [for changing from `#BFF5A5` to `#B9FB9C`](https://github.com/Automattic/themes/commit/f1a3133076e478decfcb97e2efb9f47198d015c4#diff-c6ba9c8ac9c93710eaec5f8bac0f44d1d874aaa5c1d289dd72b38960726909eaL32), and honestly can't imagine someone would visually notice the distinction between the two, so I don't think this should be a blocker.

jeryj:
I used the `columns-in-container` template to compare `trunk` to this branch and they present identically on the frontend, but in the editor the duotone from theme doesn't apply. The top image is inserted from `trunk` template with the manual duotone vs the theme.json applied duotone on the bottom which doesn't show the duotone in the editor. I'm guessing that's a separate issue, but commenting to confirm.

Editor view of skatepark columns in container template

MaggieCabrera (approved):
Yes, I think this is a good change. The bug where the filter is missing on the editor for the patterns is a duotone issue that we should fix in the editor. I will look for the issue and if there is none I'll open one.

MaggieCabrera:
This is the one! https://github.com/WordPress/gutenberg/issues/48586

jffng (approved):
Cool work! LGTM, also noted that issue y'all already caught above with the post editor.

Issue #6879, opened 2023-02-24
[Bug]: Yuga: Heading Font Set Via Global Styles Does Not Display


katiebethbrown:
_Edit: Please see https://github.com/Automattic/themes/issues/7152 for larger context and workarounds ([also noted below](https://github.com/Automattic/themes/issues/6879#issuecomment-1601797597))_

## Quick summary

When using the Yuga theme, the font family selected for Headings in Global Styles does not display in the editor or on the live site.

## Steps to reproduce

1. Open the site editor.
2. Open the Styles panel.
3. Select `Typography`.
4. Select `Headings`.
5. Choose a font.

<img width="700px" src="https://user-images.githubusercontent.com/42279181/221067626-f26e7d7d-0f7d-4f43-9bd7-d6997fb49934.png">

### What you expected to happen

For the Heading blocks to use the selected font, both in the editor and on the live site.

### What actually happened

- The Site Title block updates to show the correct font.
- The Heading blocks in the content do not change.

<img width="300px" src="https://user-images.githubusercontent.com/42279181/221067688-4a026f61-f83d-4e26-9840-be1cf1716222.png">

## Workaround details

> In order to style these elements from Global Styles, you'll have to go to `Global Styles > Blocks > Heading > Typography`;

I have confirmed that the above approach works on my site, but I'd prefer not to record that into a how-to tutorial video if we're able to correct the theme so the main Global Styles work as expected.

## Root Cause
>In the case of Yugo, we've noticed that it's caused because the theme styles the `core/heading` element with a font. Because of this Global Styles (font family in this case) defined in Typography > Headings are overridden. In order to style these elements from Global Styles, you'll have to go to Global Styles > Blocks > Heading > Typography;

> I don't have enough context of why both themes have the fontFamily set for the blocks; maybe it's something that could be discussed with the themes team.

cc: @iamtakashi b/c I see there was some previous discussion about fonts in Yuga (pNEWy-fcl-p2#comment-54800).

## Other notes

Originally reported here:
- https://github.com/Automattic/wp-calypso/issues/71641#issuecomment-1440797947

Context: Team51, recording a video to share on public socials

Reproducibility: Consistent

Severity: This issue appears to also impact the Lettre theme, and possibly others as well.

Available workarounds: Yes, easy to implement but not obvious.

iamtakashi:
Thanks for the ping. I now see the problem in the themes, Yuga and Masu. I'll update the themes asap.

[...]

PR #6885, opened 2023-02-28
Masu: Remove heading block specific font rule

iamtakashi:
Related: #6879

Yuga and Masu share the problem, and this will fix the issue in Masu. I'll fix Yuga in the premium theme repo.

### DIFF
--- a/masu/theme.json
+++ b/masu/theme.json
@@ -47,14 +47,7 @@
 			"wideSize": "1720px"
 		},
 		"spacing": {
-			"units": [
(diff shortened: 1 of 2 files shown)


jffng (approved):
Makes sense to me.

iamtakashi:
Thanks, @jffng, for reviewing and merging this change. 

Could you also review and deploy https://github.com/Automattic/theme-yuga/pull/8 as well? Thanks in advance!

PR #6766, opened 2022-12-06
Add "Course" Theme

mikeyarce:
#### Changes proposed in this Pull Request:

Add the "Course" theme that was designed to pair with Sensei LMS.

#### Notes ####
Our plan is to get this theme on dotcom first, but not enabled for everyone just yet.  We will then work on adding a showcase, demo content, and getting Headstart annotations in place. 

Please let us know if we need to make any changes here and we'd be happy to do so!  🙇🏽

PR #6888, opened 2023-03-01
Add Course theme

donnapep:
This theme is meant to be used with [Sensei LMS](https://senseilms.com/), but can also be used standalone. We are moving the code from the https://github.com/Automattic/course repo to here so that theme updates can be deployed to both WordPress.org and WordPress.com.

Replaces https://github.com/Automattic/themes/pull/6766. The original feedback left there should be resolved in this PR.


course/course.pot
```diff
@@ -0,0 +1,390 @@
+# Copyright (C) 2023 Automattic
```
jffng: Typically we do not commit the .pot files here, they just live on wpcom. Is there a reason to include them here?

course/style.css
```diff
@@ -0,0 +1,429 @@
...
+	-webkit-font-smoothing: antialiased;
+}
+
+/*
+ * Responsive menu container padding.
+ * This ensures the responsive container inherits the same
+ * spacing defined above. This behavior may be built into
+ * the Block Editor in the future.
+ */
+.wp-block-navigation__responsive-container.is-menu-open {
```
jffng: This CSS can be removed now I think.

course/style.css
```diff
@@ -0,0 +1,429 @@
...
+ */
+a {
+	text-decoration-thickness: .0625em !important;
+	text-underline-offset: .15em
+}
+
+/*
+ * Search styles
+ */
+.wp-block-search__button {
```
jffng: What's the reason for setting this in CSS? 

This is what the default search button looks like when I insert a search block:
 
Screenshot 2023-03-03 at 11 27 14 AM

course/patterns/comments.php
```diff
@@ -0,0 +1,49 @@
...
+
+<!-- wp:comments-pagination -->
+<!-- wp:comments-pagination-previous /-->
+
+<!-- wp:comments-pagination-numbers /-->
+
+<!-- wp:comments-pagination-next /-->
+<!-- /wp:comments-pagination -->
+
+<!-- wp:post-comments-form /--></div>
```
jffng, line 48: Noticing an issue with the submit button for the comment form: 

Screenshot 2023-03-03 at 11 34 23 AM

jffng (commented):
This is a nice theme! I left a few comments of some issues I noticed, but generally looking good.

course/course.pot
donnapep: I've removed it in https://github.com/Automattic/themes/pull/6888/commits/23dde8b383be94b40683307187bf5ec567a4908f.

course/style.css
donnapep: Yup! Tested and removed in https://github.com/Automattic/themes/pull/6888/commits/1ceea814319056ae138f94ba97d0a0750808e534.

course/style.css
donnapep: I wasn't able to reproduce that issue, but when I removed the CSS it didn't appear to break anything either - [1bacf4a](https://github.com/Automattic/themes/pull/6888/commits/1bacf4a63262fc9e7f5d3d72a551886b5609284a).

course/patterns/comments.php
donnapep: @jffng Hmmm...I can't reproduce this one either when I tried on a Jurassic Ninja site:

Screenshot 2023-03-06 at 8 38 25 AM

donnapep:
@jffng Any update on this one? We have some more changes ready for this theme that are blocked by this PR.

course/patterns/footer.php
```diff
@@ -0,0 +1,173 @@
...
+      <!-- /wp:group -->
+    </div>
+    <!-- /wp:group -->
+    <!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"0"}}},"layout":{"type":"constrained","contentSize":""}} -->
+    <div class="wp-block-group alignfull" style="padding-top:0">
+      <!-- wp:paragraph {"align":"center","style":{"typography":{"letterSpacing":"0.02em"},"spacing":{"padding":{"bottom":"5rem","top":"var:preset|spacing|50"},"margin":{"top":"var:preset|spacing|70"}}},"fontSize":"x-small","fontFamily":"system"} -->
+      <p class="has-text-align-center has-system-font-family has-x-small-font-size" style="margin-top:var(--wp--preset--spacing--70);padding-top:var(--wp--preset--spacing--50);padding-bottom:5rem;letter-spacing:0.02em">
+        <?php
+        // Supports Footer Credit cusotmization on WPCOM and WoA sites.
+        $is_atomic = ( defined( 'ATOMIC_SITE_ID' ) && ATOMIC_SITE_ID );
```
jffng: We try to avoid wpcom specific customization like this in themes, is there a way to remove this? I believe on simple sites we don't do anything like this and they are still able to show a custom footer.

jffng (commented):
Hi @donnapep apologies for the delay. I think I'm missing some context. Will the theme be available on the wpcom themes showcase as a free theme also? 

> We are moving the code from the https://github.com/Automattic/course repo to here so that theme updates can be deployed to both WordPress.org and WordPress.com.

It looks like the theme is already in `pub`, to confirm there is no code there that will conflict with this version? 

Thanks for bearing with me, don't mean to gatekeep, just want to make sure we're in the right spot :)

course/patterns/footer.php
donnapep: > We try to avoid wpcom specific customization like this in themes, is there a way to remove this?

@mikeyarce Would you mind chiming in on this as I believe you added the footer credit?

Also, while I have you here, any idea about the answers to these other questions? 🙂 

> Will the theme be available on the wpcom themes showcase as a free theme also?

> It looks like the theme is already in pub, to confirm there is no code there that will conflict with this version?

mikeyarce: The footer credit was customized because of how the WordPress.com Footer works - there is a specific regex in place that will not replace the custom footer we have on Sensei, which is why I opted to have a customization there.

If we want to remove our custom Sensei footer, then we don't need the specific wpcom customization.

> Will the theme be available on the wpcom themes showcase as a free theme also?

I'm not sure - I think it's up to the themes team what they want to do with it?  I would say it's probably best to make it a free theme.

>  It looks like the theme is already in pub, to confirm there is no code there that will conflict with this version?

The theme we commited to `pub` is a bit older, the only changes we would need to keep from the already committed one are the Headstart Annotations.

jffng: > If we want to remove our custom Sensei footer, then we don't need the specific wpcom customization.

👍  That would be best to remove, then I think this is good to go.

Thanks!

jffng (commented):
There are a few more pieces that should be changed before this can be launched on dotcom.

The home page template does not match the screenshot. We have been using `home.html` to achieve this consistency. It's worth noting that introducing this template will potentially change / break the home page for existing users of this theme.

image

Line-height is off in Large and Extra-large sizes.

image

The gallery block captions line-height should be corrected, and the type treatment makes it difficult to read.

image

course/patterns/footer.php
donnapep: @mikeyarce Is this something you'd prefer to handle? Feel free to commit directly to this PR.

mikeyarce: Done in https://github.com/Automattic/themes/pull/6888/commits/e73455c1621677cb4b2a6e3fab3acde606c46984 😄

course/theme.json
```diff
@@ -0,0 +1,923 @@
...
+			"blockGap": "2.5rem",
+			"padding": {
+				"right": "1.25rem",
+				"left": "1.25rem"
+			}
+		},
+		"typography": {
+			"fontFamily": "var(--wp--preset--font-family--body)",
+			"fontSize": "var(--wp--custom--typography--font-sizes--normal)",
+			"lineHeight": "1.838rem"
```
jffng: Could we use a unitless value here instead? Using rem here doesn't work well for the larger font sizes.

course/theme.json
```diff
@@ -0,0 +1,923 @@
...
+					"style": "solid",
+					"width": "1px"
+				}
+			},
+			"core/post-content": {
+				"spacing": {
+					"blockGap": "20px"
+				},
+				"typography": {
+					"lineHeight": "clamp(1.838rem, 1.763rem + 0.249vw, 1.95rem)",
```
jffng: Could we use a unitless value here instead? Using rem here doesn't work well for the larger font sizes.

course/style.css
```diff
@@ -0,0 +1,412 @@
...
+	border: 1px solid;
+	word-break: normal;
+}
+
+.wp-block-table figcaption {
+	font-size: var(--wp--preset--font-size--small);
+	text-align: center;
+}
+
+ /* Background and text color, padding around for navbar, only for mobile view. Desktop view has separate color */
```
jffng, line 48: There is a lot of navigation specific CSS, do you think this kind of customization should be available via the UI? If so, could we suggest issues in Gutenberg to improve or fix the issues?

course/theme.json
donnapep: I removed this in https://github.com/Automattic/themes/pull/6888/commits/2217dfebb24a0e9b8d1fdd089f66819cc78e7245.

course/theme.json
donnapep: This is gone too in https://github.com/Automattic/themes/pull/6888/commits/2217dfebb24a0e9b8d1fdd089f66819cc78e7245 as it's not needed.

donnapep:
> The home page template does not match the screenshot.

I don't think we're too keen on changing the screenshot to match the home page. While it's true that the Course theme can be used without Sensei, we want to do everything we can to promote its use with Sensei. We consider it the primary use case.

course/style.css
donnapep: Yes, perhaps. I didn't work on the navigation bit so it's unclear to me if some or all of this could be moved into `theme.json`. My preference would be to do any refactoring outside of this particular PR though.

donnapep:
@jffng Thanks for the review! I believe I've addressed your feedback or else responded with comments to some of your other feedback. Please let me know what you think. 🙂

m1r0:
This has become a blocker for us so I'm approving.

Any additional feedback from the Themes team is very welcome and we can address that in a separate PR. 🙏

m1r0 (approved):

jffng:
@donnapep @m1r0 apologies for the delay, yes please go ahead and merge if it meets your requirements. Docs for deploying the theme to wpcom can be found here: PNEWy-eqL-p2

As a follow up, we can also add the theme to the list of themes to automated deploy for the dotorg repo here: https://github.com/Automattic/themes/blob/trunk/deploy-dotorg.sh#L18

I also want to clarify in the future, please don't wait on @Automattic/theam to approve or merge your PRs, especially if it's become a blocker. We're happy to provide feedback and perspective. We're stewards of this repo but cannot be DRIs or approve every PR for all the themes here, since in many cases we do not have all the context and specific requirements.

PR #6890, opened 2023-03-02
Bibimbap: fix navigation submenu background

### LINKED ISSUE #6889: Bibimbap Navigation Drop Down style editor mismatch
The header navigation menu appears to display differently for the home template and the other regular templates. 

I compared the difference between header.html and hero-header.php  and the home template header is missing the overlayBackgroundColor attribute. 

There's a few issues with this;
1. The dropdown menu item text remains white on the home template  while visible on the other templates 
2. The mobile menu's are different colors for home template  and the other templates 
3. If you add in the overlayBackgroundColor attribute, it works, but then you can't see the root level text items 
4. I also tested creating a page with a simple navigation, and applying global styles. There appears to be a mismatch between the editor  and the frontend 

I'm not sure if I'm missing something in the theme.json or if there is an issue within GB or core in the navigation block. 

I'm running WP 6.2-beta4-55454 and GB 15.2.3

Related Slack Thread p1677699943036359-slack-C029FM1EH

madhusudhand:

**Home page:** (Before & After)

image image

**Other pages:** (Before & After)

image image

**Small devices:** (Before & After)

image image

#### Related issue(s):

Fixes: #6889

### DIFF
--- a/bibimbap/parts/header.html
+++ b/bibimbap/parts/header.html
@@ -8,6 +8,6 @@
 <!-- /wp:group --></div>
 <!-- /wp:group -->
 
-<!-- wp:navigation {"overlayBackgroundColor":"secondary","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /--></div>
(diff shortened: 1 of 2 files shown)


jffng (approved):
Good catch and fix @jeffikus @madhusudhand, overlay now works as expected in my testing.

PR #6902, opened 2023-03-13
Paimio: add theme

jffng:

Adds new theme designed by @beafialho 

image

#### To-do:

- [ ] `There’s an issue with the navigation that I can’t seem to change in the editor, in smaller viewport sizes.`


madhusudhand (commented):
Reviewed and addressed the following issues.

* updated min req version to 6.1 because it uses fluid typography
* included Font credits in readme
* updated patterns and templates such as removal of refs, translations, duplicate content etc..
* removed unused images
* There are conflicting definitions for button radius. removed `0` and now the buttons looks the following
image
@beafialho please let me know if this is not right.
Also, the solid button hover state isn't different from its original state. Should we make the style as outline button on hover?
image

beafialho:
Thanks @madhusudhand and @jffng for adding this theme. There are a few things that need to be done:

- Rename the theme to "Paimio"
- All button radius should be `0`
- Replace images of the posts with the two images below: see the [demo site](https://aalto.mystagingwebsite.com) for reference

chair
trolley

- Replace the theme screenshot:

paimio_screenshot


- Use this showcase image instead of the one shared above (I'm still not sure we should use the new mockup):

paimio_showcase

madhusudhand:
@beafialho renamed the theme and addressed the suggestions. It looks wonderful 🪑

madhusudhand (approved):
checks out 🚢

PR #6905, opened 2023-03-13
Kiosko: add theme

jffng:

Adds new theme by @iamtakashi 

Screenshot 2023-03-13 at 2 22 59 PM

`I built this theme as a part of a larger theme, but it has been decided to scope down, and the need should be tackled differently. Therefore, I’m now releasing this theme without additional pages, patterns, and style variations.`

`I built the front page with some Woo blocks, but as you may know, other store pages still need to rely on the legacy Woo templates, so you will see various inconsistencies in style treatment.`


pbking:
I believe this is ready to be tested on wpcom.

However since these templates have Woo blocks I believe that will prevent it from being sent to the WordPress.org collection right?  Which is OK, there are other themes in this repository that aren't on Wordpress.org.

But since Woo isn't available on simple sites, might this be a premium theme instead?

kiosko/templates/404.html
```diff
@@ -1,4 +1,4 @@
-<!-- wp:template-part {"slug":"header","tagName":"header","theme":"kiosko"} /-->
```
madhusudhand: This is coming from Create Block Theme plugin. I don't remember why the plugin adding this. 
I think it is a fix for the plugin?

CC: @matiasbenedetto

matiasbenedetto: I'm not sure what you mean @madhusudhand.

madhusudhand: Sorry, I am referring to this [commit](https://github.com/Automattic/themes/pull/6905/commits/47bfbf8f6ff76fb31a7c54adf85b1e0820982a0d)

image

vcanales:
> @pbking 
> However since these templates have Woo blocks I believe that will prevent it from being sent to the WordPress.org collection right? Which is OK, there are other themes in this repository that aren't on Wordpress.org.

@pbking just to get a full grasp of why we couldn't send it to WP.org, is this the [guideline](https://make.wordpress.org/themes/handbook/review/required/#6-plugins) we would be breaching?

> ##### https://make.wordpress.org/themes/handbook/review/required/#6-plugins
> Themes may only recommend plugins that are hosted on WordPress.org.
> Plugins may only be installed by user action, not automatically.
>
> Do not:
> 
> Include zip files or plugins in the theme folder, or download plugins automatically.
> Include plugin functionality.

> @pbking
> But since Woo isn't available on simple sites, might this be a premium theme instead?

@iamtakashi, do you have thoughts on this theme being Premium instead?

iamtakashi:
> do you have thoughts on this theme being Premium instead?

I mentioned this in Slack but I don't have a strong opinion. Feel free to launch it on WP.com only as a premium theme. I think we still make a repo for each one of our premium themes on WP.com. For example, this repo is for the Yuga theme. https://github.com/automattic/theme-yuga

iamtakashi:
Any movement here? It's been a month :)

jeffikus:
@iamtakashi I did some testing on this and compared it to the Hey theme. I think we have 2 "types" of Woo Themes that we're building;

1 - Free themes for dotcom and dotorg that meet the dotorg requirements and have template support for Woo. This would be a theme like the Hey theme.
2 - Themes like Kiosko that are intended to function best with an additional plugin. 

I wonder if the approach we are taking with Kiosko is the best approach. When we built Storefront when I was at Woo, and Canvas to a lesser degree, we gracefully degraded those themes using conditional statements to load particular parts of the theme. Now, that was done with actions for the most part, and they are classic themes, but we can conditionally load parts of a pattern in PHP in a block theme. 

My questions are;

1 - If we are building a commerce focused theme, should we be providing a fallback homepage for folks that don't want to use it as a store theme?
2 - Could we do that through the patterns conditionally?
3 - Would this impact on our workflow building through Create Block Theme?

I would love to see themes like Kiosko in the free dotorg directory. Looking at the guidelines I don't see a problem submitting it, but I'm open to correction on that too :)

pbking:
If people want to use it as something than a store front theme then all they really want is the Global Styles and a smattering of CSS.  Otherwise the templates aren't valuable unless there's conditional logic somewhere.  The concept we're working towards is that a user can pull the things they like from different themes.  That's the part a user would pick if they don't need a storefront.  Maybe we lean on that expectation rather than introducing complexity to a theme.

If the styles are truly what we want to over to users on wpcom simple sites then we should just copy the styles and use different templates.  Now we have two themes.

jeffikus:
@pbking thanks for the feedback, I agree, I think that's what would attract a user to the theme, the styles - but, the patterns help them get to the end result sooner. Do you think we should not be providing a fallback homepage?

pbking:
I think it could be helpful to a user with products to have multiple homepage options including some without those products.  So I do think that homepage patterns would be good to include.  But AFAIC that's a design decision more so than a technical one so I don't have much opinion there.  From a default template standpoint I don't think we should be including a technical solution (such as logic in patterns) to handle an environment sans-woo.

mikachan:
I agree with @pbking.

If we need the conditional logic to make it easier for some users, maybe plugins could choose to handle the default template themselves by influencing the template hierarchy. For example, in the homepage example for this theme: the default `front-page.html` template would not include any Woo blocks (i.e. the fallback homepage), and there would be another template named something like `products-front-page.html` for the commerce-specific layout. If Woo is active, then `products-front-page.html` is used over `front-page.html`, else `front-page.html` is used.

Maybe there's a better way to handle this than via the template name (something like additional meta data?), but I think it would be best if the logic sat with the plugin.

jeffikus:
@mikachan I'm not sure I understand what you mean by the logic should be in the plugin? Here's my understanding - open to correction here - but the theme holds templates and patterns, patterns can contain blocks which can be provided from many sources. Those blocks are rendered by the templates and patterns in both the editor and frontend, so the theme is controlling the presentation options. Why would the plugin decide what is output conditionally? Shouldn't the theme decide what structural parts are rendered? 

I do like your idea of having a standard for plugins to override though by template name. That's in line with how the WP template hierarchy works.

mikachan:
> Why would the plugin decide what is output conditionally? Shouldn't the theme decide what structural parts are rendered?

Yes, the theme decides which blocks are included, or rather, the theme decides which templates and patterns are included, and these may be designed to work with blocks provided by plugins. The conditional logic I was referring to was overriding a default template with a plugin-specific version. So it's still the theme that decides to include this plugin-specific template, but the conditional logic to render the template sits in the plugin, not the theme.

> That's in line with how the WP template hierarchy works.

Yes!

iamtakashi:
When I built this theme, there was a brief like "make some store themes". But I sense that's not aligned with the current vision (any theme should be able to be a store theme), and the fact that Kiosko having `front-page.html` including patterns required WC is one of the problems.

I'm happy to change this theme to be more like Hey theme, a "normal" theme with some additional templates for stores.

How does that sound?

vcanales:
> I'm happy to change this theme to be more like Hey theme, a "normal" theme with some additional templates for stores.

As a fan of Hey, that sounds great to me.

iamtakashi:
> How does that sound?

Having said that... shall we pause this theme for a bit? I don't think I describe the direction well, and I want to hear what would be the best thing from @mtias when he can.

jeffikus:
@iamtakashi any update on this?

iamtakashi:
> any update on this?

Darren is working on a post summarising the call with Matias, and I'm hoping the post will shed light on the direction for themes with Woo.

iamtakashi:
Let's "archive" this theme. I feel the leadership wants Woo folks to create themes with Woo. I don't mind it at all and, it's likely the implementation I've done in this theme isn't the best anymore.

alaczek:
After a recent discussion I'm re-opening this PR. This theme is still valuable on WordPress.com, and since it's finished it doesn't make sense to shelve it, even if it never ends up on .org (see this discussion: p1712729561809789-slack-C06H9KE62DS)

I will be making some minor updates since a few things changed.

PR #6906, opened 2023-03-13
Gramming: add theme

jffng:

Adds new theme designed by @henriqueiamarino

Screenshot 2023-03-13 at 2 10 05 PM

#### Related issue(s):

- [x] Posts on this demo have no title, just an image and a paragraph (content). I did that as I was trying to simplify posting via mobile app. Do we have a problem with that?
- [x] It would be preferable to set Aspect Ratio 1:1 on Post Featured images placed inside Query loops but I couldn’t do that. I know there’s a PR about that and it will reach the Editor anytime soon but if possible through a dev hack, I’m in favor.


pbking:
This theme relies on post content to demonstrate the index; otherwise it's quite blank.  (Can we give this kind of theme a name?  Content-dependent?)

Attaching here: 
[jacobgola.WordPress.2023-03-23.xml.zip](https://github.com/Automattic/themes/files/11051243/jacobgola.WordPress.2023-03-23.xml.zip)

pbking:
> It would be preferable to set Aspect Ratio 1:1 on Post Featured images placed inside Query loops but I couldn’t do that. I know there’s a PR about that and it will reach the Editor anytime soon but if possible through a dev hack, I’m in favor.

I was able to achieve this without a hack, just setting the height use 30 VW units:

image

It's not exact but close enough.

"If it looks right it is right."  Right?

It does change how the mobile layout works however:

Before:

image

After:

image


Is that better? Acceptable?

pbking:
I made a change to the header to use rows instead of columns.  The social bit was wrapping strangely at certain widths and any amount of social icons other than 3 was laying out strangely.  Also those icons weren't right aligned like they were in the footer:

Before:

image

image

image

After:

image

image

image

pbking:
I believe this theme is ready to go, but would like to get another's eyes on it to make sure.

gramming/parts/header.html
```diff
@@ -0,0 +1,23 @@
...
+	<div class="wp-block-group alignwide"><!-- wp:group {"align":"full","layout":{"type":"flex","flexWrap":"wrap"}} -->
+	<div class="wp-block-group alignfull"><!-- wp:site-logo {"width":80,"shouldSyncIcon":true,"className":"is-style-rounded"} /-->
+	
+	<!-- wp:group {"style":{"spacing":{"padding":{"top":"0px","right":"0px","bottom":"0px","left":"0px"},"margin":{"top":"0px","bottom":"0px"},"blockGap":"0.5rem"}},"layout":{"type":"flex","orientation":"vertical"}} -->
+	<div class="wp-block-group" style="margin-top:0px;margin-bottom:0px;padding-top:0px;padding-right:0px;padding-bottom:0px;padding-left:0px"><!-- wp:site-title {"align":"wide","style":{"spacing":{"blockGap":"0rem","padding":{"top":"0px","right":"0px","bottom":"0px","left":"0px"},"margin":{"top":"0px","bottom":"0px","right":"0px","left":"0px"}}}} /-->
+	
+	<!-- wp:paragraph {"fontSize":"small"} -->
+	<p class="has-small-font-size">Hello! I’m Jacob. I have two passions: Food and photography<br><mark style="background-color:rgba(0, 0, 0, 0)" class="has-inline-color has-foreground-color">jacobgola.com</mark></p>
```
mikachan: I think this text should probably be translatable.

gramming/readme.txt
```diff
@@ -0,0 +1,31 @@
...
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 2 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU General Public License for more details.
```
mikachan, line 31: We should add the credits for the `man-behind-the-camera-scaled.jpg` image, and possibly the images used in the screenshot too.

gramming/style.css
```diff
@@ -0,0 +1,66 @@
+/*
+Theme Name: Gramming
+Theme URI: https://github.com/wordpress/gramming/
+Author: Themes Design
+Author URI: https://wordpress.org/themes/
```
mikachan:
```suggestion
Theme URI: https://wordpress.com/theme/gramming
Author: Automattic
Author URI: https://automattic.com
```

gramming/style.css
```diff
@@ -0,0 +1,66 @@
...
+Description: Bring the Social Media experience to your site and transform it into a repository for your photos and videos. Get Gramming now.
+Requires at least: 5.8
+Tested up to: 5.9
```
mikachan:
```suggestion
Tested up to: 6.1
```

gramming/readme.txt
```diff
@@ -0,0 +1,31 @@
+=== Gramming ===
+Contributors: Themes Design
```
mikachan: I think this should be "Automattic" rather than "Themes Design".

```suggestion
Contributors: Automattic
```

gramming/readme.txt
```diff
@@ -0,0 +1,31 @@
...
+Requires at least: 5.8
+Tested up to: 5.9
```
mikachan, line 4: This can be changed to 6.1.

```suggestion
Tested up to: 6.1
```

gramming/style.css
```diff
@@ -0,0 +1,66 @@
...
+.wp-block-table th {
+	padding: var(--wp--preset--spacing--30);
+	border: 1px solid;
+	word-break: normal;
+}
+.wp-block-table figcaption {
+	font-size: var(--wp--preset--font-size--small);
+	text-align: center;
+}
+
```
mikachan:
```suggestion
```

We don't need these table styles anymore.

gramming/templates/archive.html
```diff
@@ -0,0 +1,55 @@
...
+<!-- wp:query-pagination {"paginationArrow":"arrow","align":"wide","style":{"typography":{"fontSize":"2rem","letterSpacing":"0px"}},"layout":{"type":"flex","justifyContent":"space-between"}} -->
+<!-- wp:query-pagination-previous {"label":"Newer Posts"} /-->
+
+<!-- wp:query-pagination-next {"label":"Older Posts"} /-->
+<!-- /wp:query-pagination -->
+
+<!-- wp:query-no-results -->
+<!-- wp:group {"align":"wide","style":{"spacing":{"blockGap":"1rem"}},"layout":{"type":"flex","orientation":"vertical"}} -->
+<div class="wp-block-group alignwide"><!-- wp:heading {"level":1,"align":"wide","style":{"typography":{"fontSize":"2rem","letterSpacing":"0px"}}} -->
+<h1 class="wp-block-heading alignwide" style="font-size:2rem;letter-spacing:0px"> Sorry, but nothing matched your search terms.</h1>
```
mikachan: I think the text in this file should be translatable, too.

gramming/templates/search.html
```diff
@@ -0,0 +1,55 @@
...
+<!-- wp:query-pagination {"paginationArrow":"arrow","align":"wide","style":{"typography":{"fontSize":"2rem","letterSpacing":"0px"}},"layout":{"type":"flex","justifyContent":"space-between"}} -->
+<!-- wp:query-pagination-previous {"label":"Newer Posts"} /-->
+
+<!-- wp:query-pagination-next {"label":"Older Posts"} /-->
+<!-- /wp:query-pagination -->
+
+<!-- wp:query-no-results -->
+<!-- wp:group {"align":"wide","style":{"spacing":{"blockGap":"1rem"}},"layout":{"type":"flex","orientation":"vertical"}} -->
+<div class="wp-block-group alignwide"><!-- wp:heading {"level":1,"align":"wide","style":{"typography":{"fontSize":"2rem","letterSpacing":"0px"}}} -->
+<h1 class="wp-block-heading alignwide" style="font-size:2rem;letter-spacing:0px"> Sorry, but nothing matched your search terms.</h1>
```
mikachan: Same issue here with translatable text. Maybe the no-results section could be a pattern?

gramming/templates/index.html
```diff
@@ -0,0 +1,11 @@
+<!-- wp:template-part {"slug":"header","area":"header"} /-->
+
+<!-- wp:group {"layout":{"type":"constrained"}} -->
+<div class="wp-block-group"><!-- wp:query {"queryId":8,"query":{"perPage":"12","pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":false,"taxQuery":{"category":[5]}},"displayLayout":{"type":"flex","columns":3},"align":"full","className":".add-remove-space"} -->
```
mikachan: Can we remove this class: `.add-remove-space`? I can't find the corresponding CSS.

mikachan: We shouldn't use a specific category here, in case it doesn't exist:

```suggestion
<div class="wp-block-group"><!-- wp:query {"queryId":8,"query":{"perPage":"12","pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":false,"display layout":{"type":"flex","columns":3},"align":"full","className":".add-remove-space"} -->
```

mikachan (commented):
I've left some inline comments, but otherwise this is testing well for me!

gramming/parts/header.html
pbking: Rather than translate the text I instead made it the site tagline.

gramming/templates/index.html
pbking: Indeed.  Removed.

I checked the source theme (block canvas) and didn't see it there so I'm not sure where it came from.

gramming/templates/search.html
pbking: Good call.  Moved to a pattern and used in both pages.

gramming/templates/archive.html
pbking: patternized

pbking:
Thank you for your help reviewing this @mikachan!

I've adjusted everything according to your notes with the exception of the credits which I'll ask for outside of this PR and see that it gets added before shipping.

Going to bring this into trunk so I can do some more evaluation on wpcom.

PR #6908, opened 2023-03-13
Lois: add theme

jffng:

Adds new theme designed by @henriqueiamarino

image

#### Related issue(s):

`As a special theme in deference to a figure who passed away last November Pablo wanted it to be released at a fast pace. So we would not only pay our respects but also show how updated we are to the zeitgeist of this creative era that we belong.`


madhusudhand (commented):
Thanks for working on the theme @henriqueiamarino
I have added couple of commit to fix the issues.

It looks good now.

Can you provide a screenshot for the theme?

henriqueiamarino:
Both screenshots here @madhusudhand 

lois_screenshot

lois_showcase-screenshot

PR #6909, opened 2023-03-13
Ueno: add theme

jffng:

Adds new theme designed by @beafialho 

Screenshot 2023-03-13 at 2 03 48 PM

#### To-do:

- [x] The default style’s typography is meant to be uppercase everywhere but the two variations are meant to be regular case. I defined it but after saving the variations, the text still showed up uppercase in all of them.
- [x] Duotone in images. Ueno Dark is supposed to use a duotone effect in all images, which also didn’t apply:


beafialho:
Sorry about the extra detail, but is it possible to set this color on **links hover** for the Dark Variation?

ueno links

ueno/styles/ueno-black.json
```diff
@@ -0,0 +1,652 @@
+{
+	"settings": {
+		"border": {
```
jffng, line 3: I think `appearanceTools` is missing, I wonder why this key keeps getting stripped from the export.

ueno/styles/ueno-black.json
```diff
@@ -0,0 +1,652 @@
...
+				},
+				{
+					"color": "#1b1818",
+					"name": "Tertiary",
+					"slug": "tertiary"
+				}
+			],
+			"duotone": [
+				{
+					"slug": "foreground-to-background",
```
jffng: I think should avoid the use of `foreground` and `background` as it doesn't make sense in a lot of contexts. What if we just called it `default` or `default-filter` like [Skatepark](https://github.com/Automattic/themes/blob/trunk/skatepark/theme.json#L38) does?

ueno/styles/ueno-black.json
vcanales: Added to the main theme.json file in https://github.com/Automattic/themes/pull/6909/commits/b1b5aa2794892fc0772b006ce0b9bd4597085a3b

vcanales:
@beafialho, getting back to your comments, they should be resolved on my end.

> - The default style’s typography is meant to be uppercase everywhere but the two variations are meant to be regular case. I defined it but after saving the variations, the text still showed up uppercase in all of them.

https://user-images.githubusercontent.com/1157901/232562664-d2119749-fbe8-43b6-8a8f-069efa38a22a.mov


> - Duotone in images. Ueno Dark is supposed to use a duotone effect in all images, which also didn’t apply.

I found that it does apply, but it isn't immediately reflected in the editor when switching styles; you have to navigate away from it to see it there. It is applied in the front end correctly. I will file a Gutenberg bug with this issue, but it shouldn't keep the theme from being shipped.

https://user-images.githubusercontent.com/1157901/232638577-6423fa0d-1671-4776-a35f-fd3efead2507.mp4

beafialho:
Cool, thanks a lot!

mikachan:
> I found that it does apply, but it isn't immediately reflected in the editor when switching styles; you have to navigate away from it to see it there.

I _think_ this may be covered by this issue: https://github.com/WordPress/gutenberg/issues/46397

vcanales (approved):
LTGM.

iamtakashi:
This has been available in WP.org.

PR #6910, opened 2023-03-13
Vetro: use latest duotone features

jffng:

This PR applies duotone via global styles instead of directly in the template. It also restores the duotone effects to its original intentions with each style variation. 

#### To test:

- Activate Vetro
- Switch between style variations (I noticed there is a delay for the duotone to apply, seems like the filter is cached for about a minute)
- Verify the images appear as expected with the correct duotone filters

cc @henriqueiamarino


vetro/theme.json
```diff
@@ -357,5 +361,5 @@
 		}
 	],
 	"version": 2,
-	"$schema": "https://schemas.wp.org/wp/6.1/theme.json"
-}
\ No newline at end of file
+	"$schema": "https://schemas.wp.org/trunk/theme.json"
```
ajlende: It's better to enter the minimum supported WordPress version for the theme instead of trunk here. This way, if something is added to the theme.json that doesn't work in that minimum version, the editor will flag it, or you could add an automated check with GitHub actions or pre-push or something like that.

```suggestion
	"$schema": "https://schemas.wp.org/wp/5.8/theme.json"
```

jffng: Oh good call, thanks!

jffng: Hmm on second thought — we approach the minimum version for block themes a bit differently. The approach is that the theme will still work with the minimum version, but some features may not be available.

What would you recommend in this case?

ajlende: I'd say setting the minimum version of the features that you're using is probably best in that case. Then if someone wants to add a new feature, at least you know it's another feature that won't be supported.

If theme.json ever has any breaking changes, the `version` property will need to be updated, so theme.json files with `"version": 2` can continue to work just fine. So if you're using the schema from trunk and that happens, you may suddenly have a bunch of warnings appear on the version 2 schemas since trunk has moved to version 3.

However, trunk is what you need to use if you are developing a theme using the features that haven't yet made it into a WordPress release.

The schema is just a tool that provides autocompletion and can check for correctness, so I guess it ultimately depends on how you want to use that tool.

jffng: Thanks @ajlende ! 

> I'd say setting the minimum version of the features that you're using is probably best in that case.

Makes sense, took your recommendation in 9b4599b371801755149a15544e05b815983e19e9.

mikachan (approved):
Looks good and tests well for me! I tried each variation and all the images look as expected ✨

PR #6930, opened 2023-03-21
Paimio: fix dotCom issues

### LINKED ISSUE #6929: Paimio: dotCom issues
While the theme works as expected in local env, it has the following differences on dotCom

* fontWeight is higher for paragraphs
* search buttons have default border and outline

| Local | dotCom |
| --- | --- |
| image | image |
| image | image |

madhusudhand:
This issue fixes following

* Add empty `index.php` file
* set default fontWeight to 300

Issue with search seems to be fixed in dotCom

Fixes: #6929

### DIFF
--- a/paimio/index.php
+++ b/paimio/index.php
@@ -1,0 +1,2 @@
+<?php
+	# This page intentionally left blank
--- a/paimio/index.php
+++ b/paimio/index.php
@@ -1,2 +1,0 @@
-<?php
-	# This page intentionally left blank
--- a/paimio/theme.json
+++ b/paimio/theme.json
@@ -552,6 +552,7 @@
 		"typography": {
 			"fontFamily": "var:preset|font-family|newsreader",
 			"fontSize": "var(--wp--preset--font-size--medium)",
+			"fontWeight": "300",
 			"lineHeight": "1.6"
 		}
 	},


paimio/index.php
mikachan, line 1: Why do we need a blank index.php for this theme?

madhusudhand: Seems like there is an issue in core validating parent theme in case of multi site setup. 
Adding an empty index.php does help as a workaround. Here is the issue for reference: https://wordpress.org/support/topic/child-theme-in-multisite-shows-error/?view=all

Also, theme check plugin marks `index.php` as required

mikachan: Ahh OK, as you mentioned in that forum thread, it's best if that's fixed in Core. Is there already an open trac ticket for that?

The Theme Check plugin shouldn't be marking index.php as required, the check should've been removed as part of https://github.com/WordPress/theme-check/pull/425. Which version of the plugin are you using? I'm using version 20220929, and Paimio has no errors/warnings for me.

madhusudhand: There isn't a trac ticket yet. Let me open one. 
Updated themecheck and removed `index.php` 👍

mikachan (approved):
LGTM!

PR #6953, opened 2023-03-28
Iotix: add global padding and fix footer

### LINKED ISSUE #6946: iotix: style issues
Found following issues in iotix.

Footer needs a closing tag to avoid block asking for recovery.
image

It requires Root padding
image

Archive page shows single column, but it is expected to have two columns.
image

Search block when inserted in a post vs search template
image
image

CC: @jffng @beafialho

jffng:

This PR adds root padding and fixes a block validation error with the footer.

#### Related issue(s):

#6946

### DIFF
--- a/iotix/patterns/footer.php
+++ b/iotix/patterns/footer.php
@@ -64,6 +64,5 @@
         <!-- /wp:paragraph -->
         <!-- wp:navigation {"overlayMenu":"never","layout":{"type":"flex","orientation":"horizontal"},"style":{"spacing":{"blockGap":"var:preset|spacing|40"}}} /-->
     </div><!-- /wp:group -->
-
 </div><!-- /wp:group -->
 </div><!-- /wp:group -->
--- a/iotix/patterns/footer.php
+++ b/iotix/patterns/footer.php
@@ -8,64 +8,62 @@
 ?>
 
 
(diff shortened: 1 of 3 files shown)


mikachan (commented):
The footer looks good with this PR and I'm not seeing the block recovery error, however, on the homepage I'm not seeing any horizontal padding around the post content:

image

I don't think it's related to this PR, but I did think adding the global padding would fix it 🤔 Does it just need to be wrapped in a group?

jffng:
> Does it just need to be wrapped in a group?

This question plagues my existence 😅 

@beafialho Looking at the **home** template for this theme, right now it's a post template block — should we replace this with a pattern or actual blocks?

beafialho:
@jffng are you referring to the home template in [this demo site](https://startuptheme.mystagingwebsite.com)?

jffng:
@beafialho yep! I guess I am not familiar with the approach to make the home template just use a post content block, rather than putting the blocks you want directly into the template (as a pattern or just directly). But I could be out of date or missing some context.

beafialho:
Ah yes! Here's what I did: I created a page called Home, which uses the home template then I set this page as "Front Page" in the reading settings. That way I added and edit the content in the page, not in the template.

PR #6954, opened 2023-03-28
Remove table CSS from block canvas theme as it has been addressed in Gutenberg

pbking:
Fixed here: https://github.com/WordPress/gutenberg/pull/45069


mikachan (approved):
Yes! This was on my list to handle today. Thanks! 🌸

PR #6956, opened 2023-03-29
DotCom Automation: create GlotPress project when a theme is deployed

madhusudhand:

This PR adds automation script to create and deploy GlotPress project on dotcom.

command: `npm run prelaunch:glotpress pixl`

#### Related issue(s):

https://github.com/orgs/Automattic/projects/215#card-87878291

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,6 +28,7 @@
 				"octokit": "^2.0.3",
 				"open": "^8.4.0",
 				"parse5-html-rewriting-stream": "^7.0.0",
+				"semver": "^7.5.0",
 				"string-progressbar": "^1.0.4",
 				"table": "^6.8.0"
 			},
@@ -339,6 +340,15 @@
 				"url": "https://opencollective.com/babel"
 			}
 		},
+		"node_modules/@babel/core/node_modules/semver": {
+			"version": "6.3.0",
+			"resolved": "https://registry.npmjs.org/semver/-/semver-6.3.0.tgz",
+			"integrity": "sha512-b39TBaTSfV6yBrapU89p5fKekE2m/NwnDocOVruQFS1/veMgdzuPcnOM34M6CwxW8jH/lxEa5rBoDeUwu5HHTw==",
+			"dev": true,
+			"bin": {
+				"semver": "bin/semver.js"
+			}
+		},
 		"node_modules/@babel/eslint-parser": {
 			"version": "7.18.2",
 			"dev": true,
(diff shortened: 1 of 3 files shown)


mikachan (commented):
This is great as a way to automate this from the free themes repo. Maybe this new script could be run as part of the initial deployment (or the deploy script), so we don't need to remember to run it. We'd probably need to add a check for an existing GlotPress project beforehand.

What do you think about automating this within the dotcom codebase instead? I'm thinking that this would then cover any newly created themes on dotcom, not just themes stored in this repo.

madhusudhand:
> Maybe this new script could be run as part of the initial deployment (or the deploy script), so we don't need to remember to run it. We'd probably need to add a check for an existing GlotPress project beforehand.

I am trying this fun commanline approach to launch a theme here. https://github.com/Automattic/themes/pull/6959
Its a draft version and should have all the checklist included.

What do you think about this?

> What do you think about automating this within the dotcom codebase instead? I'm thinking that this would then cover any newly created themes on dotcom, not just themes stored in this repo.

Yes. Agree that it should help all the themes on dotcom.
Which repo do you suggest this should be part of?

mikachan:
> What do you think about this?

I love it! I like that it makes launching a theme fun, and certainly more efficient than our current process.

> Which repo do you suggest this should be part of?

I'm not sure, but when I was initially thinking of automating GlotPress, I thought it would live near the GlotPress theme script file, maybe being triggered by a theme being network-activated or the showcase being published. But the solution in this PR is fab and 10x times better than what we do currently, so I'm not sure anymore. I guess I was thinking that the responsibility for GlotPress should sit with dotcom rather than this repo, but maybe this introduces unnecessary time/complications.

Edit:
Edit to add more thoughts... I'm not sure GlotPress automation should be part of a CLI script. Rather it should happen automatically, hidden in the background of another process, e.g. on theme network activation or deployment. So this maybe could still be this repo for now, but it's triggered as part of the `deploy-theme` step.

pbking:
I agree with the "hidden and automated" part of @mikachan  comments.  Can we tap that process in to the deploy process?  I think it's fine to have a command that generates the glotpress bits, but I think it might be best if that command didn't also commit that change and that instead it just rolls that change into every push if it hadn't already been created (and maybe if the theme version is >= 1.0 or something).

madhusudhand:
Thanks @mikachan @pbking for the feedback.

GlotPress creation is now made part of push-button-deployment, for the themes that changed. 
It makes sense to create every time a theme is deployed (it might have included new patterns or strings that requires translations)
It does nothing if the glotpress already exists.

As suggested by @pbking, included a version check to create only when the theme reaches its first major version.

mikachan (approved):
Thanks for addressing the feedback!

> It makes sense to create every time a theme is deployed (it might have included new patterns or strings that requires translations)

This makes sense 👍 This sounds like an improvement on our current workflow.

I think we can bring this in, as it removes the manual step from the current process and gives us a good basis for our other automation work.

PR #6961, opened 2023-03-29
Removed forced content wrapper styles

pbking:
This removes [this fix](https://github.com/Automattic/themes/pull/4190) which was ultimately causing [this bug](https://github.com/Automattic/wp-calypso/issues/69859).

It returns the original issue of content not being "all the way at the bottom" when there isn't much content.  But that's a problem in all block themes.  

The problem it was causing was that the elements were forced to be in a 'flex' container when that wasn't expected.

Before:

image

image

After:

image

image

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -184,12 +184,6 @@ body > .is-root-container > .wp-block-template-part > .wp-block-cover,
   --wpadmin-bar--height: 46px;
 }
 
-.wp-site-blocks {
-  min-height: 100vh;
-  display: flex;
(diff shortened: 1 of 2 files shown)


scruffian:
I think that code was there to keep the "Powered by WordPress" text aligned to the bottom of the page, but if it's causing other issues I'm fine removing it.

jffng:
Works as advertised

Before | After
--- | ---
Screenshot 2023-03-31 at 9 43 27 AM | Screenshot 2023-03-31 at 9 42 59 AM

jffng (approved):

PR #6967, opened 2023-04-03
Lettre: fix display of date

mtias:
Theme was using `M n` which renders the month as a number, not the actual day.


madhusudhand (approved):
Thanks for the fix @mtias 
Pushed couple of commits to fix the same issue in other patterns of Letter and Lynx theme single post template.

Checks out 🚢

mtias:
Thanks! Can close #6968

PR #6969, opened 2023-04-03
Update new theme links to point to WPCOM instead of GitHub

wojtekn:

Update all /pub themes so the theme-uri points to https://wordpress.com/theme/<slug>

Previously, they were linking to GitHub.

#### Related issue(s):

https://github.com/Automattic/wp-calypso/issues/62954


mikachan (approved):
🚢

PR #6971, opened 2023-04-03
Hey: fix avatar radius

madhusudhand:

Fixes border radius of avatar via CSS in hey theme.

Before:
image

After:
image

### DIFF
--- a/hey/style.css
+++ b/hey/style.css
@@ -93,7 +93,8 @@ a {
  * Styles for Author blocks
  * https://github.com/WordPress/gutenberg/issues/24952
  */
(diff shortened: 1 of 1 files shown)


mikachan (approved):
LGTM  👍

Would be even better if we could do it in theme.json! Looks like the styles are applied to the `.wp-block-avatar` container rather than the avatar image element.

```
			"core/avatar": {
				"border": {
					"radius": "999px"
				},
				"typography": {
					"lineHeight": "0"
				}
			},
```

iamtakashi:
Thank you @madhusudhand!

Issue #6941, opened 2023-03-22
[Bug]: Blockbase child themes & incompatible style variations


dsas:
### Quick summary

Jackson's Charcoal style variation has white background text and Jackson has a white background. This isn't easy to read.

Ruby wine | Charcoal
-------|------
<img width="1125" alt="" src="https://user-images.githubusercontent.com/93301/227045623-cd87591a-a80a-484c-99a8-ecc310b7d446.png"> | <img width="1125" alt="" src="https://user-images.githubusercontent.com/93301/227045820-ff98a0e5-ed44-42e6-8e77-d741567b45cf.png">

There are several more examples of this and similar issues with other blockbase child themes - skivers has the same style variations and they just change a thin button outline.

The style variations don't belong to the individual themes but rather their parent theme - blockbase. The variations were [added](https://github.com/Automattic/themes/pull/6167/commits/4cbb43dc10fd419f5334e0d7a768204a3c9af1dc) and [named](https://github.com/Automattic/themes/pull/6167/commits/d75c385c81f5710c77bf1073dddd8215fe8806a4) in #6167 but only started becoming visible in child theme after WP6.2 was merged to wpcom (see fbhepr%2Skers%2Sjcpbz%2Sjc%2Qvapyhqrf%2Spynff%2Qjc%2Qgurzr%2Qwfba%2Qerfbyire.cuc%3Se%3Qn8028o5p%23685-og & D104877-code for the merge and https://github.com/WordPress/gutenberg/pull/46554 for the initial GB change)

### Steps to reproduce

1. Activate the Jackson theme
2. In the site editor use the GS sidebar to set the style variation to charcoal

### What you expected to happen

To be able to read the text

### What actually happened

Couldn't read the text

### Browser

Mozilla Firefox

### Context

p1680136513400279-slack-C048CUFRGFQ passSY-5Al-p2#comment-11903

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

Yes, easy to implement

### Workaround details

 * Don't use that style variation
 * Do use it and change the text colour

autumnfjeld:
FYI on this @Automattic/lego . Discussed here https://a8c.slack.com/archives/C029FM1EH/p1680709030792759

mmtr:
Solved in https://github.com/Automattic/themes/pull/6996 + D107265-code + D107818-code.

[...]

PR #6996, opened 2023-04-06
Blockbase: Remove inherited style variations from child themes

mmtr:

As of [Gutenberg 15.1](https://github.com/WordPress/gutenberg/pull/46554)/[WordPress 6.2](https://github.com/WordPress/wordpress-develop/pull/3901), child themes inherit the style variations of the parent theme.

This is causing issues to some Blockbase child themes like Jackson or Kingsley, because the headstart content for a child theme is only created for the default style variation. When a different style variation is applied, the styles from the headstart content are incompatible with the new variation:

https://user-images.githubusercontent.com/1233880/230398566-f5abd9a1-c558-4016-9734-201c430d81d3.mov

<br>

To mitigate that, this PR introduces a new `optOutOfParentStyleVariations` setting to the Global Styles, so child themes can opt out of inheriting style variations from the parent theme. The net setting has a `false` value by default and has only been enabled on the Jackson and Kingsley themes.

#### Testing instructions:

- Apply these changes:
  - To a local/self-hosted site
  - To your WP.com sandbox (`npm run deploy:push:changes` and sandbox the API)
  - To a WoA dev site (p9o2xV-1r2-p2)
- Go to Appearance > Themes
- Activate the Jackson or the Kingsley theme
- Go to Appearance > Editor
- Open the Styles sidebar
- Make sure there are no style variations
- Go to Appearance > Themes
- Activate any other Blockbase child theme (e.g. Attar)
- Go to Appearance > Editor
- Open the Styles sidebar
- Make sure style variations are available

You can also follow instructions from D107265-code to ensure that style variations are not displayed in the WP.com design picker nor in the WP.com theme showcase.

#### Related issue(s):

https://github.com/Automattic/themes/issues/6941

### DIFF
--- a/blockbase/functions.php
+++ b/blockbase/functions.php
@@ -103,6 +103,7 @@ if ( class_exists( 'WP_Theme_JSON_Resolver_Gutenberg' ) ) {
 }
 
 require get_template_directory() . '/inc/fonts/custom-fonts.php';
+require get_template_directory() . '/inc/rest-api.php';
 
 
 // Force menus to reload
--- a/blockbase/inc/rest-api.php
+++ b/blockbase/inc/rest-api.php
@@ -15,24 +15,18 @@ function blockbase_remove_style_variations_from_child_themes( $response, $handle
 
 	$handler_class  = isset( $handler['callback'][0] ) ? $handler['callback'][0] : null;
 	$handler_method = isset( $handler['callback'][1] ) ? $handler['callback'][1] : null;
+	$opt_out        = wp_get_global_settings( array( 'custom', 'optOutOfParentStyleVariations' ) );
 
 	/*
 	 * Prevents Blockbase child themes from being considered child themes in the
 	 * `wp/v2/global-styles/themes/:theme/variations` API endpoint, so they don't
 	 * inherit the style variations from the parent theme.
 	 */
-	if ( is_a( $handler_class, 'WP_REST_Global_Styles_Controller' ) && 'get_theme_items' === $handler_method ) {
+	if ( $opt_out && is_a( $handler_class, 'WP_REST_Global_Styles_Controller' ) && 'get_theme_items' === $handler_method ) {
 		add_filter( 'template_directory', 'get_stylesheet_directory' );
 	}
 
 	return $response;
 }
 
-function blockbase_check_for_opt_out_parent_variations() {
-	$opt_out = wp_get_global_settings( array( 'custom', 'optOutOfParentStyleVariations' ) );
-	if ( $opt_out ) {
-		add_filter( 'rest_request_before_callbacks', 'blockbase_remove_style_variations_from_child_themes', 10, 2 );
-	}
-}
-
-add_action( 'init', 'blockbase_check_for_opt_out_parent_variations', 99 );
+add_filter( 'rest_request_before_callbacks', 'blockbase_remove_style_variations_from_child_themes', 10, 2 );
--- a/blockbase/inc/rest-api.php
+++ b/blockbase/inc/rest-api.php
@@ -4,62 +4,27 @@
  * Removes the style variations of Blockbase child themes inherited by the parent theme.
  *
  * @param WP_REST_Response|WP_HTTP_Response|WP_Error|mixed $response Result to send to the client.
- *                                                                   Usually a WP_REST_Response or WP_Error.
- * @param array                                            $handler  Route handler used for the request.
+ * @param array $handler Route handler used for the request.
  *
- * @return WP_REST_Response Filtered response without the inherited style variations.
+ * @return WP_REST_Response|WP_HTTP_Response|WP_Error|mixed Result to send to the client.
  */
 function blockbase_remove_style_variations_from_child_themes( $response, $handler ) {
-	if ( ! $response instanceof WP_REST_Response ) {
-		return $response;
-	}
-
 	if ( ! isset( $handler['callback'] ) || ! is_array( $handler['callback'] ) ) {
 		return $response;
 	}
 
 	$handler_class  = isset( $handler['callback'][0] ) ? $handler['callback'][0] : null;
 	$handler_method = isset( $handler['callback'][1] ) ? $handler['callback'][1] : null;
-	if ( ! is_a( $handler_class, 'WP_REST_Global_Styles_Controller' ) || 'get_theme_items' !== $handler_method ) {
-		return $response;
+
+	/*
+	 * Prevents Blockbase child themes from being considered child themes in the
+	 * `wp/v2/global-styles/themes/:theme/variations` API endpoint, so they don't
+	 * inherit the style variations from the parent theme.
+	 */
+	if ( is_a( $handler_class, 'WP_REST_Global_Styles_Controller' ) && 'get_theme_items' === $handler_method ) {
+		add_filter( 'template_directory', 'get_stylesheet_directory' );
 	}
 
-	$base_directory     = get_stylesheet_directory() . '/styles';
-	$template_directory = get_template_directory() . '/styles';
-	if ( ! is_dir( $template_directory ) || $template_directory === $base_directory ) {
-		return $response;
-	}
-
-	$variations = $response->get_data();
-	if ( ! is_array( $variations ) ) {
-		return $response;
-	}
-
-	$variation_titles_parent = array();
-	$files_parent            = new RecursiveIteratorIterator( new RecursiveDirectoryIterator( $template_directory ) );
-	$variation_files_parent  = iterator_to_array( new RegexIterator( $files_parent, '/^.+\.json$/i', RecursiveRegexIterator::GET_MATCH ) );
-	foreach ( $variation_files_parent as $path => $file ) {
-		$decoded_file = wp_json_file_decode( $path, array( 'associative' => true ) );
-		if ( ! is_array( $decoded_file ) ) {
-			continue;
-		}
-
-		if ( empty( $decoded_file['title'] ) ) {
-			$variation_title_parent = basename( $path, '.json' );
-		} else {
-			// phpcs:ignore WordPress.WP.I18n.NonSingularStringLiteralText,WordPress.WP.I18n.NonSingularStringLiteralDomain
-			$variation_title_parent = translate_with_gettext_context( $decoded_file['title'], 'Style variation name', wp_get_theme()->get( 'TextDomain' ) );
-		}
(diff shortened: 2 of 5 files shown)


pbking:
Rather than the added complexity perhaps we should just consider removing the style variations from Blockbase.  As a parent theme I'm rather keen to just remove it from general use anyway.  I don't expect we'll be making many (any) more parent block themes either but if we do they shouldn't have their own style variations.

mmtr:
> consider removing the style variations from Blockbase.

Sure, I'd be more than happy to do that as long as the Theme team doesn't identify any problem with that approach. It'll make things much simpler technically (it'd also remove the need of D107265-code), but then sites with the Blockbase theme activated (I believe Atomic and self-hosted sites have the ability to activate it) will lose access to those style variations. Is that fine?

mmtr:
Opened https://github.com/Automattic/themes/pull/7001 as an alternative approach that directly removes the style variations from the Blockbase theme.

matiasbenedetto:
Hi :wave: ! I did some testing with some Blockbase children such as Quadrat and Jackson and they seem to be working fine with the variations.

Jackson:

https://user-images.githubusercontent.com/1310626/230990985-8c8ab7d7-f424-42e0-b1d3-c5cdbcb024d5.mp4

Quadrat:

https://user-images.githubusercontent.com/1310626/230991029-edecc026-3e49-4199-bc3e-ddf646c792f2.mp4



As per the video of the PR description, it seems like what is looking odd are some colors of the elements used and that seems to be caused by the elements used on that post instead of a theme problem.

I'm not completely sure if that markup is coming from [this pattern](https://github.com/Automattic/themes/blob/6f8f4143c8670429fb5d1dfa5dba8e46c0f4fedf/blank-canvas/inc/block-patterns.php#L458) but they look pretty similar. Is it coming from there? I don't see that layout (Patricia Jackson, etc.) in the Jackson theme source code, where it is coming from?

If it's the same content I think the problem is related to the cover block used in the pattern. It has an overlay color set. When you change the palette of the site it may look odd. Apart from that, it doesn't make sense to use a cover to create that layout because it is not using a background image/media. The image (the circle with the line) is added as an image block.

For these reasons I think we should let the Blockbase variations as they are. 
Probably we should try to solve this problem by fixing the markup used in the video of the PR description and, if we are still updating blank-canvas theme, update [the pattern](https://github.com/Automattic/themes/blob/6f8f4143c8670429fb5d1dfa5dba8e46c0f4fedf/blank-canvas/inc/block-patterns.php#L458).

mmtr:
> As per the video of the PR description, it seems like what is looking odd are some colors of the elements used and that seems to be caused by the elements used on that post instead of a theme problem.

Yeah sorry, I didn't a good job explaining the underlying issue in the PR description. The culprit is actually the headstart content, as I try to better explain in [this other alternative PR](https://github.com/Automattic/themes/pull/7001):

> This is causing issues to Blockbase child themes such as Jackson in WordPress.com, because the headstart content for a child theme is only created for the default style variation. When a different style variation is applied, the styles from the headstart content are incompatible with the new variation

--- 

> solve this problem by fixing the markup used in the video

Would this work retroactively for all sites that already have the affected content generated with headstart?

mmtr:
> Is it coming from there? I don't see that layout (Patricia Jackson, etc.) in the Jackson theme source code, where it is coming from?

It comes from `wpcom-pub-themes/jackson/inc/headstart/en.json#31`.

Seems to be a `core/group` block that has a gradient style:

```html
<!-- wp:group {\"align\":\"full\",\"style\":{\"spacing\":{\"padding\":{\"top\":\"120px\",\"right\":\"0px\",\"bottom\":\"120px\",\"left\":\"0px\"}},\"color\":{\"gradient\":\"linear-gradient(180deg,rgb(244,253,255) 0%,rgb(255,255,255) 100%)\"}},\"layout\":{\"inherit\":true}} -->\n<div class=\"wp-block-group alignfull has-background\" style=\"background:linear-gradient(180deg,rgb(244,253,255) 0%,rgb(255,255,255) 100%);padding-top:120px;padding-right:0;padding-bottom:120px;padding-left:0;\"><!-- wp:image {\"align\":\"left\",\"id\":16,\"width\":152,\"height\":152,\"sizeSlug\":\"large\",\"linkDestination\":\"none\"} -->\n<div class=\"wp-block-image\"><figure class=\"alignleft size-large is-resized\"><img src=\"https:\/\/jacksondemo.files.wordpress.com\/2021\/12\/pattern-links-logo.png?w=304\" alt=\"A logo of a circle with a line through it.\" class=\"wp-image-16\" width=\"152\" height=\"152\" \/><\/figure><\/div>\n<!-- \/wp:image -->\n\n<!-- wp:spacer {\"height\":16} -->\n<div style=\"height:16px;\" aria-hidden=\"true\" class=\"wp-block-spacer\"><\/div>\n<!-- \/wp:spacer -->\n\n<!-- wp:heading {\"textAlign\":\"left\",\"level\":1} -->\n<h1 class=\"has-text-align-left\" id=\"patricia-jackson\">Patricia Jackson<\/h1>\n<!-- \/wp:heading -->\n\n<!-- wp:paragraph -->\n<p>Published work and ephemera.<\/p>\n<!-- \/wp:paragraph -->\n\n<!-- wp:spacer {\"height\":60} -->\n<div style=\"height:60px;\" aria-hidden=\"true\" class=\"wp-block-spacer\"><\/div>\n<!-- \/wp:spacer -->\n\n<!-- wp:paragraph -->\n<p><a href=\"#\">\u201cThe Lost Tricycle\u201d Book<\/a><\/p>\n<!-- \/wp:paragraph -->\n\n<!-- wp:paragraph -->\n<p><a href=\"#\">\"Why we must own our history\"<\/a> in The Atlantic<\/p>\n<!-- \/wp:paragraph -->\n\n<!-- wp:paragraph -->\n<p><a href=\"#\">\"Identity and Ownership\"<\/a> in The New York Times<\/p>\n<!-- \/wp:paragraph -->\n\n<!-- wp:paragraph -->\n<p><a href=\"#\">Sponsor: Crafty Cookies<\/a><\/p>\n<!-- \/wp:paragraph -->\n\n<!-- wp:paragraph -->\n<p><a href=\"#\">Donate to help keep us posting!<\/a><\/p>\n<!-- \/wp:paragraph --><\/div>\n<!-- \/wp:group -->
```

mmtr:
I reviewed all the Blockbase child themes and seems that only Jackson and Kingsley present major incompatibilities between the headstart content and the style variations:

- Jackson

https://user-images.githubusercontent.com/1233880/231149910-839b77c7-3192-4485-81ca-741b55f6a380.mov

<br>

- Kingsley

https://user-images.githubusercontent.com/1233880/231149961-9052eaba-b97d-4d13-8bdd-1b17d18cbbcb.mov

<br>
There are a few other child themes that have minor incompatibilities:

- Appleton (Charcoal: site title too dark, Ruby wine: Same as Default)

https://user-images.githubusercontent.com/1233880/231154618-b93dd45c-9b73-4986-b1f7-6226c3c2627b.mov

<br>

- Attar (Ruby wine: Same as Default)

https://user-images.githubusercontent.com/1233880/231154986-c4e62a30-202f-420e-9259-083525a1d889.mov

<br>

- Dorna (Ruby wine: Same as Default)

https://user-images.githubusercontent.com/1233880/231155491-a709d99f-795e-4a21-887a-36620a15db81.mov

<br>

- Farrow (Charcoal: text too dark)

https://user-images.githubusercontent.com/1233880/231155812-ed272716-8074-43c8-bad0-0a335583636a.mov

<br>

- Heiwa (Ruby wine: Same as Default)

https://user-images.githubusercontent.com/1233880/231156261-cff6bba8-3f50-405d-b2a2-88bd6e718970.mov

<br>

- Meraki (Ruby wine: Same as Default)

https://user-images.githubusercontent.com/1233880/231156496-6f442a8e-cc35-4112-9a1b-aee6f964f0e4.mov

<br>

- Russell (Charcoal: low contrast text, Ruby wine: Same as Default)

https://user-images.githubusercontent.com/1233880/231166544-851d62fd-3908-458f-8eb7-12fcc8ac9953.mov

<br>

- Zoologist (White: Same as Default)

https://user-images.githubusercontent.com/1233880/231167278-bc523cf6-6dba-43e4-bed8-50a69de00b4a.mov

---

Some of the minor incompatibilities also happen without using headstart content on a self-hosted/local site

pbking:
Consider instead this change for Jackson: D107566-code

The changes needed for Kingsley are a little more involved.  The background of the cover block should be eliminated to allow the background color to bleed through, and the text in that block should use the default foreground colors.  I'm not sure why/how the original content isn't doing that.

pbking:
Also note: The reason why most of those other child themes look “the same” with “ruby wine” variation is because they are using only TWO colors while Blockbase is making available five.  Thus those primary, secondary and tertiary color values, which are what make that variation special, aren’t used and are therefore not useful for those themes.  (Including Jackson, but also many more and noted in github)

After tinkering with these child themes I’m not very keen on that choice (made looong ago) to have only two colors for many of these child themes.  But I don’t know that changing it now is a good idea.  I just don’t like it. 

However including that style variation (Ruby Wine) for themes that only leverage two colors (foreground & background) ALSO doesn’t make much sense.

Until very recently none (most?) of these child themes had ANY style variations.  Thus removing them made sense.  But removing them via code for child themes was overly complicated and just removing them from Blockbase was done at the detriment of that theme.

I’m not sure what to do about it all.  I just wanted to get those thoughts down.

We might also want to either REMOVE Ruby Wine (since it’s so un-different for so many themes) or change it in a more dramatic way (for instance, set a slightly red background color or something).

mmtr:
> Until very recently none (most?) of these child themes had ANY style variations. Thus removing them made sense. But removing them via code for child themes was overly complicated and just removing them from Blockbase was done at the detriment of that theme.

IMO, Core's decision to expose style variations from parent themes to child themes was a bit drastic since it can have negative consequences for child themes that have been built with the (until then correct) expectation that they won't have any other style variations than the ones designed specifically for them.

It should have been an opt-in behavior. Or at the very least, provide an easy way to opt out.

That's why I think we should hide the style variation from Blockbase child themes. And I realize the code needed for that is complex (I wish I could do something simpler) but Core doesn't give us any other way.

EDIT: In https://github.com/Automattic/themes/pull/6996/commits/f51deda48147f384ade8a13e0b42a483aa59fa9b, I updated the logic to use an approach that is slightly less complicated.

PR #7007, opened 2023-04-11
Remove Videomaker and all premium theme tooling

pbking:
This change:

* Removes Videomaker from this repository
* Removes all of the "premium theme" tooling from the theme-utils.

The Videomaker theme has been relocated here: https://github.com/automattic/theme-videomaker

Because Videomaker is the ONLY "premium" theme in this repository maintaining tooling to support this (rarely updated) theme only adds complication to the process.

There are many other themes, even Blockbase child themes, that are stored in individual repositories like the one it has been moved too.

Since it's a WordPress.com premium theme it has the parent 'blockbase-premium' rather than 'blockbase' like every other Blockbase child housed in this repository.


mikachan (approved):
Thanks for doing this! I 100% agree with moving Videomaker to its own repo and removing the premium tooling.

Is there anyone we need to inform before we make this change, maybe on the premium side?

vcanales:
Fixed conflicts, and merging.

PR #7013, opened 2023-04-13
Otis: Add theme for review

matiasbenedetto:
Otis: Add theme for review
Created by @beafialho 

image

Notes  by @beafialho :

We really need more and better responsive controls to be able to set different global padding for desktop and mobile. In a case like this, we should even be able to define the size of the navigation hamburger icon and customise its thickness to match the site’s remaining elements.

The “About”, “Contact” and “Stories” should use the navigation block, but it’s not possible to use separators there and achieve this layout, which is kind of a dealbreaker. I used a query block instead.

It took a little longer than that session to polish it but this should be ready for a technical review.


matiasbenedetto (commented):
LGTM

PR #7017, opened 2023-04-13
Kigen: Add theme for review

matiasbenedetto:
Kigen: Add theme for review
Created by @iamtakashi 

image


pbking:
I'm currently running into a problem with making the position of one of the template parts works in this theme.

Because there is content that needs to be localized the entire content of the 'front-page' template resides in the `front-page.php` pattern.  (Concerns of using the front-page template aside...)

The footer for this theme is not just at the bottom of the content, therefore just moving it out of the pattern doesn't work.  So there's a `template-part` block in that pattern.

The 'theme' attribute has to be removed from template part blocks from themes (otherwise they don't work correctly and can't be found if the theme is installed in a different location than what is referenced in the 'theme' template).  HOWEVER when the theme attribute is removed from a template-part block in a PATTERN it DOESN'T work.  The only way for a template-part to work when referenced from a pattern is to INCLUDE the 'theme' attribute.  And, as we have established, since a theme can be installed in folders the 'theme' attribute CAN'T be included or it won't work in environments where the theme is installed in a folder.

pbking:
Hah.  I think typing out the issue helped me to solve it.

Rather than reference a 'footer-2' template-part which only references a 'footer-2' pattern I just instead reference the 'footer-2' pattern from the 'front-page' pattern and got rid of the 'footer-2' template-part completely.

Problem solved.

pbking:
I THINK this is good to come in, but I had one concern before doing so.  This theme uses the `front-page.html` template and that gives me some concern.  I believe that the same thing COULD be achieved by using the `home.html` template instead but still allow the user to choose between the homepage content or a blogroll style.

@iamtakashi if you could confirm that the `front-page` template is indeed the desired way to supply the content for this theme then I'm good to bring this in as it is now.

Otherwise, I believe that the `front-page` template should be refactored to `home` and THEN this will be good to bring in.

iamtakashi:
@pbking Thanks for reviewing this theme. I appreciate it. 

Yes, I'm happy to ship this with a front-page template. This theme was never designed to show a blog stream on the homepage, just like LiB or Newsletter sites.

pbking (approved):

PR #7029, opened 2023-04-20
Blockbase Children: remove conflicting `core/navigation-link` styles

### LINKED ISSUE #74984: Blockbase themes: primary theme color is used instead of Navigation block color setting
### Quick summary

Tested on Quadrat Black and Quadrat Green. When you set a "Text" (or "Submenu & overlay text") color on a Navigation block, links show up in the selected color in the editor. On the live site, however, only ~~custom links (not page links)~~ _*Submenu parents_ are affected; the rest remain in the theme's primary color.
_*edit from cuemarie, see [comment](https://github.com/Automattic/wp-calypso/issues/74984#issuecomment-1491121656)_

### Steps to reproduce

1. Activate a Quadrat theme.
2. Open the Site Editor.
3. Insert a Navigation block (if one doesn't already exist) and add links.
4. Select the Navigation block and set a "Text" color in the block settings.
5. Save.

As in the editor, all of the Navigation block's links should render in the chosen color.

### What actually happened

The color affects the editor as expected. On the live site, "page" links remain in the default "primary" theme color. Only custom links are affected.

Markup on 2023-03-27 at 15:27:07

### Impact

Some (< 50%)

### Available workarounds?

Yes, easy to implement

Simple

### Logs or notes

Not yet tested on Atomic.

The Nav block's selected color applies to the `<nav>` element, and has `!important`, but the theme applies a color to the more specific `.wp-block-navigation-link`, which takes precedence.

CSS can be used (for sites with a Premium plan or higher) to set a desired color. For free/Personal sites, additional CSS classes can be used in the editor (e.g. `has-white-color` for white links), but that needs to be set on each individual link within a Nav block (not on the Nav block itself).

mattwiebe:

Remove `core/navigation-link` styles in Blockbase children so that they properly inherit from the parent nav block. I dug for a while but was unable to figure out why the styles were present in the first place.

Prior to this, `core/navigation-link` styles would always override the parent nav block's color. I couldn't see any reason for those styles, perhaps somebody will be able to figure it out.

Note that in my testing you could only access the Navigation Link block when loading a "Classic" menu for the nav menu block. The page list menu doesn't use `core/navigation-link` it seemed. 

#### Related issue(s):

Fixes https://github.com/Automattic/wp-calypso/issues/74984

### DIFF
--- a/geologist/theme.json
+++ b/geologist/theme.json
@@ -199,12 +199,6 @@
 					"lineHeight": "1.6"
 				}
 			},
-			"core/navigation-link": {
-				"color": {
-					"background": "transparent",
-					"text": "var(--wp--custom--color--foreground)"
-				}
-			},
 			"core/post-title": {
 				"typography": {
 					"fontSize": "min(max(3rem, 7vw), 4rem)",
--- a/quadrat/theme.json
+++ b/quadrat/theme.json
@@ -210,12 +210,6 @@
 					"fontSize": "var(--wp--custom--font-size--normal)"
 				}
 			},
-			"core/navigation-link": {
-				"color": {
-					"background": "transparent",
-					"text": "var(--wp--custom--color--foreground)"
-				}
-			},
 			"core/post-title": {
 				"typography": {
 					"fontSize": "min(max(3rem, 7vw), 5rem)",
--- a/videomaker/theme.json
+++ b/videomaker/theme.json
@@ -96,12 +96,6 @@
 	},
 	"styles": {
 		"blocks": {
-			"core/navigation-link": {
-				"color": {
-					"background": "transparent",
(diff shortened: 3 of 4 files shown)


mattwiebe:
@Automattic/theam I made D108515-code and D108513-code based on this for `pub` and `premium`, respectively. It's entirely possible I did something wrong but let's get this moving because it's [bothering a lot of customers](https://github.com/Automattic/wp-calypso/issues/74984).

vcanales (approved):
Thanks! This change does make customization more intuitive. I tested with the Blockbase children affected by the PR and it LGTM.

It might be worth it to keep an eye on this for any regressions, though; folks may have been working around this quirk on their own.

PR #7031, opened 2023-04-20
Skip loading the Learning Mode styles when Sensei is not available

gabrielcaires:
Partially solve [Review spacings and alignment on the course theme#6834](https://github.com/Automattic/sensei/issues/6834)

* Skip the learning mode style when sensei is not available
* Add a temporary hook to skip the loading when Sensei is available and able to provide a new learning mode style.

#### Testing instructions
* Go to this branch
* Disable the sensei-lms plugin
* Go to the home page
* Check if the `learning-mode.css` is not loaded
* Enable the sensei-lms plugin
* Check if the `learning-mode.css` is loaded.

### Context
The sensei team is delegating the responsibility to load the learning mode style to the plugin, instead of the handle it on the theme. It will help us to avoid coupling between the theme version and the plugin version.


course/functions.php
```diff
@@ -42,33 +42,24 @@ function course_support() {
...
-		);
-
-		// Enqueue theme stylesheet.
+		wp_register_style( 'course-style', get_stylesheet_directory_uri() . '/style.css', array(), wp_get_theme()->get( 'Version' ) );
+		wp_enqueue_script( 'course-header', get_template_directory_uri() . '/assets/js/header.js', [], wp_get_theme()->get( 'Version' ), true );
 		wp_enqueue_style( 'course-style' );
 
-		// TODO: Only Load it if the the page is using learning mode
-		wp_enqueue_style( 'course-sensei-learning-mode' );
+		$has_sensei = function_exists( 'Sensei' );
```
donnapep: I think we should check if the Sensei dependency is met in the same way as our other plugins - https://github.com/Automattic/sensei-pro/blob/trunk/includes/class-sensei-pro-dependency-checker.php#L107.

course/functions.php
```diff
@@ -42,33 +42,24 @@ function course_support() {
...
+
+		/**
+		 * Temporary Hook to skip the learning mode style when the Sensei LMS is able to provide it.
+     * It only use the deprecated style if an old sensei version is installed. 
+		 */
+		$use_style_from_sensei = apply_filters( 'internal_course_theme_use_learning_mode_style_from_sensei', false );
```
donnapep: We should follow our [naming conventions](https://senseip2.wordpress.com/programming-standards-best-practices/naming-conventions/#hooks), so we could call it something more succinct like `course_learning_mode_load_styles`.

course/functions.php
```diff
@@ -42,33 +42,24 @@ function course_support() {
...
 
-		// Enqueque theme scripts.
-        wp_enqueue_script( 'course-header', get_template_directory_uri() . '/assets/js/header.js', [], wp_get_theme()->get( 'Version' ), true );
-    }
+		if ( $has_sensei && empty($use_style_from_sensei)  ) {
```
donnapep:
```suggestion
		if ( class_exists( 'Sensei_Main' ) && ! $use_style_from_sensei  ) {
```

gabrielcaires: @donnapep All suggestions was updated here 3eaa16164741d7e99b1db6bdde1529756c33d4d0

course/functions.php
```diff
@@ -10,6 +10,7 @@
  */
 
 
+// require 'wp-admin/includes/plugin.php';
```
m1r0: Is this a leftover?

gabrielcaires: Yes, thanks! Removed here 801f7d39b0ef87314de988cbc77ace323be2118d

donnapep (approved):
Looks good! When do you think we should consider removing the `learning-mode.css` file altogether? Perhaps with the next major release after the next one? If so, perhaps we could add a note in the file about it so we don't forget.

gabrielcaires:
@donnapep  I created this issue https://github.com/Automattic/themes/issues/7429 . I also added as part of our release after the current one.

PR #7034, opened 2023-04-22
Add body class for the theme variation

merkushin:
As there is no reliable way to determine the variation on backend, we save the variation in an option at the moment it the global styles are saved.

- Save theme variation on global styles update.
- Add body class for the theme variation.

#### Testing instructions:
- Go to `Appearance -> Themes` and activate the Course theme.
- Go to `Appearance -> Editor -> Templates -> Lesson (Learning Mode - Default)` (Not really important which template to select).
- Click `Styles -> Browse Styles`, select one, click `Save`.
- Go to front end and check the `<body>` tag has a class with the name `is-{lowercased-variation-title}`.
- Go to the Editor and change variation to another one. Save.
- Go to the front end and check that the body tag variation class has updated.
- Go to the Editor and reset global styles to defaults.
- Check the body tag has `is-default` CSS class.

#### Related Issues:
- https://github.com/Automattic/themes/issues/7123

### DIFF
--- a/course/functions.php
+++ b/course/functions.php
@@ -128,22 +128,6 @@ function course_save_global_styles( $post_id, $post, $update ) {
 add_action( 'save_post', 'course_save_global_styles', 10, 3 );
 
 /**
- * Delete the theme variation option when the global styles post is deleted.
- *
- * @param int     $post_id Post ID.
- * @param WP_Post $post    Post object.
- */
-function course_delete_global_styles( $post_id, $post ) {
-	if ( 'wp_global_styles' !== $post->post_type ) {
-		return;
-	}
-
-	delete_option( 'course_theme_variation' );
-}
-
-add_action( 'delete_post', 'course_delete_global_styles', 10, 2 );
-
-/**
  * Add the theme variation to the body class.
  *
  * @param array $classes Array of body classes.
--- a/course/functions.php
+++ b/course/functions.php
@@ -154,7 +154,7 @@ function course_add_variation_body_class( $classes ) {
 		return $classes;
 	}
 
-	$classes[] = 'course-variation-' . $current_variation;
+	$classes[] = 'is-' . $current_variation;
 
 	return $classes;
 }
--- a/course/functions.php
+++ b/course/functions.php
@@ -119,7 +119,7 @@ function course_save_global_styles( $post_id, $post, $update ) {
 	$current_variation  = 'default';
 	foreach ( $variations as $variation ) {
 		if ( $variation['settings'] === $global['settings'] && $variation['styles'] === $global['styles'] ) {
-			$current_variation = strtolower( $variation['title'] );
+			$current_variation = sanitize_title( $variation['title'] );
 		}
 	}
 	update_option( 'course_theme_variation', $current_variation );
--- a/course/functions.php
+++ b/course/functions.php
@@ -102,6 +102,10 @@ add_action( 'init', 'course_register_block_patterns_category' );
 
 /**
  * Determine the theme variation and save in option.
+ *
+ * @param int     $post_id Post ID.
+ * @param WP_Post $post    Post object.
+ * @param bool    $update  Whether this is an existing post being updated or not.
  */
 function course_save_global_styles( $post_id, $post, $update ) {
 	if ( 'wp_global_styles' !== $post->post_type ) {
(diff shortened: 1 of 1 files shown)


carolinan:
I really wish this could be fixed in Gutenberg / core.

mikachan:
Dropping a link to a related Gutenberg issue: https://github.com/WordPress/gutenberg/issues/43405

course/functions.php
```diff
@@ -96,4 +98,56 @@ function course_register_block_patterns_category() {
 	);
 }
 
-add_action( 'init', 'course_register_block_patterns_category' );
\ No newline at end of file
+add_action( 'init', 'course_register_block_patterns_category' );
+
+/**
+ * Determine the theme variation and save in option.
```
donnapep, line 104: The parameters are missing from the DocBlock for the new functions that were added.

course/functions.php
```diff
@@ -96,4 +98,56 @@ function course_register_block_patterns_category() {
...
+/**
+ * Add the theme variation to the body class.
+ */
+function course_add_variation_body_class( $classes ) {
+	$current_variation = get_option( 'course_theme_variation' );
+	if ( ! $current_variation ) {
+		return $classes;
+	}
+
+	$classes[] = 'course-variation-' . $current_variation;
```
donnapep: What do you think about calling this class `is-variation-name` (e.g. `is-default`) as proposed in the Gutenberg [issue](https://github.com/WordPress/gutenberg/issues/43405)? If they end up using that selector name, then we wouldn't need to rename our CSS classes to match later.

course/functions.php
```diff
@@ -96,4 +98,56 @@ function course_register_block_patterns_category() {
...
+
+	return $classes;
+}
+
+add_action( 'body_class', 'course_add_variation_body_class' );
```
donnapep, line 146: I think we'll also need to add this class to the editor so that we can apply styles there too; however, I think this might be a bit tricky so it may be better to create a separate card for it so that we don't hold this one up.

donnapep:
@merkushin Are there any testing instructions for this PR? In particular, when is it expected that `course_delete_global_styles` will fire? When I tried switching to another theme, the option wasn't cleaned up. Should we consider doing that?

Also, is it a fair statement to say that existing installs of the Course theme won't have the new styles applied that use this CSS selector until they save something in the site editor?

course/functions.php
merkushin: Created a separate card for that: https://github.com/Automattic/sensei/issues/6852

course/functions.php
merkushin: Fixed here: https://github.com/Automattic/themes/pull/7034/commits/2ca8a1e0df54aaaad4f425b29c4d09dbe09c5e96

course/functions.php
merkushin: Fixed here: https://github.com/Automattic/themes/pull/7034/commits/271357bc1d368a290b068058f12ff12939241df1

Not really sure if it is the best approach as we are supposed to have vendor prefix, etc.

merkushin:
@donnapep I added testing instruction to the description, thank you.

> In particular, when is it expected that course_delete_global_styles will fire? 

My mistake: it won't be fired. The post isn't deleted even if we delete the theme. I'll remove it from code to avoid confusion.

> When I tried switching to another theme, the option wasn't cleaned up. Should we consider doing that?

I don't think so. If the user switch back to the Course theme later, all their style changes remain the same as before switching the theme. WP keeps a `wp_global_styles` post for each theme.

> Also, is it a fair statement to say that existing installs of the Course theme won't have the new styles applied that use this CSS selector until they save something in the site editor?

That's true. I think we can try to detect the variation. For users without custom changes it will work. However, in case they have some, we can't reliably detect the variation.

Not sure if it is worth doing, so I don't do these changes now. If we decide to do it, it might worth to do in a separate PR.

course/functions.php
```diff
@@ -96,4 +98,49 @@ function course_register_block_patterns_category() {
...
+	}
+
+	$global_styles      = json_decode( $post->post_content, true );
+	$global['settings'] = isset( $global_styles['settings'] ) ? $global_styles['settings'] : array();
+	$global['styles']   = isset( $global_styles['styles'] ) ? $global_styles['styles'] : array();
+	$variations         = WP_Theme_JSON_Resolver::get_style_variations();
+	$current_variation  = 'default';
+	foreach ( $variations as $variation ) {
+		if ( $variation['settings'] === $global['settings'] && $variation['styles'] === $global['styles'] ) {
+			$current_variation = strtolower( $variation['title'] );
```
m1r0: What do you think of using the `sanitize_title` function instead of `strtolower`? Just in case we start using titles with multiple words.

Basically, the function converts a string into a slug - `Dark Blue` -> `dark-blue`.

donnapep (approved):
Works well!

PR #7046, opened 2023-05-02
Update the blue variation typography

### LINKED ISSUE #7111: Standard Layout on the Frontend (Blue): Colors and typography
According to the provided design, apply spacing, alignment and typography for the Standard (Default) Layout in Course Theme: Blue variation.

## Scope:

- Apply colors and typography based on the design.
- Desktop and mobile.
- Course Theme: Blue variation.
- Standard (Default) Layout.
- Frontend only.

## Design

Please check Figma for the design:
8mY8ui4zYK1Ylw2yIFi9Lx-fi-20020045\_37699

m1r0:
* Update the typography for the blue variation of the theme.

Testing instructions:
* Checkout the `feature/learning-mode-improvements` branch for Sensei LMS.
* Go to Appearance -> Themes and activate the Course theme.
* Go to Appearance -> Editor -> Templates -> Lesson (Learning Mode - Default).
* Click Styles -> Browse Styles -> Blue, click Save.
* Compare the typography with the design. Use an HTML vs Design tool ([example](https://chrome.google.com/webstore/detail/pixelparallel-by-htmlburg/iffnoibnepbcloaaagchjonfplimpkob?hl=en)) because the CSS provided by the Figma design is not always accurate (mainly for letter spacing).

#### Related issue(s):
Resolves: https://github.com/Automattic/themes/issues/7111


course/styles/blue.json
```diff
@@ -208,7 +208,7 @@
 					"size": "3rem",
 					"fluid": {
 						"min": "3rem",
-						"max": "4rem"
+						"max": "3.25rem"
```
m1r0, line 211: Note that I'm updating the global theme style here to make the headers font size consistent. Let me know if you think this is not expected.

donnapep:
@m1r0 I have questions. 😆 Is this the design you were working from - https://www.figma.com/file/8mY8ui4zYK1Ylw2yIFi9Lx/Learning-Mode?node-id=20020057-38548&t=7kdfaJq3nRCxsxLK-0
I want to be sure I'm looking at the right one.

If so, there are a few simple things I noticed right away that made me question if we're on the same page. I'm not planning to do more in-depth testing until I'm confident that we are:

- The site title should be bold, Noto Sans and `18px`. Right now it's Lora and `24px`.
- The case of the module names in the sidebar isn't right.
- The lesson title should be `52px` but it's `64px`.

Were there any color changes needed as part of this PR? Looks like no but just want to confirm.

It's also possible that I'm not picking up the variation changes locally for some reason. Would you mind sharing a screenshot of your changes?

m1r0:
> It's also possible that I'm not picking up the variation changes locally for some reason.

@donnapep Oh, I forgot to mention that you need to be on [this branch](https://github.com/Automattic/sensei/pull/6823) (merged a few hours ago). Sorry about that.

> Were there any color changes needed as part of this PR?

I wasn't able to find any color changes in the design, so there're no color changes.

> Would you mind sharing a screenshot of your changes?

Sure. Here're a few screenshots of the html-vs-design:

* Course title

* Navigation: Module title

* Navigation: Lesson title

* Post title

* Module title before the post title

* Progress

* Content

* Buttons

Imran92:
I've taken a look into it with our updated feature brunch. Most of it looks alright to me. Here are a few things I've noticed.

Probably the letter spacings can be changed a bit for lesson content
Screen Shot 2023-05-08 at 12 38 44 PM

Some fonts are different in the desktop and mobile versions of the design. I'm not sure if they're intentional and if we should do something about them

For example, the module title in navigation is Noto Sans in the desktop version but Montserrat in mobile

Screen Shot 2023-05-08 at 12 50 48 PM

Same for the Exit Course button
Screen Shot 2023-05-08 at 12 51 33 PM

The extra info below lessons are of different font size and probably spacing in mobile
Screen Shot 2023-05-08 at 1 01 23 PM

m1r0:
Thanks for the review @Imran92!

> Probably the letter spacings can be changed a bit for lesson content

Hmm, it looks good on my side, could you please double-check? Maybe when you took the viewport was below 1920px? At some point below 1920px the font-size [gets smaller](https://user-images.githubusercontent.com/1612178/236907123-cb4234c7-660f-43da-a532-b666def03ac8.mp4). Or maybe the issue is in my television... 🤷 

> Some fonts are different in the desktop and mobile versions of the design.

I've [asked](p1683555784526489-slack-C013QUH20TS) Andrei about it and it turned out it was a copy-paste issue in the design. The Figma file was updated so all fonts use Noto Sans now. Could you please check it again? Thank you! 🙏

course/styles/blue.json
merkushin: I was asking @Luchadores about headings. He said that in general we consider headings are already fine. Are you fixing a specific case?
And I want to mention that now Huge is almost the same as Extra large. If we look at other sizes they have some logic in those size periods: a larger size is a continuation of a smaller.

merkushin:
On desktop, everything looks good. However, font weight could be improved. In design, text is just a bit thicker.
CleanShot 2023-05-09 at 14 20 16@2x
CleanShot 2023-05-09 at 14 20 33@2x
CleanShot 2023-05-09 at 14 20 02@2x
CleanShot 2023-05-09 at 14 27 19@2x
CleanShot 2023-05-09 at 14 28 27@2x
CleanShot 2023-05-09 at 14 28 34@2x

merkushin:
Similar for mobile
CleanShot 2023-05-09 at 14 54 06@2x
CleanShot 2023-05-09 at 14 54 25@2x
CleanShot 2023-05-09 at 14 55 24@2x
CleanShot 2023-05-09 at 14 58 40@2x

Fonts in course outline on mobile look wrong. I think it is mostly about letter spacing and font weight. (Don't pay attention to the font size, it is okay when I choose a device with a smaller screen.)
CleanShot 2023-05-09 at 15 00 02@2x
CleanShot 2023-05-09 at 14 59 44@2x
CleanShot 2023-05-09 at 14 59 17@2x

Imran92:
> Hmm, it looks good on my side, could you please double-check? Maybe when you took the viewport was below 1920px? At some point below 1920px the font-size [gets smaller](https://user-images.githubusercontent.com/1612178/236907123-cb4234c7-660f-43da-a532-b666def03ac8.mp4). Or maybe the issue is in my television... 🤷

Yeah it's probably a viewport issue on different screen sizes.

> I've [asked](https://github.com/Automattic/themes/pull/p1683555784526489-slack-C013QUH20TS) Andrei about it and it turned out it was a copy-paste issue in the design. The Figma file was updated so all fonts use Noto Sans now. Could you please check it again? Thank you! 🙏

Oh nice! Won't have to use custom CSS to change font then! <3

m1r0:
@merkushin The font thickness looks good on my screen. I just double-checked the CSS properties vs the design-generated CSS properties and found just one mismatch ([fixed here](https://github.com/Automattic/themes/pull/7046/commits/f28659cb2f015c9cc8183a4def7801e49b310658)). Maybe we are using different browsers? Are you testing on Chrome? Could you compare the CSS properties generated by the design and the ones in the HTML?

I'm comparing with [this artboard](https://www.figma.com/file/8mY8ui4zYK1Ylw2yIFi9Lx/Learning-Mode?type=design&node-id=20020057-38548&t=BhnJ552G8uGcuALT-0) and here's what I see:

image

course/styles/blue.json
m1r0: This font size is used for the H1 tag which is the post title.

image

The issue for me here is that we're using a 52px (3.25rem) font size for the H1 post title in Learning Mode and a different font size (4rem) for the H1 post title outside of Learning Mode. Is this the expectation? If this is the case, I'm not sure how to implement it because it's the same element but in a different context (learning mode vs non-learning mode). I don't know if we can do it using the JSON configs. Is there another approach?

Btw, we're having the same issue with the other variants ([gold](https://github.com/Automattic/themes/pull/7047/files#diff-d085156ba069696207d9f71dc2929662df0276144b7f4ef4e1e25b49da4efa09R196-R197), [dark](https://github.com/Automattic/themes/pull/7065/files#diff-6186ce83da9d3760d88af574107425d815430ada5b56c5920ac0f227c72c883dR171-R172)).

@Luchadores Can you help us figure this out? Is it expected for the H1 post title to be the same font size in Learning Mode and outside of it? Is it OK if we use the same font size?

merkushin: I think, a way we can address this issue is to add a rule to our Learning Mode CSS and use the variation body class for that.

m1r0: I did try using the variation body class but I've noticed that the class is removed once you make a single global style change. For example, in this case, if you change the background color, the H1 font size will change as well. Are we OK with this? For me, it's a bit unreliable.

Luchadores: Hey Miro :wave-up-2: Global styles should apply everywhere. Having the same buttons, the same family-fonts are great being the same. For the font sizes, I would be careful. I believe it’s fine to have the same font size for the body of the Learning Mode content, but the course outline should stay as it is.

For example:
image

Hello and welcome can be the same h1 defined in the global styles by the user. But the content on the left side would probably break if we tie up together with the same body, h1, h2, h3… font sizes.

merkushin:
Checked font weight values in a few places. You're right, numbers are correct. So that's something on my side.
I suppose, it is related to my screen size and resolution (the largest resolution for me is 1680 x 1050).

merkushin (approved):

PR #7055, opened 2023-05-04
Corrected version numbers that should have already been at 1.x or higher

pbking:
There are a number of themes that are already in production that don't have versions of 1.x or higher.  None of these are considered "in development" or "experimental" and should have been bumped upon release.

Rather than setting all of these versions to "1.0" I have instead set the versions to reflect the changes since the beginning; thus only changing the major versions from 0 -> 1.

There were a few themes I did NOT change:

Blank Canvas Blocks (I think that's only used for development, I don't believe this is in production)
Block Canvas (Only used as a base for other themes.  Isn't in production and I don't believe should have a 1.0 version)
Gramming (Not yet in production)
Verso (Not yet in production)

### DIFF
--- a/antonia/style.css
+++ b/antonia/style.css
@@ -7,7 +7,7 @@ Description: Antonia is a theme for selling products with the help of payments b
 Requires at least: 5.8
 Tested up to: 5.9
 Requires PHP: 5.7
-Version: 0.0.5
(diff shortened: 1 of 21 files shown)


matiasbenedetto (approved):
Thanks for the update.

LGTM

PR #7059, opened 2023-05-05
Update Node version in .nvmrc & update deps

mikachan:
This PR updates the minimum required Node version from 17 to 18, as I was seeing the following error in this repo:

Screenshot 2023-05-05 at 12 36 31

This also updates all dependencies.

Testing instructions:

1. Run `rm -rf node_modules` to delete your current node_modules
2. Run `nvm use` to switch to node v18
3. Run `npm ci` to install the updated modules from the package-lock file in this PR
4. There should be no changes to the package-lock file once you've run all these commands


jeffikus:
I tried with my outdated version of node and npm. Got this error;

Screenshot 2023-05-05 at 14 16 33

Then after upgrading, I got this error;

Screenshot 2023-05-05 at 14 17 55

After that I tried to add  the --force param, but that gave a really long error as well.

vcanales (commented):
I can successfully install dependencies after running `nvm use`, but for some reason, even with no `node_modules/` present, the package-log.json file is always updated.

Running `npm ci` fails if I don't run `npm i` first, which tells me that package and package-lock might not be in sync

mikachan:
> Then after upgrading, I got this error;

That's the error I was seeing that I couldn't get past. I also had to update my version of Xcode, I wonder if you need to do the same?

> but for some reason, even with no node_modules/ present, the package-log.json file is always updated.

Hmm.. did you also delete `node_modules` in the subdirectories? I'm not sure this would make a difference though.

I'm using node v18.16.0 and npm v9.5.1. If I run `rm -rf node_modules`, and then run `npm ci`, I don't see any updates to the package-lock file, and everything installs correctly. Could it maybe be a version mismatch, and that's what's causing the differences in the package-lock file?

jeffikus:
@mikachan updated Xcode, ran both commands above, this was the output error; 

Screenshot 2023-05-08 at 08 56 06

mikachan:
> updated Xcode, ran both commands above, this was the output error;

From the error, it looks like you're using node v20.1.0, you need to be using node v18 with this PR.

mikachan:
Sorry both, I think one of my personal themes had been caught up in these updates. One of them includes the tailwindcss package and I _think_ this is what was causing the differences in the package-lock file 🤦‍♀️

I've removed these references, re-ran the commands in the testing instructions and now I think this should be good for another review. I've also updated a few more deps that have had new releases in the past few days.

vcanales (commented):
Getting this after running `npm ci`. I ran `nvm use` right before to ensure I was on the correct version.

image

`npm i` works, but it updates package-lock like before. Here's a Gist with the package-lock.json file that is generated on my machine, in case you'd like to test with it, maybe it works. https://gist.github.com/vcanales/8d3de4fab511dd2d07eace0816a083fa

jeffikus (commented):
LGTM, but before I approve it it does modify my package files. It eventually worked after switching to v18. I had some audit items, but fixed those too. 1 thing to note is that running `npm ci` took super long this time and I ran `npm audit fix` as well. Perhaps that's why the package files are now modified?

Screenshot 2023-05-11 at 11 48 44

mikachan:
Hmm.. the PR should include updates from running `npm audit fix`, as I ran this too. Super frustrating that we're seeing different package-lock files!

> Here's a Gist with the package-lock.json file that is generated on my machine, in case you'd like to test with it, maybe it works. https://gist.github.com/vcanales/8d3de4fab511dd2d07eace0816a083fa

Ooh this is helpful, thank you! I downloaded this and compared it to the package-lock in this diff, and I noticed that most of the differences were dependencies in subdirectories. I didn't update any of these as part of this, so I've now run through all subdirectories (using our scripts `batch:update` and `batch:audit:fix`), and now the package-lock here seems to include them! Hoping this will fix up the differences we're seeing! 🤞

It's probably worth noting that the package-lock file on trunk doesn't include these subdirectory dependencies either, but I think it probably should, as it should mean all the dependency versions are consistent.

Thank you, both, for helping test this.

jeffikus:
@mikachan it worked great until I ran `npm audit fix` - then my package lock showed changes. Do you need them or shall we move this to approved?

mikachan:
Gah really?! I've just tried running `npm audit fix` again and I'm not seeing any changes. It's probably fine to bring this in, we can always merge any further changes to the package-lock if needed / if it happens consistently. (But also happy to keep testing in this PR if needed.)

jeffikus:
@vcanales @pbking can you please give this 1 more test? If it works for both of you I'm happy to approve and merge this.

vcanales (approved):
It's still updating the lock file for me, but it could very well be something on my end. Also, we don't use `npm ci` in any of the automation tools, so we can fix it later.

I recommend a full `rm -rf node_modules && nvm use && npm install` to get this going on your systems :)

pbking:
`npm ci` worked fine for me.
Also my `package-lock.json` file also gets updated but... that happens a lot anyway and I wouldn't consider that problematic at all.   I'll just post-approve approve this change. :)

mikachan:
Thanks all! I'll bring this in then 🚢

PR #7075, opened 2023-05-12
Fix core theme deploy utility

pbking:
The part of the core deploy process that lands the change, deploys the theme and builds the zip was commented out.  It works fine so it was probably comment for historical debugging purposes and shouldn't have been comitted.


madhusudhand (approved):
LGTM!

PR #7088, opened 2023-05-19
Karuna: remove jQuery

sgomes:
This PR reimplements the small theme script in Karuna to use vanilla JS, dropping the `jQuery` dependency.

- Reimplement Karuna script in vanilla JS
- Remove `jQuery` dependency when enqueuing

#### Related issue(s):
See D111363-code

### DIFF
--- a/karuna/assets/js/functions.js
+++ b/karuna/assets/js/functions.js
@@ -2,40 +2,52 @@
  * Handles sticky header
  */
 
-( function( $ ) {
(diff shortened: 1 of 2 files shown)


vcanales (approved):
Sticky header behavior, which looks like the feature that should be affected by this change, tests well for me.

sgomes:
Thank you for the review, @vcanales! 👍

PR #7096, opened 2023-05-23
Iotix: make hero image much smaller

sgomes:
Iotix uses a gradient with a circular highlight as a hero image. This is a very large PNG, coming in at 741KB.

One strategy to partially solve this would be to use a JPEG instead, since we have no need for transparency. However, since the goal is simply to obtain a smooth gradient, we can instead start with a much smaller image and take advantage of browser upscaling to produce a smooth result.

This PR replaces the 741KB, 2880x1716 PNG, with a 678 *Byte*, 29x17 PNG. That's over *1000 times smaller*, while producing a better gradient, since it gets rid of the colour banding in the original.

Original:
image

This PR:
image

- Replace the hero image with a much smaller version of itself


jffng (approved):
Thanks @sgomes !

sgomes:
Thank you for the review and merge, @jffng! 👍

PR #7098, opened 2023-05-23
Lettre: Update spacer in archive header

### LINKED ISSUE #7069: Lettre: Update inline space from 200px to 100px
Via p9Jlb4-787-p2

@crisbusquets wrote:

> Right now, the div with `.wp-block-spacer` has an inline height of 200px. This seems to work well on big screens, but on mobile it takes up too much space.
>
> I would suggest reducing it a bit, maybe 100px, so that the reader can read more content without scrolling. I’m aware that our customers can modify this in the Editor, but perhaps we can adjust this in the basic Lettre theme.
> 
> Here’s a comparison:
> 
> Image
> 
> Image
> 
> And mobile:
> 
> Image
> 
> Image

edanzer:

This PR reduces the spacer height from 200px to 100px on the achive header template part.

Note: This issue started with a desire to reduce spacing above the post title on single posts. Not obviously that a single post template would be using a header called 'archive header'. 

#### Related issue(s):

Resolves https://github.com/Automattic/themes/issues/7069

#### Testing Instructions:

1) Load the Lettre theme.
2) Pull up a single post on the frontend and confirm the post loads without issue and the space above the post title is reduced as expected.
3) Go to Site Editor > Templates > Header Archive template part and confirm the title block now shows and there are no other issues.

### DIFF
--- a/lettre/parts/header-archive.html
+++ b/lettre/parts/header-archive.html
@@ -6,8 +6,8 @@
 <!-- wp:navigation {"overlayMenu":"always","overlayBackgroundColor":"background","overlayTextColor":"foreground","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right"},"style":{"typography":{"fontStyle":"normal"}},"fontSize":"x-large"} /--></div>
 <!-- /wp:group -->
 
-<!-- wp:spacer {"height":"200px"} -->
-<div style="height:200px" aria-hidden="true" class="wp-block-spacer"></div>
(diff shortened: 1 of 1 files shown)


simison (approved):
Tests well!

**Before**

Screenshot 2023-05-24 at 13 16 45

Screenshot 2023-05-24 at 13 16 54


**After**
Screenshot 2023-05-24 at 13 17 08

Screenshot 2023-05-24 at 13 17 02

PR #7119, opened 2023-05-31
Fix layout issues in course patterns

gabrielcaires:
* Remove spacing between the header and the main tag
* Remove uppercase applied on all title
* Refactor header to use min-height instead of padding to control the height 
* Fix buttons centralized by the center.

#### Related issue(s):
[Check spacing and alignment in the Course theme#6932](https://github.com/Automattic/themes/issues/7121)

### DIFF
--- a/course/style.css
+++ b/course/style.css
@@ -120,6 +120,10 @@ header {
 header .navigation-wrapper {
 	min-height: 116px;
 }
+
+h1, h2, h3, h4, h5, h6 {
+	text-transform: revert;
+} 
 /*
  * Comments
  */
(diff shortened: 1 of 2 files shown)


course/style.css
```diff
@@ -266,9 +277,6 @@ a {
 	text-decoration: none;
 }
 
-.wp-block-button {
-	text-align: center;
```
merkushin, line 270: Not really sure about this change. It shouldn't affect button's position, but the position of text inside a container.
And as we can see [here](https://github.com/Automattic/sensei/pull/6938#issuecomment-1570929148), buttons remain aligned by center.

merkushin: Okay, I see. Those buttons are anchors, and this change affects them.
But for some reason I don't see that change.
Maybe a cache issue :(

merkushin: Yes, it was cache :(

PR #7128, opened 2023-06-08
Fotograma: Add theme

iamtakashi:
Fotograma: Add theme

fotograma-main-2


vcanales (approved):

iamtakashi:
Thanks for the review @vcanales !

I think you meant [the title of the pattern](https://github.com/Automattic/themes/blob/trunk/fotograma/patterns/header.php#L3) `header`.

Now that we have the header pattern, shall we include the pattern in [the header template part file](https://github.com/Automattic/themes/blob/trunk/fotograma/parts/header.html) like the footer?

`<!-- wp:pattern {"slug":"fotograma/header"} /-->`

PR #7130, opened 2023-06-08
Bute: add theme

iamtakashi:
Bute: add theme

bute-main


madhusudhand:
Tested and Fixed the issues 🌟 

@iamtakashi Only UI issue observed was the color of social icons.

image

Can you confirm if we can change it to same color as site-title and menu?

iamtakashi:
> Can you confirm if we can change it to same color as site-title and menu?

@madhusudhand Yes, cice catch again! I'm not sure how I missed it. it's weird that they appear in white when `"iconColorValue":"#000000"` But #000000 isn't the right colour anyway. Please change it to `contrast` which is `#191919` Thanks in advance!

madhusudhand (approved):
Updated social icon color value to `contrast`.

Theme is now ready to 🚢

PR #7132, opened 2023-06-08
Al Dente: Add theme

iamtakashi:
Al Dente: Add theme

al-dente-device-1


vcanales (approved):
LGTM, only with the observation that the Subscribe Block won't be available for all users:

image

This hasn't stopped other themes, such as Bitácora.

PR #7138, opened 2023-06-13
Update George Lois theme default styles to dark

madhusudhand:
#### Changes proposed in this Pull Request:

Update George Lois default styles to dark. Light version is offered through style variation.

image

#### Related issue(s):

Fixes: https://github.com/Automattic/wp-calypso/issues/78104

PR #7139, opened 2023-06-19
Revert "update george-lois default styles to dark (#7138)"

vcanales:
This reverts commit 3fcd94eeb731a6f4b0392752e97512ff9c93e7b9 introduced in #7138 due to concerns with the impact for users who have already activated the theme.

Also: 
- Adds a Dark style as a variation instead of default.
- Removes light.json — I tested and this doesn’t impact the experience. If they user has explicitly set the Light variation, they will be shown the Default instead, which is the same visually.


madhusudhand (approved):
Tested the change. It works as intended!

PR #7145, opened 2023-06-21
Creatio: Add theme

alaczek:
Creatio is meant to be the new base theme used in Pattern Assembler (see: pekYwv-1Yg-p2). It is based on the refreshed version of Shawburn (pNEWy-gqF-p2).


creatio/theme.json
```diff
@@ -0,0 +1,728 @@
...
+					"color": "#f7f9fd",
+					"name": "Tertiary",
+					"slug": "tertiary"
+				}
+			],
+			"text": true
+		},
+		"layout": {
+			"contentSize": "620px",
+			"definitions": {
```
pbking: I think these "definitions" are added by Gutenberg right now (and shouldn't be).  I think they should be removed.  


... or maybe I just don't understand the purpose because my brain is out of date?

pbking (approved):

autumnfjeld:
cc: @Automattic/lego

PR #7163, opened 2023-06-27
Bibliophile: add theme

madhusudhand:

Adds the new theme `bibliophile` designed by @beafialho 

image


madhusudhand:
Fixed the general issues.

@beafialho The post-meta shows `By <categores>` as follows

image

The prefix `BY` suggests it to be author name instead of categories. Can you confirm?

beafialho:
Yes @madhusudhand, that's correct. Good catch. 

I set it up as categories because I wanted to have different "authors" in the demo, and naming them as categories was the only way to do that.

Captura de ecrã 2023-06-28, às 10 36 15

beafialho:
The only thing is that I don't remember if I set Post Author to look exactly like the Categories, could you confirm?

madhusudhand:
There isn't a post-author-name definition in any of the templates, but if I place one, yes it looks exactly like categories.

image

Also, in the `single` template, there isn't a comments block. is it intentional?

beafialho:
Ok, it looks as intended. Thanks for checking.

> Also, in the single template, there isn't a comments block. is it intentional?

Yes, that's intentional.

madhusudhand:
Ready to 🚢

PR #7165, opened 2023-06-27
Bugfix: replaced bitacora with tu where needed.

nvourva:
Due to typo bitacora_support was hooked to after_sebitacorap_theme.

Replacing instances of the string bitacora with tu in lines 15, 19, 32 & 41.
In the above mentioned lines the string tu was replaced with the theme's slug thus disabling the after_setup_theme action by turning it to after_sebitacorap_theme resulting in the editor style not being added and the theme's text domain not being loaded properly.

This PR replaces the string bitacora with tu where needed to mitigate the issue.

### DIFF
--- a/bitacora/functions.php
+++ b/bitacora/functions.php
@@ -12,11 +12,11 @@
 if ( ! function_exists( 'bitacora_support' ) ) :
 
 	/**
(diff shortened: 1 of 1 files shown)


vcanales (approved):
Nice catch, thanks for the PR!

Issue #7072, opened 2023-05-10
Lettre: Fix weird alignment on mobile


simison:
Via p9Jlb4-787-p2

@crisbusquets wrote:

> Mobile navigation
>
> While the navigation is working on desktop, on mobile it has a weird alignmement —every third element seems to have different CSS (you can check this in [my demo site](https://href.li/?https://cbusquets1989.wordpress.com/)). Or maybe it’s just the navigation elements that are too long.
>
> ![Image](https://github.com/Automattic/themes/assets/87168/bbaffe74-e80b-48fb-a13c-329f65409d1c)
>
> A smaller font-size seems to do the trick, and it’s also easier to read:
>
> ![Image](https://github.com/Automattic/themes/assets/87168/6779f0e5-d2a8-4323-a6b0-4951e7d69065)

simison:
@crisbusquets would this work better?

- Right-aligned the text
- Tightened line-height

<img width="464" alt="Screenshot 2023-06-07 at 14 04 48" src="https://github.com/Automattic/themes/assets/87168/d4f433f2-c6b1-4122-8835-83c51e9c6039">

Another option would be to adjust font-size for mobile, as it was the same as on desktop:

<img width="509" alt="Screenshot 2023-06-07 at 14 06 59" src="https://github.com/Automattic/themes/assets/87168/ba1d97ea-eae5-4947-9ebd-673ecad2e0b9">

Lemme know which you like better or if there's another solution.

crisbusquets:
Thanks for sharing, @simison !

I prefer the second one, it's easier to read.
On mobile devices, we don't need fonts as big as the first screenshot.

edanzer:
This was resolved in https://github.com/Automattic/themes/pull/7167

Issue #7071, opened 2023-05-10
Lettre: Respect “current” class on desktop


simison:
Via p9Jlb4-787-p2

@crisbusquets wrote:

> General nav
>
> On desktop, the `current` class doesn’t seem to work properly. The first item is always underlined, no matter if I’m on a post or on a different page.
>
>  https://github.com/Automattic/themes/assets/87168/ad54915c-758b-41f5-bc4e-557e581488d8

edanzer:
@crisbusquets @simison - This issue also traces back to a core block, in this case the navigation block. I can duplicate the issue on a vanilla WordPress site with the navigation block.

The issue is that as soon as you open the menu, the first item in the menu has 'focus', and [there's a css rule](https://github.com/WordPress/gutenberg/blob/97ab6b167cc0d9da6c5f0ecf6be9457570d33731/packages/block-library/src/navigation/style.scss#L65) that applies underline to any element with focus.

Is it worth me trying to open a PR in Gutenberg for the nav block for this? That would take a bit of learning, but I wouldn't mind getting up and running for Gutenberg. I do think the PR process is slower there. 

Other thoughts?

simison:
at least opening, and checking if there is one already, a bug report in Gutenberg would be good

I think there's a team focusing on navigation block specifically

simison:
also if you wanna fix a core block, go for it! :-)

crisbusquets:
> The issue is that as soon as you open the menu, the first item in the menu has 'focus', and [there's a css rule](https://github.com/WordPress/gutenberg/blob/97ab6b167cc0d9da6c5f0ecf6be9457570d33731/packages/block-library/src/navigation/style.scss#L65) that applies underline to any element with focus.

I'm not sure I understood this correctly: the `focus` overrides `current` (or the other way around)?

+1 to @simison, if that's the way to fix it 🙏🏼

Thank you @edanzer !

edanzer:
@crisbusquets @simison - Long over due, but I opened an issue to start a conversation about this on the Gutenberg repo here: https://github.com/WordPress/gutenberg/issues/51982

@crisbusquets - to your question above, :focus is overriding current. Indeed, from what I can tell, the block isn't applying a current class to the nav menu item for the current page. But they're definitely setting the focus to the first item on the menu each time user opens the mobile menu. 

[...]

PR #7167, opened 2023-06-28
Lettre: Fix styling issues for top nav menu

edanzer:

* Reduces the font size of items in the nave menu. Among other things this fixes an issue with text for medium-to-longer page titles wrapping on small screens.
* Ensure wrapping text doesn't align left. 
* Remove underline from :focus items. This was producing a situation where the first item in the menu was always underlined, even when it was not current. 
* Add underline to current item. The current item in the menu will now be underlined. 

**Before**
lettre nav before

**After**
lettre-nav-after

#### Related issue(s):
https://github.com/Automattic/themes/issues/7071
https://github.com/Automattic/themes/issues/7072

#### Testing instructions
1) Checkout this branch to your local WordPress instance.
2) Ensure you have multiple items on your menu, including one page with a long page title.
3) Confirm: 
   - First item isn't always underlined by default
   - Current page is underlined
   - Font size is smaller
   - If longer item does wrap, it is still aligned right.

### DIFF
--- a/lettre/style.css
+++ b/lettre/style.css
@@ -49,7 +49,7 @@ a {
 .wp-block-search__button:hover,
 .wp-block-file .wp-block-file__button:hover,
 .wp-block-button__link:hover {
-	opacity: 0.90;
+	opacity: 0.9;
 }
 
 /*
@@ -128,7 +128,7 @@ input:not([type="submit"]):not([type="button"]):focus {
  * Comment form refinements.
  */
 .comment-form .comment-subscription-form,
-.comment-form .post-subscription-form{
+.comment-form .post-subscription-form {
 	margin: 0;
 }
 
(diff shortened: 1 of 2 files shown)


lezama (approved):

lettre/style.css
```diff
@@ -138,10 +138,29 @@ input:not([type="submit"]):not([type="button"]):focus {
...
\ No newline at end of file
+}
+
+/*
+ * Refine mobile navigation styling.
+ */
+.wp-block-navigation a:where(:not(.wp-element-button)):focus,
+.wp-block-pages-list__item:focus.wp-block-navigation a:where(:not(.wp-element-button)):focus {
+	text-decoration: none;
+}
```
simison, line 155: Removing focus-underline so that wrong link isn't highlighted as "current" element could impact browsing the navigation with keyboard? Also weird how first element was focused when the navigation opens?

cc @scruffian hi! Is this something that should be improved in Gutenberg? (See PR description for screenshots)

PR #7168, opened 2023-06-28
Tomoni: replace with noto JP with variable font

jffng:
This PR replaces the otf files with the variable font face, which is 67% smaller.

### DIFF
--- a/tomoni/theme.json
+++ b/tomoni/theme.json
@@ -249,7 +249,7 @@
 							]
 						}
 					],
-					"fontFamily": "Noto Sans, Noto Sans JP",
+					"fontFamily": "Noto Sans",
 					"slug": "noto-sans"
 				},
 				{
--- a/tomoni/theme.json
+++ b/tomoni/theme.json
@@ -105,60 +105,6 @@
 				{
 					"fontFace": [
 						{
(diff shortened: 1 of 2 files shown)


tomoni/theme.json
```diff
@@ -303,8 +249,22 @@
 							]
 						}
 					],
-					"fontFamily": "Noto Sans, Noto Sans JP",
```
iamtakashi, line 306: I think we need to keep this reference to Noto Sans JP. Otherwise, Noto Sans JP will never be used as `var(--wp--preset--font-family--noto-sans)` is just `Noto Sans`.

jffng: You're right! I restored it: [25abb58](https://github.com/Automattic/themes/pull/7168/commits/25abb58b6686023682c35771bae4b211faf3327e)

For context, I removed it because I thought it was strange to have the fontFamily listed like this: 

Screenshot 2023-06-29 at 12 51 12 PM

But I agree it's better to have the fallback available, so you don't have set a different fontFamily when using English and Japanese together.

iamtakashi: Oh, I see. It's a little weird, but it's also not too bad either.

Thanks for the performance improvement!

PR #7169, opened 2023-06-29
MENU: Rename menu theme to the-menu

madhusudhand:

Menu theme name is not available in dot-org. Renaming to the-menu


madhusudhand:
Tested the theme. It works great!
ready to 🚢

PR #7173, opened 2023-06-29
Luminance: Add theme

iamtakashi:

Luminance add theme

Luminance


mikachan (approved):
Great theme! I really like the typography. Tests well for me 🎉

iamtakashi:
@mikachan I'm unsure if it was an error, but what did you mean by assigning this to me?

Regardless, I think I need to adjust the theme because of the feedback I've received, so hold on merging this for a bit, please.

mikachan:
> I'm unsure if it was an error, but what did you mean by assigning this to me?

I assigned it to you because you opened the PR, and on the Theme Launcher it initially looks like there's no associated PR when there is no user assigned. But I didn't mean anything else, e.g. for you to review. I probably should've assigned myself as I always intended to review!

Will hold off merging for now 👍

iamtakashi:
@mikachan I made a minor adjustment, and if that looks ok to you, merge and deploy it on Monday or whenever you have time. Thank you!

mikachan:
Looks good to me! Just one note - with the addition of `text-wrap: balance;` to the post titles, should we also add a `max-inline-size` size? It's suggested in [this post](https://developer.chrome.com/blog/css-text-wrap-balance/) to use these two properties together.

iamtakashi:
Oh, I didn't read about `max-inline-size`, thanks! But I'm struggling to set a value for it. Because it's a "fluid" layout and we won't be sure of the length of the title, Any arbitrary number for it could lead to an unexpected gap like this.

Screenshot 2023-07-11 at 13 58 51

This is pretty new for me so I might be making a wrong assumption. What do you think?

mikachan:
> This is pretty new for me so I might be making a wrong assumption. What do you think?

It's new for me too 😅 It's a pretty cool feature. I've just had a play around with different `max-inline-size` values and I'm not sure that we need it here. Like you say, it's a fluid layout so I think a good value to use is `100%`, which seems to be the default.

Also, I can see `text-wrap: balance;` is improving the word balance on its own, so I think this is good as it is. Sorry for confusing things, although it was fun to learn more about this!

I'll merge and deploy - we can always revisit if needed!

PR #7186, opened 2023-07-04
Fewer: add theme

matiasbenedetto:
Adding fewer theme created by  @beafialho


matiasbenedetto (commented):
**Changes I made from the original exported zip:**
- Remove not needed CSS (GB PR already merged).
- Bump version from 0.0.1 to 1.0.0
- Remove the unwanted or not needed parameters for the query block.
- Deleted the patterns folder because none of the patterns was being used.
- Deleted the images folder because non of the was being used in a template.

To launch in dotOrg we would need to include the screenshot images credits. cc. @beafialho 

Apart from that everything seems to be working as expected.

I'm merging this.

PR #7188, opened 2023-07-04
Exmoor: Fix footer credit so it will be replaced correctly for WP.com

iamtakashi:
Initially, I wanted to make the footer credit span two lines, but that prevented it from being replaced correctly for the WP.com-specific credit.

**before**
Screenshot 2023-07-04 at 15 20 06

**after**
Screenshot 2023-07-04 at 15 22 01


mikachan (approved):
Looks good to me 👍

PR #7197, opened 2023-07-05
Removed unnecessary and unexpected styles from heading element

pbking:
The font 'rubik' is not installed by the font and not expected to be used.  The color defined is the default and unneeded to be set.

This change removes those settings completely.

There are no visual changes unless Rubik font is available (for example via Jetpack fonts, etc)


mikachan (approved):
Makes sense!

Issue #7172, opened 2023-06-29
Ueno: Add missing info and files to the theme


beafialho:
The following feedback in [WordPress.org](https://themes.trac.wordpress.org/ticket/143821) is blocking Ueno's launch:

> License of the images used in the screenshot and the bundled image in the theme is missing.
Font License is missing.
WARNING: assets/images/ueno_1.png is 2.0 MB in size. Large file sizes have a negative impact on website performance and loading time. Compress images before using them.
WARNING: assets/images/ueno_2.png is 1.9 MB in size. Large file sizes have a negative impact on website performance and loading time. Compress images before using them.
WARNING: screenshot.png is 916.1 KB in size. Large file sizes have a negative impact on website performance and loading time. Compress images before using them.

**Image licenses**
https://publicdomainreview.org/collection/unkindo-kimono-designs

**Font**
[IBM Plex Mono](https://fonts.google.com/specimen/IBM+Plex+Mono?query=ibm+plex+mono)

**Compressed images and screenshot**
![ueno_1](https://github.com/Automattic/themes/assets/65220155/bb0162d8-98aa-433d-9527-7702b32d2f74)
![ueno_3](https://github.com/Automattic/themes/assets/65220155/c62fe140-aa67-42c7-99bc-3a0d794e468b)
![ueno_2](https://github.com/Automattic/themes/assets/65220155/bc20db8f-1112-48d7-aed1-0de32720df5a)
![screenshot](https://github.com/Automattic/themes/assets/65220155/c3977b08-59dd-4e18-8d41-c6edc4c6555e)

PR #7198, opened 2023-07-05
Ueno: update assets and their license

vcanales:
I've addressed the problems raised during the theme's dotOrg submission. In the process, I noticed that the image assets included are not used within the theme other than in the screenshot. Is this correct @beafialho? If so, I can remove them. 

Compress images
Add licensing information for images and font.

#### Related issue(s):
#7172


beafialho:
Thanks @vcanales, these images are used for posts and the about page.

PR #7199, opened 2023-07-05
Add the `style-variations` tag to themes with style variations

mikachan:
Adds the `style-variations` tag to existing themes that include style variations.

More info on the tag here: https://meta.trac.wordpress.org/ticket/6545


vcanales (approved):
Good catch!

PR #7203, opened 2023-07-06
Bute: Update navigation so they are vertically centre to avoid wpcom banners overlapping

iamtakashi:

This theme suggests having an "always-icon" navigation. 
butedemo wordpress com_(MacBook Pro 14_) (1)

When you open the navigation, the demo or the marketing bars overlap the navigation, and the close button gets hidden.
butedemo wordpress com_(MacBook Pro 14_) (2)

Usually, this is not a problem because of the z-index of `100000` while the banners are `99998` and `99998`. However, when a header is inside a cover block, I don't think the z-index is actually `100000`, thus, it goes underneath the banners :/

I thought it'd be best to make the navigation vertically centre to avoid that. There is no option to modify the layout inside the modal, so I needed to use css though.

localhost local_(MacBook Pro 14_) (2)

What do you think? Any other solution that we don't need to hold up the launch?

### DIFF
--- a/bute/theme.json
+++ b/bute/theme.json
@@ -724,7 +724,7 @@
 				}
 			},
 			"core/navigation": {
-				"css": " .wp-block-navigation__responsive-container.is-menu-open{justify-content:center; .wp-block-navigation__responsive-container-close{right:50%;transform:translate(50%, 0%)}}",
+				"css": " .wp-block-navigation__responsive-container.is-menu-open{justify-content:center;}}",
 				"elements": {
 					"link": {
 						"typography": {
--- a/bute/theme.json
+++ b/bute/theme.json
@@ -724,7 +724,7 @@
 				}
 			},
 			"core/navigation": {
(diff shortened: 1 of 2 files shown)


bute/theme.json
```diff
@@ -724,6 +724,7 @@
 				}
 			},
 			"core/navigation": {
+				"css": " .wp-block-navigation__responsive-container.is-menu-open{justify-content:center;}}",
```
mikachan:
```suggestion
				"css": " .wp-block-navigation__responsive-container.is-menu-open{justify-content:center;}",
```

I think there's an accidental extra bracket here.

mikachan (approved):
It'd be good to fix this upstream in Gutenberg, but for now, we can change the alignment of the menu in this theme so the launch isn't blocked. Apart from a tiny change, the CSS looks good to me 👍

iamtakashi:
Thanks @mikachan!

I opened an issue here: https://github.com/WordPress/gutenberg/issues/52511

PR #7214, opened 2023-07-07
Block Canvas: Remove patterns

matiasbenedetto:
 I moved the content of the patterns to the templates and removed the patterns from Block Canvas theme.

### Why?
Block Canvas is used as a theme starter to create other themes. Usually, the content of the patterns is overwritten in the templates, and the patterns are no longer used. When the new theme is exported the patterns are included but have nothing to do with the theme created.


mikachan:
@iamtakashi @beafialho @henriqueiamarino Do you still use Block Canvas as a base theme, and would these changes improve your theme workflow?

beafialho:
I still use Block Canvas as a base theme, yes. If removing the patterns will simplify the process and files, I think it's a good change.

iamtakashi:
I don't use Block Canvas any more, but this is a good idea for a base theme 👍👍👍

henriqueiamarino:
Just like @beafialho, I still use it, and if these changes make things better, I am OK with them.

mikachan:
Thanks for the feedback, all! In that case, I think we should bring this in.

mikachan (approved):

PR #7217, opened 2023-07-07
Adding Theme: Dawson

pbking:
Adding Dawson to Free Themes Repo


pbking:
This theme doesn't have a footer.  Is that a requirement for the free themes?

pbking:
image

@beafialho I added a super-simple footer to the site (a right-aligned link to wordpress.org).  

I'm pretty sure that all the themes in this collection need a footer, but if that turns out to be not true then there's no other reason I suggest it.

Please let me know if that addition is OK.  I believe this is good to bring in otherwise.

beafialho:
Thanks @pbking, this is for WordPress.com? If that's a requirement, let's just align it to the left side and set font size to `0.7rem`:

Captura de ecrã 2023-07-13, às 16 20 16

For WordPress.org I don't find the footer requirement in the documentation.

pbking:
No, not a requirement for dotorg at all.  I think it's something we may need to ship for wpcom though.  I can't find an example of a theme without one.

beafialho:
Yes, for Dotcom I believe it just adds footer credit automatically.

iamtakashi:
I think WP.com _swap_ it with the variations suitable for WP.com. I haven't tested it, but the swapping mechanism (regex) would probably fail if there is no text in the footer.

pbking:
Yes, that's how the code works (or at least it did last I was in it).  It swaps out based on looking for either a wordpress.com or a wordpress.org link.

I've changed the footer to match the design given; thank you.

PR #7229, opened 2023-07-12
Update Creatio

alaczek:
This is a bulk PR with discreet commits for the following changes:
* update theme screenshot & add image credits
* turn single post byline into a pattern
* adjust the post title in archive/index templates
* adjust the line-height of headings


madhusudhand (commented):
Screenshot Image size is bit large. Please compress the image.

Everything else looks good!

PR #7237, opened 2023-07-13
Creatio: Bulk fixes

alaczek:
* Change the layout setting on the post content to all full-width/wide blocks inside post/page content.
* Add top padding to content in index template (required after removing gaps between header/content/footer, but also it felt a bit tight
* Update footer pattern - remove the space at the bottom.
* Remove blockGap between header/sections/footer for Pattern Assembler (special case, see https://github.com/Automattic/wp-calypso/issues/78097 for reference).

### DIFF
--- a/creatio/style.css
+++ b/creatio/style.css
@@ -7,11 +7,11 @@ Description: Creatio is a simple, minimal theme that supports full-site editing 
 Requires at least: 6.0
 Tested up to: 6.2.3
 Requires PHP: 5.7
-Version: 0.0.2
+Version: 0.0.1
 License: GNU General Public License v2 or later
 License URI: http://www.gnu.org/licenses/gpl-2.0.html
 Text Domain: creatio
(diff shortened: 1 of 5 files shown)


creatio/style.css
```diff
@@ -7,11 +7,11 @@ Description: Creatio is a simple, minimal theme that supports full-site editing
 Requires at least: 6.0
 Tested up to: 6.2.3
 Requires PHP: 5.7
-Version: 0.0.2
+Version: 0.0.1
```
madhusudhand: Let's mark it as `1.0.0`?

creatio/style.css
```diff
@@ -7,11 +7,11 @@ Description: Creatio is a simple, minimal theme that supports full-site editing
 Requires at least: 6.0
 Tested up to: 6.2.3
```
madhusudhand: dotOrg suggest to have only upto minor version. Update it to `6.2`.

Also, theme uses fluid typography. Means the required min version should be `6.1`
Update the same in `readme.txt`

madhusudhand (approved):
LGTM!

creatio/style.css
alaczek: Done!

creatio/style.css
alaczek: Done!

PR #7242, opened 2023-07-13
Adding Raw theme from @henriqueiamarino

pbking:
image


raw/patterns/home.php
```diff
@@ -0,0 +1,165 @@
...
+<!-- /wp:column --></div>
+<!-- /wp:columns -->
+
+<!-- wp:spacer {"height":"4rem","style":{"spacing":{"margin":{"top":"0px","bottom":"0px"}}}} -->
+<div style="margin-top:0px;margin-bottom:0px;height:4rem" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- /wp:spacer -->
+
+<!-- wp:columns {"align":"wide"} -->
+<div class="wp-block-columns alignwide"><!-- wp:column -->
+<div class="wp-block-column"><!-- wp:block {"ref":307} /--></div>
```
pbking: There are a few of these `wp:block` elements in the templates.  I'm guessing those are... reusable blocks?  I can't say I've seen them in templates before.

If so, are those reusable blocks that you would like available to users as patterns?

pbking: image
This is what I see from the editor.

pbking: Confirmed that was the case.

It will be grand for CBT to export these (or even better, for Gutenberg to manage this for themes).  

Until then I just copied them out to theme patterns.

PR #7253, opened 2023-07-21
Nested: Cleanup query property

iamtakashi:
These seem to be overlooked during the review, and it's causing an issue in the Design Picker in WP.com. More info: p7DVsv-i3x-p2#comment-46764

cc: @henriqueiamarino


pbking (approved):
Looks like that got them all.  Change works as expected.  Thanks @madhusudhand for also fixing the metadata.

pbking:
🚢 

Thanks for fixing @iamtakashi

PR #7256, opened 2023-07-21
Nested: Fix cover blocks on homepage

iamtakashi:
As per p7DVsv-i3x-p2#comment-46770, this intends to fix the cover blocks on the homepage as we see in the demo site.


madhusudhand (approved):
Thanks for fixing the issue. 
Ready to 🚢

PR #7262, opened 2023-07-25
Montagna batch updates

alaczek:
This ended up being a pretty big overhaul. After looking over the design, there were many issues that I wanted to address, mainly with spacing and typography, which then required updating all the templates. I also added a 404, and no-results patterns, and changed layout of single page template to two columns.


pbking (approved):
Other than a few minor fixes I made the changes look good.  

🚢

PR #7264, opened 2023-07-25
Nested: Fix the excerpt colour

iamtakashi:
I've noticed the excerpt colour is left not styled as default. That should be white as default, like the heading below.

Before
localhost local_ (17)

After
localhost local_ (16)


mikachan (approved):
LGTM 👍

PR #7265, opened 2023-07-25
Add course theme single lesson template

Imran92:
Lesson is a Custom Post Type in sensei. So, the lessons were supposed to use the `Single` template. But as Sensei forced lessons to be rendered using the `Page` template instead, though it showed in the editor that Lesson is using the Single template, in reality, it was using the Page template of Course theme.

Now that we've fixed the issue for FSE themes in sensei https://github.com/Automattic/sensei/pull/7045, it'll no longer force the Page template and it'll start using the Single template.

But the Single template for post in Course theme is a two column design, so we'll get only less than 60% of the page's real estate for the lesson content to show, which doesn't look very good, and also it'll be a sudden unintended side effect.

To solve that, we're adding a template only for the lesson post type which is an exact copy of the page template. So, now even after the change in core sensei, lessons in Course theme will still look as they were before. With the ability for the user to edit the template as they wish unlike before.

#### Related issue(s):
https://github.com/Automattic/sensei/pull/7045

### DIFF
--- a/course/functions.php
+++ b/course/functions.php
@@ -92,6 +92,17 @@ add_filter(
 	3
 );
 
+/**
+ * Filter the list of templates for the single lesson page.
+ *
+ * @param array   $page_templates Array of page templates.
+ * @param string  $theme          The current theme.
+ * @param WP_Post $post           The post being edited, provided for context, or null.
+ *
+ * @since Course 1.3.1
+ *
+ * @return array Array of page templates.
+ */
 function course_theme_filter_single_lesson_template_for_sensei_learning_mode( $page_templates, $theme, $post ) {
 	// In case some other plugin has a post type called lesson.
 	if ( ! $post || ! class_exists( 'Sensei_Main' ) ) {
--- a/course/functions.php
+++ b/course/functions.php
@@ -87,23 +87,24 @@ add_action( 'init', 'course_register_block_patterns_category' );
 
 add_filter(
 	'theme_lesson_templates',
-	function( $page_templates, $theme, $post ) {
-
-		// In case some other plugin has a post type called lesson.
-		if ( ! $post || ! class_exists( 'Sensei_Main' ) ) {
-			return $page_templates;
-		}
-
-		$course_id                = Sensei()->lesson->get_course_id( $post->ID );
-		$is_learning_mode_enabled = Sensei_Course_Theme_Option::has_learning_mode_enabled( $course_id );
-
-		if ( $is_learning_mode_enabled ) {
-			unset( $page_templates['single-lesson'] );
-		}
-
-		return $page_templates;
-	},
+	'course_theme_filter_single_lesson_template_for_sensei_learning_mode',
 	11,
 	3
 );
+
+function course_theme_filter_single_lesson_template_for_sensei_learning_mode( $page_templates, $theme, $post ) {
+	// In case some other plugin has a post type called lesson.
+	if ( ! $post || ! class_exists( 'Sensei_Main' ) ) {
+		return $page_templates;
+	}
+
+	$course_id                = Sensei()->lesson->get_course_id( $post->ID );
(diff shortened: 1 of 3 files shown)


course/theme.json
```diff
@@ -82,6 +82,13 @@
 				"post",
 				"course"
 			]
+		},
+		{
+			"name": "single-lesson",
+			"title": "Single Lesson",
```
donnapep: Let's change the title of this one to "Lesson" instead of "Single Lesson" to better match how Learning Mode templates are named.

donnapep:
Nice find!

I know this repo doesn't prompt for testing instructions, but we should always be sure to include those so people know what's expected of them. 🙂 

Here's what I did:
- Switched to [this Sensei branch](https://github.com/Automattic/sensei/pull/7056).
- Opened a lesson and ensured the editor looked as it did before.
- Verified that the template is set to _Single Lesson_.
- Edited the template and checked that it looked as it did before.
- Checked the frontend to ensure it looked as it did before.

I noticed that, when LM is enabled, _Single Lesson_ appears in the template dropdown. Is it possible to hide it in that scenario? I don't think we want to allow that (and it also doesn't look good if they use that template with LM).

Imran92:
Thanks for checking @donnapep ! 

> Let's change the title of this one to "Lesson" instead of "Single Lesson" to better match how Learning Mode templates are named.

I've changed the name from Single Lesson to just Lesson here https://github.com/Automattic/themes/pull/7265/commits/7b8c09cd98412cff808c6f938cff5c06eae29b87.

> Is it possible to hide it in that scenario? I don't think we want to allow that (and it also doesn't look good if they use that template with LM).

It was possible but found it super tricky. I've opened a PR here for it https://github.com/Automattic/sensei/pull/7056

Thanks!

donnapep (approved):

course/functions.php
```diff
@@ -85,3 +85,25 @@ function course_register_block_patterns_category() {
 
 add_action( 'init', 'course_register_block_patterns_category' );
 
+add_filter(
+	'theme_lesson_templates',
+	function( $page_templates, $theme, $post ) {
```
donnapep: We should try to avoid using inline functions with filters (and actions). It makes it next to impossible to remove them when you need to using `remove_filter`.

Imran92: Thanks @donnapep , I've moved it to a separate function here https://github.com/Automattic/themes/pull/7265/commits/6ad7328bb94b4f3b15b8219dbcf0e4a2de3d13b0

course/functions.php
```diff
@@ -85,3 +85,26 @@ function course_register_block_patterns_category() {
...
 add_action( 'init', 'course_register_block_patterns_category' );
 
+add_filter(
+	'theme_lesson_templates',
+	'course_theme_filter_single_lesson_template_for_sensei_learning_mode',
+	11,
+	3
+);
+
+function course_theme_filter_single_lesson_template_for_sensei_learning_mode( $page_templates, $theme, $post ) {
```
donnapep, line 106: This function is missing a DocBlock.

donnapep (commented):
~Unfortunately, I'm still seeing "Lesson" in the dropdown when LM is enabled. I can see it being `unset`, but I guess something must be firing later to add it back in?~

Never mind! I was on the wrong Sensei branch. It's getting confusing. 😅

course/functions.php
Imran92: Thanks! Updated it here https://github.com/Automattic/themes/pull/7265/commits/9015d270c250378f71c842018b6aff0ec2cdab77

donnapep (approved):

madhusudhand:
@Imran92 Along with another theme changes, I have deployed these changes here D118587-code

Imran92:
Thanks a lot @madhusudhand !

PR #7267, opened 2023-07-25
Block Canvas: Remove Rubik font from heading definition

mikachan:
Block Canvas is currently setting Rubik as the heading font in its theme.json file, however, this font isn't included with the theme and I believe the theme is meant to default to the system font.

This theme is generally used as a starting point for many other block themes and we have seen Rubik accidentally included in other themes, so I think it should be removed to prevent this from happening.

### DIFF
--- a/block-canvas/theme.json
+++ b/block-canvas/theme.json
@@ -339,7 +339,6 @@
 			},
 			"heading": {
 				"typography": {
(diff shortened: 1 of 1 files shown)


scruffian (approved):
LGTM

PR #7271, opened 2023-07-26
Strand: Add theme

alaczek:
Initial commit. 

Ready for launch post: pNEWy-gTF-p2


pbking (approved):
Well designed and well executed!  Nicely done @alaczek!  I love the sticky side panel.

I made only one small change to remove the reference on a navigation block.

The theme itself is ready to bring in, however the point you made about the jetpack subscription block we might want to see resolved before we ship the theme.  I see that you opened that [here](https://github.com/Automattic/jetpack/issues/31063).

PR #7276, opened 2023-07-27
Update theme.json

alaczek:
Add custom spacing sizes.

Could someone double check for me that will not affect existing sites?

### DIFF
--- a/creatio/theme.json
+++ b/creatio/theme.json
@@ -57,7 +57,38 @@
 				"steps": 7,
 				"unit": "rem"
 			},
-			"spacingSizes": [],
+			"spacingSizes": [
+				{
+					"name": "1",
+					"size": "0.5rem",
+					"slug": "30"
+				},
+				{
+					"name": "2",
+					"size": "1rem",
+					"slug": "40"
+				},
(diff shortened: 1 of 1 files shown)


mikachan (commented):
> Could someone double check for me that will not affect existing sites?

I think these changes will affect existing sites slightly. The default spacing sizes from Core are:

- --wp--preset--spacing--30: 0.67rem
- --wp--preset--spacing--40: 1rem (same as this PR)
- --wp--preset--spacing--50: 1.5rem
- --wp--preset--spacing--60: 2.25rem
- --wp--preset--spacing--70: 3.38rem
- --wp--preset--spacing--80: 5.06rem

These new sizes are quite close to the default sizes, it may be worth comparing the templates with the two different sets of sizes and seeing what impact they have visually.

alaczek:
That is a great point! I tried it out locally and yes, there are those differences, but they are fairly minor, and mostly noticeable when viewing side-by-side. So if there are no code issues I'd like to merge this change.

mikachan (approved):
Sounds good to me!

PR #7277, opened 2023-07-27
Add navigation block markup provided by CBT

### LINKED ISSUE #79622: Iotix - Critical Error on pages when using PHP 8.0
### Quick summary

When `Iotix` is active on a site using PHP 8.0, it causes a critical error when viewing static pages. 

This is the error from CLI:

`[06-Jul-2023 18:00:38 UTC] PHP Fatal error:  Uncaught TypeError: array_key_exists(): Argument Automattic/themes#2 ($array) must be of type array, null given in /wordpress/core/6.2.2/wp-includes/block-supports/align.php:48`

Reported in this interaction: 6507135-zd-woothemes

### Steps to reproduce

1. On an Atomic site, activate `Iotix`.
2. Make sure PHP version is 8.0.
3. Visit live site.

The site to load without errors.

### What actually happened

Seeing message about critical error on the site.

### Browser

_No response_

### Context

6507135-zd-woothemes

Atomic

### Other notes

_No response_

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

Yes, easy to implement

### Workaround details

Switching to PHP 7.4 resolves the error.

vcanales:
Fixes https://github.com/Automattic/wp-calypso/issues/79622

Missing properties on the Navigation block for the header template part were causing sites with PHP@8 or higher to break.

#### Testing
0. Make sure you're using PHP version 8 or higher, which is where the error happens.
1. Activate Iotix.
2. Go to Settings > Reading, set a to Static Page, and choose any existing page.
3. Visit site; it should not break.
4. Click on a few other pages for good measure — it was breaking for me even when skipping the Reading settings instruction, but the problem was reported using the steps above.

### DIFF
--- a/iotix/parts/header.html
+++ b/iotix/parts/header.html
@@ -3,7 +3,7 @@
 <div class="wp-block-group alignwide" style="padding-top:var(--wp--preset--spacing--70);padding-right:0;padding-bottom:var(--wp--preset--spacing--60);padding-left:0"><!-- wp:site-title /-->
 
 <!-- wp:group {"style":{"spacing":{"padding":{"top":"0","right":"0","bottom":"0","left":"0"}}},"layout":{"type":"flex","flexWrap":"nowrap"}} -->
-<div class="wp-block-group" style="padding-top:0;padding-right:0;padding-bottom:0;padding-left:0"><!-- wp:navigation {5} /-->
+<div class="wp-block-group" style="padding-top:0;padding-right:0;padding-bottom:0;padding-left:0"><!-- wp:navigation {"layout":{"type":"flex","justifyContent":"left","orientation":"horizontal"}} /-->
 
 <!-- wp:buttons -->
 <div class="wp-block-buttons"><!-- wp:button {"fontSize":"small"} -->


mikachan (approved):
This works for me! I tested with PHP 8.0.0.

I was able to replicate the error following your testing steps without this PR. I see this on my static homepage:

```
Fatal error: Uncaught TypeError: array_key_exists(): Argument #2 ($array) must be of type array, null given in /wp-includes/block-supports/align.php on line 48
```

With this PR, the homepage (and all other pages I've tested) work successfully without error.

I think we should merge this PR, but I also wonder if this should be handled more defensively upstream. I don't think it's ideal that incorrect markup is causing a fatal error, maybe it should be contained/handled by the navigation block.

vcanales:
> I think we should merge this PR, but I also wonder if this should be handled more defensively upstream. I don't think it's ideal that incorrect markup is causing a fatal error, maybe it should be contained/handled by the navigation block.

Agreed 100%; the original error message was misleading and not helpful at all in finding the actual issue. I'll look into how to follow-up with a definitive fix.

PR #7278, opened 2023-07-27
Bitacora: Edit sidebar markup to use aside

iamtakashi:
Change sidebar markup to use aside for better accessibility. Ref: pdXvLu-1X2-p2

### DIFF
--- a/bitacora/readme.txt
+++ b/bitacora/readme.txt
@@ -12,6 +12,9 @@ Bitácora is a simple old-school blog theme.
 
 == Changelog ==
 
+= 1.0.8 =
+* Edit sidebar markup to use aside (#7278)
+
 = 1.0.7 =
 * Small tweaks to force update following `style-variations` tag fix (#7209)
 
--- a/bitacora/readme.txt
+++ b/bitacora/readme.txt
@@ -12,9 +12,6 @@ Bitácora is a simple old-school blog theme.
 
 == Changelog ==
 
-= 1.0.8 =
-* Edit sidebar markup to use aside (#7278)
-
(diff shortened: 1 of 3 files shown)


bitacora/readme.txt
```diff
@@ -12,6 +12,9 @@ Bitácora is a simple old-school blog theme.
 
 == Changelog ==
 
+= 1.0.8 =
+* Edit sidebar markup to use aside (#7278)
+
```
mikachan: The changelog will be updated automatically when the theme is deployed, so we don't need to add a manual update here.

bitacora/style.css
```diff
@@ -7,7 +7,7 @@ Description: Bitácora is a simple old-school blog theme.
 Requires at least: 6.0
 Tested up to: 6.1.1
 Requires PHP: 5.7
-Version: 1.0.7
+Version: 1.0.8
```
mikachan: This will also be updated automatically on deployment.

mikachan (approved):
Thanks for this update! It looks good to me.

Just left a few comments about the changelog and version updates, as these don't need to be included in the PR.

bitacora/readme.txt
iamtakashi: Sorry about that!

bitacora/style.css
iamtakashi: I'll fix them now. Sorry!

iamtakashi:
@mikachan I've revert the version bump that wasn't necessary. Thanks for letting me know about it.

mikachan:
No worries, LGTM!

PR #7316, opened 2023-08-15
Startfit: Remove CSS solution that is no longer necessary for the avatar block

iamtakashi:
Since the block was fixed in https://github.com/WordPress/gutenberg/pull/53007, the css solution is no longer needed, and in fact, it gets in the way of Global Styles.

**Before**
Screenshot 2023-08-15 at 16 09 56

**After**
Screenshot 2023-08-15 at 16 09 41

This also changes `Tested up to` to 6.3.


mikachan (approved):
Nice!

PR #7319, opened 2023-08-16
Add new theme Hall

pbking:
Add new theme Hall

image


pbking:
I've done some cleanup of the theme; There are a few more things to do before this is ready to ship.

The content in the patterns needs to be in english so that it can be localized.

Patterns need to be made and referenced from template parts.  Perhaps the "Home" content should be a full page pattern?

I think there are a few spacing inconsistencies that remain.

henriqueiamarino:
[hall.zip](https://github.com/Automattic/themes/files/13350188/hall.zip)

Here's the new theme zip file, thank you.

mikachan:
I've updated this PR to include the latest changes and I've fixed a spacing issue I could see on the home template. I'm going to merge this so @henriqueiamarino can continue working on it.

PR #7323, opened 2023-08-17
Tenaz: Make sure header top always has side paddings

iamtakashi:
I realised the social links and the search field bump the edge of the screen.

Add a non-constrained group block wrapping the column blocks.

**Before**
localhost local_ (12)

**After**
localhost local_ (11)


mikachan (approved):
✨

PR #7326, opened 2023-08-18
Add Lineup to dotorg deploy list

mikachan:
This adds Lineup to the automated Dotorg deploy list, now that Lineup is live on the Dotorg themes directory.

cc. @iamtakashi (Just looping you in, but this work should be automated in future!)


iamtakashi:
Ok, good to know. Thanks!

PR #7333, opened 2023-08-28
Course Theme: Add gravatar hovercard in comments

### LINKED ISSUE #7306: Integrate Gravatar Hovercards with comments
We would like to integrate [Gravatar Hovercards](https://github.com/gravatar/hovercards) with comments.

> We could also experiment with adding something like 'Profile' under the Avatar that also shows the Hovercards. 

Context - p1691699311741419-slack-C02NWDZBL0H

Imran92:

- Added gravatar hovercard to comment query block

#### Gravatar hovercard library documentation
https://github.com/gravatar/hovercards/blob/0.5.8/README.md

#### Testing instructions:

- Enable this branch of course theme
- Enable comments on a page/post
- Go to the page/post from frontend
- Comment as a user who's email has a gravatar account
- Hover on the gravatar image you see in the comment
- Make sure you see the hovercard

https://github.com/Automattic/themes/assets/6820724/449f4f86-e996-45cf-b199-1c6c10e375c2

#### Related issue(s):
https://github.com/Automattic/themes/issues/7306


course/assets/js/hovercard-setup.js
```diff
@@ -0,0 +1,9 @@
+document.addEventListener('DOMContentLoaded', () => {
```
donnapep, line 1: Could we follow the same whitespace rules in this JS file as we do for Sensei?

donnapep:
This doesn't look so great on mobile. I wonder if we could use `wp_is_mobile` to not load the Gravatar files in that case? Not sure how reliable that function is, but if it shows on tablets and not phones, that's actually better.

Screenshot 2023-10-19 at 11 37 01 AM

donnapep:
Is there any P2 or similar that documents how to integrate hovercards that we could link to in the description?

course/functions.php
```diff
@@ -119,3 +124,28 @@ function course_theme_filter_single_lesson_template_for_sensei_learning_mode( $p
...
+ * Defer the parsing of specific JS scripts.
+ *
+ * @param string $tag The <script> tag for the enqueued script.
+ * @param string $theme The script's registered handle.
+ *
+ * @since Course 1.3.3
+ *
+ * @return string The <script> tag for the enqueued script with defer attribute.
+ */
+function defer_parsing_of_js( $tag, $handle ) {
```
donnapep: What do you think about bumping the minimum supported WP version to 6.3 and specifying a [loading strategy](https://make.wordpress.org/core/2023/07/14/registering-scripts-with-async-and-defer-attributes-in-wordpress-6-3/) instead?

donnapep:
@Imran92 Can you provide an update here?

Imran92:
Thanks for the ping Donna. Some theme PRs got left out when we moved to other project. I'll update this PR as per the suggestions above. Adding it to my to-do

Imran92:
> This doesn't look so great on mobile. I wonder if we could use `wp_is_mobile` to not load the Gravatar files in that case? Not sure how reliable that function is, but if it shows on tablets and not phones, that's actually better.
> 
> Screenshot 2023-10-19 at 11 37 01 AM

Yap you are absolutely right. Looks like gravatar-hovercard library fixed it themselves on [0.5.8](https://github.com/gravatar/hovercards/releases/tag/0.5.8) release to not load the hovercard on mobile! So I've just pointed to the updated version here https://github.com/Automattic/themes/pull/7333/commits/c3f3ebad7e8f9e8157979856309a3870c4299ec9. On my testing, it looks like it doesn't detect resizing, so we have to reload the page to understand that it's working. Thanks!!

course/functions.php
Imran92: I've used loading strategy and updated the version number here https://github.com/Automattic/themes/pull/7333/commits/4ee873da89ba5facafb67841af86bc4e2c374e35. Very nice idea Donna! Thank you!!

course/assets/js/hovercard-setup.js
Imran92: Yapp, good point. I've updated it here https://github.com/Automattic/themes/pull/7333/commits/07d099ef4bb5b9512279d90d6ef4707ba6216076

Imran92:
> Is there any P2 or similar that documents how to integrate hovercards that we could link to in the description?

I found their [readme](https://github.com/gravatar/hovercards/blob/0.5.8/README.md) very helpful and contained all the required information. I'm adding it to the PR description.

course/assets/js/hovercard-setup.js
Imran92: Oh, looks like the spaces get removed by this pre-commit hooks https://github.com/Automattic/themes/pull/3596

Imran92: Oh, looks like the spaces get removed by this pre-commit hooks https://github.com/Automattic/themes/pull/3596

PR #7334, opened 2023-08-28
Independent Pub 2: use get_avatar

sgomes:
Use `get_avatar` to get the full avatar `` element, instead of manually composing it in the theme.

This gets us a `srcset`, so that we can pick a resolution-appropriate image size. I also took the opportunity to explicitly add `loading="eager"`, given that this avatar is a likely candidate for the LCP (Largest Contentful Paint) element, and should thus not wait until layout occurs to start being fetched.

- Replace manual `` tag generation with `get_avatar`.

#### Related issue(s):

For wpcom, this change was initially done in D114229-code, but it appears to have later been overwritten by the version of Independent Publisher 2 in this repo.


taipeicoder (approved):
👍

sgomes:
Thank you for the quick review, @taipeicoder! 👍

PR #7339, opened 2023-08-30
Optimismo: Update for dotorg submission

iamtakashi:
Update the theme for dotorg submission.

**Changes**
* Updated the `readme.txt`
* Updated the description and the tags in the stylesheet
* Removed the css fix to the avatar nlock
* Improved the post template in `index` to give the spacing instead of a CSS fix
* Removed unnecessary layout definition from `theme.json` 
* Minor format improvements in `theme.json`


jffng (commented):
Hey @iamtakashi, the changes here look good, I noticed one thing that is a bit unrelated to this PR but the theme bundles [many images](https://github.com/Automattic/themes/tree/trunk/optimismo/assets/images) that are not used? Maybe we should remove these, or if they are used, we should add credits.

Other than that, LGTM.

iamtakashi:
@jffng Oh! Thanks for catching that. They were from the base theme, and I now deleted them. If it looks ok now, can you merge and deploy? Thank you!!

PR #7354, opened 2023-09-08
Remove category from taxQuery to avoid a not found error

miksansegundo:

- Remove the `category` from `taxQuery` in queries to avoid a not found error

This fixes broken query blocks because the attribute `taxQuery` include a category ID that doesn't exist in user sites.


mikachan (approved):
Thank you! ✨

PR #7370, opened 2023-09-14
Montagna: Make sure the elements don't bump to the screen edges

alaczek:
In this case elements were not bumping against browser edges, but there definitely was misalignment.

| Before | After |
| ----------- | ----------- |
| image|image|


montagna/templates/single.html
```diff
@@ -1,8 +1,10 @@
-<!-- wp:group {"style":{"spacing":{"blockGap":"0"}},"layout":{"type":"constrained"}} -->
-<div class="wp-block-group"><!-- wp:template-part {"slug":"header","tagName":"header","align":"full"} /-->
+<!-- wp:group {"style":{"spacing":{"blockGap":"0","padding":{"right":"0","left":"0"}}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group" style="padding-right:0;padding-left:0"><!-- wp:template-part {"slug":"header","theme":"montagna","tagName":"header","align":"full"} /-->
```
mikachan:
```suggestion
<div class="wp-block-group" style="padding-right:0;padding-left:0"><!-- wp:template-part {"slug":"header","tagName":"header","align":"full"} /-->
```

We don't need to include the `theme` attribute here as it's added automatically in the editor.

mikachan (commented):
Apart from one small change, LGTM!

alaczek:
Thanks for the review!

PR #7373, opened 2023-09-15
Update styles of question feedback

Imran92:

- .grade class is common in both question header and feedback. We've specified it more to be applied only in the header.

#### Related issue(s):
https://github.com/Automattic/sensei/issues/7115

### DIFF
--- a/course/theme.json
+++ b/course/theme.json
@@ -824,7 +824,7 @@
 						}
 					}
 				},
-				"css": "& .grade { font-family: var(--wp--preset--font-family--system) } & .sensei-lms-question__answer-feedback--incorrect{ --color: #f2ddbe; }"
+				"css": "& .grade { font-family: var(--wp--preset--font-family--system) } & .sensei-lms-question__answer-feedback--incorrect { --color: #f2ddbe; }"
 			}
 		},
 		"color": {
--- a/course/theme.json
+++ b/course/theme.json
@@ -824,7 +824,7 @@
 						}
 					}
 				},
-				"css": "& .grade { font-family: var(--wp--preset--font-family--system) }"
+				"css": "& .grade { font-family: var(--wp--preset--font-family--system) } & .sensei-lms-question__answer-feedback--incorrect{ --color: #f2ddbe; }"
 			}
 		},
 		"color": {
--- a/course/styles/blue.json
+++ b/course/styles/blue.json
@@ -300,7 +300,7 @@
 						}
 					}
 				},
-				"css": "& .grade { font-weight: 600; font-family: var(--wp--preset--font-family--heading); } "
+				"css": "& .sensei-lms-question-block__header .grade { font-weight: 600; font-family: var(--wp--preset--font-family--heading); }"
 			}
 		},
 		"elements": {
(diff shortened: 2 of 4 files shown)


donnapep:
We don't want to target `trunk`, do we?

Imran92:
> We don't want to target trunk, do we?

As we'll have more PRs in the release, we shouldn't, as that'll trigger an auto-release. I've created another feature branch and targeted this PR to that. Thanks!

donnapep:
@Imran92 I spent some time auditing and testing this (and the related PRs). I noticed there are some inconsistencies in the design that made figuring out the correct colors a bit tricky. 😅 

The only small thing I noticed that I think we should change (if it's not too difficult) is the border color of the ordering question for the Gold variation. Right now they don't match the border colors of the "Correct" and "Incorrect" feedback, and I think they should. So the ordering question borders would be:

- Incorrect - `#d69557`
- Correct- `#68a071`

course/theme.json
```diff
@@ -824,7 +824,7 @@
 						}
 					}
 				},
-				"css": "& .grade { font-family: var(--wp--preset--font-family--system) }"
+				"css": "& .grade { font-family: var(--wp--preset--font-family--system) } & .sensei-lms-question__answer-feedback--incorrect{ --color: #f2ddbe; }"
```
donnapep: Curious if the default theme needs the more specific `& .sensei-lms-question-block__header .grade` selector or no?

donnapep: ~I'm also wondering if we should start putting this Sensei-specific CSS in Sensei's 3rd-party CSS folder instead of embedding it in the theme. I believe we started doing that to try to keep Sensei stuff out of the theme, since Sensei is not required to use it, and it doesn't make sense to bloat the theme with CSS it will never need.~

UPDATE: Thinking about this more, currently it's not possible to know which variation is currently being displayed, so until we (or Gutenberg) adds the variation to the `body` tag or something, I guess this won't be possible. So never mind. 😅

Imran92:
> I noticed there are some inconsistencies in the design that made figuring out the correct colors a bit tricky. 

Agreed! 💯 

I've updated the border color here https://github.com/Automattic/themes/pull/7373/commits/7a6e5acab58157bd2f60738cc908637935063428 . The color in the plugin was too specific and we'd have write a lot of selectors in `gold.json` to override that. To avoid that, I've reduced the specificity in the Pro PR https://github.com/Automattic/sensei-pro/pull/2421/commits/994b9bc1f6c06752444c1d9df31d18e3bd74edfb, but that unfortunately dismissed the review 😓 

Thank you!

donnapep:
@Imran92  Just wondering about my other question https://github.com/Automattic/themes/pull/7373#discussion_r1341291162?

donnapep (approved):

Imran92:
> @Imran92 Just wondering about my other question https://github.com/Automattic/themes/pull/7373#discussion_r1341291162?

Hi @donnapep! 👋 Sorry, missed that question. In Figma, the font family and weight are the same for both title grade and feedback grade for the default variation. So we didn't need to specify it more for this one as we're avoiding adding more specificity unless needed.

Thank you! ☺️ 


image

PR #7377, opened 2023-09-21
Creatio 2: Add theme

alaczek:
This is an updated version of the Creatio theme, used as the base theme in the Pattern Assembler.
See pNEWy-hl2-p2 for details.


vcanales (approved):
LGTM!

PR #7380, opened 2023-09-22
Awburn: Make sure full width elements stretch edge to edge. This also fix 404 template.

### LINKED ISSUE #7375: Awburn: Full width alignment doesn't stretch to edge to edge
### Quick summary

Awburn's layout can afford a full-width element to stretch edge to edge. However, there are always gaps between the element and the window. This happens to both the page and the single templates.

### Steps to reproduce

1. Publish a post and a page with a full-width element in it.
2. Open the page and the post on the front of the site.

A full-width element stretches fully edge to edge.

### What actually happened

There are gaps between the element and the window.

localhost local_width-test-post_-1

iamtakashi:

Fixes #7375 

I realised the 404-html was also badly broken. So I fixed that as well.

| Before | After |
| ----------- | ----------- |
| localhost local_core-blocks-post_ (1) | localhost local_core-blocks-post_ |

### DIFF
--- a/awburn/templates/404.html
+++ b/awburn/templates/404.html
@@ -2,10 +2,10 @@
 <div class="wp-block-group" style="border-top-color:var(--wp--preset--color--secondary);border-top-style:solid;border-top-width:10px"><!-- wp:template-part {"slug":"header","tagName":"header"} /--></div>
 <!-- /wp:group -->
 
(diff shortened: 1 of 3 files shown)


vcanales (approved):
LGTM!

Issue #7402, opened 2023-10-02
Creatio: Single template has no comments / comment form


supernovia:
### Quick summary

I wasn't sure whether to file this as an enhancement request or bug, but since Creatio isn't turning off comments for posts, I figure it's probably a bug not to have a comment form at all without editing the template to add one.

### Steps to reproduce

1. Use Creatio on any site
2. View a post that has comments open
3. Notice there is no comment form
4. Edit the Single template
5. Notice there is no comment form there either

### What you expected to happen

I would expect Single to have a comment form for posts, and be able to show comments, without requiring editing

### What actually happened

Users have to add their own comment form and comments. Since there's no simple insertion method for that, it's a bit of work for them to add those in.

### Browser

Google Chrome/Chromium, Mozilla Firefox, Microsoft Edge, Apple Safari, iOS Safari, Android Chrome

### Context

User report, volunteer report, this thread: https://wordpress.com/forums/topic/comments-box-15/?

### Platform (Simple, Atomic, or both?)

Simple, Atomic, Self-hosted

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

Yes, difficult to implement

### Workaround details

User has to build the comments section, adding a form, comments, any needed spacers, etc.

I wouldn't say it's terribly hard for an HE, but it's a fair bit of work for an end user.

cuemarie:
📌 REPRODUCTION RESULTS
- Tested on Simple – **Replicated**

📌 FINDINGS/SCREENSHOTS/VIDEO
Activating Creatio on a simple site with default Comment settings removes the comments form on the Single Posts template.

https://github.com/Automattic/themes/assets/27249804/b4606e08-2078-48eb-b55b-db4eb99d0724

📌 ACTIONS
- Triaged
- Added to Bug Scrub Planning
- flagged on the Creatio Themedevs p2 post as well, as there's related feedback about this there: pNEWy-h3y-p2#comment-56685

pak-cia:
+1 - 71239160-zen

alaczek:
Hey @supernovia 👋 I have a PR that adds the comment form to single posts. Do you think it should also be added to page template?

supernovia:
> Hey @supernovia 👋 I have a PR that adds the comment form to single posts. Do you think it should also be added to page template?

[...]

PR #7422, opened 2023-10-10
Update single.html

alaczek:
Add comment form to single posts.

#### Related issue(s):
#7402


vcanales (approved):
Looks good

vcanales:
I see a pending question in https://github.com/Automattic/themes/issues/7402, on whether or not to add comments to pages as well. If the answer is yes, I suggest making this comment template into a template part before merging, so that we may reuse it instead of copy-pasting the code.

PR #7423, opened 2023-10-10
Fix/raw

pbking:

This makes a few changes related to colors and their usage in the theme.

Removes the unused 'tertiary' color (which was the same as 'foreground' but with an alpha layer)
Removed references to 'primary' color (which isn't provided by the theme) and instead use 'foreground'.
Remove some unnecessary color assignments.
Invert some color assignments in order to make the theme match the [demo](https://rawthemedemo.wordpress.com/2023/04/04/boston-city-hall/) and achieve the expected design.
Include a missing image.


vcanales (approved):
Looks good

PR #7432, opened 2023-10-17
Add body class for theme variations

Imran92:

- We've added a custom property in each variation which we later parse and add in body class from sensei. Allows us to add CSS using CSS files.
- Added a hook `course_theme_variation_loaded`, description is added below.
- Added stylesheets hooked to the action above, loading variation specific stylesheets.

#### Added hook
`course_theme_variation_loaded` - Fires after the variation is determined, can be used for different use cases, including loading variation specific stylesheets

Details in this PR https://github.com/Automattic/sensei/pull/7225


course/theme.json
```diff
@@ -145,6 +145,7 @@
 			"commentGapSmall" : "10px",
 			"commentGapMedium" : "1.25rem",
 			"courseNewsletterGap": "clamp(2.5rem, -0.292rem + 9.306vw, 6.688rem)",
+			"courseThemeVariation": "default",
```
donnapep, line 148: I think we could call this `variation` since we know we're already using the Course theme. 🙂 Unless it has some special meaning in `theme.json` (but I don't think so)?

donnapep (approved):

Imran92:
Hi @donnapep 👋 

As per our discussion here https://github.com/Automattic/sensei/pull/7225, I've moved the functionality here.

It gives us the added advantage of not having to check which theme is in use.

Also, I've added a hook which can be used from sensei or any other place to enqueue variation specific stylesheets.

I've also added stylesheets for each variation.

Once we have this PR approved, we can close the associated PR in Sensei because we won't need that anymore. Also, we can share our approach in the theme team's issue.

Thanks!

course/functions.php
```diff
@@ -119,3 +111,65 @@ function course_theme_filter_single_lesson_template_for_sensei_learning_mode( $p
...
+
+	return $classes;
+}
+
+/**
+ * Enqueue the specific stylesheet for the variation.
+ *
+ * @param string $variation_name The current theme variation.
+ *
+ * @since Course $$next-version$$
```
donnapep: I think we should go ahead and add the proper version number here and above since we don't have a script we can run to replace `$$next-version$$` in this repo.

donnapep (approved):
Left a wee comment about not using `$$next-version$$` in this repo.

course/functions.php
Imran92: Replaced it here with 1.3.5 https://github.com/Automattic/themes/pull/7432/commits/5c0a9b42710274d767638149e57baa72902ecdd6 ! Thanks! I'll go ahead and merge it as it is a very small change

PR #7457, opened 2023-11-03
Sunderland: add theme

jffng:
Adds theme designed by @beafialho .


jffng:
@beafialho where does the `summertime_image.png` come from?

beafialho:
@jffng here's the image source: https://www.rawpixel.com/image/8799420/photo-image-plant-fire-public-domain

PR #7459, opened 2023-11-03
Renamed 'Spearhead' to 'Spearhead (Classic)' and 'Spearhead Blocks' to 'Spearhead'

pbking:
Only the NAME has changed.  The Theme URI, Text Domain, folder/slug and everything else remains as it was before.

image


mikachan (approved):
Changes look good to me! I also tested activating and deactivating these themes, just in case this highlighted any odd problems with the name change. All looks good!

iamtakashi:
After this getting merged and deployed, can we change the names in the showcase?

PR #7462, opened 2023-11-06
Fix extra padding issue in outline buttons

Imran92:

In Course theme, some extra padding get added to the outline button from the core. This overrides the spacing we've added for buttons in theme.json, making the outlined button different than normal button.
Screenshot 2023-11-06 at 4 04 17 PM
Screenshot 2023-11-06 at 4 04 47 PM

We've added the padding we add for buttons in variation CSS files here to get more specificity over the other padding so that our padding is used.

#### After
image

#### Testing Instructions
- Add a normal button and an outline button in a page
- In trunk, they'll have different size vertically
- In this branch, they should look similar

#### Related issue(s):
https://github.com/Automattic/sensei/issues/7243


merkushin (approved):
👍

donnapep:
@Imran92 What's the update for this PR? Do we still need it?

Imran92:
Looks like we'll still need it. The outline buttons are still bigger than the filled buttons -

#### In trunk
Screenshot 2024-01-25 at 8 49 41 PM

#### In this branch
image




But looks like WP core has added a new style for outline button padding that's taking precedence over our selectors' specificity. I'll update this PR to make it one step more specific so that our style gets applied instead -
image

PR #7481, opened 2023-11-12
Dara: Fix TikTok Embed Block

### LINKED ISSUE #7341: Dara: Tiktok block has blockquote formatting
### Quick summary

Tiktok blocks are inserted as `<blockquote>` elements as mentioned in this Gutenberg issue: https://github.com/WordPress/gutenberg/issues/47104

For themes that apply specific styles to blockquotes, these styles also apply to Tiktok blocks specifically. Given the lack of movement on that Gutenberg issue, adding a fix into affected themes might be worth exploring.

### Steps to reproduce

1. Activate Dara theme. 
2. Insert a Tiktok block anywhere on the site.
3. View the site.

The Tiktok embed should appear on its own as a standard embed.

### What actually happened

On Dara, the Tiktok block has extra left padding and a large opening quotation mark as if it's a Quote block. 

### Browser

_No response_

### Context

6663892-zd-a8c

Simple, Atomic, Self-hosted

### Other notes

_No response_

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

Yes, easy to implement

### Workaround details

CSS can act as a workaround for sites with the Premium plan or higher. Those without are forced to either switch themes or embrace the odd appearance.

Aurorum:
Ideally, Core would change the oEmbed for TikTok so that the `<blockquote>` is replaced with a `<div>` - but it seems like they're not keen on that. Dara seems to be the only theme where significant styling is done to quotes which damages such embeds, so it's probably worth fixing here.

| Before | After |
|--------|--------|
| Screenshot 2023-11-12 at 21 50 09 | Screenshot 2023-11-12 at 21 49 29 |

Edit: actually, I changed my mind on targeting all embeds - it probably makes more sense to only target videos instead. I can imagine there's some embeds where a quote might be useful (eg. a tweet which has subsequently been deleted).

#### Related issue(s):
Fixes #7341

### DIFF
--- a/dara/blocks.css
+++ b/dara/blocks.css
@@ -191,6 +191,16 @@ blockquote cite {
 	color: #fff;
 }
 
+/* Embed */
+
+.wp-block-embed blockquote {
+	padding: 0;
+}
+
+.wp-block-embed blockquote:before {
+	content: none;
+}
+
 /*--------------------------------------------------------------
 3.0 Blocks - Formatting Blocks
 --------------------------------------------------------------*/
--- a/dara/blocks.css
+++ b/dara/blocks.css
@@ -193,11 +193,11 @@ blockquote cite {
 
 /* Embed */
 
-.wp-block-embed blockquote {
+.wp-block-embed.is-type-video blockquote {
 	padding: 0;
 }
 
-.wp-block-embed blockquote:before {
+.wp-block-embed.is-type-video blockquote:before {
 	content: none;
 }
 


Aurorum:
Actually, I changed my mind on targeting all embeds - it probably makes more sense to only target videos instead. I can imagine there's some embeds where a quote might be useful (eg. a tweet which has subsequently been deleted).

vcanales (approved):
Thanks, this works as advertised.

PR #7490, opened 2023-11-15
Update theme twentytwentyfour

adamziel:
Changes made in WordPress Playground


adamziel:
Oh yay, it actually worked!

PR #7518, opened 2023-11-26
Varia: Include Jetpack Content Options

### LINKED ISSUE #1754: Varia: [Improvement] Add "Content options" to Varia and its Childs
I have had some users asking to hide featured images or authors.

Other themes have this option, and not sure if that's something in the scope of Global Styles:

Screen Shot 2020-01-18 at 10 36 18

For Premium plan users we can give some CSS code. But the only option for Personal plan users is to pick a different theme.

### LINKED ISSUE #3544: Varia themes: Add option to also hide automatic featured images on posts
Currently, Varia child themes have an option to show/hide featured images on single pages:  
Screen Shot on 2021-03-28 at 10:08:37

It would be handy to have a similar option to allow users to prevent featured images from automatically showing at the top of single posts. Currently, the only way to do this is by adding CSS code, which is not something all users know how to do or have access to, depending on their plan.

`````
.single-post . post-thumbnail  {
  display:none;
}
`````

Context 28653328-hc
Related https://github.com/Automattic/themes/issues/1754

### LINKED ISSUE #3637: Varia: add Content Option checkbox to show/hide featured images on posts
Could users have a Content Option to hide featured images on posts, much like the option to hide or [show featured images on pages](https://github.com/Automattic/themes/pull/3220)? I would recommend defaulting to "on" since that's the current behavior.

We're often asked us how to either hide or resize post featured images using CSS. Support and volunteers have always just given the users the CSS needed to hide the images (or told them they'd have to upgrade to do that) but perhaps it would be better to give users an option they can find on their own.

This came up today for a user [in this thread](https://wordpress.com/forums/topic/maywood-and-balasana-content-option-featured-image/) who thought our themes were buggy, since the Content Options checkbox did not affect posts. I can see how they would misunderstand.

Aurorum:
This includes Jetpack Content Options to Varia and therefore its child themes. In #3220, an option was added to hide the Featured Images from pages. This was included in the section `jetpack_content_options`, but Jetpack Content Options weren't actually registered. 

cc @scruffian: I believe this was an oversight, and not intentional - would you mind confirming that? I'm reverting that change here in favour of adding Jetpack Content Options, which means that users can choose to hide Featured Images on posts and the homepage as well, rather than just pages. Thanks!

---

I've set the defaults based on how they are currently, so that sites using this theme right now won't be affected without changes being manually made in the Customizer. 

Screenshot 2023-11-26 at 16 03 24
Screenshot 2023-11-26 at 16 03 39

#### Related issue(s):
Fixes #3637
Fixes #3544
Fixes #1754

### DIFF
--- a/varia/functions.php
+++ b/varia/functions.php
@@ -225,7 +225,17 @@ if ( ! function_exists( 'varia_setup' ) ) :
 				'enable_theme_default' => true,
 			)
 		);
-
+		
+		// Add support for Content Options.
+		add_theme_support( 'jetpack-content-options', array(
+			'blog-display' => 'content',
+			'featured-images' => array(
+				'archive'         => true,
+				'archive-default' => true,
+				'post'            => true,
+				'page'            => true,
+			),
+		) );
 	}
 endif;
 add_action( 'after_setup_theme', 'varia_setup' );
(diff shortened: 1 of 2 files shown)


scruffian (approved):
LGTM

Aurorum:
Hi @scruffian - thanks very much for the review! We just had a user follow-up about this in 7380619-zen, so if you have some time, would you mind merging and syncing with WP.com? I’m afraid that I don’t have the perms to do so.

mreishus:
D131935-code : Not deployed yet, will need to do some testing

mreishus:
Deployed now.

Aurorum:
Thanks @mreishus - much appreciated :)

pbking:
This seems to be causing problems for users who had previously hidden featured images.

With this change implemented users who previously didn't have featured images on their pages and posts have them showing unexpectedly.

They can still turn them off, but that's not desired.

I think perhaps this should be reverted.

Aurorum:
If it gets reverted, wouldn't that cause problems for people who have made changes subsequent to Content Options being added? The old implementation was flawed which caused this issue to start with, but users are at least able to have more control over Featured Images now. 

I also don't think that `get_theme_mod( 'show_featured_image_on_pages', false )` would be saved, unless I'm mistaken? So it wouldn't have any effect.

PR #7521, opened 2023-11-29
Indice: Add theme

iamtakashi:
Add Indice theme created by @mtias and @pablohoneyhoney. pNEWy-hIc-p2 I'm following up from here.

screenshot


carolinan:
This theme name has the protentional to be very confusing, since the index file is highly relevant to theme development.
Try searching for "WordPress theme index", the theme would be competing directly with the theme developer handbook: https://developer.wordpress.org/themes/getting-started/your-first-theme/
But also the .com documentation: https://wordpress.com/support/templates/edit-the-index-and-archive-templates/

iamtakashi:
@carolinan Thanks for the feedback on the name. 

I will rename it `Indice`, which means "index" in French. I'll add a comment when I'm done with the renaming. Hold on.

iamtakashi:
Renaming is done, and it's now ready for review. Thanks in advance!!

vcanales (approved):
Looks good!

PR #7528, opened 2023-12-06
Assembler: Add theme

fushar:

Add a new base theme for the Assembler flow.

For now, this theme just reuses most of the template parts and patterns from the existing Creatio 2 theme.

#### Related issues

- https://github.com/Automattic/dotcom-forge/issues/4778

cc: @richtabor @SaxonF


assembler/patterns/byline-date.php
```diff
@@ -0,0 +1,20 @@
+<?php
+/**
+ * Title: Byline and Date Post Meta
+ * Slug: assembler/byline-date
+ * Inserter: no
+ */
+?>
+<!-- wp:group {"style":{"spacing":{"blockGap":"0.26rem","padding":{"bottom":"20px"}},"typography":{"fontSize":"x-small"}},"layout":{"type":"flex","flexWrap":"nowrap","verticalAlignment":"center","justifyContent":"center"}} -->
+<div class="wp-block-group" style="padding-bottom:20px;font-size:x-small"><!-- wp:paragraph {"fontSize":"x-small"} -->
+<p class="has-x-small-font-size"><?php echo esc_html_x( 'Published by', 'assembler' ); ?></p>
```
miksansegundo: > [!TIP]
> When we are not passing context information for the translators, we should use [esc_html_e](https://developer.wordpress.org/reference/functions/esc_html_e/) rather than [esc_html_x](https://developer.wordpress.org/reference/functions/esc_html_x).
>
> Otherwise, `esc_html_x` uses "assembler" as the context information for the translators, and "default" as the domain when "assembler" should be the domain. The domain is a unique identifier for retrieving translated strings.

assembler/parts/header.html
```diff
@@ -0,0 +1,13 @@
+<!-- wp:group {"layout":{"type":"constrained"}} -->
```
miksansegundo: Adding the header as a pattern in a PHP file and then including it in the template part, as we do for the footer, would be better because that will allow us to add these comments:

```
<?php
/**
 * Title: header
 * Slug: assembler/header
 * Categories: featured, header
 * Block Types: core/template-part/header
 */
?>
```
The [header in TT4](https://github.com/WordPress/twentytwentyfour/blob/trunk/parts/header.html) doesn't add these comments. Do you have any idea why? @alaczek @richtabor

fushar: > Adding the header as a pattern in a PHP file and then including it in the template part

But does that really matter for this theme, as this template part will be (almost) always overwritten by the header pattern that's selected in the Assembler flow?

assembler/templates/page.html
```diff
@@ -0,0 +1,50 @@
...
+	<!-- wp:group {"style":{"spacing":{"blockGap":"var:preset|spacing|40","margin":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group" style="margin-top:var(--wp--preset--spacing--60);margin-bottom:var(--wp--preset--spacing--60)"><!-- wp:post-title {"textAlign":"left","align":"wide","fontSize":"xxxx-large"} /--></div>
+<!-- /wp:group -->
+
+<!-- wp:group {"layout":{"type":"default"}} -->
+<div class="wp-block-group"><!-- wp:post-content {"lock":{"move":false,"remove":false},"layout":{"type":"constrained"}} /--></div>
+<!-- /wp:group -->
+
+<!-- wp:group {"layout":{"type":"constrained"}} -->
+<div class="wp-block-group"><!-- wp:comments {"style":{"spacing":{"margin":{"top":"var:preset|spacing|70"}}}} -->
```
miksansegundo: > [!WARNING]  
> Let's remove the comments block `wp:comments` from this template. Starts from `<!-- wp:comments` until `<!-- /wp:comments -->` but we have to remove the group wrapping that block too.

assembler/parts/header.html
miksansegundo: I'd say yes because:
- `Categories` is required to show that pattern in the right list in the inserter.
- `Block Types` is required to show that pattern in the modal when adding/replacing a header.

assembler/theme.json
```diff
@@ -0,0 +1,344 @@
...
+					"fontSize": "var(--wp--preset--font-size--medium)",
+					"fontWeight": "600"
+				}
+			}
+		},
+		"color": {
+			"background": "var(--wp--preset--color--theme-1)",
+			"text": "var(--wp--preset--color--theme-5)"
+		},
+		"css": ".no-underline a { text-decoration: none; }",
```
miksansegundo, line 249: This is interesting. I'm just curious about this class name `no-underline`. Where/how is it used? @richtabor

assembler/parts/header.html
fushar: Done!

assembler/templates/page.html
fushar: Done. I forgot about this!

assembler/patterns/byline-date.php
fushar: Done. TIL. I just copied from Creatio 2 😂

alaczek:
Looks good to me! LGTM and start testing!

assembler/theme.json
richtabor: In the case where you don't want an underline applied, but the core control applies the CSS to parent container. Such as the post title block.

assembler/parts/header.html
richtabor: Template parts don't need the headers; only patterns do. :)

PR #7532, opened 2023-12-06
Add proper templates to Assembler (WIP)

richtabor:
Updates the templates for Assembler, as a follow-up for https://github.com/Automattic/themes/pull/7528.

## An exploration into simpler templates v. page content

This PR reduces the template load to index, archive, and single, as follows: 

### Index
- Template file consists of header part, main element with post content, and footer part. 
- Loads the page set as homepage in the reading settings. 
- All pages resemble this template; it's the source.

### Archive
- Template file contains a generic blog layout for archive views. 
- We can provide alternate archive templates to replace with (like TT4 does).

### Archive
- Template file contains a generic single post layout. 
- We can provide alternate single post templates to replace with  (like TT4 does).

### Caveats
In the Assembler flow on WordPress.com, we would need to create a "Home" page and assign it to the homepage via the reading settings option—instead of creating/augmenting a home.html template.

- The reading settings **must** be set to a static page. 
- A homepage must be set for the static page. 

Props @SaxonF.


fushar:
@richtabor Sorry could you rebase; there's conflict after I merged https://github.com/Automattic/themes/pull/7529 😄

autumnfjeld:
@ianstewart

SaxonF:
The template hierarchy can be a confusing concept to grasp even for experienced developers. I would advocate for moving in the direction of treating templates more like simple wrappers leaving pages to contain your sites content. This is something core needs to address but the approach here at least begins to move in that direction. 

It also plays nicely into the idea that the showcase can become a much more composable / mix and match type experience that leads into the assembler. If most themes are driven by the same pattern library, then applying a homepage from the showcase is as simple as copying the pattern or page.

All that to say, given the confusion on Dotcom surrounding templates, I think this approach would be a worthwhile experiment.

scruffian:
This makes me think of an idea connected to something @SaxonF shared earlier. It seems like we could never use home.html and front-page.html, and instead rely on reading settings to control the contents of the home page. This would make it more important to have reading settings in the Site Editor.

richtabor:
I made [an alternate PR](https://github.com/Automattic/themes/pull/7538) to get Assembler templates in a decent place; we can discuss/explore this method in the meantime.

richtabor:
Closing for now, as the alternate was merged. https://github.com/Automattic/themes/pull/7538

PR #7534, opened 2023-12-07
Add strict types declaration to latest themes

vcanales:
Deploy script won't allow themes to be commited without strict types declaration.
Recent previous case: https://github.com/Automattic/themes/pull/7529

### DIFF
--- a/beep/functions.php
+++ b/beep/functions.php
@@ -7,7 +7,7 @@
  * @package Beep
  * @since Beep 1.0
  */
-
+declare( strict_types = 1 );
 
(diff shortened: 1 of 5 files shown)


fushar (approved):

iamtakashi:
@vcanales Thank you for taking care of this (and for all the reviews)! I'll remember to add these going forward.

vcanales:
There are still a few missing ones; I didn't realize we also need to add them to the patterns, or, essentially, to _all_ PHP files. I can get on that later today, unless you'd like to do it :)

PR #7538, opened 2023-12-07
Add templates to Assembler (alternate)

richtabor:
Alternate to #7532, following up on https://github.com/Automattic/themes/pull/7528. 

This PR follows the existing model for using a `home.html` template as the home page, rather than using the page content as proposed in #7532. It also applies the other tweaks as added to the alternate PR. 

This model assumes the reading settings of the site have not changed, and that you create a page if you'd like a blog. 

### Home page:

It's odd that this renders "Blog home" when the home.html template is in use, though there is [an issue](https://github.com/WordPress/gutenberg/issues/56244) to clear this up:

CleanShot 2023-12-07 at 13 48 29

In this model it's also odd to have the posts per page controls in the details view on the left, as the theme may not have a blog-oriented home page:

CleanShot 2023-12-07 at 13 48 42

### Other pages: 

Work as expected, ready to receive pattern input: 

CleanShot 2023-12-07 at 13 49 57

### Blog page: 

If you want a blog, you add a "Blog" page and apply a Query loop block to it. The caveat here is that WordPress does not know which page is your blog page, therefore permalinks are not applied to subsequent posts, nor is the details view populated with the blog details (as shown in the "Home page" section above):

CleanShot 2023-12-07 at 14 06 55

*Note that I added spacer blocks in the visuals, to help indicate where pattern content is intended.


richtabor:
There's some detail work to clean up on the archive/search templates - but they're ok for now.

assembler/templates/home.html
```diff
@@ -0,0 +1,7 @@
+<!-- wp:template-part {"slug":"header"} /-->
```
fushar, line 1: We don't need to include a Home template, because the Assembler flow will always add it (see Creatio 2).

richtabor: Shouldn't Assembler just override the default? Otherwise the theme can't be used as-is, right?

fushar: This theme (and the existing Creatio 2) won't be discoverable outside the Assembler flow, so this template will never be possibly usable.

We had this convo before, see: p1665497100719849/1665152933.113869-slack-CRWCHQGUB.

alaczek:
Aside from the home.html convo this looks good to me!

alaczek (approved):

assembler/templates/home.html
richtabor: In the future it may be available elsewhere (perhaps leveraged by WooCommerce). Would the theme + Assembler on dotCom function properly with home.html existing in the theme files, even if it's not necessarily used?

fushar: I see. I believe it should work -- let's 🚢 and try!

fushar (approved):

fushar:
@alaczek, will you be willing to merge and deploy this to Assembler theme? Let's see it in action ASAP.

alaczek:
I merged but can't deploy, likely because we're missing strict type declarations: https://wp.me/p3topS-1Q6 can take care of it a bit later, unless someone beats me to it.

assembler/theme.json
```diff
@@ -340,5 +388,19 @@
 			"fontSize": "16px",
 			"lineHeight": "1.65"
 		}
-	}
-}
+	},
+	"templateParts": [
```
miksansegundo, line 392: The template part "comments" is used in single.html, but it's not included in `templateParts`. I remember we had issues when template parts were missing here. 

@alaczek shall we add "comments" here?   

From the [docs](https://developer.wordpress.org/themes/block-themes/templates-and-template-parts/#adding-template-parts-to-theme-json):

> If you would like the template part to be selectable in the block inserter, you also need to include it in the templateParts section of theme.json.

alaczek: I opened a new PR to take care of this.

PR #7555, opened 2023-12-14
Assembler: Fixing minor issues

alaczek:
Following the changes in #7538, I'm adding a strict type declaration to php files, and comments template part declaration to theme.json.


fushar (approved):
🚢

PR #7556, opened 2023-12-14
Infield: Add theme.

alaczek:
This is the initial PR to add the Infield theme. More info: pNEWy-hQ6-p2


pbking:
I removed some fonts that didn't seem to be used but otherwise this theme seems ready to bring in.

pbking (approved):

PR #7565, opened 2023-12-20
Update theme erma

beafialho:
Changes from WordPress Playground


beafialho:
I wanted to have renamed the `wp-custom-template-center-aligned-post.html` but I didn't know if it was that simple and didn't want to break anything. If possible, when reviewing this, let's rename it to `center-aligned-post.html`.

mikachan (approved):
> If possible, when reviewing this, let's rename it to center-aligned-post.html.

I've made this change ✅

The theme changes test well for me, thank you!

PR #7567, opened 2023-12-20
Default featured images to false and leverage theme mod values that have already been set

pbking:
A possible alternative to #7566 

Default Jetpack Content options to FALSE
Includes any previously set values as the default value for Jetpack Content Options

#### Related issue(s):
https://github.com/Automattic/wp-calypso/issues/85369

### DIFF
--- a/varia/functions.php
+++ b/varia/functions.php
@@ -225,15 +225,16 @@ if ( ! function_exists( 'varia_setup' ) ) :
 				'enable_theme_default' => true,
 			)
 		);
-		
+
 		// Add support for Content Options.
 		add_theme_support( 'jetpack-content-options', array(
 			'blog-display' => 'content',
 			'featured-images' => array(
-				'archive'         => true,
-				'archive-default' => true,
-				'post'            => true,
-				'page'            => true,
+				'archive'         => false,
+				'archive-default' => false,
+				'post'            => false,
+				//featured images on pages boolean was previously managed by Varia rather than jetpack.  If that value has been set default to that.
+				'page'            => get_theme_mod( 'show_featured_image_on_pages', false ),
 			),
 		) );
 	}
--- a/varia/functions.php
+++ b/varia/functions.php
@@ -230,11 +230,13 @@ if ( ! function_exists( 'varia_setup' ) ) :
 		add_theme_support( 'jetpack-content-options', array(
 			'blog-display' => 'content',
 			'featured-images' => array(
-				'archive'         => false,
-				'archive-default' => false,
-				'post'            => false,
+				'archive'         => true,
+				'archive-default' => true,
+				'post'            => true,
+				'post-default'    => true,
+				'page'            => true,
 				//featured images on pages boolean was previously managed by Varia rather than jetpack.  If that value has been set default to that.
-				'page'            => get_theme_mod( 'show_featured_image_on_pages', false ),
+				'page-default'    => get_theme_mod( 'show_featured_image_on_pages', false ),
 			),
 		) );
 	}


varia/functions.php
```diff
@@ -225,15 +225,16 @@ function varia_setup() {
...
 			'featured-images' => array(
-				'archive'         => true,
-				'archive-default' => true,
-				'post'            => true,
-				'page'            => true,
+				'archive'         => false,
+				'archive-default' => false,
+				'post'            => false,
+				//featured images on pages boolean was previously managed by Varia rather than jetpack.  If that value has been set default to that.
+				'page'            => get_theme_mod( 'show_featured_image_on_pages', false ),
```
Aurorum: As I understand it, this would remove that option when you're only seeking to set the default value: https://jetpack.com/support/content-options/

```suggestion
				'page-default'    => get_theme_mod( 'show_featured_image_on_pages', false ),
```

You could add `post-default` as `false` too, if you'd prefer.

Aurorum (changes requested):

pbking:
Hrm, looking at Varia it looks like additional changes would be needed to a number of templates to avoid Featured Images from being displayed multiple times.

https://github.com/Automattic/themes/blob/trunk/varia/template-parts/content/content-single.php#L25
https://github.com/Automattic/themes/blob/trunk/varia/template-parts/content/content.php#L28
etc...

Aurorum:
> Hrm, looking at Varia it looks like additional changes would be needed to a number of templates to avoid Featured Images from being displayed multiple times.

Sorry - not trying to be disruptive here, but only trying to help! :) That was the reason why I leveraged Jetpack in the first place because it handles all this - the problem with the original Varia implementation is it only dealt with pages. As such, #7566 would cause Featured Images to be displayed by default, and the user wouldn't be able to hide them. 

If Jetpack Content Options is removed, I'd argue that settings need to be implemented for those templates (which they don't with this PR, hence why I favour it).

pbking:
I see what you're saying, that in order to make Varia work better would require more changes.

However considering the age of Varia and how much effort should be taken to change Varia in ANY way I don't believe we should be enabling this functionality to users.  If they are currently seeking more control over their content I believe we should direct them to block themes where this option is in their hands.

Closing this.  I don't believe we should be changing varia at all and the original change was more disruptive than helpful.

pbking:
Ok, I stepped through all the scenarios I could think of and, other than now needing Jetpack as a dependency this seems to be working nicely as expected.

Reopening for evaluation.

varia/functions.php
pbking: Looking at the existing behavior it looks like all instances DO display the featured image, so (except for pages when had already introduced that option defaulting to false) the rest I believe are correct defaulting to 'true'.

Aurorum (approved):
This is much better, I'd say! But if you'd like to remove the Jetpack dependency entirely, give me 10 mins and I'll have a separate PR. :)

varia/functions.php
```diff
@@ -225,15 +225,18 @@ function varia_setup() {
...
-		
+
 		// Add support for Content Options.
 		add_theme_support( 'jetpack-content-options', array(
 			'blog-display' => 'content',
 			'featured-images' => array(
 				'archive'         => true,
 				'archive-default' => true,
 				'post'            => true,
+				'post-default'    => true,
```
Aurorum, line 236: Oh, on second thought - I thought the original issue recommended setting this to false?

```suggestion
				'post-default'    => false,
```

Aurorum (changes requested):

varia/functions.php
pbking: I think only setting 'page-default' to false is necessary.

As you mentioned there is no option to turn it off before the Jetpack change... and from what I observe if a Post has a featured image it is displayed, no option otherwise.  So setting the 'post-default' value to false means that a site doesn't change based on this code change.

Aurorum: Ah, yes, whoops, I got muddled. Yes, I think this is good.

Aurorum (approved):

pbking:
:/ You're gonna laugh...

But now I'm worried about LOSING the Jetpack Dependency... If customers used the existing Jetpack functionality to change their site... and now we change how that's configured... will they lose those Jetpack-related changes?

I think... I'm fine going with what is in this change rather than eliminating the Jetpack Dependency.

Aurorum:
> But now I'm worried about LOSING the Jetpack Dependency... If customers used the existing Jetpack functionality to change their site... and now we change how that's configured... will they lose those Jetpack-related changes?

Heh, I'm glad that I was persuasive! Yes, they would. The only benefit that I can think of to losing the Jetpack dependency (other than not needing Jetpack, but I doubt that will really affect anyone) is that https://github.com/Automattic/jetpack/issues/29286 wouldn't be an issue - I tried fixing it yesterday, and had no luck. :) 

That's very much your call though.

pbking:
Thanks tons for your assist @Aurorum!

Gonna go ahead and ship this now.

Aurorum:
@pbking would you like to still proceed with removing the Jetpack dependency or does this approach work for you?

PR #7568, opened 2023-12-20
Varia: Remove Jetpack Dependency For Featured Images

### LINKED ISSUE #85369: Featured Images defaulting ON for pages and posts in Varia and Varia child themes.
### Quick summary

Multiple users reporting that featured images are showing up on pages and posts where they're not wanted. This is due to the "Featured Images" checkboxes in the Content Options of the Customizer being selected in the theme without user action.

### Steps to reproduce

1. Start on a site with Varia or a child theme active. (Confirmed on Maywood, Stow, Alves.)
2. Go to Customizer →  Content Options.
3. Look under "Featured Images. Many sites will have these checkboxes selected without having done so themselves.

If options are selected, featured images will appear on pages and posts.

If options are deselected, featured images will not appear on pages and posts, but will also not display image previews in post blocks (Blog Posts, Latest Posts, etc.)

### What actually happened

If options are selected, featured images will appear on pages and posts, sometimes duplicating images already present.

If options are deselected, featured images will not appear on pages and posts, but will also not display image previews in post blocks (Blog Posts, Latest Posts, etc.)

### Impact

Some (< 50%)

### Available workarounds?

_No response_

### Logs or notes

Known cases so far:
* 7458679-zd-a8c
* 7458223-zd-a8c
* 7449348-zd-a8c

Aurorum:
@pbking - this is how I'd go about #7566 :) Up to you if you'd prefer it to #7567, but here's the advantages. 

**Pros:**
A lot of the reports in Automattic/wp-calypso#85369 were about the fact that Content Options automatically hides the Featured Images set in the Latest Posts block. There is an open issue in Jetpack about this - Automattic/jetpack#29286 - but I tried fixing it, and it seems super tricky to do. This would at least remove Varia from the list of themes affected by that bug.

Also doesn't require Jetpack anymore, but I don't think that's a problem - can't imagine many self-hosted sites without Jetpack using Varia, to be honest. 

**Cons:**
User who have made changes in the last two weeks for the Posts and Blog settings will need to do so again. 

Feel free to close if, in your judgement, the cons outweigh the pros here. 

#### Related issue(s):
Closes Automattic/wp-calypso#85369

### DIFF
--- a/varia/functions.php
+++ b/varia/functions.php
@@ -227,17 +227,9 @@ if ( ! function_exists( 'varia_setup' ) ) :
 		);
 
 		// Add support for Content Options.
+		// We handled Featured Images within Varia directly.
 		add_theme_support( 'jetpack-content-options', array(
 			'blog-display' => 'content',
-			'featured-images' => array(
-				'archive'         => true,
-				'archive-default' => true,
-				'post'            => true,
-				'post-default'    => true,
-				'page'            => true,
-				//featured images on pages boolean was previously managed by Varia rather than jetpack.  If that value has been set default to that.
-				'page-default'    => get_theme_mod( 'show_featured_image_on_pages', false ),
-			),
 		) );
 	}
 endif;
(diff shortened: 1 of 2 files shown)


pbking:
Ooh I like it!

Giving this a solid test.  Hadn't quite yet shipped the alternative yet.

pbking:
Thank you @Aurorum I checked this branch out.  It's not behaving quite like I was expecting.  Images aren't shown in situations where I expect them to.  I think it could ultimately be made to work and not having the Jetpack dependency would be great but it's not there and with the many child themes that varia has might not be the most efficient fix.

I'm going to close this one and go with the #7567 option instead; it works as expected as I stepped through the timeline of changes as a customer.

Thank you for your help and for presenting another option.

Aurorum:
No worries at all, that makes sense! Glad to have a solution which fixes Automattic/wp-calypso#85369 while still giving users control of their Featured Images. :)

PR #7578, opened 2023-12-21
Bedrock: Initial commit

iamtakashi:
Add Bedrock theme. pNEWy-hSd-p2

screenshot


vcanales (changes requested):
I'm not sure what's happening with the markup of the Index template, but it's not working for me:

image

Attempting recovery by clicking the button works, but it moves the post-list pattern into the template.

iamtakashi:
Thanks for catching that! All the templates were broken on the site without the Gutenberg plugin. It turns out the missing role attribute from one of the `div`s for the cover block was the culprit. I've added it, and now all the templates work with and without the Gutenberg plugin. Could you test it again? Thanks in advance.

vcanales (approved):
Works well now!

PR #7586, opened 2024-01-05
Assembler: Another round of tweaks

richtabor:
- [x] Add custom CSS for flex-grow, to allow for header contents to flow better on mobile.
- [x] Remove template locking from page.html template. 
- [x] Update fonts and font families.
- [x] Add better fallback for colorless social icons (potentially overriding `.wp-block-social-links.is-style-logos-only` colors to use currentColor instead).


richtabor:
@alaczek this is ready for review + push to prod whenever you have a few. 🥳

alaczek (approved):

PR #7600, opened 2024-01-10
Bedrock: add missing strict types declaration

vcanales:
Quick fix: add missing strict types declaration on post meta pattern.


iamtakashi:
Oops, thanks for catching that!

PR #7637, opened 2024-02-06
Feelin Good: Add theme

iamtakashi:
Add Feelin' Good theme. pNEWy-i0p-p2

screenshot


vcanales (commented):
There seems to be a problem with the cover block markup on all templates — using the markup after running the block recovery fixes the issue, but maybe you have another solution, @iamtakashi> 

image

iamtakashi:
Thanks for catching that. Cover block markup seems to be changed in the trunk. I changed the markup to the recovered markup without the Gutenberg plugin, and that seems to work with/without the plugin.

vcanales (approved):

PR #7643, opened 2024-02-09
Assembler: Minor spacing and button margin styling

richtabor:
Minor tweaks to make columns and buttons work better without needing spacing adjustments in patterns. Also a tiny utility class to leverage in patterns. 

### Before
CleanShot 2024-02-09 at 16 26 34

### After
CleanShot 2024-02-09 at 16 26 03

### DIFF
--- a/assembler/style.css
+++ b/assembler/style.css
@@ -88,6 +88,10 @@ footer {
 
 }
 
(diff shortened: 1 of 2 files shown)


alaczek (approved):
Looks good to me!

PR #7653, opened 2024-02-20
Perenne: remove period from version

vcanales:
Extra period causes deploy script to fail when comparing versions.


iamtakashi:
Oops, sorry about that! I'll be careful about it going forward.

PR #7659, opened 2024-02-22
Feelin good: Add strict types to footer

vcanales:
Add one missing strict types declaration.


iamtakashi:
Nice catch! Phew, thank you!!

PR #7671, opened 2024-03-11
Lossless image optimization

danluu:
In principle, this shouldn't change the visual appearance of any images. ImageOptim was run with the highest compression setting except that JPEG metadata stripping was disabled because some users prefer to avoid stripping JPEG metadata.

Overall, this reduces image sizes by 12.4% (`17.8 MB` saved on `147.1 MB` in images).

I haven't edited the changelog section of the readme for the changed themes. I'd be happy to do that as well, but I wanted to put this up for discussion before making readme changes since it will be quite a few changes and if people prefer a different structure (multiple pull requests, not committing optimizations to certain image types, etc.), it would be nice to get that feedback before making a bunch of readme changes.

While this change isn't zero risk because lossless image compression software can have bugs, there's a `wpcom` job that does this automatically for images on `wpcom`, so I think this adds a negligible amount of risk on top of what's already happening to images served from `wpcom`.

Shrink image sizes using lossless compression.

#### Related issue(s):

#5357, which appears to be a less global version of this proposed change.


dsas:
It'd be cool to ensure that new themes don't reintroduce unoptimized images. Is there some documentation on what you did?

>  if people prefer a different structure (multiple pull requests, not committing optimizations to certain image types, etc.), it would be nice to get that feedback before making a bunch of readme changes.

I can't imagine why it would be a problem, maybe someone from @Automattic/pixel has some thoughts though.

danluu:
@dsas For this, I used https://imageoptim.com/mac and ran it on the entire repo. For people on Macs, it's pretty easy to use (you can drag and drop entire folders into it).

In terms of making sure unoptimized images don't make it in by accident, one possibility would be to have a hook that runs an image optimizer automagically. For Linux, one possibility would be to use https://trimage.org/. IME, it doesn't compress things quite as much, but it's fairly close. Both `ImageOptim` and `Trimage` call other open source tools optipng, pngcrush, etc., and those tools could also be called individually.

Since this repo already uses node to manage some stuff, I wonder if the simplest way to do it would be via some node package, which there seem to be a lot of.

scruffian:
I think it would be great to try creating the hook and then commit it and watching these image be compressed as part of that hook. Then we know it's a robust solution going forward rather than a one time fix.

danluu:
@scruffian Sure. I have a few questions since I'm not familiar with this repo or php or javascript or any kind of web development.

I see that there's a [pre-commit-hook.js](https://github.com/Automattic/themes/blob/trunk/pre-commit-hook.js). Is that where I'd want to add this? And how can I see this the hook to see what's happening?

Also, when I ran `ImageOptim` locally, it took a while to run. I didn't time it, but I'd guess it was an hour or two. If it's going to run on everything instead of only new or changed files, then it's going to add a bit latency to commits. Is that considered an ok amount of latency?

There's some value in running it on all files because improvements in image optimizers give some gains over time (for my personal stuff, I effectively run this stuff on a cron and, even after an image can't be improved for a while, it will often later get a bit smaller as someone improves one of the tools), but that's relatively minor compared to the large savings we get on some images in the first run, so it seems fairly reasonable to only run on new or changed files, although that's sort of a different thing.

scruffian:
> I see that there's a [pre-commit-hook.js](https://github.com/Automattic/themes/blob/trunk/pre-commit-hook.js). Is that where I'd want to add this? And how can I see this the hook to see what's happening?

Yes. The best way to test it to just commit things to a branch and it will run.

>  Is that considered an ok amount of latency?

No a few hours is definitely not ok! I think there are two options:

1. Only run the hook on newly added images
2. Use a bot to create PRs which compress the images automatically every X day (for example we have one that updates dependencies like this: https://github.com/Automattic/themes/pull/7655)

If you think this is going to be too tricky then I'm also happy to just merge this PR and we can add these optimisations as a followup.

danluu:
If it's something that shouldn't be slow, then I think a bot that occasionally runs is probably the right approach  — even if we only look at newly added images, if someone adds a theme that has some large images, it could be quite slow (I haven't really looked into how these things scale, but I've had a single really large image take over 10 minutes).

Yeah, I think it makes sense to just merge this and then have the bot thing be a separate issue. There's a related change we could make that I think is higher impact, can't really be done automatically, and will also be less work that setting up a bot to do this automatically, which is replacing some of the losslessly compressed image with much smaller lossy compressed images.

One place this should make a large difference is when going to wordpress.com and then using the theme picker to select a theme for a new blog. @niranjan-uma-shankar noted that this step downloads 70MB of images and @dsas noted that a lot of this size comes from the images bundled with themes. On looking at the images, there are a lot of this big "screenshot" images that are lossless PNGs, which seems somewhat non-optimal for a number of reasons and I think we should be able to get images that are (just a guess based on trying this with one image) ~1/10th the size, which should make the entire sign-up / theme picker flow a lot more usable for anyone who doesn't have a high-end connection.

Anyway, I didn't want to submit that change since it conflicts with this one, so merging this one so we can do this other change makes sense to me, with an eye on eventually adding some kind of bot that ensures that we don't have accidental lingering unoptimized images for too long.

PR #7681, opened 2024-03-15
Craftfully: Fix for missing "contrast" preset color

miksansegundo:
- Rename the color preset named `contrast` as `foreground` because `contrast` doesn't exist.

@alaczek I wasn't sure if you wanted to create a `contrast` preset or use `foreground` instead. 

#### Related issue(s):

Screenshot 2567-03-15 at 16 02 38


alaczek:
Not sure what you mean by "because contrast doesn't exist."?

I think with this theme I was following the naming convention from TT4, and I did want to use `contrast`.

miksansegundo:
I am confused now. I think is the other way around  as how I describe it in this PR.

The markup uses "foreground" but that color is not [defined](https://github.com/Automattic/themes/blob/b2e10d6db1537abc143f95ca3741008765bb9f9d/craftfully/theme.json#L14) in the theme.json.

miksansegundo:
See "foreground" being used [here](https://github.com/Automattic/themes/blob/b2e10d6db1537abc143f95ca3741008765bb9f9d/craftfully/patterns/newsletter-subscription.php#L25)

alaczek:
Oh yeah, in that case it should be switched to contrast.

miksansegundo:
Okay, closing this in favor of https://github.com/Automattic/themes/pull/7787 Could you review and merge that?

PR #7687, opened 2024-03-19
Cookbook: Add theme

alaczek:
Add Cookbook theme. See more: pNEWy-hjH-p2

image

cc @henriqueiamarino


alaczek:
I activated the theme and I noticed a few things I wanted to double check with you @henriqueiamarino:
* in home.html template the blog post query is not inheriting query from the template, which results in an empty space, and need for extra config. Shall I change this to inhert query from template?
* single page template doesn't include the page template. What do you think about adding one?
* archive template is missing page navigation, so it's not possible to access all posts.
* the search template looks off - the header colors are different, and the layout of search results is not a grid - would it make sense to reuse the archive templates here to make it more consistent?
image

Let me know how you would like to address those and I can help!

henriqueiamarino:
I just replied to it on Slack.
I am reviewing all the themes I designed in my first year using the Editor, as I learned many best practices and will apply them. The same goes for style variations; my idea is to make them available in all my themes.

As soon as I finish the Cookbook update, I'll have it linked to this post; thanks.

henriqueiamarino:
The theme was thoroughly reviewed, and the following final is the final version; thanks, @alaczek.
[cookbook-update.zip](https://github.com/Automattic/themes/files/14903722/cookbook-update.zip)

alaczek:
This is looking good to me (I pushed a tiny fix). I have two questions (not blockers):
* could this text color be the same as header color? It's clashing a bit with the background in the Gecko's Dream variation
image

* I was wondering why the header.php pattern is hidden?
```
<?php
/**
 * Title: header
 * Slug: cookbook/header
 * Categories: hidden
 * Inserter: no
 */
?>
```

henriqueiamarino:
@alaczek I have no idea why the Header is hidden. I exported the theme using Create Block Theme, as I do regularly.

You're right about the black/green contrast on Gecko's style; it's not WCAG **AAA** as it should be. I will fix it as soon as I can export a theme again. Changes to the plugin are blocking me right now. Hopefully, I'll have the chance to fix it on Friday. Thank you.

Screenshot 2024-04-24 at 12 16 52

alaczek:
I fixed the text color for you @henriqueiamarino and wrapped some texts in a translation function. I think we're good to go!

PR #7694, opened 2024-03-21
Deploy script: Improve npm versioning step

mikachan:
Sometimes there can be issues with node modules which will cause the deploy script to fail. This PR removes the `--silent` parameter in the npm versioning step for each theme, so if there is an error, there should be a helpful error message printed to help debug the problem. Currently, it's difficult to see what's causing the script to fail.

I've also removed the step that adds the root package-lock.json file for each theme that has a package.json file, as the root package-lock.json file is added at the end of the updates on [line 642](https://github.com/Automattic/themes/blob/trunk/theme-utils.mjs#L642).

To test, ensure the deploy script runs successfully by running `npm run deploy`.

#### Related issues:
p1710970837782649-slack-C029FM1EH

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -781,9 +781,6 @@ async function versionBumpTheme(theme, addChanges) {
 		}
 		if (addChanges) {
 			await executeCommand(`git add ${file}`);
-			if (isPackageJson) {
-				await executeCommand(`git add package-lock.json`);
-			}
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


vcanales (approved):
LGTM. It might be worth it to look into committing the package-lock files for the themes that are npm projects as well, otherwise I'm not sure what it's usefulness is.

madhusudhand (approved):
LGTM!

mikachan:
> It might be worth it to look into committing the package-lock files for the themes that are npm projects as well, otherwise I'm not sure what it's usefulness is.

Yeah, I'm not sure what the reasoning was behind not committing the package-lock files for each theme 🤔

I'll merge this in for now, thanks both!

PR #7716, opened 2024-04-04
Add a script that inserts strict type declarations

vcanales:
Checks all php files in a given directory for existing strict type declartions, and if it doesn't find one, adds it.

Tasks:

- [x] Run the script as a part of the deployment script.

### DIFF
--- a/add-strict-types.sh
+++ b/add-strict-types.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+# Verify an argument is provided
+if [ "$#" -ne 1 ]; then
+    echo "Usage: $0 <directory>"
+    echo "Error: Directory argument missing."
+    exit 1
+fi
+
+# Directory containing PHP files, passed as an argument
+DIRECTORY="$1"
+
+# Loop over each PHP file in the specified directory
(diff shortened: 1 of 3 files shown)


pbking:
I added an npm command to execute the logic so I didn't have to wait for a deploy to do it.   I found that you have to pass the "last deployed hash" to get the list of changed themes (since it's a "changed from when" type of logic) and added that as well.

I seem to be having problems with my sandbox right now so I haven't been able to fully test this yet though. :/

pbking:
Got my sandbox working and gave this a whirl.  I tweaked the call to the bash command, it seems the file had changed at some point.

This seems to be working as expected now.  The script is doing what it should.

My only remaining suggestion is to make the output optional and pass `-quiet` or something from the deploy script; there's quite a lot there.

With or without that change I believe this is good to go!

pbking (approved):

vcanales:
> My only remaining suggestion is to make the output optional and pass `-quiet` or something from the deploy script; there's quite a lot there.

Thanks for the suggestion! I've added a condition to the output, which was purely for debug use; if a `DEBUG` env variable is set to anything truthy, it'll show up again.

PR #7719, opened 2024-04-05
Generate playground previews for theme changes

vcanales:
Adds a GitHub action that creates Preview Links for changed themes within a PR.
The action leverages two key tools:
- WordPress Playground's [URL fragment blueprints.](https://wordpress.github.io/wordpress-playground/blueprints-api/using-blueprints#url-fragment)
- github-proxy.com by @stoph — used to allow the Playground to download a zip of the changed themes in the PR.

The created links are then added as a comment on the PR thread, as you can see below.
The comment is updated after every push, accounting for changes to other themes.

### Testing
Feel free to push changes to other themes to this branch.
I'll remove all theme-related changes before merging.

### DIFF
--- a/.github/scripts/create-preview-links.js
+++ b/.github/scripts/create-preview-links.js
@@ -1,0 +1,151 @@
+const fs = require('fs');
+
+/*
+ * This function creates a WordPress Playground blueprint JSON string for a theme.
+ *
+ * @param {string} themeSlug - The slug of the theme to create a blueprint for.
+ * @param {string} branch - The branch where the theme changes are located.
+ * @returns {string} - A JSON string representing the blueprint.
+ */
+function createBlueprint(themeSlug, branch) {
+	const template = {
+		steps: [
+			{
+				step: 'login',
+				username: 'admin',
+				password: 'password',
+			},
+			{
+				step: 'installTheme',
+				themeZipFile: {
+					resource: 'url',
+					url: `https://github-proxy.com/proxy.php?action=partial&repo=Automattic/themes&directory=${themeSlug}&branch=${branch}`,
+				},
+			},
+			{
+				step: 'activateTheme',
+				themeFolderName: themeSlug,
+			},
+		],
+	};
+
+	return JSON.stringify(template);
+}
+
+/*
+ * This function reads the `style.css` file of a theme and returns the theme name.
+ *
+ * @param {string} themeSlug - The slug of the theme to get the name of.
+ * @returns {string} - The name of the theme as defined in the `style.css` file.
+ */
+function getThemeName(themeSlug) {
+	const styleCss = fs.readFileSync(`${themeSlug}/style.css`, 'utf8');
+	const themeName = styleCss.match(/Theme Name:(.*)/i)[1].trim();
+	return themeName;
+}
+
+/*
+ * This function reads the `style.css` file of a theme and returns the name of the parent theme.
+ * If the theme is not a child theme, it returns an empty string.
+ *
+ * @param {string} themeSlug - The slug of the theme to get the parent theme name of.
+ * @returns {string} - The name of the parent theme as defined in the `style.css` file.
+ */
+function getParentThemeName(themeSlug) {
+	const styleCss = fs.readFileSync(`${themeSlug}/style.css`, 'utf8');
+	const parentTheme = styleCss.match(/Template:(.*)/i);
+	const isChildTheme = parentTheme && '' !== parentTheme[1].trim();
+
+	if (!isChildTheme) {
+		return '';
+	}
+
+	return parentTheme && '' !== parentTheme[1].trim()
+		? parentTheme[1].trim()
+		: '';
+}
+
+/*
+ * This function creates a comment on a PR with preview links for the changed themes.
+ * It is used by `preview-theme` workflow.
+ *
+ * @param {object} github - An authenticated instance of the GitHub API.
+ * @param {object} context - The context of the event that triggered the action.
+ * @param {string} changedThemeSlugs - A space-separated string of theme slugs that have changed.
+ */
+async function createPreviewLinksComment(github, context, changedThemeSlugs) {
+	const changedThemes = changedThemeSlugs.split(' ');
+	const previewLinks = changedThemes
+		.map((themeSlug) => {
+			const parentThemeName = getParentThemeName(themeSlug);
+			const note = parentThemeName
+				? ` (child theme of **${parentThemeName}**)`
+				: '';
+
+			return `- [Preview changes for **${getThemeName(
+				themeSlug
+			)}**](https://playground.wordpress.net/#${createBlueprint(
+				themeSlug,
+				context.payload.pull_request.head.ref
+			)})${note}`;
+		})
+		.join('\n');
+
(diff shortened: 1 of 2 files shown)


.github/workflows/preview-theme.yml
```diff
@@ -0,0 +1,75 @@
...
+          # Retrieve list of all changed files
+          git fetch origin trunk:trunk
+          changed_files=$(git diff --name-only HEAD origin/trunk)
+          
+          # Loop through changed files and identify parent directories
+          declare -A unique_dirs
+          for file in $changed_files; do
+              dir_name=$(dirname "$file")
+              echo $dir_name
+              if [[ -f "$dir_name/style.css" ]]; then  # Check if the directory contains a theme
```
MaggieCabrera: I think this makes it so if a change is inside a subfolder within the theme (ie: a pattern) this check won't pass, cause $dir_name will be something like `quadrat/patterns` instead of just `quadrat`. I'm testing this script in the community themes repo: https://github.com/WordPress/community-themes/pull/162

MaggieCabrera: I made a change on that PR that solves it for me, maybe you want to replicate here

vcanales: Good catch! I've applied you fix here. Thank you!

pbking:
This is the best kind of magic.

pbking:
image

I'm not sure how to address (or that we should) but CHILD themes (which we probably won't build any more) don't work since the parent themes aren't also installed.

pbking:
Should we also have the Gutenberg plugin activated as well?

vcanales:
> I'm not sure how to address (or that we should) but CHILD themes (which we probably won't build any more) don't work since the parent themes aren't also installed.

I've added a check and a warning for child themes, advising the user to install the parent theme in order for the preview to work.

> Should we also have the Gutenberg plugin activated as well?

For what purpose? I'd imagine that if we intend on shipping themes compatible with the latest version that might be necessary, but I don't know if that's the case.

.github/scripts/create-preview-links.js
```diff
@@ -0,0 +1,151 @@
...
+I've detected changes to the following themes in this PR: ${changedThemes
+		.map((themeSlug) => getThemeName(themeSlug))
+		.join(', ')}.
+
+You can preview these changes by following the links below:
+
+${previewLinks}
+
+I will update this comment with the latest preview links as you push more changes to this PR.
+**⚠️ Note:** The preview sites are created using [WordPress Playground](https://wordpress.org/playground/). You can add content, edit settings, and test the themes as you would on a real site, but please note that changes are not saved between sessions.
```
mikachan, line 108: Small suggestion to add an extra line break here:

```suggestion
I will update this comment with the latest preview links as you push more changes to this PR.

**⚠️ Note:** The preview sites are created using [WordPress Playground](https://wordpress.org/playground/). You can add content, edit settings, and test the themes as you would on a real site, but please note that changes are not saved between sessions.
```

mikachan (commented):
This is ace! Thank you for working on it.

I tested this by making changes to LowFi and Programme, and I can see the preview links were generated correctly and the GH comment was updated, but I don't think I'm seeing the changes made to the themes correctly. For Programme, I updated the theme.json color presets, but they're not reflected in the preview link version of the theme:

image

The palette should include a new "accent" color (#E619CE) and "secondary" should be changed to #000000.

vcanales:
> I tested this by making changes to LowFi and Programme, and I can see the preview links were generated correctly and the GH comment was updated, but I don't think I'm seeing the changes made to the themes correctly. For Programme, I updated the theme.json color presets, but they're not reflected in the preview link version of the theme:

Thanks for noticing this. It took a bit of debugging to figure out what was happening, and it's caused by the fact that the branch name includes a `/`, and how `github-proxy` reads and uses branch names. I've filed a PR to get that fixed upstream: https://github.com/stoph/github-proxy/pull/3

mikachan:
> Thanks for noticing this. It took a bit of debugging to figure out what was happening, and it's caused by the fact that the branch name includes a /, and how github-proxy reads and uses branch names.

Oh wow, good detective work there. Is this ready to test again? I saw your PR/proposed changes for `github-proxy` were merged 🎉

pbking:
It doesn't seem to have fixed things yet... I made some changes in Blockbase and don't see them reflected yet.

vcanales:
Things should be working now, this is good for another review.

pbking:
Yesss!  

Works for me.  This is beautiful.

pbking:
I have removed my Blockbase changes and approve this message.

pbking (approved):
Works great now.

fotograma/functions.php
```diff
@@ -9,7 +9,8 @@
  */
 	declare( strict_types = 1 );
 
-
+// adding a change
+// another change
```
pbking:
```suggestion
```

fotograma/patterns/404.php
```diff
@@ -1,15 +0,0 @@
-<?php declare( strict_types = 1 ); ?>
-<?php
-/**
- * Title: A 404 page
- * Slug: fotograma/404
- * Inserter: no
- */
-?>
```
pbking: Need to bring these files back?

fotograma/templates/404.html
```diff
@@ -1,27 +0,0 @@
-<!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|50","bottom":"var:preset|spacing|50"}},"position":{"type":"sticky","top":"0px"}},"backgroundColor":"base","layout":{"type":"constrained","justifyContent":"left"}} -->
-<div
-	class="wp-block-group has-base-background-color has-background"
```
pbking: Bring it back?

mikachan:
I'm still not seeing changes to Programme, am I testing in a weird way? 😅 I changed the Secondary colour and removed the 404 template, but I'm not seeing either change in the Playground link.

pbking:
Weird.  I also made some changes to Programme and am not seeing them reflected in the link.  Strange.  I'm certain I saw Blockbase changes earlier (the same type).

I'm not sure what's going on. :P

vcanales:
@pbking @mikachan the problem with Programme in particular is caused upstream, by the github-proxy service, because there seem to be temp files around from before the bug related to branch naming was fixed. I'm working with the author of that tool to get it fixed.

Could you try with another theme? I believe this PR is ready to be shipped if a theme that we haven't tried before works fine.

pbking:
Hrm...

I made font and color changes to Pendant.  However I'm not seeing either change now.

vcanales:
> Hrm...
> 
> I made font and color changes to Pendant. However I'm not seeing either change now.

I see. Yeah, still an upstream issue. I'll let you know when to review again.

vcanales:
It's been fixed upstream now. @pbking your changes to Pendant are actually being loaded correctly now. 

image

I think we're good for one more — final? — test now :) 
cc. @mikachan — your changes to Programme also seem to be previewed correctly.

mikachan (approved):
This is working great for me now, when testing Programme again and a different theme, Raw 🎉

Nice work @vcanales!

PR #7722, opened 2024-04-08
Update theme grammerone-update

henriqueiamarino:
Massive changes to the current theme. And a necessary update to the title.
[grammerone-update.zip](https://github.com/Automattic/themes/files/14955367/grammerone-update.zip)

Please change it from **GrammerOne** to **Grammer**.


henriqueiamarino:

PR #7723, opened 2024-04-08
Update theme the-menu-update

henriqueiamarino:
[the-menu-update.zip](https://github.com/Automattic/themes/files/14955376/the-menu-update.zip)


henriqueiamarino:

PR #7725, opened 2024-04-08
Update theme curriculum-update

henriqueiamarino:
[curriculum-update.zip](https://github.com/Automattic/themes/files/14955389/curriculum-update.zip)


henriqueiamarino:

PR #7726, opened 2024-04-08
Update theme boxedbio-update

henriqueiamarino:
Massive changes to the theme.
[boxedbio.zip](https://github.com/Automattic/themes/files/14955429/boxedbio.zip)


henriqueiamarino:
I will refrain from updating themes for a while. I'll be launching this new version in the Dotorg showcase anytime soon.

PR #7727, opened 2024-04-08
Update theme raw-update

henriqueiamarino:
[raw-update.zip](https://github.com/Automattic/themes/files/15073210/raw-update.zip)


henriqueiamarino:

PR #7730, opened 2024-04-09
Low Fi: Remove the space in the theme display name

iamtakashi:
This removes the space in the theme display name, but there is no change in the slug and the text domain.

The display name was `Low Fi`, and Dotorg wants the theme slug and the text domain to be `low-fi` with that name. Since naming changes in an already launched theme is a nightmare, I'm changing the display name to `LowFi` (removing the space between `Low` and `Fi`) so the current slug and text domain, `lowfi`, makes better sense.

### DIFF
--- a/lowfi/readme.txt
+++ b/lowfi/readme.txt
@@ -1,4 +1,4 @@
-=== Low Fi ===
+=== LowFi ===
 Contributors: Automattic
 Requires at least: 6.0
 Tested up to: 6.4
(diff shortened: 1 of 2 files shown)


mikachan (approved):
LGTM, I can't see any other references to "Low Fi" in the theme files 👍

iamtakashi:
Thanks, @mikachan, for the quick review. Merging is a click of a button, but should I worry about deploying this change to dotcom, too? Can't that be automated?

mikachan:
I didn't merge this immediately as it was late where I was and I didn't want to deploy in the evening. I'll deploy this now.

We're looking into how we can automate the deploy to dotcom. I wonder if this sits well with the Theme Launcher on dotcom - I've opened an issue for it here: https://github.com/Automattic/mc-theme-launcher/issues/2.

PR #7746, opened 2024-04-17
Update theme overlaid-update

henriqueiamarino:
[overlaid-update.zip](https://github.com/Automattic/themes/files/15009472/overlaid-update.zip)


henriqueiamarino:

PR #7754, opened 2024-04-18
Update theme screenplay-update

henriqueiamarino:
[screenplay-update.zip](https://github.com/Automattic/themes/files/15025335/screenplay-update.zip)


henriqueiamarino:

PR #7759, opened 2024-04-22
Prepare theme for WP.org submission

beafialho:


mikachan (commented):
I noticed all of the patterns in this theme don't have any outer spacing:

image

I'm guessing it's either missing from the theme.json settings or there's a missing Group block.

beafialho:
I believe I did this on purpose because, that way, a user would just be able to add the pattern to a page/post and it already has wide width. I just tested it on the demo site.

mikachan:
Ah OK, in that case I think everything looks good to merge!

mikachan (approved):

iamtakashi:
In this PR, some pattern files and images have been added to the root of the theme. It was an oversight, wasn't it? Otherwise, let me know. I'll clean up in a PR.

PR #7760, opened 2024-04-23
Radcliffe 2: Removes logo resize reset testing code

creativecoder:

Removes testing code that deletes the `logo_size` theme mod with a query param, as it's not needed any longer.

There's no other references to the `remove_logo_size` param I can find, and the theme and related customizer features are not being actively developed, so this code shouldn't be missed.

See p3btAN-2Id-p2 for more context.


mikachan (approved):
Thanks for this! I also can't see any other references to `remove_logo_size`, so this looks good to remove.

creativecoder:
Deployed to wpcom with D146349-code

PR #7768, opened 2024-04-24
Update theme.json

beafialho:

This fixes wrongly set default font "Rubik" and replaces it with the correct font "Albura".


mikachan (approved):
Thanks for the fix, LGTM!

PR #7783, opened 2024-04-29
Dara: move away from deprecated class

jeherve:

Jetpack's `Jetpack_User_Agent_Info` class was deprecated 4 years ago in https://github.com/Automattic/jetpack/pull/16129. It is now recommended to use its replacement, `User_Agent_Info`, also shipping with the Jetpack plugin.

This PR makes the change to avoid deprecation notices in the logs:

```
PHP Deprecated:  Function Jetpack_User_Agent_Info::is_ipad is deprecated since version Jetpack 8.7! Use \Automattic\Jetpack\Device_Detection\User_Agent_Info::is_ipad from the `automattic/jetpack-device-detection` package instead. in /srv/htdocs/__wp__/wp-includes/functions.php on line 6078
```

### DIFF
--- a/dara/inc/jetpack.php
+++ b/dara/inc/jetpack.php
@@ -70,10 +70,10 @@ add_action( 'after_setup_theme', 'dara_jetpack_setup' );
 /**
  * Footer widgets Callback for Infinite Scroll
  */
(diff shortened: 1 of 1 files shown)


coder-karen (approved):
Tests well (footer widget testing doesn't appear impacted), and deprecation notices are removed.

PR #7794, opened 2024-05-09
Cakely: Fixed submenu colours, and added a link to Midjourney's TOS page to readme.txt

iamtakashi:
* Fixed submenu colours. Also adjusted the distribution of the paddings for submenu.
* Add a link to Midjourney's TOS page to readme.txt

| Before | After |
| ----------- | ----------- |
| CleanShot 2024-05-09 at 17 00 47@2x | CleanShot 2024-05-09 at 16 40 42@2x |

#### Related issue(s):
https://themes.trac.wordpress.org/ticket/173805


madhusudhand (approved):
It fixes the issue in themes that uses default navigation menu. If user already changed to a different navigation and saved the changes, it keeps `overlayBackgroundColor` in the user changes and it would still be an issue.

Looks good for dotOrg. A css override may be required for dotCom (based on active installations)

iamtakashi:
@madhusudhand Thanks for your review. I appreciate it.

Yes, this only fixes it on new sites and the sites with the default. I'm not going to try fixing the existing sites because I think the bug was obvious for those who had a navigation with a submenu and they have already fixed it in the editor by themselves 🙏

PR #7806, opened 2024-05-15
Refactor dotorg deploy script to use an EXCLUSION list instead of INC…

pbking:
This script runs on every commit of trunk.  It attempts to push any version of themes that are NEWER in the github repo than are present in wporg.

Because this currently uses an INCLUSION list only the themes in that list are automatically updated on dotorg.  This has resulted in a lot of themes falling behind and necessitates a manual submission process.

This refactors the dotorg deploy script to 

* use an EXCLUSION list instead of INCLUSION list
* ignore ALL classic themes
* check existence of themes and version using HTTP instead of SVN

The result of this change would (immediately) push the following themes:

➤ Upgrading archivo/ from 1.0.4 to 1.0.6
➤ Upgrading bibliophile/ from 1.0.3 to 1.0.4
➤ Upgrading bsoj/ from 1.0.2 to 1.0.3
➤ Upgrading common/ from 1.0.4 to 1.0.5
➤ Upgrading freddie/ from 1.0.1 to 1.0.2
➤ Upgrading jinjang/ from 1.0.0 to 1.0.1
➤ Upgrading kigen/ from 1.0.2 to 1.0.3
➤ Upgrading loudness/ from 1.0.5 to 1.0.10
➤ Upgrading optimismo/ from 1.0.2 to 1.0.3
➤ Upgrading tomoni/ from 1.0.2 to 1.0.3
➤ Upgrading trellick/ from 1.0.2 to 1.0.4


deploy-dotorg.sh
```diff
@@ -11,96 +11,198 @@ if [[ -z "$SVN_PASSWORD" ]] && [[ $1 != "preview" ]]; then
...
-	"upsidedown"
-	"vivre"
-	"zoologist"
-)
+# Do things for all of the themes
+for THEME_SLUG in */ ; do
 
-for THEME_SLUG in ${THEMES_TO_DEPLOY[@]} ; do
+	# Skip any classic themes (assuming that none of our classic themes use a theme.json file)
+	if ! test -f "./${THEME_SLUG}/theme.json"; then
```
mikachan, line 40: As it's possible that classic themes may use a theme.json file, I'm wondering if it might be better to check for an `index.html` file instead?

deploy-dotorg.sh
```diff
@@ -11,96 +11,198 @@ if [[ -z "$SVN_PASSWORD" ]] && [[ $1 != "preview" ]]; then
 	exit 1
 fi
 
+# These themes either:
+# 	are not (and not expected to be) deployed to wporg
+# 	have been deployed to wporg, however they are not expected to be updated
+# Note that all classic themes (those without a theme.json file) are ALSO ignored
+declare -a THEMES_TO_IGNORE=(
+        'aldente'
```
mikachan: It would be great if @beafialho @henriqueiamarino and @iamtakashi could check this list, just in case we're excluding any themes that should be deployed to dotorg.

I think the following are possible candidates: foam, cortado, grammerone, hall, raw.

mikachan (commented):
Great idea! This works much better than an inclusion list, as it's so tricky to remember to update that list. I've left some small comments to check before we bring this in.

deploy-dotorg.sh
pbking: I THINK that I checked all of those and found that they hadn't been updated in > 2 years; that's why they are on the list.

Not ALL themes that haven't been updated in > 2 years are in that list... just themes that HAVE updates.

However the list gone through a few iterations so I'm not sure that I trust that any more.

deploy-dotorg.sh
pbking: Yeah, I guess that's what CORE is looking for to determine eligibility (though that also takes into account a theme's PARENT too right?  That would be harder to check and I'm not sure that it would matter for our purposes.)

mikachan: > though that also takes into account a theme's PARENT too right?

Yeah I think so, although wouldn't that be the case with a theme.json file too?

I would guess that all our block themes include a theme.json file so it's probably fine to use this, I was just being mindful of any future themes that may for some reason not include a theme.json file but do include an index.html file, following Core's rules for block themes.

deploy-dotorg.sh
mikachan: foam, cortado, grammerone, hall, and raw have all been added in the last year, and I can't see anything on the original "ready to launch" posts about them being dotcom-exclusives. But I'm also not 100% sure of the intention with these themes, which is why I thought it might be best to check in with the designers first.

deploy-dotorg.sh
```diff
@@ -11,96 +11,198 @@ if [[ -z "$SVN_PASSWORD" ]] && [[ $1 != "preview" ]]; then
...
+        'twentytwentytwo-mint'
+        'twentytwentytwo-pink'
+        'twentytwentytwo-red'
+        'twentytwentytwo-swiss'
+        'verso'
+        'vetro'
+        'winkel'
+        'xanadu'
+)
+
```
vcanales: We could read ignored themes from a file, such as `.dotorg-ignore`. This follows conventions — using files to define ignored paths — and makes maintenance easier.

```suggestion
# Get the directory of the script
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
IGNORE_FILE="$SCRIPT_DIR/.dotorg-ignore"

# Initialize an empty array for themes to ignore
THEMES_TO_IGNORE=()

# Read themes to ignore from .dotorg-ignore file, ignoring comments
if [[ -f "$IGNORE_FILE" ]]; then
    while IFS= read -r line || [[ -n "$line" ]]; do
        [[ "$line" =~ ^#.*$ ]] && continue
        THEMES_TO_IGNORE+=("$line")
    done < "$IGNORE_FILE"
else
    echo ".dotorg-ignore file not found, proceeding without ignoring any themes."
fi
```

mikachan: +1 to this, it would be easier to maintain this way.

deploy-dotorg.sh
vcanales: I think we can follow-up with feedback on the definitive list; for now we would benefit from focusing on getting the mechanism in, so that we can maintain that list from an ignore file.

vcanales:
I've refactored the script to read ignore paths from a `.dotorg-ignore` file. I also refactored `rsync` commands that were using a separate `dotorg-excludes.txt` file, to use the same file as the theme deployment.

I think it makes sense to keep the list consolidated, with both general paths and theme paths that we don't want to include in our dotorg deployments.

deploy-dotorg.sh
mikachan: Sounds good 👍

mikachan (commented):
Thanks @vcanales.

> I think it makes sense to keep the list consolidated, with both general paths and theme paths that we don't want to include in our dotorg deployments.

Agreed, I think a single list here works best.

I've tested this locally (using incorrect SVN credentials to avoid committing changes) and it looks like it'll work well. Should we bring this in to test it out?

vcanales (approved):

PR #7808, opened 2024-05-16
Update theme otis

beafialho:
Changes from WordPress Playground, to improve the theme and prepare for Dotorg submission:

- Set templates to use font size presets instead of local customizations
- Added font license
- Fixed some spacings that weren't optimized before


beafialho:
Some changes were missing, so I closed this one and opened #7809

PR #7809, opened 2024-05-16
Otis: Update theme and prepare for Dotorg submission

beafialho:
Changes from WordPress Playground, to improve the theme and prepare for Dotorg submission:

- Set templates to use font size presets instead of local customizations
- Added font license
- Fixed some spacings that weren't optimized before


mikachan:
👋 @beafialho It looks like this style.css file is missing a theme URI, which will be needed for the dotorg submission. Do you know what the URI is for this theme?

beafialho:
Thank you @mikachan! I added it here: https://github.com/Automattic/themes/commit/d82db8e14a7b22ed06419154f797a5c686d6a1df

mikachan:
Thanks @beafialho! This looks good to bring in now.

> I added it here: https://github.com/Automattic/themes/commit/d82db8e14a7b22ed06419154f797a5c686d6a1df

It looks like this was committed to the main branch, `trunk`. For future changes, it's best to commit to the branch of the specific PR rather than `trunk`. In this case, the branch is `playground-changes-2024-05-16T11-28-25-343Z`.

mikachan (approved):

PR #7829, opened 2024-05-24
Add version bump workflow

vcanales:
#### Todo
- [x] Make it possible to bump versions without connecting to the sandbox.
- [x] Create a PR with the automated changes.
- [x] Write proper PR title and message

#### Changes proposed in this Pull Request

Adds a GitHub workflow that makes it possible to bump theme versions, as well as the repo's tag, on-demand.

It was necessary to refactor the function that executes the version bump to use the latest tag in the repository instead of connecting to the sandbox to fetch the last deployed hash.

#### Review
I'd like input, in particular from @pbking, to figure out if making this change will affect the .com part of the deployment in any way.

I believe that the `versionBumpThemes` function does not impact what gets deployed to .com, because the check for changes is ran against the `hash` stored in the sandbox and not locally.

### DIFF
--- a/.github/workflows/version-bump.yml
+++ b/.github/workflows/version-bump.yml
@@ -1,0 +1,132 @@
+name: Version bump themes
+on:
+  workflow_dispatch:
+    inputs:
+      theme-slug:
+        description: 'Theme slug'
+        required: false
+        default: 'all'
+  
+jobs:
+  version-bump:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          ref: trunk
+      
+      - name: Create branch
+        id: create-branch
+        run: |
+          BRANCH_NAME="automated-version-bump/${{ github.run_number }}"
+          git checkout -b $BRANCH_NAME
+          echo "BRANCH_NAME=$BRANCH_NAME" >> $GITHUB_ENV
+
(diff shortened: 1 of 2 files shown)


pbking:
> I believe that the versionBumpThemes function does not impact what gets deployed to .com, because the check for changes is ran against the hash stored in the sandbox and not locally

Yes, that's correct.  For this reason you can't version bump a theme manually (for instance changing the major or minor version outside of the automated version bumping, etc).  So however we change the version bumping actions it shouldn't effect the wpcom deployment behavior.

It looks like using the last tag is a fine way to determine what might need to be version bumped, and probably a better measurement that "whatever version has been last deployed to wpcom".

pbking (approved):
I ran this on a fork and the action worked as expected.

I ran the action locally and it integrated with the automated deployment as it did before the changes.

I think that we can bring this in and give it a try.

mikachan:
Is this ready to merge and try out? 👀

vcanales:
> Is this ready to merge and try out? 👀

I'm figuring out a bug with the PR body, where for some reason the themes that have been changed are not being included. Working on that on my fork so that I can test the workflow — https://github.com/vcanales/themes

pbking:
This continues to work as expected (both the action as ran on my fork and locally).  I'm not sure what error was fixed but you mentioned it was to note what themes changed and I see that it is indeed happening.

I think this is ready to come in and give it a few runs in the real world.

PR #7831, opened 2024-05-24
Tenaz: Use one query block per section in front page template

creativecoder:

- Simplifies the front page template of the Tenaz theme to use one query loop block per section
- Uses CSS to replicate that layout

| **Before** | **After** |
| - | - |
| image | image |

### DIFF
--- a/tenaz/patterns/latest-posts.php
+++ b/tenaz/patterns/latest-posts.php
@@ -10,59 +10,28 @@
 	<!-- wp:group {"align":"wide","style":{"border":{"bottom":{"color":"var:preset|color|contrast","style":"solid","width":"3px"}},"spacing":{"margin":{"bottom":"var:preset|spacing|50"}}},"layout":{"type":"default"}} -->
 	<div class="wp-block-group alignwide" style="border-bottom-color:var(--wp--preset--color--contrast);border-bottom-style:solid;border-bottom-width:3px;margin-bottom:var(--wp--preset--spacing--50)">
 		<!-- wp:heading {"style":{"typography":{"textTransform":"uppercase"}},"fontSize":"medium"} -->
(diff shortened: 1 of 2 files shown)


iamtakashi:
Interesting! I don't know how I feel about the necessity of using CSS to control the font size and hide the excerpt, though. Having a single query loop block allows pagination to work, but I'm not sure if we want to pay that price to have a pagination in this specific theme.

Also, on the front of the site, I can see one extra post appears. It seems like the number of sticky post is added to the number of posts specified in the query loop block. This isn't caused by this PR, but it surfaced the issue.

CleanShot 2024-05-28 at 18 44 48@2x

creativecoder:
Thanks for taking a look @iamtakashi !

This is definitely an experiment, and I'm also unsure it's worth the tradeoff of simplifying the block layout for the template. As you point out, hiding the excerpt feels particularly hacky.

Another idea that occurred to me is to enhance the Query Loop block so that it can support multiple Post Template child blocks. This way you could apply different layouts to different posts in a single loop (the Post Template block would need some way to target a subset of the posts query, maybe something like the CSS nth-child syntax). But that would add a lot of complexity, so again I'm not sure it's a worthy tradeoff.

I'll open an issue or discussion in Gutenberg and see if there are other ideas for how the Query Loop could better handle multiple layouts for a single query.

With a quick search, I did find another idea: [Query Loop Block: Adding Multiple Query Loop Blocks with the option to not repeat posts already shown by previous Query Loop on same page #40903](https://github.com/WordPress/gutenberg/issues/40903)

creativecoder:
I converted this to a draft to indicate it's still experimental.

creativecoder:
Closing this until there's a better solution on the editor side.

PR #7842, opened 2024-05-31
Assembler: Add support for upcoming section style variations

richtabor:
Note that not every theme style variation has a "Dark 2" style (intentionally). Some themes are more minimal than others.  

## Testing 

1. Activate the theme. 
2. Add three patterns. 
3. Add a different section style to each one.
4. Switch to different style variations in the site editor. 
5. See changes. 

## Visuals 
Blueberry: 
CleanShot 2024-05-31 at 17 19 12

Tangerine: 
CleanShot 2024-05-31 at 17 19 58

Mauve: 
CleanShot 2024-05-31 at 17 19 47

Linen: 
CleanShot 2024-05-31 at 17 19 36

Rose: 
CleanShot 2024-05-31 at 17 19 00

Pearl: 
CleanShot 2024-05-31 at 17 19 23

Lilac: 
CleanShot 2024-05-31 at 17 21 33


richtabor:
> You can preview these changes by following the links below:

* with the next version of Gutenberg.

richtabor:
While running Gutenberg trunk (and with the next version of Gutenberg) you will see these block style variations on Group, Columns, and Column blocks. 

CleanShot 2024-05-31 at 17 28 31

richtabor:
I've updated all the variations with section styles, added a few, removed a few. 

I'm currently using `light`, `dark`, and `light-2`, etc. I'm thinking about just using something generic, like `section-1` (resolving to `is-style-section-1`. As not every one has every type of section style (light, light 2, dark, dark 2, and bright).

apeatling:
I think I prefer the section-N naming for classnames, although I don't have a strong feeling on it.

richtabor:
The tricky part (and maybe it's ok) is that the UI will render "Section 1" rather than "Light". The titles for each variation do not get applied when color-only variations are selected. 

Perhaps that can be tweaked on the Gutenberg side though.

miksansegundo:
How are you testing these? I'll review this in-depth tomorrow.

I'm guessing we can test this on the editor, the Dotcom onboarding for themes without style variations, and Assembler flow.

richtabor:
> The tricky part (and maybe it's ok) is that the UI will render "Section 1" rather than "Light". The titles for each variation do not get applied when color-only variations are selected.

I'm leaning towards using more generic style names, like "is-style-section-1", etc. 

It's a bit odd in the short term, while these are expressed as buttons with text labels, but in a near iteration we will depict these visually with the actual colors of the variations—so you see each one before you apply the style. 

Also, it resolves the current clunkiness around primarily dark variations, where "is-style-dark" and "is-style-dark-2" can only be darker than the default (reducing the amount of design variance). To have as much variance as the primarily light variations, we need to use something more generic.

richtabor:
> How are you testing these? I'll review this in-depth tomorrow.

Activate the theme. 
Add three patterns. 
Add a different section style to each one.
Switch to different style variations in the site editor. 
See changes.

miksansegundo (approved):
Looking good to me 👍 


https://github.com/Automattic/themes/assets/1881481/ff47f78f-57c5-48fc-9277-843b5c420dd9



The generic titles and classes are clear, and the [reasons](https://github.com/Automattic/themes/pull/7842#issuecomment-2158449498) make sense.  

Screenshot 2567-06-11 at 18 43 37


I would remove the spacer in the Footer template part that is out of the container ([here](https://github.com/Automattic/themes/blob/de77f898535b35f5bc0ba4331fb342d8ba823848/assembler/parts/footer.html#L1)).

Screenshot 2567-06-11 at 18 52 36

That spacer shows the background color of the global style variation while the Footer container modifies the background using `#B6B6B61F` [here](https://github.com/Automattic/themes/blob/de77f898535b35f5bc0ba4331fb342d8ba823848/assembler/parts/footer.html#L6C78-L6C87).

Screenshot 2567-06-11 at 18 51 32

Screenshot 2567-06-11 at 18 51 48

richtabor:
I was thinking of removing that footer color too, and assigning one of the section styles instead as a default.

richtabor:
Looks good, let's merge.

PR #7870, opened 2024-06-18
LeanCV: add theme

henriqueiamarino:
LeanCV is a blog theme that echoes the structure of a professional profile with original visuals and interesting navigation. It's suitable for the general public to display information, experiences, and education. And it's super easy to customize.


jasmussen:
Theme looks good! I noticed a scroll jump for the 100vh sticky image:
 
scroll

But this appears _only_ when the adminbar is present. So that's okay, you the sticky is actually a good solution to this.

---

A couple of visual bits of feedback. The block gap in this stack looks a little tight: 
Screenshot 2024-06-25 at 09 57 49

It's not quite big enough to look intentional, and bumping it just a bit would help. The spacing feedback applies to some of the rest of the page too:
Screenshot 2024-06-25 at 09 57 58

That may be a personal opinion, but just bumping the gaps a little bit might aid legibility. Here's a quick test:
Screenshot 2024-06-25 at 10 02 09

Not strong opinions, let me know if it resonates.

---

Templates look good. I think across all the theme work, it might be good to do a separate exercise to improve our "no search results found" text. While it can be different from theme to theme, in most cases IMO it can be the same. And these two read a little flowery to me:
Screenshot 2024-06-25 at 10 05 16

Screenshot 2024-06-25 at 10 05 47

Something simpler like "No results found." and "You can try a search:" might be better for both cases. I don't think you necessarily have to address that for this theme, but it's something to generally consider across themes. CC: @beafialho.

---

Style variations look good. Blue is a good professional default, but the red wine and forest green look great too. Can we find a different name than vampirella?
Screenshot 2024-06-25 at 10 06 13

Vampirella is a trademark! 😅 

---

In a quick code review, I found this lovely picture:

Screenshot 2024-06-25 at 10 04 44

Is that used anywhere? Should it be? I assume the featured person is generated using Midjourney. Same feedback as for "Outnow" for the readme, would be good to change 

> All images used in the templates were generated with AI using Midjourney.

to

> Generated with AI on Midjourney, released under the CC0 license

So in summary, looks good, nice CV theme. Let's get the style variation renamed, the README updated, check if the mary shelley picture needs to be there, and consider whether to do any spacing changes or not, and then this one is good to go. Nice work!

henriqueiamarino:
Thanks, @jasmussen. The demo site is live [here](https://leancv.mystagingwebsite.com/). I conducted the following changes:
- Removed the Mary Shelley portrait from the assets. It was the V1 launched in Dotcom, but Ian required a contemporary photography to make the theme more appealing;
- Made spacing more abundant;
- Fixed no-results text;
- Renamed style variations.

henriqueiamarino:
@jasmussen I closed this PR as I can't revert a commit. I created a new one: #7893

PR #7878, opened 2024-06-20
Moviescript: add theme

henriqueiamarino:
Moviescript is a simple blog theme that perfectly suits writers or the general public who want their sites to be formatted as movie scripts.
[Demo site](https://moviescript.mystagingwebsite.com/)

screenshot


MaggieCabrera:
This theme has a hidden pattern called header-rate that doesn't seem to be used anywhere, can you confirm if it needs deleting, @henriqueiamarino ?

MaggieCabrera (changes requested):
The heading hierarchy in this theme is hard to maintain. We generally don't want the site title to be the H1, since it's preferable to have the main heading be something that tells you what each template/page is about, ie: the post title on a single post or the archive title on archive templates, instead of being the same for every single page. But this theme lacks that main heading in many of the templates like the pages template and the index

henriqueiamarino:
Thanks for your review, @MaggieCabrera. Sometimes, I add a pattern that some users might use, but I don't use it in my take on the theme. I never really knew if that was okay or not. If it's not a good practice, you can delete it; no problem.

I am afraid I have no control over the site title as an H1. As I design in the Editor, that is how the site title is available to me; I have other current choices. Also, I can add a title for the Page template; it's no problem. I can also make post titles H1 instead of H2 for the Index. Would that be enough? Working on it.

henriqueiamarino (commented):
Changes approved. I tried to access this branch using `git checkout add/moviescript` but apparently my side is outdated. Maybe this is due to the fixes you made. How can I check properly to make the corrections you mentioned? Sorry, I am a newbie using GH, but I am eager to get good at it.

Screenshot 2024-08-30 at 16 14 21

MaggieCabrera:
> Thanks for your review, @MaggieCabrera. Sometimes, I add a pattern that some users might use, but I don't use it in my take on the theme. I never really knew if that was okay or not. If it's not a good practice, you can delete it; no problem.
> 

It's ok to have but since it has the hidden attribute, it won't show anywhere in the UI for users! I will change that to make it available

MaggieCabrera (approved):
Pushed a couple of changes to show the hidden header/footer patterns and some translation and a11y stuff. Good to go!

PR #7879, opened 2024-06-20
WP-DOS: add theme

henriqueiamarino:
WP-DOS is a blog theme designed for nostalgic ones. It is a tribute to the folks who invented computing as we know it today. It is an exciting opportunity to test blogging as if we were back again to the green (or amber) phosphor.

[Demo site](https://wpdos.mystagingwebsite.com/)

screenshot


henriqueiamarino:
The `background-image` inside the Cover in Home is not showing, and I can't fix it.
I ask the reviewer to check it, please.

jasmussen:

Love it, looks great. I particularly like the use of a plus grid for the metadata, and the glitchy background:

Screenshot 2024-08-16 at 09 19 13

I'm not seeing that on the demo site, btw—is that what you're referring to with the cover background?

Mobile looks great, fonts look great. The indentation for quotes may be slightly large:

Screenshot 2024-08-16 at 09 21 21

Maybe it doesn't even need any indentation to the left of the border?

The pages can get pretty side:

Screenshot 2024-08-16 at 09 21 39

I wonder if you can't add some of the page metadata that's present on blog posts? It can even be as little as the plus grid, shown on the left:

Screenshot 2024-08-16 at 09 22 40

That would let pages have the same centered look as posts, and I think it would work great.

There's a question of what that would do to full-wide images. But I guess that would already be a question for posts. So maybe best to just sync up there. 

**Template review**

Templates look great:

Screenshot 2024-08-16 at 09 24 17

Keep the ascii four-oh-four! Even if it might glitch on Mobile! This is a glitch theme after all.

I wish we could do more with the form widgets:

Screenshot 2024-08-16 at 09 25 17

This would look much better if it was just the green border, green text, and black background. But this is a limitation in the block editor, I'm aware. And if we write some local CSS to override, it'll likely break for the style variations.

**Style variations**

Speaking of, they look great! 

image

I specifically appreciate that you included the two classic monitor shades, as well as a fully neutral monochrome version.

For this theme it's okay the font doesn't change, it truly is a specific theme that might fall apart otherwise. At the same time, it's such a nice layout, it'd be nice with a neutral style that wasn't quite as programmery. Light background, sans-serif font, plusses-aligning-be-damned. 

**File and readme reviews**

Looks like the README needs a Public Domain notice for the background texture. Other than that, looks good.

Font names were messed up, unfortunately:

Screenshot 2024-08-16 at 09 30 01

There's a double comment here:

Screenshot 2024-08-16 at 09 30 39

You mention a cover background image, but in my quick test I see you using the Cover block mainly to afford 100vh height to the content, and that the background is smartly set in Global Styles:

Screenshot 2024-08-16 at 09 34 18

So perhaps that's fine?

**Summary**

Let me know what's up with the missing Cover background, but otherwise, this looks gready to ship after you've renamed the fonts and tweaked the README.

henriqueiamarino:
Thanks, @jasmussen. As I can see from your screenshots, the background is now showing great. I'll iterate it following your general feedback.

wp-dos/theme.json
```diff
@@ -0,0 +1,582 @@
...
+			]
+		},
+		"useRootPaddingAwareAlignments": true
+	},
+	"styles": {
+		"background": {
+			"backgroundImage": {
+				"id": 1290,
+				"source": "file",
+				"src": "file:./assets/images/dos_noise-texture.png",
```
jasmussen, line 158: Just noting, the background is defined here, so this looks fine.

henriqueiamarino:
@jasmussen I fixed everything you mentioned, but I couldn't figure out a better way to set an inverted (white to black) style variation. I created it, but as the background image is not affected by duotones, it doesn't change. Should we keep it like that and make it clear on the page showcase for users to replace the bg image? Or is it better to remove it?

In time, I tried an image based on transparency, but the effect didn't work anyway.

jasmussen:
I see, yes, the light style is difficult with the dark background. One option would be if the PNG for the background had transparency, rather than a black fill. Or perhaps that transparency is already present, but the duotone just makes it black? In that case, perhaps it might work if there's a solid fill instead of a transparent background? If it's easy to test, see if it works. But otherwise, it's fine to launch this without the light style! Thanks for trying it.

henriqueiamarino:
I really tried, and I am ready to keep trying; I love this kind of challenge. What I learned is that duotones do not affect our background image, even if they are applied to all images in Global Styles. Strange, right? Do you think I should create an issue requiring a fix?

This is the transparent image I used. It should work with a simple Black to White duotone:
dos_noise-texture

jasmussen:
It sounds like it's non-trivial. Skip the light theme, it's fine!

I am also realizing in this, that you can't apply duotone to the site-global background image, I believe the filter gets inherited for all contents inside, unless separate mechanisms are applied. We can't have that.

henriqueiamarino:
Ok, it's done. Font and style variation removed; thank you.

PR #7881, opened 2024-06-20
Excerpts: add theme

henriqueiamarino:
Excerpts is a text-only blog theme that trusts post excerpts rather than titles and featured images and has many variations.
[Demo site](https://excerptsdemo.mystagingwebsite.com/)

screenshot


alaczek:
I'm seeing a few UX issues with this theme:
* Site title isn't displayed anywhere, while the nav is used twice in the header.
* The header is invisible on internal pages, and lacks left/right spacing:
image

* the info about theme download in the footer doesn't make sense from the customer point of view. While it's true that customers can remove this via the editor, we also know it's quite hard to navigate for inexperienced users. We can keep that info on the demo page, but I recommend removing it from the theme.

* the footer has a broken block in the editor:
image

* single posts have a different sidebar content than other templates - was this intentional?
* sidebar.php has a hard-coded link that won't work, unless customer site happens to have a page with the slug `posts`: `<p class="has-text-align-right has-medium-font-size"><?php echo __('<a href="/posts/">All articles</a> ↘', 'excerpts');?></p>`
* while I understand the choice not to display post titles on the homepage, not including them on single posts feels like a bug.
* in the post editor the content has no left/right spacing:

image

henriqueiamarino:
I committed all the necessary changes and added some diversity to the existing style variations.

Later, I noticed an **Excerpts** staging site, which may have already been submitted to Dotor. To avoid a naming conflict, I am closing this PR for the benefit of #8067. It was renamed, and it's now **Titleless**.

PR #7887, opened 2024-06-21
Pierian: add theme

henriqueiamarino:
Pierian is a minimalist typography-focused theme that embraces a low-key styling approach. With equivalent headings, paragraphs, links, and navigation, it features columns that elegantly divide navigation, blog posts, and a profile block.
[Demo site](https://pieria.mystagingwebsite.com/)

screenshot


alaczek:
A found a few issues  with this theme:
* the header.html template part is not used anywhere, and contains empty blocks only.
* looking at the design the left side feels like it should be a header template part, and reused across all templates (at the moment that part is hard coded in each template). 
* The same thing applies to the right side - it's hardcoded in all the templates. A pattern called sidebar.php exists and it should be used across the board instead. Addressing this is important - otherwise customers would have to update each and every template when making changes to either the left or right sidebar.
* If you address the two issues above you will be able to to use templates, as opposed to making every template content a pattern (for example archive.html template contains reference to pattern like so `<!-- wp:pattern {"slug":"pierian/archive"} /-->`, and ideally the template itself should include header pattern, content of the template, and the sidebar pattern.

MaggieCabrera (approved):
I made the changes from the review plus a few more that I found myself, this should be good to go

henriqueiamarino:
Thanks, @MaggieCabrera. I'm merging it, then.

henriqueiamarino:

[pierian.WordPress.2024-09-12.xml.zip](https://github.com/user-attachments/files/16982239/pierian.WordPress.2024-09-12.xml.zip)

PR #7888, opened 2024-06-21
Message: add theme

henriqueiamarino:
Message is a theme that brings the mobile messaging experience to your WordPress site.
[Demo site](https://messagerie.mystagingwebsite.com/)

screenshot


alaczek:
I found a few issues with this theme:
* query loop in the front page template is set to custom, and won't follow posts per page settings
* this section in the front page template feels excessive for a blog theme. I appreciate you wanted to educate the customers, but these aren't particularly instructive. We know interacting with the editor can be hard for customers - we shouldn't include sections that have to be manually removed:
image

* the theme includes both home and front-page templates, that are virtually the same. Blog theme doesn't need both.
* footer has the word "Seen" - what is the purpose of that?
image

henriqueiamarino:

[message.WordPress.2024-09-12.xml.zip](https://github.com/user-attachments/files/16982212/message.WordPress.2024-09-12.xml.zip)

PR #7889, opened 2024-06-21
Feature: add theme

henriqueiamarino:
The Feature is a magazine theme focused on sizable typography and imagery to grow your blog posts, reviews, artwork, and news.
[Demo site](https://feature.mystagingwebsite.com/)

screenshot


alaczek:
I found a few UX issues, similar to the Excerpts theme:
* site title isn't included in the header.
* the info about the theme and download in the footer doesn't make sense from the customer point of view. While it's true that customers can remove this via the editor, we also know it's quite hard to navigate for inexperienced users. We can keep that info on the demo page, but I recommend removing it from the theme.
image
* Can the page-about.html template can be removed?

MaggieCabrera:
Is it ok to submit themes to .org using 3rd party blocks on the templates instead of the patterns? I don't have jetpack locally and I see a placeholder prompting me to install instead. I think this doesn't block the submission, but wanted to point it out

MaggieCabrera (approved):

henriqueiamarino:

[feature.WordPress.2024-09-12.xml.zip](https://github.com/user-attachments/files/16982287/feature.WordPress.2024-09-12.xml.zip)

PR #7909, opened 2024-07-01
Hola: add theme

henriqueiamarino:
**Hola** is a text-only blog design, a new interpretation of the theme **Hey** with diverse style variations.
[Demo site](https://hola.mystagingwebsite.com/)

screenshot


henriqueiamarino:
@jasmussen I am pointing out some themes that might be quicker to review. This one is a good example.

jasmussen:
Thanks for that! 

As usual, it is becoming a running joke that I'm unable to not comment on details I spot as I see them. I will continue to defer to you on whether they are necessary to fix in each theme here, or whether the feedback can be absorbed for future themes. So for all the visual pieces, no blockers. 

In this case, nice and simple theme. I like the very simple footer, it's seductively simple:

Screenshot 2024-07-04 at 11 15 47

The outer "border" is also a nice touch, I suspect that detail can work in varous ways for other themes.

The top padding above the header here is nice and spacious, but the padding above the pagination is smaller, can they be the same?

Screenshot 2024-07-04 at 11 16 16

Type-spacing here looks good. There's something about the font sizes, and font weights:
Screenshot 2024-07-04 at 11 16 47

"January 18, 2024" is 17px, and the intro paragraph is 20px. They're pretty close in font size, which for me makes the slight difference in _font weight_ stand out. I could see this going a few ways. One option, simply have the same font size for the datetime:

Screenshot 2024-07-04 at 11 19 05

Or, make the font size smaller and make the weight a little thicker. Here's 14px, and font-weight: 600;

Screenshot 2024-07-04 at 11 19 56
—the above is not quite right, ideally it would be font-weight 500, with the larger text being 400, and ideally the two of them feel like they are the same font weight across the font size differential. Make sense?

The site logo looks good. I'm assuming that it's pushed downwards a bit is intentional:

Screenshot 2024-07-04 at 11 21 24

If intentional, I'd lean in a bit, push it down slightly more, so it looks more intentional. There's also an issue here with the out of box theme (not the demo site). Here, the site title sits pretty close to the edge of the blue header:

Screenshot 2024-07-04 at 11 22 39

I'm guessing the offset image and the trick necessary to make that happen makes this tricky. If you can't make it work, I wonder: would the theme still work if the full site logo was inside the blue header, without sticking out at the bottom? 

The blue border does present a brain-scratcher on short pages:

Screenshot 2024-07-04 at 11 26 11

Is it possible to put all the templates inside a container that has a 100vh min-height, so the bottom blue border always touches the bottom of the browser?

**Template review**

Templates look good, mobile works well.

**Style variations**

I like that you pushed the style variations a bit on this one! This font is fun:

Screenshot 2024-07-04 at 11 27 41

**File and readme reviews**

Readme looks good!

**Summary**

There's a lot of feedback here. Let me know what category this all lands in, as far as how easy or hard it might be to address. I don't know that any of it is blocking, perhaps only the part about the header text when there's no site logo would be good to think about.

henriqueiamarino:
Thanks, @jasmussen. The more feedback, the better. This theme will be adjusted with my new way of dealing with type scales. I learned it from Takashi and have applied it to all new themes unless I want something unique.

The header and site logo are tasty details, but they were built in a tricky way. I'll see what I can do to fix that.

henriqueiamarino:
I am closing this PR as I commit unwanted changes.
I've cloned the Trunk a couple of times today already, but every time I create a PR, there's a risk of it needing to be updated. **I am still trying to figure out how to fix this, so I'll create a new PR.**

PR #7863, opened 2024-06-17
MyMenu: add theme

henriqueiamarino:
My Menu is a simple theme designed to facilitate restaurant owners’ site-building experiences. It is clean, direct, and customizable. Test the cool style variations that have been added to the theme.
[Demo site](https://mymenu.mystagingwebsite.com/)

screenshot

PR #7920, opened 2024-07-03
Specials: add theme

henriqueiamarino:
A clean, beautiful, and customizable theme for restaurant menu websites.

[Demo site](https://specials.mystagingwebsite.com/)

screenshot


jasmussen:

Delightful monochromatic theme with good use of duotone photography. It's also nice to see something east-asian as far as demo content. 

Visually it is a bit close to [MyMenu](https://github.com/Automattic/themes/pull/7863), not quite close enough to be a style variation, but nearing that point. There's something quite bold about this large heading and the whitespace, that could make for some interesting layouts in various places:

Screenshot 2024-07-15 at 14 57 41

Speaking of, it says "Specials", but I was expecting a section of menu demo content below that. Is something missing there?

It's nice to use named HTML anchors for jumping to sections of the page. I wonder, should there be a "back to top" link somewhere at the bottom? You can add one by simply linking to `#`:
Screenshot 2024-07-15 at 14 55 25

Not a blocker, but a consideration for future themes! It's a nice trick and it doesn't require a custom block.

There's a small space between major sections here:
Screenshot 2024-07-15 at 14 56 29

If that's intentional, it should probably be a bit larger, perhaps? 

I'm seeing an issue with some links in the footer:

Screenshot 2024-07-15 at 14 58 59

They appear to have broken. 

I'm also wondering if the border should be around teh whole footer box, or not, there's something unfinished feeling about it as-is. Also not a blocker, mainly a thought.

I wonder if we can tweak this text a bit, the grammar seems a bit off:

> A menu theme designed for those people who love to eat, which are always the best.

Perhaps:

> Discover authentic Chinese cuisine, crafted for food lovers.

Mobile looks fantastic, the theme feels made for it, which is reasonable given it's a restaurant site.

I wonder: should we put an address and a phone number at the bottom? Normally I'd scroll to the footer for a restaurant website for those things specifically. They may take the place of Tumblr and Instagram?

**Template review**

404 page could use some text tweaks:

Screenshot 2024-07-15 at 15 03 53

Perhaps:

> Sorry!
> We are apparently out of that item, or the page you are looking for doesn't exist. What do you think about a new selection? 

The post meta includes a $ sign:

Screenshot 2024-07-15 at 15 05 56

It's a subtle line between helpful demo content and going a step too far. I think in this case it might be better to just omit that $.

**Style variations**

These are fantastic. I particularly like the font and stark colors of Black Void:

image

Chewing Blue, Dark Caramel, Nautical Frost are nice too.

Red star is a favorite of mine because it's so out there, ketchup and mustard!

image

It's too crazy to be default, but it's definitely opinionated, I love that. However perhaps we should call it something other than Red Star. Perhaps:

* Fireburst
* Fireworks
* Solar Flare
* Radiant Ember
* Sunflare
* Sunburst

**File and readme reviews**

Files look good.

**Summary**

This is a nice one. The style variations, especially, give this one potency, very nice work. There are a couple of technical followups, but this one nicely bookends the category of single page vertical menu themes.

henriqueiamarino:
Thanks, @jasmussen. The big 'Specials' word should be the restaurant logo before the footer block. I used an image because we can't do that with text right now, but we may be able to in the future.
Anyway, I fixed the footer, and it's now quite different, following your ideas.

jasmussen:
> The big 'Specials' word should be the restaurant logo before the footer block.

Oh! Makes sense then. All good.

henriqueiamarino:
@jasmussen Anyway, I've changed it, and it's now the site title in a considerable size, using VW.
I also added an address and a back-to-top link.

PR #7923, opened 2024-07-03
LeanCV: Use shared color variations

richtabor:
Exploring sharing variations between Assembler and other block themes. 

I picked a few [color variations from Assembler](https://github.com/Automattic/themes/tree/trunk/assembler/styles/colors) that closely resembled the original ones, and made a few slight adjustments to the duotones to allow for better interoperability across styles/themes.  

Also made a few changes to markup while I was in there. 

https://github.com/Automattic/themes/assets/1813435/f3745a4b-f7c3-4c50-964f-72287019b3cc


henriqueiamarino (commented):
Everything is fine with the variations, they are good to go @richtabor. Thank you.

PR #7924, opened 2024-07-03
Epi: Use shared color variations 

richtabor:
Add shared color variations from Assembler. 

https://github.com/Automattic/themes/assets/1813435/c1fa2bd3-fd6c-43b5-8e81-bcaf8c166977


richtabor:
Note I've added a duotone color to the images, to showcase the duotone changing.

PR #7733, opened 2024-04-11
Update theme George Lois to Maverick

henriqueiamarino:
Massive changes to the current theme. And a necessary update to the title.
Please change it from **George Lois** to **Maverick**.
[maverick.zip](https://github.com/Automattic/themes/files/14955436/maverick.zip)

This issue is a new version of an existing one, #7592, which is now closed.

PR #7933, opened 2024-07-05
Try: George Lois theme: Delete the whole thing.

jasmussen:

Alternative to #7733. Deletes the George Lois theme, with the intent to add a different-name version of the same.


jasmussen:
Let me know if I did this right Sarah/Henrique!

mikachan (commented):
I can see this removes the `george-lois` theme directory, which is correct if we want to remove this theme entirely.

However, I can see the George Lois theme still exists on dotcom, so we probably need to coordinate there first before deleting the theme from this repo. cc @miksansegundo

miksansegundo:
Hi, I believe we have to retire the theme from Dotcom first. @alaczek, will you create a task for retiring it?

cc: @dsas I don't have experience retiring themes. There is a guide at PNEWy-tR-p2

dsas:
@miksansegundo agreed that we need to retire the theme but we don't need to block this PR to do it. The sync script won't delete the theme directory from dotcom, even though it gets removed in this repo (we should carefully double check this when syncing though).

jasmussen:
Thanks for the input. What's the next step here, then? Delete the folder? Maverick is already present, same theme: https://github.com/Automattic/themes/tree/trunk/maverick

miksansegundo:
Yes, it seems that you can merge this PR safely and later we'll retire the theme from Dotcom.

jasmussen:
Cool, thanks for the help. @henriqueiamarino should we merge this one and close #7733?

henriqueiamarino:
Ok, thanks @jasmussen. I'm merging it.

henriqueiamarino (commented):
Everything is fine here, merging the PR.

PR #7955, opened 2024-07-17
Sprinkler: add theme

beafialho:
Add theme via WordPress Playground.

**WARNING from Theme Check plugin**

Found `="<?php esc_html_e` in patterns/front-page.php. Use `esc_attr_e()` inside HTML attributes, and `esc_url()` for link attributes. A manual review is needed.

`Line 111: <div class='wp-block-cover alignfull has-tertiary-color has-text-color has-link-color' style='padding-right:var(--wp--preset--spacing--50);padding-left:var(--wp--preset--spacing--50)'><span aria-hidden='true' class='wp-block-cover__background has-black-background-color has-background-dim-30 has-background-dim'></span><?php esc_html_e(' src='<?php echo esc_url( get_styles`


MaggieCabrera:
I fixed the warning on this one. I'm reviewing it and so far I've found that the post title has no space below it:

Screenshot 2024-08-29 at 16 24 40

This is because we've reset margin bottom to 0. I understand the reason, but this change affects the title on the post editor. It's better to just tweak the template (in this case by removing the specific margins of the group that contains the title) and leave some actual margin on the block. I will change this while maintaining the spacing on the template for this theme, but I thought I'd warn you for future themes @beafialho 

Screenshot 2024-08-29 at 16 25 27

MaggieCabrera (approved):
This one is looking good and ready to go! The theme check plugin gives me this warning though:

WARNING: assets/images/image_3-scaled.jpg is 599.2 KB in size. Large file sizes have a negative impact on website performance and loading time. Compress images before using them.

I'm not sure if that will block it on the .org repo, @beafialho I'll leave that to you to check

Issue #7921, opened 2024-07-03
Stage: Remove sticky condition from home page query loops


richtabor:
If you don't have sticky posts, then the home page of your blog looks broken/empty. 

### Steps to replicate

Activate the theme without sticky posts. 

### Expected

See my blog's latest posts, regardless of sticky status or not. 

### Visual 

![CleanShot 2024-07-03 at 11 12 20](https://github.com/Automattic/themes/assets/1813435/83305e07-51d8-42fe-9502-d997420e0965)

mrfoxtalbot:
I raised the priority on this, it is likely to confuse users.

alaczek:
Stage has an updated version called OutNow, but it still has the original issue.

@iamarinoh could you fix OutNow, so that we can launch it, and delist Stage? I think in this case selecting the "Inherit query from template" makes the most sense.

henriqueiamarino:
I fixed it, @alaczek. The only reason I haven't selected **Inherit query from template** is that I want only one post displayed in each niche.

dsas:
Is it worth fixing Stage or shall we just close this issue? It's installed on ~4k sites, some of which will be unlaunched or never updated.

alaczek:
I was going to close this issue when we delist Stage.

lsl:
Why wait :smile:

PR #7956, opened 2024-07-17
OutNow: Remove Sticky Post settings

henriqueiamarino:
Fixing OutNow as required at #7921.
This will allow folks to delist Stage, an old version of this theme.


henriqueiamarino:
The detected changes on sites other than **OutNow** are errors already reported. Please do not consider them.

PR #7960, opened 2024-07-19
Cottage: add theme

beafialho:
**Description:**
Cottage is blogging theme with a cottagecore aesthetic.

_Issues caused by CBT described [here](https://github.com/WordPress/create-block-theme/issues/691) need to be addressed._


alaczek:
I fixed the issues manually.

jasmussen:
There are a couple of .DS_store files here as well. 

I haven't actually found out how to push to these playground branches, my git-fu is failing me, otherwise I'd help! Let me know if I can help in other ways.

jasmussen:
I wonder why there isn't a Playground link for this one.

As a quick gut-check, this appears to work well, so happy to ship this one.

PR #7969, opened 2024-07-24
Handy: add theme

henriqueiamarino:
Handy is a blog theme designed for individuals focusing on users who are into a side hustle on home improvement and handyperson services.

[Demo site](https://handy.mystagingwebsite.com/)

screenshot


alaczek:
The theme is looking good to me. One thing I would change is the blog on the homepage - I would align both posts and use a single query loop block. I know it adds a bit of visual interest with the way the posts are laid out, but I still think we should simplify the setup, especially since it suffers from the same query issues we discussed for OutNow.

henriqueiamarino:
@alaczek, I just updated the QL, as you mentioned. No strings attached to the previous design; we can make it more functional.

PR #7971, opened 2024-07-25
Goodskin: add theme

beafialho:
Adds **Goodskin** theme

Goodskin is a blogging theme designed for skincare blogs.

screenshot


alaczek:
I made some fixes, mostly to image and font paths.

I did notice the theme has footer.php pattern that isn't referenced anywhere.
I'm also wondering about the "Subscribe" section of the sidebar, since it's a purely static button. Should this be the Jetpack subscription block?

PR #7972, opened 2024-07-25
theme-utils: Update to use internal GitHub

ehg:
This changes all usage of `arc` to GitHub's `gh` utility.

I haven't tested this, and I'm not quite sure how to – but I'm pushing it as a starting point. Maybe someone could set up `gh` on their sandbox and give it a spin?

To set up `gh`:
 - Login to your sandbox
 - Run `gh auth login` and follow the instructions.

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -309,6 +309,20 @@ async function pushButtonDeploy() {
 		await updateLastDeployedHash();
 
 		let commitMessage = await buildGithubCommitMessageSince(hash);
+
+		// Make sure the themes/pub repo in sandbox is ready to create a PR to the A8C GitHub Host  
+		prompt = await inquirer.prompt([{
+			type: 'confirm',
+			message: 'Before you can create the GitHub PR, login in A8C GitHub Enterprise Server from the themes/pub repo in your sandbox with the command "gh auth login" and using your SSH key.\nAre you logged in?',
+			name: "continue",
+			default: false
+		}]);
+
+		if (!prompt.continue) {
+			console.log(`Aborted Automated Deploy Process at require to login in into A8C GitHub Enterprise Server in sandbox.`);
+			return;
+		}
+
 		let prUrl = await createGithubPR(commitMessage);
 		let prId = prUrl.split('pull/')[1];
 
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -290,7 +290,7 @@ async function pushButtonDeploy() {
 		if (thingsWentBump) {
 			prompt = await inquirer.prompt([{
 				type: 'confirm',
-				message: 'Are you ready to push this version bump change to the source repository (Github)?',
+				message: 'Are you ready to push this version bump change to the source repository (GitHub.com)?',
 				name: "continue",
 				default: false
 			}]);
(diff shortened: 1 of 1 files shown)


theme-utils.mjs
```diff
@@ -1014,66 +1014,64 @@ Subscribers:
...
+	console.log('creating GitHub Pull Request');
 
 	let result = await executeOnSandbox(`
 		cd ${sandboxPublicThemesFolder};
 		git branch -D deploy
 		git checkout -b deploy
 		git add --all
 		git commit -m "${commitMessage}"
-		arc diff --create --verbatim
+		gh pr create --fill
```
miksansegundo: The deploy process fails here with the message: `none of the git remotes configured for this repository point to a known GitHub host.`  when creating the PR.

At this point of the process, the commands are run in the sandbox in the path for the `themes/pub` repo `/home/wpcom/public_html/wp-content/themes/pub`

I tried adding the login command `gh auth login` before creating the PR and while the login is successful, the PR creation still fails with the same message.   

Screenshot 2567-07-26 at 19 17 13

I'm now trying to figure out from what means that error message means and what's required here. Any thoughts?

miksansegundo: The remotes look good

Screenshot 2567-07-26 at 19 22 47

miksansegundo: Reading this issue that seems related https://github.com/cli/cli/issues/7269

Noting that our `$GH_HOST` constant is empty and needs to contain the A8C Enterprise GitHub host.

The error comes from `gh browse`:
Screenshot 2567-07-26 at 19 28 10

[gh browse](https://cli.github.com/manual/gh_browse) needs the GitHub HOST url to open the PR. 

The script needs to log in into the A8C Enterprise GitHub host with a valid key to read. Your user must be added first to the repos in the Automattic organization visiting a MC page.

After logging in with SSH key, the PR command `gh pr create --fill` works:

Screenshot 2567-07-26 at 20 08 12

Screenshot 2567-07-26 at 20 11 37

miksansegundo: We can add instructions during the process saying "When prompted login in GitHub Enterprise Server typing the A8C GitHub host" 

Screenshot 2567-07-26 at 20 14 46

@ehg Maybe you know how to set the $GH_HOST with the A8C Enterprise GitHub host auto-magically in all Dotcom sandboxes.

miksansegundo: I tried adding a message with instruction but it does't work because during the script process there is no prompt to choose the Enterprise Server.

```
echo "When prompted login in 'GitHub Enterprise Server' typing the A8C GitHub host and using your SSH key."
gh auth login
gh pr create --fill
```
We could pass the host name to `gh auth login --hostname` but I believe that's not an option for safety reasons. See p1721998260238249-slack-C02ECE3ML

The only option I can imagine now is logging in manually before creating the PR. I have added a prompt to confirm that step. We can add more detailed instructions to the field guide page.

miksansegundo:
I have added changes to allow us testing the deploy process.

Steps for testing:

- checkout this branch in local
- run `npm run deploy`
 - you are supposed to deploy from trunk but that check is skipped 
- follow the prompts

I have only tested until the PR step and created a deploy PR during testing.

theme-utils.mjs
ehg: > Noting that our $GH_HOST constant is empty and needs to contain the A8C Enterprise GitHub host.

I don't think that helps much, with `gh auth login` it didn't default to the `GH_HOST` when logging in. Also `gh browse` works OK without `GH_HOST`.  I think maybe you could instruct people to read a set up page on the Fieldguide?

dsas: Resolving, because @miksansegundo has a prompt above that users should already have used `gh auth login`, that worked for me and seems like a good enough solution.

theme-utils.mjs
```diff
@@ -1014,66 +1029,64 @@ Subscribers:
...
+	console.log('Creating GitHub Pull Request');
 
 	let result = await executeOnSandbox(`
 		cd ${sandboxPublicThemesFolder};
 		git branch -D deploy
 		git checkout -b deploy
 		git add --all
 		git commit -m "${commitMessage}"
-		arc diff --create --verbatim
+		gh pr create --fill
```
dsas: Will need to add `--head deploy` here to tell it to push to the `deploy` branch, otherwise it will give an interactive prompt for whether to push to `Automattic/wpcom-themes` or whether to create a fork and push there.

This is where gh diverges from arc. arc doesn't care that there are multiple diffs all created from a `deploy` branch, whereas gh will error out if there are two PRs trying to merge `deploy` with `trunk`. I think this behaviour is fine though, it makes sense to stop and evaluate the existing open PR.

dsas (approved):
I think this is good to be merged. I tested until and including a `deploy pub outnow;exit;`

dsas:
Updated the existing theme deployment howto PNEWy-eqL-p2

PR #7984, opened 2024-07-31
Vows: Fix the footer

alaczek:
Make footer.html template part reference the footer.php pattern. 
Fix the footer.php pattern to remove static RSVP link.


henriqueiamarino (approved):
The CBT plugin should remove hard-coded links, but it has failed here. I am okay with the fixes, and I’ll check other themes after this issue.

henriqueiamarino:
@alaczek, the changes on the theme look good to me. But the preview link is not working. Do you know why?

PR #7985, opened 2024-07-31
Blissed: Fix the footer

henriqueiamarino:
Fix the footer.php pattern to remove the static RSVP link following #7984.


jasmussen (approved):
Looks right to me!

PR #7928, opened 2024-07-04
Adonay: add theme

beafialho:
Adonay is crafted for single page websites that want to leave a stunning and memorable first impression. It also provides post and page templates for those looking to customize and broaden their website's functionality. Adonay comes with 3 distinctive style variations and 12 vibrant color options.

screenshot

PR #7987, opened 2024-07-31
Theme previews: use external action

vcanales:

This PR changes the current theme preview workflow to use an external function that executes the same process, but has two advantages:

1. It runs using `pull_request_target`, meaning that PRs created from forks will get previews too.
2. The action can be shared with other repos with a similar structure to get Playground previews as well, such as the [WordPress Community Themes](https://github.com/WordPress/community-themes) repo.

We could also consider moving the repo, which is currently under my account, to the Automattic org.

#### Related issue(s):

Related to the discussion [here](https://github.com/Automattic/themes/pull/7928#issuecomment-2260474085) cc. @adamziel @mikachan

### DIFF
--- a/.github/workflows/preview-theme.yml
+++ b/.github/workflows/preview-theme.yml
@@ -7,105 +7,17 @@ permissions:
   pull-requests: write
 
 jobs:
-  check-for-changes-to-themes:
+  preview-theme-changes:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


mikachan (approved):
This is fab, thanks so much @vcanales! I think we should bring this in to test it out and allow previews to be created from forks.

Issue #7970, opened 2024-07-25
Multiple themes: Remove copyright symbol


alaczek:
A few of our themes have a copyright and date inserted in the footer, as outline here: pekYwv-4ok-p2
This issue is to consolidate this task in one place.

The affected themes: https://github.com/search?q=repo%3AAutomattic%2Fthemes+%C2%A9+language%3APHP&type=code&l=PHP

- [x] Vitrum
- [x] Fontaine
- [x] Pendant
- [x] Bitacora
- [x] Jaida
- [x] Upsidedown
- [x] Course

dsas:
Opened https://github.com/Automattic/themes/pull/7981 to deal with Course and Upside Down - they're the simplest cases as the copyright statement there didn't make any sense.

I'm not 100% clear on what we want to do with the others. Just remove the year (Fontaine, Pendant, Jaida, Bitacora), or remove the copyright para altogether (+vitrum)?

**If** removing the copyright para completely, then I think we'll want "Designed with WordPress" aligned differently, I've made suggestions below but maybe a design eye would have a better suggestion.




Theme | Design | Suggestion
-------|------|-----
Vitrum | <img width="1461" alt="Screenshot 2024-07-30 at 18 30 22" src="https://github.com/user-attachments/assets/548586b6-1cb0-4c9e-a1f5-1d160b185819"> | Move "Designed with" in place of copyright
Fontaine | <img width="731" alt="Screenshot 2024-07-30 at 18 33 13" src="https://github.com/user-attachments/assets/525ef6f3-1848-44d0-b50b-d4fbb04a86f3"> | Just remove copyright, leave everything else where they are.
Pendant | <img width="731" alt="Screenshot 2024-07-30 at 18 56 32" src="https://github.com/user-attachments/assets/6c999acf-2dcc-41ea-be94-dd2a3d4882d3"> | Just remove copyright, leave "Powered by" where it is
Bitacora | <img width="731" alt="Screenshot 2024-07-30 at 18 57 10" src="https://github.com/user-attachments/assets/4ca84b3a-bc07-4f80-8e04-416930459ac8"> | Centre align "Designed with"
Jaida | <img width="1461" alt="Screenshot 2024-07-30 at 18 31 34" src="https://github.com/user-attachments/assets/4ce19faf-7ffe-4892-861a-a28e317ca43e"> | Maybe center align "Designed with..." in the same row as the site title and social icons

alaczek:
[...]

PR #7988, opened 2024-07-31
Remove copyright statement

dsas:
Remove copyright statement from the following themes:
 * Vitrum
 * Fontaine (x2)
 * Pendant
 * Bitacora
 * Jaida

The standard for a while has been to use "Designed with WordPress" (or something similar) in the footer, rather than a copyright statement. Additionally the copyright line is often incorrect - where it hardcodes the year, this is obviously not accurate and where it credits the name of the blog, or states "All Rights Reserved", this may not be accurate either. We should leave it to users to set their own copyright statements appropriately.

Theme | Before | After
-------|------|-----
Vitrum | Screenshot 2024-07-30 at 18 30 22 | image
Fontaine | Screenshot 2024-07-30 at 18 33 13 | image
Fontaine (dark) | image | image
Pendant | Screenshot 2024-07-30 at 18 56 32 | image
Bitacora | Screenshot 2024-07-30 at 18 57 10 | image
Jaida | Screenshot 2024-07-30 at 18 31 34 | image

#### Related issue(s):
#7970

### DIFF
--- a/bitacora/patterns/footer.php
+++ b/bitacora/patterns/footer.php
@@ -61,22 +61,8 @@
 	</div>
 	<!-- /wp:columns -->
 
-	<!-- wp:group {"align":"wide","layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
+	<!-- wp:group {"align":"wide","layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"center"}} -->
 	<div class="wp-block-group alignwide">
-		<!-- wp:group {"style":{"spacing":{"blockGap":"4px"}},"layout":{"type":"flex","flexWrap":"nowrap"}} -->
-		<div class="wp-block-group">
-			<!-- wp:paragraph {"fontSize":"small"} -->
-			<p class="has-small-font-size"><?php _e( '©', 'bitacora' ); ?></p>
-			<!-- /wp:paragraph -->
-
-			<!-- wp:site-title {"style":{"typography":{"letterSpacing":"0px","textTransform":"none"}},"fontSize":"small"} /-->
-
-			<!-- wp:paragraph {"fontSize":"small"} -->
-			<p class="has-small-font-size"><?php _e( '2022, All Rights Reserved.', 'bitacora' ); ?></p>
-			<!-- /wp:paragraph -->
-		</div>
-		<!-- /wp:group -->
-
 		<!-- wp:paragraph {"align":"left","className":"has-small-font-size","fontSize":"small"} -->
 		<p class="has-text-align-left has-small-font-size">
 			<?php
--- a/fontaine/patterns/footer-dark.php
+++ b/fontaine/patterns/footer-dark.php
@@ -52,11 +52,6 @@ declare( strict_types = 1 );
 	<div class="wp-block-group">
 		<!-- wp:group {"style":{"spacing":{"blockGap":"var:preset|spacing|30"}},"layout":{"type":"flex","flexWrap":"nowrap"}} -->
 		<div class="wp-block-group">
-			<!-- wp:paragraph -->
-			<p><?php echo esc_html__( '© 2023', 'fontaine' ); ?></p>
-			<!-- /wp:paragraph -->
-
-			<!-- wp:site-title {"level":0} /-->
 		</div>
 		<!-- /wp:group -->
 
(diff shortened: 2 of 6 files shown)


fontaine/patterns/footer-dark.php
```diff
@@ -52,11 +52,6 @@
 	<div class="wp-block-group">
```
mmtr, line 52: This block group seems unnecessary.

With block group:
Screenshot 2024-08-01 at 11 41 45



Without block group (paragraph only):
Screenshot 2024-08-01 at 11 42 54

fontaine/patterns/footer.php
```diff
@@ -52,11 +52,6 @@
 	<div class="wp-block-group">
```
mmtr, line 52: Same as above

bitacora/patterns/footer.php
```diff
@@ -61,22 +61,8 @@
 	</div>
 	<!-- /wp:columns -->
 
-	<!-- wp:group {"align":"wide","layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
+	<!-- wp:group {"align":"wide","layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"center"}} -->
```
mmtr, line 64: I think we can get rid of this group and just display a paragraph which is aligned to the center

With block group:
Screenshot 2024-08-01 at 11 46 21

Without block group (paragraph only aligned to the center):
Screenshot 2024-08-01 at 11 46 49

jaida/patterns/footer.php
```diff
@@ -34,7 +34,7 @@
 
 <!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"right":"0","left":"0","top":"var:preset|spacing|80","bottom":"var:preset|spacing|70"}}},"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
 <div class="wp-block-group alignwide" style="padding-top:var(--wp--preset--spacing--80);padding-right:0;padding-bottom:var(--wp--preset--spacing--70);padding-left:0"><!-- wp:paragraph {"style":{"typography":{"fontSize":"0.8rem"}}} -->
```
mmtr, line 36: If we move the padding to the inner paragraph that contains the credit, I think we can simplify this.

With block group:
Screenshot 2024-08-01 at 11 49 56

Without block group (paragraph only with extra padding):


Screenshot 2024-08-01 at 11 52 37

pendant/patterns/footer.php
```diff
@@ -10,7 +10,7 @@
 <!-- wp:group {"style":{"spacing":{"padding":{"top":"40px","bottom":"40px"}}},"layout":{"inherit":true}} -->
```
mmtr, line 10: Another unnecessary block group here.

With block group:

Screenshot 2024-08-01 at 11 54 19

Without block group (paragraph only with extra padding):

Screenshot 2024-08-01 at 11 55 59

vitrum/patterns/footer.php
```diff
@@ -9,15 +9,7 @@
 ?>
 <!-- wp:group {"style":{"border":{"top":{"width":"1px"},"right":[],"bottom":[],"left":[]}},"layout":{"type":"constrained"}} -->
```
mmtr, line 10: This one looks harder to remove because of the constrained layout

mmtr (commented):
This tests well for me. But I think we can remove some unnecessary group blocks now. Let me know what do you think.

pendant/patterns/footer.php
dsas: I did remove this block group originally, but decided that keeping the group and the empty paragraph makes it easier for people to customise the footer if they did want to revert back to something like how it was before. They just need to click where it says "Type / to choose block" and begin typing

dsas: (same goes for many of the others)

mmtr: Fair enough 👍

mmtr (approved):

mmtr:
Just realized that, for Vitrum, the credit was on the right, and it's now on the left. Should we keep it on the right as we did for the other themes?

Before | After
------|-----
Screenshot 2024-07-30 at 18 30 22 | image

dsas:
> Just realized that, for Vitrum, the credit was on the right, and it's now on the left. Should we keep it on the right as we did for the other themes?

@mmtr, that would've been more consistent 🤔. I changed this deliberately though, it (subjectively) looked better on the left to me and [Henrique confirmed that's fine](https://github.com/Automattic/themes/issues/7970#issuecomment-2260075050).

PR #8007, opened 2024-08-08
Olympique: add theme

henriqueiamarino:
A theme to celebrate the history of the international games and the milestones that have shaped their legacy. 

[Demo site](https://olympique.mystagingwebsite.com/)

screenshot


jasmussen:
@mikachan @adamziel Curious why playground links stopped appearing for new themes. They were really useful!

For now I'll review this manually.

jasmussen:

I love this one. It's so lovely.

2024-08-13 11 57 17 local-wordpress local f9c5cc6841f7

The stripe up here is intentionally taller, that works better with the world map logo:

Screenshot 2024-08-13 at 11 57 26

But looking at the demo site, I wonder if the navigation on the right should be top aligned?

image

I like the style variations of this one a lot. Even if it doesn't go wild with typography: there's something so very olympic about the base helvetica-esque font, that it seems _right_ for this one to not go wild there.

I like these two:

Screenshot 2024-08-13 at 11 58 36

Screenshot 2024-08-13 at 11 57 40

The latter, light gray version, thanks for including a monochrome one. The contrast is a bit low though, I wonder if the text color could be a tad darker and still feel "light"?

Templates overall look good:

Screenshot 2024-08-13 at 11 58 55

Mobile looks fantastic. 

Except for the navigation overlay, it's lacking padding:

Screenshot 2024-08-13 at 12 02 52

I believe there was a trick to adjust that padding?

On the demo site, and you don't need to fix this because it's not shipping with the theme, but I notice this image is low resolution/blurry:

Screenshot 2024-08-13 at 12 04 33

The list style here, I wonder if the left padding of the list block should be adjusted to push the bullets inwards? Or is it intentionally offset to the left?

Screenshot 2024-08-13 at 12 04 27

For this quote in the About page, there's a curious little .. that I don't know where is coming from:

Screenshot 2024-08-13 at 12 04 49

Probably also in the demo site category, but good to look at.

**Summary**

This is a really nice one. Beautiful, clean, olympic. It's virtually ready. There were a couple of small tweaks worth evaluating, a consideration for contrast on one of the style variations, but otherwise, this is ready to go! Nice one.

mikachan:
> Curious why playground links stopped appearing for new themes. They were really useful!

We've just merged a PR that will hopefully fix them 🤞 https://github.com/Automattic/themes/pull/8011

henriqueiamarino:
@jasmussen, thanks again. All your feedback was considered on this new commit. The only exception is the quote block that apparently carries a `blockquote-class` that I couldn't find and remove.

I also decided to use photos on the demo site instead of pictograms. This way, the image duotones are more apparent and appealing.

Please review it again asap. I am planning to have it submitted today.

jasmussen:
Theme looks good. 

The only piece is the inner padding of the navigation menu. Based on the code, it looks like we can't easily fix that without applying a root padding, which is not fitting for this theme.

So one thing we can do is just _remove_ the navigation at the bottom (it's fine, there's a menu at the top after all). OR, default it to not be a burger menu.

This will, longer term stop being a problem as it should become possible to style the overlay separately.

If you can either remove that navigation, or just default it to not be the hamburger, you can merge and ship this theme. Nice work!

henriqueiamarino:
Ok, I'll remove that hamburger one, @jasmussen.

In addition to what you said about Typography, I'll try to find options similar to Helvetica until the end of this week. This way, we'll have different (but coherent) options in our Style variations. Once we launch this, it's much easier to update.

jasmussen:
The font is fine as is! Doesn't have to be helvetica. In fact, the overused grotesk feels very olympic. No change needed from my end, pure appreciation.

henriqueiamarino:
@jasmussen theme and demo site updated.

jasmussen:
Nice! Ship it!

adamziel:
> Curious why playground links stopped appearing for new themes. They were really useful!

The integration was set up by @vcanales so let me CC him.

vcanales:
> > Curious why playground links stopped appearing for new themes. They were really useful!
> 
> The integration was set up by @vcanales so let me CC him.

[It is working on new themes](https://github.com/Automattic/themes/pull/8154#issuecomment-2350946693). I'm not sure why it didn't run on this one, but there were some changes to the workflow between when this PR was created and now, so it might have gotten caught in the middle of the change. 

edit: I think the report was made previous to the fixes I mentioned earlier, so I would say that it's all good now :) the action [did run a bit after the comment was made.](https://github.com/Automattic/themes/pull/8007#issuecomment-2286289071)

PR #8009, opened 2024-08-13
Blissed: Separate strings from HTML.

alaczek:
Separate strings from HTML in header and footer patterns to fix this issue:

image


alaczek:
@henriqueiamarino I had to make some more changes to the footer, as it was showing HTML (see screenshot)

Overall though, these static links are not a great customer experience, since they require manual linking. What do you think about switching them over to a nav menu (styled the same way) instead? This would make it much easier, as new pages would automatically show up there, already linked.

PR #8008, opened 2024-08-12
Musee: add theme

henriqueiamarino:
This is another renamed version of the [Hall](https://github.com/Automattic/themes/pull/7607) theme,  formerly named Expo (https://github.com/Automattic/themes/pull/7976), and Atrium (https://github.com/Automattic/themes/pull/7883). Musee is a blog theme that presents posts in different Query Loops using tags—the perfect pick for museums or galleries.

[Demo site](https://musee.mystagingwebsite.com/)

screenshot

PR #7976, opened 2024-07-29
Expo: add theme

henriqueiamarino:
This is the renamed version of Atrium (#7883), an update of the Hall theme.
Expo is a blog theme that presents posts in different Query Loops using tags—the perfect pick for museums or galleries.

[Demo site](https://expo.mystagingwebsite.com/)

screenshot

PR #8012, opened 2024-08-13
Musee: Style variations updated

henriqueiamarino:
This commit complements the merged theme #8008.
I am not sure why the style variations and assets were not present in the previous version but this is now fixed. Just as as it was on the former version of this theme, Expo #7976.


henriqueiamarino:
As this was reviewed and approved yesterday, it's okay to be merged.
I just checked the style variations using the preview link, and they are all there now, as expected.

PR #8015, opened 2024-08-14
Hola: Round of fixes

alaczek:
Round of fixes to the templates and patterns. cc @henriqueiamarino


henriqueiamarino:
Thanks for fixing it, @alaczek. I've checked it in the preview, and it looks fine.

PR #8019, opened 2024-08-14
Tweak rootHasVersionBump logic

mikachan:
Tweaks the version bump logic so the root project version is only bumped if there are changes.

Hopefully prevents unnecessary version bumps from happening, like this: https://github.com/Automattic/themes/pull/8018

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -674,7 +674,11 @@ async function versionBumpThemes() {
 	//version bump the root project if there were changes to any of the themes
 	const rootHasVersionBump = await checkProjectForVersionBump(hash);
 
-	if (versionBumpCount > 0 && !rootHasVersionBump) {
+	if (!rootHasVersionBump) {
+		console.log('No changes detected; version bump not required');
+	}
+
+	if (versionBumpCount > 0 && rootHasVersionBump) {
 		await executeCommand(`npm version patch --no-git-tag-version && git add package.json package-lock.json`);
 		changesWereMade = true;
 	}


vcanales (commented):
Thanks! LGTM

theme-utils.mjs
```diff
@@ -674,7 +674,11 @@ async function versionBumpThemes() {
 	//version bump the root project if there were changes to any of the themes
 	const rootHasVersionBump = await checkProjectForVersionBump(hash);
 
-	if (versionBumpCount > 0 && !rootHasVersionBump) {
+	if (!rootHasVersionBump) {
+		console.log('No changes detected; version bump not required');
+	}
+
+	if (versionBumpCount > 0 && rootHasVersionBump) {
```
miksansegundo, line 681: I have had to update this logic because otherwise the version wasn't bumped when theme changes are made. See https://github.com/Automattic/themes/commit/2ba4dd2aa9e29d6ad566564befb974f18479c674#diff-877b4284538ebb4cf1411cf4487c456bda52082b9e72e4a0b6770d5a80d7d5a9R677

`versionBumpCount` is a count for the theme bumps done for changed themes.

`rootHasVersionBump` is a boolean that tell us if the root package version has changed from the previews release tag.

Therefore, we can skip the root package bump only when `versionBumpCount === 0` and we must bump it when it's not done yet `rootHasVersionBump === false`.

cc: @dsas

mikachan: Thanks @miksansegundo 👍

PR #8020, opened 2024-08-14
Fix dotorg deploy script

### LINKED ISSUE #7936: [Bug]: The deploy-dotorg GH action fails if there is an error with a single theme
### Quick summary

The deploy-dotorg GH action seems to be failing here: https://github.com/Automattic/themes/actions/runs/9837238698/job/27154613541.

It fails with this error: `sed: can't read s/, auto-loading-homepage//g: No such file or directory`

We could update the script so that it skips over themes that cause an error, so that the script can continue rather than exiting out.

### Steps to reproduce

Run the deploy-dotorg GH action.

I would expect the script to continue to run over all possible themes rather than exiting on the first error.

### What actually happened

The script exited and therefore stopped on the first error.

### Browser

_No response_

### Context

_No response_

_No response_

### Other notes

_No response_

### Reproducibility

Consistent

### Severity

None

### Available workarounds?

None

### Workaround details

_No response_

mikachan:
This adds a check to see if a theme exists in the wordpress.org directory before attempting to deploy any updates. It also removes any slashes from the theme slug and checks if the version that is being added is actually the latest version. This should fix the failures with this GH action. Here's an example run: https://github.com/Automattic/themes/actions/runs/10391716725/job/28775266698

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/7936


vcanales:
I looked into this lightly as well. I'm not sure if it's part of the cause for errors, but the script is not removing some trailing slashes. You could cherry-pick [this commit](https://github.com/Automattic/themes/commit/11b87c993ac1f33cd5e27578d009c9b08296659c) if you'd like to remove slashes.

Example: https://github.com/Automattic/themes/actions/runs/10385090310/job/28753468617#step:3:9

You can see that the theme is called `expo/`, which is wrong. I saw it with some other themes as well, down the commit history.

mikachan:
I don't think that's part of the errors, but I'll cherry-pick that anyway as it sounds like that needs fixing too. Thank you!

Edit: 

> the theme is called expo/

Maybe it is part of the errors!

deploy-dotorg.sh
```diff
@@ -36,6 +36,9 @@ rm -rf ./deploy
 # Do things for all of the themes
 for THEME_SLUG in */ ; do
 
+	# Remove slashes from the theme slug
+	THEME_SLUG=${THEME_SLUG%/}
```
vcanales: Going to one-up my own suggestion, and suggest this other snippet, which removes _all_ slashes

```suggestion
	THEME_SLUG=$(echo $THEME_SLUG | tr -d '/')
```

vcanales (approved):
Triggered another run with the trailing slash removal, and it does well: https://github.com/Automattic/themes/actions/runs/10391823637

LGTM

deploy-dotorg.sh
mikachan: Nice! Is `$slug` a new variable? Or should this be `THEME_SLUG=$(echo $THEME_SLUG | tr -d '/')`?

mikachan: Looks like you updated the suggestion, I'll test it out, thanks!

Issue #7910, opened 2024-07-02
[Luminance] - Featured images are misaligned. 


Greatdane:
### Quick summary

Check the demo site; https://luminancedemo.wordpress.com/

The featured images within the query loop are misaligned. Just reported by a user and can reproduce. 

![CleanShot 2024-07-02 at 13 27 38@2x](https://github.com/Automattic/themes/assets/5075840/197b8c91-3c35-4634-8581-777af379b6be)

### Steps to reproduce

Activate Luminance and check the homepage template in the site editor. 

### What you expected to happen

Featured image should be aligned. 

### What actually happened

Featured image is not aligned. 

### Browser

Google Chrome/Chromium, Mozilla Firefox, Microsoft Edge, Apple Safari

### Context

Customer Report.

### Platform (Simple, Atomic, or both?)

Simple

### Reproducibility

Consistent

### Severity

All

### Available workarounds?

Yes, difficult to implement

### Workaround details

Likely CSS. Not Available to Starter plans.

Greatdane:
8404697-zd-a8c

marutim:
Another report here: 8404682-zen

jartes:
📌 REPRODUCTION RESULTS
- Tested on Simple – Replicated
- Tested on Atomic – Could Not Replicate

📌 FINDINGS
On atomic sites, this is not happening, as they have the following CSS to fix that: 

```
.wp-block-post-featured-image.wp-block-post-featured-image img {
    vertical-align: top;
width: auto;
}
```

📌 ACTIONS
- Triaged
- Adding normal priority

lizswafford:
8428324-zd-a8c

supernovia:
Another report here. 8460124-zd-a8c

This is an atomic site. We'll add the CSS but noting it does occur there by default.

dsas:
[Luminance sets `core/post-featured-image` to `vertical-align:top`](https://github.com/Automattic/themes/blob/86579e7425456c170553814456273f305c987416/luminance/theme.json#L510) which would correct the display but this is being overridden to `vertical-align:bottom` by the [more specific CSS in post-features-image/style.css](https://github.com/WordPress/WordPress/blob/master/wp-includes/blocks/post-featured-image/style.css) since [WP 6.6 downgraded CSS specificity](https://make.wordpress.org/core/2024/06/21/wordpress-6-6-css-specificity/).

dsas:
[...]

PR #8023, opened 2024-08-19
Luminance: Fix display of featured images

dsas:
The featured images in Luminance were completely misaligned due to changes in how Gutenberg handles specificity of style rules. Rather than the custom block-specific CSS for the `core/featured-image` block taking priority, the default block library styles do.

The problem has been reported upstream but in the meantime, using the theme-wide custom CSS fixes the problem.

Before | After
-------|------
image | image

(ignore the devtools in the screenshots)

#### Related issue(s):
 * https://github.com/Automattic/themes/issues/7910 
 * https://github.com/WordPress/gutenberg/issues/64453


alaczek:
This does work for me! Closing as fixed.

PR #8028, opened 2024-08-22
Luxus: Add theme

alaczek:
More context & announcement post: pNEWy-i79-p2


luxus/style.css
```diff
@@ -0,0 +1,197 @@
...
+}
+
+.contact-form__input-error {
+	font-size: var(--wp--preset--font-size--small);
+}
+
+/*
+ * Decorative border for headings with .wave-underline class.
+ */
+:is(h1, h2, h3, h4, h5, h6).wave-underline::after {
```
dsas: This is causing a jarring hover bug inside the editor where the wavy underline stops underlining


https://github.com/user-attachments/assets/4b87ffb6-c3f5-4d3b-b6b6-42fb825f6258



The editor styles for `.block-editor-block-list__layout` is overriding nearly all of these rules. Maybe the specificity needs increasing here?

luxus/style.css
```diff
@@ -0,0 +1,197 @@
...
+	mask: url(assets/images/wave.svg) no-repeat center / cover;
+	height: 7px;
+	content: " ";
+	width: 200px;
+	background: var(--wp--preset--color--theme-4);
+	display: block;
+	mask-position: center;
+	position: relative;
+	bottom: -9px;
+	left: calc( 100% - ( 50% + 100px ) );
```
dsas: Not sure that this is important 🤔  but this won't work with left or right aligned headings. The text will be left/right aligned and the wave will be centered.

Maybe this rule should include something like `:is(h1, h2, h3, h4, h5, h6).has-text-align-center.wave-underline::after`

luxus/theme.json
```diff
@@ -0,0 +1,659 @@
...
+				},
+				{
+					"fontFace": [
+						{
+							"fontFamily": "Hanken Grotesk",
+							"fontStyle": "normal",
+							"fontWeight": "100",
+							"src": [
+								"file:./assets/fonts/HK-grotesk/hanken-grotesk_normal_100.ttf"
+							]
```
madhusudhand: instead of including all individual font files, can we try using [variable](https://github.com/marcologous/hanken-grotesk/tree/master/fonts/variable) font instead?

`"fontWeight": "100 900",`

luxus/style.css
```diff
@@ -0,0 +1,197 @@
+/*
+Theme Name: Luxus
+Theme URI: https://wordpress.com/theme/luxus
+Author: Automattic
+Author URI: https://automattic.com/
+Description: A WordPress theme for nail and beauty salons.
+Requires at least: 6.0
```
madhusudhand: theme is using `fluid` typography (theme.json) which was introduced in [6.1](https://wordpress.org/news/2022/11/misha/).
```suggestion
Requires at least: 6.1
```

luxus/theme.json
alaczek: Yep, that's a good idea - I'll do that!

luxus/style.css
alaczek: This is a bit dodgy, isn't it? I'll try to re-do it by creating a custom separator with the wave, should be easier for customers to grasp.

alaczek: I created a custom separator instead and got rid of this dodgy CSS solution :)

luxus/style.css
alaczek: I removed this entirely.

PR #8031, opened 2024-08-22
Olympique: fixes following Dotorg review

henriqueiamarino:
This update fixes the following issues:
- I found link that in the single post is not underlined, please fix this in the update.


jasmussen:
I've checked in on whether those `<?php declare( strict_types = 1 ); ?>` that are being removed in this PR are okay, and intended. But best guess, yes, those should not be there. Changes look good otherwise 👍 👍

PR #8053, opened 2024-09-02
Escape patterns in directory

vcanales:
A lot of changes from the code formatted; actual changes for review [here.](https://github.com/Automattic/themes/commit/5e7ce6a2f727c63090fefcf2ef123a16f9df193f)

This enables using our existing script for applying escaping functions on patterns from the cli by passing in a directory:

`npm run patterns:escape -- <theme_name>`

A known issue is that the function applies escaping to tags already wrapped in an escaping function—this is not a trivial issue to fix, and I suggest solving it in a separate effort.

### DIFF
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -1464,17 +1464,18 @@ async function escapePatterns( themeSlug ) {
 		} );
 	} else {
 		// If no theme slug is provided, detect changed files via Git
-		patternFiles = await glob( '**/*.php', {
-			ignore: [ 'node_modules/**', 'vendor/**' ], // Exclude node_modules and vendor directories
-		} );
+		const stagedFiles = await executeCommand(
+			'git diff --name-only --cached'
+		);
+		const unstagedFiles = await executeCommand( 'git diff --name-only' );
 
-		// Further filter the files to only those that have changed
-		patternFiles = patternFiles.filter( ( file ) => {
-			const result = execSync( `git diff --name-only HEAD ${ file }` )
-				.toString()
-				.trim();
-			return result !== '';
-		} );
+		// Combine staged and unstaged files, remove duplicates, and filter for pattern files
+		patternFiles = [
+			...new Set( [
+				...stagedFiles.split( '\n' ),
+				...unstagedFiles.split( '\n' ),
+			] ),
+		].filter( ( file ) => file.match( /.*\/patterns\/.*.php/g ) );
 	}
 
 	// Arrange patterns by theme
--- a/theme-utils.mjs
+++ b/theme-utils.mjs
@@ -1454,16 +1454,16 @@ export async function executeCommand( command, logResponse ) {
 	} );
 }
 
-async function escapePatterns( directory ) {
+async function escapePatterns( themeSlug ) {
 	let patternFiles;
 
-	if ( directory ) {
-		// If a directory is provided, use fast-glob to find all PHP files in the specified directory
-		patternFiles = await glob( `${ directory }/**/*.php`, {
+	if ( themeSlug ) {
+		// If a theme slug is provided, use fast-glob to find all PHP files in the specified theme's directory
+		patternFiles = await glob( `${ themeSlug }/**/*.php`, {
 			ignore: [ 'node_modules/**', 'vendor/**' ], // Exclude node_modules and vendor directories
 		} );
 	} else {
-		// If no directory is provided, detect changed files via Git
+		// If no theme slug is provided, detect changed files via Git
 		patternFiles = await glob( '**/*.php', {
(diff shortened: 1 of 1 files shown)


theme-utils.mjs
```diff
@@ -1180,228 +1367,265 @@ Host wpcom-sandbox
...
-
-	// arrange patterns by theme
-	const themePatterns = patterns.reduce((acc, file) => {
-		const themeSlug = file.split('/').shift();
+async function escapePatterns( directory ) {
+	let patternFiles;
+
+	if ( directory ) {
+		// If a directory is provided, use fast-glob to find all PHP files in the specified directory
+		patternFiles = await glob( `${ directory }/**/*.php`, {
```
jffng: If we wanted to keep it `npm run patterns:escape -- [argument]` I think it would make more sense to accept the theme folder name / slug as the argument rather than a specific directory, and glob for the `{directory}/patterns/*.php`. The other theme-utils scripts accept the theme slug as the argument also.

jffng (commented):
When I tried running this, I ran `npm run patterns:escape -- adonay` and it escaped the `functions.php` file which wasn't my intention. 

Should we make the argument more specifically named, since it is expecting a directory and not the theme folder name? 

For example `npm run patterns:escape --directory=adonay/patterns` or `npm run patterns:escape --theme=adonay`. Or `npm run patterns:escape -- [themeslug]`.

theme-utils.mjs
vcanales: >  I think it would make more sense to accept the theme folder name / slug as the argument rather than a specific directory

That was the intent, but I see that the parameter naming was confusing; I've updated it.

vcanales:
> When I tried running this, I ran `npm run patterns:escape -- adonay` and it escaped the `functions.php` file which wasn't my intention.

I've modified it so that we only look within `patterns/` for the given theme https://github.com/Automattic/themes/pull/8053/commits/629eaa31cdd2ccdab1281875d834e76513f45fa6. I tried with Adonay again, and the output seems to be what we expect this time: only patterns being escaped.

```sh
╔══════════════════════════════════════════╗
║              THEME: adonay               ║
║                                          ║
║  Following patterns may get updated with ║
║    escaped strings and/or image paths    ║
╟──────────────────────────────────────────╢
║ adonay/patterns/404.php                  ║
╟──────────────────────────────────────────╢
║ adonay/patterns/comments.php             ║
╟──────────────────────────────────────────╢
║ adonay/patterns/footer.php               ║
╟──────────────────────────────────────────╢
║ adonay/patterns/hidden-no-results-conten ║
║ t.php                                    ║
╟──────────────────────────────────────────╢
║ adonay/patterns/home.php                 ║
╟──────────────────────────────────────────╢
║ adonay/patterns/homepage-with-a-small-ro ║
║ unded-image.php                          ║
╟──────────────────────────────────────────╢
║ adonay/patterns/homepage-with-full-heigh ║
║ t-image.php                              ║
╚══════════════════════════════════════════╝

**Comment by jffng** [comment]
Nice, thanks! 

Did you try testing the normal `npm run patterns:escape` script after modifying a pattern file? It still works, but is running very slow for me on this PR. Any idea why that might be happening?

**Comment by vcanales** [comment]
> Did you try testing the normal `npm run patterns:escape` script after modifying a pattern file? It still works, but is running very slow for me on this PR. Any idea why that might be happening?

Thanks for checking that scenario; the way in which I was searching and running git was really inefficient. In 60811f5a887ed8662e6b63b1c2b7038f74cd3237, I've reverted it back to something closer to what the original script did when no theme slug is provided.

**Review by jffng — APPROVED** [review]

File: theme-utils.mjs [hunk 2]
```diff
@@ -1534,12 +1759,16 @@ async function validateThemes( themes, { format, color, tableWidth } ) {
...
+		const themeRequires = getThemeMetadata(
+			styleCss,
+			'Requires at least',
+			true
+		);
 		const wpVersion = themeRequires
 			? `${ themeRequires }.0`.split( '.', 2 ).join( '.' )
 			: undefined;
-		const hasThemeJsonSupport = wpVersion && semver.valid( `${ wpVersion }.0` ) && semver.gte( `${ wpVersion }.0`, '5.9.0' )
-		const hasThemeJson = fs.existsSync( themeJsonPath );
```
MaggieCabrera, line 1542: I think this has broken the theme validation

MaggieCabrera: See this PR https://github.com/Automattic/themes/pull/8116

jffng: Thanks for catching @MaggieCabrera — it seems there were some changes we thought were formatting, but broke some of the theme validation functions. Fixing in #8174.

PR #8062, opened 2024-09-04
Blissed: Fix page jumps

alaczek:
Fixes page jumps in Blissed theme by adding the URL before the anchor link


miksansegundo (approved):
LGTM 👍

PR #8039, opened 2024-08-27
Add automated JSON schema validation

ajlende:
## Changes proposed in this Pull Request:

Adds an npm script and GitHub action for validating JSON for themes.

1. Gets the schema from the "Requires at least" minimum WP version.
2. Warns when the `$schema` version doesn't match the minimum WP version.
3. Validates the theme.json and style variations against the minimum WP version schema.

### Validation script

I may have had a bit too much fun with building the tables and adding colors, so the code exploded a bit. 😅

```sh-session
$ node theme-utils.mjs help validate-theme                    

validate-theme [--format=FORMAT] [--color=WHEN] [--table-width=COLUMNS] <array of theme slugs>

Validates a theme against the WordPress theme requirements.

--format=FORMAT

        Output format. Possible values: *table*, json, dir.

--color=WHEN

        Colorize the output for table or dir formats. The automatic mode only enables colors if 
        an interactive terminal is detected. Possible values: *auto*, always, never.

--table-width=COLUMNS

        Explicitly set the width of the table format instead of determining it automatically.
        Will default to 120 if omitted and width cannot be determined automatically.
```

But I think the new result is _really_ nice. ✨ 

image

And it works with pagers.

```sh-session
$ # It can take a while to process longer lists of themes
$ node theme-utils.mjs validate-theme --color=always --table-width=$(( $(tput cols) )) atheme,adventurer,grammer,skatepark | less -R
```

image

Or without if you only have a theme or two to validate.

```sh-session
$ node theme-utils.mjs validate-theme grammer                                                                  

Progress: [ '■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■', 100 ] 1/1

╔═════════════════════════════════════════════════════════════╤══════════════════════════╗
║ ERROR                                                       │ Message : missing 'Requi ║
║ Theme : Grammer                                             │ res at least' header met ║
║ File  : style.css                                           │ adata                    ║
╟─────────────────────────────────────────────────────────────┼──────────────────────────╢
║ WARNING                                                     │ Actual   : undefined     ║
║ Theme : Grammer                                             │ Expected : 5.9 or greate ║
║ File  : style.css                                           │ r                        ║
║                                                             │ Message  : the 'Requires ║
║                                                             │  at least' version does  ║
║                                                             │ not support theme.json   ║
╟─────────────────────────────────────────────────────────────┼──────────────────────────╢
║ WARNING                                                     │ Instance path : /setting ║
[...]

PR #8065, opened 2024-09-04
fix a minor issue with theme.json validation error reporting

madhusudhand:

It doesn't exit with code `1` even though there are errors in the output.
It fixes it by checking problems with `error` type

#### Related issue(s):

#8039


mikachan (approved):
Code changes look good to me 👍

PR #8069, opened 2024-09-04
Remove themes launched with different names

dsas:

 * Atrium, Expo and Hall were old names for Musee
 * Trailblazer is now Maverick

 None of the themes have been launched on dotorg or dotcom under these
 old names, and so they are safe to remove.

Once this issue is merged, the same will need to be carried out in the wpcom public themes repo


dsas:
@alaczek can you just sanity check these are the right ones to delete? None of them have wp.com/theme/ pages.

dsas:
Validate Themes failure is fine - it can't validate the themes if the change is that they've been deleted as there's nothing to validate :)

alaczek:
Yep, these are the themes to remove 👍

alaczek (approved):

PR #8071, opened 2024-09-05
Sprinkler: addressing dotorg feedback

MaggieCabrera:
This PR addresses the dotorg feedback for this theme:

https://themes.trac.wordpress.org/ticket/187578

- [x] Skip links
- [ ] Image license


MaggieCabrera:
The image license problem is summarized here: https://themesbycarolina.com/how-to-check-the-license-on-pexels-com/

Should we look for other images for this theme? Also, Kafleg is asking to credit the images don't he screenshot, that are not loaded in the theme.

@beafialho @henriqueiamarino @jasmussen

beafialho:
@MaggieCabrera to provide some context. This theme was part of an effort of theme "spin-offs" for WordPress.com. Essentially reusing existing themes and applying different demo content, different styles to fill the gap for particular verticals. We were told we could use Unsplash, Pexels, etc for those. I don't recall how many images use the Pexels license in this theme, but perhaps we don't launch this one for now on Dotorg? Also curious about what other think.

MaggieCabrera:
I'm going to merge the skip links fixes and for now leave the image issue unresolved. We can submit to dotcom and then revisit this for .org

PR #8076, opened 2024-09-05
The Shore: addressing dotorg feedback

MaggieCabrera:
This PR addresses the dotorg feedback for this theme:

https://themes.trac.wordpress.org/ticket/187526

- [x] Skip links
- [ ] Image license

### DIFF
--- a/the-shore/patterns/home.php
+++ b/the-shore/patterns/home.php
@@ -8,23 +8,24 @@
 ?>
 <!-- wp:template-part {"slug":"header-homepage"} /-->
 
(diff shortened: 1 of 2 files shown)


MaggieCabrera:
This one has the same problem with the midjourney images. I made the other fixes, will leave this PR here in case we want to change the images for dotorg @beafialho @jasmussen

jasmussen:
Current instinct: we can merge and launch this on WP.com pending further clarification on requirements, and then hope a theme of the same name doesn't land on wp.org in the mean time. Depending on the outcome, we can then either replace the demo content images for the core version, not launch it at all, or otherwise.

PR #8081, opened 2024-09-06
Fictionary: Add new theme

MaggieCabrera:
pNEWy-hZB-p2


alaczek:
A had a quick look at this one:
* I'm not sure we need to include images with this theme as they are not used in templates/patterns.
* We had a discussion a while back and decided not to include copyright messages in themes: pekYwv-4ok-p2
* the home template has a static link that goes to staging site (I think the link can be removed)
* the footer template part is not using the footer pattern

iamtakashi:
I'm cleaning up this theme now.

iamtakashi:
I addressed all the feedback and cleaned it up so it has a better chance of being approved in the theme review.

iamtakashi:
@alaczek I thought you might want to review it with the changes I made, but It passed the automated check, so I'll merge it and move on. 🏃‍♂️

PR #8082, opened 2024-09-06
Impressionist: Add new theme

MaggieCabrera:
pNEWy-i2a-p2


alaczek:
I found some super old theme with the same name when I googled "Impressionist WordPress theme" - not sure if that's going to be picked up by theme check or not.

alaczek:
A few more things jumped out at me:
* some images are too large and need compressing
* the front page template needs a corresponding pattern to insert the path to the cover image
* the 404 template isn't using the 404 pattern
* there are two custom page templates in there that I'm not sure are needed?

iamtakashi:
@alaczek I have made a fair amount of code changes to address your feedback and improve the theme. Could you take a look at it? Thanks in advance!

I suspect we may need to change the name. I wish we could check it before submitting a theme.

alaczek:
Yep, this is looking good to me! How about you try submitting this to .org, and if we hear back about the name we can revert the PR and start a new one under a different name?

MaggieCabrera:
I can't approve cause I opened this PR but this looks good to me, let's submit to .org and as soon as we know we can use the name, put it on the .com launch queue

impressionist/readme.txt
```diff
@@ -0,0 +1,55 @@
...
+
+Cardo Font
+Copyright (c) 2002-2011, David J. Perry (hospes02@scholarsfonts.net) 
+This Font Software is licensed under the SIL Open Font License, Version 1.1. This license is available with a FAQ at: http://scripts.sil.org/OFL  
+Source: http://scholarsfonts.net
+-- End of Cardo Font credits --
+
+impressionist-home-bg.jpg
+CC0, Public Domain
+https://www.rawpixel.com/image/2677419/free-illustration-image-winter-monet-snow
```
MaggieCabrera: We probably want to change the format of this, since the .org reviewers always ask for it to be something like:

```
Credit Jakub Kapusnak
License: Creative Commons Zero (CC0) license
License URL: https://www.foodiesfeed.com/license/
Source: https://www.foodiesfeed.com/free-food-photo/sea-crab-at-fish-market/
```

MaggieCabrera: And also say where the images are used, including the theme screenshot

iamtakashi:
I submitted it here: https://themes.trac.wordpress.org/ticket/189282. The automated check didn't tell me anything worrying. It might just be ok. I'll update it here when I hear something from them.

impressionist/readme.txt
iamtakashi: Good call. I edited it before submitting it.

alaczek:
I believe we're in the clear on the theme name - it comes up automatically as a part of initial theme check.

I'll deploy to .com today and you can proceed with demo and showcase page (and launch too if you want).

PR #8090, opened 2024-09-10
Bump node version for theme version workflow

vcanales:
The minimum version required by the project is not compatible with what we were currently using.


mikachan (approved):
Looks good, thank you!

PR #8099, opened 2024-09-11
Theme rename: Farmhouse to Greenseed

MaggieCabrera:
Farmhouse's name [was taken](https://themes.trac.wordpress.org/ticket/187528), so we need to rename it.


MaggieCabrera:
@alaczek @beafialho the only thing that's missing is the screenshot change, but I don't think I have access to the demo site

Issue #8107, opened 2024-09-12
[DotCom Launch]: Greenseed


MaggieCabrera:
### DotCom launch checklist

- [x] Create [demo](https://greenseeddemo.wordpress.com/) site
- [x] Create [theme showcase](https://theme.wordpress.com/themes/greenseed/)
- [x] Network activate
- [x] Launch theme

The demo site has been created but the theme has been renamed (old Farmhouse), we need to delete the old one and activate the new one

alaczek:
This issue is a blocker for launch: https://github.com/Automattic/themes/issues/8102

iamtakashi:
#8102 fixes the missing image issue, and everything looks good on the front of the site. But, in the editor, the images are still down the page initially, and I think we'll need to fix it before launching it.

The trick the theme does is the group block with a custom CSS class `overlay` has an [absolute positioning](https://github.com/Automattic/themes/blob/trunk/greenseed/style.css#L70). That makes the wrapper's height (the group block with a sticky positioning) is zero, and everything else below comes at the top of the browser. That works well on the front of the site, but not in the editor at the moment.

Is there a way to add a custom style `position: absolute;` to the Stack block that has the `overlay` class **in the editor** like simulated in the screenshot?

![CleanShot 2024-09-17 at 19 17 00@2x](https://github.com/user-attachments/assets/d359e5eb-02c6-4b2d-b3ad-c58a92c1fa52)

iamtakashi:
https://themes.trac.wordpress.org/ticket/188382

iamtakashi:
There doesn't seem to be a way to fix it with the current approach. I'll work on the theme differently to achieve the effect.

iamtakashi:
I fixed the theme by taking a different approach to achieve what Bea wanted. I'm launching and submitting this today.

iamtakashi:
Launched on dotcom and submitted to dotorg https://themes.trac.wordpress.org/ticket/193734

PR #8123, opened 2024-09-13
Greenseed: fix images

### LINKED ISSUE #8102: Greenseed: Broken images
I think during the renaming process the asset images got changed in patterns, while the actual file names were not. Example:
https://github.com/Automattic/themes/blob/b5eb7db123801a9b6eb700b722a8a8334adfc7be/greenseed/patterns/home.php#L29

cc @MaggieCabrera

MaggieCabrera:
The images needed renaming when we renamed the theme. 

I still need to scroll to see the images when on the site editor. I'm unsure if this si an editor bug or if it's intended, in the frontend we do see the images without having to scroll down.

#### Related issue(s):
https://github.com/Automattic/themes/issues/8102


iamtakashi:
The rename solves #8102. Thanks @MaggieCabrera. I'll investigate what's happening in the editor in #8107.

PR #8130, opened 2024-09-13
Colorloops: add theme

scruffian:


henriqueiamarino (approved):

iamtakashi:
@henriqueiamarino Very cool aesthetic!

- I'd recommend that each query loop block on the front page template not include sticky posts, especially in this theme, where multiple query loop blocks are used. We don't see sticky posts in the editor, but the front of the site actually shows them and the multiple sticky posts seem like a bug. I believe this is an editor bug, and it's filed in [here](https://github.com/WordPress/gutenberg/issues/55317) and [there](https://github.com/WordPress/gutenberg/issues/64083).

CleanShot 2024-09-25 at 15 32 38@2x

CleanShot 2024-09-25 at 15 34 27@2x

- The post titles on the index template are tiny compared to other templates. It seems like an oversight.
- There is a single-line break tag in the heading on the 404 template. Do you actually want a line break there? 
- I'd recommend not using "lorem ipsum" filler in a theme template/template part. It feels less intentional. Something small, like a made-up paragraph, would be much nicer.

CleanShot 2024-09-25 at 16 05 54@2x

- The following patterns are not used anywhere in the theme. They should be removed.
   - `colorloops/hidden-no-results-content`
   - `colorloops/separator`
   - `colorloops/comments` 
   - `colorloops/page`

Let me know if I need to clarify anything!

iamtakashi:
Sorry, I forgot to add. This probably has something to do with [the bug in the CBT plugin](https://github.com/WordPress/create-block-theme/issues/573), but the name WordPress in the footer credit is no longer a link. I needed to fix it manually in Cubico. See [here](https://github.com/Automattic/themes/blob/trunk/cubico/patterns/footer.php#L35). But it's annoying that next time you save the theme, the HTML appears. This is also a bug in CBT. I'll add a comment again in the issue.

henriqueiamarino:
Thanks, @iamtakashi. I checked and fixed everything you mentioned.
I also removed the unused Patterns, although I believe CBT has been adding them to all themes.

iamtakashi:
Thanks for the update. I see a single line break in the paragraph on the sidebar. Everything else looks good.
CleanShot 2024-09-26 at 14 17 32@2x

> I believe CBT has been adding them to all themes.

No, they are patterns from the base theme you worked on to make this theme. At this time, we need to manually remove those unused patterns inherited from the base theme after exporting the theme.

PR #8134, opened 2024-09-13
Dossier: add theme

scruffian:


henriqueiamarino (approved):

iamtakashi:
@henriqueiamarino  I love the aesthetic, and the theme is solid. Just a few notes. 

* The `dossier/comments` pattern isn't used. The theme currently uses the `comments` template part. Since the template part uses the labels `Older` and `Newer`, which should be translated, I think the template part should be a pattern. Here is what you could do.
  1.  Delete everything in the `dossier/comments` pattern apart from the header. 
CleanShot 2024-10-02 at 16 17 11@2x
  2. Copy all the markup in the `comments` template part and paste it to the `dossier/comments` pattern. 
CleanShot 2024-10-02 at 16 27 47@2x
  3. Make the labels translatable. It should look like this. 
CleanShot 2024-10-02 at 16 20 50@2x
--
* I expected these to be clickable with the arrows. Maybe we could add links to them or use the details blocks? 
CleanShot 2024-10-02 at 16 50 02@2x
* A pagination is missing on the index template.
* The license info about the bundled image (used in the homepage template) is missing from the `readme.txt`.

henriqueiamarino:
What I did:
- Fixed the comments pattern following @iamtakashi  observations;
- Made the labels translatable;
- Added the pagination missing to the Index;
- Adjusted the license information of the bundled image.

What I couldn't do:
- Set the two headings displayed on the 'About the Chef' section as links.  CBT is still breaking the HTML markup when we try to add a # to it. Anyway, I'll link them to the About page at my demo site to show how they should work.

iamtakashi:
@henriqueiamarino Thanks for the update. I made a few small changes. You can see the details in each commit above.
* I've added a reference to the comments pattern from the comments template part(I didn't include this last step in the instruction. My fault)
* Fixed typos in the comments and the 404 patterns.
* Removed unnecessary theme tags and the theme URI from the CSS header (the Themeshaper page didn't seem to exist, but this is optional, so it should be fine for the review.)

If you are happy with the theme, merge and ship it!

henriqueiamarino (approved):

PR #8148, opened 2024-09-13
Add Startorg

scruffian:

Has an error:

```
Notice: Function gutenberg_get_typography_value_and_unit was called incorrectly. Raw size value must be a string, integer or a float. Please see [Debugging in WordPress](https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/) for more information.
```


iamtakashi:
I worked on this theme to make it launchable. It's good to go.

PR #8154, opened 2024-09-14
Bysshe: Add theme

iamtakashi:
Add the Bysshe theme created by @mtias and @pablohoneyhoney. pNEWy-hIc-p2

This theme carries an image from Unsplash, this will not be submitted to WP.org and be available on WP.com only.

screenshot


iamtakashi:
> Found a reference to unsplash.com

Yes, the designer, Matías, is fine with this theme being only on WP.com. Let me know when it's ok to merge.

MaggieCabrera:
This is looking good to me. I don't know about the paragraphs with the dots and translations, since they are decorative, maybe we could use `esc_html_x` for those to add context for translators so they know that they are just symbols? It's still not super accessible but I think it shouldn't be a blocker

`esc_html_x( 'Title', 'context for translators', 'text-domain' );`

MaggieCabrera (approved):

iamtakashi:
@MaggieCabrera Thanks for the review. That is a good point to use `esc_html_x` for those decorative text elements, but I wonder if they need to be translated. What do you think?

iamtakashi:
Anyway, I've changed it to use `esc_html_x()` for the symbols. I'll merge it when it passes the automated checks.

PR #8170, opened 2024-09-17
Add theme-check action

matiasbenedetto:

Add a theme-check github action that runs theme-check plugin checks on the themes that were modified in a particular PR. The action post their reports as a comment. 

It's using the [theme-check wp-cli](https://github.com/WordPress/theme-check/pull/458) from this PR until it's merged.
I created a [demo repo to test it](https://github.com/matiasbenedetto/themes-test/pull/2).

## Screenshot
The action should post a message like this:

image

### DIFF
--- a/.github/workflows/theme-check.yml
+++ b/.github/workflows/theme-check.yml
@@ -201,13 +201,12 @@ jobs:
             fullComment += '--- \n';
           }
           
-          github.rest.issues.createComment({
-            issue_number: context.issue.number,
-            owner: context.repo.owner,
-            repo: context.repo.repo,
-            body: fullComment
-          });
-
           if (hasErrors) {
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: fullComment
+            });
             core.setFailed('One or more themes have required changes. Please review the comments and make necessary updates.');
           }
--- a/.github/workflows/theme-check.yml
+++ b/.github/workflows/theme-check.yml
@@ -1,0 +1,236 @@
+name: WordPress Theme Check
+
+on:
+  pull_request:
+    types: [opened, labeled, synchronize]
+
+# Add this permissions block
+permissions:
+  pull-requests: write
+
+jobs:
+  theme-check:
+    runs-on: ubuntu-latest
+    
+    services:
+      wordpress:
+        image: wordpress:php8.2
+        ports:
+          - 8080:80
+        env:
+          WORDPRESS_DB_HOST: db
+          WORDPRESS_DB_USER: wordpress
+          WORDPRESS_DB_PASSWORD: wordpress
+          WORDPRESS_DB_NAME: wordpress
+      db:
+        image: mysql:8.0
+        env:
+          MYSQL_ROOT_PASSWORD: rootpassword
+          MYSQL_DATABASE: wordpress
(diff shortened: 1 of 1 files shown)


.github/workflows/theme-check.yml
```diff
@@ -0,0 +1,236 @@
+name: WordPress Theme Check
+
+on:
+  pull_request:
```
vcanales, line 4: Using `pull_request` means that the [action won't be able to post comments on PRs created from a fork.](https://github.com/marketplace/actions/add-pr-comment#proxy-for-fork-based-prs) A workaround for this is using `pull_request_target` instead, but we would have to make sure that the audit is free from exploit opportunities before merging, as that event runs with higher permissions.

.github/workflows/theme-check.yml
```diff
@@ -0,0 +1,236 @@
...
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+
+    - name: Install WP-CLI inside WordPress container
+      run: |
+        docker exec $(docker ps -qf "ancestor=wordpress:php8.2") curl -O https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar
+        docker exec $(docker ps -qf "ancestor=wordpress:php8.2") chmod +x wp-cli.phar
+        docker exec $(docker ps -qf "ancestor=wordpress:php8.2") mv wp-cli.phar /usr/local/bin/wp
```
vcanales: Could we use `@wordpress/env` instead? I tested the theme-check cli integration using `npx @wordpress/env run cli`

vcanales (approved):
Works as advertised!

For other iterations, could we consider only posting a message if there are errors in order to avoid extra comments?

.github/workflows/theme-check.yml
matiasbenedetto: Yes, that is good suggestion; I replaced the plain docker definition by the use the `@wordpress/env.` The code looks a bit simpler now. https://github.com/Automattic/themes/pull/8170/commits/2840740f6b1f01977616281e974427978f5fc4bb

.github/workflows/theme-check.yml
matiasbenedetto: I'm not sure what's the way to go around this. I think you researched this before for the playground preview action, right? Could you provide more context, please?

matiasbenedetto:
> For other iterations, could we consider only posting a message if there are errors in order to avoid extra comments?


:+1:  Yes, implemented that here: https://github.com/Automattic/themes/pull/8170/commits/6f6161863131636097ff2b2c318043922909a9fa

.github/workflows/theme-check.yml
vcanales: Yeah, this action looks safe because we are not loading code from outside the YML file. The risk we faced when using `pull_request_target` earlier came from the fact that the YML file was checking out a file from the same repo, which could potentially be modified by an attacker.

Here's a more in-depth and general explanation: https://github.com/CycodeLabs/raven/blob/main/docs/Pull%20Request%20Injections/README.md#2-pull_request_target--checkout

> The real danger arises when such a workflow is combined with checkout action, which checks out code from an incoming, potentially untrusted pull request and then executes scripts or runs commands based on that code.

matiasbenedetto: Thanks for the context! I think we can merge this PR and research more in case we need forked repos run this action.

PR #8188, opened 2024-09-19
The Shore: Added Site Title Block to Headers

### LINKED ISSUE #8168: The Shore: Header missing site title
The Shore does not include the site title in the header, only Logo and navigation. This causes confusion for the customers, and we should avoid this design pattern.

Could we please add site title to the header? It's ok if the demo site displays the logo only.

cc @iamtakashi @MaggieCabrera

iamtakashi:
Added Site Title Block to Headers. Fixes #8168.

### DIFF
--- a/the-shore/patterns/header-homepage.php
+++ b/the-shore/patterns/header-homepage.php
@@ -15,9 +15,9 @@
 <!-- wp:group {"align":"full","layout":{"type":"default"}} -->
 <div class="wp-block-group alignfull"><!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|40","bottom":"var:preset|spacing|40","left":"5vw","right":"5vw"}}},"layout":{"type":"flex","justifyContent":"space-between","verticalAlignment":"center"}} -->
 <div class="wp-block-group alignfull" style="padding-top:var(--wp--preset--spacing--40);padding-right:5vw;padding-bottom:var(--wp--preset--spacing--40);padding-left:5vw"><!-- wp:group {"layout":{"type":"flex"}} -->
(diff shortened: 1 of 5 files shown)


the-shore/patterns/header-homepage.php
```diff
@@ -15,9 +15,9 @@
 <!-- wp:group {"align":"full","layout":{"type":"default"}} -->
 <div class="wp-block-group alignfull"><!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|40","bottom":"var:preset|spacing|40","left":"5vw","right":"5vw"}}},"layout":{"type":"flex","justifyContent":"space-between","verticalAlignment":"center"}} -->
 <div class="wp-block-group alignfull" style="padding-top:var(--wp--preset--spacing--40);padding-right:5vw;padding-bottom:var(--wp--preset--spacing--40);padding-left:5vw"><!-- wp:group {"layout":{"type":"flex"}} -->
-<div class="wp-block-group"><!-- wp:image {"width":"200px","sizeSlug":"full","linkDestination":"none"} -->
```
alaczek, line 18: Should we keep the `"width":"200px"` param? Without it the logo is tiny.

iamtakashi:
> Should we keep the "width":"200px" param? Without it the logo is tiny.

I'm not sure. I kept it initially, but it felt too large when the logo was 1:1. I removed it and left it as default because it depends on the aspect ratio of the logo, and the site admin is expected to adjust the size accordingly.

localhost local_ (16)

alaczek:
>  and the site admin is expected to adjust the size accordingly.

Fair enough, I'll make a note about this on the showcase page.

PR #8198, opened 2024-09-24
Fix theme validation and theme check workflow errors

madhusudhand:
#### Theme Validation workflow

1. This PR includes `style.css` changes as PR paths to validate. Without it, the validation workflow doesn't not validate themes with style.css only changes.

2. Also fixes minor issue where args is empty array.

```
 file:///home/runner/work/themes/themes/theme-utils.mjs:148
			while ( args[ 0 ].startsWith( '--' ) ) {
			                  ^

TypeError: Cannot read properties of undefined (reading 'startsWith')
```

#### Theme Check workflow

It fixes the following issue with theme check [workflow](https://github.com/Automattic/themes/actions/runs/11011254270/job/30574917285) (also simplifies the directory finding logic)

```
Run IFS=',' read -ra FOLDERS <<< ".github"
Processing theme folder: .github
ℹ Starting 'wp theme activate .github' on the cli container. 

Error: The '.github' theme could not be found.
✖ Command failed with exit code 1
Command failed with exit code 1
Error: Process completed with exit code 1.
```

### Testing instructions:

A test PR #8199 should run successfully.
Please add any other scenarios to validate.

### DIFF
--- a/.github/workflows/theme-check.yml
+++ b/.github/workflows/theme-check.yml
@@ -68,7 +68,7 @@ jobs:
             const rootFolders = new Set();
             for (const file of allFiles) {
               const parts = file.filename.split('/');
-              if (parts.length > 1) {
+              if (parts.length > 1 && !parts[0].startsWith('.')) {
                 rootFolders.add(parts[0]);
               }
             }
--- a/.github/workflows/theme-check.yml
+++ b/.github/workflows/theme-check.yml
@@ -41,41 +41,9 @@ jobs:
         wp-env run cli wp plugin install https://github.com/WordPress/theme-check/archive/refs/pull/458/head.zip --activate
 
     - name: Get changed root folders
-      uses: actions/github-script@v6
-      with:
-        script: |
-            const { owner, repo, number } = context.issue;
-            let allFiles = [];
-            let page = 1;
-            
-            while (true) {
-              const { data: files } = await github.rest.pulls.listFiles({
-                owner,
-                repo,
-                pull_number: number,
-                per_page: 100,
-                page: page
-              });
-              
-              allFiles = allFiles.concat(files);
-              
-              if (files.length < 100) break;
-              page++;
-            }
-
-            console.log('files', allFiles.map(file => file.filename));
-            
-            const rootFolders = new Set();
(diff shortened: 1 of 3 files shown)


matiasbenedetto (approved):
Looks good to me :+1:

.github/workflows/theme-check.yml
```diff
@@ -41,41 +41,9 @@ jobs:
...
-                rootFolders.add(parts[0]);
-              }
-            }
-            
-            const rootFoldersArray = Array.from(rootFolders);
-            core.exportVariable('THEME_FOLDERS', rootFoldersArray.join(','));
-            console.log(`Changed root folders: ${rootFoldersArray.join(', ')}`);
+      run: |
+        THEME_FOLDERS=$(git diff --name-only HEAD^ | awk '!/^\./ && /\//' | awk -F/ '{print $1}' | uniq | paste -s -d, -)
+        echo "THEME_FOLDERS=${THEME_FOLDERS}" >> $GITHUB_ENV
```
vcanales: This logic might run into the same issue that we had in the theme previews workflow, where changes made to `trunk` after the branch and PR are created would show up—i.e. changes that are not being made in the PR.

I suggest comparing to the the merge base commit instead, which you can get with `gh`, [like here](https://github.com/vcanales/action-wp-playground-pr-preview/blob/trunk/src/get-changed-themes.ts#L18), because `action/checkout` [does not support merge-base](https://github.com/actions/checkout/discussions/423).

vcanales: Moreover, `git diff --name-only HEAD^` compares the most recent commit (`HEAD`) with its parent commit (`HEAD^`). If there are more than 2 commits, changes made in earlier ones might be missed.

matiasbenedetto: Oh, I missed that when reviewing. I think we can remove that change and continue using the GitHub API as in the trunk version of this action.

madhusudhand: Thanks for pointing. Reverted the workflow to previous state and added f1a5330e80ee10057be99bd5bd137b758f21195e to exclude non-theme directories.

PR #8200, opened 2024-09-24
Use theme-check plugin master branch instead of a particular PR

matiasbenedetto:
## What?
Use theme-check plugin master branch instead of a particular PR

## Why?
We were using a particular PR that included the wp-cli integration to run the theme-check plugin in the CI environment. That branch is merged now, so we can use master.

### DIFF
--- a/.github/workflows/theme-check.yml
+++ b/.github/workflows/theme-check.yml
@@ -38,7 +38,7 @@ jobs:
 
     - name: Install and activate Theme Check plugin
       run: |
-        wp-env run cli wp plugin install https://github.com/WordPress/theme-check/archive/refs/pull/458/head.zip --activate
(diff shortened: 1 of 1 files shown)


mikachan (approved):
Changes LGTM 👍 

As a side note, looks like we should probably only run the WordPress Theme Check workflow on PRs that change files in themes - maybe check for file changes in subfolders, with an exclude list for folders like .github?

matiasbenedetto:
> As a side note, looks like we should probably only run the WordPress Theme Check workflow on PRs that change files in themes - maybe check for file changes in subfolders, with an exclude list for folders like .github?

Yes, that's in progress in https://github.com/Automattic/themes/pull/8198

mikachan:
> Yes, that's in progress in https://github.com/Automattic/themes/pull/8198

Ah sorry, I hadn't seen this PR. Nice!

PR #8204, opened 2024-09-25
Raconteur: general fixes after dotorg review

henriqueiamarino:
- This PR fixes some type sizes and spacings to improve the theme's responsiveness.
- It also removes an HTML that was displayed in the front: https://i.rankmath.com/i/8VPxa9


iamtakashi:
* The header template part is no longer used in the theme. If there are templates that share the same element such as header and footer, they should be template parts. Otherwise, the user must customise those separately on all the templates.
* The same is true for the footer.
* The comments template part is also unused in this theme.
* The `templateParts` in the theme.json should match with the templates parts used in the theme.
* The following patterns should be deleted if they are not used in the theme.
  * `raconteur/hidden-404`
  * `raconteur/hidden-archive`
  * `raconteur/hidden-index`
  * `raconteur/hidden-page`
  * `raconteur/hidden-search`
  * `raconteur/hidden-single`
  * `raconteur/page`
  * `raconteur/posts-post-list`

Could you fix these things first? I can check the front of the site once the theme files are in order.

henriqueiamarino:
@iamtakashi, the header is used as the first column, where the site title and tagline are. You are right about the Footer, but I didn't want to remove it because it was mandatory to have this pattern somewhere.
The other patterns, as usual, were not added by me. I suppose CBT has been adding some to several themes. But if you think it's okay, I can remove them; thanks.

henriqueiamarino:
@iamtakashi unused parts and patterns removed.

iamtakashi:
Thank you! I'll take a look at it again.

iamtakashi:
* You might have already seen in my Colorloops review, but they are from the base theme you used to build this theme. We need to remove them manually after exporting the theme. We still need to remove the following patterns because I don't think they are used in any template.
  * `raconteur/footer`
  * `raconteur/hidden-comments`
* There are two `main` tags in `404`. They are probably better as default`div`, and the one with the orange arrow should be `main`. 
CleanShot 2024-09-26 at 16 35 59@2x
* These should match the actual template parts in the theme. CleanShot 2024-09-26 at 16 40 08@2x
* The same as Colorloops let's not include sticky posts in the query loops on the front page. 
localhost local_ (14)
* It's better to include the query no results block with a message so we can avoid things like this.  
CleanShot 2024-09-26 at 16 49 53@2x
* The drop cap makes several blocks look broken. Can we do something about it?
CleanShot 2024-09-26 at 16 52 30@2x 
CleanShot 2024-09-26 at 16 54 39@2x 
CleanShot 2024-09-26 at 16 57 58@2x

Let me know if I can clarify anything!

henriqueiamarino:
Thanks, @iamtakashi; I'll check that. Unfortunately, I'll have to remove the drop cap. I can't make it work better than that.

henriqueiamarino:
I have fixed all the issues mentioned by @iamtakashi, and I am merging this update:

- Removed the drop cap from the content element;
- Fixed the duplicated <main> tag of the 404 page;
- Deleted unused patterns: footer and hidden comments;
- Created a template part for comments;
- Fixed template parts section on theme.json;
- Set sticky posts as **excluded** for all QLs;
- Added **No Results** to all QL, including the Homepage.

henriqueiamarino:
In time, I'll document the drop cap issue here to make it available in the future. I tried to make it work by adding a class to the **Content** element, but the results showed it was a bad idea. How can we have it working for the first paragraph only?

Screenshot 2024-09-27 at 10 07 02

iamtakashi:
Thanks for the update on fixing the issues! I'll see what we can do about the drop cap.

PR #8027, opened 2024-08-21
Assembler: Improving styles

richtabor:
#### Changes proposed in this Pull Request:
- Set color inheritance for headings with links at the theme level.
- Remove text-wrap balance, as its conflicting with the editor.
- Add subtle transition effect for buttons. 
- Reduce footprint of styles, which allow for simpler section styles (and simpler overriding of styles).
- Update style variations.

### Visual 

https://github.com/user-attachments/assets/f7e996b8-03d8-48fb-9585-562b192591e8

PR #8206, opened 2024-09-26
Assembler: Remove inherited color styling from linked headings

### LINKED ISSUE #8193: [HC] Assembler: Link hover color not updating on Post titles
### Quick summary

Changes do not apply when changing the Hover color on Post titles in the Assembler theme.

#hc

### Steps to reproduce

1. Load up the Assembler theme.
2. Add the Query Loop block and choose any pattern.
3. Click on a Post Title and change the Link Hover color on block's Styles section.
4. Hover over the post title to see if the color changes.

Post title color should change when hovering over it.

### What actually happened

Post title color doesn't change.

image

### Impact

Some (< 50%)

### Available workarounds?

_No response_

Simple, Atomic

### Logs or notes

Reported in 8759313-zen

Tested this on another block theme (Dawson) but could not reproduce.

image

ivan-ottinger:
Fixes https://github.com/Automattic/themes/issues/8193.

- Assembler: Remove inherited color styling from linked headings

There doesn't seem to be a need for the `!important` inherited color override that is causing small regression with the hover link color.

#### Testing instructions:

1. Add `Title` and `Heading` blocks to a page / post and make sure these titles are linked somewhere (`#` should be enough).
2. Try changing blocks' colors, including hover color through the editor sidebar.
3. Adjust link and text colors through the Global Styles.
4. There should be no regressions or unexpected color styling on the published page / post.
5. Try to reproduce https://github.com/Automattic/themes/issues/8193. It shouldn't be possible.

#### Related issue(s):

- https://github.com/Automattic/themes/issues/8193
- related comment: https://github.com/Automattic/themes/pull/8027#pullrequestreview-2328174006

### DIFF
--- a/assembler/style.css
+++ b/assembler/style.css
@@ -14,12 +14,6 @@ Text Domain: assembler
 Tags: blog, one-column, three-columns, wide-blocks, block-patterns, custom-colors, custom-logo, custom-menu, editor-style, featured-images, full-site-editing, rtl-language-support, style-variations, template-editing, theme-options, threaded-comments, translation-ready
 */
 
-/* Make link colors match text colors. */
-
-h1 a, h2 a, h3 a, h4 a, h5 a, h6 a {
-    color: inherit;
-}
-
 /* Remove auto-applied padding on headings when color is applied. */
 
 h1.has-background, h2.has-background, h3.has-background, h4.has-background, h5.has-background, h6.has-background {
--- a/assembler/style.css
+++ b/assembler/style.css
@@ -14,12 +14,6 @@ Text Domain: assembler
 Tags: blog, one-column, three-columns, wide-blocks, block-patterns, custom-colors, custom-logo, custom-menu, editor-style, featured-images, full-site-editing, rtl-language-support, style-variations, template-editing, theme-options, threaded-comments, translation-ready
 */
 
-/* Make link colors match text colors. */
-
-h1 a, h2 a, h3 a, h4 a, h5 a, h6 a {
-    color: inherit !important;
-}
-
 /* Remove auto-applied padding on headings when color is applied. */
 
 h1.has-background, h2.has-background, h3.has-background, h4.has-background, h5.has-background, h6.has-background {
--- a/assembler/style.css
+++ b/assembler/style.css
@@ -14,6 +14,12 @@ Text Domain: assembler
 Tags: blog, one-column, three-columns, wide-blocks, block-patterns, custom-colors, custom-logo, custom-menu, editor-style, featured-images, full-site-editing, rtl-language-support, style-variations, template-editing, theme-options, threaded-comments, translation-ready
 */
 
+/* Make link colors match text colors. */
+
+h1 a, h2 a, h3 a, h4 a, h5 a, h6 a {
+    color: inherit;
+}
+
 /* Remove auto-applied padding on headings when color is applied. */
 
 h1.has-background, h2.has-background, h3.has-background, h4.has-background, h5.has-background, h6.has-background {


assembler/style.css
```diff
@@ -14,12 +14,6 @@ Text Domain: assembler
 Tags: blog, one-column, three-columns, wide-blocks, block-patterns, custom-colors, custom-logo, custom-menu, editor-style, featured-images, full-site-editing, rtl-language-support, style-variations, template-editing, theme-options, threaded-comments, translation-ready
 */
 
-/* Make link colors match text colors. */
-
-h1 a, h2 a, h3 a, h4 a, h5 a, h6 a {
-    color: inherit !important;
```
zaguiini, line 20: What happens if we remove the `!important` but keep the inheritance?

allilevine:
Changing the link color and hover color worked for me in a post, but I'm not seeing the hover color when editing a template:

Screen Shot 2024-09-26 at 9 16 13 AM

Is that expected?

assembler/style.css
ivan-ottinger: Thanks, Luis. Good point. I did not observe any issues when we remove the `!important`. It might be a safer approach to keep the color inheritance in place. → Added it back in the latest commit.

ivan-ottinger:
> Changing the link color and hover color worked for me in a post, but I'm not seeing the hover color when editing a template:

Thanks for catching that, Allison! Keeping the color inheritance in and removing just the `!important` as Luis [suggested above](https://github.com/Automattic/themes/pull/8206#pullrequestreview-2331214067) fixes that as well. 🙂

allilevine (approved):
This is working well for me now! I see the hover colors while editing and they show up correctly in the published post. The hover color is also visible when editing global styles. ✅

ivan-ottinger:
Thank you for additional review, Allison!

ivan-ottinger:
Hey @Automattic/theam 👋🏼🙂 Just wanted to check if we can merge this PR - or - whether you will be merge and deploying it later?

(I already read the docs here: https://github.com/Automattic/themes/blob/trunk/CONTRIBUTING.md, but haven't contributed to this repo yet)

Thanks!

mikachan:
Thanks for the ping, @ivan-ottinger! Feel free to merge this PR whenever you're ready. As this is a dotcom theme, either @alaczek or @miksansegundo would usually handle the deployment, but I'm also happy to help deploy.

richtabor:
Let me double check; I'm trying to recall why it was added in the first place.

richtabor:
I actually think we can just remove the whole bit, as link color set to currentColor [like this](https://github.com/Automattic/themes/blob/trunk/assembler/theme.json#L966C1-L970C14). I'm pretty sure this was for some specificity issue, but that has since been resolved.

richtabor (approved):
LGTM

ivan-ottinger:
> This is working well for me now! I see the hover colors while editing and they show up correctly in the published post. The hover color is also visible when editing global styles. ✅

Hey @allilevine 👋🏼 Interestingly, all is working well even with the whole color inheritance gone. I think it must have been some kind of glitch before.

The only issue I noticed (not related to this PR, nor the theme either) is that when the Title block is added when editing templates, the anchor isn't being added to the markup:

Markup on 2024-09-27 at 09:03:00

As a result, the styles don't get applied.

It is not clear to me if this is a bug or done on purpose for some reason, so I opened a new issue in the Gutenberg repo: https://github.com/WordPress/gutenberg/issues/65691

---

Thank you for the additional review and merge @mikachan, @alaczek  and @richtabor!

For completeness, here's the difference between having color inheritance in vs not:

### with inheritance - heading link color is inherited from the text color:

Markup on 2024-09-27 at 08:41:53

### without inheritance - actual link color is being used:

Markup on 2024-09-27 at 08:43:23

Issue #8191, opened 2024-09-20
[DotCom Launch]: Colorstream


iamtakashi:
### DotCom launch checklist

- [x] Create [demo site](https://colorstreamdemo.wordpress.com/)
- [x] Create [theme showcase](https://wordpress.com/theme/colorstream)
- [x] Network activate
- [x] Launch theme

iamtakashi:
I'm waiting for the demo content.

In the meantime, @alaczek, could you add the theme tier info to the showcase page?

iamtakashi:
The demo site was created. I can see the Personal tier has been added to the showcase page. If that is correct, it's ready to go.

alaczek:
Is there are a reason this theme comes with both front-page and home templates?

I also noticed the sidebar in single page template is different than in single post temple - was this intentional and could we possibly unify this? Ideally the sidebar would be a reusable template part, and we wouldn't use "Lorem ipsum" filler text.

Single post:
![Image](https://github.com/user-attachments/assets/6d16b324-2379-4a5e-9b16-72e519cf8be6)

Single page:
![Image](https://github.com/user-attachments/assets/d1ee0a9e-6072-4eb8-a355-959c25245e57)


Finally, could we please remove the hard coded navigation ref?
https://github.com/Automattic/themes/blob/trunk/colorstream/parts/header.html#L12

iamtakashi:
Let's not include sticky posts from the query loops on the front page. ![CleanShot 2024-09-26 at 17 15 52@2x](https://github.com/user-attachments/assets/b595eb7d-846a-4ca0-9ba7-0b048ffc54ee)

henriqueiamarino:
@alaczek and @iamtakashi, [here's](https://github.com/Automattic/themes/pull/8207) the merge that fix the theme following your comments. I hope the hard-coded element is gone now; I tried to remove it after Takashi's instructions.

iamtakashi:
@alaczek This should be ready to go after the latest updates get deployed to dotcom.

PR #8207, opened 2024-09-26
Colorstream: unifies sidebars and removes frontpage

henriqueiamarino:
This PR fixes the observations commented [here](https://github.com/Automattic/themes/issues/8191).

- It unifies the two sidebars in a single template part;
- Removes the hard-coded navigation;
- Delete the front page once the homepage works the same way.


iamtakashi:
Can we exclude sticky posts from the query loops on the home template?

henriqueiamarino:
Yes, sure. It's done, @iamtakashi.

PR #8214, opened 2024-09-30
Podbase: add theme

henriqueiamarino:
Sometimes, your podcast episode cover arts deserve more attention than regular thumbnails offer. If you think so, PodBase is the design for your podcast site.

[Demo site](https://podbase.mystagingwebsite.com/)

screenshot


iamtakashi:
CleanShot 2024-09-30 at 14 49 55@2x
* `general` as the area throws errors. `uncategorized` should be used for those patterns. When you are there, if you add a title to each, the name will appear nicely in the editor. For example, 

CleanShot 2024-09-30 at 14 53 59@2x

--

* Does the theme use 'crimson-pro`? If not, let's remove the files and the reference from the JSON file.

--

CleanShot 2024-09-30 at 15 20 20@2x

* To be honest, I didn't expect the inverted colour scheme on other templates. It's unique, but I'd probably consider making other templates light-on-dark, like the homepage. It's up to you, just a thought.
* In my opinion, the homepage, the archive, the index, and the search templates could be better if they share a uniform layout. It makes the theme complicated unnecessarily without a huge gain. The differences I noticed were...
  * the grid
  * the size of the featured image
  * the little text "> listen"
  * the vertical alignment between the featured image and the title.

--

* We could add the sidebar to the 404 and the page templates, too, maybe?

--

* When The "Menu" is clicked on the homepage, we don't see anything because it's light-on-light

--

CleanShot 2024-09-30 at 15 31 35@2x
* Should they be buttons? Maybe we don't need the icons... The vertical alignment between the icon and the text seems to be awkward, anyway.

--

CleanShot 2024-09-30 at 15 50 19@2x
* This might be intentional, but the pagination seems bare. 

--

* There is more than one `main` tag on the archive, the home, the index, the search templates.

--

CleanShot 2024-09-30 at 16 06 48@2x
* Since the border in the search block works oddly, I'd remove it.

--
screenshot
The stylised featured images on the screenshot are cool, but they might give the use a false anticipation.

henriqueiamarino:
My updates have addressed the following:
- Fixed the template part areas;
- Removed unused Crimson Pro;
- Added a sidebar to 404 — yes, I am sure this is useful here;
- Fixed the Menu link colors
- Removed stroke from Search element

henriqueiamarino:
Thanks for your review, @iamtakashi. I am detailing some decisions below:

**Inverted Color Scheme**
This is definitely opinionated, but once I created this system, I couldn't alternate it on variations. It will be constantly inverted, no matter the colors. Since this is a third podcast theme, I want to proceed with a bold approach here; I believe the others ([this](https://castcore.mystagingwebsite.com/) and [this](https://casthub.mystagingwebsite.com/)) will follow your expectations closely.

**Footer Buttons**
I added containers to links in the footer to ensure the black icons would work with both backgrounds. Once I have icons, I need such a safe space. This is nice to have as the design is neat and typographical. There are no decorations here, and the icons work to give users a visual break.
The icons became a problem once I couldn't use social links because PocketCasts was not there. I am using PNGs with an overlay, but I would love a better solution to remove the button aspect. I tried FontAwesome, an icon font, but it didn't work out.

**Query loop designs**
Following your advice, I've streamlined the variations of QL. Now, there are only two designs: the default and the Homepage version, which showcases the podcast cover in a larger size.

**Featured Image**
The stylized featured images emulate podcast covers. I chose not to use plain images here because that's not how podcasts work. Podcast episodes typically have unique covers uploaded along the audio track; I've done it. Therefore, users of a podcast theme will expect to showcase their art. This is not a false anticipation but a connection to the public's intention.

We're good to move on to the next ones, I believe.

iamtakashi:
Thanks for your reply and the update. Okay, understood. There is no blocker to launch. Ship it 🚢 !!

PR #8225, opened 2024-10-02
Kawaii-chan: add theme

henriqueiamarino:
Kawaii-Chan is a simple theme that supports full-site editing. It comes with a set of minimal templates and design settings that can be manipulated through Global Styles. Use it to build something beautiful.

[Demo site](https://kawaii-chan.mystagingwebsite.com/)

screenshot


henriqueiamarino:
@iamtakashi, I know CBT is messing up with background links, but I couldn't fix this. I tried diverse ways to link it to the image in the assets folder, but it didn't work. Please help me here.

Screenshot 2024-10-02 at 11 47 02

iamtakashi:
@henriqueiamarino This makes the background work for me. Let me know if this is not what you wanted to achieve.
```
"background": {
	"backgroundImage": {
		"source": "file",
		"src": "file:./assets/images/kawaii_pattern-bg.png",
		"title": "kawaii_pattern",
		"url": "file:./assets/images/kawaii_pattern-bg.png"
	},
	"backgroundPosition": "50% 0",
	"backgroundSize": "1080px"
}
```
CleanShot 2024-10-02 at 15 13 14@2x

henriqueiamarino:
Thanks @iamtakashi. I tried that before, but as the background disappeared in my version, I thought there was a better solution. Anyway, I fixed it. Maybe it's some local setting that is affecting my side.

I am not sure, though. Apparently, it's also gone on the Playground preview.

iamtakashi:
* As we discussed, let's replace the AI-generated images of the pins. 
* The `rubik-bubbles` and `bellota-text` don't seem to be used anywhere. If that's correct, let's remove the files, the mentioning in `readme.txt` and from the JSON.
*  Both the comments pattern and the comments template part don't seem to be used anywhere.
* There are strings that should be translation-ready in the `404.html`. There is a `404` pattern, but that's not called in the `404.html`.
* The `post-meta` pattern doesn't seem to be used anywhere.
* For a new theme, we only need this in the changelog. 
```
= 1.0.0 =
Initial release
```
* `Designed with WordPress` in the `footer` pattern is not translation-ready.
* The colour variations could use a little polish.
CleanShot 2024-10-03 at 23 46 51@2x
CleanShot 2024-10-03 at 23 49 01@2x
CleanShot 2024-10-03 at 23 49 51@2x

henriqueiamarino:
Thanks for the review, @iamtakashi. Here's my take:
- I replaced the AI-generated images. As I created the new ones myself, the existing credits are okay;
- The two unused fonts should be included in the style variations. They are now adjusted;
- Comments pattern is now used on the single, just as Posta-meta;
- I removed the 404 pattern as it was not being used;
- I fixed the changelog info on the readme.text;
- I fixed the 'Design with WordPress' footer signature.

Finally, on the style variations, I did what was possible. The images that create rounded bottom details use duotones that can't get updated on variations. One day, it will be possible to make duotones variable (like gradients; VAR Base to VAR Contrast, f.i.), but I can't fix it while it is still impossible. I see two possible solutions here:

1 - The first is to provide all the necessary duotones on the theme; users will manage to switch them manually. As this is the less impactful solution, I picked this one.
2 - The more dramatic solution is to remove all style variations while the variable duotone doesn't work.

iamtakashi:
Thanks for the update.

* Now, the theme uses the `comments.html` template pattern with labels (`older` and `newer`) that should be translation-ready. So, just like Dossier, the markup of the template pattern needs to move to the `comment.php` pattern, and you can wrap the labels with the translation functions. And then, you can call the pattern from the template part.
* The `404.html` template still has strings that should be translation-ready, which means it needs to be a pattern. So, create a `404.php` pattern with the markup of the `404.html` template and wrap the strings with translation functions. And then, call the pattern on the template. You've done that in Dossier; look at it if this is confusing.
* I now can see what's going on with the duotones. I probably would do the same. Let's make sure we have an issue filed in the Gutenberg repo.
*

henriqueiamarino:
I fixed the first part, copying the 404.html markup into a new 404.php file.
I could not wrap the strings with translation functions because I didn't know what those were. I could repeat what I did with Dossier if I had pagination elements here, but I don't.

Sorry, @iamtakashi, I can't do better than that and will need extra help here.
Please let me know what to do next.

iamtakashi:
No worries! I'm happy to help. I'll push my fixes so that you can see what I meant in the review.

iamtakashi:
@henriqueiamarino, I've pushed three changes ([for the comments](https://github.com/Automattic/themes/pull/8225/commits/7a715ed88e5fa03b2aba633beca1fea02085d0e4), [the 404](https://github.com/Automattic/themes/pull/8225/commits/e24f1e7f9a1eb79c33d4fe23ccb5fc6a4488d9bc), and one just to remove typos). I hope those are clearer for you to see. Let me know if I can clarify anything further.

iamtakashi:
I think this is ready to merge now. Ship it 🚢

PR #8235, opened 2024-10-04
Podbase: adds sidebar to page template

henriqueiamarino:
This PR adds a sidebar to the page template and also fixes a minor detail on the menu overlay color.


iamtakashi:
Looks good to me. Merging.

PR #8259, opened 2024-10-11
Conference: add theme

henriqueiamarino:
A theme designed for meetups and conferences. Use Conference to create captivating event websites with user-friendly customization options to ensure your gatherings leave a lasting impression.

[Demo site](https://conference.mystagingwebsite.com/)

screenshot


alaczek:
Hi Henrique 👋 I had a quick look at the code and noticed a few things:
* there seem to be both patterns and templates for things like single, page, index, etc. I assume the patterns are redundant, since they are not being used?
* On the other hand there are templates like archive.html and search.html that import a pattern - I'm not sure of this is necessary - the code could live in the template. I think the only exception to this is home template.
* the footer has a broken block:
image

henriqueiamarino:
Thanks for the ping, @alaczek.
- These pattern/template mixes have been happening a lot lately. Can you tell what's causing them? I don't know how to fix those, but I'll ask for help.
- The broken block seems easy to fix. As Takashi recommended, I added the WordPress link differently, but it didn't work. I'll go back to my old fix. CBT keeps breaking links if we add them via Editor.

iamtakashi:
* The theme check plugin says, "WARNING: screenshot.png is 659.8 KB in size. Large file sizes have a negative impact on website performance and loading time. Compress images before using them."
* The Newsreader font isn't working. There is no `newsreader` folder, while the paths for the normal 700, italic 400, italic 700 have the folder. Create a newsreader folder inside the fonts folder, and change the path for the normal 400 to `./assets/fonts/newsreader/newsreader_normal_400.ttf`
CleanShot 2024-10-18 at 18 57 04@2x
* The same for Inter Tight. The font isn't working because there is no folder called `inter-tight` in the fonts folder. 
CleanShot 2024-10-18 at 19 04 36@2x I'm wondering why things like this happen repeatedly in your themes. I don't think I have had anything like this while using the CBT.
* The `comments` pattern, `patterns/comments.php` is unused. It's likely from the base theme. Delete the file.
* Change the `<?php esc_html_e('Designed with WordPress', 'conference');?>` in `footer-pages.php` to `<?php echo wp_kses_post( __ ( 'Designed with <a href="https://wordpress.org" rel="nofollow">WordPress</a>', 'conference' ) ); ?>`
* Change the `<<?php echo wp_kses_post( __ ( 'Designed with <a href="[https://wordpress.org](https://wordpress.org/)" rel="nofollow">WordPress</a> ', 'conference' ) );?>` in `footer.php` to `<?php echo wp_kses_post( __ ( 'Designed with <a href="https://wordpress.org" rel="nofollow">WordPress</a>', 'conference' ) ); ?>` There is a copy and paste error there, that's why it's broken as @alaczek pointed out. 
* The sizes for `h1` and `h3` are static. Use the variables so it can use the fluid size. 
CleanShot 2024-10-18 at 19 29 44@2x
* The size of the heading `Oops! This page...` stays the same on small screens because it's a static value `5.5 rem`. It's better to avoid using a static size and, instead, make a new preset, `xx-large`, for example, and use that in the template. 
CleanShot 2024-10-18 at 19 41 20@2x
* The content on the footer bumps to the edge of the screen. 
CleanShot 2024-10-18 at 19 50 26@2x
* One of the duotone presets, `cobalt`, uses an uppercase C in the slug. This must be a small case, and the reference to the duotone must be changed, too. 
CleanShot 2024-10-18 at 19 55 40@2x
CleanShot 2024-10-18 at 19 58 50@2x

henriqueiamarino:
Thanks, @iamtakashi. I followed your observations and believe I fixed all but the `php echo wp_kses_post` line.
This is still breaking the element... can you still see something wrong?

iamtakashi:
Just to note, this is now a dotcom-only theme because the review team doesn't like the theme name.

PR #8263, opened 2024-10-14
Update blockbase to better ensure type safety

dsas:

The Blockbase custom font migration doesn't deal well with bad data. This increases the ability to deal with it a bit, by checking that some array keys exist before using them.

The error was `Warning: Undefined array key "slug"`, and now strict type checking is enabled for this file on wpcom, this is a fatal type error rather than a warning.

#### Related issue(s):

See p1728779070911429-slack-C4N88L95W

#### How to test

 1. Checkout this file locally
 2. Add `<?php declare( strict_types=1 );?>` as the new first line, to replicate what happens on wpcom
 3. scp to your sandbox
 4. Sandbox the site mentioned in the slack thread linked above
 5. access it via `curl --insecure <url> > test.html`
 6. Check your sandboxes error log, and see there are no errors.
 7. Unsandbox the site
 8. Open test.html in your browser
 9. See that the site is legible.


matiasbenedetto (approved):
LGTM

PR #8268, opened 2024-10-15
Optimise asset images on themes starting with A

sgomes:
This PR optimises some of the images in themes starting with the letter `A`. The changes were generated using `ImageOptim`, and images were visually inspected to ensure no obvious visual artefacts were introduced.

These assets are either loaded by the theme directly (in the case of classic themes), or are part of patterns / templates in the case of block themes.

In the case of classic themes, the images will have a direct performance impact when they're loaded.

In the case of block themes, the images will only have a performance impact if the user does not replace them and does not upload them to their media library. So while they won't have an impact every time, it's still best to make the images smaller in the first place, as a form of mitigation.

- Optimise image assets in themes starting with the letter `A`


sgomes:
Reviewers: please note that any issues above are pre-existing, as the only thing this PR does is optimise a few existing images.

PR #8290, opened 2024-10-18
CoachAva: add theme

henriqueiamarino:
Coach Ava is a theme designed for coaching professionals. It is tailored to showcase services, blog posts, and podcasts. It is the ideal platform to establish a solid online presence and connect coaches with their audience.

[Demo site](https://coachava.mystagingwebsite.com/)

screenshot


iamtakashi:
@henriqueiamarino Sorry for the delay. I'm currently making a review checklist while I review themes. The checklist should help us going forward.

* Since this theme has a custom page template, let's add that to theme.json so the template appears nicely in the editor. 
CleanShot 2024-10-28 at 14 48 02@2x
CleanShot 2024-10-28 at 14 49 17@2x
* The `main` tag is currently used for the cover block on the home template, but I think it'd make more sense if we use it for the content. 
CleanShot 2024-10-28 at 15 33 42@2x
* The `main` tag is missing from `page-with-sidebar`, `page`, and `single` patterns. A `main` tag is necessary on all the templates for the `skip to content` link to work. 
CleanShot 2024-10-29 at 11 22 22@2x
* Let's fix the footer credit with a link in the Footer Basic pattern.
* Remove `comments.php`
* The font `Besley` is missing.
* In the JSON, the site title block uses `Poppins`, which seems to be an oversight.
* The image `landscape_linkedin-1251827.jpg` isn't used in the theme.
* In the JSON, `h2` is set to `3em`. It'd be better if it's one of the preset sizes.
* Let's remove the following tags from the style.css. They are only relevant to classic themes. `custom-colors`, `custom-header`, `custom-logo`, `custom-menu`, `editor-style`, `featured-image-header`, `flexible-header`, `full-width-template`, `post-formats`, `theme-options`.
* Some credits for some of the bundled fonts and the images are missing from the readme.txt`

Let me know if I can clarify anything!

henriqueiamarino:
Thanks, @iamtakashi. The checklist idea will help us a lot.
For now, all the following recommendations have been reviewed:

- Added Besley font pack;
- Formatted custom page template title;
- Fixed main tag on the Homepage; added to Page, Page-with-sidebar, and Single;
- Removed comments.php;
- Fixed WordPress link on Footer Basic;
- Fixed site-title typeface (removing Poppins);
- Removed unused image;
- Fixed H2 yet for the third time — not sure why it was reverted to a single value like that;
- Removed the tags as recommended;
- Added the credit that was missing to the font credits section.

iamtakashi:
The font Basley is still not working. The font files are using uppercase. I recommend using always small case. And `600` aren't bundled. Also, check the `regular` and `normal`.
CleanShot 2024-10-29 at 19 32 52@2x

iamtakashi:
I think we need to clarify that the images I assume you created (the icons and the logos for hypothetical companies) are credited to you and released as public domain.

henriqueiamarino:
You're right, @iamtakashi. I'm still trying to figure out why the Besley folder disappeared, but the ones I used should have been checked before calling it ready. I checked the Playground link, and the fonts are showing.
I also updated the credits for the icons and placeholder logos; thank you.

iamtakashi:
@henriqueiamarino Thank you for the updates. It looks good to me. Merging.

iamtakashi:
@henriqueiamarino This should be good to submitting to the dotorg. Just to make sure, you don't need to bump up the version since this is the first time to submit 👍

PR #8305, opened 2024-10-29
CoachBen: add theme

henriqueiamarino:
Coach Ben is a theme designed for coaching professionals. It is tailored for showcasing services, blog posts, and podcasts — the ideal platform to establish a solid online presence and connect coaches with their audience.

[Demo site](https://coachben.mystagingwebsite.com/)

screenshot


iamtakashi:
I'll review this today!

iamtakashi:
@henriqueiamarino Here are my notes.

* Let's add a No result block to the archive and the search results templates like the index template.
* The query type for the query loop block in the search results template should be the default. 
CleanShot 2024-10-30 at 15 24 49@2x
* The sidebar in the single post template is misaligned. I assume the `page about` template is what you wanted to achieve. If I'm correct, the block spacing is overlooked.
CleanShot 2024-10-30 at 15 31 12@2x
* The page-about template could be called a `Page with Sidebar` template. Let's add it in the JSON, just like you did for CoachAva.
* A `main` tag is missing in the pages template.
* The single posts template there are two `main` tags. Only the outer, `Content` is good.
* The button colour is in low contrast with the text. Especially in the `Lichen` and the `Tin` style variations.
CleanShot 2024-10-30 at 16 13 04@2x 
CleanShot 2024-10-30 at 16 14 08@2x
* The font sizes for the `button`, `buttons`, `post-content`, `post-terms`, and `query-pagination` have absolute size. None of them are big in size, so it's less likely to become a problem with small viewports, but it'd provide a better UX for customisation if they were all presets.
* I think `wp-brand_7.png` is unused. If I'm correct, remove it.
* Let's edit the body text in the Events section on the homepage template so they are hypothetical.
* This is [a bug in the CBT plugin](https://github.com/WordPress/create-block-theme/issues/742), but the custom labels in the pagination blocks are left untouched after saving the theme, and we need to make it internationalised manually :/ The steps are:
1. Create a pattern called `pagination.php` in the `patterns` folder and paste this:
```
<?php
/**
 * Title: pagination
 * Slug: coachben/pagination
 * Inserter: no
 */
?>
<!-- wp:query-pagination {"align":"full","layout":{"type":"flex","justifyContent":"left","flexWrap":"wrap"}} -->
<!-- wp:query-pagination-previous {"label":"<?php esc_html_e( 'Previous', 'coachben' ); ?>"} /-->
<!-- wp:query-pagination-next {"label":"<?php esc_html_e( 'Next', 'coachben' ); ?>"} /-->
<!-- /wp:query-pagination -->
```
2. The pagination is used in three templates, `archive.html`, `index.php`, and `search.php`, so we need to replace it with the newly created pattern above.
3. For example, in `archive.html`, you will delete the pagination block altogether. 
CleanShot 2024-10-30 at 17 32 48@2x
4. Once you deleted the pagination block, you will call the pattern by `<!-- wp:pattern {"slug":"coachben/pagination"} /-->` 
5. This is how it should look when you replace it with the pattern. 
CleanShot 2024-10-30 at 17 28 54@2x
6. Repeat the same in `index.php` and `search.php`
* The `comments.php`, `page-about.php`, `page.php`, and `single.php` aren't used. Remove them.
* Tiny detail: Could the name be something else? 
CleanShot 2024-10-30 at 17 51 09@2x
* Let's add credits to all the fonts bundled in `readme.txt`
* This is probably something the CBT used to do at some point, but there is a full SIL font license, let's remove it.
* Let's add credits to all the images, including the ones you made in the `readme.txt`.
* Let's remove irrelevant theme tags. `block-styles`, `custom-colors`, `custom-header`, `custom-logo`, `custom-menu`, `editor-style`, `featured-image-header`, `flexible-header`, `full-width-template`, `post-formats`, `sticky-post`, and `theme-options`.

Let me know if I can clarify anything further!

henriqueiamarino:
Thanks for the thorough review, @iamtakashi. Here are my fixes:

- Fixed coach name and description on the sidebar;
- Made the Search template QL the same as the Archive;
- Fixed the misalignment on the single template;
- Renamed the page-about as indicated;
- Added a main tag to the Page template and removed an extra from the Single;
- Fixed the button contrast issue: the text color should be the Primary;
- Removed remaining absolute sizes from elements: button, buttons, post-content, post-terms, and query-pagination;
- Delete image wp-brand_7;
- Replaced the copy text of the events section;
- Fixed and repeated the process of making pagination internationalized for archive.html, index.php, and search.php;
- Removed the following unused PHP files: comments, page-about, page, and single;
- Updated style.css tags as required;
- Added credits related to icons and placeholder logos;
- Removed SIL font license info and update the missing font information.

iamtakashi:
@henriqueiamarino Thanks for the update! Most things were addressed correctly 👍 

* You've correctly created the pagination pattern, but the old pagination was still on the templates that needed archive, search, and index templates.
* No results block wasn't added to the archive and the search templates. But my instruction wasn't sufficient—since the paragraph "Sorry, but nothing [...]" in the no results block needed to be internationalised, the archive template needed to be converted to a pattern, and be called from the `archive.html`.

I've pushed my edits, [1](https://github.com/Automattic/themes/pull/8305/commits/4a710daace78a763fc0386f6a5a2c9350b32b995) and [2](https://github.com/Automattic/themes/pull/8305/commits/de1daac5dc7991c30af085dc7dedf2fd2c53c3cc) to address the above, take a look at it for the future. Let me know if I can clarify anything further!

This should be ok to merge now.

henriqueiamarino:
Thanks for the push, @iamtakashi. Sometimes, I don't know what's going on with my fixes. I use Visual Studio Code and save everything correctly, but when you later check, it's not there.
This time, I give you my word: I followed your steps, removed old pagination, and added new to the archive, search, and index templates. Anyway, it's now done.

iamtakashi:
You're welcome, @henriqueiamarino.

I wondered what might have happened to the calls to the pagination pattern on the templates. It looks like a call to a pattern adds the markup of the pattern to the editor, even if nothing has changed in the pattern. For example,

Let's say this is in a template file.
```
<foo>
<!-- wp:pattern {"slug":"coachben/pagination"} /-->
</foo>
```
That becomes this in the editor.
```
<foo>
<!-- wp:query-pagination {"align":"full","layout":{"type":"flex","justifyContent":"left","flexWrap":"wrap"}} -->
<!-- wp:query-pagination-previous {"label":"Previous"} /-->
<!-- wp:query-pagination-next {"label":"Next"} /-->
<!-- /wp:query-pagination -->
</foo>
```
So, if you saved the theme with CBT after you added the call to the pattern to the file, the call will be gone. I think this could be what happened. It's annoying. I'll look into whether this is an expected behaviour or not. I don't think this has always been like it. The call to a template part stays in the editor all the time, and that makes sense to me. Anyway, the theme is good to submit to the dotorg now.

PR #8376, opened 2024-11-06
Fix cover hover reveal style on touch devices

renatho:

When using the hover reveal style in the Cover block, and the `hover` is not available in the browser, like in touch devices, it reveals the content when it's visible in the viewport while scrolling.

#### Known issues:

Effect in the layer is not working: https://github.com/Automattic/themes/issues/8377

#### Test Instructions:

1. Add multiple cover blocks with content to a page.
2. Select the style "Hover Reveal" for the Cover blocks (under styles in the editor sidebar).
3. Navigate through it in a normal browser, and make sure it continues working as previously (revealing the content on hover).
4. Navigate through it in a touch browser (works on mobile simulator in dev tools), and make sure the content is revealed as you scroll through it.

See a video of the result in the editor and in the frontend using the 2 types of scenarios (touch and mouse).

https://github.com/user-attachments/assets/8908c0a5-93fe-4684-8e8d-e456202069b2

#### Related issue(s):
https://github.com/Automattic/wp-calypso/issues/96070

### DIFF
--- a/dalston/block-extends/extend-cover-block.css
+++ b/dalston/block-extends/extend-cover-block.css
@@ -9,6 +9,10 @@
 	opacity: 0;
 }
 
+#editor .wp-block-cover-image.is-style-hover-reveal.dalston-in-viewport.has-background-dim.has-background-dim-10 .wp-block-cover__gradient-background,
+#editor .wp-block-cover-image.is-style-hover-reveal.dalston-in-viewport.has-background-dim.has-background-dim-10:not(.has-background-gradient):before,
+#editor .wp-block-cover.is-style-hover-reveal.dalston-in-viewport.has-background-dim.has-background-dim-10 .wp-block-cover__gradient-background,
+#editor .wp-block-cover.is-style-hover-reveal.dalston-in-viewport.has-background-dim.has-background-dim-10:not(.has-background-gradient):before,
 #editor .wp-block-cover-image.is-style-hover-reveal.has-background-dim.has-background-dim-10 .wp-block-cover__gradient-background:hover,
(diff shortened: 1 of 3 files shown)


dalston/functions.php
```diff
@@ -206,7 +206,7 @@ function dalston_editor_styles() {
 
 	// Hide duplicate palette colors
 	$colors_array = get_theme_mod( 'colors_manager', array( 'colors' => true ) ); // color annotations array()
-	if ( ! empty( $colors_array ) && $colors_array['colors']['txt'] != '#1E1E1E' ) { // $config-global--color-foreground-light-default;
+	if ( ! empty( $colors_array ) && ! empty( $colors_array['colors'] ) && ! empty( $colors_array['colors']['txt'] ) && $colors_array['colors']['txt'] != '#1E1E1E' ) { // $config-global--color-foreground-light-default;
```
renatho, line 209: It's not directly related to this feature. It just fixes a warning that I noticed in the editor.

dalston/block-extends/extend-cover-block.css
```diff
@@ -122,58 +167,82 @@
 	transition: opacity 0.2s ease-in-out;
 }
 
+.wp-block-cover.is-style-hover-reveal.dalston-in-viewport.has-background-dim.has-background-dim-0:before,
```
renatho, line 170: As described in the issue, the `dim` styles are not working and I created an issue to check this.

Anyway, I added `.dalston-in-viewport` to all the `:hover` styles to keep the styles consistent. And when that issue is fixed, it should work.

phcp (approved):
Nice work! Tested on mobile and desktop browsers, and it works as expected in both cases.

https://github.com/user-attachments/assets/c34d73c4-b8aa-4a71-9cf2-30be24a66cb7

PR #8416, opened 2024-11-14
Assembler: Move the style variation specific font families from the main theme.json to the variation json file

### LINKED ISSUE #8079: Assembler: Move the style variation specific font families from the main theme.json to the variation json file.
## What?
Move the style variation specific font families from the main theme.json to the variation JSON file.

## Why?
Currently, the font families must be defined in the main theme.json file to load as expected in the editor when the style variation is switched. If this issue https://github.com/WordPress/gutenberg/issues/59965 is fixed (in progress https://github.com/WordPress/gutenberg/pull/65019), we could avoid this 'hack' that's producing the printing of not needed CSS in the frontend. So, this change would be a tiny performance improvement, and it could lead to simpler-to-define typography variations.

## Related
The same issue in TT5: https://github.com/WordPress/twentytwentyfive/issues/233

alaczek:
Closes #8079


richtabor:
I pulled these changes into https://github.com/Automattic/themes/pull/8421 to go with a number of typography improvements, to prevent conflicts. Thanks!

PR #8418, opened 2024-11-14
Fix: Contact form block dropdown arrow wrapping under field in Assembler theme 

### LINKED ISSUE #8035: Contact form block dropdown arrow wrapping under field in Assembler theme
### Quick summary

The dropdown arrow is under the dropdown field. It should be inline on the right side of the field.

contact-form-dropdown

### Steps to reproduce

1. Install and activate Assembler theme
2. Add a contact form to any page.
3. Add a dropdown field to the form
4. View the arrow under the form.

The arrow should be displayed on the right side of the field, inline with the text.

### What actually happened

The arrow is displayed under the dropdown field.

### Impact

Some (< 50%)

### Available workarounds?

Yes, easy to implement

Using this CSS as a workaround:

```
.contact-form .contact-form__select-wrapper::after {
  right: 15px;
  top: 50%;
}
```

Simple, Atomic

### Logs or notes

I have tested the dropdown field on other themes like Twenty Twenty-Four, Annalee, and Outland. The issue does not occur on those themes. So it seems to only be an issue when using the Assembler theme.

MaggieCabrera:

It looks like grunion (I thought that was deprecated, but some of the code lives in jetpack apparently) is doing some weird class swapping around, and if we modify the padding of the input element it preaks the arrow for the select. It's really hard to debug because the plugin is doing something with class swapping and rewriting the CSS, so it's not as simple as overriding the class. It was a matter of trial and error to actually verify that the selector apparently causing this problem is in fact `input:not([type="submit"]):not([type="checkbox"])` when the element we are targeting is a select!!!

This PR removes the padding altogether from the general input element to avoid this.

Before | After
--- | ---
Screenshot 2024-11-14 at 11 34 22 | Screenshot 2024-11-14 at 11 34 28

#### Related issue(s):

Closes https://github.com/Automattic/themes/issues/8035

### DIFF
--- a/assembler/style.css
+++ b/assembler/style.css
@@ -72,8 +72,15 @@ input:not([type=submit]):not([type=checkbox]),
     font-weight: inherit;
     line-height: 1.5;
     min-height: 42px;
-    padding: 0.8rem 1rem;
     width: 100%;
+}
+
+textarea,
+.wp-block-post-comments-form textarea,
+.wp-block-post-comments-form input:not([type=submit]):not([type=checkbox]),
+.jetpack-contact-form .jetpack-field .jetpack-field__input,
+.jetpack-contact-form .jetpack-field .jetpack-field__textarea {
+    padding: 0.8rem 1rem;  
 }
 
 textarea:focus,


MaggieCabrera:
it looks like without targetting the general inputs we miss the padding on all of them. I'm trying to untangle this in Jetpack, but it might be that this is the quick solution if that gets complicated to solve in there.

richtabor:
Closing in support of https://github.com/Automattic/jetpack/pull/40206

PR #8420, opened 2024-11-15
Improve spacing within post content with content and patterns

richtabor:
Improves the spacing between content and header/footer areas, as well as fullwidth patterns within content areas. This way patterns can both be fullwidth, right against header and footer template parts, AND paragraph content (standard post and pages for example) can have proper spacing apart from the header and footer. 

### Before
If post content contains a paragraph last, rests against the footer: 

CleanShot 2024-11-15 at 16 43 32

### After 
Paragraphs always have proper spacing:  

LOCATION

### Before
Fullwidth patterns with zero-ed out margin rest right against paragraph content:

SERVICES

### After
Better spacing: 

SERVICES


alaczek (approved):
Looks good to me!

PR #8421, opened 2024-11-15
Improve Assembler typography styles

### LINKED ISSUE #8079: Assembler: Move the style variation specific font families from the main theme.json to the variation json file.

richtabor:
A few improvements I discovered as I research what folks are doing with the Assembler theme.  

- Closes https://github.com/Automattic/themes/issues/8079
- Adjust size of default H4 to add more visual hierarchy.
- Disable background color picker on heading blocks.
- Reduce size of `xx-large` just a little.
- Remove "Rufina" variation; it's not strong and too similar to others.
- Add tweaks for site tagline in typography styles.
- Encompasses https://github.com/Automattic/themes/pull/8416 into this PR to prevent conflicts.

### Site tagline
https://github.com/user-attachments/assets/71793297-5fe0-456f-88a9-03ef371f8435

### H4 Before 
CleanShot 2024-11-15 at 15 54 33

### H4 After
CleanShot 2024-11-15 at 15 54 01

+ props to @alaczek.


alaczek:
This is looking good to me, but we should also address these two things:

* in the 05-fahkwang.json the headings are supposed to use 200 font weight, but we only register 400 and 600.
* in 01-figtree.json the headings are supposed to use 900 font weight, but we only register 400-700.

We either need to add the missing font weights or update the headings definition.

There are also unused fonts - Libre Baskerville, and Rufina.

richtabor:
> We either need to add the missing font weights or update the headings definition.

> There are also unused fonts - Libre Baskerville, and Rufina.

Addressed, thanks!

alaczek (approved):
LGTM!

Issue #7942, opened 2024-07-09
Task: Delist old versions of updated themes


alaczek:
This task a reminder to delist old versions, and keep track of, themes that received significant updates recently and are being released under new names:

```[tasklist]
### Tasks
- [x] Delist Atrium → Musee
- [x] Delist Curriculum → Launch LeanCV
- [x] Delist Stage → Launch OutNow
- [x] Delist Vetro → Launch Vitrum
- [x] Delist UpsideDown → Launch Inversum
- [x] Delist Cookbook → Launch Recipebook
- [x] Delist Expo → Musee
- [x] Delist The Menu → MyMenu
- [x] Delist Grammer One → Grammer
- [x] Delist Overlaid → Layover
- [x] Delist Blogorama → Brightblog
- [x] Delist DOS → WP-DOS
- [x] Delist Issue → Feature
- [x] Delist Nested → Matrioska
- [x] Delist Ritratto → Retrato
- [x] Delist Screenplay → Moviescript
- [x] Delist Trailblazer > Maverick
- [x] Delist Texty → Titleless
- [x] Delist Raw → Brute
- [x] Delist Programme → Massimo
- [x] Delist Eventual → Happening
- [x] Delist Pieria → Pierian
- [x] Delist Messagerie → Message
- [x] Delist BoxedBio → JazzyGrid
- [x] Delist Organizer → Promoter
```

henriqueiamarino:
@alaczek, we also have themes that must be removed from our repo. They were all committed and submitted to Dotorg under new names.

- Atrium
- Expo
- Trailblazer

dsas:
> @alaczek, we also have themes that must be removed from our repo. They were all committed and submitted to Dotorg under new names.
> 
>     * Atrium
> 
>     * Expo
> 
>     * Trailblazer

https://github.com/Automattic/themes/pull/8069

dsas:
We should double check this delisting list and also see if we need to close or move any issues opened under the old names (they should have a relevant label).

iamtakashi:
LeanCV was launched on dotcom, so feel free to delist the Curriculum when it's ready for dotcom.

iamtakashi:
Vitrum, Inversum, and Outnow were launched on dotcom. Please delist the old versions (Vetro, Stage, and Upsidedown) whenever convenient.

iamtakashi:
MyMenu has launched.

iamtakashi:
Layover has launched.

[...]

PR #8424, opened 2024-11-18
Urbem: add theme

henriqueiamarino:
Urbem is a spin-off of Vitrum, a theme with wide-width layouts, generous space for images, and bold type sizes.

[Demo site](https://urbem.mystagingwebsite.com/)

screenshot


henriqueiamarino:
@iamtakashi, I renamed **Downtown** and submitted it again, as the theme name has already been taken. I can't explain that happening again. I am pretty sure I checked it in Belgrad, right next to you, and there were no theme names like it. Anyway, they do exist, and I will have to rename them; I am sorry.

I already added Downtown to [this list](https://github.com/Automattic/themes/issues/7942) to be delisted in the future. Let me know if I can do more to fix that issue.

iamtakashi:
Sorry that happened to you again :( But don't worry, Downtown has yet to launch on Dotcom, so it doesn't have to be on the list, and once I review this quickly, we can launch it as Urbem on both Dotorg and Dotcom.

iamtakashi:
Let's use a capitalised name in the `style.css` and the `readme.txt` So the name will appear nicely here.
CleanShot 2024-11-18 at 12 06 50@2x
CleanShot 2024-11-18 at 12 02 58@2x
CleanShot 2024-11-18 at 12 03 06@2x

For the theme description in the `style.css`, I think you meant `Vitrum` rather than `Vetro`.

Other than those, it looks good to ship!

PR #8439, opened 2024-11-19
Streamer: add theme

henriqueiamarino:
Elevate your esports squad's online presence and connect with fans like never before using this theme design.

[Demo site](https://streamer.mystagingwebsite.com/) | [Site content](https://github.com/user-attachments/files/17817588/streamer.WordPress.2024-11-19.xml.zip)

screenshot


iamtakashi:
@henriqueiamarino Here are my notes.

* There is nothing wrong with using the `vh` unit for a font size, but I wonder if you can find an optimal font size using the `vw` unit for the `x-large` size. I don't know the exact reason, but the templates look broken in the preview. I assume the preview has a large viewport height for some reason. Regardless of the issue, I tend to find using the `vw` unit for font size makes more sense for a variety of viewports.
CleanShot 2024-11-19 at 16 17 32@2x

* I don't think the Blog Home template `templates/home.html` and `patterns/home.php` will be used when a theme has a `front-page.html`. Let's remove both.
* Using the overlay menu with an icon could help with a situation like this. 
CleanShot 2024-11-19 at 16 37 10@2x
* I think the side paddings for the content has been set wrongly on all the templates other than the front-page template. 
CleanShot 2024-11-19 at 17 49 05@2x
* On the Page with Sidebar template, it would make sense to add a sidebar for the name's sake.
* There are a few uses of numeric font sizes in the `parts/comments.html` and the Jetpack contact form in the `patterns/footer.php`. Let's use presets wherever we can.
* `patterns/single.php` is unused. Let's remove it.
* Do we need `/assets/fonts/lc-mogi/lc-mogi_normal_vf.ttf`? If the `.woff2` version is sufficient, let's remove the `.ttf` version.
* The font `TMT Limkin Pixel` is registered in the `theme.json`, but it's not used, and there are no font files bundled. Let's remove them from the JSON.
* The theme can have the following theme tags. `wide-blocks`, `block-styles`, `featured-images`, `full-site-editing`, `right-sidebar`, `sticky-post`, `style-variations`.

henriqueiamarino:
Thanks, @iamtakashi, that was a fast one 😀

I reviewed and fixed the following:
- Made all the font size presets in REMs instead of VH or VW;
- Fixed the side paddings on the templates;
- Deleted TMT Limkin Pixel from theme.json;
- Deleted unused patterns/single.php;
- Made the custom template named 'Pages Full-Width' and fixed the 'Pages' template;
- Deleted unused font LC Mogi;
- Added necessary tags to the style.css;
- Replaced all the numeric font sizes I found at comments.html and footer.html.

I want to discuss the Menu and the Blog Home issues tomorrow.

iamtakashi:
@henriqueiamarino Thanks for the update. I [removed](https://github.com/Automattic/themes/pull/8439/commits/4c38220baeba2796d5c9e3ac7d65ef27a46d2978) the `right-sidebar` tag since you decided not to have a sidebar.

It looks good to me now. Ship it! 🚢

PR #8453, opened 2024-11-25
Streamer: fixes Plunge style variation

henriqueiamarino:
This PR fixes the units used on the H1s of a style variation. The (Virtual Height) `VH` values were causing the theme preview to break, but they are now fixed.


iamtakashi:
Looks good to me 🚢 !

Issue #8378, opened 2024-11-07
Spiel theme: Changing link color on first Query Loop block included in Front Page template doesn't hold


tanjoymor:
### Quick summary

Specifically on the first Query Loop block that is included in the Front Page template of the Spiel theme, changing the Link color doesn't hold after you refresh, for either the Title block or the Categories block. This seems to be a very specific issue to the design and this one particular instance of the Query Loop block.

### Steps to reproduce

1. On any Dotcom site (Simple or Atomic) activate the Spiel theme
2. Open the Site Editor to the Front Page template
3. Select the Title block in the **first** Query Loop (the larger one on the left)
4. Make sure that the "Make title a link" toggle is on
5. Change the Link Default Color 
6. Save
7. Refresh
8. Repeat for the Categories block (same thing happens)
9. The Excerpt link works as expected

### A clear and concise description of what you expected to happen.

I expected the color I switched to to remain as the active applied color.

### What actually happened

The color reverts back to the default color.

https://github.com/user-attachments/assets/82cafaed-0476-471f-b590-41631c9db2f2

### Impact

Some (< 50%)

### Available workarounds?

Yes, difficult to implement

### If the above answer is "Yes...", outline the workaround.

I have not tried creating a custom CSS class to see if CSS could override this.

Delete the included Query Loop and add a new one, the new instance works as expected (but is no easy feat for many users to reconfigure properly).

### Platform (Simple and/or Atomic)

Simple, Atomic

alaczek:
I can replicate. The site title and categories [link color is set via theme.json](https://github.com/Automattic/themes/blob/trunk/spiel/theme.json#L869), and weirdly, cannot be reset via the editor :| It also cannot be overriden on the block level.

![Image](https://github.com/user-attachments/assets/ca1665d4-c67d-4878-ba22-5e8f87ff01f1)

I fond a Gutenberg issue that seems to cover this case: https://github.com/WordPress/gutenberg/issues/66148

alaczek:
Ok, I think I might have been wrong there - I tested this with TT4 by adding the styles for Title block to the theme.json file, but I was still able to override them on the block level.

I believe the issue stems from the fact that there are block level styles applied to the title and terms block. Removing them doesn't impact the default appearance, but allows changing the colors. I will have a fix for this shortly.

iamtakashi:
> I believe the issue stems from the fact that there are block level styles applied to the title and terms block.

[...]

PR #8455, opened 2024-11-25
Spiel: Removed local block level colours from sticky post in front page

iamtakashi:
Removed the local block-level colours from the sticky post in the front page template.


iamtakashi:
Fixes #8378.

Issue #8467, opened 2024-11-28
Rockfield theme - Proceed to checkout button doesn't have any styles, shows up as a link


vinnykaur:
### Quick summary

When Rockfield theme is active, `Proceed to Checkout` shows up as a link. It doesn't have any button styles like the rest of the buttons on the site. 

### Steps to reproduce

1. Activate Rockfield theme on an eCommerce site.
2. Add a product and view cart page.
3. Notice the Proceed to checkout button. It appears as a link rather than a button.

![Image](https://github.com/user-attachments/assets/d2538d5e-75d5-414f-ab70-54b72b9d779a)

### A clear and concise description of what you expected to happen.

The Proceed to Checkout should be rendered as a button similar to the rest of the buttons to the site. 

### What actually happened

The Proceed to Checkout displayed as a link to a text. 

### Impact

One

### Available workarounds?

No but the platform is still usable

### If the above answer is "Yes...", outline the workaround.

I shared the below CSS with the user to make the link look like a button:

```
/* Customize cart link to add a background and hover zd-9066244-ck*/
.wc-block-cart__submit-button {
    color: #fff ;
    background: #8b4513;
}
.wc-block-cart__submit-button.contained:hover {
    opacity: .9;
    color: #fff;
}
```

### Platform (Simple and/or Atomic)

Atomic

vinnykaur:
User reported here: 9066244-zd-a8c

Robertght:
I'm going to mark this as Triaged on the One Board, as I can see we have someone assigned to it.

@ouikhuan feel free to do the same in the future with new issues. Thanks!

ouikhuan:
Fixed in #8469

PR #8469, opened 2024-11-28
Rockfield: Fix cart button styles

ouikhuan:

This PR fixes the button styles in the Cart page for Proceed to checkout button and also remove the outline from the remove item link.

Before:

Screenshot 2024-11-28 at 6 21 47 PM

After:

Screenshot 2024-12-06 at 2 33 23 PM

#### Related issue(s):
#8467

### DIFF
--- a/rockfield/style-rtl.css
+++ b/rockfield/style-rtl.css
@@ -2850,7 +2850,7 @@ body:not(.fse-enabled) .site-description {
 	z-index: 1;
 }
 
-.main-navigation > div > ul li:hover, .main-navigation > div > ul li[focus-within] {
+.main-navigation > div > ul li:hover, .main-navigation.js-focus-within > div > ul li[focus-within], .js-focus-within .main-navigation > div > ul li[focus-within] {
 	cursor: pointer;
 	z-index: 99999;
 }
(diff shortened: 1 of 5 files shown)


ouikhuan:
Hi @Automattic/theam 👋 I'm working on this issue from GroundsKeeping rotation so first timer here! The PR is a CSS fix, but from the theme check above it seems to trigger quite a few error that's not related to the proposed fix, what would you suggest for us to move this forward?

alaczek:
Could we style this button to match the theme?

image

I had a quick look at the failing checks, and I don't think we have to worry about this - I think they are looking for theme.json which this theme doesn't have as it's older.

ouikhuan:
> I had a quick look at the failing checks, and I don't think we have to worry about this - I think they are looking for theme.json which this theme doesn't have as it's older.

Thanks for having a look @alaczek !

> Could we style this button to match the theme?

Sure, I've updated the CSS so it's matching the theme style now and ready for re-review, could you have another look once you have some time? Thank you!

Screenshot 2024-12-06 at 2 33 23 PM

alaczek (approved):
This is looking good to me now. LGTM!

rockfield/style-rtl.css
```diff
@@ -2850,7 +2850,7 @@ body:not(.fse-enabled) .site-description {
 	z-index: 1;
 }
 
-.main-navigation > div > ul li:hover, .main-navigation > div > ul li[focus-within] {
+.main-navigation > div > ul li:hover, .main-navigation.js-focus-within > div > ul li[focus-within], .js-focus-within .main-navigation > div > ul li[focus-within] {
```
scruffian, line 2853: This PR changed some selectors accidentally I believe (see https://github.com/Automattic/themes/pull/8570)

PR #8480, opened 2024-12-04
Dadaist: add theme

henriqueiamarino:
Dadaist is a theme designed to celebrate Dadaism and its great artists. A theme design full of diverse typefaces and background images to wrap a unique blog design.

[Demo site](https://dadaist.mystagingwebsite.com/) | [Export content](https://github.com/user-attachments/files/18008877/dada.WordPress.2024-12-04.xml.zip)

screenshot


iamtakashi:
@henriqueiamarino I love the aesthetic. It's very inspiring to see the techniques that achieve all the effects. Kudos!

* Let's change the URL for this link to `#` since we won't know the exact URL. 
CleanShot 2024-12-04 at 18 14 30@2x
* This could be a part of the charm, but is the first sentence supposed to use a title case so it's meant to be "Minute"? 
CleanShot 2024-12-04 at 18 26 34@2x
* Let wide and full-width blocks stretch on the page and the single posts templates. The one-column layout expects they can stretch. If you need help with this, ping me in Slack.
* The featured image in the page template is narrower than in the single posts template. It's probably intentional, though.
* The font sizes, medium and large, are probably meant to be `1.25rem` and `2rem` since the labels for other sizes seem to be the same as the max sizes. 
CleanShot 2024-12-04 at 23 16 19@2x
* There are two `main` in `patterns/404.php`.
* Let's remove unused patterns `patterns/comments.php`, `patterns/hidden-no-results-content.php`.
* I assume you've created the images used in the theme. Let's add the credit to `readme.txt`.
* Let's add credits for the fonts to `readme.txt`.
* We can see a thumbnail used as a featured image. Include a credit for that, too (I know...)
* Let's make the theme description the same on both files, `readme.txt` and `style.css`.
* The theme tags, `two-column` and `three-column` need to be `two-columns` and `three-columns`.
* Can we use the preset sizes for the input and the label in the Jetpack contact form?

henriqueiamarino:
Thanks, @iamtakashi. I have two pieces of information worth mentioning. I am curious if they are relevant enough to become PRs, but I'll let you decide.

- Once we clone a theme, we lose image and font credits. CBT should keep this or display the option to do it.
- When we change the <max> and <min> values of font presets over the Editor, the "size" info on the theme.json becomes obsolete. It should be updated there, too.
- Finally, CBT does not record the fixes on the Jetpack block. I tried saving edits on the demo site and returning them to the local version, but it didn't work. We need to fix it manually in the PHP file.

henriqueiamarino:
@iamtakashi -san, here's what I changed on the latest commit:

- Fixed 'All posts' link to #;
- Fixed the subscribe copy to be title case;
- The feature images are intentionally different on single and page templates;
- Fixed the font sizes and left a note about it in my previous comment;
- Removed one <main> tag from 404;
- Removed unused patterns;
- Added image and font credits to readme.txt and made the description unique;
- Fixed the theme tags on style.css;
- Added preset sizes to the Jetpack contact form and left a note about it, too.

iamtakashi:
@henriqueiamarino Thanks for the update. It looks good to me. Ship it 🚢 !

iamtakashi:
> I have two pieces of information worth mentioning.
* Thanks for sharing the info, @henriqueiamarino. For the first one, it looks like we can add/edit it whenever we need. 
CleanShot 2024-12-06 at 13 33 25@2x
* I can confirm that the font size preset doesn't update. I'll open an issue if it's not in the Gutenberg repo.
* I can confirm the Jetpack Form block doesn't save the presets. I'll open an issue if it's not in the Jetpack repo.

PR #8489, opened 2024-12-04
Fix Assembler outline variation size and style

richtabor:
Making the outline button variation the same size as the default. Also removes the brightness effect on the button; was a bit finicky. 

### Visual 

| Before  | After |
|CleanShot 2024-12-04 at 18 13 11|CleanShot 2024-12-04 at 18 12 03|


MaggieCabrera (approved):
LGTM

PR #8490, opened 2024-12-04
Remove forced post title font size for Assembler

richtabor:

If a post title does not have a post title, it renders quite large. This is fine for single post templates etc, but not for patterns. This removes the forced post title font size. 

#### Before 
CleanShot 2024-12-04 at 18 17 01

#### After 
CleanShot 2024-12-04 at 18 17 36


MaggieCabrera (approved):
Makes sense to me! the theme.json declaration seemed redundant in this case

PR #8491, opened 2024-12-05
Add blank template to Assembler

richtabor:
Adds a standard blank template to Assembler, which is necessary to support one-page patterns.


alaczek (approved):
Looks good! LGTM

PR #8510, opened 2024-12-06
Varia: handle `$colors_array` entries not being set

alaczek:
Handle `$colors_array` entries not being set.


josephscott (approved):
Thank you for upstreaming this 👍

PR #8512, opened 2024-12-07
Assembler: Add blueprint for playground

richtabor:
Allows for easier testing of section/pattern styles.


richtabor:
@alaczek should this be omitted from the build when shipping the theme?

PR #8513, opened 2024-12-09
Psychedeli: add theme

henriqueiamarino:
PsycheDeli is the ultimate WordPress theme for bloggers and artists seeking to showcase their vibrant journeys through the captivating visuals of the '60s. Designed with passion and creativity, PsycheDeli is your gateway to effortlessly share your psychedelic adventures and thought-provoking creations with the world. Get ready to ignite imaginations and captivate audiences like never before with PsycheDeli.

[Demo site](https://psychedelic.mystagingwebsite.com/) | [Content Export](https://github.com/user-attachments/files/18127265/psychedeli.WordPress.2024-12-13.xml.zip)

screenshot


iamtakashi:
@henriqueiamarino here are my notes. Let me know if I can clarify anything further.

* For all the featured images inside a cover block for the circle effect, it'd be better to set the width with `%` (leave the height blank) rather than an absolute `px` value to avoid looking like this on small screens.
CleanShot 2024-12-10 at 18 55 08@2x 
CleanShot 2024-12-10 at 19 25 27@2x
* It'd be better to set the border radius with `%` rather than an absolute value to avoid looking like this. `50%` would make it a circle regardless of the size of the featured image. 
CleanShot 2024-12-10 at 19 00 16@2x
* There is a left padding accidentally in the first cover block in the blog home template, so the feature image inside isn't the centre.
CleanShot 2024-12-10 at 18 48 54@2x
* Same as the last few themes, let's use the Jetpack Subscribe block for the subscription. 
CleanShot 2024-12-10 at 19 11 10@2x
* I wonder if the posts in the article section on the home page can look similar or the same as the ones in the archives template because those look nicer to me.
CleanShot 2024-12-10 at 19 14 46@2x 
CleanShot 2024-12-10 at 19 18 42@2x
* It's subjective, but I didn't expect the navigation has a white background. It could have the yellow background.
CleanShot 2024-12-10 at 19 32 25@2x
* I know these are meant to be changed, but I'm unsure if Github and Linkedin are good defaults for the hypothetical profile. Tumblr and something else might be better.
* The pullquote block seems a little too large compared to the body text. Could it be the medium size?
CleanShot 2024-12-10 at 19 59 33@2x
* In all the templates except Blog Home template, the gradient used in the "content" group doesn't change with the style variations. The group block doesn't have the gradient slug. I think you can fix it by reapplying the gradient in the editor and saving the templates again.
CleanShot 2024-12-10 at 23 03 49@2x
* Since the image path of the psychedelic flower background needs to be localised, the `header-home` and the `header` template parts need to be patterns. I can see they exist as patterns, but each of the template part file doesn't have the call to it. Like the `parts/footer.html`, they should have a call to their patterns. For example, the `parts/header-home.html` should only have `<!-- wp:pattern {"slug":"psychedeli/header-home"} /-->`.
* `main` tag is missing in the Blog Home template. It makes sense if the "static content" group has it, but we probably should rename it just "content" since it's not static.
* Let's change the version of the minimum PHP version to 7.0. in `readme.txt`.
* There are lots of images are listed in `readme.txt` Let's only list the images bundled and seen in `screenshot.png`. But we need to change the three images from Pexels in `screenshot.png` because they aren't CC0 ([more info](https://make.wordpress.org/themes/handbook/review/resources/#recommended-websites-for-images)). I find the most images in Pexel aren't compatible with GPL.
* Let's change the version of the minimum PHP version to 7.0. in `style.css`
* There are a number of numeric font sizes in `theme.json`. (`site-title`, `h1`, `h2`, and `h3`.) Can all be presets to provide a better customising experience?
* Use only [allowed theme tags](https://make.wordpress.org/themes/handbook/review/required/theme-tags/). They should all be lower case and have no spaces. (e.g. `Two Columns` > `two-colums`) I don't think the following tags are appropriate for this theme. `Entertainment`, `Portfolio`, `Three Columns`, `Editor Style`, `flexible-header`, `Site Editor`, `Full Width Template`, `Template Editing`, `Theme Options`.
* Instead the should have the following tags `block-styles`and `full-site-editing`.
* When a theme is ready to export, it's good idea to run the theme check with [the plugin](https://en-gb.wordpress.org/plugins/theme-check/). It tells you if there is a problem with the theme tags.

henriqueiamarino:
Thanks, @iamtakashi. I knew this would not be easy to fix, so I left it as the last one.
I followed and fixed all the issues you mentioned. But as you can see from the working theme on the demo site, we still have problems showing the background image on <header> and <header-home>.

- Set cover widths and corners to % for all Featured images inside all cover blocks;
- Removing left padding from the first cover block on the Home template;
- Replaced Jetpack block for <Subscriptions> as indicated;
- Made all QL posts the same as we have for the Archive, Search, and Index templates;
- Added Base color as the background color of the Navigation for <header> and <header-home>;
- Replaced social links as mentioned;
- Fixed Pullquote block size;
- Fixed the Content block for all templates, using the one from Blog Home, which was working well;
- Followed instructions to make <header> and <header-home> patterns by fixing the HTML files on parts — not sure if this went well as the background image is not showing on the demo site;
- Checked all <main> tags again;
- Minimum or Required PHP version set to 7.0 on readme.text and style.css;
- Replaced all(?) the numeric font sizes mentioned;
- Fixed tags following indications.

Before wrapping up this review, I'll wait for your thoughts on the background issue.

iamtakashi:
I've just pushed changes to the header and the header home template parts. After pulling my changes, can you test it?

I'll review the rest of your updates from now.

iamtakashi:
@henriqueiamarino, the only things that need to change again are;
* Swapping the Jetpack Form block with the Subscription block.
* The theme tags. There are some tags with spaces and capital letters. I think these are good to have for this theme `blog, one-column, two-columns, block-styles, featured-images, full-site-editing, rtl-language-support, sticky-post, style-variations, threaded-comments, translation-ready, wide-blocks`

henriqueiamarino:
Thanks again, @iamtakashi. I will work on the demo site and the showcase page before submitting it tomorrow. As soon as I finish, I'll update the content export here and let you know.

iamtakashi:
Ok, I'll ship it! 🚢

PR #8516, opened 2024-12-11
Independent Publister 2: Add spacing between button blocks

### LINKED ISSUE #8511: Independent Publisher 2: Buttons block are overlapping
### Quick summary

The Buttons Block are overlapping when using the theme Independent Publisher 2. 

### Steps to reproduce

1. On a Simple or Atomic site, activate Independent Publisher 2.
2. Go to Page or Post editor and add a Buttons block with two buttons.
3. Observe that buttons are overlapping.

The buttons should not be overlapping each other. 

### What actually happened

The buttons were overlapping. It was not possible to adjust the spacing between the buttons. 

Screenshots: 

Image

Image

### Impact

All

### Available workarounds?

Yes, easy to implement

You can apply the following CSS code: 

.wp-block-button .wp-block-button__link {
box-sizing: border-box;
}

However, sites on Free plan won't be able to use this workaround. 

Atomic, Simple

### Logs or notes

User report: https://wordpress.com/forums/topic/buttons-block-overlaps-the-buttons-with-each-other

sixhours:

* Add a left margin to button blocks that directly follow other button blocks to prevent overlap. 
* Also adds the corresponding RTL styles.

**Before**

Screenshot 2024-12-11 at 1 14 52 PM

**After**

Screenshot 2024-12-11 at 3 23 21 PM

#### Related issue(s):

Fixes https://github.com/Automattic/themes/issues/8511

### DIFF
--- a/independent-publisher-2/css/blocks.css
+++ b/independent-publisher-2/css/blocks.css
@@ -195,7 +195,7 @@ p.has-drop-cap:not(:focus)::first-letter {
 	background: #0087be;
 	border: solid 1px transparent;
 	border-radius: 3px;
-	box-sizing: content-box;
+	box-sizing: border-box;
 	color: #fff;
 	cursor: pointer;
 	-webkit-transition: background 120ms ease-in-out, box-shadow 120ms ease-in-out;
@@ -204,7 +204,7 @@ p.has-drop-cap:not(:focus)::first-letter {
 	font-size: 16px;
 	font-weight: 400;
 	font-style: normal;
-	padding: .4375em .875em;
+	padding: .4375em 1.25em;
 	text-decoration: none;
 	-webkit-font-smoothing: antialiased;
 	-moz-osx-font-smoothing: grayscale;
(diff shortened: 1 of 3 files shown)


sixhours:
There's something unique to WordPress.com that's causing this behavior, because when I apply this patch to a different test site, I get extra spacing between the buttons:

Screenshot 2024-12-11 at 2 01 44 PM

The markup hasn't changed. I'll keep digging and see what the difference is.

sixhours:
Not sure what the difference is, but updating button styles to `border-box` and updating the padding solves this problem on WP.com without making the spacing wonky on self-hosted installations.

Should I worry about the Theme Check results? Do we usually make updates to older themes to comply or is it safe to let those slide?

alaczek:
> Do we usually make updates to older themes to comply or is it safe to let those slide?

Let is slide in this case, I don't think we ever pushed this theme to .org repo. Anyway, looks good to me!

alaczek:
I merged this as it solves the issue on the front end, but in the editor the buttons are still overlapping (both on self-hosted and .com sites):

Editor view:
image

I believe the same fix needs to be applied here:
https://github.com/Automattic/themes/blob/trunk/independent-publisher-2/css/editor-blocks.css#L543

Issue #8517, opened 2024-12-11
LeanCV Review


iamtakashi:
### Quick summary

Since we found out LeanCV has not yet launched on dotcom, I'm reviewing the theme.

@henriqueiamarino, here are my notes. Since the theme has already been launched in dotorg, I'll list the issues we can fix without disrupting the sites that already use this theme. Let me know your thoughts.

* `duotone-default` isn't registered, but `avatar`, `cover`, `post-featured-image`, and `site-logo` blocks were meant to be used it. Therefore, any page with any of the blocks throws those notices. Ideally, we want to add the working duotone to those blocks, but that will change how those blocks appear on the live sites :( Shall we remove the duotone from those blocks since it has yet to work since the beginning? What do you think?  ![Image](https://github.com/user-attachments/assets/f75f9557-491e-4103-8292-bd0a633ba9f6)

* I'd recommend turning on the debug mode by editing the `wp-config.php` to see things like this.  ![Image](https://github.com/user-attachments/assets/4da41a65-637d-45dc-89e8-4b78d2dd2731)
![Image](https://github.com/user-attachments/assets/dd0e2d95-2507-4573-88d3-1f120d16ebbb)

* Let's turn on the overlay menu for small screens to avoid this. ![Image](https://github.com/user-attachments/assets/18d97058-a335-40ff-840e-c861ca0ece18)

* It's a CBT fault, but there are `strong` tags inside the localised string. Let's move the tags outside of the escape function like this![Image](https://github.com/user-attachments/assets/f391d94b-6bc1-4e69-9791-0e093223a7ee)

* I know it's a Gutenberg bug, but let's edit the size label for `small`, `medium`, and `large` font size presets. ![Image](https://github.com/user-attachments/assets/cb103f22-ea8d-4c77-826c-61ba97cf75b2)

* Let's remove unused template parts, `parts/header-pages.html` and `parts/portrait.html`. But let's add all the other template parts in `theme.json` with a `title`.

* The footer template part doesn't seem right. Edit the vertical paddings in one of the group blocks, and resaving the theme should fix it.

* Let's remove unused patters, `patterns/page.php`, `patterns/single.php`, `patterns/bio.php`, `patterns/comments.php`, `patterns/hidden-404.php`, `patterns/portrait.php`, `patterns/self.php`.

* A `main` tag is missing in 404, archives, front page, blog home, index, and search templates.

* Let's update the `Tested up to:` to 6.7 in `readme.txt` and `style.css`.

[...]

Issue #7785, opened 2024-04-29
Update "Supports PHP:" value for all themes


creativecoder:
We've been adding `declare( strict_types=1 );` to theme PHP files because of WP.com deploy requirements.

This mode was added in PHP 7. Currently most (all?) themes in this repo declare "Support PHP: 5.x" in the style.css header.

We need to update this to correctly require PHP 7.0+, to indicate that themes may trigger errors if used with older PHP versions.

creativecoder:
I've restored `[Type] Bug` for this issue--fixing incorrect PHP compatibility information is not an enhancement.

inaikem:
@iamtakashi, do you have a sense of who/which team this issue would be assigned to now?

@Automattic/fusion: Same question to you + how would you rate the priority of this now our minimum requirement on Dotcom is PHP 8.x?

iamtakashi:
> We need to update this to correctly require PHP 7.0+

That's good to know. We'll make sure about the themes going forward. cc @beafialho @iamarinoh 

> do you have a sense of who/which team this issue would be assigned to now?

Thanks for the ping. The decision that I know about is that; (pNEWy-i6e-p2#comment-57267)
* The themes launched before 2023 will be maintained by dotcom teams (Marvel? T-rex?)
* The themes launched after 2023 will be maintained by dotorg teams, including the theme designers

inaikem:
@iamtakashi 👋 We currently support [PHP 8.1, 8.2, 8.3](https://wordpress.com/support/php-environment/) on WPcom. It's highly possible 8.1 will be deprecated next year.

Question: What's the best way for us to loop in your team with the most up-to-date guidance?

inaikem:
@Automattic/t-rex, based on Takashi's comment above, are you able to pick up all pre-2023 theme updates?

dsas:
> We need to update this to correctly require PHP 7.0+, to indicate that themes may trigger errors if used with older PHP versions.

Clarification: so far as I know a strict types declaration will trigger a warning, not an error. The impact is a *lot* of noisy messages, not a fatal error stopping the site from being displayed.

This affects pre-2023 themes that were updated within roughly the last 12 months that are being used on non-dotcom site using a PHP version that was EOL by PHP six years ago, and widely EOL'd in linux distros more than four years ago. Unless there's something I'm missing, it seems like a very low priority for us.

> @Automattic/t-rex, based on Takashi's comment above, are you able to pick up all pre-2023 theme updates?

[...]

PR #8518, opened 2024-12-12
LeanCV: fixes issues mentioned on #8517

henriqueiamarino:
This PR addresses the issues mentioned by @iamtakashi in review #8517.

- The only solution I found for the duotone issue will impact users, I am aware. But I do think this is better than simply removing duotones, which would impact much more the theme's concept;
- Mobile menu set;
- The strong tags were removed;
- Font size presets adjusted, and numeric font values on theme.json were all turned into presets;
- Removed unused template parts;
- Removed unused patterns;
- Fixed Footer padding;
- Checked and fixed all the <main> tags. They are all set to the Content block now;
- Updated readme.txt and style.css for the Version number, Tested Up To, and Requires PHP;
- Removed tags and set the necessary ones.


iamtakashi:
@henriqueiamarino Thanks for the update! I've pushed [a change](https://github.com/Automattic/themes/pull/8518/commits/04ae122e152cf2e82f6f1db855f992bdc31e76ad) to fix the things that seem overlooked.

* Removed `patterns/hidden-404.php`, `patterns/portrait.php`, and `patterns/self.php`.
* Added `wide-blocks` tag.
* Changed the version number to `1.0.0` inside the changelog.
* Changed the minimum PHP version to 7.0 (it was my fault) in the `readme.txt` and the `style.css`.
* Added back the theme tag `blog` (since the theme description says a blog theme).

--
Now, I noticed the duotoned elements don't change immediately in the editor. (It changes on the front of the site, though) The Professio theme is also suffering from this. You have overcome this issue in AlleyOop and Raconteur. Should we do the same in this theme? 
CleanShot 2024-12-13 at 01 41 23

henriqueiamarino:
Thanks, @iamtakashi. It's really disturbing how some changes I make vanish from the commits. I am 100% sure I noticed those. I did remove those patterns and added <wide-blocks>. That's a bummer. Anyway, yes, the default-duotone replacement works only after you activate the theme.json, for instance, after refreshing the Editor.

Several ready-to-use duotone presets will work for this one, but we should not abandon the unique default duotone technique entirely on other themes. It's helpful, and eventually, the Editor will be faster at displaying that change.

I can add more presets and fix the style variations for this one. It's restrictive for users but probably the faster solution.

iamtakashi:
Beautiful. Thanks, @henriqueiamarino.

I just pushed a change to update the minimum PHP version again. Sorry for being back and forth. It looks like 8.2 will be the minimum PHP version going forward. ([more info](https://github.com/Automattic/themes/issues/7785))

iamtakashi:
I'll merge this and bump up the version. The dotorg version should be updated automatically. I'll launch this on dotcom soon, so it'd be great if you could send me the demo content.

PR #8526, opened 2024-12-16
Psychedeli: Update headers to show a long site title properly

iamtakashi:
I realised a long site title didn't appear properly. This was due to the line height being `0`. I've changed this to `0.8` and adjusted several paddings to compensate. I also changed the alignment of the site title to the centre.

| Before | After |
| ----------- | ----------- |
| CleanShot 2024-12-16 at 17 31 34@2x | CleanShot 2024-12-16 at 17 28 05@2x |

Also, I realised that some form elements from the Jetpack contact form weren't removed correctly, so I cleaned that up.

@henriqueiamarino, let me know if this is good to merge.


henriqueiamarino:
I think this is a great solution, thanks @iamtakashi. As far as I can see, it's ready to be merged.

Issue #8252, opened 2024-10-09
Themes linking/embedding resources from development websites


dsas:
### Quick summary

Several themes are linking to the site used to develop the theme, or is embedding images & assets from that theme development site. This shouldn't be happening for two reasons:
 1. Users have to find and fix all of the links, otherwise their visitors will be taken to your development website. For many of these links they're probably changing them anyway, but they might miss a 'home' link, assuming it'll just work.
 2. If you delete these sites, the people using the fonts and images linked from them will have the appearance of their websites changed.

Currently affected themes:
 - [x] adonay
 - [x] allez
 - [x] artly
 - [x] common
 - [x] eventual
 - [x] goodskin
 - [x] grammerone
 - [ ] ~~greenseed~~ (I think it got fixed during launch)
 - [x] raw
 - [ ] ~~screenplay~~
 - [x] stage
 - [x] ueno

### Steps to reproduce

Search the code on github for 'mystagingwebsite': https://github.com/search?q=repo%3AAutomattic%2Fthemes%20mystagingwebsite&type=code

### A clear and concise description of what you expected to happen.

I expected there to be zero results

### What actually happened

It found 18 files across 

### Impact

One

### Available workarounds?

There is no user impact

Robertght:
@dsas, it looks like you have this one on your plate already.

I'm going to mark this as Triaged on our board, but let us know if you need our help checking some things in the process.

alaczek:
Adonay is on the list, but I'm not seeing resources from staging site in the code. There was an issue on the demo site specifically, which I fixed.

alaczek:
BTW, here's the search results for `localhost`:

https://github.com/search?q=repo%3AAutomattic%2Fthemes+localhost&type=code

alaczek:
I began working through these, starting with GrammerOne. Boy, is it a can of worms 😬 Essentially, it wasn't using site title, logo and tagline blocks, but static heading and images, so I ended up fixing it in every header. I would appreciate a review though.

Btw, should I create a new branch for every theme I fix, or is it ok to keep it all on one branch with seperate commits for each theme @dsas?

dsas:
> Boy, is it a can of worms 

😬 

> should I create a new branch for every theme I fix

Whichever works for you really. It's ok to keep it all on one branch. I prefer many smaller branches as it's quicker to review them one-by-one.

alaczek:
Hey @dsas 👋 I believe this is ready for your review.

[...]

PR #8528, opened 2024-12-16
Blissed: Remove static links

dsas:
See #8252

I'm removing the image background completely, because it hasn't been working from the very beginning.


dsas (commented):
✅ looks good to me

PR #8531, opened 2024-12-16
GrammerOne: Remove static links

dsas:
See #8252


dsas:
I think we should use https://github.com/Automattic/themes/pull/8538 instead

PR #8533, opened 2024-12-16
Spiel: Update latest-posts.php

### LINKED ISSUE #8378: Spiel theme: Changing link color on first Query Loop block included in Front Page template doesn't hold

dsas:
Remove the text and link color props from the blocks so that they can properly inherit colors and allow for making chnages. Fixes #8378

### DIFF
--- a/spiel/patterns/latest-posts.php
+++ b/spiel/patterns/latest-posts.php
@@ -29,10 +29,10 @@ declare( strict_types = 1 );
 
 					<!-- wp:group {"style":{"spacing":{"blockGap":"0.5rem"}},"layout":{"type":"constrained"}} -->
 					<div class="wp-block-group">
-						<!-- wp:post-terms {"term":"category","style":{"elements":{"link":{"color":[],":hover":{"color":{"text":"var:preset|color|primary"}}}},"color":[]},"textColor":"contrast"} /-->
-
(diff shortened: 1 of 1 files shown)


dsas:
whoops opened accidentally

iamtakashi:
Oops, I didn't realise this also needed to change. Thanks for looking after it.

PR #8534, opened 2024-12-16
Stage: Remove static links.

dsas:
See #8252


dsas (commented):
I can't see where this file is used, but the change itself looks fine

PR #8536, opened 2024-12-16
Artly: Remove static links.

dsas:
See #8252


dsas (commented):
✅ lgtm. Ideally it would be translated, but it also wasn't before, so it's no worse off.

PR #8538, opened 2024-12-16
Fix static links in GrammerOne

dsas:
See #8252


dsas (commented):
@alaczek This change to **Grammer-one** makes sense to me. **Grammer** has nearly the exact same change to its headers, except it still hardcodes a tagline.

PR #8539, opened 2024-12-16
Eventual: remove static links

dsas:
See #8252


eventual/patterns/index.php
dsas, line 1: Was this intententionally removed @alaczek ? As far as I can tell it's hidden and unused, but just wanted to double check.

Change looks fine to me

alaczek: apologies, I should have communicated this clearly - yes, this has been my conclusion as well and reason for removing.

Issue #7974, opened 2024-07-28
Rainfall: spacing issues on WordPress 6.6


sajmes:
### Quick summary

User reports that the 6.6 update has caused spacing issues on Rainfall. Reverting to 6.5.x resolved the issues. 

### Steps to reproduce

1. Install Rainfall on a site with 6.6 active. 

### What you expected to happen

Spacing should match the theme demo. 

### What actually happened

Spacing is compressed. 

Demo:
![6 5](https://github.com/user-attachments/assets/a69d4427-a4b2-48b7-93ac-f3ddb3e7f3d5)

Live: 
![6 6](https://github.com/user-attachments/assets/b7de16b4-3eb6-4057-af04-59a291ab0570)

### Reproducibility

Consistent

### Severity

None

### Available workarounds?

None

retnonindya:
📌 REPRODUCTION RESULTS
- Tested on Simple – Replicated
- Tested on Atomic – Replicated

📌 FINDINGS/SCREENSHOTS/VIDEO
I noticed the Block Spacing tool looks different (I can't find a proper expression. My apologies.) It seems like there are only two options for block spacing: 100 and 110.

<img width="1704" alt="Screenshot 2024-08-08 at 5 48 40 PM" src="https://github.com/user-attachments/assets/7577d363-c01f-4951-92b2-4ef05a7d3732">

📌 ACTIONS
- Triaged

iamtakashi:
This should be fixed with #8551. 

I'm closing this as I no longer repro the issue with 6.6. Feel free to reopen if that's not the case for you.

PR #8551, opened 2024-12-19
Rainfall: Add paddings and other visual theme improvements

beafialho:
Rainfall is currently looking badly because it has no paddings on the site at all:

https://github.com/user-attachments/assets/9a30454b-97d9-4335-8ab0-2baf711a5b4b

I went ahead and fixed it, along with other visual improvements I thought should be done to the theme overall.

@iamtakashi when you have a moment, can you review these changes? Also, let me know if there's anyone else I should ask for a review. Thanks in advance.


iamtakashi:
@beafialho Thanks for the PR! The changes look good, but there seem to be extra side paddings around the title in the archive and the page with black header templates. 
CleanShot 2024-12-19 at 20 21 14@2x

Also, for some reason, the height of the spacers in this change seems broken. [Example](https://github.com/Automattic/themes/pull/8551/files#diff-627d609c16637ab6d49996a02d912fed88d076e68d6c04c7062c575441382594R8) 
You can fix them if you re-save the templates with CBT after setting the height in the editor. If you go to the [files changed tab](https://github.com/Automattic/themes/pull/8551/files) and search `\u002d`, you should be able to find which spacers are broken.
CleanShot 2024-12-19 at 23 11 57@2x


And `"ref":1909` is added back in the navigation block in this change. That makes the preview broken. The CBT should automatically remove that when we save it with CBT.
CleanShot 2024-12-19 at 20 40 55@2x

Let me know if you need help with anything.

beafialho:
Thanks for reviewing @iamtakashi! I submitted a fix for the spacer issues and the extra side paddings you pointed out. 

Regarding the navigation block, I can't find the `"ref":1909` bit anywhere in the theme files, where do you see that?

But yes, CBT was supposed to handle that.. I don't understand why that's happening because I clicked in this option when I saved it: 

Captura de ecrã 2024-12-20, às 09 27 46

beafialho:
Correction: I spotted one ref in the footer part. Submitted a fix for that too!

iamtakashi:
Thanks for the update, @beafialho.

There are still broken spacers in `parts/query.html`(which is used in the archive and search templates) and `templates/single.html`. Could you check all the spacers used in the single posts template?

There are some templates with images and text that need to be localised. Can you make sure the Localize Text and Images options are checked, and the Process Only Modified Templates option is unchecked the next time you save the theme?
CleanShot 2024-12-20 at 13 28 41@2x

beafialho:
> There are still broken spacers in parts/query.html(which is used in the archive and search templates) and templates/single.html. Could you check all the spacers used in the single posts template?

Thanks for your help @iamtakashi.

I double checked the `parts/query.html` and `templates/single.html` and found the issues. I saved it, unchecked the Process Only Modified Templates option. Can you try now?

iamtakashi:
Thanks for the update, @beafialho. The change looks good.

I've also pushed some further changes. Check the commits above if you want to make sure.
* Reformatted some dodgy escaping in the single posts template
* Changed the image paths in the templates
* Removed and replaced non-defined spacing sizes

dd32:
This should resolve this issue: Fixes https://github.com/Automattic/themes/issues/7974

beafialho:
Thank you for pushing the changes. I re-checked and submitted one last update to the Single Post template (the comments section was set to content width, and it should be wide). If this change looks good, this is good to merge.

iamtakashi:
@beafialho Thanks for the update to the comments section. That made me realise the width setting was broken in the query template part and index template. I fixed them now.

iamtakashi:
@dd32 Thanks for the link. I can confirm this PR will fix https://github.com/Automattic/themes/issues/7974.

iamtakashi:
The PR is getting a lot larger than expected. In my testing with this PR, I no longer see something very wrong on the front of the site. Let's merge this and fix small things if we find them later.

PR #8561, opened 2025-01-07
LeanCV: Add section styles, clean up colors

richtabor:
Support section styles to provide more variation and better interoperability across color variations. Also cleans up duotones.

#### Visual

https://github.com/user-attachments/assets/82a1d98d-185e-4f3f-a8a7-360e2d47dfd0


alaczek:
When I applied this locally, the duotone on the image was removed, and I couldn't see the block styles. Am I missing something?

PR #8569, opened 2025-01-13
Varia: Fix nested bullets in menus

### LINKED ISSUE #7756: Shawburn: Submenu Hyphens Don't Display on iPhone
### Quick summary

On iPhone, hyphens and indentation don't appear on the submenu items. Works fine on Android though.

### Steps to reproduce

1. Activate the Shawburn theme
2. Create a menu with submenu items
3. Observe submenu items are hyphenated on desktop and Android phones
4. Observe submenu items are not hyphenated on iPhone

I expected to see hyphens and indents like on Android: 
Screenshot 2024-04-21 at 14 17 18

### What actually happened

Submenu items had no hyphens or indents on iPhone 12 and 15:
Screenshot 2024-04-21 at 15 03 24

### Browser

iOS Safari, Android Chrome

### Context

#8074605-zen

_No response_

### Other notes

_No response_

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

### Workaround details

_No response_

scruffian:
This is a fix for https://github.com/Automattic/themes/issues/7756

The previous code `content: "– " counters(nested-list, "– ", none);` isn't valid CSS. It works in some browsers but not all. 

(I found it didn't work in Firefox, which makes this a lot easier to test than iOS.)

The last parameter to the `counters` function is a `style`: https://developer.mozilla.org/en-US/docs/Web/CSS/counters. None doesn't seem to be a valid style, so intsead I have defined a style called `empty` which works the same as `none`.

## Testing instructions

1. Switch to a Varia based theme (e.g. Shawburn)
2. Add a menu, which has nested items
3. Open the front end your site in a mobile preview (small browser window should be fine)
4. Open the menu
5. Confirm that the nested items show `-`s before them like this
Screenshot 2025-01-13 at 09 53 23

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/7756

### DIFF
--- a/alves/style-editor.css
+++ b/alves/style-editor.css
@@ -1542,6 +1542,12 @@ p:not(.site-title) a:hover {
 	margin: 0 0 5px 0;
 }
 
+@counter-style empty {
+	.fse-template-part {
+		symbols: "– ";
+	}
+}
+
 .fse-template-part .main-navigation {
 	color: #394d55;
 }
@@ -1731,7 +1737,7 @@ p:not(.site-title) a:hover {
 	/* Increment the dashes */
 	counter-increment: nested-list;
 	/* Insert dashes with spaces in between */
-	content: "– " counters(nested-list, "– ", none);
+	content: "– " counters(nested-list, "– ", empty);
 }
 
 @media only screen and (min-width: 560px) {
--- a/alves/style-editor.css
+++ b/alves/style-editor.css
@@ -1542,14 +1542,14 @@ p:not(.site-title) a:hover {
 	margin: 0 0 5px 0;
 }
 
-@counter-style empty {
-	.fse-template-part {
-		symbols: "– ";
-	}
-}
-
 .fse-template-part .main-navigation {
 	color: #394d55;
+}
+
+@counter-style empty {
+	.fse-template-part .main-navigation {
+		symbols: "– ";
+	}
 }
 
 .fse-template-part .main-navigation > div {
--- a/alves/style-editor.css
+++ b/alves/style-editor.css
@@ -1546,12 +1546,6 @@ p:not(.site-title) a:hover {
 	color: #394d55;
 }
 
-@counter-style empty {
-	.fse-template-part .main-navigation {
-		symbols: "– ";
-	}
-}
-
 .fse-template-part .main-navigation > div {
 	display: none;
 }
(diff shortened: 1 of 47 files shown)


shawburn/style-editor.css
```diff
@@ -1602,7 +1608,7 @@ pre.wp-block-verse {
 	z-index: 1;
 }
 
-.fse-template-part .main-navigation > div > ul li:hover, .fse-template-part .main-navigation > div > ul li[focus-within] {
+.fse-template-part .main-navigation > div > ul li:hover, .fse-template-part.js-focus-within .main-navigation > div > ul li[focus-within], .js-focus-within .fse-template-part .main-navigation > div > ul li[focus-within] {
```
MaggieCabrera: this is coming from the sass compiler, which we must have updated at some point and haven't built the theme since. I say this doesn't hurt to include, but we could do it separately, your call!

scruffian: Won't this actually break things, because the selector is different?

MaggieCabrera: looking at it more closely, I'm not so sure anymore. Looks like something is missing according to [the docs](https://www.npmjs.com/package/postcss-focus-within) (we need the polyfill paired with the CSS change). postcss-focus-within was last updated in https://github.com/Automattic/themes/pull/7059

scruffian: Support for `focus-within` looks pretty good: https://caniuse.com/?search=focus-within. Maybe we should just remove the polyfill?

scruffian: Fix here: https://github.com/Automattic/themes/pull/8570

mikachan: Will this need updating once https://github.com/Automattic/themes/pull/8570 is merged (I'm guessing `.js-focus-within` won't be needed with that PR)?

scruffian: yes

alves/style-editor.css
```diff
@@ -1546,6 +1546,12 @@ p:not(.site-title) a:hover {
 	color: #394d55;
 }
 
+@counter-style empty {
+	.fse-template-part .main-navigation {
```
scruffian: I don't get why this is being prefixed with `.fse-template-part`.

scruffian: Ah, this is because in the editor CSS we wrap everything with the .fse-template-part selector.

scruffian:

mikachan (approved):
This fixes the issue described in https://github.com/Automattic/themes/issues/7756 for me, testing in Firefox:

| Before | After |
| ---- | ---- |
| image | image |

The code changes also LGTM 👍

PR #8575, opened 2025-01-14
Blockbase: Fix previewing post editor colors from style variations in child themes

### LINKED ISSUE #6963: [Theme] Attar - In Style variations, Theme colors (primary, secondary, tertiary) apply as Black to blocks on pages and posts
### Quick summary

When you're using one of the style variations in the Attar theme, choosing one of the default theme colors for the text doesn't work. The palette in the **Default** variation works as expected though.

### Steps to reproduce

1. Switch to the Attar theme.
2. Select a style variation other than **Default**.
3. In a new page, try to update the text color and choose one of the default theme options.

text color default

4. Next unselect the default color option and manually add the same color code. The color will update as expected now. 

text color custom

I expected the color of the text to change and match the default color that I had selected.

### What actually happened

The text color didn't change. 

### Impact

Some (< 50%)

### Available workarounds?

Yes, easy to implement

Simple, Atomic

### Logs or notes

_No response_

adamwoodnz:
Fixes https://github.com/Automattic/themes/issues/6963

This PR removes the `!important` from the fallback color rules in the Blockbase stylesheet. The actual colors are defined in theme.json and generated in global styles, and these `!important` fallbacks cause problems in child themes with style variations loaded, when previewing colors in the post editor. The issue was raised referencing the Attar theme, but it is reproducible with other Blockbase child themes, like Ames.

**Please note that these rules do not cause the same issues when the Blockbase theme itself is activated, or in the site editor, or on the frontend.**

With a child theme like Attar activated and a style variation selected, the Blockbase theme styles are inlined in the post editor document after the style variation styles, overriding them and causing the color to be set to the Blockbase theme default custom colors:

Screenshot 2025-01-14 at 7 03 15 PM

Lowering the specificity of these rules allows the correct color presets generated from the style variation to be used, as expected.

An alternative fix could be to adjust the loading order, so that the Blockbase stylesheet is inlined before the global styles. I'm not aware of this happening with other non Blockbase themes however.

### Screenshots (Theme: Attar, Style variation: Charcoal)

| | Before | After |
|-|-|-|
| Text | Screenshot 2025-01-14 at 6 46 08 PM | Screenshot 2025-01-15 at 2 50 03 PM |
| Button | Screenshot 2025-01-14 at 6 45 52 PM | Screenshot 2025-01-15 at 2 47 57 PM |

## Testing

We need to test both the Blockbase theme and child themes.

1. Activate the Attar theme
2. Under Appearance -> Editor -> Styles, select a style variation and save
3. Edit a post, changing the color of elements using colors from the color palette (should be the colors from the style variation)
4. Observe that the colors are applied in the post editor preview
5. Save and check that the post on the frontend matches the post editor preview
6. Return to the admin and use the site editor to edit a page or template, changing colors
7. Observe that the colors are applied in the site editor preview
8. Save and check that the page on the frontend matches the site editor preview
9. Activate the Blockbase theme and repeat steps 2 - 8

Further regression testing:
- Activate other child themes and repeat steps 2 - 8
- Repeat steps 3 - 8 with Blockbase or a child theme, and **no** style variation selected

### DIFF
--- a/blockbase/assets/ponyfill.css
+++ b/blockbase/assets/ponyfill.css
@@ -123,43 +123,43 @@ body > .is-root-container > .wp-block-template-part > .wp-block-cover,
 }
 
 .has-primary-background-color {
-  background-color: var(--wp--custom--color--primary) !important;
+  background-color: var(--wp--custom--color--primary);
 }
 
 .has-secondary-background-color {
-  background-color: var(--wp--custom--color--secondary) !important;
+  background-color: var(--wp--custom--color--secondary);
 }
 
 .has-foreground-background-color {
-  background-color: var(--wp--custom--color--foreground) !important;
+  background-color: var(--wp--custom--color--foreground);
 }
 
 .has-background-background-color {
-  background-color: var(--wp--custom--color--background) !important;
+  background-color: var(--wp--custom--color--background);
 }
 
 .has-tertiary-background-color {
-  background-color: var(--wp--custom--color--tertiary) !important;
+  background-color: var(--wp--custom--color--tertiary);
 }
 
 .has-primary-color {
-  color: var(--wp--custom--color--primary) !important;
+  color: var(--wp--custom--color--primary);
 }
 
 .has-secondary-color {
-  color: var(--wp--custom--color--secondary) !important;
+  color: var(--wp--custom--color--secondary);
 }
 
 .has-foreground-color {
-  color: var(--wp--custom--color--foreground) !important;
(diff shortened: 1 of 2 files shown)


mikachan (approved):
These changes fix the issue described in https://github.com/Automattic/themes/issues/6963 for me, and I agree with this reasoning:

> Lowering the specificity of these rules allows the correct color presets generated from the style variation to be used, as expected.

LGTM 👍

PR #8584, opened 2025-01-16
Outnow: Add Inter Tight font

henriqueiamarino:
This PR removes Urbanist from Heading and adds Inter Tight as intended. 
The Homepage QL were set to display default posts, with offset settings only.


iamtakashi:
@henriqueiamarino Thanks for the update! Though, I can only see typographic changes in this PR at the moment. https://github.com/Automattic/themes/pull/8584/files

henriqueiamarino:
Oh, okay, @iamtakashi. So, the QL changes were already added to the last update and, therefore, were not committed as new changes. I had a note on my desk about it; I probably forgot to check it as updated. Anyway, this is now free to go once it shows only Inter Tight on the default style variation.

iamtakashi:
Okay, it looks good. I'll merge it, bump the version, and deploy the theme to dotcom (I want to practice it.)

PR #8585, opened 2025-01-16
Varia: Strengthen button CSS

### LINKED ISSUE #2327: Hever: Homepage button color not updating properly
## Steps to replicate

Video: https://share.getcloudapp.com/GGuRvJDP

- Create a new site on WP.com
- Select the Hever theme
- Edit the homepage and try to change the button colors.
- If you add a new button and update the colors as well, it is not reflected correctly but it displays correctly in the editor

## Result

Button color was not reflected correctly

## Expected

Button color should be reflected as displayed on the editor.

* User report: [#22118057-hc](https://mc.a8c.com/support-stats/happychat/transcript.php?id=22118057)

### LINKED ISSUE #7309: Varia children & Page Optimize: Button styling overwritten by inline styles
### Quick summary

Initially reported in 6652110-zd-a8c and p1691767299173069-slack-C03TY6J1A. User reported this change only began recently (possibly starting around August 7 or 8).

Buttons added to sites using Varia child themes appear to ignore some styles set in the editor in favor of default styling coming from inline CSS (`classic-theme-styles-inline-css`) when _Page Optimize_ is active (or specifically the `concatenate styles` setting is enabled). 

### Steps to reproduce

1. Activate a Varia child theme (e.g. Shawburn, also tested with Stratford).
2. Make sure Page Optimize is deactivated.
3. Add a Buttons block with several buttons to a post or page. Test with different button styles and colors (using theme palette colors and custom colors).
4. View the post/page.
5. Activate Page Optimize and view the post/page again.

The buttons should look consistent with Page Optimize activated or deactivated. 

### What actually happened

With Page Optimize active, the buttons revert to a specific default style. E.g. on Shawburn, buttons have rounded corners (max border radius), gray background, and white text unless those settings are set to _custom_ values in the editor. (Preset/theme palette colors are overridden.)

Shawburn theme with Page Optimize deactivated:

Shawburn theme with Page Optimize active:

### Browser

_No response_

### Context

6652110-zd-a8c

Atomic

### Other notes

_No response_

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

Yes, easy to implement

### Workaround details

Disabling the `concatenate styles` option in Page Optimize's settings or disabling Page Optimize entirely resolves the issue.

scruffian:
This stengthens the button CSS in Varia based themes so that when the CSS is concatenated by the page-optimize plugin, the theme CSS still wins.

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/7309
Fixes #2327

Before:
Screenshot 2025-01-16 at 13 00 57

After:
Screenshot 2025-01-16 at 13 00 27

To test:
1. Install the page optimize plugin - https://github.com/Automattic/page-optimize/
2. Activate the plugin and confirm that button styles are now grey
3. Checkout this PR and confirm that buttons appear as expected again

### DIFF
--- a/alves/style-rtl.css
+++ b/alves/style-rtl.css
@@ -198,7 +198,8 @@ $grid-configuration: map-extend($grid-configuration-default, $grid-configuration
 button[data-load-more-btn],
 .button, button,
 input[type="submit"],
-.wp-block-button__link,
+a.wp-block-button__link,
+button.wp-block-button__link,
 .wp-block-file__button, .a8c-posts-list__view-all, .wp-block-search .wp-block-search__button, body .widget_eu_cookie_law_widget #eu-cookie-law input.accept {
 	line-height: 1;
 	color: #ffffff;
@@ -217,11 +218,13 @@ input[type="submit"],
 button[data-load-more-btn]:before,
 .button:before, button:before,
 input[type="submit"]:before,
-.wp-block-button__link:before,
+a.wp-block-button__link:before,
+button.wp-block-button__link:before,
 .wp-block-file__button:before, .a8c-posts-list__view-all:before, .wp-block-search .wp-block-search__button:before, body .widget_eu_cookie_law_widget #eu-cookie-law input.accept:before, button[data-load-more-btn]:after,
 .button:after, button:after,
 input[type="submit"]:after,
-.wp-block-button__link:after,
+a.wp-block-button__link:after,
+button.wp-block-button__link:after,
 .wp-block-file__button:after, .a8c-posts-list__view-all:after, .wp-block-search .wp-block-search__button:after, body .widget_eu_cookie_law_widget #eu-cookie-law input.accept:after {
 	content: '';
 	display: block;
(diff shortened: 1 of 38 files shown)


matiasbenedetto (approved):
The change looks good technically and can fix the issue, but changing the specificity of a selector could create visual regressions in custom CSS written to extend these themes. I think we should usually try to avoid this type of change, but given that there's already a known issue with those styles (https://github.com/Automattic/themes/issues/7309) affecting multiple sites, I think the trade-off could be acceptable.

Another option would be to avoid rendering the styles conflicting with the theme ones. 


```php
add_action( 'wp_enqueue_scripts', function() {
    wp_dequeue_style( 'classic-theme-styles' );
}, 20 );
```

It would avoid rendering this piece of CSS (including the conflicting CSS) that's probably unnecessary.
```
<style id="classic-theme-styles-inline-css">
/*! This file is auto-generated */
.wp-block-button__link{color:#fff;background-color:#32373c;border-radius:9999px;box-shadow:none;text-decoration:none;padding:calc(.667em + 2px) calc(1.333em + 2px);font-size:1.125em}.wp-block-file__button{background:#32373c;color:#fff;text-decoration:none}
</style>
```

That would require more testing though so probably the option added in this PR is good enough to ship.

PR #8588, opened 2025-01-16
Varia: Fix Jetpack subscription button

### LINKED ISSUE #6785: [Bug]: Hever - Subscribe block not styling correctly
### Quick summary

On Hever theme, when a user applies a custom style to the subscribe block, the form is displayed correctly in the Block Editor but it isn't on the preview or published page.

Specifically, it's like if alignment, height and border color preferences are not taken into account — even if they're being saved (and visible) in the Editor.

This has been reported on a support ticket (not sure if and what details I'm allowed to share here), but here are a couple of screenshots I took from our customer's website and on a test website of mine too. Hope it helps to clarify.

Our customer's website, Hever theme: Block Editor view
Customer Website - Block Editor View

Our customer's website, Hever theme: live page view
Customer Website - Page View

As you can see, the form is misaligned (should be center-aligned but it's left-aligned) and color preferences are not displayed. The button's `border-color` is set to `transparent` on the live page, but it should be the same as the one displayed in Block Editor.

My test website, Hever theme: Block Editor view
Test Website - Block Editor View

My test website, Hever theme: live page view
Test Website - Page View

Then I switched to Twenty Twentythree, updated the colors and looks ok.

My test website, Twenty Twentythree theme: live page view
Test Website with Twenty Twentythree Theme - Page View

You can see that the form fields height is correct, the form is center-aligned and the button border color matches the text field one.

Please note that this doesn't depend on any plugin or custom CSS preferences as the user is on a Personal plan and I'm trying it on an empty website created from scratch.

### Steps to reproduce

1. Create a website and install Hever theme.
2. On any page, add the Subscribe block and update some of the styling preferences.
3. Publish the changes and compare the editor view with the live page.

Styling preferences should be applied to the Subscribe block so that it matches what I see in the Block Editor view. Form shouldn't be misaligned, borders color should match my preference and the button height should match the text field.

### What actually happened

Only a few styling settings are applied to the Subscribe block form. Its look on the published page doesn't match the one in Block Editor.

### Browser

Google Chrome/Chromium, Microsoft Edge, Apple Safari, iOS Safari

### Context

Customer report via support ticket

Simple, Atomic

### Other notes

_No response_

### Reproducibility

Consistent

### Severity

Most (> 50%)

### Available workarounds?

Yes, easy to implement

### Workaround details

All of the above mentioned issues are cosmetic, so they can be easily fixed via CSS. However, each time a user decides to change a color from the Block Editor, the custom CSS code should be updated accordingly. Moreover, this issue affects Free and Personal plans too so it's not possible to add custom CSS via the Customizer there.

scruffian:
Override the :: before and ::after selectors for the button in the Jetpack subscription block. This CSS is added to all buttons in Varia themes. I'm not sure why it's there, but removing it changes a lot of buttons. Instead this small change only targets buttons in the Jetpack subscription block, so its scope is much smaller.

Before - notice that the button isn't inline with the input
Screenshot 2025-01-16 at 15 35 43

After: - notice that the button is now inline with the input
Screenshot 2025-01-16 at 15 35 17

#### Related issue(s):
Fixes https://github.com/Automattic/themes/issues/6785

### DIFF
--- a/alves/style-rtl.css
+++ b/alves/style-rtl.css
@@ -2299,6 +2299,11 @@ hr.wp-block-separator.is-style-dots:before {
 	width: 100% !important;
 }
 
+.wp-block-jetpack-subscriptions button::before,
+.wp-block-jetpack-subscriptions button::after {
+	display: none;
+}
+
 table,
 .wp-block-table {
 	width: 100%;
--- a/alves/style.css
+++ b/alves/style.css
@@ -2299,6 +2299,11 @@ hr.wp-block-separator.is-style-dots:before {
 	width: 100% !important;
 }
 
+.wp-block-jetpack-subscriptions button::before,
+.wp-block-jetpack-subscriptions button::after {
+	display: none;
+}
+
 table,
 .wp-block-table {
 	width: 100%;
--- a/balasana/style-rtl.css
+++ b/balasana/style-rtl.css
@@ -2299,6 +2299,11 @@ hr.wp-block-separator.is-style-dots:before {
 	width: 100% !important;
 }
 
+.wp-block-jetpack-subscriptions button::before,
+.wp-block-jetpack-subscriptions button::after {
+	display: none;
+}
+
 table,
 .wp-block-table {
 	width: 100%;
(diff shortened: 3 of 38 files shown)


mikachan (approved):
This tests well for me and I like the idea of the fix being specific to the Jetpack subscription block 👍

PR #8595, opened 2025-01-17
Zoologist: Correct color classes for foreground and secondary

### LINKED ISSUE #5787: Zoologist: Secondary color not being applied to header or paragraph block
### Quick summary

When applying the secondary theme color to a paragraph or header block, the color change is not visible either in the editor, in a draft preview, or on a published page. 

### Steps to reproduce

1. Activate the Zoologist theme
2. Set a custom secondary color from Customize > Colors
3. Create a new post or page and add a paragraph or header block with text
4. Change the text color to the secondary theme color

I expect this to change the color immediately in the editor and then be reflected on the published page or post

### What actually happened

The primary color is applied instead.

### Context

Reported in #4875035-zen.   I was able to test and reproduce.  I tested on Zoologist and Russel and only Zoologist exhibited the behavior.

Screen Capture on 2022-03-19 at 10-56-16
Markup 2022-03-19 at 11 04 38

### Simple, Atomic or both?

Simple

### Theme-specific issue?

Zoologist

_No response_

### Reproducibility

Consistent

### Severity

_No response_

### Available workarounds?

_No response_

### Workaround details

This CSS seems to work:

/* Temporary fix for secondary color bug */
.has-secondary-color {
  color: var(--wp--preset--color--secondary) !important;
}

pkevan:
Fixes #5787.

### DIFF
--- a/zoologist/theme.json
+++ b/zoologist/theme.json
@@ -78,10 +78,10 @@
 				}
 			},
 			"color": {
-				"foreground": "var(--wp--preset--color--primary)",
+				"foreground": "var(--wp--preset--color--foreground)",
 				"background": "var(--wp--preset--color--background)",
 				"primary": "var(--wp--preset--color--primary)",
-				"secondary": "var(--wp--preset--color--primary)",
+				"secondary": "var(--wp--preset--color--secondary)",
 				"tertiary": "var(--wp--preset--color--tertiary)"
 			},
 			"form": {


zoologist/theme.json
```diff
@@ -78,10 +78,10 @@
 				}
 			},
 			"color": {
-				"foreground": "var(--wp--preset--color--primary)",
+				"foreground": "var(--wp--preset--color--foreground)",
 				"background": "var(--wp--preset--color--background)",
 				"primary": "var(--wp--preset--color--primary)",
-				"secondary": "var(--wp--preset--color--primary)",
+				"secondary": "var(--wp--preset--color--secondary)",
```
MaggieCabrera, line 84: I think quadrat and geologist needs the same change

pkevan: Seems like there are a few of these - i'll go through and create some PRs to fix the ones I see.

Issue #6611, opened 2022-09-22
Illustratr, Toujours, other themes: No margins between Image blocks on live site


thekingsprojects:
### Quick summary

On Illustratr and several other themes, when a page or post has two or more Image blocks one above another, there is normally a space between those images in the block editor. On a live site (and assuming the Image blocks do not have captions), those margins disappear, and the images touch directly.

A short list of themes I found that are affected by this; I haven't checked all themes so there may be more. There doesn't seem to be a pattern other than that they're all Classic themes; no FSE themes that I tried were affected, and other Classic themes like Twenty Twelve and Stratford were also unaffected.

- Espied
- Illustratr
- Hemingway Rewritten
- Circa/Quadra
- Pictorico
- Isola
- Eighties
- Toujours
- Independent Publisher (added by @inaikem)

This issue was reported specifically for Toujours back in June: https://github.com/Automattic/wp-calypso/issues/64994

It was supposedly fixed by this Gutenberg merge: https://github.com/WordPress/gutenberg/pull/42005

But I also tried Toujours on my test site, and there are no margins between the images there, either.

This appears to be caused by a

```
figure {
  margin: 0;
}
```

coming from one of the inline stylesheets; deactivating the `margin: 0` in the inspector, or overriding it with CSS, restores the margins.

On my Atomic test site, I notice the "figure margin: 0" is already being overridden by this (tested with both Illustratr and Toujours):

```
.wp-block-image {
  margin: 0 0 1em;
}
```

So on Atomic the margins seem to work, while on Simple the images run together. Maybe the Gutenberg fix above only applied to Atomic sites for some reason, or has somehow been reverted for Simple?

### Steps to reproduce

1. On a Simple site, create and save a post that has two or more Image blocks directly above/below one another. Remove any captions from the images.
2. Switch the site's theme to Illustratr (or Toujours, or another theme listed above).
3. View the live post with the Image blocks.

### What you expected to happen

The images should display with margins between them, as they do in the editor:

![image-margins-error-editor](https://user-images.githubusercontent.com/82622339/191642763-949f5108-2c82-412c-986a-0e715e301e39.png)

### What actually happened

The images do not have any margins/space separating them.

<img width="1345" alt="image-margins-error-live" src="https://user-images.githubusercontent.com/82622339/191642980-f69f777f-5510-43bc-811f-afad092960cb.png">

### Context

Customer report in 5573795-zen

### Platform (Simple, Atomic, or both?)

Simple

### Reproducibility

[...]

PR #8594, opened 2025-01-17
Dara: add margin in block editor

pkevan:
Fixes #6965. 

Fixes #6611 on Dara theme.

PR #8598, opened 2025-01-20
Toujours: Fix margin top-bottom on frontend

cbravobernal:
#### Related issue(s):

Fixes this issue only in the Toujours theme.

 https://github.com/Automattic/themes/issues/6611

#### How to test it

Before the PR: 
- Use Toujours theme.
- Create or edit a page, add a couple of images on the editor.
- Check that there is a margin in the editor.
- Preview or publish the page, check that there are no margins.
Screenshot 2025-01-20 at 16 51 03

After the PR:

- Preview or publish the page, check that there are margins.
Screenshot 2025-01-20 at 16 51 34

### DIFF
--- a/toujours/header.php
+++ b/toujours/header.php
@@ -21,11 +21,6 @@
 </head>
 
 <body <?php body_class(); ?>>
-<?php
-if ( function_exists( 'wp_body_open' ) ) {
-	wp_body_open();
-}
-?>
 <div id="page" class="site">
 	<a class="skip-link screen-reader-text" href="#content"><?php esc_html_e( 'Skip to content', 'toujours' ); ?></a>
 
--- a/toujours/header.php
+++ b/toujours/header.php
@@ -21,6 +21,11 @@
 </head>
 
 <body <?php body_class(); ?>>
+<?php
+if ( function_exists( 'wp_body_open' ) ) {
+	wp_body_open();
+}
+?>
 <div id="page" class="site">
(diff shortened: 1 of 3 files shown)


toujours/header.php
```diff
@@ -21,6 +21,11 @@
 </head>
 
 <body <?php body_class(); ?>>
+<?php
```
MaggieCabrera: Is this change related to the PR?

toujours/inc/custom-header.php
```diff
@@ -43,9 +43,9 @@ function toujours_header_style() {
 	$header_text_color = get_header_textcolor();
 	/*
 	 * If no custom options for text are set, let's bail.
-	 * get_header_textcolor() options: Any hex value, 'blank' to hide text. Default: HEADER_TEXTCOLOR.
+	 * get_header_textcolor() options: add_theme_support( 'custom-header' ) is default, hide text (returns 'blank') or any hex value.
 	 */
-	if ( HEADER_TEXTCOLOR === $header_text_color ) {
+	if ( get_theme_support( 'custom-header', 'default-text-color' ) === $header_text_color ) {
 		return;
```
MaggieCabrera: Same question

pkevan:
I chatted with @mikachan yesterday, and we can ignore the theme-check errors, so probably best to have these adjustments removed from all the PRs to avoid extra noise.

toujours/header.php
cbravobernal: It is a theme-chek list fix. We can remove it.

pkevan (approved):
Approving as same as: https://github.com/Automattic/themes/pull/8594

PR #8606, opened 2025-01-21
Timestream: add theme

henriqueiamarino:
Timestream is a compact theme design that turns your blog into a visually captivating timeline. It is perfect for showcasing events, milestones, or historical content in a chronological flow.

[Demo site](https://timestream.mystagingwebsite.com/) | [Site content export](https://github.com/user-attachments/files/18492461/timestream.WordPress.2025-01-21.xml.zip)

screenshot


iamtakashi:
@henriqueiamarino, here are my notes. There are fewer issues to note in this theme. Good job, man!

* In the header template part, the site title is floating, and the negative margin makes a little awkward misalignment. Was it intentional?
CleanShot 2025-01-27 at 17 07 19@2x
* The header overall has room for improvement.
CleanShot 2025-01-27 at 17 24 22@2x CleanShot 2025-01-27 at 17 23 27@2x 
CleanShot 2025-01-27 at 17 26 17@2x
CleanShot 2025-01-27 at 17 26 51@2x
* It'd look neater if the alignments were tighter. 
Front of the site CleanShot 2025-01-27 at 17 17 19@2x 
Editor CleanShot 2025-01-27 at 18 36 03@2x
* Let's add a pagination and a no-results block with the usual helpful text to the archives, index, and search results templates.
* A link at the bottom of the query loop block on the blog home template, would provide a better continuity. 
CleanShot 2025-01-27 at 17 36 42@2x
* You can remove the system font registration from the theme.json as it's not used in the theme.
* Remove `rena-vf.woff2` from `/assets/fonts/neutral-sans/`.
* Let's fix the sizes for those sizes. 
CleanShot 2025-01-27 at 18 17 59@2x
* The footer template part has the credit that needs to be internationalized. Resaving the theme should fix it.
* Remove the `patterns/header.php`, `patterns/index.php` `patterns/page.php`
* The tree's image is actually on the blog home template, so we could say that instead of `Images for theme screenshot:` in the `readme.txt`.
* I've pointed out this a few times before, but we can't use different cases or two-word-with-a-space tags as theme tags. Remove the following, `Custom Colors`, `custom-menu`, `Custom Logo`, `Editor Style`, `Featured Images`, `Site Editor`, `Block Editor Patterns`, `Theme Options`, `Wide Blocks`, `Two Columns`, `Three Column`
* Instead, add `featured-images`, `wide-blocks`, `two-columns`, `three-columns`, `block-styles`, `full-site-editing`, and `style-variations`.
* It'd be good to run [the theme check plugin](https://en-gb.wordpress.org/plugins/theme-check/) before adding a theme with a PR to catch the problem with the theme tags above. 
CleanShot 2025-01-27 at 18 55 44@2x

henriqueiamarino:
Thanks, @iamtakashi.

- When I designed this, my mobile view was very different. Something was updated, which must have broken my layout idea. It's now fixed;
- Alignments throughout the Home were fixed. I found the issue and changed the Footer to suit the body;
- Removed the system font register and deleted the font standing in the wrong folder;
- Fixed the preset sizes;
- Remove the patterns/header.php, patterns/index.php patterns/page.php;
- Updated the readme.txt credit line;
- Update all the tags.

You're right about the Theme Check plugin. I forgot about it. It's been a long time since I launched a theme from scratch. As the plugin was not part of the backlog work, I hardly used it. Anyway, I'll use it from now on.

iamtakashi:
Thanks for the update! It looks like resaving the theme with CBT didn't make a footer pattern. I've edited the footer (nothing has changed; I just added and removed a period, so the editor thinks I've changed the footer template pattern) and resaved the theme with CBT to create the footer pattern. I also cleaned up the theme tag.

I think this is good to go. I'll merge this and launch it on dotcom.

PR #8615, opened 2025-01-22
Marl: Correct primary and secondary color classes

### LINKED ISSUE #8524: Marl: Primary and Secondary colors aren't used, causing confusion with Styles
### Quick summary

By default, Marl doesn't have a primary and secondary color. If you select a swatch with primary and secondary color and apply those to blocks they don't work as intended.

### Steps to reproduce

1. Activate Marl
2. In the Site Editor, go to Styles > Browse Styles and select the 3rd option (“rainforest”)
3. From the main Styles menu again, go to Colors > Edit palette, and confirm that the first color (“primary”) is green and the second (“secondary”) is dark blue.
4. Set the background on a Group block to the Primary color. Set text in the Group block to the Secondary color.
5. View the live site.

The correct colors should show up.

### What actually happened

Primary and secondary colors do not appear and show up as gray instead.

Image

### Impact

Some (< 50%)

### Available workarounds?

Yes, difficult to implement

Setting any other custom color will allow them to display properly. (if the primary color is #12665B , you can set a blocks color to #12665C it’ll be imperceptibly different from the actual primary color and actually render properly on the site.)

Simple

### Logs or notes

Tested on simple only so far.

pkevan:
Fixes #8524.

### DIFF
--- a/marl/theme.json
+++ b/marl/theme.json
@@ -27,8 +27,8 @@
             "color": {
                 "foreground": "var(--wp--preset--color--foreground)",
                 "background": "var(--wp--preset--color--background)",
-                "primary": "var(--wp--preset--color--foreground)",
-                "secondary": "var(--wp--preset--color--foreground)",
+                "primary": "var(--wp--preset--color--primary)",
+                "secondary": "var(--wp--preset--color--secondary)",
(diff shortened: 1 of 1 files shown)


cbravobernal (approved):
LGTM

Issue #8505, opened 2024-12-06
[DotCom Launch]: Brightblog


iamtakashi:
### DotCom launch checklist

- [x] Theme review
- [x] Create demo site
- [x] Create theme showcase
- [x] Network activate
- [x] Add theme to the curated list
- [x] Verify theme in [showcase](https://wordpress.com/theme/brightblog)
- [x] Verify theme in [onboarding](https://wordpress.com/start)

iamtakashi:
@henriqueiamarino, here are my notes!
* Add a no-results block with helpful text to the archives, blog home, index, and search results templates.
* Can we reduce the side paddings on the main content area? It seems excessive for small screens, and it'd be great if it aligns with the header and the footer. This applies to the archives, 404, pages, search results, and posts templates.
![Image](https://github.com/user-attachments/assets/04f87eda-e02b-4b89-8f95-1c44c1b83475)  ![Image](https://github.com/user-attachments/assets/a9bc0db4-eee7-4e86-b70d-e7f9501eb2a9)
* For the blog home and the index templates, can we use a columns block instead of a row block? At the moment, the layout on small screens has room for improvement. Also, can we make the right-side padding for the content the same as the header?  ![Image](https://github.com/user-attachments/assets/ef48f521-3a11-415f-b46f-8eb7e27aa023)
* The featured image feels like floating without an anchor. It would be nicer if they aligned at the top.  ![Image](https://github.com/user-attachments/assets/1f4aa142-1885-47df-97c7-3c3bdcba1ec6)
* I'd recommend adding a comments section, at least in the single posts template.
* Let's fix the value of `size` for the custom font size presets in the theme.json.
* Remove `parts/header.html` and `parts/header-pages.html`.
* A `main` tag is missing in the 404, pages, and single posts templates.
* Let's add a title for each template part in the theme.json.
* Remove `patterns/about.php`, `patterns/comments.php`, `patterns/header-pages.php`, and `patterns/page.php`.
* Add `6.7` to `Tested up to:` in the `readme.txt` and `style.css`.
* Add `7.2` to `Requires PHP:` in the `readme.txt` and `style.css`.
* Remove the credit for Overused Grotesk.
* Change the version number to use the semantic versioning `1.0.0` in style.css
* Remove the following theme tags: `entertainment`, `block-patterns`, `custom-background`, `custom-colors`, `custom-logo`, `editor-style`, `post-formats`, `template-editing`, and `theme-options`.
[...]

PR #8623, opened 2025-01-22
Brightblog: essential fixes for Dotcom launch

henriqueiamarino:
On this commit, I fixed the following issues commented on #8505:

- Added the required No Results paragraph block;
- Reduced the overall layout paddings;
- Fixed the layout of _Blog Home_ and _Index_ by the use of columns;
- Fixed the values for the custom font size presets in the theme.json;
- Checked the **main** tags in all the templates;
- Listed the templates in the theme.json, including the new _comments_;
- Removed unused patterns: about, header-pages, page.php;
- Removed unused parts: header.html, header-pages.html;
- Updated Tested up to and Requires PHP as needed; 
- Removed incorrect tags listed;
- Removed duplicated comment from style.css;


iamtakashi:
Thanks for the update, @henriqueiamarino ! The changes look good. Merging!

PR #8045, opened 2024-08-28
Update Node/npm and dependencies

ajlende:
#### Changes proposed in this Pull Request:

Update node version and dependencies that support the new version. The node version bump was needed to support `URL.canParse()` in #8039, but I figured updating the dependencies would be nice while we're at it.

Node version is now 20 to match core.

Most of the major version bumps were just dropping support for node 18 and 19, so no action was needed for many of them. There were a few exceptions.

- `husky@9` [[changelog]](https://github.com/typicode/husky/releases/tag/v9.0.1): Required changes to `.husky/pre-commit` and `prepare` in `package.json`.
- `octokit@4` [[changelog]](https://github.com/octokit/octokit.js/releases): Dropped support for some features that we weren't using. Only used by `create-a-theme-checklist.js`, but it looked like no code changes were needed.
- `inquirer@10` [[changelog]](https://github.com/SBoudrias/Inquirer.js/releases/tag/inquirer%4010.0.0): Rewrite in TS, but was intended as a drop-in replacement. Used by some release workflows in `theme-utils.mjs`. Seemed like no code changes needed.
- `@wordpress/scripts@27` was intentionally left at v27 for WP <=6.5 support. See the [v28 changelog](https://github.com/WordPress/gutenberg/blob/trunk/packages/scripts/CHANGELOG.md#2800-2024-05-31) for more info.
- `wp-prettier` had a different installation method than was used for the current version, so I followed those instructions to reinstall the latest version. See the [readme](https://www.npmjs.com/package/wp-prettier?activeTab=readme) for more info.

#### Related issue(s):

- #8039

PR #8624, opened 2025-01-22
Update deps and remove node-sass

mikachan:
This PR updates all the deps in this repo, and also replaces `node-sass` with `sass` in any themes that still use it. This is because I'm seeing lots of issues with `node-gyp` and `node-sass` depending on where I run this repo, and updating `node-sass` is probably long overdue.

I haven't included any updated CSS files, as I thought it would be easier to review this way, and I think we can update each individual CSS file as the themes are updated. I did test whether the CSS still compiled successfully by running `npm run build:style` for each theme that has been updated.

To test:

- Run `npm i` and ensure there are no errors
- cd into a theme directory that uses `sass`, e.g. `alves`, and run `npm run build:style`
- Ensure that the style.css file is compiled successfully. You will probably see changes in the way its compiled, but I think we can commit these changes each time a theme's styles need to be updated. 

#### Related issue(s):
https://github.com/Automattic/themes/pull/8045


MaggieCabrera:
Thank you for the ticket instructions! It works fine for me

MaggieCabrera (approved):
🚢

scruffian (approved):
Looks good to me. I think we should update the CSS in a subsequent PR so that people don't see unexpected changes in the future.

Issue #8506, opened 2024-12-06
[DotCom Launch]: WP-DOS


iamtakashi:
### DotCom launch checklist

- [x] Theme review
- [x] Create demo site
- [x] Create theme showcase
- [x] Network activate
- [x] Add theme to the curated list
- [x] Verify theme in [showcase](https://wordpress.com/theme/wp-dos)
- [x] Verify theme in [onboarding](https://wordpress.com/start)

iamtakashi:
@henriqueiamarino, here are my notes!

* How about not using the columns block there so that the navigation extends to the right freely? 

![Image](https://github.com/user-attachments/assets/bad0ae30-33d6-4ace-b799-a547ff486a78)

* What do you think about making the pagination as wide as it can get? At the moment, it feels like floating a bit. 

![Image](https://github.com/user-attachments/assets/15339131-6875-44d0-a3a6-fd1dbd980524)

* Add a no-results block with helpful text to the archives, blog home, index, and search results templates.
* For small screens, it's probably better not to specify the width of the search field in the 404 and the search results templates. Either use a colmuns block or a group block and let it extend to the content width. 

![Image](https://github.com/user-attachments/assets/e8daeb3b-fdb8-4855-968d-82d4b08cbfaa)
* The colour of the overlay is different in the blog home template. The background appears darker. Was it intentional? 

![Image](https://github.com/user-attachments/assets/34920e14-421a-4830-8465-4ba15ce6cce1)

* The content of the single posts template doesn't flow well—the post meta comes after the comments. 

![Image](https://github.com/user-attachments/assets/45c9c5f3-9e3b-4c33-af8e-b8bd3cf42d36)

How about having another four columns block underneath the current one and placing the comments section in it? 

![Image](https://github.com/user-attachments/assets/e0034f85-48b7-4687-af05-aa7340487baf)

* Remove bundled Overused Grotesk.
* I don't think we need the font registration for the `system-font` in `theme.json`.
* Let's fix the size value for the custom font size presets in the theme.json.
* Let's use the JetBrains Mono for the code block, too.
* The header and the footer template parts have text strings that need to be internationalized. Resaving the theme should make the patterns and call it from the template parts.
* The theme currently doesn't use the post-meta template part, but the user will benefit if the theme uses it since the post-meta is quite elaborate and appears on many different templates. It will simplify the experience around customisation and the theme files.
* Let's add a title for each template part in the theme.json.
[...]

PR #8626, opened 2025-01-23
WP-DOS: essential fixes for Dotcom launch

henriqueiamarino:
On this commit, I fixed the following issues commented on #8506:

- Removed the navigation and pagination blocks from the column structure and made both wide;
- Added the required No Results paragraph block;
- Fixed the Search block, removing the width spec;
- Adjusted the overlay in all cover elements, making the background one for all templates;
- Made _Comments_ a template and moved it to a new column;
- Removed bundled Overused Grotesk and system-font;
- Fixed the values for the custom font size presets in the theme.json;
- Fixed Jetbrains font presets;
- Listed the templates in the theme.json, including the new comments;
- Hopefully, after the CBT save Header, Footer, and the 404 template are now internationalized; 
- Used _Post Meta_ correctly on the single post template, as required;
- Fixed the main tags in all the templates;
- Removed unused patterns: _hidden-no-results-content.php_. The other, _comments.php_ is now being use;
- Updated info: _Tested up to_ and _Requires PHP_ as needed;
- Changed the version number;
- Removed incorrect tags listed.


iamtakashi:
@henriqueiamarino Thanks for the updates! As we discussed, I pushed [a change](https://github.com/Automattic/themes/pull/8626/commits/926109c3640785819ee7d578cbfb46dabd68cfe8) with a new post meta template part (the "+" sign inside the template part doesn't need to be internationalised), and changed the archives template just to show you how you could do the same to other templates. Could you do the same for the templates could share the post meta template part? (if you don't have time, I don't mind doing that myself at all.)

henriqueiamarino:
Thanks, @iamtakashi. I did this by copying the whole column structure into the Query Loop. I used _Archive_ as the source of truth and repeated it on the _Index_ and _Search Results_ templates. Now that I see the _Blog Home_ should also have one, I'll commit to a final fix.

iamtakashi:
Nice one. This is solid. Merging!

PR #8631, opened 2025-01-24
Updates the essential fixes for Dotcom launch

henriqueiamarino:
- Removed the negative margin from the post excerpt;
- I checked the other columns for discrepancies, and they all appear aligned right now.

Screenshot 2025-01-24 at 20 01 05


iamtakashi:
I made a separate PR for the negative margin here: https://github.com/Automattic/themes/pull/8632

Issue #8500, opened 2024-12-06
[DotCom Launch]: Vitrum


iamtakashi:
### DotCom launch checklist

- [x] Create [demo site](https://vitrumdemo.wordpress.com/)
- [x] Create [theme showcase](https://wordpress.com/theme/vitrum)
- [x] Network activate
- [x] Add theme to the curated list
- [x] Verify theme in [showcase](https://wordpress.com/theme/vitrum)
- [x] Verify theme in [onboarding](https://wordpress.com/start)

iamtakashi:
@henriqueiamarino, I thought this is the best place to leave my notes to review before the dotcom launch. 

* The height difference is a bit awkward. Have a look at [this](https://github.com/Automattic/themes/blob/trunk/onyxpulse/theme.json#L248); if you subtract the thickness of the border from the padding, you can make both the same dimension. ![Image](https://github.com/user-attachments/assets/fa8563fc-7ce7-4580-b4b1-3e60bfc037b9)
* The `contentSize` is set to `900`, which doesn't make sense. The width for posts seems to be 1051px, which is probably the best default content width.
* In `templates/404.html`, `templates/archive.html`, `templates/index.html`, `templates/search.html`, `templates/single.html`, there are strings that are not internationalised. Re-save the theme with CBT so that they become patterns.
* Remove unused patterns `patterns/about.php`, and `patterns/hidden-404.php`
* Change `Tested up to:` to `6.7` in `readme.txt` and `style.css`
* Change `Requires PHP:` to `7.0` in `readme.txt` and `style.css`
* Let's review the image credits to remove unnecessary links. Since the theme doesn't bundle any image, let's add credits for the two images we see in the `screenshot.png`
* Let's change the version to `1.4.0` in `style.css`
* Remove the following theme tags, `entertainment`, `portfolio`, `block-patterns`, `custom-colors`, `custom-header`, `custom-menu`, `editor-style`, `full-width-template`, `template-editing`, `theme-options`
* Add `wide-blocks` tag
* This comment is duplicated in `style.css`
```
/*
 * Font smoothing.
 * This is a niche setting that will not be available via Global Styles.
 * https://github.com/WordPress/gutenberg/issues/35934
 */
```
* There are several uses of numeric font sizes in the theme. We could make `x-small` for the blocks using `0.8rem` and similar sizes. We should be able to use other presets existing in the theme for others with small differences.

henriqueiamarino:
[...]

PR #8637, opened 2025-01-28
Lexionic: add theme

henriqueiamarino:
Lexiconic is a sleek, versatile theme that empowers book authors, publishers, and storytellers. Its particular layout makes it easy to create stunning landing pages and offers a seamless way to showcase a piece of writing.

[Demo site](https://lexionic.mystagingwebsite.com/) | [Site content export](https://github.com/user-attachments/files/18574513/lexionic.wordpress.2025-01-10.xml_.zip)

screenshot


iamtakashi:
@henriqueiamarino Here are my notes. The number of things I need to mention is getting less and less. Great job!

* I've noticed that the custom spacing sizes defined in the `theme.json` aren't working. If the JSON file is version 3, the `defaultSpacingSizes` must be `false` to override the default spacing size. [Example](https://github.com/WordPress/twentytwentyfive/blob/trunk/theme.json#L58). I reviewed the theme as if the custom spacing sizes are actually working, but please fix it in the next update.
* Most of the custom duotones aren't used in the theme, do we need them?
* I like the unique-looking CTA in the footer, but it's not looking great on small screens. Can we take a different approach to make it work on both large and small screen sizes?
CleanShot 2025-01-29 at 14 42 11@2x
* It's not a problem, but why do the images on the blog home template use the cover blocks but not the image blocks?
* It's probably better to use a heading for the text "Homer". It's awkward in the variations that use a font with a strong character for headings. 
CleanShot 2025-01-29 at 15 17 44@2x
* A `main` tag is missing in all templates.
* Remove the following unused patterns. `patterns/single.php`, `patterns/page.php`, 
* All the style variations have custom duotones. Do we actually need them? If we do, it's probably better if they were registered in the default json.
* Change the `Requires PHP:` to `7.2` in both `readme.txt` and `style.css`
* Let's add credits to the rest of two images bundled with the theme.
* Thanks for paying attention to the theme tags! You can also add `block-styles` and `full-site-editing`.

henriqueiamarino:
Thanks @iamtakashi. Although the check has failed, I believe I fixed all the issues.
Here's the checklist:

- I tried to fix the custom spacing sizes issue by adding the line `defaultSpacingSizes: false`. I am not sure if this is enough. Please let me know;
- I removed all the duotones. They belong to an initial idea that is now gone. I did it on the style variations, as well;
- The footer CTA was repaired to look better. Thanks for checking it;
- I am using a `cover` instead of an _image_ to ensure the aspect ratio is set to 3:2. We can't control AR on images; we can only upload cut ones. But that would be a hack, and users wouldn't follow it. Do you think this is a problem?
- Made the author's text into a Heading;
- Added a main tag to all the `Content wrappers`;
- Removed the unused patterns: single and page.php;
- Changed the Required PHP to 7.2;
- Added the credit to the image that was missing;
- Added the extra tags you mentioned.

iamtakashi:
@henriqueiamarino, thanks for the update! Most of the issues were addressed correctly 👍 

* Thanks for removing the custom duotones, but it's still used in [here](https://github.com/Automattic/themes/pull/8637/files#diff-ba990d44b546e7c8c6b5f423e0ed9ad902fbcc982f7fbecba6e763fb5379e91cR11) and [here](https://github.com/Automattic/themes/pull/8637/files#diff-04c325b7a9520bf50e1dfbcd6dffd29998a4568d74ad24f209139d0bee6e54aaR100).
* The credit for the shadow image is missing.
* The height of the button is now different between the block variations. You have fixed the same issue [here](https://github.com/Automattic/themes/issues/8500#issuecomment-2555663994). However, this time, the border thickness is 2px. 
CleanShot 2025-01-30 at 12 18 31@2x

PR #8638, opened 2025-01-28
Global styles: allow override to work when Gutenberg plugin is active

### LINKED ISSUE #8636: Blockbase: style variation opt-out not always working
### Quick summary

When running the Gutenberg plugin within a site, the functionality to opt-out of style variations from a parent theme no longer functions as expected.

This is due to the `WP_REST_Global_Styles_Controller` now being `WP_REST_Global_Styles_Controller_Gutenberg` and therefore https://github.com/Automattic/themes/blob/trunk/blockbase/inc/rest-api.php#L25 is not longer triggered.

### Steps to reproduce

Themes which use `optOutOfParentStyleVariations` should not show parent theme style variations.

Possible to reproduce on `Simple` sites, and presumably `Atomic` but could not test this easily.

### Site owner impact

More than 60% of the total website/platform users

### Severity

Major

### What other impact(s) does this issue have?

_No response_

### If a workaround is available, please outline it here.

_No response_

Simple, Atomic

pkevan:
Fixes #8636

### DIFF
--- a/blockbase/inc/rest-api.php
+++ b/blockbase/inc/rest-api.php
@@ -17,12 +17,15 @@ function blockbase_remove_style_variations_from_child_themes( $response, $handle
 	$handler_method = isset( $handler['callback'][1] ) ? $handler['callback'][1] : null;
 	$opt_out        = wp_get_global_settings( array( 'custom', 'optOutOfParentStyleVariations' ) );
 
+	// Global styles are overridden when Gutenberg plugin is used.
+	$check_class = class_exists( 'WP_REST_Global_Styles_Controller_Gutenberg' ) ? 'WP_REST_Global_Styles_Controller_Gutenberg' : 'WP_REST_Global_Styles_Controller';
+
 	/*
 	 * Prevents Blockbase child themes from being considered child themes in the
 	 * `wp/v2/global-styles/themes/:theme/variations` API endpoint, so they don't
 	 * inherit the style variations from the parent theme.
 	 */
-	if ( $opt_out && is_a( $handler_class, 'WP_REST_Global_Styles_Controller' ) && 'get_theme_items' === $handler_method ) {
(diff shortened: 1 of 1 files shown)


merkushin:
@pkevan is it ready for testing and review?
Are there any testing instructions?

pkevan:
> @pkevan is it ready for testing and review?

Yes, it can be ready - I was going to attempt to see some numbers on how many users might be affected by fixing this, since they could be currently using the inherited color palettes.

> Are there any testing instructions?

It can be tested with any child theme of `blockbase` on an environment where Gutenberg plugin is activated - with the patch applied and the theme.json of the child theme updated to include `"optOutOfParentStyleVariations": "true",` in the `custom` section.

merkushin:
@pkevan 

Perhaps a silly question: how can I use it in my sandbox (a different repo is used there)?
Meanwhile, I'm looking for an answer in docs.

pkevan:
> @pkevan
> 
> Perhaps a silly question: how can I use it in my sandbox (a different repo is used there)? Meanwhile, I'm looking for an answer in docs.

No such thing as a silly question 😄 

Locally, the requests to the `variations` REST API endpoint on the same domain, but when testing via a `Simple` site, you would need to sandbox requests to `public-api` also.

MaggieCabrera:
I applied this change to my sandbox and tested Jackson and I still see the variations from Blockbase. Do we need to do anything else to test this? 

We will be archiving this repo soon (TM) in favor of the enterprise repo and I was trying to get this merged but it seems it's not there yet, or I'm missing out how to test it!

pkevan:
> I applied this change to my sandbox and tested Jackson and I still see the variations from Blockbase. Do we need to do anything else to test this?

Strange - did you sandbox public-api also?

It works as expected when I change to the `Jackson` theme when sandboxed for me.

MaggieCabrera (approved):
It was my public api not being sandboxed on FF :D this works as intended

pkevan:
Thanks @MaggieCabrera!

PR #8642, opened 2025-01-28
Kiosko: Add changelog.txt for woocommerce.com

iamtakashi:
Context: p9h1Lv-feI-p2

@mikachan @dsas Do you think adding a `changelog.txt` to this theme will cause any issues in Dotcom? I don't think we should add it to all the themes as it's redundant because a change log lives in the `readme.txt`, but currently, this is the only theme that needs the file.

We might need to update the scripts/workflow that updates the readme.txt so that it updates `changelog.txt` too if it exists.

### DIFF
--- a/kiosko/changelog.txt
+++ b/kiosko/changelog.txt
@@ -1,0 +1,4 @@
+== Changelog ==
+
+= 1.0.0 =
+* Initial release


dsas:
It shouldn't cause any issues. 

It's a bit of an annoying requirement. Given it's just kiosko (so far) does it make sense to just continue to manage this by hand for now rather than automating it? I'm not sure how/if changelog generation might change in the brave new one-repo world.

iamtakashi:
> Given it's just kiosko (so far) does it make sense to just continue to manage this by hand for now rather than automating it?

Yeah, let's continue editing this by hand for now. I doubt Kiosko gets many updates. But I wonder how best to remind us to update the `changelog.txt`.

How about adding this in the changelog in the `readme.txt`? I am hoping to see it :)

`[We must also update the changelog.txt ](https://github.com/Automattic/themes/tree/trunk/kiosko/changelog.txt)`

MaggieCabrera:
I'm bringing this in, let's open an issue on the other repo about this to potentially add a script that automates this step, it shouldn't be terribly hard, more a matter to scrape some time to do it

MaggieCabrera (approved):

Issue #8667, opened 2025-01-30
Theme Layover: Title Block Invisible In Editor


mxhassani:
### Quick summary

When creating a post or a page, the title (and its placeholder) in the title block isn't visible. 

![Image](https://github.com/user-attachments/assets/69b36e17-3bac-4192-8e13-9fffbd3ad5bf)

When selecting it:

![Image](https://github.com/user-attachments/assets/98a22c1f-d0b0-444c-8ba4-691b1ce47dbb)

This is caused by the fact it's using the same color as the background color as the editor.

### Steps to reproduce

1. Activate Layover
2. Create a new post
The usual location of the title is clickable and editable, but not visible. 
Also replicable on the demo site.

### Site owner impact

Fewer than 20% of the total website/platform users

### Severity

Minor

### What other impact(s) does this issue have?

No revenue impact

### If a workaround is available, please outline it here.

Enabling [Show template](https://wordpress.com/support/preview-your-content/#view-content-within-the-template) addresses it. 

The source of the issue is this CSS code:
`color: var(--wp--preset--color--foreground);` that's inline on the editor. 
It can be changed to `color: var(--wp--preset--color--secondary);` for a quick fix.

### Platform (Simple and/or Atomic)

Simple, Atomic

mxhassani:
9353054-zen

fushar:
@iamtakashi / @henriqueiamarino could you look into this issue? I see this is a recently launched theme. Thanks!

iamtakashi:
@henriqueiamarino I think all the headings and the links should use `background` and override the colour of them to `foreground` where we need in the templates. Are you happy to make a PR?

henriqueiamarino:
Thanks @fushar and @iamtakashi. This one is tricky because it has a dark `Home` and a light `Single` template. Anyway, I'll think about it and find a solution.

supernovia:
📌 REPRODUCTION RESULTS
- Tested on Simple – Replicated
- Tested on Atomic – Replicated

📌 FINDINGS/SCREENSHOTS/VIDEO

<img width="729" alt="Image" src="https://github.com/user-attachments/assets/c0d2cc5f-695a-46f0-b39e-e63a8b91f8f9" />
<img width="752" alt="Image" src="https://github.com/user-attachments/assets/0442466b-cce7-4ea9-885a-3d413882626e" />


📌 ACTIONS
- Triaged

[...]

PR #8682, opened 2025-02-06
Layover: Fixes link color issue

henriqueiamarino:
This PR fixes the **Layover** theme by replacing the link colors.
The issue was previously mentioned and discussed in #8667.


iamtakashi:
Thanks for the update, @henriqueiamarino! LGTM👍 Merging.

